PR #5, opened 2015-04-22
Changes top bar of patient view for patients with multiple samples

inodb:
This PR does the following:
- Adds sample clinical attributes for multiple samples on a single line
- Change styling sightly
- Add extra tab with downloadable patient/sample table

An example with many patients can be seen on the dev server at ```patient-view-topbar/case.do?cancer_study_id=prad_fhcrc&case_id=05-144``` (not sure if it's alright to post the full link publicly).

Problems I ran into:
- [x] What to do when there is no clinical data in the database? That is actually an error on the data level but not sure if that should be handled anywhere. Tips?
- [x] In single sample view the more about this tumor button doesn't work anymore


inodb:
Fixed the bugs I found. Haven't found new ones yet.

inodb:
Doing this PR to master branch now where new features should be introduced first

PR #31, opened 2015-05-22
Collapse and expand samples in patientview topbar

inodb:
Current implementation shows drop down mouseover to show all samples in case
there are too many to fit in the topbar. This feature changes the topbar to
expand/collapse the samples on click with a more/less button.

topbar_collapse_expand


ecerami:
wow.  +1 for embedding a video describing the feature, and another +1 for making it a pull request.  awesome!

inodb:
Thanks!

PR #51, opened 2015-06-02
Fix issue #50.

jjgao:


jjgao:
Title was wrong. Function is correct :)

PR #68, opened 2015-06-04
Import timeline files without stop date column

inodb:
Change timeline importer to take input files without stop date column.


inodb:
Should merge this to hotfixes now

PR #74, opened 2015-06-09
Arman graduates and becomes an alumnus

armish:
Thanks for all the fish, team :hand:


ecerami:
nooooooo.  there's no leaving the portal, though, as I have now discovered.  we will see you again in a few years :-)

armish:
Still find myself referring the team as us and puzzling people about my affiliations, but yeah: once in, never out ;)

PR #103, opened 2015-07-07
Fix unit tests

morungos:
This contains the bulk of the unit testing logic, although it isn't completely ready for merging yet. The main issue I've found is in the extended mutation parsing, which isn't picking up most of the information from the test file I have. I'm guessing these are out of date with respect to the input expectations. This also needs the Spring static injection magic added to the repository and security context files, as we've switched out the dependency pull from explicit context use to dependency injection -- this primarily affects JdbcUtil and SpringUtil, and the added test context shows the pattern needed.


jjgao:
@morungos could you please use cBioPortal:rc branch as the target branch for the pull request? We are using master branch for released code only.

morungos:
Closed this. It isn't possible to change the target, so I just made a new one that targets cBioPortal:rc

PR #104, opened 2015-07-07
Fix unit tests

morungos:
This contains the bulk of the unit testing logic, although it isn't completely ready for merging yet. The main issue I've found is in the extended mutation parsing, which isn't picking up most of the information from the test file I have. I'm guessing these are out of date with respect to the input expectations. This also needs the Spring static injection magic added to the repository and security context files, as we've switched out the dependency pull from explicit context use to dependency injection -- this primarily affects JdbcUtil and SpringUtil, and the added test context shows the pattern needed.

Created again to target the rc branch.


serverhorror:
Hi,

would it be possible to separate 5931ac3b0185d43e68bfe4fbd3 out of this and provide a migration script so that we can finally get rid of the dreaded MyISAM tables?

Even better would be if it would be supporting Postgresql :)

morungos:
I'd agree on both, but Postgres might be a *much* bigger task. 

On another project (https://github.com/pughlab/tracker) I've used Querydsl SQL for most of my SQL query generation. I used this because our IT people start to look worried when they see JDBC calls with generated SQL, and the more I use it, the more I like it. But I think the MyBatis used in the business component might be a better way forward as some work on that already exists. 

HSQLDB is also worth putting into the mix, as then the unit and integration tests can run without server configuration, which makes testing and easy deployment very much simpler.

What I would suggest we might consider is a move to ANSI SQL quotes in the meantime. I actually used this for my seed data, MySQL supports it with a simple URL/session option, and it covers about 90% of the work porting to Postgres/HSQLDB without needing to be done in one go. Basically, that means changing backticks to double quotes and setting a session option on the URLs. 

So we can park full database portability as a future issue to discuss, and I'll split out the InnoDB as a first PR (the tests depend on the transactions enabled by that). We do have a migration script, I'll dig that out to include.

serverhorror:
The postgres thing was just wishful thinking. I wouldn't have expected it to really happen.

I think that supporting multiple RDBMS is something that will take time. Yes having the option to run tests with some kind of in memory DB would be perfect (also for small test installations), still if cBioPortal supports multiple databases that will be a huge task. Especially since there is basically no relational integrity in the database as of now (MyISAM + `SET FOREIGN_KEY_CHECKS=0;` in cgds.sql), I guess I have seen to many long nighters because of the MyISAM and it's problems :)

n1zea144:
Stuart,

Can I get access to the pughlab repository and this branch?  I'd like to pull it into my local repository for testing before merging into the rc branch.

B

n1zea144:
Never mind Stuart, I was able to find the pughlab repository by following through your ID (a search on Github did not show me the pughlab repos).

morungos:
Oh yes, the joys of Github not indexing forks. At all. :-(

n1zea144:
So I merged this repos/branch into the master branch of my forked version of the repos.  I'm getting the following stats after the merge:

Tests run: 206, Failures: 14, Errors: 17, Skipped: 6

I've imported the new schema into my cgds_test db and added db credentials into my portal.properties is there something else I should be setting?

There is one other thing that I can see will need to be done for completeness - business/src/main/resources/applicationContext-business.xml will need to be updated to inject the data source into JdbcUtil.java via org.springframework.beans.factory.config.MethodInvokingFactoryBean.  One call should be added at the section at line 38 (used by our importer tool) and the other at line 53 ( when the code is running within tomcat).  Did you intentionally leave this out?

morungos:
Hmmm, that's much worse than I was getting. I'll have a rummage though. I really don't like the errors, so I'd aim to avoid them entirely.

n1zea144:
In case it may be helpful, I just emailed you my maven output.

morungos:
Nothing's shown up in email sadly. If you emailed to uhnresearch.ca and zipped it or anything sensible, likely our mail portal silently erased it. Maybe try morungos at gmail.com?

n1zea144:
Sent.

morungos:
Thanks, got em. Quick hypothesis: are you still running with default tables in MyISAM? I know writing a script to do that was part of our list, but we haven't finalized that yet. What I suspect is: MyISAM tables aren't really transactional, so no rollback happens and then new data will conflict when something similar is inserted. 

Now I re-read 5931ac3, I removed the MyISAM but didn't explicitly set to InnoDB, mainly because I stupidly overplanned and wanted something that might support other DBs some day. If the MySQL is set to default to MyISAM it's probably back where it started, and a quick fix to the schema file should make everything very much happier.

n1zea144:
Isn't that what the following line in core/pom.xml is suppose to do:

<sqlCommand>SET storage_engine=INNODB</sqlCommand>

morungos:
You're right: that's exactly what it's for and now I'm puzzled as they should be. Can you check in the DB and see if they really are InnoDB? `SHOW CREATE TABLE cancer_study;` should do it.

morungos:
I've tried almost everything I can think of that might cause this. Can you confirm whether the tables in  cgds_test are using InnoDB or MyISAM? 

I've replicated it by setting the server to disable InnoDB entirely. Then you get MyISAM tables even if we do select the default storage engine, and we're back into a nontransactional mode. Check the `my.cnf` settings for things that might disable InnoDB, although (IIRC) it can also be dropped during compilation. 

Interestingly, on some OSes (e.g., CentOS), InnoDB does seem to be disabled by default!! http://serverfault.com/questions/414642/mysql-my-cnf-shows-innodb-as-disabled-skip-innodb-should-i-enable-it-for-larg

n1zea144:
The tables are using MyISAM.  Thanks for the tip.  We are running CentOS here.  I'll have our sysadmin take a look.

n1zea144:
Merged #138 which included these changes.

PR #121, opened 2015-07-17
Warning Icon & Tooltip for the Protein Change column

onursumer:
Adds a warning icon and a tooltip where the annotated protein change value is different than the original value on the MAF file. For now, this feature is only enabled for MSK-IMPACT cancer study.


jjgao:
close it now -- code has merged to rc branch.

PR #126, opened 2015-07-24
removed reference to proprietary oncokb server

j-hudecek:


j-hudecek:
When oncokb URL is non null potentially sensitive data (sample ids, patient ids, mutations) are sent to it. It shouldn't be present in the default properties file

PR #159, opened 2015-08-07
trivial change to test travis-ci

jjgao:


ecerami:
Great experiment!  And, interesting that it also failed.  Must be something specific to the way we are building pull-requests on travis. pulling in @cataphract

Issue #153, opened 2015-08-03
Home page: Treemap stops working for D3 v3.3.8


inodb:
In the release notes from D3 v3.3.8:
```
- Fix stale children array when recomputing a hierarchical layout.
```
This removes empty children arrays from the nodes and results that in
``./portal/target/portal/js/src/d3.right-menu-stats.js`` the following code doesn't work anymore:
```
var nodes = treemap.nodes(root)
    .filter(function(d) { return !d.children; });
```

PR #164, opened 2015-08-12
Update timeline to use cbioportal/clinical-timeline

inodb:
Changes the timeline to use cbioportal/clinical-timeline@f84d34c6e5c4f54f6bd8aa078ff0db99aa24627a that was developed in collaboration with Mellinghoff et al.

The new version gives more options for changing the timeline. One can split and color tracks by clinical attribute in the tooltip, hide certain tracks, add them again and collapse/stack the circles on a track. All these features can be performed by either calling methods on a timeline object or by the user through a tooltip on each track. Since the tooltip is not as visually appealing as the rest of the portal we've decided to disable user manipulation of the timeline for now and apply some default settings. That is always split the TREATMENT track on the AGENT attribute if it exists and order the tracks by:
```
["SPECIMEN", "SURGERY", "STATUS", "DIAGNOSTICS", "IMAGING", "TREATMENT"]
```
The other ones alphabetically. All tracks without duration data (i.e. all tracks with circles) are collapsed on default except the SPECIMEN track. This is because we want the 1,2,3 labels to appear on that track in case there is mutation data available. That wouldn't work if one circle represents multiple samples.

The original idea was to use jiahuang/d3-timeline@82622c4cc35bac7283b3a317826b0709ac1ae476 as a library, but since this project is currently not actively maintained it was not possible to get all the required features accepted. Therefore some small changes have been made to the d3-timeline and the source is included. Hopefully once that project gets more active again we can send a PR and use the d3-timeline as a library instead. Using the d3-timeline source code however required updating the D3 version to v3.5.6. This resulted in issue #153, which has a fix included in this PR.

The underscore version also had to be updated to v1.8.3 for some of the new data manipulation functions used in clinical-timeline. Didn't notice any regression in the rest of the portal.

What still needs to be done before merging:
- [x] Make the su2c data look good on default by sizing the circles based on their test result
- [x] Add a deceased dot to the timeline using clinical attributes from the patient
- [ ] Create a JAR for testing on the dev server
- [ ] Use displayName of clinical attributes for track names to make them more visually appealing


jjgao:
Thanks, Ino. Could you recreate the pull request pointing to rc branch?

inodb:
Sure!

PR #170, opened 2015-08-14
new webapi and datamanager

adamabeshouse:
Backend implementation of new web api
Frontend implementation of wrapper, and query data manager


n1zea144:
I think we need to break ApiService.java into separate service classes (you used to have it this way) per model class.  I suggest this because I'm starting to reimplement some Dao* code in mybatis and some of it is not relevant to our current web api, like DaoCancerStudy.addCancerStudy().  Lets discuss.

PR #263, opened 2015-09-24
add NA to genetic alteration tracks

adamabeshouse:
With this pull request, you can mark a datum for oncoprint as being unsequenced or otherwise having unavailable data by adding to that object the member "na" with any value ("true" may be a logical choice going forward)


leedonghn4:
yes. We worked on that future

adamabeshouse:
@leedonghn4: Right, this was my contribution to it. Let me know it works as expected.

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/274

PR #298, opened 2015-10-05
Fix oncoprint overcounting # patients

adamabeshouse:
..by handling edge case of split where the string is empty


jjgao:
@adamabeshouse: could you make another pull to rc? Thanks.

adamabeshouse:
@jjgao: should I still?

PR #322, opened 2015-10-08
Bug fix #311

### LINKED ISSUE #311: HTML tags in the breadcrumbs if a gene has Mutsig or Gistic result
image

http://www.cbioportal.org/beta/study.do?cancer_study_id=ucec_tcga_pub

aderidder:
Was caused by unexpected extra elements in the table cell
Solved by storing the geneId as an attribute and using that instead of qtip contents.


jjgao:
@aderidder: We merged some new code and it looks like we have some conflicts. Could you try to resolve? Thanks.

aderidder:
Hi @jjgao,
We've looked into it and should be solved now. Let me know if there are more problems.

jjgao:
Issue fixed.

Issue #216, opened 2015-09-04
support patient id in the custom case set input box


jjgao:

jjgao:
Currently only sample IDs are supported. 

![image](https://cloud.githubusercontent.com/assets/840895/10232296/dc2103a0-6855-11e5-906b-199713cc155d.png)

We should add an option to support sample IDs too, similar to what we do in the study view select box.

![image](https://cloud.githubusercontent.com/assets/840895/10232328/10934738-6856-11e5-815d-4b9fc69df713.png)

jjgao:
@zheins: please start from the rc branch.

zheins:
Pull request #323

PR #323, opened 2015-10-08
Support of patient ID in custom case set input box. Issue #216

zheins:
Also adds ability to get samples based on patient ID and case ID in the SampleMapper and api service and controller classes (ibatis).

Signed-off-by: Zack Heins <zackheins@gmail.com>


jjgao:
Thanks, @zheins.

Is it possible to move the mapping code from QueryBuilder to frontend? In general, we wanted to remove code from QueryBuilder, and maybe one day we can remove the QueryBuilder class :)

@adamabeshouse: How does oncoprint switch between samples/patients? Is the patient-sample mapping already in session manager?

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -290,4 +290,17 @@
 	public List<DBStudy> getStudies(List<String> study_ids) {
 		return studyMapper.getStudies(study_ids);
 	}
+        
+        @Transactional
+        public List<DBSample> getSamplesByStudyAndPatientId(String study_id,String patient_id) {
```
adamabeshouse: I'm not sure if we need this

web/src/main/java/org/mskcc/cbio/portal/web/ApiController.java
```diff
@@ -146,14 +146,19 @@
...
     @RequestMapping(value = "/samples", method = {RequestMethod.GET, RequestMethod.POST})
-    public @ResponseBody List<DBSample> getSamples(@RequestParam(required = true) String study_id, @RequestParam(required = false) List<String> sample_ids) {
+    public @ResponseBody List<DBSample> getSamples(@RequestParam(required = true) String study_id, @RequestParam(required = false) List<String> sample_ids, @RequestParam(required = false) List<String> patient_ids, @RequestParam(required = false) String patient_id) {
 	    if (sample_ids != null) {
 		    return service.getSamples(study_id, sample_ids);
-	    } else {
+	    } else if (patient_ids != null) {
+                    return service.getSamplesByStudyAndPatientId(study_id,patient_ids);
+            } else if (patient_id != null) {
+                    return service.getSamplesByStudyAndPatientId(study_id,patient_id);
```
adamabeshouse: Do we really need this? A GET request will look exactly the same (?patient_ids=id1  vs ?patient_id=id1) and for POST you'd only need to wrap it in brackets, not a huge deal. Is there another reason to keep this?

adamabeshouse:
@jjgao the patient-sample mapping is in the setupOncoprint code at the moment

PR #335, opened 2015-10-14
Make promises fail if anything goes wrong

adamabeshouse:


pieterlukasse:
Now that I can catch a failed promise, how would you like me to present the error on the screen? Should I use the yellow error banner that is used in other situations?

PR #338, opened 2015-10-15
enrichment analysis improvements

yichaoS:
possible solutions: 

1. disable gene list as SQL input
2. multi threading for calculation


jjgao:
It is much better. Thanks. 

Mutation: 12s
CNA: 24s
mRNA: 80s
RPPA: <1s

I'll also review the code and see if we can improve it further.

PR #371, opened 2015-10-28
Duplicate columns removed in corrected files.

zheins:
Signed-off-by: Zack Heins <zackheins@gmail.com>


zheins:
Closing - need to determine which column to remove. Manual curation.

PR #382, opened 2015-10-29
code refactor (enrichments)

yichaoS:
- Added interface in DaoGeneticAlteration class to process calculation on the fly 
- re-apply multi-threading


core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java
```diff
@@ -77,6 +78,15 @@ public static DaoGeneticAlteration getInstance() throws DaoException {
         return daoGeneticAlteration;
     }
 
+    public static interface AlterationProcesser {
+        ObjectNode process(
+            long entrezGeneId,
+            String[] values,
+            ArrayList<Integer> orderedSampleList,
+            HashMap mutHm
```
jjgao: What is mutHm?

yichaoS: mutation hashmap --> result from calling DaoMutation.getSimplifiedMutations, instead querying for database

PR #391, opened 2015-11-03
Bug fix of importUsers.py

zheins:


zheins:
Good catch @fedde-s. I just made the change and put in a pull request. Thanks!

PR #397, opened 2015-11-05
Fix for Active Directory NullPointerException

tlangs:
Just a simple edit to the ```applicationContext-security.xml``` file so Spring will initialize the accessControl bean properly. This is my first ever pull request (I'm still a student), so let me know if there's anything more I need to do.

### DIFF
--- a/portal/src/main/resources/applicationContext-security.xml
+++ b/portal/src/main/resources/applicationContext-security.xml
@@ -462,6 +462,15 @@
             <session-management session-fixation-protection="none"/>
         </http>
 
+        <b:bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
+            <b:property name="staticMethod" value="org.mskcc.cbio.portal.util.SpringUtil.setAccessControl"/>
+            <b:property name="arguments">
+                <b:list>
+                    <b:ref bean="accessControl"/>
+                </b:list>
+            </b:property>
+        </b:bean>
+
         <!-- Add Authentication Manager -->
         <authentication-manager>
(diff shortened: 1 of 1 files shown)


smowton:
I found this was necessary to use the AD auth module too; recommend merging this.

pieterlukasse:
I reviewed the code and this seems correct (even though my experience with Spring is limited and I have not tested it as such). Not setting `accessControl` in `org.mskcc.cbio.portal.util.SpringUtil` will indeed give NullPointerException in code such as `org.mskcc.cbio.portal.servlet.CancerStudyView.validate()`. 

One recommendation to this pull request would be to test whether you can move this part (which now seems to be present in all authentication "profiles" in the XML): 
```xml
	<b:bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
	    <b:property name="staticMethod" value="org.mskcc.cbio.portal.util.SpringUtil.setAccessControl"/>
	    <b:property name="arguments">
	        <b:list>
	            <b:ref bean="accessControl"/>
	        </b:list>
	   </b:property>
	</b:bean>
```
to the "shared" profile found at the start of the XML:
```xml
<b:beans profile="googleplus,openid,saml,ad">
```
This should help avoid unnecessary code duplication. 

@jjgao , @n1zea144: what do you think? Can we merge this one?

tlangs:
I somehow borked my fork. I've moved the bean to the shared profile and committed to a new branch. The pull request can be found [here](https://github.com/cBioPortal/cbioportal/pull/1067).

PR #398, opened 2015-11-05
get color from front of google colors array, not from back

adamabeshouse:


adamabeshouse:
@schultzn we already had the google colors in - the problem was we've been getting them from the end instead of from the beginning

Issue #510, opened 2015-11-17
Select Patient/Case Set items hidden if list too long


zheins:
![screen shot 2015-11-17 at 6 08 44 pm](https://cloud.githubusercontent.com/assets/10240498/11227843/9751754a-8d56-11e5-8913-87d5adfdd18c.png)

zheins:
PR #511

PR #370, opened 2015-10-28
White space fix

aderidder:
Two small style related changes, which so far seems to have fixed the whitespace issue

PR #511, opened 2015-11-18
Issue #510

zheins:
@zhx828 Will removing this overflow style be ok? I don't notice any problems taking it out.

### DIFF
--- a/portal/src/main/webapp/css/global_portal.css
+++ b/portal/src/main/webapp/css/global_portal.css
@@ -1123,7 +1123,6 @@ cursor: default;
     padding:10px;
     margin-top:6px;
     padding-top:10px;
-    overflow: hidden;
 }
 
 /* Main Query Form */


zhx828:
@zheins I think so. Only dynamicQuery is using this style, if you tested it already, it's ready to merge

jjgao:
It was related to PR https://github.com/cBioPortal/cbioportal/pull/370. I'd ask @aderidder to review it.

aderidder:
Hi,
I think you'll have to remove the overflow hidden from the tabs div of visualize.jsp as well. I did indeed add those overflow hiddens to solve the white-space problem, but the content of the dropdown now overflows the element's boundaries and as an effect is also hidden.
I removed both overflow hiddens on my local instance and it restored the dropdown functionality. Oddly, the white-space did not reappear; not sure why. If in your case it does reappear, maybe we could try to set the overflow hidden at a lower element, clear the floats some other way or add the dropdown at a different level.

zheins:
@aderidder I think the overflow style in visualize.jsp was already removed. 

What was the original whitespace issue? I don't see a screenshot or good description of it in the previous merge.

ecerami:
@zheins:  in the original whitespace bug, the query results page had two screenfulls of blank, gray area.  So, the screenshot would be a big gray rectangle at the bottom of the page.

zheins:
I think for the moment, removing this styling makes sense. It's better to have whitespace than not be able to select items from the dropdowns.

aderidder:
Yes I agree. Also, the white-space bug was extremely small on the main page.

2015-11-18 17:39 GMT+01:00 Zachary Heins <notifications@github.com>:

> I think for the moment, removing this styling makes sense. It's better to
> have whitespace than not be able to select items from the dropdowns.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cBioPortal/cbioportal/pull/511#issuecomment-157771300>
> .
>



-- 
E.   sander@thehyve.nl
T.   +31(0)30 700 9713
W.  www.thehyve.nl

the hyve

PR #533, opened 2015-11-23
Fix bug cased by #364 fixing

zhx828:
In issue 364, we used component chartID instead of dc charID which
cased a issue in breadcrumb.


jjgao:
@aderidder could you help review this?

aderidder:
Hi I looked at it and also at #364 since I was curious about the mistake I made.
Seems ok to me!

Issue #229, opened 2015-09-11
Patient view: mouse over on cBioPortal column in mutation table not working


jjgao:
We used to have a cross-cancer histogram when hovering the number in the cBioPortal column.

adamabeshouse:
Can you put a screenshot/replication instructions?

jjgao:
e.g. in the page below, nothing happens after hovering the numbers on the cBioPortal column. 

http://www.cbioportal.org/case.do?cancer_study_id=ucec_tcga_pub&case_id=TCGA-BK-A0CC

adamabeshouse:
Do you think this is related to my code?

jjgao:
@inodb: could you help look into this? We also want to change the number for truncating mutations. Let's discuss.

jjgao:
@inodb if you haven't started working on this, I'll take this issue. I need to make some other changes as well.

inodb:
@jjgao Haven't started on this yet, so please do

Issue #501, opened 2015-11-17
display matches based on position for the cbioportal column


jjgao:
Currently a truncating mutation will be matched to all truncating mutations. We want to match mutations in the same position only.

PR #536, opened 2015-11-25
fixed issues #229, #501 -- cBioPortal column mouse over

jjgao:


jjgao:
@onursumer this is about the cbioportal column in the patient view. Could you review the code?

core/src/main/java/org/mskcc/cbio/portal/servlet/PancancerMutationsJSON.java
```diff
@@ -136,19 +139,31 @@ private static synchronized AccessControl getaccessControl() {
...
+                            CanonicalGene gene = daoGeneOptimized.getGene(symbol);
+                            if (gene!=null) {
+                                long entrezId = gene.getEntrezGeneId();
+                                
+                                // create the query string: (<ENTREZ ID>,<POSITION>)
+                                posWithEntrez.add("(" + entrezId + "," + position + ")");
+                            }
+                        }
+                        
 			String[] parts = proteinPos.split("_");
```
onursumer: Looks like this line of code is no longer needed

onursumer:
We have another (duplicate) PancanMutationHistogram.js in the mutation-mapper library. I think we should try to refactor the patient view to use the mutation-mapper library before it is too late...

jjgao:
Yes, we should consolidate the two tables. We can remove a lot of code in PancanMutationHistogram.js. The code to display thumbnails are not useful any more. Let's not handle that for this PR. I'll create another ticket for this.

Issue #538, opened 2015-11-26
Charts on study summary display attribute ID instead of display name


j-hudecek:
Even though the clinical attribute is defined as 

a1[0].attributes[i]: Object
attr_id: "OS_MONTHS"
datatype: "NUMBER"
description: "Overall survival in months since diagnosis"
display_name: "Overall Survival (Months)"

The chart on Study summary is labeled OS_MONTHS and not Overall Survival (Months)

j-hudecek:
Proposed solution #539

PR #539, opened 2015-11-26
Use display name to label graphs on study view

j-hudecek:
Only if display name is not defined or is blank fallback to using attribute id. Solves #538


jjgao:
@inodb: could you review this PR?

inodb:
Looks fine to me. It does change the study view quite a bit, since all the headers for the graphs now use the `display_name` instead of the `attr_id`. E.g.
screen shot 2015-12-02 at 1 15 56 pm
Is there a reason why you didn't use `display_name` before @zhx828 ?

zhx828:
Probably made a mistake. It should use display_name instead.

jjgao:
@zhx828 could you check if it looks good for the mskimpact study.

PR #554, opened 2015-12-03
Added IE=Edge directive to prevent quirks mode

j-hudecek:
It is there for the normal header but the login page doesn't have it

### DIFF
--- a/portal/src/main/webapp/login.jsp
+++ b/portal/src/main/webapp/login.jsp
@@ -36,6 +36,7 @@
 <head>
 <title><%= GlobalProperties.getTitle() %>::cBioPortal Login</title>
 <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
+<meta http-equiv="X-UA-Compatible" content="IE=Edge"/> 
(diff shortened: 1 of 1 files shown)


jjgao:
Here is a post recommending against using x-ua-compatible...

http://stackoverflow.com/questions/6771258/whats-the-difference-if-meta-http-equiv-x-ua-compatible-content-ie-edge-e

j-hudecek:
We needed it at NKI as our IE 11 decides to go into quirks mode for internal (i.e. on local intranet) websites if it is not there which breaks the openid login buttons. I thought other people might encounter this problem so I've made the PR. As mentioned the main page also has this header so it improves consistency.

j-hudecek:
I haven't really found where they recommend against using it in the post. In fact there is one answer which describes my scenario as well http://stackoverflow.com/questions/6771258/whats-the-difference-if-meta-http-equiv-x-ua-compatible-content-ie-edge-e/15623604#15623604

inodb:
@j-hudecek @jjgao Yeah as described in the commit this is already in `global/header.jsp`. It's just missing from `login.jsp`. I think we should merge this

jjgao:

Issue #475, opened 2015-11-16
Improve performance of study view


pieterlukasse:
Tasks:
- profile study view to find bottlenecks: profile frontend and backend code, including DB queries.

Goals:
- report with main bottlenecks identified
- bottlenecks solved or addressed in specific new issues (add reference to these issues)

pieterlukasse:
@aderidder : @zhx828 started an [RFC document (RFC14)](https://docs.google.com/document/d/19xaYBvciCCYjY1j244Dl1wCpiQn6f0r4DzmKajdWE04/edit) for this.

zhx828:
ScatterPlot Optimization SHA: d8b6573943bd28d9c1b32ff4d4238f294ec42d86

zhx828:
Help needed on the Table.js initDataTable function. Looks like most of time has spend on redrawing the dataTable. I couldn't find any better solution for this except using other library.

You might want to use my branch to do the further improvement. (https://github.com/zhx828/cbioportal/tree/study-view)

zhx828:
Help also needed on drawPlots function in StudyView ScatterPlot(https://github.com/zhx828/cbioportal/blob/study-view/portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js). This is more like d3 performance issue. May need to find a better method.

pieterlukasse:
Hi @zhx828 , thanks for sharing the improvements. For your information, @aderidder and I are also looking at ways to improve some bottlenecks. We have worked on improving performance related to:
- qtips in query form case lists
- qtips and timeout in query results page: Plots view, Enrichment tab, 
- improvements in query results page to ensure each tab only gets loaded when clicked

We are also planning to improve some parts of Study view like the Datatables .

zhx828:
@pieterlukasse Should I send a pull request including about all my edits and maybe later you can merge with yours. Then we can discuss about the next common goal?

PR #560, opened 2015-12-04
Study view optimization

zhx828:
Related to #475

### DIFF
--- a/portal/src/main/webapp/js/src/study-view/component/PieChart.js
+++ b/portal/src/main/webapp/js/src/study-view/component/PieChart.js
@@ -95,8 +95,8 @@ var PieChart = function(){
     function addPieLabels() {
         var _filters =[];
 
-        $('#' + DIV.mainDiv + ' .study-view-pie-label').html("");
-        $('#' + DIV.mainDiv + ' .study-view-pie-label-copy').html("");
+        d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label').selectAll('*').remove();
+        d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label-copy').selectAll('*').remove();
 
         initLabelInfo();
         labelFunction();
--- a/portal/src/main/webapp/js/src/study-view/component/PieChart.js
+++ b/portal/src/main/webapp/js/src/study-view/component/PieChart.js
@@ -1024,15 +1024,11 @@ var PieChart = function(){
         _tableDiv.push('</tbody></table>');
         _tableDivStr = _tableDiv.join('');
 
-        var pieLabelTable = $('#' + DIV.mainDiv)
-            .find('.study-view-pie-label')
-            .append(_tableDivStr);
+        d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label').html(_tableDivStr);
 
-        pieLabelTable.find('table').attr('id', DIV.labelTableID+'-0');
+        $('#' + DIV.mainDiv + ' .study-view-pie-label table').attr('id', DIV.labelTableID+'-0');
 
-        $('#' + DIV.mainDiv)
-            .find('.study-view-pie-label-copy')
-            .append(_tableDivStr);
+        d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label-copy').html(_tableDivStr);
 
         if(selectedAttrDisplay.length > maxLabelNameLength && selectedAttrDisplay.length > _labelHeaderLT) {
             var _th = $('#' + DIV.mainDiv).find('#' + DIV.labelTableID+'-0 thead th:nth-child(1)');
@@ -1041,7 +1037,7 @@ var PieChart = function(){
             _th.text(selectedAttrDisplay.substring(0, maxLabelNameLength<=_labelHeaderLT?_labelHeaderLT-2:maxLabelNameLength-3) + '...');
         }
 
-        if(tableAttrs.hasOwnProperty(selectedAttr)){
+        if(tableAttrs.hasOwnProperty(selectedAttr) && currentView !== 'table'){
             $("#"+DIV.chartDiv+"-table-icon").click();
         }
     }
--- a/portal/src/main/webapp/js/src/study-view/component/PieChart.js
+++ b/portal/src/main/webapp/js/src/study-view/component/PieChart.js
@@ -89,6 +89,9 @@ var PieChart = function(){
     var labelLT = 5; //label length threshold
     var labelHeaderLT = 4; //label header length threshold
 
+    //The attributes will be converted to table view after initialization
+    var tableAttrs = {'CANCER_TYPE':1, 'CANCER_TYPE_DETAILED':1};
+
     //This function is designed to draw Pie Labels based on current color the
     //Pie Chart has. Pagging function will be added when the number of labels
     //bigger than 5.
@@ -206,7 +209,9 @@ var PieChart = function(){
         });
 
         if(category[1] === 'h1' && currentView === 'pie') {
-            initMainDivQtip();
+            $('#' + DIV.mainDiv).one('mouseover', function () {
+                initMainDivQtip();
+            });
         }
         //else if(category === 'extendable'){
         $("#"+ DIV.chartDiv +"-extend").css('display', 'block');
@@ -226,7 +231,7 @@ var PieChart = function(){
             style: {
                 classes: 'qtip-light qtip-rounded qtip-shadow qtip-lightyellow forceZindex qtip-max-width study-view-qtip'
             },
-            show: {event: "mouseover", solo: true, delay: 0},
+            show: {event: "mouseover", solo: true, delay: 0, ready: true},
             hide: {fixed:true, delay: 300, event: "mouseleave"},
             position: {my:'left center',at:'center right', viewport: $(window)},
             content: copyPieLabelTable(),
@@ -464,7 +469,9 @@ var PieChart = function(){
                });
             $('#' + DIV.chartDiv ).css('display','block');
             $('#' + DIV.titleDiv ).css('display','block');
-            initMainDivQtip();
+            $('#' + DIV.mainDiv).one('mouseover', function () {
+                initMainDivQtip();
+            });
         });
 
         $("#"+DIV.chartDiv+"-reload-icon").click(function() {
@@ -960,21 +967,22 @@ var PieChart = function(){
 
     //Called when the number of label biggen than 6, used by addPieLabels()
     function labelFunction() {
-        var _tableDiv = '';
+        var _tableDiv = [];
+        var _tableDivStr = '';
         var _labelLT = labelWLT; //label length threshold
         var _labelHeaderLT = labelHeaderWLT; //label header length threshold
 
-        _tableDiv += '<table';
+        _tableDiv.push('<table');
 
         if(category[0] === 'w1') {
             _labelLT = labelLT;
             _labelHeaderLT = labelHeaderLT;
-            _tableDiv += ' class="smallTable" ';
+            _tableDiv.push(' class="smallTable" ');
         }else{
-            _tableDiv += ' class="regularTable" ';
+            _tableDiv.push(' class="regularTable" ');
         }
 
-        _tableDiv += '><thead><th>'+selectedAttrDisplay+'</th><th>#</th><th>Freq</th></thead><tbody>';
+        _tableDiv.push('><thead><th>'+selectedAttrDisplay+'</th><th>#</th><th>Freq</th></thead><tbody>');
 
         for(var i=0; i< label.length; i++){
             var _tmpName = label[i].name;
@@ -984,28 +992,28 @@ var PieChart = function(){
             }
 
             if(i % 1 === 0){
-                _tableDiv += "<tr>";
+                _tableDiv.push("<tr>");
             }
 
-            _tableDiv += '<td  uniqueId="'+
+            _tableDiv.push('<td  uniqueId="'+
                         DIV.labelTableTdID +label[i].id+'-'+i+
                         '"  style="font-size:'+fontSize+'px;white-space: nowrap;">'+
                         '<span class="pieLabel"><svg width="'+(labelSize+3)+'" height="'+
                         labelSize+'"><rect width="'+
                         labelSize+'" height="'+labelSize+'" style="fill:'+
-                        label[i].color + ';" /></svg><span';
+                        label[i].color + ';" /></svg><span');
             if(_tmpName !== label[i].name) {
-                _tableDiv +=  ' class="hasQtip"';
+                _tableDiv.push(' class="hasQtip"');
             }
-            _tableDiv += ' oValue="'+ (label[i].name.replace(/'/g, "&#39;")).replace(/"/g, "&#34;") + '" style="vertical-align: top">'+
+            _tableDiv.push(' oValue="'+ (label[i].name.replace(/'/g, "&#39;")).replace(/"/g, "&#34;") + '" style="vertical-align: top">'+
                 _tmpName+'</span></span></td><td class="pieLabelValue" uniqueId="'+
                 DIV.labelTableTdID +label[i].id+'-'+i+
                 '"><span style="margin: 3px 0;padding-top: 3px;">'+label[i].value+
                 '</span><input type="checkbox" style="float:right; margin: 3px 0;" /></td>' +
-                '<td>' + (Number(label[i].value)/sampleSize* 100).toFixed(1) + '%' + '</td>';
+                '<td>' + (Number(label[i].value)/sampleSize* 100).toFixed(1) + '%' + '</td>');
 
             if(i % 1 === 0){
-                _tableDiv += '</tr>';
+                _tableDiv.push('</tr>');
             }
 
             if(maxLabelNameLength < _tmpName.length) {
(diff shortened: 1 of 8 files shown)


zhx828:
@pieterlukasse since you are familiar with study view, could you help me review my code? Thanks so much

portal/src/main/webapp/js/src/study-view/component/PieChart.js
```diff
@@ -206,7 +209,9 @@ var PieChart = function(){
         });
 
         if(category[1] === 'h1' && currentView === 'pie') {
-            initMainDivQtip();
+            $('#' + DIV.mainDiv).one('mouseover', function () {
```
pieterlukasse: I would put this check in the initMainDivQtip() itself. See https://github.com/thehyve/cbioportal/commit/c4ee712310cb9c5824fb0d74ccb79812f0b75cf6

portal/src/main/webapp/js/src/study-view/component/PieChart.js
```diff
@@ -89,14 +89,17 @@ var PieChart = function(){
...
+
     //This function is designed to draw Pie Labels based on current color the
     //Pie Chart has. Pagging function will be added when the number of labels
     //bigger than 5.
     function addPieLabels() {
         var _filters =[];
 
-        $('#' + DIV.mainDiv + ' .study-view-pie-label').html("");
-        $('#' + DIV.mainDiv + ' .study-view-pie-label-copy').html("");
+        d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label').selectAll('*').remove();
```
pieterlukasse: does this have to be .selectAll ? I think you could use d3.select. So: 
d3.select('#' + DIV.mainDiv).selectAll('.study-view-pie-label').selectAll('*').remove();

portal/src/main/webapp/js/src/study-view/component/PieChart.js
```diff
@@ -960,21 +967,22 @@ var PieChart = function(){
...
     //Called when the number of label biggen than 6, used by addPieLabels()
     function labelFunction() {
-        var _tableDiv = '';
+        var _tableDiv = [];
+        var _tableDivStr = '';
         var _labelLT = labelWLT; //label length threshold
         var _labelHeaderLT = labelHeaderWLT; //label header length threshold
 
-        _tableDiv += '<table';
+        _tableDiv.push('<table');
```
pieterlukasse: instead of building up a string, I would suggest using _.template functionality together with the possibility to add new table rows via .append or appendTo functions in jquery. 

e.g. using jquery, something like this should work I think: 
var tableDiv = d3.select('#' + DIV.mainDiv).selectAll('.study-view-pie-label-copy');
$('<table />', {class: (category[0] === 'w1'? "smallTable" : "regularTable"}).appendTo(tableDiv);

etc

portal/src/main/webapp/js/src/study-view/component/PieChart.js
```diff
@@ -960,21 +967,22 @@ var PieChart = function(){
...
             _labelHeaderLT = labelHeaderLT;
-            _tableDiv += ' class="smallTable" ';
+            _tableDiv.push(' class="smallTable" ');
         }else{
-            _tableDiv += ' class="regularTable" ';
+            _tableDiv.push(' class="regularTable" ');
         }
 
-        _tableDiv += '><thead><th>'+selectedAttrDisplay+'</th><th>#</th><th>Freq</th></thead><tbody>';
+        _tableDiv.push('><thead><th>'+selectedAttrDisplay+'</th><th>#</th><th>Freq</th></thead><tbody>');
```
pieterlukasse: so, based on comment above, this could be done with something like: 

```javascript
table.append(
    $('<thead/>').append(
        $('<th/>', {text: selectedAttrDisplay})
    )
    .append(
        $('<th/>', {text: '#'})
    )
    .append(
        $('<th/>', {text: 'Freq'})
    );
var tableBody = table.append('<tbody/>');
```

portal/src/main/webapp/js/src/study-view/component/PieChart.js
```diff
@@ -984,28 +992,28 @@ var PieChart = function(){
             }
 
             if(i % 1 === 0){
-                _tableDiv += "<tr>";
+                _tableDiv.push("<tr>");
             }
 
-            _tableDiv += '<td  uniqueId="'+
+            _tableDiv.push('<td  uniqueId="'+
```
pieterlukasse: here you could use the _.template functionality. See for example how templates from pancancer_study_summary_templates.html are used

portal/src/main/webapp/js/src/study-view/component/PieChart.js
```diff
@@ -1013,17 +1021,14 @@ var PieChart = function(){
...
+        _tableDiv.push('</tbody></table>');
+        _tableDivStr = _tableDiv.join('');
 
-        var pieLabelTable = $('#' + DIV.mainDiv)
-            .find('.study-view-pie-label')
-            .append(_tableDiv);
+        d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label').html(_tableDivStr);
 
-        pieLabelTable.find('table').attr('id', DIV.labelTableID+'-0');
+        $('#' + DIV.mainDiv + ' .study-view-pie-label table').attr('id', DIV.labelTableID+'-0');
```
pieterlukasse: this deviates from you line below. Shouldn't this be something like: 
```
d3.select('#' + DIV.mainDiv).select...

```
?

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
```diff
@@ -428,14 +435,22 @@ var ScatterPlots = function() {
...
                     fill: fill,
-                    stroke: stroke
-                });
-            })
-            .attr("d", d3.svg.symbol()
-                .size(style.size)
-                .type(style.shape))
-            .attr("stroke-width", style.stroke_width)
-            .attr("z-index", "100");
+                    stroke: stroke,
```
pieterlukasse: FYI - in other performance tests we made, we found that stroke attribute slows down the plot. You could consider not adding it by default and only add it when the circle has been clicked/selected.

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
```diff
@@ -1152,47 +1168,25 @@ var ScatterPlots = function() {
...
-                    if(_datumArr[_index].fill === style.fill && _datumArr[_index].stroke === 'red'){
-                        $(this).attr("clicked", 'clicked');
-                    }else if(_datumArr[_index].fill === 'red' && _datumArr[_index].stroke === 'red'){
-                        $(this).attr("clicked", 'shiftClicked');
-                    }else if(_datumArr[_index].fill === 'red' && _datumArr[_index].stroke === style.stroke){
-                        $(this).attr("clicked", 'both');
-                    }
+            for(var i=0 ; i< dataCopyL ; i++){
+                var _case_id = dataCopy[i].case_id;
+                if(highlightCases.hasOwnProperty(_case_id)) {
```
pieterlukasse: maybe this is faster: 
```javascript
if(highlightCases[_case_id]) 
```

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
```diff
@@ -1152,47 +1168,25 @@ var ScatterPlots = function() {
...
+                    _highlightTmpDataArr.push(jQuery.extend(dataCopy[i],_datumArr[highlightCases[_case_id]]));
+                }else if(caseIdList.hasOwnProperty(_case_id)){
```
pieterlukasse: same comment as above:
```javascript
}else if(caseIdList[_case_id]){
```

portal/src/main/webapp/js/src/study-view/component/Table.js
```diff
@@ -135,8 +135,8 @@ var Table = function() {
 
     function initTable(data) {
         var table = $('#' + divs.tableId);
-        var tableHeaderStr = '';
-        var tableBodyStr = '';
+        var tableHeaderStr = [];
```
pieterlukasse: same comments as for PieCharts.js above

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -412,6 +413,117 @@ var StudyViewProxy = (function() {
...
+
+    function getCNABasedOnSampleIds(sampleIds) {
+        var data = {
+            alter: [],
+            caseIds: [],
+            cytoband: [],
+            gistic: [],
+            gene: []
+        };
+        if (sampleIds instanceof Array) {
```
pieterlukasse: why this check? If this is some kind of validation which is really necessary, then I would also add a respective error message when sampleIds is not an Array.... e.g. 
```
else {
 throw "error in getCNABasedOnSampleIds: sampleIds should be an array";
}
```

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -412,6 +413,117 @@ var StudyViewProxy = (function() {
...
+                            var key = obtainDataObject.cnaSampleBased[sampleIds[i]][j].gene + obtainDataObject.cnaSampleBased[sampleIds[i]][j].alter;
+                            if (!geneSpecific.hasOwnProperty(key)) {
+                                geneSpecific[key] = obtainDataObject.cnaSampleBased[sampleIds[i]][j];
+                                geneSpecific[key].caseIds = [];
+                            }
+                            geneSpecific[key].caseIds.push(sampleIds[i]);
+                        }
+                    }
+                }
+                _.values(geneSpecific).map(function (item) {
```
pieterlukasse: this loops again over the geneSpecific list. It is better (and also faster) to do this immediately after line 460, right? So just after line 460 above, add:
```javascript
                   data.alter.push(geneSpecific[key].alter);
                   data.caseIds.push(geneSpecific[key].caseIds);
                   data.cytoband.push(geneSpecific[key].cytoband);
                   data.gistic.push(geneSpecific[key].gistic);
                   data.gene.push(geneSpecific[key].gene);
```

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -412,6 +413,117 @@ var StudyViewProxy = (function() {
...
+
+    //Change the mutated gene data base on sample ID
+    function convertMutatedGeneData(data) {
+        var converted = {};
+
+        if (data) {
+            for (var i = 0, dataL = data.length; i < dataL; i++) {
+                var caseIds = data[i].caseIds;
+
+                for (var j = 0, caseIdsL = caseIds.length; j < caseIdsL; j++) {
```
pieterlukasse: there is no need to make this extra caseIdsL variable. I would write
```
j < caseIds.length
```
This was also found in other for loops by the way

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -412,6 +413,117 @@ var StudyViewProxy = (function() {
...
+        return converted;
+    }
+
+    function getMutatedGeneDataBasedOnSampleIds(sampleIds) {
+        var data = [];
+        if (sampleIds instanceof Array) {
+            if(sampleIds.length === this.getSampleIds().length) {
+                return obtainDataObject.mutatedGenes;
+            }
+            if(obtainDataObject.hasOwnProperty('mutatedGenesSampleBased')){
```
pieterlukasse: I think it is not such a good practice to use globals like is done here with obtainDataObject. A *small* improvement to this would be to wrap access to this object via a getter. Something like: 
```javascript
function getDataObject() {
   return obtainDataObject;
}
```
still not pretty....but is an improvement. .

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -460,6 +573,8 @@ var StudyViewProxy = (function() {
         getArrDataBySampleIds: getArrDataBySampleIds,
         getAttrData: function(){ return obtainDataObject.attr;},
         getMutatedGenesData: getMutatedGenesData,
+        getMutatedGeneDataBasedOnSampleIds: getMutatedGeneDataBasedOnSampleIds,
```
pieterlukasse: General comment: 
I would favor StudyViewProxy connecting in some way to the new API exposed in cbioportal-client.js and cbioportal-datamanager.js (also known as QuerySession)

These methods like StudyViewProxy.getMutatedGeneDataBasedOnSampleIds seem to repeat things that are (or should be) also available in QuerySession API. 

@adamabeshouse, @jjgao : what do you think?

adamabeshouse: I certainly agree. @zhx828, are we right in assuming that getMutatedGeneDataBasedOnSampleIds is doing the same thing as QuerySession.getGenomicEventData({genetic_profile_ids: [..], genes:[..], sample_ids:[..]}) ?

portal/src/main/webapp/js/src/study-view/component/PieChart.js
zhx828: I've changed in here: https://github.com/zhx828/cbioportal/commit/2f06bf7dac17a6b625550346f5b334e54e90f5ec

portal/src/main/webapp/js/src/study-view/component/PieChart.js
zhx828: I changed from appendTo actually. jQuery appendTo is really slow sometimes especially when you try to append several times. Instead, building up an array could be one of better solution for this.

portal/src/main/webapp/js/src/study-view/component/PieChart.js
zhx828: Same as above.

portal/src/main/webapp/js/src/study-view/component/PieChart.js
zhx828: I only replaced some time consuming functions like html and remove.

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
zhx828: Oh fascinating, is the stroke only the component effect the performance?

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
zhx828: You are right!! Check this out: http://jsperf.com/key-exist-check/3 I will start to use your way from now on.

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
zhx828: Same as above.

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
zhx828: Function documentation has been added in here : https://github.com/zhx828/cbioportal/commit/7155054ce2735af66db3332330f1e79efd729ec6
Unnecessary has been removed in here: https://github.com/zhx828/cbioportal/commit/02a3320e43b29a881da97f10bdb89b01f7079ba0

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
zhx828: It has been changed in here: https://github.com/zhx828/cbioportal/commit/04af2fb837c3e96a7eefa2e3517b911c6271e421

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
zhx828: This is kind of my coding style. There is performance test that the way it is may have better performance: https://jsperf.com/array-length-vs-cached
Open to discussion. :)

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
zhx828: :+1:  this may require lots of modification. I would rather postpone this suggestion after integrating with QuerySession. What do you think?

pieterlukasse: ok, you could leave this for later

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
pieterlukasse: surprising! I didn't expect this performance difference!

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
pieterlukasse: ok!

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
pieterlukasse: cool. Now, I see you added _.isUndefined to one of the lines but used uncondition in the other ones. Maybe it is better to stick to one of them?

portal/src/main/webapp/js/src/study-view/component/ScatterPlots.js
pieterlukasse: one of the most important ones as it seems...

portal/src/main/webapp/js/src/study-view/component/PieChart.js
pieterlukasse: I mean, in line 1029 you write:
```
$('#' + DIV.mainDiv + ' .study-view-pie-label table')
```
and on line 1027, 1031 yo write: 
```
d3.selectAll('#' + DIV.mainDiv).selectAll('.study-view-pie-label-copy')
```
two different ways for selecting. Makes the code a bit messy.

portal/src/main/webapp/js/src/study-view/component/PieChart.js
pieterlukasse: hm...that is a pity that jquery would be so slow...but does it have a dramatic impact here? This part of the code does not seem like a place that should have big impact in the performance (i.e. a small table is being rendered here)....do we really need this kind of optimization here?

PR #566, opened 2015-12-08
make service get sample id lists instead of patient ids

adamabeshouse:


jjgao:
@zheins since this is related to the patient/sample list refactoring, could you review this PR?

Issue #573, opened 2015-12-09
Create targetedQTip in cbio-util


zhx828:
Create util to initialize qTip only if mouse over the target.

PR #575, opened 2015-12-09
Add addTargetedQTip for initializing qTip when first time mouseenter the target

zhx828:


zhx828:
This close #573

zhx828:
@onursumer could you review?

PR #578, opened 2015-12-10
Set minimum rectangle width on timeline to 5px

j-hudecek:


inodb:
Thanks for the PR. Do you mean to set the minimum width of a rectangle on the timeline instead of the ticks' min width? It seems like that's what this change does. Timeline specific PRs are better done here: https://github.com/cbioportal/clinical-timeline. The code is included as a library. I still have to use some js minimizer and comment the file to make that more clear.

j-hudecek:
Yes it's the rectangle, that's what I meant. Patients sometimes get a drug only once and it created 1px wide line which were impossible to hover over.
 
What is the relation ship between this code and the clinical-timeline repo? I mean - this js file gets deployed and actually creates some timeline. Can I take the code from clinical-timeline repo and drop it in our installation and expect it to work? It seems it works differently than the old one (I can't even find the place where to make this change in it)

inodb:
At the top of the file you can see the version:
https://github.com/j-hudecek/cbioportal/blob/bca2a7692739f07168616a82872ebb01a80c524e/portal/src/main/webapp/js/src/patient-view/clinical-timeline.js#L1

The line you edited corresponds to this line in version v0.4.0:
https://github.com/cBioPortal/clinical-timeline/blob/e9ac960c1e061e4cacceb9a896e511d62648c463/js/lib/d3-timeline.js#L425

So actually this rabbit hole goes deeper and is an edit in the d3-timeline library, but since that [repo](https://github.com/jiahuang/d3-timeline) does not seem to be under active development anymore it is fine to change it in the clinical-timeline repo. Afterwards we can do a PR to the cbioportal repo and update the clinical-timeline version (i.e. ```make clean && make build``` then copy over the file). Anyway I understand it's a bit convoluted for such a small change, so I can update it myself if you want.

j-hudecek:
Yes please go ahead and perform the needed magical incantations. I'm not sure if you want to solve the problem exactly in this way in the new version of the library, but it's better than nothing.

PR #587, opened 2015-12-14
updates to python scripts

zheins:
@priti88 @fedde-s please test/review. I don't have my normal work laptop and can't test properly.

Import cancer type information needs to be further defined in the meta_study file - I just used the type_of_cancer field for most of the information.


zheins:
@fedde-s You're right, sorry. I'll put them back!

zheins:
I thought we just assumed meta would be somewhere in the filename? I suppose it would be possible to determine if its a metafile based on the structure of the file, meaning if every line follows the <key>:<value> format (ignoring #).

zheins:
Meaning the files no longer need to be meta_<data_type>, but meta files are specified by meta in the filename, data files can be named whatever as the path is described in the meta.

fedde-s:
Ah, sorry, I think I misread what the string you try to find in metadata.values() represents. It still doesn’t look like you’re getting the data type from the value of a specific field in the meta file though.

fedde-s:
Dropping the use of the word meta in the file name as a signal is not what I meant to suggest, sorry. I misread what you were doing in the commit, partially confused because you were not looking at the value of any particular meta file field. (would be `meta_file_type`, and @priti88 cleverly decided to use the old file name conventions this field replaces as the values for this field)

Issue #591, opened 2015-12-14
PORTAL_HOME should not be required at run time


inodb:
Currently if no `PORTAL_HOME` is set, running the portal war crashes with:
```
SEVERE: Exception sending context initialized event to listener instance of class org.springframework.web.util.Log4jConfigListener
java.lang.IllegalArgumentException: Could not resolve placeholder 'PORTAL_HOME' in string value "file://${PORTAL_HOME}/src/main/resources/log4j.properties"
```
Since we use multiple wars on a single Tomcat it does not make sense to have dependencies on PORTAL_HOME during run time. In this case it would seem that multiple instances of the portal on the same server might be using the same log4j properties file.

inodb:
I don't think this has been fixed yet, for latest discussion see https://github.com/cBioPortal/cbioportal/pull/605

PR #605, opened 2015-12-15
Fix #546 Allow deployment on Heroku

### LINKED ISSUE #546: automate building war files for pull requests
- [x] build a new war file for every pull request.
 - [x] set up a public test database
 - [x] deploy the new war file, named pr-1234.war?
 - [x] ideally delete the war file after a PR is closed.

Deploy to OpenShift: 
https://blog.openshift.com/how-to-build-and-deploy-openshift-java-projects-using-travis-ci/

Deploy to Heroku:
https://blog.travis-ci.com/2013-07-09-introducing-continuous-deployment-to-heroku/

You can also deploy Java Applications with the Heroku Maven Plugin:
https://devcenter.heroku.com/articles/deploying-java-applications-with-the-heroku-maven-plugin

inodb:
- Add webapp-runner.jar as dependency (used by heroku to run war)
- Add deploy to heroku button to README
- Remove unnecessary target directories after building war to decrease heroku's
  slug size (basically size repo + war)
- Add dbconnector spring property to allow using jndi or dbcp at runtime.
  Heroku's webapp-runner.jar does not seem to work with our setup of jndi.
- There is a heroku profile in pom.xml that copies all heroku specific files to
  the root dir of the portal when building the war. Only app.json has to be in
  the root dir for the deploy to heroku button to work. Copy EXAMPLE property
  file to src/main/resources/portal.properties. Set custom properties on
  command line with maven in app.json (i.e. use amazon public db).
- There was a problem with PORTAL_HOME both being required at run time and build
  time. At Heroku the build directory is different from the run directory and
  currently there is no dynamic way to get the name of this dir. Scripts in
  profile.d are run on heroku before the war is started. This allows one to set
  the PORTAL_HOME variable there. The PORTAL_HOME variable at build time is set
  through a buildpack on github inodb/cbioportal-buildpack. This is why the
  extra buildpack is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
 # cBioPortal
 
 [![Build Status](https://travis-ci.org/cBioPortal/cbioportal.svg?branch=master)](https://travis-ci.org/cBioPortal/cbioportal)
--- a/app.json
+++ b/app.json
@@ -1,0 +1,31 @@
+{
+  "name": "cBioPortal",
+  "description":"The cBioPortal for Cancer Genomics provides visualization, analysis and download of large-scale cancer genomics data sets.",
+  "repository":"https://github.com/cbioportal/cbioportal",
+  "logo":"http://www.cbioportal.org/images/cbioportal_logo.png",
+  "keywords": [
+    "java",
+    "tomcat",
+    "javascript",
+    "cancer",
+    "genomics"
+  ],
+  "env": {
+    "JAVA_OPTS": {
+      "description":"options for jvm",
+      "value": "-Xmx300m -Xms300m -XX:+UseCompressedOops"
+    },
+    "MAVEN_CUSTOM_OPTS": {
+        "description":"set heroku profile for mvn",
+        "value":"-Pheroku,public -DskipTests -Ddb.user=cbio_user -Ddb.password=cbio_pass -Ddb.portal_db_name=public_test -Ddb.connection_string=jdbc:mysql://cbioportal-public.c1xhhbwn8izk.us-east-1.rds.amazonaws.com:3306/ -Ddb.host=cbioportal-public.c1xhhbwn8izk.us-east-1.rds.amazonaws.com"
+    }
+  },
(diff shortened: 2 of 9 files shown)


n1zea144:
First, thanks for fixing core/pom.xml.  I don't understand this notation in web.xml:
${dbconnector:jndi}
Is dbconnector:jndi a new property?

Is PORTAL_HOME still required at runtime?  We required it at one point when we were allowing properties to be read outside the WAR (before our code went open source and groups could not build their own war).  I didn't think we still had that dependency.

morungos:
I still think this is a build issue worth discussing some time. We struggled with it a bit, as I generally don't like having to bake configurations into a war file. It's hard to test in staging, for example, as you need to recompile to change the DB. My concern isn't having the external config hook (I think that's important) but having the same variable used at build time and run time, sometimes being used to template values into the war, and sometimes later. It needs clarification as much as anything, and separating these functions might be advisable.

inodb:
Thanks for the comments @n1zea144 @morungos 

The `dbconnector` is a new run time property. This was required for deactivating the jndi and using dbcp instead. It seems the webapp-runner.jar from heroku does not support jndi, or at least not the way we use it. By setting it as `dbconnector:jndi` it will use jndi as the default value leaving everything as is. In the Procfile you can see how it is used when the war is run: `-Ddbconnector=dbcp`.

 `PORTAL_HOME` is still used at run time, see issue https://github.com/cBioPortal/cbioportal/issues/591. It would be good to remove at least the runtime dependency on the `PORTAL_HOME` env variable. We could perhaps expand the [portal.properties wiki page](https://github.com/cBioPortal/cbioportal/wiki/portal.properties-Reference). Divide it in run time and build time parameters. I agree that it would be very useful to be able to specify the db at run time. That being said I think that is outside the scope of this PR.

n1zea144:
I think the current behavior is to look for an external PORTAL_HOME var and a portal.properties file within and if not found, look on the classpath (actually that may not be true, at compile time it may be looking for PORTAL_HOME/src/main/resources/portal.properties, but at runtime, PORTAL_HOME/portal.properties).

In any event, yes, this is a good discussion to have.

PR #607, opened 2015-12-16
Adjustments to OQL and Oncoprint

adamabeshouse:
Adjust how sample data is processed and thus how OQL filtering is done, all in the service of fixing bug where INFRAMEs registered as TRUNCs in the oncoprint.


jjgao:
- [ ] fusions should be ordered after other mutations
image
 - [ ] when "show all mutations with the same color," fusions should still show with different color and glyph

jjgao:
I'll merge it now. An issue has been created: https://github.com/cBioPortal/cbioportal/issues/609

PR #613, opened 2015-12-16
Jinja html validator output

fedde-s:
I didn’t finish the formatting of long lists of aggregated line numbers (for errors that appear on multiple lines in a file) just yet, but here’s basically all other functionality of the new HTML report. Fee free to hit merge if you trust it, @zheins (and @priti88).


fedde-s:
I didn’t mean there’s a half-finished feature in the code that will prevent it from running, these aggregated fields simply appear as empty cells in the table (until I look into it again tomorrow).

Issue #614, opened 2015-12-16
Authentication makes assumptions about Spring Security providers


morungos:
I've discovered an issue with OpenID Connect re: commit: https://github.com/cBioPortal/cbioportal/commit/64d3cd79c2f332ef4aa8f1e7529f392c3817bec3#diff-ab796ae4f68402c7773da4d59de399a7R105

The underlying problem is in the commit message:

> All AuthenticationProviders return principals which implement interface UserDetails

This is not actually true. If it was, `getPrincipal()` would return a `UserDetails` and it doesn't. The Spring Security docs say: "Many of the authentication providers will create a UserDetails object as the principal" (see: http://docs.spring.io/autorepo/docs/spring-security/4.0.3.RELEASE/apidocs/org/springframework/security/core/Authentication.html#getPrincipal--)

This is throwing a cast error for the OpenID Connect providers from Mitre OIDC, that actually returns an immutable map as the principal. 

I can't see a reason for the cast here needing to be anything other than `Object`, as the authentication is passed directly on anyways.

PR #617, opened 2015-12-16
Don't always assume getPrincipal() is a UserDetails

morungos:
A small fix to the unnecessary assumption of UserDetails from Spring Security authentication providers.

See #614


morungos:
Sorry, this is garbage. I'm attempting to backport from a fork which has this issue, but it needs a bigger fix because it actually needs to pass the authentication instance to other methods to verify permissions. Same basic issue, but I need to improve the fix

morungos:
I've made the fixes now, and it should be better, except for maybe an update against rc.

Issue #424, opened 2015-11-10
Ghost case on LUAD_TCGA


jjgao:
http://www.cbioportal.org/case.do?cancer_study_id=luad_tcga&sample_id=Val.Stat

jjgao:
Fixed in code.. still need to fix the db manually.

jjgao:
@zheins has this been fixed in the code?

PR #622, opened 2015-12-16
Skip looking at first item in ProteinArrayData

zheins:
This field is not data - it is a header. This caused weird fields to show up as Patients.


zheins:
Fixes issue #424

PR #624, opened 2015-12-16
Added validator and meta script

priti88:
merge with fedde's html


fedde-s:
I think the root `scripts/` directory is a better place for the meta script than the `import_data_validator` package. Currently you are proposing to have both, while giving them different command line arguments and making them use different import scripts. Which version did you mean to keep?

priti88:
We can remove the meta script from import_data_validator directory. 

/scritps/meta_import.py
/scripts/cbioportalImporter.py
/scripts/import_data_validator/validateData.py

These will be the final files.

fedde-s:
Ah, okay, thanks. I’ll remove the other one and add the commit to my pull request!

PR #639, opened 2015-12-18
Improvements based on Pieter's feedback in #560

zhx828:


zhx828:
@pieterlukasse I think I will stick with underscore then.

pieterlukasse:
OK

PR #529, opened 2015-11-20
changed meta-file flag for ImportCopyNumberSegmentData

pieterlukasse:
small fix to make sure the import process via  cbioportalImporter.py is aligned with the requirement for a meta file in the underlying java process (specific fix for ImportCopyNumberSegmentData)

PR #642, opened 2015-12-21
Performance improvements in some of the query results tabs

pieterlukasse:
Notes:
- Please ignore first 2 commits, they are also part of #529 

Most important changes here are:
- making sure tab contents only load when tabs are clicked. Fixes made for Network tab, Mutex tab (Mutex content was being loaded when any tab was clicked) and pancancer tab.
- pancancer retrieves data from 2 services. These are now fired in parallel.


jjgao:
@adamabeshouse could you please review this?

jjgao:
@pieterlukasse I like the idea of loading the tab after click. There is only one problem: we used to be able to go to a tab directly (http://bit.ly/1HrZBrX#pancancer_study_summary). In your new code, it's not working.

portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary.js
```diff
@@ -790,12 +790,20 @@ function DataManagerPresenter(dmInitCallBack)
 			
 			//do the next call:
 			console.log(new Date() + ": CALL to get sample clinical atttributes (cancer types)");
-			return window.QuerySession.getSampleClinicalData(["CANCER_TYPE","CANCER_TYPE_DETAILED"]);
+			//this and next call below are in parallel, so the last one should call the dmInitCallBack:
+			nrResponsesReceived++;
+			if (nrResponsesReceived == 2) {
+				dmInitCallBack(self);
+			}
```
adamabeshouse: instead, you could do:

At line 769:
```
var genomic_data_fetched = new $.Deferred();
var clinical_data_fetched = new $.Deferred();
```

then here, ```genomic_data_fetched.resolve()```. Below, ```clinical_data_fetched.resolve()```. And instead of calling the callback,

``` $.when(genomic_data_fetched, clinical_data_fetched).then(function() {
// body of dmInitCallBack
});

Comment (pieterlukasse) [inline portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary.js]: hi Adam, thanks for the tip. I was aware of this option, however I chose to not do it like this since the function in then() will only be called when both genomic_data_fetched, clinical_data_fetched are ready. By splitting it up, I can now do the respective javascript processing of each part independently. So if e.g. genomic_data_fetched returns faster than clinical_data_fetched, I can process this data while the download of clinical_data_fetched continues. If there is a better way of doing this it would be great. I would hope there is something like:
```
$.when(data1, data2).then(functionA, functionB, functionC);
```
where functionA is called when data1 is in, functionB is called when data2 is in, and functionC is called when functionA and functionB are both done... I checked quickly but could not find something like this.

**Comment by pieterlukasse** [comment]
hi @jjgao , thanks for testing it and reporting the issue. I pushed up the fixes for the tabs I worked on as part of this PR: plots, mutex, pancancer and network.

File: portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary.js (hunk 1 above)
Comment (adamabeshouse) [inline portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary.js]: I think I wasn't clear enough - my solution accounts for this problem. You resolve the promises after your individual callbacks have been called. So it really is functionally equivalent to what you have.

A function that does this might look like:
```
var promiseHandler = function(promiseA, promiseB, callbackA, callbackB, finally) {
    var callbackA_done = new $.Deferred();
    var callbackB_done = new $.Deferred();
    promiseA.then(function(response) {
        callbackA(response);
        callbackA_done.resolve();
    });
    promiseB.then(function(response) {
        callbackB(response);
        callbackB_done.resolve();
    });
    $.when(callbackA_done, callbackB_done).then(finally);
}
```

pieterlukasse:
@adamabeshouse , @jjgao 

thanks for the detailed explanation. I implemented this (and some extra fixes to a bug I found during testing) and pushed it to the PR. 

You can find the implementation in https://github.com/thehyve/cbioportal/commit/2862ecc4ba5d59cb03b72122e6508123a6a459b5
(and after this commit I made a small optimization and 2 fixes related to bugs in how 2 parameters/filters were being set). 

I think that because I pushed this to the same PR, your comments to the original commit got lost :(  sorry for the inconvenience.

sheridancbio:
I've just begun looking at the code changes here, but I think changes on this fork are related to issue #709 which I reported yesterday: In the patient_view display, there is a tab for viewing tissue images. There is a direct URL link shown on the news page to the anchor #tab_images on patient_view. The tab opens through the link, but the contents of the tab are not initialized. (related commit: thehyve@7fe1bc5)

Apologies if this is not related.

pieterlukasse:
Hi @sheridancbio, thanks for linking this. Issue #709 is indeed about another page (Patient View), so it is not related to this pull request. However, the solution for #709 could be the same as done here for the Query results tabs. I made the fix for this tab in PR #721 as an example. @jjgao : can someone in your team pick this up for the other tabs?

jjgao:
@pieterlukasse could you please resolve the conflicts?

pieterlukasse:
@jjgao : done, I've resolved the conflict.

PR #649, opened 2015-12-22
Plotstab performance improvements

pieterlukasse:
significant improvements in performance in the plots tab for medium sized studies (~1K samples)


yichaoS:
@jjgao @pieterlukasse  I just tested it. It does improves the speed a lot. But I couldn't find the radio buttons on the left side to choose between "Genetic Profile" and "Clinical Attribute" therefore couldn't switch to clinical attribute view at all?? In the source you can see the display for the buttons are set to be none. 
screen shot 2015-12-23 at 11 29 20 am

pieterlukasse:
@yichaoS @jjgao  : strange...I tested this and the radio buttons appear on my screen (see screenshot). Can you send more details like which browser you used, whether you get any javascript errors, etc. Maybe clear your cache just to be sure? Thanks!

image

PR #651, opened 2015-12-22
Remove fix option, gene info download by default

zheins:


fedde-s:
Ah, one more request: could you make the validator emit an info message (through the logger, with no `data_filename`, `line_number`, `column_number` or `cause` as neither is applicable) before and after downloading or reading the file? It may take a moment (or fail), and it could help people who are looking at the info messages to understand why it hasn’t started validating the first file yet.

zheins:
Closing - really out of date

Issue #665, opened 2015-12-23
scatter plot not working for samples without mutations


jjgao:
http://www.cbioportal.org/beta/case.do?sample_id=TCGA-VS-A9V4-01&cancer_study_id=cesc_tcga

PR #674, opened 2015-12-23
patient view scatter plot for samples without mutations

JiaoJiao123:
1: pass qtip content to scatterplot, instead of generating qtip in the
scatterplot
2: add empty object check for samples without mutations


jjgao:
solved https://github.com/cBioPortal/cbioportal/issues/665

PR #687, opened 2015-12-26
fixed issue #685

jjgao:
When combining sample data into patient data, the un-process sample data was being combined, but makePatientData seemed only working on oql-process data because it was combining events (e.g. mrna "UPREGULATED"). Also, running OQL process twice seemed unnecessary. 

So I made some change. Now makePatientData combines oql-processed data and also combines altered samples into altered patients and similar unaltered.

### DIFF
--- a/portal/src/main/webapp/js/api/cbioportal-datamanager.js
+++ b/portal/src/main/webapp/js/api/cbioportal-datamanager.js
@@ -62,47 +62,47 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 		};
 		
 		var getMatchingMutationsByType = function(mutations, oql_target_type, relation) {
-		    var ret = [];
-		    for (var i=0; i<mutations.length; i++) {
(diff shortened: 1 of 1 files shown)


jjgao:
@adamabeshouse sorry for the indentation problem. The main change is in the makePatientData function (line 570) and line 703.

jjgao:
@adamabeshouse I am going to merge and release it to fixed #685. Feel free to review and make changes.

adamabeshouse:
ah yes this was a major oversight on my part. I think we should change back makePatientData to operate on unprocessed data though.

jjgao:
> I think we should change back makePatientData to operate on unprocessed data though.

Any reason for that? I thought running OQL twice is not necessary. Maybe I missed something...

adamabeshouse:
You're right, we should add an intervening step, so that it goes like:

Webservice sample data -> sample data filtered through OQL -> sample data
                                                                                                   \-> patient data

PR #693, opened 2015-12-29
Check for undefined sample ids while validating data files

fedde-s:
Instead of collecting them and checking after all files are read. This way the error message can be much more helpful, informing the user of where in the study the id was found.

It means clinical data files are now read before all others (and case lists after), as would be done when loading the data into cBioPortal.


fedde-s:
Never mind, I’ll implement a few more such improvements and make a pull request with several in one go. That way I will not bother you this often!

Issue #517, opened 2015-11-18
Clean up database schema


jjgao:
- [x] change patient_list to sample_list
 - [x] change patient_list_list to sample_list_list
 - [x] fix Java code naming
 - [x] remove micro_rna and micro_rna_alteration
 - [x] remove mutation_frequency
 - [x] change field name within patient_list_list
 - [x] create version table in database
 - [x] migration script - account for versioning
 - [x] seed database
 - [ ] ER diagram for documentation

jjgao:
@zheins let's try to finish it.

morungos:
Might be worth looking at sqitch: https://github.com/theory/sqitch (although it's Perl so you'll probably all hate it)

pieterlukasse:
@jjgao : I would love to review this

zheins:
PR #700 addresses some of the immediate concerns. We have it merged in our development environment.

Aside from changing some of the names of classes and db tables to be consistent with our naming conventions, there is a python script which takes a sql script as input to update the database. The script is executed based on the version found in stored in an information table and the version assigned to the statements.

```sql
##version: 1.0.0
CREATE TABLE info (DB_SCHEMA_VERSION VARCHAR(8));
INSERT INTO info VALUES ("1.0.0");

##version: 1.1.0
CREATE TABLE sample_list LIKE patient_list;
INSERT sample_list SELECT * FROM patient_list;
...
```

jjgao:
@pieterlukasse: sure, please review #700. No need to ask :) Thanks.

pieterlukasse:
@zheins regarding the seed DB (in https://github.com/cBioPortal/cbioportal/wiki/Import-the-Seed-Database) can you please let me know where I can find the respective test seed DB file so that I can review this as well?

jjgao:
I am closing this one. Please log remaining issues in another one.

pieterlukasse:
@jjgao : Ok, I moved the diagram part to #1041

Issue #606, opened 2015-12-15
Most of the core tests are not working


inodb:
There seem to be quite some Exceptions raised in the latest master branch test https://travis-ci.org/cBioPortal/cbioportal/builds/95919781. Travis however still says the test was successful. Any idea @cataphract ?

cataphract:
Apparently maven is still deeming the build successful. The most obvious cause would be something like `maven.test.failure.ignore` set to true. Maybe running maven with `-X` gives a clue.

inodb:
Thanks for the reply. It is indeed because the failure results are ignored: https://github.com/cBioPortal/cbioportal/blob/master/core/pom.xml#L347. Basically the tests aren't working at all, so changing the title of this.

ecerami:
@inodb:  is this still open, e.g. travis is reporting all green, but unit tests are failing?

inodb:
@ecerami yup, the unit tests are still failing. The only test we currently have is the E2E test that takes a screenshot of the study view and the patient view.

ecerami:
@inodb:  that's just crazy!  thought we had fixed this months ago.  am going to assign this to @ersinciftci as his first big task.

inodb:
Great! Yeah it would be wonderful if we can get this fixed. Good luck @ersinciftci! As a side note, I've been told a bunch of the unit tests are pretty pointless. I'm guessing the data model has changed as well, so feel free to deprecate tests that are no longer useful. Ideally if you have the time, we would also be able to get some test coverage stats so we can get a better idea of what's untested. Maybe code climate https://docs.codeclimate.com/docs/setting-up-test-coverage or coveralls https://github.com/trautonen/coveralls-maven-plugin

Sjoerd-van-Hagen:
If Maven is set to ignore the failed tests, they are also ignored by the CI. If you run the unit tests directly, say from your IDE, they will not be ignored and thus may fail.

What needs to be done is update the pom so the failed tests are not ignored. Failing tests can be labeled with @Ignore if necessary. They will then be ignored both by jUnit and Maven, and therefore also by Travis.

morungos:
I've run JaCoCo for test coverage before, and I'm starting to backport some of my changes -- I can add this into the mix if anyone wants. It works OK, and JaCoCo is free and isn't dependent on external services. FYI, the coverage was around 20% when I last ran it

inodb:
[...]

PR #700, opened 2015-12-31
Database cleanup

zheins:
Issue #517 

@jjgao We need to coordinate running the migration script on our dbs with the merging of these changes, and communicate to users to do the same. The ER diagram needs to be updated as well on the wiki - I'll take care of this next week.


zheins:
@n1zea144 please review

core/src/main/resources/db/sql_migration_scripts/1.0.0.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE sample_list LIKE patient_list;
+INSERT sample_list SELECT * FROM patient_list;
+CREATE TABLE sample_list_list LIKE patient_list_list;
+INSERT sample_list_list SELECT * FROM patient_list_list;
+DROP TABLE micro_rna;
+DROP TABLE micro_rna_alteration;
+DROP TABLE mutation_frequency;
+ALTER TABLE sample_list_list CHANGE PATIENT_ID SAMPLE_ID INT(11);
+CREATE TABLE version (VERSION_NUMBER VARCHAR(8));
```
jjgao: Maybe rename table to info with a column DB_SCHEMA_VERSION. We can incorporate other information in the table.

core/src/main/resources/db/sql_migration_scripts/1.0.0.sql
```diff
@@ -0,0 +1,10 @@
...
+INSERT into version values ("1.0.0");
```
jjgao: Can we put all migration sql into one script, e.g. migration.sql? We can put versions inside the sql, e.g.

```sql
##version: 1.0.0
CREATE TABLE info (DB_SCHEMA_VERSION VARCHAR(8));
INSERT into version values ("1.0.0");

##version: 1.1.0
CREATE TABLE sample_list LIKE patient_list;
INSERT sample_list SELECT * FROM patient_list;
CREATE TABLE sample_list_list LIKE patient_list_list;
INSERT sample_list_list SELECT * FROM patient_list_list;
DROP TABLE micro_rna;
DROP TABLE micro_rna_alteration;
DROP TABLE mutation_frequency;
ALTER TABLE sample_list_list CHANGE PATIENT_ID SAMPLE_ID INT(11);
```

The python code has to be modified.

core/src/main/resources/db/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+##version: 1.1.0
+CREATE TABLE sample_list LIKE patient_list;
+INSERT sample_list SELECT * FROM patient_list;
+CREATE TABLE sample_list_list LIKE patient_list_list;
+INSERT sample_list_list SELECT * FROM patient_list_list;
+DROP TABLE micro_rna;
+DROP TABLE micro_rna_alteration;
+DROP TABLE mutation_frequency;
+ALTER TABLE sample_list_list CHANGE PATIENT_ID SAMPLE_ID INT(11);
+INSERT INTO info VALUES ("1.1.0");
```
jjgao: Maybe set instead of insert because we don't need multiple versions here.

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
...
+    except MySQLdb.Error, msg:
+        print >> ERROR_FILE, msg
+        return None
+    
+    if not version_table_exists:
+        return (0,0,0)
+
+    # Now query the table for the version number
+    try:
+        cursor.execute('select version_number from version')
```
jjgao: update this to use the info table.

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
...
+        CREATE TABLE ...
+    
+    """
+    
+    sql_file = open(sql_filename, 'r')
+    sql_version = (0,0,0)
+    run_line = False
+
+    for line in sql_file:
+        if line.startswith('#'):
```
jjgao: Maybe look for "##version:" instead of just "#" so that "#" can still be used for other comments.

pieterlukasse:
@zheins : now that you removed the tables micro_rna, what is the correct way to import micro_rna data? Could you update https://github.com/cBioPortal/cbioportal/wiki/File-Formats as well?

zheins:
micro_rna data is not supported in the portal right now. I'll go ahead and remove it from the file-formats page.

pieterlukasse:
@zheins : does that mean we will get rid of all the negative entrez_ids in `genes` table?

pieterlukasse:
@zheins : actually, I doubt that this is the case. It seems it just migrated to another place. See ImportGeneData.java main method. mirna should now be imported via this class, right? See also my comment on #805 regarding the class ImportMicroRNAIDs.java

jjgao:
@pieterlukasse the micro_rna and micro_rna_alteration tables are legacy code for support miRNA. They are not being used any more. Removing them won't affect any of the pipeline or function of the portal. miRNA data is being processed and merged into the gene table as you have noticed.

pieterlukasse:
@jjgao : exactly. That's why I think it would be good to update the documentation on how microRNA is supported and how one can ensure it gets loaded into his local instance.

jjgao:
@pieterlukasse  The current way of supporting miRNA is very hacky. As I said, Rob is working on a proper implementation. We will document that once we have the new implementation.

pieterlukasse:
@jjgao great, that is good news! One of our clients was actually complaining about this (i.e. the fact that miRNAs were in the same table as genes with negative entrezids). Looking forward to the solution.

core/src/main/resources/db/cgds.sql
```diff
@@ -185,14 +183,13 @@ CREATE TABLE `patient_list` (
...
--- TODO rename the PATIENT_ID column, which is misleading
-CREATE TABLE `patient_list_list` (
+CREATE TABLE `sample_list_list` (
   `LIST_ID` int(11) NOT NULL,
-  `PATIENT_ID` int(11) NOT NULL,
-  PRIMARY KEY  (`LIST_ID`,`PATIENT_ID`),
-  FOREIGN KEY (`PATIENT_ID`) REFERENCES `sample` (`INTERNAL_ID`) ON DELETE CASCADE
+  `SAMPLE_ID` int(11) NOT NULL,
+  PRIMARY KEY  (`LIST_ID`,`SAMPLE_ID`),
+  FOREIGN KEY (`SAMPLE_ID`) REFERENCES `sample` (`INTERNAL_ID`) ON DELETE CASCADE
```
pieterlukasse: missing foreign key to sample_list. Please add:
```sql
 FOREIGN KEY (`LIST_ID`) REFERENCES `sample_list` (`LIST_ID`)
```
I am not sure about the `ON DELETE CASCADE` part though...do we want this? I think that this could have some performance issues when removing large studies as a huge amount of deletes would have to be part of the same transaction (think of tables like `sample_cna_event`). @jjgao : what do you think about the `ON DELETE CASCADE` part?

pieterlukasse:
another thing I am missing is the foreign keys to `clinical_attribute`, i.e. `clinical_sample` and `clinical_patient` used to refer to this table via `ATTR_ID`. It does not seem you have removed these foreign keys, but I think they should be there. So if you can add them, this would be great.

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
+#! /usr/bin/env python
+
```
pieterlukasse: I think this script is a great idea. A few comments though:
- I still miss the part where the web application backend checks if it is running on the correct version of the DB. My idea is that, at startup, the backend should check if the DB has been migrated to the version that it is expecting it to be in. And if the migration did not yet occur, ask the user to run DB migration (this script) first. See also [my comment on 796] (https://github.com/cBioPortal/cbioportal/issues/796#issuecomment-177917780) here. 
@jjgao : any thoughts? Or will this be part of another PR?

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
...
+    if version2[1] > version1[1]:
+        return False
+    if version1[2] > version2[2]:
+        return True
+    return False
+
+def run_migration(db_version, sql_filename, cursor):
+    """
+
+        Goes through the sql and runs lines based on the version numbers. SQL version should be stated as follows:
```
pieterlukasse: I foresee some problems with this approach, like someone forgetting to add this version line. Also, having to put your SQL statement in 1 line can be quite painful. I would like to suggest that we agree on a file name convention that contains the version number and then the whole file is executed in this step.

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
...
+    portal_properties = get_portal_properties(properties_filename)
+    cursor = get_db_cursor(portal_properties)
+
+    if cursor is None:
+        print >> ERROR_FILE, 'failure connecting to sql database'
+        sys.exit(1)
+
+    # execute - get the database version and run the migration
+    db_version = get_db_version(cursor)
+    run_migration(db_version, sql_filename, cursor)
```
pieterlukasse: db_version will be None if an error occurs. This should be checked here.

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
...
+                sys.exit(1)
+            
+
+def usage():
+    print >> OUTPUT_FILE, 'migrate_db.py --properties-file [portal properties file] --sql [sql migration file]'
+
+def main():
+    """ main function to run mysql migration """
+    
+    try:
```
pieterlukasse: maybe before starting, warn the user that he needs to backup his DB in case something goes wrong...(it can happen that there is still a bug in the migration script)?

core/src/main/scripts/migrate_db.py
```diff
@@ -0,0 +1,204 @@
...
+
+    for line in properties_file:
+        line = line.strip()
+
+        # skip line if its blank or a comment
+        if len(line) == 0 or line.startswith('#'):
+            continue
+        
+        # store name/value
+        property = line.split('=')
```
pieterlukasse: could be simpler if using ConfigParser. See http://stackoverflow.com/questions/3595363/properties-file-in-python-similar-to-java-properties/26221097#26221097

portal/src/main/webapp/WEB-INF/jsp/global/global_variables.jsp
```diff
@@ -545,17 +545,17 @@ $(document).ready(function() {
...
     // genetic profile ids
     var geneticProfiles = "<%=geneticProfiles%>";
 
     // plots_tab.jsp
     var cancer_study_id = "<%out.print(cancer_study_id);%>",
-            patient_set_id = "<%out.print(patient_set_id);%>";
-    patient_ids_key = "";
-    if (patient_set_id === "-1") {
-        patient_ids_key = "<%out.print(patientIdsKey);%>";
+            sample_set_id = "<%out.print(sample_set_id);%>";
```
pieterlukasse: Can't we just clean-up all this documented out code?

core/src/main/resources/db/cgds.sql
jjgao: Good point on FOREIGN KEY on LIST_ID.

I think we should have ON DELETE CASCADE in order to maintain the integrity of the database, so that you can delete one sample with all its associated data. The same for deleting a cancer study. When deleting a cancer study, you do want to delete all associate data.

We haven't turn on the foreign key constraint yet because we are using MyISAM instead of InnoDb. Currently we run several sql delete queries to delete data from multiple tables when deleting a study. In the future we want to turn on the foreign key constraint and cascade the deletion.

core/src/main/scripts/migrate_db.py
jjgao: Good point. We need to put a version number that the web app is compatible. This number should be accessible by Java code. Where should we put it @zheins @pieterlukasse ? Once we have that, we can display a message on the web page when the migration script needs to be run.

pieterlukasse:
@zheins : nice work! I added comments for some improvements.

core/src/main/scripts/migrate_db.py
jjgao: Database schema change should be very rare and any change will be reviewed. I'd not worry about someone forgetting to add the version line.

You can add multiple lines following the version line.

core/src/main/scripts/migrate_db.py
jjgao: Good point.

core/src/main/resources/db/cgds.sql
pieterlukasse: @jjgao : the FOREIGN KEY will ensure the integrity. ON DELETE CASCADE will just *automate* the deletion cascade, so when you want to delete a sample, all you have to do is `delete from sample where...` . This is a great feature if the cascade does not involve deleting huge amount of records. I am just worried that this will also be used to delete a complete study...I am not sure if DB can handle this well....

core/src/main/scripts/migrate_db.py
pieterlukasse: still each SQL statement needs to be on a single line...which is a bit cumbersome (to write and to review)

core/src/main/scripts/migrate_db.py
zheins: If I used that, we would have to change how our properties files are structured. ConfigParser needs section headers, which aren't currently in our properties files.

core/src/main/scripts/migrate_db.py
jjgao: @zheins could you allow a sql statement in multiple lines (using ";" to separate)? Ideally all statements in one version should be in one transaction.

core/src/main/scripts/migrate_db.py
pieterlukasse: @jjgao a global static final variable would do the job. I like the idea of the message on the web page.

zheins:
@jjgao I'm using a digit in the app version from portal.properties. Do you agree with this approach or should we do it some other way?

pieterlukasse:
@jjgao @zheins I think it should be baked into the code and not easy to change (so not a property). People might be tempted to change the property (lower the version number) instead of running the migration scripts.

zheins:
https://obscure-lowlands-71457.herokuapp.com/

If you'd like to see the warning generated.

jjgao:
Agreed with @pieterlukasse. @zheins has put it in cgds.sql.

zheins:
@jjgao Should we maybe take care of #606 before doing any other tests. I added one to just check that db.version is not null.

zheins:
@pieterlukasse @jjgao I added multiple line support in the migration script, and each versions set of statements are committed in blocks.

jjgao:
@zheins could you resolve the conflicts?

zheins:
@jjgao All conflicts resolved.

core/src/main/resources/db/migration.sql
```diff
@@ -0,0 +1,14 @@
+##version: 1.0.0
+CREATE TABLE info (DB_SCHEMA_VERSION VARCHAR(8));
+INSERT INTO info VALUES ("1.0.0");
+
+##version: 1.1.0
+CREATE TABLE sample_list LIKE patient_list;
+INSERT sample_list SELECT * FROM patient_list;
+CREATE TABLE sample_list_list LIKE patient_list_list;
+INSERT sample_list_list SELECT * FROM patient_list_list;
+DROP TABLE micro_rna;
```
jjgao: @zheins could you remove the drop statements?

pieterlukasse: @jjgao : regarding your last comment to not have drop statements, do you want to have another procedure for this? Just wondering you are planning to handle this scenario.

tweep:
Hello @jjgao

have you considered using Flyway ( https://flywaydb.org/ ) for db migrations and to track schema changes ? I've used it for various other projects and it is very handy and self explanatory. 
Ping me if you want to integrate it and I  can add it and create a pull request.

jjgao:
@tweep Sure. We would be interested in learning about Flyway. Feel free to create a pull request. Thanks!

tweep:
Ok, great. Flyway is pretty awesome as it keeps tracks of all schema changes and does the migration for you itself if your db is missing patches. 

To make this the most useful,we have to understand how cbioportal is currently creating schema patches. @inBrightLight - any thoughts ? 
I suggest 
- we create pull request for flyway integration only for cbioportal ( NOT CGDS for now ) 
- initial request contains only the addition of a single table to cbioportal schema for schema management 
- cbioportal team integrates, after successful integration / testing, the SQL to create basic schema 
- last step is the modification of how the seed-file is created. I suggest to keep the seed file separate, but to take the drop / create table statements out. 

Question: How is the seed file currently produced ?

inBrightLight:
Hi, can you give us any hints how do you perform patching right now? 
Is there any worlflow for it? Flyway is a great tool, widely used across java apps, so I hope you'll make use of it.

PR #706, opened 2016-01-04
Keep code style consistent - related to #639

zhx828:
To keep consistency, use underscore in if condition. @pieterlukasse could you review?

### DIFF
--- a/portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
+++ b/portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
@@ -159,7 +159,7 @@ var StudyViewProxy = (function() {
                 for(var i= 0; i < a1[0].attributes.length; i++){
                     var caseAttr = new CaseAttr();
                     caseAttr.attr_id =  a1[0].attributes[i].attr_id.toUpperCase();
-                    if(a1[0].attributes[i].display_name){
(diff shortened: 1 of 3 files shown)


portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -257,7 +257,7 @@ var StudyViewProxy = (function() {
                 }else {
                     var cnaLength = obtainDataObject.arr.length;
                     for(var i = 0; i < cnaLength; i++) {
-                        if(obtainDataObject.arr[i]['MUTATION_COUNT']) {
+                        if(!_.isUndefined(obtainDataObject.arr[i]['MUTATION_COUNT'])) {
```
pieterlukasse: In line 258, should variable be called cnaLength or something like mutationLength? 

I would also advocate for keeping the same style in referring to attributes, so either ['MUTATION_COUNT'] OR .MUTATION_COUNT. Now you use both mixed up in this and next line.

pieterlukasse:
@zhx828 : minor comment above

zhx828:
@pieterlukasse Good catch, please see the changes in the commit above

pieterlukasse:
@zhx828 OK!

PR #707, opened 2016-01-04
Enable jQuery Tabs with hashtags; Mis-align issue fixed

zhx828:


zhx828:
@jjgao there is a major change about the jQuery tabs with hashtags support. Could you help review? Thanks.

Issue #709, opened 2016-01-05
Old news link to Tissue Images example not working, also page refresh fails


sheridancbio:
On the [News Page](https://github.com/cBioPortal/cbioportal/wiki/News) there is a link to an example for Tissue Images (see [here](https://github.com/cBioPortal/cbioportal/wiki/News#july-25-2013)). The link was to page anchor _#images_, which did not seem to exist. By updating it to _#tab_images_ the browser enters the example page and opens the tissues images tab, but the images do not load.

A manual click on the tab does display images.

To allow an incoming page request with anchor to open the referenced tab and have it be properly displayed may be an enhancement, if that functionality was not working in the past. If this is not enabled, the News page example link should be changed to not use a page anchor, and users should also get an instruction telling them to click the _Tissue Images_ tag.

PR #721, opened 2016-01-08
fix to get tissue_images tab to load directly when url contains a reference to this tab

pieterlukasse:
This is related to issue #709

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/tumormap/patient_view/tissue_images.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/tumormap/patient_view/tissue_images.jsp
@@ -41,7 +41,28 @@
             tissueImageLoaded = true;
         }
     }
-    $("#link-tissue-images").click(loadImages);
+    
+    $(document).ready( function() {
+    	//whether this tab has already been initialized or not:
+    	var tab_init = false;
+    	//function that will listen to tab changes and init this one when applicable:
+    	function tabsUpdate() {
+	        if ($("#tissue-images-div").is(":visible")) {
+		    	if (tab_init === false) {
+		    		loadImages();
+		            tab_init = true;
+		            console.log("tab_images tab initialized");
+		        }
+	    	}
+    	}
+        //this is for the scenario where the tab is open by default (as part of URL >> #tab_name at the end of URL):
+    	window.setTimeout(tabsUpdate,1);
+        //this is for the scenario where the user navigates to this tab:
+        $("#patient-tabs").bind("tabsactivate", function(event, ui) {
+        	tabsUpdate();
+        });
+    });        
+    
 </script>
 
 <div id="tissue-images-div"><img src="images/ajax-loader.gif"/></div>


sheridancbio:
Code works for tissue images tab. One concern:

tissue_images.jpg adds function into document.ready() handler which binds tabsactivate event to image loading function which checks if tab is visible. patient_view.jpg adds function into document.ready() handler which calls switchToTab() function if there is a supplied #tab_* anchor in URL. switchToTab() activates tab for tissue images. So this approach will break if the order of execution of the two document.ready() handlers changes (since the images tab will not be visible yet).

This fix does appear to work, and all tabs seem to display correctly on direct URL access.

There may be a simpler fix in patient_view.jpg ... which I will test. In the switchToTab() function, the test:
    if (toTab==='images') {                                                                                                                                                               
        loadImages();                                                                                                                                                                     
    }                                                                                                                                                                                     

seems to be broken because the tab id changed from 'images' to 'tab_images' at some point. Fixing this test here may fix the problem of loading images without introducing a new handler. If not, we should go ahead with this code.

pieterlukasse:
@sheridancbio good catch! I missed this completely in patient_view.jsp 
Looking at the code in patient_view.jsp, I think that the change below in patient_view.jsp would also suffice
` ` ` 
if (toTab==='images') {   
` ` `
change to 
` ` ` 
if (toTab==='tab_images') {
` ` `

However, I would advocate for keeping the initialization logic of a tab within the files related to this tab. This improves code clarity and maintenance. That means that I prefer to see that the code above in patient_view.jsp is removed and that we keep something similar to what I proposed in tissue_images.jsp . Let me know what you think.

sheridancbio:
There are 8 tabs shown on the example patient-view:
  tab_summary "Summary"
  tab_mutations "Mutations"
  tab_cna "Copy Number Alterations"
  tab_drugs "Drugs"
  tab_clinical_trials "Clinical Trials"
  tab_images "Tissue Images"
  tab_path-report "Pathology Report"
  tab_samples-table "Clinical Information"
There also seem to be 2 tabs which are not shown for the linked example:
  tab_pathways "Network"
  tab_similar-patients "Similar Patients"

Fortunately, each tab <div> includes a distinct jsp:
  tab_summary : summary.jsp
  tab_mutations : mutations.jsp + mutations_table_template.jsp
  tab_cna : cna.jsp + cna_table_template.jsp
  tab_images : tissue_images.jsp
  tab_path-report : path_report.jsp
  tab_pathways : pathways.jsp
  tab_similar-patients : similar_patients.jsp
  tab_drugs : drugs.jsp
  tab_clinical-trials : clinical_trials.jsp
  tab_samples-table : samples_table.jsp

I think the tab initialization should be done only on one level -- either the patient-view.jsp level, or in each lower level included .jsp

Doing the initialization in patient-view.jsp seems to me to reduce code complexity and code duplication. For instance, if a new tab is added to this display, the tab creator would need to know to add tab-initialization code into the included lower level .jsp if initialization were done on the lower level. And if the approach to initializing tabs is changed, it will would need to be changed in all lower level tabs. But one good argument for lower level individual tab initialization code is that some tabs are initialized differently than others. For instance there is a lengthy document.ready handler for cna.jsp which constructs a complicated table (whether or not the tab is visible is seems). Also there is a check for a .pdf viewer browser capability in path_report.jpg. So maybe moving initialization to the lower level makes sense.

So if we wish to move the initialization to the lower level .jsp, some .jsp pages which have no document.ready handler would need to have one added to activate the tab referred to in the URL anchor. Others which already have a document.ready handler should be updated to activate the proper tag. I also think it would be good to (on tab clicks) set the URL to match the anchor for the tab: that way a browser refresh will work correctly (and not take the user back to the original entry point)

Sometimes there seems to be delayed initialization/loading. But I'm not sure I understand what is intended. For example, in drugs.jsp we "loadDrugsClinicalTrialsTable()" on clicks on #link-drugs or #link-clinical-trials .. so I would assume that the population of these tables should not happen until the tab is made visible (either by click, or by direct URL navigation). This is similar to the tissue images. Logic for this delayed loading should be made explicit (good naming, and separation of code which is executed on entry to patient-view from code which is executed on activation of the lower level tab)

Lastly, there is a at least one other way that a user could switch between tabs … by keyboard tabbing the focus to the tab control and then using left and right arrows to change the tab. We should consider adding event handlers so that whichever way the user navigated to the tabs, the initialization and loading of the content occurs properly (and if possible, the URL in the browser address bar would be changed to match the currently open tab.) Currently the tabs Drugs, Clinical Trials, Tissue Images, and Clinical Information fail when navigating this way.

But maybe some of these other issues should be separated into separate pull-requests and addressed later.

sheridancbio:
I can try to work out preliminary code and show you what I am picturing. (It is likely that some errors will happen, since I don't fully understand this code yet)

sheridancbio:
I am working on some sample code now ... trying to remove all of the tab display management from the upper level, and put tab-specific initialization (on page load / tab create, and on tab activate) on the lower level jsp pages.

pieterlukasse:
ok, sounds good. 
- I agree that the downside is some code duplication...maybe you can think of a solution for this as well. 
- I also agree that it is good to set the URL to match the anchor for the tab
- regarding your comment "cna.jsp which constructs a complicated table (whether or not the tab is visible is seems)" : that is exactly what we would like to avoid, i.e. tab specific items should be done only if the tab is selected. So if you can also fix that, it would be great ;)

sheridancbio:
I have worked on this for a few days [also trying to understand my local portal build .. so a bunch of irrelevant debugging was going on].

I'm going to send you my code in email ... it is not yet working, but I'll explain the high level idea here. Feel free to use it, adapt it, or not --- whatever makes sense to you.

----

Originally I thought that *all* initialization code could be moved into the lower level tabs' .jsp files, but changed my mind. There is some initial data construction for the higher level patient-view page itself, and also, there is probably some initialization code needed for the tab control itself which is unrelated to the data on the included tab code. My idea was to make the patient-view jsp "ignorant" of the contents of the included tabs ... just as tissue_images.jsp.

For each lower level tab, there are at least 2 time points to do appropriate initialization. Using the jquery tabs widget terminology (which I tried to learn) there is a "create" event and an "activate" event. There is also a "load" event, but I don't fully understand that --- it sounds like it is tied to an auto-load ability of the widget to pull remote content into the tab content div. I don't think we use that feature, so I'm ignoring it.

Instead of a create handler, each tab would attach a document.ready handler which would do any primary initialization if the tab was referenced in the URL using the tab's anchor name. [if a create handler makes more sense though, feel free to switch this functionality into there instead of document.ready]

Expensive / Unnecessary initialization for not yet visible tabs would be done by the "activate" event handler. If the tab anchor is present in the URL on page load, then we can force an activate event using the .tabs("option", "active", <index>) method of the widget during page load (via a scheduled call to an activateTab() function). For tabs without the matching anchor in the URL, they would not call their own activateTab() functions .. and would just register the tabsactivate handler for future navigation.

In patient-view.jsp, there would simply by a catch-all initialization of the tabs widget ... to get the widget initialized in the case that none of the anchors were seen in the URL. Alternatively, this could be put into the default tab (tab_summary) document.ready handler. That code is here:
function setUpPatientViewTabs() {
    $('#patient-view-tabs').tabs({
        show: false
    }); //initialize, and with no animation effects
    fixCytoscapeWebRedraw();
}

I'm sending you code via email ... note that somehow I broke the actual image loading on the tissue_images.jsp, and have not yet applied this approach to the other tabs; just the tissue_images.jsp include.

There were two tabs that I played with but I guess they are not actually live yet, so I ignored network and similar-patients. I also did some renaming and text formatting, but I'm dropping most of that from the diff so you can see the important changes.
Renamings:
patient-tabs -> patient-view-tabs
patient-section -> patient-view-section
setUpPatientTabs -> setUpPatientViewTabs

jjgao:
#793 was a quick fix for this. We will put some framework in place later for a better solution.

PR #753, opened 2016-01-14
Volcano plot part1

aderidder:
Volcano Plot (ploty) + Mini-onco functionality added to the Enrichtments tab's subtabs

@yichaoS since you're the author of the over representation analysis view and also working with plotly, are you willing to do the code review?


yichaoS:
Some suggestions:

@aderidder I tried the box selection to select a subset of the dots, but I couldn't figure out a good way to deselect it except refresh the page all together; and I don't find any reset button for that. Maybe a more convenient UI design for this part? 

@aderidder maybe we can make it more explicit when click on a gene in the table, some bar charts will pop up? I found that feature by accident....otherwise I wouldn't notice. Maybe by default selecting the first gene and show the bar chart from the very beginning?  

@jjgao Do you think maybe we could just get rid of q-value? And we can get rid of "tendency towards" and just say "co-occurrence"/"mutual exclusive"? Just this page is getting really crowded.

aderidder:
Hi @yichaoS thanks for the feedback

Concerning your first point: yes you're right. Unfortunately I haven't been able to find a good way to do this in the current plotly version. I you'd like more details, let me know and I'll elaborate. I've created an issue for this in the plotly github (https://github.com/plotly/plotly.js/issues/181). 

Concerning the second issue: I've made some changes and now the mini-onco is visible straight away with the second bar suggesting that the user can select a gene in the table or in the plot. The new version is available on our test server. 

Thanks,
Sander

jjgao:
@aderidder very nice work!

About Yichao's first point, I think double click on the plot reset the selection, but the table is not updated. 

It's good idea to remove "tendency towards" in the last column. Let's still keep the q-value there now.

aderidder:
Hi @jjgao thanks, glad you like it!
I think that to fix the issue we'll have to change the plotly library. The problem is with the selection box and the 'plotly_selected' event. On a single-click, the selection box stays, whereas on a double-click the selection box is removed. However, the 'plotly_selected' event in both cases ends up with 'undefined' eventData, so there is currently no distinction between a single-click and a double-click there. 
Currently the behaviour is: if eventData is undefined, don't do anything. This makes sense if you single-click, as the selection box still surrounds the points, but it doesn't make sense for the double-click, as the selection box is gone. If I change the code to: if eventData is undefined, remove the filtering, then it makes sense for the double-click, but not for the single-click. 
Interested in hearing your thoughts.

jjgao:
@yichaoS do you have any other comments? If not, I'll move it to beta testing.

yichaoS:
@jjgao I think it looks good.

portal/src/main/webapp/js/src/over-representation-analysis/view.js
```diff
@@ -253,6 +284,86 @@ var orTable = function() {
...
+        return profile_type === orAnalysis.profile_type.mutations || profile_type === orAnalysis.profile_type.copy_num;
+    }
+
+    /**
+     * check whether the datatype is LOG-VALUE to prevent logging it again
+     * @returns {boolean}
+     */
+    this.hasLogData = function(){
+        return data_type === "LOG-VALUE";
+    }
```
pieterlukasse: @jjgao : be aware this could be a pitfall! You need to check if all the relevant genetic profiles on your server have the data type correctly set to LOG-VALUE, otherwise they will not be selected here!

PR #758, opened 2016-01-14
Alteration Count API - RFC17: Alteration Count API

JiaoJiao123:
Speed test for the POST method in postman
I tested with 1500 input records (gene, start and end) and the result is as below.
If per study is false, it took around 100ms to return 1500 records. If per study is true, it took around 400ms to return around 20K records.

GET method
/api/mutation?type=count&per_study=false&inputIDs=firstRecord,secondRecord&genes=GAGE10,PRR33&starts=1,1&ends=253303360,209223700


jjgao:
Thanks, @JiaoJiao123! Could you fix the following issues and send a pull request to rc branch? 

 - [ ] add echo parameter to allow use-defined ID
 - [ ] change /api/mutation? to /api/mutation_count?

zhx828:
@JiaoJiao123 Doesn't DBAltCount need get/set functions?

JiaoJiao123:
@zhx828 yeah right, I will add get and set functions. Thank you!

zhx828:
@JiaoJiao123 You should send pull request to rc, right? @jjgao

PR #761, opened 2016-01-15
Alteration Count API - RFC17: Alteration Count API

JiaoJiao123:
Please follow the RFC link below for detailed API description
RFC: https://docs.google.com/document/d/1V--MU2-vQMhWhTF1oy-FAOE0hFoT7WXf1ft4mR443U0/edit

the GET method can be accessed like below
/api/mutation_count?type=frequency&per_study=true&gene=BRAF,EGFR&start=1,1&end=10000,20000&studyId=acc_tcga,brca_bccrc,brca_bccrc_xenograft_2014&echo=gene

the input data for POST method should be in the format as below:
{
              "type": "frequency",
              "per_study": true,
             "echo": ["gene", "end", "myID", "end"],
             "studyId": ["brca_tcga_pub","acc_tcga","brca_tcga"],
             "data": [ {"myID":"firstID","gene": "BRAF", "start": "1", "end": "10000"}, 
                          {"myID":"secondID","gene": "EGFR", "start": "1", "end": "20000"} ]
}


business/src/main/java/org/mskcc/cbio/portal/model/DBAltCount.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class DBAltCount implements Serializable {
+    public int count;
+    public int studyID;
+    public String id;
+    public String gene;
+    public int start;
+    public int end;
+
+    public int getCount() {
```
adamabeshouse: since all these properties are public (and will stay so), no need to have getters and setters I think

business/src/main/java/org/mskcc/cbio/portal/model/DBAltCountInput.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author jiaojiao
+ */
+public class DBAltCountInput implements Serializable {
+    
+    public String type;
+    public Boolean per_study ;
+    public List<DBAltCountInputData> data; 
+    public List<String> echo;
+
+    public List<String> getEcho() {
```
adamabeshouse: Same as above, I don't think we need getters and setters

business/src/main/java/org/mskcc/cbio/portal/model/DBAltCountInputData.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author jiaojiao
+ */
+public class DBAltCountInputData implements Serializable {
+    public String gene; 
+    public String id;
+    public Integer start;
+    public Integer end ;
+
+    public String getGene() {
```
adamabeshouse: same as above

business/src/main/java/org/mskcc/cbio/portal/persistence/MutationMapper.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+import org.apache.ibatis.annotations.Param;
+import org.mskcc.cbio.portal.model.DBAltCount;
+
+/**
+ *
+ * @author jiaojiao
+ */
+public interface MutationMapper {
+    List<DBAltCount> getMutations(@Param("gene") String gene, @Param("start") Integer start, @Param("end") Integer end);
```
adamabeshouse: maybe, for clarity, ```getMutationCounts``` and ```getMutationCountsPerStudy``` ?

adamabeshouse:
Looks great! Only small comments. :+1:

business/src/main/java/org/mskcc/cbio/portal/model/DBAltCount.java
jjgao: Or maybe it's better to make them private.

adamabeshouse: Either way, though AFAIU this is just the container for SQL results so might not be necessary

JiaoJiao123:
@adamabeshouse  Thank you very much for your comment.
yeah I guess the gettera and setters are unnecessary for those public attributes. I will remove them. Also I will rename the functions to make them more make sense.

jjgao:
@onursumer could you also review this and see if it covers all the use cases in the mutations table.

business/src/main/java/org/mskcc/cbio/portal/model/DBAltCount.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ *
+ * @author jiaojiao
+ 
+@JsonInclude(JsonInclude.Include.NON_NULL)*/
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+
+public class DBAltCount implements Serializable {
+    public int count;
+    public int studyID;
```
jjgao: We should return the stable study ID instead of internal study ID.

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -74,6 +80,165 @@
 		return cancerTypeMapper.getCancerTypes(cancer_type_ids);
 	}
 
+        @Transactional
+	public List<DBAltCount> getMutations(String type, Boolean per_study, List<String> ids,  List<String> genes, List<Integer> starts, List<Integer> ends, List<String> echo) {
```
jjgao: The code here should not assume there are IDs. The user may not provide an ID. They can provide anything in the URL to echo back.

jjgao: The method name should be renamed: maybe to getAlterationCount

jjgao: Can we put the parameters in hashmap instead of list them all so that we can incorporate any other parameters fro the user easily?

web/src/main/java/org/mskcc/cbio/portal/web/ApiController.java
```diff
@@ -45,6 +47,19 @@
...
 
     @Transactional
+    @RequestMapping(value = "/mutation_count", method = {RequestMethod.GET})
+    public @ResponseBody List<DBAltCount> getMutations(@RequestParam(required = true) String type, @RequestParam(required = true) Boolean per_study, @RequestParam(required = true) List<String> ids, @RequestParam(required = true) List<String> genes, @RequestParam(required = true) List<Integer> starts, @RequestParam(required = true) List<Integer> ends, @RequestParam(required = false) List<String> echo) {
+        return service.getMutations(type, per_study, ids, genes, starts, ends, echo);
+        
+    }
+    
+    @Transactional
+    @RequestMapping(value = "/mutation_count", method = {RequestMethod.POST})
```
jjgao: Please try to follow the spec in the doc including the parameter names.

For example, can we support post of json? e.g.

```
[
  { 
    ‘data’: [
        {‘gene’: ‘geneA’,
          ‘startPos’: <startPosA>,
          ‘endPos’: <endPosA>,
          ‘myid’: ‘idA’},
          …
      ],
    ‘type’: ‘count’,
    ‘category’: ‘none’ or ‘study’,
    ‘echo’: [‘myid’]
  },
  {...},
]
```

onursumer:
Current PancancerMutationsJSON servlet allows querying by gene symbols only: http://www.cbioportal.org/pancancerMutations.json?cmd=byHugos&q=GAGE10,PRR33

I think the equivalent query should be something like this: https://cbioportal-pr-761.herokuapp.com/api/mutation_count?type=count&per_study=false&gene=GAGE10,PRR33&id=firstRecordID,secondRecordID&echo=id

But it complains about a missing start parameter. Is there currently a way to query with gene symbols only?

JiaoJiao123:
@onursumer For now, gene, start and end are required parameters for the API. But I think it's a good idea to allow query only by gene. I will work on that and thank you for your suggestion.

jjgao:
@adamabeshouse has another API call to get gene level counts. @adamabeshouse we should integrate that into this API.

JiaoJiao123:
@jjgao @onursumer 
I made the start and end parameters optional by setting default value to them.
If user didn't input start and end, we will use 0 as start and 1,000,000,000 as end, which I assume any start-end region will fall into. Let me know if any issues.

adamabeshouse:
@JiaoJiao123 would it be possible to use infinity instead of 1,000,000,000?

onursumer:
I guess <code>Integer.MAX_VALUE</code> can also be used in place of infinity in Java.

JiaoJiao123:
Thank you for you guys' great suggestion! I set the default end position value to be Integer.MAX_VALUE

adamabeshouse:
It's not a big deal, but I think Integer.MAX_VALUE currently looks like a magic number in the service output. I think for clarity it should either be the length of the gene, or maybe if no position is specified, there should be no start or end attribute in the return object.

jjgao:
Agreed with @adamabeshouse, for query without positions, the position fields should not be queried from database.

JiaoJiao123:
@jjgao @adamabeshouse 
I added some queries for input without start and end values and removed the default position setting which might be confusing.

adamabeshouse:
@JiaoJiao123 I think it would still be useful to have support for just putting the start or just putting the end position.

JiaoJiao123:
@adamabeshouse Yeah right. I think so too. I will add the feature to support only start input or end input.

JiaoJiao123:
I realized we can use dynamic SQL in mybatis, which enable me to write one dynamic SQL based on parameters instead of 8 messy SQL statements. Support for only start or only end position is added. check it out and let me know if any other suggestions. @adamabeshouse

Issue #772, opened 2016-01-21
Importing a mutation sample without clinical data causes issues


j-hudecek:
If mutation data contains samples which are not listed in the clinical data file cBioPortal counts mutations in the sample but doesn't add them to the total. This causes genes to have (on study view) mutation frequency > 100% and when a gene is mutated only in these samples selecting it in the mutated genes view results in 0 samples selected. 

Importing mutation sample without clinical data should be an error.

jjgao:
@zheins we should add this to the validator.

PR #773, opened 2016-01-21
Stop import of mutation data on encountering unknown sample

j-hudecek:
Solves #772 
I've also removed an assert - by default asserts are not executed (you need to specify -ea) so an if is better


jjgao:
@n1zea144 could you review this change?

zheins:
Do we really want to stop importing on finding an unknown sample? @n1zea144 what do you think? My opinion is throw a warning and keep going if possible.

jjgao:
@zheins @j-hudecek: it think it's a good idea to throw a warning.

jjgao:
@n1zea144 @zheins let me know what you'd like to do with this? It's been there for too long.

PR #722, opened 2016-01-08
Replace dataTable with react table in study view  clinical tab - DO NOT MERGE

zhx828:

PR #779, opened 2016-01-22
Replace dataTable with react table in study view clinical tab 

zhx828:
Using Enhanced Fixed Datatable to replace DataTable library in Study View Clinical Tab

Notice: This pull requests contains some libraries which reply on Bootstrap. It needs to be tested after #775  having been merged.


zhx828:
This relates to #722

PR #790, opened 2016-01-25
Quick fix for proper direct URL access to tab_images (Tissue Images)

sheridancbio:
- fix element name references in javascript code to match actual element names
- also corrected pathways tab, although this is not currently in use


sheridancbio:
This is a short term fix to allow direct URL access to the tissue images tab on the patient view to load properly.

Other improvements are needed, such as adopting the approach of only doing expensive tab loading when tabs are visible. (see https://github.com/cBioPortal/cbioportal/pull/642, https://github.com/cBioPortal/cbioportal/pull/721)

jjgao:
@sheridancbio could you send another pull request to hotfix and close this one?

sheridancbio:
will do

Issue #801, opened 2016-01-29
mysql server slowdowns occurring : high CPU load during mutation queries


sheridancbio:
Slowdowns - sometimes for several minutes seem to be occurring in the mysql server.
"show processlist" on mysql show queries running which involve mutations ... in particular, joins of the mutation table to the mutation_event table.

sheridancbio:
The current query suspected of causing slowdowns is:
DaoMutation : countSamplesWithProteinPosStarts()
with SQL: 
  SELECT ONCOTATOR_PROTEIN_POS_START, ...
    FROM mutation, mutation_event
    WHERE GENETIC_PROFILE_ID IN (... list of profile ids ...)
    AND mutation.MUTATION_EVENT_ID=mutation_event.MUTATION_EVENT_ID
    AND (mutation.ENTREZ_GENE_ID, ONCOTATOR_PROTEIN_POS_START) IN (... list of gene,pos pairs)
    GROUP BY ONCOTATOR_PROTEIN_POS_START, GENETIC_PROFILE_ID


The problem is that for queries across most/all studies, nearly the entire (3M) mutation table is returned in a temporary result table before joining with the other large (2M) mutation_event table and being filtered by gene,pos

sheridancbio:
Improvement was applied with pull request #803. This reduced the size of the mutation table by using filtering based on geneId. The join operation to mutation_event happened in about one tenth the time for the worst case observed query. However, in the time since the fix there have been moments of high mysqld load observed when 10 parallel SQL queries which merge mutation to mutation_event are happening. Further performance improvements may be needed in the future. For now, the excessive delays in webserver response do not seem to be happening.

PR #803, opened 2016-01-29
Fix DaoMutation call countSamplesWithProteinPosStarts to help with mysql slowdown

sheridancbio:
This will partially address issue #801

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -1193,10 +1193,10 @@ public final class DaoMutation {
 		ResultSet rs = null;
 		try {
 			con = JdbcUtil.getDbConnection(DaoMutation.class);
-
-			String sql = "SELECT ONCOTATOR_PROTEIN_POS_START, GENETIC_PROFILE_ID, mutation.ENTREZ_GENE_ID, count(DISTINCT SAMPLE_ID) FROM mutation, mutation_event " +
-			             "WHERE GENETIC_PROFILE_ID IN (" + StringUtils.join(internalProfileIds, ",") + ") " +
-			             "AND mutation.MUTATION_EVENT_ID=mutation_event.MUTATION_EVENT_ID " +
+                        //TODO: create where clause for mutation table to allow use of ENTREZ_GENE_ID index to filter results
+                        String sql = "SELECT ONCOTATOR_PROTEIN_POS_START, GENETIC_PROFILE_ID, mutation.ENTREZ_GENE_ID, count(DISTINCT SAMPLE_ID) FROM mutation, mutation_event " +
+                                     "WHERE GENETIC_PROFILE_ID IN (" + StringUtils.join(internalProfileIds, ",") + ") " +
+                                     "AND mutation.MUTATION_EVENT_ID=mutation_event.MUTATION_EVENT_ID " +
 			             //"AND concat(concat(mutation.ENTREZ_GENE_ID, '_'), ONCOTATOR_PROTEIN_POS_START)" +
 			             //"IN ('" + StringUtils.join(proteinPosStarts, "','") + "') " +
 			             "AND (mutation.ENTREZ_GENE_ID, ONCOTATOR_PROTEIN_POS_START)" +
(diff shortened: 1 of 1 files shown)


sheridancbio:
From the mysql logs, a reproducible "Slow" query was found and rewritten to avoid large temporary tables before merging.

From the logs, the query was:
SELECT ONCOTATOR_PROTEIN_POS_START, GENETIC_PROFILE_ID, mutation.ENTREZ_GENE_ID, count(DISTINCT SAMPLE_ID) FROM mutation INNER JOIN mutation_event ON mutation.MUTATION_EVENT_ID=mutation_ev
ent.MUTATION_EVENT_ID WHERE GENETIC_PROFILE_ID IN (2079,4534,3722, ... ) AND (mutation.ENTREZ_GENE_ID, ONCOTATOR_PROTEIN_POS_START)IN ((10395,1510),(7259,193),(51421,558), ...);
Using mysql "EXPLAIN", it showed that no table index would be used when querying the mutation table:
+----+-------------+----------------+--------+--------------------------------------------------------------------------------+---------+---------+--------------------------------------+---------+-------------+
| id | select_type | table          | type   | possible_keys                                                                  | key     | key_len | ref                                  | rows    | Extra       |
+----+-------------+----------------+--------+--------------------------------------------------------------------------------+---------+---------+--------------------------------------+---------+-------------+
|  1 | SIMPLE      | mutation       | ALL    | GENETIC_PROFILE_ID,GENETIC_PROFILE_ID_2,GENETIC_PROFILE_ID_3,MUTATION_EVENT_ID | NULL    | NULL    | NULL                                 | 2847650 | Using where |
|  1 | SIMPLE      | mutation_event | eq_ref | PRIMARY                                                                        | PRIMARY | 4       | dbname.mutation.MUTATION_EVENT_ID |       1 | Using where |
+----+-------------+----------------+--------+--------------------------------------------------------------------------------+---------+---------+--------------------------------------+---------+-------------+

Here the "ALL" type of query means that the full table content of mutation (3M records) would be extracted and then filtered by a WHERE clause iteration testing for GENETIC_PROFILE_ID in a (longish) list. The problem is that given the set of profileIds in the IN() clause, almost all mutations were pulled in to the temporary table (2847606 out of 2847650)

When the server was relatively quiet, this mysql query took 11 seconds to complete. At other times when the mutation table was being queried by other threads, it could be much slower.

Example output from "show processlist":
+--------+------------+-----------------+-------------+---------+------+----------------------+------------------------------------------------------------------------------------------------------+
| Id     | User       | Host            | db          | Command | Time | State                | Info                                                                                                 |
+--------+------------+-----------------+-------------+---------+------+----------------------+------------------------------------------------------------------------------------------------------+
| 685767 | userid | localhost:51434 | dbname | Query   |  140 | Copying to tmp table | SELECT ONCOTATOR_PROTEIN_POS_START, GENETIC_PROFILE_ID, mutation.ENTREZ_GENE_ID, count(DISTINCT SAMP |

Since the mutation table has a field ENTREZ_GENE_ID, and the argument to the DaoMutation function (countSamplesWithProteinPosStarts) includes a set of geneIds to filter for, I have added a separate WHERE clause to the SQL statement of the form:

WHERE mutation.ENTREZ_GENE_ID IN (10395, 7259, 51421 ...)

This allows filtering to be done using the index on ENTREZ_GENE_ID. The size of the resulting tmp table is reduced to 37212 rows (about 100 times smaller). And the query runs in about 1 second (10 times faster).

Explain for query with additional WHERE clause:
+----+-------------+----------------+--------+-----------------------------------------------------------------------------------------------+----------------+---------+--------------------------------------+-------+-------------+
| id | select_type | table          | type   | possible_keys                                                                                 | key            | key_len | ref                                  | rows  | Extra       |
+----+-------------+----------------+--------+-----------------------------------------------------------------------------------------------+----------------+---------+--------------------------------------+-------+-------------+
|  1 | SIMPLE      | mutation       | range  | GENETIC_PROFILE_ID,GENETIC_PROFILE_ID_2,GENETIC_PROFILE_ID_3,ENTREZ_GENE_ID,MUTATION_EVENT_ID | ENTREZ_GENE_ID | 4       | NULL                                 | 37212 | Using where |
|  1 | SIMPLE      | mutation_event | eq_ref | PRIMARY                                                                                       | PRIMARY        | 4       | dbname.mutation.MUTATION_EVENT_ID |     1 | Using where |
+----+-------------+----------------+--------+-----------------------------------------------------------------------------------------------+----------------+---------+--------------------------------------+-------+-------------+

Issue #528, opened 2015-11-20
Improvements for the data loading scripts


pieterlukasse:
A number of improvements are needed to make the process of loading new datasets into cBioPortal more robust and simpler. These are listed below:
- [x] Improve validation of input data files prior to loading them. These are being tracked in #479 
- [x] simplify the use of cbioportalImporter.py: 
 - [x] Once all necessary files are ready, one still needs to call the importer script in a specific order. This can be either simplified or the process could guide the user with better validation/error messages. 
Order should probably be: meta_study,clinical data,(per-gene copy number data?),everything else, case lists
 - [x] a calling routine for running the whole pipeline (preferably in Python), connecting the validation and loading parts

- [x] reporting improvements: 
 - [x] the process in general should be more verbose, reporting more details at each step of the loading process. 
 - [x] validation messages need to be improved for easier troubleshooting

pieterlukasse:
another improvement is #494

pieterlukasse:
In general, these steps are implemented. We are now in the phase of testing and rework based on issues found, or small enhancements

pieterlukasse:
largely solved by #1020, #1074

PR #807, opened 2016-02-02
Added sorting of warnings during import and displaying them with counts

j-hudecek:
This pull request changes the way warnings are handled during data import. ProgressMonitor now stores them in a TreeSet (stores each warning only once in order in log(n)). At the end of import they are output sorted by alphabet with a count of how many times they occured during import. Example:

    Gene not found:  TXNP7 [0]. Ignoring it and all mutation data associated with it!; 2x
    Gene not found:  U3 [0]. Ignoring it and all mutation data associated with it!; 12x
    Gene not found:  U40455.1 [0]. Ignoring it and all mutation data associated with it!; 2x
    Gene not found:  U66059.29 [0]. Ignoring it and all mutation data associated with it!; 2x
    Gene not found:  U8 [0]. Ignoring it and all mutation data associated with it!; 4x

This limits the output of the command and prints similar pseudogenes close to each other.
This pull request also propagates ProgressMonitor to getNonAmbiguousGene and     getSampleByCancerStudyAndSampleId methods as an optional parameter, if it specified errors and warnings are stored in the ProgressMonitor, otherwise System.err is used.
ProgressMonitor now also writes the warnings to a log file as the output can be sometimes overwhelming the console.


core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -338,7 +339,16 @@ public CanonicalGene getNonAmbiguousGene(String geneId) {
...
+    public CanonicalGene getNonAmbiguousGene(String geneId, String chr) { return getNonAmbiguousGene(geneId, chr, null); }
+
+        /**
+         * Look for gene that can be non-ambiguously determined.
+         * @param geneId an Entrez Gene ID or HUGO symbol or gene alias
+         * @param chr chromosome
+         * @param pMonitor for import warnings
+         * @return a gene that can be non-ambiguously determined, or null if cannot.
+         */
+    public CanonicalGene getNonAmbiguousGene(String geneId, String chr, ProgressMonitor pMonitor) {
```
jjgao: I don't like to put pMoniter in the function signature. Maybe we should have a ProgressMonitor singleton class that script and dao classes can call? And the singleton class is only initialized when importing (not the web code).

j-hudecek: I agree that singleton would be useful. The getUnambiguousGeneName is also called on tomcat so we have to make sure that case is also handled - only print the warning to the log


________________________________

j-hudecek: So I looked into making it a singleton and while it makes sense it is a really big change that touches almost a hundred of files and not all the changes are straightforward.
Are you sure you want me to do this?

For deciding whether we are running in a web server or not this method could be used: http://docs.liferay.com/portal/5.1/javadocs/portal-kernel/com/liferay/portal/kernel/util/ServerDetector.java.html

j-hudecek:
So I've updated the pull request to make ProgressMonitor a singleton, what do you think of it now?

jjgao:
Thanks @j-hudecek! I think it's great.

@zheins could you also review the change?

zheins:
@j-hudecek I really like these changes, great job! @jjgao @n1zea144, this PR requires some changes in our pipelines, which I went ahead and made.

zheins:
@j-hudecek, could you merge the conflicts?

j-hudecek:
thanks, this should work right?

fedde-s:
@pieterlukasse might want to see if this overlaps any of the work he planned or did in making the output of the import process' Java layer more user-friendly, as part of issue https://github.com/cBioPortal/cbioportal/issues/528

fedde-s:
Looks great though!

jjgao:
I think this looks great. Thanks, @j-hudecek! I'll merge it now.

@pieterlukasse if you have any further improvements, we can address them in another pull request.

pieterlukasse:
@j-hudecek , @jjgao : awesome news, thanks! We actually have a client asking for improvements in this part, so I will definitely merge it in our branch and test it as well! I will make a new PR in case of improvements from our side.

PR #834, opened 2016-02-10
display multiline clinical attributes

j-hudecek:
and , set max-width on sample clinical attribute value to 800px.
This allows us to use very long clinical attributes (pathology reports) for samples and have sane formatting 
image


zhx828:
@jjgao ready to merge

Issue #814, opened 2016-02-03
incorrect # of profiled samples in the header of mutations and cna tables of study view


jjgao:
![image](https://cloud.githubusercontent.com/assets/840895/12793295/e94f0ccc-ca7d-11e5-96ca-9bbcde406d9e.png)

PR #841, opened 2016-02-11
Filter out blank sample ids

zheins:


zheins:
Fixes #814

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -221,11 +221,11 @@ var StudyViewProxy = (function() {
                         var _parts = _lists[i].split('\t');
                         if(_parts.length < 5) continue;
                         if (_parts[0] === parObject.studyId+"_sequenced") {
-                            _sequencedSampleIds = _parts[4].split(' ');
+                            _sequencedSampleIds = _parts[4].split(' ').filter(Boolean);
```
jjgao: maybe do a trim() before split instead of filter() to avoid the loop?

portal/src/main/webapp/js/src/study-view/data/StudyViewProxy.js
```diff
@@ -221,11 +221,11 @@ var StudyViewProxy = (function() {
                         var _parts = _lists[i].split('\t');
```
pieterlukasse: @zheins @jjgao : this part could at least use some inline comments explaining why we skip when _parts.length < 5, why we can expect an empty space at the end of the line and what are parts [0] to [4]....could you please add it so that future troubleshooting is easier?

zheins: The structure of that object, which is based on the case lists:

0: "case_list_id"
1: "case_list_name"
2: "case_list_description"
3: "cancer_study_id"
4: "case_ids"

If the length isn't 5, there is something wrong with the case list and the sample Ids won't be able to be extracted.

I just hit the webservice for case lists with wget (http://www.cbioportal.org/webservice.do?cmd=getCaseLists&cancer_study_id=brca_tcga), and it looks like there is an extra space at the end of each case list returned. Looking in the database (sample_list_list) there are no blank entries. @jjgao should we be doing anything else for this or just trim the response from the webservice?

pieterlukasse: @zheins thanks for the details. If you add these as comments to the code I think it is already an improvement. Another improvement would be to give an error message in case the length isn't 5 (since you indicate that in this case there is something wrong - or maybe we can add a strict validation to this, i.e. does it happen when one of these 5 fields is missing in DB record? ).

Issue #828, opened 2016-02-09
Submit button on the query page can be edited in the browser


schultzn:
The Submit button on the main page can be edited - see the screenshot below. 
Once edited, the button no longer works.

![image](https://cloud.githubusercontent.com/assets/11892704/12927310/8990e23a-cf35-11e5-98a2-37f1b1cfd3a0.png)

PR #845, opened 2016-02-12
submit button style relative to issue #828

JiaoJiao123:
Cause: submit button style was not specified, which made it editable as input element.
Fix: set type value to be submit and set width and height to adjust button size


JiaoJiao123:
@schultzn @jjgao 
I specified the input type to be submit. It should be fine now.
Please check it out and let me know if any issue
https://cbioportal-pr-845.herokuapp.com/

jjgao:
Looks good. Thanks, @JiaoJiao123! fixed #828

zhx828:
The code looks good. But I would suggest to move the style into css file.

JiaoJiao123:
@zhx828 Thanks for your suggestions! I moved the css style to the global_portal.css
https://cbioportal-pr-845.herokuapp.com/

zhx828:
@JiaoJiao123 I would suggest to rename the class name with more specific meaning one. Current name may easily causes duplication issue since it is in global css.

jjgao:
@JiaoJiao123 could you address this as well?

JiaoJiao123:
@jjgao @zhx828 
class name changed to homePageSubmitButton

zhx828:
@JiaoJiao123 there is no need to give the size. I've updated to Bootstrap button. Please see commit https://github.com/zhx828/cbioportal/commit/87c12454bf4683be9692d43cb979700d0eaa0115

jjgao:
oops I merged it before see @zhx828's comments. @JiaoJiao123 could you create another pull?

JiaoJiao123:
@zhx828 
so just use the bootstrap class style "btn btn-default btn-xs" for the submit button, and no need to write our own css at all right?

JiaoJiao123:
@jjgao I will create another pull request

zhx828:
@JiaoJiao123 yes. I already changed that in my merged pull request. You can just remove your newly styles and do a test.1

PR #851, opened 2016-02-17
Add configuration for pathology tissue images

j-hudecek:
Currently "Tissue images" tab only appears for patients with id starting with "TCGA-" (string in code) for who a particular meta url returns 200 and html that when matched against a pattern (hardcoded) produces a number >0 in the first match group.
This pull request moves these two constants into the properties file so that institutions can use it for non-TCGA samples.


jjgao:

PR #866, opened 2016-02-24
Add gene panels

zheins:
Gene panels added to database, import script, api, and oncoprint.


n1zea144:
I don't think we should roll gene panel data into the apiController class.  Can we put it back into a separate controller?  In fact, the apiController should probably be broken up (another discussion).

jjgao:
+1 on break apiController. There's no reason to put all into one controller.

jjgao:
@zheins could you add some sql to the migration code and increase the version to 1.2.0?

adamabeshouse:
@zheins lgtm

@jjgao,@n1zea144: Why prefer more controllers/services? The controller just passes calls back to the service, which delegates to the relevant mapper.

jjgao:
@zheins @n1zea144 Let's push this forward and release it in 1.2.0.

jjgao:
@zheins what's left on this pull request?

zheins:
@jjgao, nothing. I'd just have to resolve the conflicts I think.

jjgao:
@zheins when merging, could you increase the db version to 1.2.2?

zheins:
@jjgao this PR now breaks the oncoprint. Since Adams new implementation was merged in, the gene panel information isn't being handled in the proper manner. I'll try to get this done by the end of the week.

zheins:
Closing this for now unmerged - new pull request for the api, db schema, and import: https://github.com/cBioPortal/cbioportal/pull/1194

Issue #822, opened 2016-02-05
Validator: enable 3 modes of validation


pieterlukasse:
The validator script should have the following 3 modes:

- [x] \(1) Basic checks mode: validator only checks the basic things in the files, like format, presence of required fields, valid sample ids. Basically all standalone checks that do not require external data.
- [x] \(2) All checks mode, with external data in static files: validator does all checks done in 1 plus extra checks to validate the used gene symbols, the specified clinical attributes and the given cancer type (if given). For these checks, it expects gene, clinical attributes and cancer types data to be available in a set of predefined local files. An extra script should also be added to cbioportal to easily generate these files to be shared with other users.
- [x] \(3) All checks mode, with external data coming from cbioportal server: same as 2, but external data is retrieved via the /api/* webservices from the local cbioportal server. **This is the way the validator currently works**

fedde-s:
@pieterlukasse @priti88 @ecerami In commit thehyve/cbioportal@8c67f3f141dba43ec22e8cb57a1c3bd762fbecb2 I merged in the `--no_portal_checks` flag to make the validation script skip any tests requiring data from the portal, i.e. 'mode 1'. These commits are included in pull request #798 which is under review by @priti88.

priti88:
@fedde-s : Mode 1 looks great except for one comment:
1) Mutation file throws error for missing columns: Variant_Type, Protein_position, SWISSPROT
These does not seem to be required columns and user might not have them.

fedde-s:
@priti88 These columns were established to be required by cBioPortal in an email conversation with Ben and Onur, please see and comment on issue #806

priti88:
@ecerami Any thoughts?

fedde-s:
[...]

PR #880, opened 2016-03-01
Validator and loader prelastrc

fedde-s:
Hi @priti88, I hadn't sent a pull request yet, but here's the script to dump the JSON files required by the validator in one go. It uses the Java layer (reusing the functions that serve the information to the web API itself), so you'll have to recompile. After that, you can `cd` to `core/src/main/scripts/`, set and export the `JAVA_HOME` and `PORTAL_HOME` environment variables as appropriate, and run `./dumpPortalInfo.pl`. The usage message should be clear enough, and I will also document the functionality on The Hyve's fork of the wiki.  @pieterlukasse @ecerami


fedde-s:
And I believe this closes issue #880.

fedde-s:
And I believe this closes issue #822.

PR #885, opened 2016-03-02
Remove MYCANCERGENOME_URL, use local file only

inodb:
The portal uses a local tsv file of mycancergenome. This removes
MYCANCERGENOME_URL references. Includes two other fixes i.e.: fix the EXAMPLE
properties file such that heroku/travis shows mycancergenome, and fix links in
mycancergenome tsv file (were pointing to cbioportal instead of
mycancergenome).


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -114,7 +114,7 @@ oncokb.url=
 show.hotspot=true
 
 # Link to My Cancer Genome. Please disable (set to false) when using cBioPortal with patient identifiable data due My Cancer Genome license restrictions.
-mycancergenome.show=false
+mycancergenome.show=true
```
jjgao: let's not turn on this by default.

jjgao:
Thanks for take care of this. It was the fault of my sloppy code..

jjgao:
@inodb I'll merge the code now to fix the auto build. Don't worry about disable mycancergenome if you haven't started.

inodb:
:ok_hand:

PR #892, opened 2016-03-03
prevented 3D panel resize event to also trigger window resize event

onursumer:
This fixes a bug where mutation table re-renders itself every time 3D panel is resized


onursumer:
With the revision 95c4bc3, now the conflict between jquery resizable plugin and latest bootstrap update has been resolved.

See https://github.com/cBioPortal/mutation-mapper/pull/28 for details.

PR #896, opened 2016-03-04
Classifier Web API

angelicaochoa:


n1zea144:
- I don't think we need the classifier class - we can just return the JSON as a string (but we will probably have to change the return type out of the controller by adding - produces = "application/json; charset=utf-8" to the @RequestMapping tag).
- Why is getInternalIdBySampleId(@Param("sampleId") String sampleId); is on the ClassifierMapper?
- Whats SampleMapper.java?
- Whats this for on the Classifier service: getInternalIdBySampleId?
- getClassifierDataBySampleId(@PathVariable int internalId) on the Controller should take a stable sample id (string)
- Do you have an ImportClassifier.java class in the scripts package?  Actually we should probably change Classifier references to GDD.  Like ImportGDD.java

PR #897, opened 2016-03-04
Revert to old CSS styles

inodb:
The bootstrap upgrade causes a change to the style. Some newlines have disappeared from the top and the bold style has been removed: 

screen shot 2016-03-04 at 11 42 19 am

This reverts things back to the way it was, but that is probably incorrect, so somebody should provide a proper solution instead. @zhx828 could you check?


jjgao:
@inodb @zhx828 : can we resolve this soon? We need to bring travis back.

zhx828:
@jjgao why is this issue related to Travis?

jjgao:
@zhx828 travis detects the changes of the styles / screenshot and failed the build.

zhx828:
Hi @inodb and @jjgao I've changed the bootstrap styles. Most of issues happened on bootstrap global styles for html tags overwrote the default styles. You can see from here: https://rc-style.herokuapp.com/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04 and http://www.cbioportal.org/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04 I think they almost identical, but travis still failed due to phantomjs comparison. Could you guys take a look and give me some suggestions? Thanks.

inodb:
I think it's because the old page is slightly wider than the new one. It doesn't look like a huge issue, so you could update the screenshot with your PR. Github has some image comparison build in that will also make it easier to compare the figures.

PR #901, opened 2016-03-07
Fix the style issues due to Bootstrap upgrates

zhx828:


jjgao:
@zhx828 Instead of directly modifying bootstrap.min.js, is it possible to override outside.

zhx828:
@jjgao Yes we could. Like @inodb mentioned in the email, we should include bootstrap.min.css before including global.css globally. Currently, it will only be included in in some pages. Once we do that, we can override the styles in global.css

zhx828:
@jjgao @inodb All overwrites have been moved to global_portal.css. I also make the bootstrap.min.css globally included. And made sure global_portal.css will be included after all other public libraries have been included. Please see the testing portal https://rc-style.herokuapp.com

inodb:
@zhx828 Looks fine to me. On the patient view the text of recurrence/primary is no longer bold, but it's not a huge issue.

As a side not you might also want to squash the commits to a single one, so the log doesn't become as huge. Now there are multiple update patient view screenshots etc that are not super informative.
```
git rebase -i
```
Then force push to your own repo

zhx828:
@inodb @jjgao Could you please review the code? Thank you.

inodb:
@zhx828 Looks fine to me

PR #909, opened 2016-03-09
Add swagger api specification generation using springfox

sheridancbio:
working on swagger-ui output for web-api through main war file


sheridancbio:
Superseded by #920

PR #925, opened 2016-03-14
Added docker files

als2076:


Dockerfile
```diff
@@ -0,0 +1,41 @@
+# Dockerfile for cBioPortal
+FROM elementolab/jamm:1.0
```
fedde-s: Wouldn’t it be both safer and more efficient to use the official Tomcat image as a base, rather than a manually maintained image based on a different base operating system (CentOS) than the one official images use (Debian)? I think we could get the size of the stack down from 3 GiB to about half of that, and then even share the base OS with the MySQL container. Moreover, this is a requirement if we aim to make it an official image one day: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#from

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+MAINTAINER Alexandros Sigaras <als2076@med.cornell.edu>
+#===========================#
+# Clone cBioPortal			#
+#===========================#
+#RUN git clone https://github.com/cBioPortal/cbioportal.git; \
+ADD / /cbioportal
```
fedde-s: Small note: the `ADD` command is disfavoured for simple cases like this in favour of the `COPY` command (https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy). In addition, I do not believe I’ve seen this use of absolute paths as source operands documented anywhere, does it do the same thing as this?
```Dockerfile
COPY . /cbioportal
```

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN	echo "export PORTAL_HOME=/cbioportal" >> /root/.bash_profile;
```
fedde-s: An `ENV` command would probably be a lot clearer, especially since practically every single command run in a container using this image will need this variable, and they probably will not want to run a bash shell along with them to explicitly parse .bash_profile.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Setup MySQL				#
+#===========================#
+RUN chkconfig mysqld on; \
+	service mysqld start; \
+	mysql_secret=$(cat /var/log/mysqld.log | grep "A temporary password is generated for" | awk '{print $NF}'); \
+	new_pass="P@ssword1"; \
+	cbio_user="cbio"; \
+	cbio_pass="P@ssword1"; \
+	mysql -u root -p$mysql_secret --connect-expired-password -e "set password=password('$new_pass'); flush privileges;"; \
+	mysql -u root -p$new_pass -e "create database cbioportal; create database cgds_test; CREATE USER '$cbio_user'@'localhost' IDENTIFIED BY '$cbio_pass'; GRANT ALL ON cbioportal.* TO '$cbio_user'@'localhost'; GRANT ALL ON cgds_test.* TO '$cbio_user'@'localhost'; flush privileges;" ;
```
fedde-s: Is this step really necessary to build cBioPortal, without running integration tests? Since (I believe) we’re provisioning an a filesystem and process environment to run either 1) a Tomcat server running cBioPortal, 2) a Git pull and Maven build of cBioPortal or 3) the cBioPortal import scripts, I think we can be a lot more efficient by not including MySQL in it at all.

docker/build/gene_sets.txt
```diff
@@ -0,0 +1,22 @@
+Prostate Cancer: AR Signaling=SOX9 RAN TNK2 EP300 PXN NCOA2 AR NRIP1 NCOR1 NCOR2
```
fedde-s: Is this file specific to Docker provisioning? Or are you suggesting it as a new default?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+#===========================#
+# ADD Default Config		#
+#===========================#
+RUN cp /cbioportal/docker/config/* /cbioportal/src/main/resources/; \
+	source /root/.bash_profile; /bin/cp -u --force /cbioportal/docker/config/context.xml $CATALINA_HOME/conf/context.xml;
+#===========================#
+# Build From Source			#
+#===========================#
+RUN cp /cbioportal/docker/build/gene_sets.txt /cbioportal/core/src/main/resources/; \
```
fedde-s: Is the default version of this file that’s included in the repository somehow not applicable to Docker builds? Or do you really mean to suggest changing this default file?

docker/scripts/checkCustomFiles.sh
```diff
@@ -0,0 +1,6 @@
+if [[ -e /custom_files/ ]]; then
+	source /root/.bash_profile; \
+	sleep 5; \
+	/bin/cp --force /custom_files/cbiologo.png $CATALINA_HOME/webapps/cbioportal/images;
```
fedde-s: Would it be possible to customise these settings using the [cBioSkins](https://github.com/cBioPortal/cbioportal/wiki/Customizing-your-instance-of-cBioPortal) configuration options? Then there would be no need to run an additional shell script concurrently with Catalina.

docker/scripts/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+source /root/.bash_profile; \
+/bin/bash /cbio_scripts/addcBioAliases.sh; \
+/bin/bash /cbio_scripts/checkCustomConfig.sh; \
+$CATALINA_HOME/bin/startup.sh; \
```
fedde-s: Is it possible to use an `exec` statement to run Catalina in the foreground, and get rid of the concurrently running shell process? Since Catalina is the process for which the container was started, it should ideally run as process id 1.

docker/customization/StudyViewInitCharts.js
```diff
@@ -0,0 +1,1237 @@
+/*
```
fedde-s: I don’t believe this file is used for anything, is it?

docker/scripts/checkCustomFiles.sh
fedde-s: In addition, I think you meant `/cbioportal/docker/customization/cbiologo.png`, although the default you suggest there is specific branding for Weil-Cornel, which is probably not the intended default to merge to rc.

Dockerfile
fedde-s: Ah, almost forgot: the image should also be provisioned for running the database update script.

inodb: Yeah this has been changed recently. You don't need MySQL anymore to build unless you are running the core tests.

Dockerfile
als2076: Yes. does the same thing. Will switch to:

```Dockerfile
COPY . /cbioportal
```

Dockerfile
als2076: Agree 100% in adopting an official image and any improvements to reduce size.

To the best of my knowledge tomcat is lacking maven and mysql which are both required components during build.

Current [dockerfile] (https://hub.docker.com/r/elementolab/jamm/~/dockerfile/) builds from official centos and installs all cBioPortal prereqs from scratch.

We can switch to tomcat but we would still need to install all prereqs. Any suggestions?

Dockerfile
als2076: Currently this loads from .bashrc.
The `ENV` command is the correct approach for variables configured during runtime.
Would configuring PORTAL_HOME, which is practically our project dir add any value or should we simply define a path in the dockerfile?

Personally leaning towards the latter but open to suggestions.

Dockerfile
als2076: Great. I will go ahead and remove mysql related lines 13-20.
Thanks for the feedback

Dockerfile
als2076: context.xml with mysql configuration is no longer required. We can definitely remove this.
Thanks

docker/scripts/checkCustomFiles.sh
als2076: 100% agree to modify using the cBioSkins. Only thing required is to pass the customization filepath using the -v during runtime and updating the portal.properties file accordingly

docker/scripts/entrypoint.sh
als2076: Agree. Docker has greatly improved over the past and we should use:

```Dockerfile
ENTRYPOINT exec $CATALINA_HOME/bin/catalina.sh run
```

docker/customization/StudyViewInitCharts.js
als2076: No. Part of internal testing for the IPM cBioPortal that we were changing that file.
I will remove it. Thanks

docker/scripts/entrypoint.sh
fedde-s: I would suggest `CMD` instead, to allow for running different commands (such as the data loading script) instead of the web server. `CMD` sets the default command if no other command is given to `docker run`, while `ENTRYPOINT` seems to be intended for sh scripts that configure the container (the part that varies between installations) before they exec the command. For instance, the Dockerfile for MySQL uses it to set up the database and user permissions before it execs `mysqld`: https://github.com/docker-library/mysql/tree/master/5.7

Dockerfile
fedde-s: I’d say both. Several commands in the scripts package, as well as `mvn test`, use the `PORTAL_HOME` variable.

Dockerfile
fedde-s: I have not tested everything, but I believe these are the Debian packages that contain all the dependencies, including runtime dependencies of the database upgrade script and dataset validator:

```Dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends \
                git \
                maven \
                libmysql-java \
                openjdk-8-jdk="$JAVA_DEBIAN_VERSION" \
                python \
                python-jinja2 \
                python-mysqldb \
                python-requests \
        && rm -rf /var/lib/apt/lists/*
# set up Tomcat to use the MySQL Connector/J Java connector
RUN ln -s /usr/share/java/mysql-connector-java.jar "$CATALINA_HOME"/lib/
```

fedde-s: That is, if using `tomcat:8-jre8` as a base.

Dockerfile
als2076: Fixed in dockerfile

Dockerfile
als2076: Fixed in dockerfile

Dockerfile
als2076: Fixed in dockerfile

PR #926, opened 2016-03-14
Timeline use .*OS_STATUS$ and .*OS_MONTHS$ for timepoint of death

inodb:
Before only CAISIS_OS_STATUS/CAISIS_OS_MONTHS was used for timepoint of death,
which are MSKCC internal names. This changes the determination of the timepoint
by allowing any type of prefix before OS_STATUS and OS_MONTHS (including no
prefix as is the case in the public portal).


inodb:
@pieterlukasse, @aderidder This should fix the timeline issue

pieterlukasse:
hi @inodb , thanks for picking this up so quickly. I do have a few remarks: 
- I would like to avoid such flexibility in the portal code. I prefer to agree on a specific set of names  and leave any harmonization to the scope of a transformation step that should be executed *before* actually loading the data in the portal. This should simplify the documentation and the portal code itself (and also the data validation script).

inodb:
Thanks for checking @pieterlukasse. You mean to change the front end timeline code to support only `OS_STATUS` and `OS_MONTHS` and change e.g. `CAISIS_OS_STATUS` to `OS_STATUS` in the transformation step, correct? That sounds reasonable. One challenge is that currently the data model is such that certain clinical attributes are prefixed with their data source (e.g. `CAISIS_` or `DARWIN_`). In our case we still have a couple of studies that use this style for `OS_STATUS` and `OS_MONTHS` in the private portal. I'm not sure if the idea is to stop tracking of the data source and remove all entries of `CAISIS_OS_STATUS` in the database completely (@jjgao @n1zea144 ?). That would be easiest. If we do want to keep track of the data source for these particular clinical attributes it might be good to have a separate discussion of how we might change the data model to better support that. In the meantime I would be okay with using the code from the PR and support the data model as it currently is until we find a proper solution.

pieterlukasse:
@inodb : thanks for the explanation. It makes sense to want to keep track of the source, I like that. 

I would argue that now would be a good time to improve this since we are currently implementing the clinical data validation and @n1zea144 will be adjusting the data loading part in the coming days. My idea would be to add an extra optional attribute to the clinical data telling what the source system is. Something like `SOURCE_SYSTEM` in table `clinical_attribute`. 

We could perhaps follow these steps: 
- keep your PR as it is, since it solves a current bug
- implement simple and strict validation only for OS_STATUS and OS_MONTHS and SOURCE_SYSTEM
- extend data loading and DB model to accept the new SOURCE_SYSTEM attribute
- in a *new* PR simplify the javascript code to only consider  OS_STATUS and OS_MONTHS

inodb:
Sounds like a plan. There are some more challenges to this:

- Make study view handle `SOURCE_SYSTEM`. Currently it's easy to spot where a clinical attribute comes from due to the prefix.
- Similar for tooltips on patient view
- Think a bit about how `SOURCE_SYSTEM` differs between clinical data and genetic data. Maybe there is some generalization possible? We will be setting a `SOURCE_SYSTEM` attribute for clinical data, but not for genetic data. We currently assume all the genetic data is from the same source (the study), but this is probably not the case in the future. Seems to me like these problems are very similar.

inodb:
FYI looks like we might be able to change all prefixes of `*_OS_MONTHS` and `*_OS_STATUS`, e.g. `CAISIS_OS_MONTHS` to `OS_MONTHS`, so it might be better to wait with merging this PR. We can discuss the `SOURCE_SYSTEM` in another issue in the future.

pieterlukasse:
Nice. This will simplify your code. 
:+1:

inodb:
Since the `CAISIS_` prefixes still haven't completely been purged from the database and I doubt we will be able to solve this soon, I think it's better to merge this PR for now and log an issue.

Issue #922, opened 2016-03-13
Cross Cancer Tab, x-labels get truncated


ecerami:
x-labels get truncated...

![screenshot_767](https://cloud.githubusercontent.com/assets/1009066/13731548/4dff86f0-e942-11e5-846c-5bebd0d4a501.png)

aderidder:
Hi @ecerami, 
I've added some code which changes the left padding and the height based on the length of the text. It now looks like this if you have a very long cancer type:
![image](https://cloud.githubusercontent.com/assets/9414840/13746396/98e61de4-e9f3-11e5-95c1-b238e67ffa91.png)

Is this ok?

ecerami:
thanks @aderidder.  looks much better!

pieterlukasse:
It also happens in the regular multi-study cross-cancer histogram (see screenshot). 
Maybe @JiaoJiao123 can apply the same fix there since she is working on #802 ? 

![image](https://cloud.githubusercontent.com/assets/2900303/13807062/84371192-eb5f-11e5-85e1-50fa2455cde7.png)

JiaoJiao123:
@pieterlukasse 
Thank you for letting me know. I will fix it on my pull request

PR #927, opened 2016-03-15
Fix 922

aderidder:
#922. left-padding and height based on the maximum length of the cancer type

### DIFF
--- a/portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary_histogram.js
+++ b/portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary_histogram.js
@@ -26,7 +26,7 @@ function PancancerStudySummaryHistogram()
 	// Some semi-global utilities
     // Here are some options that we will use in this view
     var width = 1100;
-    var height = 550;
+    var height;
     var paddingLeft = 80;
     var paddingRight = 50;
     var paddingTop = 10;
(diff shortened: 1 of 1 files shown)


aderidder:
Hi @JiaoJiao123 
Since you will be doing a similar fix for cross-cancer, are you willing to review this one?
Thanks!

JiaoJiao123:
@aderidder  Yes sure!

portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary_histogram.js
```diff
@@ -67,7 +67,63 @@ function PancancerStudySummaryHistogram()
...
+        var fontSpace = maxLabelLength*fontSize.substr(0, fontSize.length-2)*0.66;
+        calculatePaddingLeft(fontSpace);
+        calculateHeight(fontSpace);
+    }
+
+    /**
+     * Calculate the left padding
+     * @param fontSpace
+     */
+    var calculatePaddingLeft = function(fontSpace){
```
JiaoJiao123: Just to make the code concise, maybe you can just use the following statement for the calculatePaddingLeft() function? You already initialized paddingLeft with 80. there is no need to define minPaddingLeft again.
paddingLeft = Math.max(Math.cos(0.33*Math.PI) * fontSpace, paddingLeft);

JiaoJiao123:
@aderidder The code looks good overall ! I made comment about a minor thing. You can see it in the files changed.

portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary_histogram.js
```diff
@@ -67,7 +67,63 @@ function PancancerStudySummaryHistogram()
...
+        var maxLabelLength = 0;
+        for(var i=0; i<histData.length; i++){
+            var curType = histData[i].typeOfCancer;
+            if(curType.length > maxLabelLength) maxLabelLength = curType.length;
+        }
+        // determine the font-size used
+        var studyWidth = getStudyWidth(histData);
+        var fontSize = getFontSize(studyWidth);
+        // determine the space used by the label and calculate the padding en height
+        var fontSpace = maxLabelLength*fontSize.substr(0, fontSize.length-2)*0.66;
```
JiaoJiao123: I am not sure why you apply substr() to a number. You can use round() instead

aderidder: The function getFontSize adds a "px" suffix. For the calculation I need to get rid of that.

portal/src/main/webapp/js/src/pancancer_study_summary/pancancer_study_summary_histogram.js
aderidder: Yes you're right I will change this. However, I think I do need to store a minimum, as otherwise if I filter the data based on a cancer type, it won't reset the padding to the minimum. I'll get back to you about this.

aderidder: I've fixed it; both paddingLeft and height are now calculated using your formula.

JiaoJiao123:
ok get it. looks good!

PR #928, opened 2016-03-15
GDD service.

angelicaochoa:
- Added GDD files for web api
- Changed classifier references to GDD.
  @n1zea144


core/src/main/java/org/mskcc/cbio/portal/util/ProgressMonitor.java
```diff
@@ -58,7 +58,7 @@
     /**
      * Private ctor for enforcing singleton
      */
-    private ProgressMonitor() {
+    public ProgressMonitor() {
```
zheins: I don't think this should be public. ImportGDD.importData should not need an instance of ProgressMonitor passed to it. If you look in your ImportGDD main method, you are using ProgressMonitor correctly:

`ProgressMonitor.setCurrentMessage("No GDD data found in directory, skipping import: " + dataFile.getCanonicalPath());`

Try doing this throughout the importData method as well.

jjgao:
closing this since we have a new implementation.

Issue #839, opened 2016-02-11
Staging files for TCGA according to new validation standards


pieterlukasse:
The expected staging files format has changed a bit in #798 , so we need the new export of TCGA staging files to comply to that. 

Logging this issue to keep track of the last issues to resolve.

pieterlukasse:
@n1zea144 : I found a new issue regarding RPPA data. According to #730 one should also provide the Z-SCORES file, next to the LOG2-VALUE file. Could you add this to the next test release of the test study you will be generating?

pieterlukasse:
@n1zea144 : for  brca_tcga_data_cna_hg19.seg I get:  Unknown chromosome "23" : shall we map '23' to X and '24' to Y? I couldn't find any mapping in ImportCopyNumberSegmentData.java

pieterlukasse:
@n1zea144 : in the last copy of http://cbio.mskcc.org/~grossb/brca-tcga.zip the PATIENT_ID is now missing in clinical samples file.

pieterlukasse:
@n1zea144 : these are the remaining issues in the last copy of http://cbio.mskcc.org/~grossb/brca-tcga.zip:
- [ ]  the PATIENT_ID is now missing in clinical samples file.
- [ ] for brca_tcga_data_cna_hg19.seg I get: Unknown chromosome "23" : shall we map '23' to X and '24' to Y? I couldn't find any mapping in ImportCopyNumberSegmentData.java
- [ ] I found a new issue regarding RPPA data. According to #730 one should also provide the Z-SCORES file, next to the LOG2-VALUE file. Could you add this to the next test release of the test study you will be generating?
- [ ] in multiple files there seem to be a few wrong entrez ids remaining. E.g. in data_expression_median.txt: line 10136: Gene symbol does not match given Entrez id; found in file: '(MIR16,-284)'. According to the NCBI file, the correct entrez id for MIR16 is 51573, where the symbol is GDE1 and 363E6.2|MIR16 are the synonyms
- [ ] ERROR: data_methylation_hm450.txt (also similar problem in data_linear_CNA.txt): lines [6450, 7537, 12322, (2 more)]: Entrez gene id not known to the cBioPortal instance.; found in file: ['105376839', '105378503', '654780', '(2 more)']. 
[...]

PR #937, opened 2016-03-16
Meta file for clinical data support

zheins:
#839 

Option parsing now using jopt.

Maintains compatibility with old style clinical data files. If metafile passed in, use it and do not expect attribute type header line in the datafile. If one is not passed in, expect old style data clinical file and use attribute type in clinical data file header for each attribute.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java
@@ -57,7 +57,7 @@ public class ImportClinicalData {
 
     private File clinicalDataFile;
     private CancerStudy cancerStudy;
-    private String attributesDatatype;
+    private AttributeTypes attributeType;
 
     public static enum MissingAttributeValues
     {
@@ -93,6 +93,19 @@ public class ImportClinicalData {
         }
     }
     
+    public static enum AttributeTypes 
+    {
+        PATIENT_ATTRIBUTES("PATIENT"),
+        SAMPLE_ATTRIBUTES("SAMPLE"),
+        MIXED_ATTRIBTUES("MIXED");
+        
+        private String attributeType;
+        
+        AttributeTypes(String attributeType) {this.attributeType = attributeType;}
+        
+        public String toString() {return attributeType;}
+    }
+    
     private static void quit(String msg)
     {
         if( null != msg ){
@@ -110,7 +123,7 @@ public class ImportClinicalData {
     {
         this.cancerStudy = cancerStudy;
         this.clinicalDataFile = clinicalDataFile;
-        this.attributesDatatype = attributesDatatype;
+        this.attributeType = AttributeTypes.valueOf(attributesDatatype);
     }
 
     public void importData() throws Exception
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportClinicalData.java
```diff
@@ -170,7 +170,7 @@ public void testImportClinicalDataParameters() throws Exception {
 
 		// TBD: change this to use getResourceAsStream()
         File clinicalFile = new File("target/test-classes/clinical_data.txt");
-        ImportClinicalData importClinicalData = new ImportClinicalData(study, clinicalFile);
+        ImportClinicalData importClinicalData = new ImportClinicalData(study, clinicalFile, "SAMPLE_ATTRIBUTES");
         importClinicalData.importData();
```
pieterlukasse: @zheins would be good to also add a unit test for PATIENT_ATTRIBUTES.

pieterlukasse:
Assigning to @fedde-s to test this with his current version of the validator.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java
```diff
@@ -104,7 +124,7 @@ public void importData() throws Exception
         FileReader reader =  new FileReader(clinicalDataFile);
         BufferedReader buff = new BufferedReader(reader);
         List<ClinicalAttribute> columnAttrs = grabAttrs(buff);
-
+
```
pieterlukasse: I think there should be an `if` here that does this check based on the value of `this.attributesDatatype`. E.g.
```java
int patientIdIndex = -1;
int sampleIdIndex = -1;
if (this.attributesDatatype.equals("SAMPLE_ATTRIBUTES") //use a constant here
  sampleIdIndex = findSampleIdColumn(columnAttrs);
else if (this.attributesDatatype.equals("PATIENT_ATTRIBUTES") 
  patientIdIndex = findPatientIdColumn(columnAttrs);
...
```

fedde-s: @pieterlukasse  Only the sample id is specific to sample-level files; patient ids are required in both. Patient id is actually the *only* column that can be allowed in both since sample attributes and patient attributes are stored in the same table and cannot have the same name.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java
```diff
@@ -104,7 +124,7 @@ public void importData() throws Exception
         FileReader reader =  new FileReader(clinicalDataFile);
         BufferedReader buff = new BufferedReader(reader);
         List<ClinicalAttribute> columnAttrs = grabAttrs(buff);
-
+        
         int patientIdIndex = findPatientIdColumn(columnAttrs);
         int sampleIdIndex = findSampleIdColumn(columnAttrs);
```
pieterlukasse: this check on row 131 needs to be adjusted: PATIENT_ID always required, SAMPLE_ID is only required when `this.attributesDatatype.equals("SAMPLE_ATTRIBUTES")`

n1zea144:
I think a cleaner way to implement this is to use an Enum with values for a patient attribute file, sample attribute file or mixed attribute file instead of a string.  The values that the datatype property (in the meta file) can be assigned are the same you would find in this Enum.  That would make mapping the datatype property to a specific Enum value straightforward.  If the metafile is not provide the Enum would be assigned the mixed datatype.

fedde-s:
Commit thehyve/cbioportal@189c0490d110aec66cd91f80820775ecf8a61432 merges this pull request into the branch in which we are developing the new validation and full-study loading workflow, which means it will be included in our big pull request to the rc branch soon. As this is what the change was requested for, merging it to rc independently now would probably be superfluous work. Closing.

PR #947, opened 2016-03-21
Separate the table creation function and change table content function in Study View Pie js file

zhx828:
This change will fix the issue that the short name not always attached
with a tooltip in pie chart label tooltip.


jjgao:
It is strange that some times the tooltips do not show up for some pie charts when you mouse over.

zhx828:
Hm... The probably still exist? Could you give me some examples?

Issue #951, opened 2016-03-22
Invalid use of reserved word as identifier in SQL


morungos:
In `DaoGeneticAlteration#addGeneticAlterations`, the reserved word `VALUES` is used as an identifier. It needs identifier escaping. This throws an error in MySQL, at least on my system -- and this is documented as being expected so it ought to be amended to use identifier escaping.

fedde-s:
I wondered why nobody else seemed to have any trouble with that unquoted column… Figured that the parsing logic was different in the version of MySQL I was running, from a release repo of a GNU/Linux distro. Those tend to lag behind patches that are not security-related especially if they affect interfaces.

fedde-s:
Ah, @pieterlukasse knew why it wasn’t an issue for more people: when loading data from files, the function contacts the database via `MySQLbulkloader.insertRecord(String...)` and never gets to this SQL statement. So presumably it’s only used in unit tests.

pieterlukasse:
can be closed, right?

PR #952, opened 2016-03-22
Added identifier escaping to VALUES. See #951

morungos:


morungos:
Hmm, this error is nowt to do wi' me. I'm guessing there's something borked in rc

pieterlukasse:
just ran into this now. Thanks for fixing!

Issue #938, opened 2016-03-16
Consolidate search paths for portal.properties (and log4j.properties)


n1zea144:
- [x] 1. Remove references to $PORTAL_HOME in pom.xml files
- [x] 2. Change orders of search path in all spring context to $PORTAL_HOME/... and then classpath:....
- [x] 3. Check GlobalProperties.java, but it should be checking in this order too
- [x] 4. Turn off resource property filtering for all spring config files

Note - need to consider affect this has on db versioning code - we may have to filter one or more java classes with a db version - via maven/resources-plugin.

morungos:
Just to comment that probably the syntax might be: 

```
            <property name="locations">
              <list>
                <value>file://#{systemEnvironment['PORTAL_HOME']}/portal.properties</value>
               ....
              </list>
            </property>
```

As I know the syntax used to use $PORTAL_HOME directly in Spring files as is currently the case (cf https://github.com/cBioPortal/cbioportal/blob/72acaf5e0a8719dd763750f8d2b8418bcb991bc4/web/src/main/resources/applicationContext-web.xml), because then there's a risk they get templated by Maven to a build-time location which defeats the purpose of using environment variables.

n1zea144:
I think the reference to systemEnvironment is not need because of the following:
PropertyPlaceholderConfigurer.searchSystemEnvironment setting to true.   In any event, including in this issue - turn off all "filtering" of Spring configs so that templating happens at runtime.  Note, some of the complexity introduced here was a result of providing user configurable settings before the code was open source and all we could provide was binaries.  Now that we are open source, we have an opportunity to clean things up.

I am afraid some of this cleanup make cause exceptions in some environments.

morungos:
Okay, that's great!! I'm going to have to work to sync some of my weirder deployment issues back into pull requests.

morungos:
Whoo @zheins -- this is looking good :-)

morungos:
Note the comment from the closed ticket. Sometimes Spring configures from `file:///${PORTAL_HOME}/src/main/resources/portal.properties` and sometimes from `file:///${PORTAL_HOME}/portal.properties`. We're also using duplicate `propertyPlaceholderConfigurer` beans in different files. All of these issues remain to be addressed, and this is still a significant obstacle for us.

n1zea144:
This is a mistake.  @zheins can you look at this?

We only want to reference /src/main/resources/portal.properties at build time and via relative path, not $PORTAL_HOME. 

[...]

PR #953, opened 2016-03-22
Issue #938

zheins:
Filter at runtime spring config files

Add filtering of portal.properties for core module

Search order for spring context is $PORTAL_HOME/.. then classpath:...

Removed $PORTAL_HOME from poms

Put spring security intercepts directly into applicationcontext-security to simplify filtering.

### DIFF
--- a/business/pom.xml
+++ b/business/pom.xml
@@ -74,7 +74,7 @@
               <outputDirectory>${project.build.directory}/classes</outputDirectory>
               <resources>
                 <resource>
-                  <directory>${PORTAL_HOME}/src/main/resources</directory>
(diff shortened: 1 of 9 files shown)


n1zea144:
Lets get rid of references to PORTAL_HOME/src/main resources in spring context files.

inodb:
This seems to have broken travis: https://travis-ci.org/cBioPortal/cbioportal/builds/120946584

inodb:
This should fix it: https://github.com/cBioPortal/cbioportal/pull/1016

n1zea144:
Yes, the file no longer has to be copied.

PR #963, opened 2016-03-28
Unignored and fixed failing unit tests. See #606

### LINKED ISSUE #606: Most of the core tests are not working

ersinciftci:
Added JaCoCo maven plugin.
Removed testFailureIgnore from maven surefire plugin.
Fixes #606


ecerami:
@ersinciftci:  thanks.  this looks good to me.  Only question I had:  can we tie this pull request to the original issue:  https://github.com/cBioPortal/cbioportal/issues/606?

@jjgao:  have you already scheduled this for the next release / merge?  I am excited to finally get the unit tests working again!

ecerami:
Perhaps @onursumer or @morungos could review?

jjgao:
@ecerami I scheduled it for 1.2.0. I'll release 1.1.0 this week. We can do a hotfix release of this if necessary.

ersinciftci:
I realized that there was another -DskipTests in .travis.yml, so travis didn't run the tests the first time, I removed it and now travis gets "Access denied for user 'cbio_user'@'localhost'" error, I'm trying to understand the problem.

ersinciftci:
Tests should use the travis user as specified in .travis/settings.xml but it's trying to use cbio_user, something is overriding it.

ersinciftci:
It seems @inodb changed the db.user in .travis.yml to cbio_user to run end-to-end tests with an external database, but travis' database doesn't have cbio_user.

inodb:
Yeah you should run maven twice. Basically the first time it should run the core unit tests using the local mysql db on Travis for testing (maybe you can only build core or smth for this). The second time the war should be built with the external database on amazon for making the screenshot. Another option would be to make the database option modifiable post build.

morungos:
We should maybe talk on the 11th. Personally, I'd love to see core testable with HSQLDB for unit tests. Or let's put it another way: we should probably distinguish unit and integration testing better. Running Maven twice shouldn't be necessary in the long term, even if we do need to do it now. That reflects the problems we're still baking values into the jars and wars from Maven.

ersinciftci:
@morungos I agree that we should distinguish unit and integration testing better, but I think unit tests shouldn't connect to any database, in-memory or not, we should mock the part that connects to a database, and integration tests should use an in-memory database like HSQLDB.
@inodb Ok, I'll do a temporary fix to make it work.

ersinciftci:
Now tests are passing in travis as well.

core/src/main/java/org/mskcc/cbio/portal/dao/JdbcUtil.java
```diff
@@ -244,4 +245,30 @@ static Double readDoubleFromResultSet(ResultSet rs, String column) throws SQLExc
...
         return rs.wasNull() ? null : d;
     }
+
+    /**
+     * Tells the database to ignore foreign key constraints, effective only for current session.
+     * Useful when you want to truncate a table that has foreign key constraints. Note that this
+     * may create orphan records in child tables.
+     * @param con Database connection
+     * @throws SQLException
+     */
```
inodb: hurray doc comments :+1:

inodb:
:+1: Wonderful! Thanks! This looks good to me. Maybe someone else with more experience of the core tests can take another look.

In terms of commit history. Maybe try to rebase to latest version and squash the commits to a single one.

inodb:
Also, I agree with the points raised before. We should def start separating unit/integration test logic and do something about those config values baked into the war.

ersinciftci:
Rebased and squashed the commits.

morungos:
This looks like a fair solution -- for now. I avoided the foreign key hacking just by ordering table handling. It's basically because MySQL has a broken model of referential integrity, that requires consistency at every single query. Other DBs don't, IIRC. Turning foreign key checks off and on is not a good long-term strategy, even just to get tests working. But this is the least of our problems with the core.  

My solution, when I tackled this, was to remove `deleteAllRecords` methods and all the logic for resetting the database, and to use InnoDB to put the tests into transactions which roll back. After all, `deleteAllRecords` is not something we will ever want in production, and even having code there to wipe the database is something I get antsy about.

ersinciftci:
I agree, I didn't think about a long term solution because I think we shouldn't connect to a database(in-memory or not) in unit tests, and we should use a high level data access technology, like Spring Data JPA, instead of low level JDBC code, so the future of these unit tests and JdbcUtil class are questionable.

morungos:
@ersinciftci Yes, I know where you're coming from, although in practice, the repository layer does need to be tested too, so I think there's always a small place for DB testing. I'm not a huge fan of JPA, personally -- I'm liking QueryDSL SQL right now, as it's proving a decent way to write safe DB code while still allowing some interesting queries without JPA. 

Maybe we should speedily hack out the whole DAO layer into a repository soon, so we can deprecate all the statics.

ersinciftci:
I think repository layer should be tested with integration tests. QueryDSL is a must if we need interesting queries(like dynamic queries), I don't know if we do, I need to get more familiar with the project, but if our queries are simple, Spring Data JPA's interface methods that doesn't require implementation are so convenient.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoTypeOfCancer.java
```diff
@@ -131,20 +131,22 @@ public static int getCount() throws DaoException {
...
-      try {
-         con = JdbcUtil.getDbConnection(DaoTypeOfCancer.class);
-         pstmt = con.prepareStatement("TRUNCATE TABLE type_of_cancer");
-         pstmt.executeUpdate();
-      } catch (SQLException e) {
-         throw new DaoException(e);
-      } finally {
-         JdbcUtil.closeAll(DaoTypeOfCancer.class, con, pstmt, rs);
-      }
-   }
```
onursumer: Although some lines are identical, they were still marked as modified. A possible IDE auto formatting behavior related to white spaces.

core/src/test/java/org/mskcc/cbio/portal/dao/TestDaoCancerStudy.java
```diff
@@ -85,18 +85,16 @@ public void testDaoCancerStudy() throws DaoException, IOException {
         assertEquals("GBM", cancerStudy.getName());
         assertEquals("Glioblastoma", cancerStudy.getDescription());
 
-        cancerStudy.setName("Breast");
-        cancerStudy.setCancerStudyStablId("breast");
-        cancerStudy.setDescription("Breast Description");
-        DaoCancerStudy.addCancerStudy(cancerStudy);
+        CancerStudy cancerStudy2 = new CancerStudy("Breast", "Breast Description", "breast", "brca", false);
```
onursumer: nice catch!

core/src/test/java/org/mskcc/cbio/portal/scripts/TestCutInvalidCases.java
```diff
@@ -44,14 +42,12 @@
...
-
-	// TODO Fix this test following the change in data file
-	@Ignore
+    
 	@Test
     public void testCutInvalidCases() throws Exception {
 		// TBD: change this to use getResourceAsStream()
-        File casesExcludedFile = new File("target/test-classes/cases_excluded_test.txt");
-        File dataFile = new File("target/test-classes/cna_test.txt");
+        File casesExcludedFile = new File("src/test/resources/cases_excluded_test.txt");
```
onursumer: I don't remember what was the initial reason to refer the files in target/test-classes/ instead of src/test/resources, but I think this is the right way to refer to test files.

onursumer:
Added a few comments for some minor points. Looks good to me in general.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoTypeOfCancer.java
ersinciftci: The indentation had 3 spaces, so IntelliJ converted it to 4 spaces proper indentation.

jjgao:
@ersinciftci could you please resolve the conflicts?

ersinciftci:
After merging it gives 

> Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.16:test (default-test) on project core: ExecutionException; nested exception is java.util.concurrent.ExecutionException: java.lang.RuntimeException: The forked VM terminated without saying properly goodbye. VM crash or System.exit called ?

error, I commented out all System.exit's in the code to test but still fails. I increased surefire memory with vm args but still. Anyone has any idea?

morungos:
We've nailed down the VM exit issues in https://github.com/cBioPortal/cbioportal/issues/1113, and there should be a PR on that soon. It's basically due to changes in command line arguments for importers coupled with improved validation systems now calling `System.exit()` more, and we needed to factor those into `main` methods which are mostly stubs + exception handlers around a `Runnable`. That means we can test virtually everything without calling a `System.exit()`, and even if a command line parse fails, we don't exit the VM.

ersinciftci:
Trying to retrigger build by closing an reopening to solve "Waiting for status to be reported" issue but it doesn't work anymore (it worked in the morning). Might have to close this and create a new PR if it doesn't get solved.

ersinciftci:
Opened a new PR with the same changes https://github.com/cBioPortal/cbioportal/pull/1272, still has the "Waiting for status to be reported" problem though. Probably github has some connection issues because  it was so slow in the morning.

ersinciftci:
Closed the new one and reopened this since it seems it's a github issue https://status.github.com/messages

ersinciftci:
Ok, finally Travis build succeeded, with real tests! Had to revert @inodb 's docker related changes back because we can't access the Travis mysql from Docker container, I talked to him, we will find a way in the coming days, but now all tests are running. There is 1 ignored test, will unignore it today or tomorrow, but this PR is ready to be merged.

morungos:
Wonderful. Great work :-)

ecerami:
Hooray!!!

Issue #968, opened 2016-03-29
Add model deployment under Jetty


morungos:
At UHN we don't use Tomcat, but Jetty. This issue is about adding a couple of nonintrusive samples files which show how to build and deploy a Jetty version of the portal. Jetty has a few advantages, it's designed for running in simple daemons and doesn't require the management layers of Tomcat.

PR #969, opened 2016-03-29
Add a model Jetty deployment

morungos:
This relates to #968, and provides a Jetty XML file and a web.xml override that can deploy an entire cBioPortal using Jetty from the command line. This is ideal for daemons and small deployments without a larger web container.

### DIFF
--- a/src/main/etc/jetty-cbioportal.xml
+++ b/src/main/etc/jetty-cbioportal.xml
@@ -1,0 +1,198 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure.dtd">
+<Configure id="Server" class="org.eclipse.jetty.server.Server">
+
+    <New id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
+      <Set name="outputBufferSize">32768</Set>
+      <Set name="requestHeaderSize">8192</Set>
+      <Set name="responseHeaderSize">8192</Set> 
+    </New>
+    
+    <!-- Fairly awful configuration of the class list -->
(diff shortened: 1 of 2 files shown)


morungos:
This doesn't touch any existing code, so it should be very safe to merge. It simply creates a few small additional files in the repo, none of which get built into the web application.

n1zea144:
@morungos Should we keep this separate from src/main/resources or consider moving other configuration files here?

morungos:
I kept it separate as things in `src/main/etc` shouldn't get into any jars/wars, and although I think that's probably true for the stuff in the top level, I'd like to be a bit cautious about it. It could be renamed, but I wouldn't put them into `src/main/resources`.

inodb:
I was using heroku's `webapp-runner.jar` for the same purpose. I think jetty should work as well on heroku: https://devcenter.heroku.com/articles/deploy-a-java-web-application-that-launches-with-jetty-runner. Might be better to have only one of these command line runners. It does seem that the `webapp-runner` takes up less LOC: https://github.com/cBioPortal/cbioportal/blob/master/portal/pom.xml#L124-L171, but maybe the jetty runner has more features. Seems like a bunch of IDEs use it.

morungos:
If heroku's runner is production stable I'd be open to switching if it's simpler. But we have a couple of non-trivial requirements:

 * Must be able to contain two war-based webapps (for us, because we side-serve an OIDC provider)
 * Must be able to JNDI (because DB configuration is currently very messy)
 * And, as I said, must be production stable as a Linux daemon (because Tomcat is admin impossible here)

We didn't search far for a match, because I have templates for Jetty on Debian. We'll be deploying cBioPortal into OICR shortly, and they're a Debian place, so I'm hoping to make dpkgs for easy updates at their end.

Issue #884, opened 2016-03-02
bug in import gistic_genes_amp and _del?


aderidder:
Hi,

I was looking at "Select Genes from Recurrent CNAs (Gistic)" in the brca provisional on cbioportal.org 
I looked at the genes_in_region from the staging file data_gistic_genes_amp and compared it to the values given in the popup. 
For some it seems ok. For example, DDAH1 has a qvalue of 4.1251e-13
For others genes are missing. For example, for the qvalue 2.7818e-43 the popup says there are 0 genes. According to the file there is quite a list:
```
SNORA40|ENSG00000253047.1,RN7SL600P,RN7SL473P,C1orf138,LINC00568,CTSS,ECM1,ENSA,MCL1,RPRD2,ADAMTSL4,GOLPH3L,TARS2,HORMAD1,MIR4257,
```

We checked some other values, and it looks like it only add genes until it encounters the pipe symbol at which point it seems to fail?

pieterlukasse:
@zheins , @n1zea144 could you please take a look at this and let us know what is the best way to fix this?

PR #980, opened 2016-03-30
GisticReader bugs, fixes Issue #884

zheins:
Split genes by | and take only front portion, and if no genes found in database we should skip the event instead of adding invalid entry to database.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GisticReader.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GisticReader.java
@@ -59,6 +59,8 @@ public class GisticReader {
      * @return                  CancerStudyId
      * @throws DaoException
      */
+    
+    private static org.apache.log4j.Logger logger = org.apache.log4j.Logger.getLogger(GisticReader.class);
 
     public int getCancerStudyInternalId(String cancerStudy_str)
             throws DaoException {
@@ -156,7 +158,9 @@ public class GisticReader {
             Gistic gistic;
             try {
                 gistic = this.parseLine(line, cancerStudyId, chromosomeField, peakStartField, peakEndField, genesField, qvalField, ampField, cytobandField);
-                gistics.add(gistic);
+                if (gistic != null) {
+                    gistics.add(gistic);
+                }
             } catch(Exception e) {
                 e.printStackTrace();
             }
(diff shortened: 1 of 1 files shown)


zheins:
#884

core/src/main/java/org/mskcc/cbio/portal/util/GisticReader.java
```diff
@@ -200,25 +204,26 @@ private Gistic parseLine(String line, int cancerStudyId, int chromosomeField, in
...
 
-            if (canonicalGene == null) {
-                canonicalGene = new CanonicalGene(gene);
-
-//                    System.out.println("gene not found, skipping: " + gene);
-//                    throw new DaoException("gene not found: " + gene);
-            }
+            if (canonicalGene != null) {
+                if (canonicalGene.isMicroRNA()) {
+                    System.err.println("ignoring miRNA: " + canonicalGene.getHugoGeneSymbolAllCaps());
```
pieterlukasse: Please use
`ProgressMonitor.logWarning()` method. In this way all warnings are grouped together and reported at the end of the import step

core/src/main/java/org/mskcc/cbio/portal/util/GisticReader.java
```diff
@@ -200,25 +204,26 @@ private Gistic parseLine(String line, int cancerStudyId, int chromosomeField, in
...
+                genes.add(canonicalGene);
             }
-
-            genes.add(canonicalGene);
         }
         // -- end parse genes --
-
+        
+        if (genes.size() == 0) {
+            logger.info("No genes found in database - skipping gistic event");
```
pieterlukasse: Please use
`ProgressMonitor.logWarning()` method. In this way all warnings are grouped together and reported at the end of the import step

pieterlukasse:
@zheins : thanks for the fix. I added some minor comments regarding the logging.

core/src/main/java/org/mskcc/cbio/portal/util/GisticReader.java
pieterlukasse: I will do it.

core/src/main/java/org/mskcc/cbio/portal/util/GisticReader.java
pieterlukasse: I will do it.

PR #981, opened 2016-03-30
Made spring configuration locations consistent

zheins:
Issue #938


n1zea144:
We want searching from $PORTAL_HOME to come first (already coming from #953 pull request)

PR #1010, opened 2016-04-04
Gdd ui

angelicaochoa:


jjgao:
it didn't break if connected to the current database without the gdd table.

PR #1015, opened 2016-04-05
Patient View Clinical Attributes Customization

inodb:
Make it easier to customize the styling of clinical attributes at the top of the patient view. Mostly moving all the JS rules for patient and sample styling to CSS, which is much easier to interpret. A different institution could replace the CSS file to show different clinical attributes on the patient view if preferred. The CSS file is `portal/src/main/webapp/css/patient-view/clinical-attributes.css`.

screen shot 2016-06-21 at 6 49 06 pm

Follow up to [RFC20](https://docs.google.com/document/d/1LcSUT1eCsNtoxCLowX4VEyqnQFaihiN0YgqoOfmMDo0/edit#heading=h.i7nj1f7d3b6x)

Includes:
- [x] Patient formatting from JS to CSS
- [x] Sample formatting from JS to CSS
- [x] Study Specific Customization
- [x] Add a data cleaning step (should this be changeable as well?)
- [x] Flexbox does not work on travis ci version of phantomJS, either upgrade or use docker container/selenium for FF & chrome screenshot (https://github.com/travis-ci/travis-ci/issues/3225)
- [x] Add documentation for custom clinical attributes styling on patient view page
- [x] Works on Firefox, Chrome, IE 11

# Suggested reviewers
@adamabeshouse @zhx828 @aderidder @yichaoS @angelicaochoa


JiaoJiao123:
@inodb  Generally it looks great. It’s nice to be able to edit clinical attributes through the css file, and have the data process part in util folder. 
Just one minor thing to point out, in util/clinicalAttributesUtil.js, you have a function derive, and inside this function, you have a new function normalizedCaseType declared. From my understanding, it’s not a very good format since you are not able to reuse the function outside, and also hard to track all functions. In the google javascript style guide, Function Declarations Within Blocks is suggested as no
https://google.github.io/styleguide/javascriptguide.xml#Function_Declarations_Within_Blocks

inodb:
Thanks so much for reviewing @JiaoJiao123 ! That's a good suggestion about not using function declarations within blocks, though I'm not sure if they mean within function scope or within a for/while/if block. I will change it to be like `var normalizedCaseType = function(caseType) {}` as proposed in the google style guide just in case. Regarding reuse of the function: I prefer to not reuse the function outside of `derive()`, because it should be removed in the future. That's why I made it private to `derive()` and added the TODO comment that normalizedCaseType should be a clinical attribute at the database level instead.

I'm not sure about the other methods, what the preferred way is, this:
```
function clinicalAttributesUtil() {
  function clean() {
  }
}
```
or:
```
function clinicalAttributesUtil() {
  var clean = function() {
  }
}
```
I have been using the former, but if there's anything in the style guide on this I would be happy to change it.

portal/src/main/webapp/js/src/patient-view/util/util.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+/**
+ * Get first key found in object. Otherwise return null.
+ * @param {object} object - object with key/value pairs
+ * @param {array} keys - array of keys
+ */
+window.getFirstKeyFound = function(object, keys) {
```
zhx828: I would suggest to add all functions into patientViewUtil instead of registering into global.If you think other pages can benefit from this function, put it under cbio-util.js

portal/src/main/webapp/js/src/patient-view/util/util.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return null;
+}
+
+/**
+ * Add format method to JS string, similar to printf. Only added if it isn't
+ * implemented yet.
+ * http://stackoverflow.com/questions/610406/
+ */ 
+if (!String.prototype.format) {
```
zhx828: Really not recommend to attach it into String since you only have one place to use. Suggest to move to patientViewUtil

portal/src/main/webapp/js/src/patient-view/util/util.js
inodb: Thanks, yeah I think `getFirstKeyFound` might be general enough to move to `cbio-util.js`

portal/src/main/webapp/js/src/patient-view/util/util.js
inodb: Good point

portal/src/main/webapp/js/src/patient-view/util/clinicalAttributesUtil.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+  }
+
+
+  return {
+    cleanAndDerive: cleanAndDerive,
+    getSpans: getSpans,
+    addFirstOrderClass: addFirstOrderClass
+  };
+})();
```
zhx828: Since you already wrap them into a singleton, could you pass in getFirstKeyFound as a parameter?

inodb: Yeah I could also just add it as a method in `clinicalAttibutesUtil`, since that's the only place it is used now.

portal/src/main/webapp/css/patient-view/clinical-attributes.css
```diff
@@ -0,0 +1,131 @@
...
+.clinical-attribute[attr-id~="OS_MONTHS"] {
+    order: 4;
+}
+.clinical-attribute[attr-id~="DFS_STATUS"] {
+    order: 5;
+}
+.clinical-attribute[attr-id~="DFS_MONTHS"] {
+    order: 6;
+}
+#more-patient-info a .clinical_attribute[attr-id~="PRIMARY_SITE"] {
```
angelicaochoa: `PRIMARY_SITE` is often used in clinical data interchangeably to mean (1) the anatomic source of the tumor sample or (2) the anatomic source of the _primary_ tissue sample (as opposed to the anatomic source of the metastatic tissue sample, in which case `PRIMARY_SITE` might be null and `METASTATIC_SITE` should be used instead if that is the case). 

Should this clinical attribute be reserved for `more-sample-info` instead? If more than one sample exists (and are different types), then it might make more sense to throw it into `more-sample-info` instead. Or if we want to keep it in `more-patient-info`, perhaps the attribute id should depend on the sample type if the patient has only one sample?  (i.e., if sample type is primary then select from `PRIMARY_SITE`, `TUMOR_TISSUE_SITE`, etc.).

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
```diff
@@ -0,0 +1,181 @@
...
+      'not available',
+      'pending',
+      'discrepancy',
+      'completed',
+      '',
+      'null',
+      'unknown',
+      'na',
+      'n/a'
+    ];
```
angelicaochoa: also include:
- [unknown]
- [not submitted]
- [not evaluated]

These might also be helpful to include although @zheins has been doing some post-processing with tcga data (where most of these attr values appear):
- [not applicable]
- [not available]
- [undefined]

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
```diff
@@ -0,0 +1,181 @@
...
+      derivedClinicalAttributes.DERIVED_NORMALIZED_CASE_TYPE = normalizedCaseType(caseType);
+
+      // TODO: DERIVED_SAMPLE_LOCATION should probably be a clinical attribute.
+      if (derivedClinicalAttributes.DERIVED_NORMALIZED_CASE_TYPE === 'metastasis') {
+          loc = getFirstKeyFound(clinicalData, ['METASTATIC_SITE', 'TUMOR_SITE']);
+      } else if (derivedClinicalAttributes.DERIVED_NORMALIZED_CASE_TYPE === 'primary') {
+          loc = getFirstKeyFound(clinicalData, ['PRIMARY_SITE', 'TUMOR_SITE']);
+      } else {
+          loc = getFirstKeyFound(clinicalData, ['TUMOR_SITE']);
+      }
```
angelicaochoa: should also check `TUMOR_TISSUE_SITE`

portal/src/main/webapp/css/patient-view/clinical-attributes.css
inodb: Well spotted. On `more-sample-info` we should use `DERIVED_SAMPLE_LOCATION`. We should make that a clinical attribute at the database level instead at some point. The code currently moves all clinical attributes that are the same in all samples to the patientInfo, so if the `PRIMARY_SITE` is the same in all samples it will show up in `more-patient-info`.

inodb: Or actually it only copies `CANCER_TYPE` and `CANCER_TYPE_DETAILED`, so maybe we should remove `PRIMARY_SITE` from the `more-patient-info` all together.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
inodb: Thanks, will do

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
inodb: Great, this was actually not in the original code afaik, but yes let's do that

JiaoJiao123:
@inodb oh if you put normalizedCaseType inside of derive function just for temporary use, then that's fine. And if I see any guide about this later, I will share with you.

inodb:
Incorporated all the requested changes, thanks for reviewing @zhx828 @JiaoJiao123 @angelicaochoa. Let me know if there's anything else

portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
```diff
@@ -125,7 +125,7 @@ boolean showDrugs = GlobalProperties.showDrugsTab();
 boolean showSamplesTable = isPatientView;
 String userName = GlobalProperties.getAuthenticatedUserName();
 
-String darwinAccessURL = CheckDarwinAccess.checkAccess(request);
+// String darwinAccessURL = CheckDarwinAccess.checkAccess(request);
```
fedde-s: Is this comment necessary? Is it an old temporary fix for the issue already solved by PR #1389?

fedde-s:
I’ve just spent days documenting these attributes and making the validation script recognise them, unaware of this pull request. Good to see that someone took up the task of cleaning up this code!

docs/Customizing-your-instance-of-cBioPortal.md
```diff
@@ -257,3 +257,8 @@ skin.documentation.markdown=false
 
 If your documentation contains a relative link, cBioPortal will assume it uses the same format and, when clicked, will try to interpret the page as such. If your documentation contains an absolute link, cBioPortal will open it in a new window when clicked.
 Please be aware that the links may be case-sensitive! E.g. https://github.com/cBioPortal/cbioportal/wiki/News.md is not the same as https://github.com/cBioPortal/cbioportal/wiki/news.md
+
+# Custom styling of the patient view's clinical data
+The [Patient View](http://www.cbioportal.org/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04) shows several [clinical attributes](File-Formats.md#clinical-data) at the top of the page, e.g. `AGE`, `SEX`:
+![test](../test/end-to-end/screenshots/firefox/patient_view_lgg_ucsf_2014_case_id_P04.png)
+The order, styling and visibility of those [attributes](File-Formats.md#clinical-data) at the top can be changed by editing the [patient view's clinical attributes CSS file](../portal/src/main/webapp/css/patient-view/clinical-attributes.css).
```
fedde-s: Might it be cleaner to override them in the custom CSS file specified in the as-yet-undocumented portal property ‘`special_css`’ instead, instead of configuring the portal by patching source code under version control? Or is that not feasible with the way CSS works?

fedde-s: Maybe there should be a property to specify which CSS file to use for this otherwise, with `patient-view/clinical-attributes.css` being the default.

fedde-s:
Also, the sample attributes `GLEASON_SCORE_1` and `GLEASON_SCORE_2` were never displayed because of a bug—they were only displayed if the latter was called `GLEASON_SOCRE_2`. I was going to fix that in my PR, but I’ll revert it.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
```diff
@@ -0,0 +1,203 @@
...
+      'not available',
+      'pending',
+      'discrepancy',
+      'completed',
+      '',
+      'null',
+      'unknown',
+      'na',
+      'n/a',
+      '[unkown]',
```
fedde-s: Typo, please fix

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
```diff
@@ -0,0 +1,203 @@
...
+window.ClinicalAttributesUtil = (function() {
+  /**
+   * Clean clinical attributes. Useful for rounding numbers, or other types of
+   * data cleanup steps. Probably differs per institution.
+   * @param {object} clinicalData - key/value pairs of clinical data
+   */
+  function clean(clinicalData) {
+    // Shallow Copy clinicalData
+    var cleanClinicalData = $.extend({}, clinicalData);
+    var NULL_VALUES = [
```
fedde-s: I think there’s no need to add a third place where these values are listed, and can potentially get out of synch with the documentation (it already introduces one here: `[unkown]`). @pieterlukasse told me that the loading code now translates all of them to the same database representation, and even for previously-loaded databases that do already contain these values it would be cleaner to translate them in the Java layer using the same enum before sending them to the client.

pieterlukasse: I agree we should keep this at the java layer (actually I would prefer it to be even before that, in the T step of the ETL...and we are making progress with this...but not there yet, and apparently there is still a need to "harmonize on the fly" some old data that is already in the DB). See also PR #1397 where new "NULL_VALUES" were added to the java enum and the validator list.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
```diff
@@ -0,0 +1,203 @@
...
+          caseTypeLower = caseType.toLowerCase();
+
+          if (caseTypeLower.indexOf('metasta') >= 0) {
+            caseTypeNormalized = 'metastasis';
+          } else if (caseTypeLower.indexOf('recurr') >= 0) {
+            caseTypeNormalized = 'recurrence';
+          } else if (caseTypeLower.indexOf('progr') >= 0) {
+            caseTypeNormalized = 'progressed';
+          } else if (caseTypeLower.indexOf('prim') >= 0 || 
+                    caseTypeLower.indexOf('prim') >= 0) {
```
fedde-s: This conditional looks like a typo.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
```diff
@@ -0,0 +1,203 @@
...
+      return caseTypeNormalized;
+    }
+
+    var caseTypeNormalized = normalizedCaseType(clinicalData, ['SAMPLE_TYPE', 'TUMOR_TISSUE_SITE', 'TUMOR_TYPE']);
+    if (caseTypeNormalized !== null) {
+      var loc;
+
+      derivedClinicalAttributes.DERIVED_NORMALIZED_CASE_TYPE = caseTypeNormalized;
+
+      // TODO: DERIVED_SAMPLE_LOCATION should probably be a clinical attribute.
```
fedde-s: What do you mean by this comment? Is `TUMOR_SITE` a bad name for this attribute?

portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
fedde-s: I see that it was reverted in PR #1433, good.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
pieterlukasse: @inodb I think we were too late with our comments here, so maybe you missed them. Could you take a look at them?

pieterlukasse: @inodb will the custom css be another file or do we have to overwrite one of the files in source control? We already have a global property for overwriting the portal's main css file to change the look and feel, pointing to a custom css file. See code

```java
    String global_style = GlobalProperties.getProperty("global_css");
    String special_style = GlobalProperties.getProperty("special_css");
```
in `css_include_standard.jsp` and `css_include.jsp`.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
inodb: The `prim` one? It's short for `primary`

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
inodb: @fedde-s I think `TUMOR_SITE` is some attribute directy from the clinical database, so not sure if we should normalize it before importing. Maybe @angelicaochoa can comment. I figured it might be better to add an extra attribute that follows this logic, prolly at the java layer. I'm not sure how that is done now at the java layer, but I think it makes sense to allow for some form of derived attributes.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
fedde-s: But am I misreading it if I think the two sides of the `||` clause are identical?

docs/Customizing-your-instance-of-cBioPortal.md
inodb: @fedde-s Yes I think something like that makes sense if we continue to follow current config practices. I do wonder if it's not easier to allow for some type of config that allows you to simply replace source files, so we don't have to make a separate config value for each of them (use at your own risk).

fedde-s: Just a thought: might it be a sensible compromise to look for the instance-specific file in one set place, and if a file by that name exists, use it instead of the default `portal/src/main/webapp/css/patient-view/clinical-attributes.css`?

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
inodb: Good point

docs/Customizing-your-instance-of-cBioPortal.md
inodb: Yes, I like that approach.

portal/src/main/webapp/js/src/patient-view/util/ClinicalAttributesUtil.js
inodb: @pieterlukasse @fedde-s Sorry I actually didn't see these comments until today.

I made some comments further up regarding the config. If you have an immediate need for changing the styling of the patient view page, we can facilitate something like that through changing the properties file as proposed. Otherwise I feel hesitant to introduce new config until we have better defined the current config procedure. I'd like to see consensus on these first https://github.com/cBioPortal/cbioportal/pull/1517 and https://github.com/cBioPortal/cbioportal/pull/1516 .

inodb: Btw @pieterlukasse @fedde-s I agree that these `NULL_VALUES` should prolly be at the java layer (or preferably T step). When I coded this there were still some of these values in the db. Not sure about now, it might be possible to get rid of it.

PR #1016, opened 2016-04-05
Don't copy url intercepts on travis

inodb:
The spring security intercepts are now in applicationcontext-security, so copying url intercepts on travis is no longer necessary.


inodb:
Travis works now, but heroku is still broken

PR #1018, opened 2016-04-06
a fix for initially invisible 3Dmol canvas

onursumer:
see https://github.com/cBioPortal/mutation-mapper/pull/33 for additional information


onursumer:
Now this PR also includes a temporary workaround for webGL error caused by 3Dmol.js

Issue #848, opened 2016-02-17
Data Loading Script : Standardize exit code and error handling 


priti88:
The cbioportalImporter.py should be able to catch any errors in the java classes that it triggers to import data. Currently these different classes in the `core/org.mskcc.cbio.portal.scripts` package do not have a standard way of working. Some will report an error in stderr others will give a stack trace. Stderr is also not a reliable source as a lot of success messages are also printed there.

So it is important to standardize the way the individual java Import* classes behave and agree on the exit codes returned in case of error or success.

pieterlukasse:
A test version is now available at https://github.com/thehyve/cbioportal/tree/fix_848.
Following classes still need to be reviewed and/or refactored:
- IMPORT_STUDY_CLASS = "org.mskcc.cbio.portal.scripts.ImportCancerStudy"
- REMOVE_STUDY_CLASS = "org.mskcc.cbio.portal.scripts.RemoveCancerStudy"
- IMPORT_CANCER_TYPE_CLASS = "org.mskcc.cbio.portal.scripts.ImportTypesOfCancers"
- IMPORT_CASE_LIST_CLASS = "org.mskcc.cbio.portal.scripts.ImportSampleList"
- ADD_CASE_LIST_CLASS = "org.mskcc.cbio.portal.scripts.AddCaseList"
- VERSION_UTIL_CLASS = "org.mskcc.cbio.portal.util.VersionUtil"

pieterlukasse:
Items above now also refactored. Solution is ready for testing and merging on top of #1020

sambrightman:
Could this be extended to the `metaImport.py` and `validateData.py`? It looks like a small number of `sys.exit(exitcode)` statements could be added to allow more easily checking import status programmatically. For example, one in `validateData.py` when running standalone and two in `metaImport.py` when there are errors or warnings with no override. I can submit a PR if you like.

pieterlukasse:
Hi @sambrightman , I believe this has been done already in `metaImport.py` for the scenario you mentioned. But maybe you can make a PR indeed to show the specific part of the code you mean.

sambrightman:
Sure, #1629 shows what I mean. Submitted against `rc`, which I think is what is expected (even though I'm actually having the problem in Docker, and would like to see it merged and rebuilt with such a change). Currently I get validation errors, shown on the console and in the HTML report, but exit code 0. My import scripts detects a successful import due to the 0 exit code.

PR #1020, opened 2016-04-06
Validator and loading improvements

pieterlukasse:
Large pull request with the *new* (python) validation script and respective adjustments in the (java) data loading classes. Summary of main points in this PR:

- refactoring and further implementation of `validateData.py` to validate a complete study directory containing the staging files
- many unit test files and system test files for `validateData.py`
- implementation of the `metaImport.py` script to validate and load a complete study with a single command
- necessary adjustments to the `cbioportalImporter.py` script
- *moved* all these .py script to a new subfolder: `core/src/main/scripts/importer`
- improvements in the output messages given by the *java* data loading layer, also silencing some of the excessive "progress %" messages
- better and more uniform load result statistics messages given by the *java* data loading layer
- stricter validation of the gene id columns (both their names and their contents) in the *java* data loading layer, aligning it with the new validation script. 
- various fixes in the data loading code, including GISTIC, RPPA
- db changes, added to `migration.sql`, moving this code to release `1.2.0`

Documentation ( :warning: this will also be merged to main wiki once the code is merged to `master`, so please review this as well):

- List of relevant changes you need to make in existing staging files: [Data Loading: What You Need To Change](https://github.com/thehyve/cbioportal/wiki/Data-Loading-:-What-You-Need-To-Change)
- Upgrading to this version, including DB changes: [updating your cBioPortal installation](https://github.com/cBioPortal/cbioportal/wiki/Updating-your-cBioPortal-installation)
- Test the validator: follow the validator documentation available at [Data Loading](https://github.com/thehyve/cbioportal/wiki/Data-Loading) 

Other minor changes made in this PR, to improve usability of the validation and loading process by removing unnecessary warning/error messages, simplifying interfaces, etc:

- removed C12ORF9 from `cbio_cancer_genes.txt` configuration file
- deprecated `ImportDataUtil` (all clinical data should now be added via `ImportClinicalData`)
- removal of "germline_white_list_file"  option in `MutationFilter`, simplifying `ImportExtendedMutationData`


pieterlukasse:
@ecerami : the long awaited PR!

ecerami:
@pieterlukasse:  awesome!  the only outstanding issue then may be #848.

pieterlukasse:
@ecerami : yes, that is correct. I would like to do #848 as an improvement on top of this one. Just afraid the necessary code rearrangements will make the review harder. So if people can review and approve this first and we do #848 right after that, it would be ideal. I do agree that #848 is quite essential for the `metaImport` script. I had an incident myself while trying to load BRCA study during development....the loader had crashed on patient clinical file, but the `metaImport` script did not detect it and continued...and it took me a while to notice this.

ecerami:
@pieterlukasse:  sounds reasonable, as long as can get both done by middle of next week.

core/src/main/resources/db/cgds.sql
```diff
@@ -727,3 +729,10 @@ CREATE TABLE `info` (
 );
 -- THIS MUST BE KEPT IN SYNC WITH db.version PROPERTY IN pom.xml
 INSERT INTO info VALUES ('1.1.0');
+
+--1.2.0 specific changes:
```
jjgao: I would put all the db changes into the corresponding tables directly. We don't have to separate the 1.2.0 specific changes in one block (they are already in migration.sql).

pieterlukasse: Yes, I thought about this as well. My reasoning for doing it like this was that I think it will just make it harder for people to correctly share their DB changes. I.e. it would mean one will have to register his changes in a set of `ALTER` statements (which he probably tested locally, so this is fine) and then think of how to translate them to the other `CREATE table` context syntax...which can be error prone and harder to test (i.e. to really test it one has to run the create statements in a separate DB).  

Let me know if you think this makes sense. I can translated this into the `CREATE table` parts if you still prefer that.

jjgao: I think cgds.sql should be just a snapshot, otherwise, it becomes another migration.sql over time. Let's keep it clean and keep everything in create table statement.

core/src/main/resources/db/migration.sql
```diff
@@ -9,3 +9,9 @@ CREATE TABLE sample_list_list LIKE patient_list_list;
 INSERT sample_list_list SELECT * FROM patient_list_list;
 ALTER TABLE sample_list_list CHANGE PATIENT_ID SAMPLE_ID INT(11);
 UPDATE info SET DB_SCHEMA_VERSION="1.1.0";
+
+##version: 1.2.0
+ALTER TABLE cna_event AUTO_INCREMENT=1;
+UPDATE info SET DB_SCHEMA_VERSION="1.2.0";
+ALTER TABLE mutation add UNIQUE KEY `UQ_MUTATION_EVENT_ID_GENETIC_PROFILE_ID_SAMPLE_ID` (`MUTATION_EVENT_ID`,`GENETIC_PROFILE_ID`,`SAMPLE_ID`) COMMENT 'Constraint to block duplicated mutation entries.';
```
n1zea144: Having a problem running this command.  Is a COMMENT only valid when a UNIQUE INDEX is created (I can get this to run by dropping COMMENT component.

core/src/main/resources/db/migration.sql
```diff
@@ -9,3 +9,9 @@ CREATE TABLE sample_list_list LIKE patient_list_list;
...
+ALTER TABLE sample_profile add UNIQUE KEY `UQ_SAMPLE_ID_GENETIC_PROFILE_ID` (`SAMPLE_ID`,`GENETIC_PROFILE_ID`) COMMENT 'Constraint to allow each sample only once in each profile.';
```
n1zea144: Having a problem running this command.  Is a COMMENT only valid when a UNIQUE INDEX is created (I can get this to run by dropping COMMENT component.

core/src/main/resources/db/migration.sql
pieterlukasse: Hi @n1zea144 : does the problem also occur when you use the `migrate_db.py` script? Or did you try to run it manually? 

Anyway, I removed these in a new commit.

core/src/main/resources/db/cgds.sql
pieterlukasse: Ok, I updated the `cgds.sql`. Actually I had apparently already done that before, so I just needed to remove these lines and update the 1.1.0 to 1.2.0.

jjgao:
@pieterlukasse could you resolve the conflicts?

pieterlukasse:
@jjgao : done, was just `core/pom.xml`.

PR #1040, opened 2016-04-11
Fixed issues in Study View Clinical tab table

zhx828:
- The patient/sample ID link won't be truncated
- Fixed CNA column with only two digits and for number lower than 0.01, show '< 0.01'

### DIFF
--- a/portal/src/main/webapp/js/lib/EnhancedFixedDatatable.js
+++ b/portal/src/main/webapp/js/lib/EnhancedFixedDatatable.js
@@ -144,10 +144,15 @@ var QtipWrapper = React.createClass({displayName: "QtipWrapper",
         }
 
         if (attr === 'CASE_ID') {
-            label = React.createElement("a", {target: "_blank", href: cbio.util.getLinkToSampleView(cancerStudyId, this.props.rawLabel)}, label)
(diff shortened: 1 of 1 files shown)


portal/src/main/webapp/js/lib/EnhancedFixedDatatable.js
```diff
@@ -144,10 +144,19 @@ var QtipWrapper = React.createClass({displayName: "QtipWrapper",
         }
 
         if (attr === 'CASE_ID') {
-            label = React.createElement("a", {target: "_blank", href: cbio.util.getLinkToSampleView(cancerStudyId, label)}, label)
+            label = React.createElement("a", {target: "_blank", href: (window.hasOwnProperty('cbio') && cbio.hasOwnProperty('util'))?cbio.util.getLinkToSampleView(cancerStudyId, this.props.rawLabel):''}, label)
```
jjgao: Is it necessary to check if cbio.util exists?

zhx828: Yes. This is the library copy. In the library itself, it doesn't have the cbio.util

PR #1061, opened 2016-04-14
Display cBioPortal Version number in the footer

stachnim:

### DIFF
--- a/test/end-to-end/test_make_screenshots.sh
+++ b/test/end-to-end/test_make_screenshots.sh
@@ -22,6 +22,9 @@ phantomjs --ignore-ssl-errors=true --web-security=false ${DIR}/make_screenshot.j
                               "${DIR}/screenshots/study_view_lgg_ucsf_2014.png" \
                               5000
 
+# upload the second image
+curl -F "clbin=@${DIR}/screenshots/study_view_lgg_ucsf_2014.png" https://clbin.com							  
+							  
 # make sure screenshot is still the same as the one in the repo, if not upload
 # the image
 git diff --quiet -- ${DIR}/screenshots/study_view_lgg_ucsf_2014.png || \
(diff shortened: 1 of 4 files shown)


stachnim:
There is a "Version 1.2.0-SNAPSHOT" in footer. This is probably a cause why check on travis-ci fail.

jjgao:
@stachnim thanks for the pull request. Please update the screenshots for testing. @inodb could you write some document about how to do that?

stachnim:
The version number will change after every release. Maybe we should think about some mechanism to ignore this change on screenshots? Here is an interesant talk about good method for that: https://www.youtube.com/watch?v=UMnZiTL0tUc

inodb:
Nice @stachnim, adding the version number to the page will be super useful for debugging.

Could you summarize the proposed solution from the video? I was thinking to wrap the version in a span. That way I can parse the page with phantomjs and remove the version span before taking the screenshot.

Will also write some documentation on how to update the screenshot.

stachnim:
@inodb I wrapped the version in a span with id for easier remove form screenshot. I added also uploading a second screenshot (someone probably forget that) and remove version span element from taking a screenshot. I was unable to really test this last improvement (-> Windows) - if it's not work I can delete this two line.

For summarize the video. We can use [ImageMagick](http://www.imagemagick.org/) for [comparing two images](http://www.imagemagick.org/script/compare.php) and generate a diff, e.x.: https://clbin.com/Rkp4HH.png Then analysis what is wrong will be simple.

test/end-to-end/test_make_screenshots.sh
```diff
@@ -22,6 +22,9 @@ phantomjs --ignore-ssl-errors=true --web-security=false ${DIR}/make_screenshot.j
                               "${DIR}/screenshots/study_view_lgg_ucsf_2014.png" \
                               5000
 
+# upload the second image
+curl -F "clbin=@${DIR}/screenshots/study_view_lgg_ucsf_2014.png" https://clbin.com
```
inodb: The second image does not need to be uploaded, because in the statements below you can see that it is actually only uploaded in case the images differ (using `git diff`).

inodb:
@stachnim Great, thanks for fixing that!

Also thanks for summarizing the video. I'm using ImageMagick locally to do the diff atm (http://www.akikoskinen.info/image-diffs-with-git/), but yes it would be great to do this on Travis instead.

stachnim:
I removed unnecessary uploading of screenshot, but making a screenshot doesn't work.
```
TypeError: 'null' is not an object (evaluating 'spanVersion[0]')
```
Should I remove this lines in make_screenshot.js

inodb:
I wasn't really sure how to manipulate the DOM in phantomjs. Had to take a closer look to fix it. Made a new PR here: https://github.com/cBioPortal/cbioportal/pull/1079. Thanks again for your contribution!

PR #1062, opened 2016-04-14
Add monitoring health endpoint

stachnim:


n1zea144:
@stachnim Thank you for your contribution, but for various reasons, we don't want to extend the web service in this manner (at least at this time).

PR #102, opened 2015-07-02
Implementation of LDAP authentication mode.

holtgrewe:
Tested on Active Directory server with a relatively complex directory
structure.  Should work with other AD instances but also with OpenLDAP
with hopefully only configuration and no code change.

PR #117, opened 2015-07-17
Active directory

juleskers:
Add initial Active Directory authentication support to cBio.

This is the basic spring configuration to allow cBio to log-in via Active Directory accounts.
It adds the relevant portal.properties settings, spring Config and makes minor changes to the CancerStudyPermissionEvaluator to enable alternative spring-security authentication-providers.
APP.NAME:permission filtering was made optional (portal.properties setting) so that existing AD-groups need not be prefixed with "CBIOPORTAL:" to make them work.

The first whitespace-fix is really only running the IntelliJ codeformatter on the file.
This fixes several odd indents and the mixed use of tab-indenting and space-indenting. The tabs-to-spaces is sadly enough to confuse the github comparator, which makes this diff look needlessly complex.
Local diffing tools like meld handle this better.

Still ToDo: update the login screen with username and password boxes, currently uses the auto-generated barebones spring-security password entry screen.

PR #1072, opened 2016-04-15
Support for authentication with LDAP

holtgrewe:
This PR is a rewrite of #102 for the current `master` of cBioPortal. Also, it fixes the "ad" authentication by setting the `accessControl` into `SpringUtil`.

We could not get the "ad" authentication method by @juleskers to work with the clinic-wide ActiveDirectory controller. Also, setting up an LDAP-backed OpenID/SAMML solution (as proposed by @morungos) is not possible in our environment.

Thus, supporting LDAP by this patch is the only way to use cBioPortal in our environment.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/authentication/ldap/LDAPUserDetailsContextMapper.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/authentication/ldap/LDAPUserDetailsContextMapper.java
@@ -1,0 +1,199 @@
+package org.mskcc.cbio.portal.authentication.ldap;
+
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.mskcc.cbio.portal.authentication.PortalUserDetails;
+import org.mskcc.cbio.portal.dao.PortalUserDAO;
+import org.mskcc.cbio.portal.model.User;
+import org.mskcc.cbio.portal.model.UserAuthorities;
+import org.springframework.ldap.core.DirContextAdapter;
+import org.springframework.ldap.core.DirContextOperations;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
+import org.springframework.security.ldap.DefaultSpringSecurityContextSource;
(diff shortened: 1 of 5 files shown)


n1zea144:
@holtgrewe We don't merge PR's into the master branch directly.  Can you recreate PR into the RC branch?

holtgrewe:
Sure, recreated as #1076

n1zea144:
@holtgrewe Can we just get the following fix in this PR:
"Also, it fixes the 'ad' authentication by setting the accessControl into SpringUtil"

We can consider the LDAP support in a separate PR?  I'm wondering what it does that is not already covered in our LDAP support provide here:
[https://github.com/juleskers/cbioportal/commit/0248e12f8afc943c0f0cebe936583805d188589e](https://github.com/juleskers/cbioportal/commit/0248e12f8afc943c0f0cebe936583805d188589e)

holtgrewe:
To clarify AD vs LDAP for our AD installation:

- We never got #117 to work for our AD setup (which is maintained by the hospital IT services).
- We got the LDAP to work nicely, however. I don't know what is causing the problems with the Spring AD authentication that are solved/worked around by using LDAP. I tried to give the AD support in #117 the same attention as I did our path through LDAP but somehow the code paths appeared to be even more cryptic than the Spring LDAP.

Regarding the comment of setting `accessControl`:

- Adjusting my original patch for LDAP had to include setting `accessControl` through `setAccessControl`. I'm 99% sure that this is also required for AD to work correctly after the recent changes in the `applicationContext-security.xml` file. I did not test it, just tried to follow boyscout's rule here.

I can take out this part of the patch and maybe @juleskers can comment on it being necessary for getting his AD patch to work with the current `rc` branch?

n1zea144:
@holtgrewe We do want the updates  which set accessControl via the call to setAccessControl - I like how you make the call at the top in the shared spring profile section and removed them from individual protocol profiles (and I think it was missing from the AD section).

If @juleskers can assist in getting AD working in the current rc branch, that would be great.

n1zea144:
@holtgrewe I don't want to wait any longer for @juleskers.  I think we should just go ahead and merge this PR.  Can you first rebase with RC and resolve any merge issues?  I'll then go ahead and merge.

holtgrewe:
@n1zea144 Recreated this PR as #1200.

n1zea144:
Closing this PR as its been merged via [1200](https://github.com/cBioPortal/cbioportal/pull/1200)

morungos:
Good this is now here. I might do a second PR for the OpenID Connect stuff that we're using locally. I don't think we can switch to use LDAP directly, because we have four different LDAP servers for different parts of our user community, and packing that logic into cBio isn't appropriate.

Issue #844, opened 2016-02-12
Not all CNA data gets imported


pieterlukasse:
When running the importer, I get the following messages

```
Reading data from:  /media/hyve/ssd/datasets/brca/tcga/data_CNA.txt
Recaching...
Finished recaching...
--> profile id:  157
--> profile name:  Putative copy-number alterations from GISTIC
--> genetic alteration type:  COPY_NUMBER_ALTERATION
--> total number of lines:  23310
Import tab delimited data for 1080 samples.
23218 records inserted into genetic_alteration
648723 records inserted into sample_cna_event
Error: only 648723 of the 651121 records were inserted.
```
Notice the last line: `Error: only 648723 of the 651121 records were inserted.`  This is resulting from the check in `MySQLbulkLoader.loadDataFromTempFileIntoDBMS` function:

```java
if (nLines!=updateCount && !processingClinicalData()) {
             System.err.println("Error: 
```
Question is: how can we check what is the problem? Why are some rows not added? Has this been troubleshooted before?

pieterlukasse:
@n1zea144 : can you maybe help with this?

zheins:
@pieterlukasse In that cna file there are rows per gene with a different hugo_symbol (a different open reading frame) but the same entrez_gene_id (eg. Hugo:CXorf22 Entrez:286464; Hugo:CHDC2 Entrez:286464, the open reading frame maps to the same entrez gene ID). When these get processed they end up just being a duplicate in the temp loading file.

These rows with different ORFs also do not have the same events per sample - meaning some events are being lost or changed.

@jjgao @n1zea144 How do we want to proceed? Do we need to set up a way to handle genes that map to the same entrez id?

jjgao:
@zheins The data does not make sense. `CXorf22` and `CHDC2` are both aliases of `CFAP47` (entrez: `286464`). But the same gene, there should be only one copy number value per sample. I am wondering if there is any bug in the GISTIC code. Could you download the Firehose GISTIC result and check if the error is there?

pieterlukasse:
@jj, @zheins : any updates? I discussed this with @priti88 today and I think it makes sense to throw an exception in this case. I am in the process of letting the java layer abort with non zero exit status in case of errors (as part of #848) and this would be such an error. It currently occurs when loading the staging files from BRCA, which means that it is a real problem for #839, especially if we start throwing exception and aborting the import process (which again, I think we should in this case).

jjgao:
[...]

PR #1074, opened 2016-04-15
Improvements to data loading code

pieterlukasse:
Main change:
- Improvements to `org.mskcc.cbio.portal.scripts.*` classes and related utility classes to ensure standard behavior, i.e. return exit code `-1` when something went wrong and exit code `-2` when wrong cmd line arguments are used.  This is a solution for https://github.com/cBioPortal/cbioportal/issues/848. Mainly `try/catch` and `system.exit()` statements added.

Other fixes:
- Now setting `cancer_study.STATUS` field in import process. At the start of the import process a study is added with status `UNAVAILABLE` and at the end, if nothing fails (as described above), the status is set to `AVAILABLE`. This is apparently not used yet in the UI, but could now be used to avoid displaying studies that are not yet fully loaded. 
- The `migration.sql` script adds `STATUS` field to `cancer_study` **if it does not yet exist** 
- Improvements to `MySQLbulkLoader` to correctly throw an Exception, as discussed in https://github.com/cBioPortal/cbioportal/issues/844. It also does not add the `REPLACE` statement for clinical patient data, as now all study related data is expected to be removed (including `clinical_patient` and `clinical_sample`) when re-importing a study (i.e. `DaoCancerStudy.deleteCancerStudy` is called).
- Fix to skip duplicated records in data_CNA.txt (was causing constraint violations in `sample_cna_event table`)
- small fix to allow Entrez_gene_id column to be empty if Hugo is provided in `ImportExtendedMutations` (MAF data)


pieterlukasse:
@ecerami , @jjgao , @n1zea144, @priti88  : this is the PR that includes the solution for #848

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -330,7 +327,10 @@ public static void addCancerStudy(CancerStudy cancerStudy, boolean overwrite) th
                 pstmt.setString(8, StringUtils.join(groups, ";"));
             }
             pstmt.setString(9, cancerStudy.getShortName());
-
+          	//status is UNAVAILABLE until other data is loaded for this study. Once all is loaded, the 
+            //data loading process can set this to AVAILABLE:
+            //TODO - use this field in parts of the system that build up the list of studies to display in home page:
+            pstmt.setInt(10, Status.UNAVAILABLE.ordinal());
```
n1zea144: Can we comment this out until we start updating other parts of the system to us it?

pieterlukasse: We could do that. Do you want me to disable the part where we set it to AVAILABLE at the end as well? However, we would like to implement the missing parts for this *soon* as the current import process is not transactional (this means that whenever there is a problem and the process stops - which is the key requirement of #848 for which this PR was made - you end up with a corrupted study in the study list). Ideally we don't display such studies in the UI, to avoid unexpected issues when users start querying half loaded studies. I believe @priti88 and @ecerami would be interested in this feature in the context of their requirements for #848.

fedde-s:
Extended automated tests for the import data validator to make them check that it generates the expected HTML reports.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
n1zea144: Yes please.  I agree on the need to do this, in fact we attempt to do this, but it has to integrate propertly with the primitive cache/hashmaps that are populated in memory and it was here that the code became problematic.  I would prefer to implement a complete solution in the PR request and not spread out across multiple ones.

pieterlukasse:
:tada: :smiley: :tada::tada::tada: :+1:

core/src/main/resources/db/migration.sql
```diff
@@ -15,3 +15,10 @@ ALTER TABLE cna_event AUTO_INCREMENT=1;
 UPDATE info SET DB_SCHEMA_VERSION="1.2.0";
 ALTER TABLE mutation add UNIQUE KEY `UQ_MUTATION_EVENT_ID_GENETIC_PROFILE_ID_SAMPLE_ID` (`MUTATION_EVENT_ID`,`GENETIC_PROFILE_ID`,`SAMPLE_ID`);
 ALTER TABLE sample_profile add UNIQUE KEY `UQ_SAMPLE_ID_GENETIC_PROFILE_ID` (`SAMPLE_ID`,`GENETIC_PROFILE_ID`);
+
+##version: 1.2.1
+SET @s = (SELECT IF(    (SELECT COUNT(*) FROM INFORMATION_SCHEMA.COLUMNS WHERE table_name = 'cancer_study' AND table_schema = DATABASE() AND column_name = 'STATUS') > 0,  "SELECT 1", " ALTER TABLE cancer_study ADD STATUS int(1) DEFAULT NULL"));
```
jjgao: Pieter, what is this code for? Could you add some comments?

Since we haven't release 1.2.0, I think we should squeeze it into 1.2.0? Since the code is already merge. I'll do a quick fix now.

pieterlukasse: Hi JJ, this code is the way I found for adding column `STATUS` to `cancer_study` only if it is not already there. I noticed that this column was missing in our DB, probably because it was introduced at a time where we did not yet have such migration scripts. So, basically the code runs `ALTER TABLE cancer_study ADD STATUS int(1) DEFAULT NULL` IF needed, and `SELECT 1` if not. 

Yes, this could also be set into 1.2.0....but then people who already migrated this will get the error that their DB is ahead of the code. Maybe better to keep as 1.2.1? I noticed I forgot to set 1.2.1 in cgds.sql (I set it to 1.2.0, while in pom.xml I correctly set it to 1.2.1 - should I fix it there?)

fedde-s: @pieterlukasse I think the error was in setting it to 1.2.1 in the first place, as we’re still building release 1.2.0. It seems a bit odd to skip release 1.2.0 altogether to fix it as you seem to propose, isn’t there another solution to have developers/testers revert the number in their databases to 1.2.0?

pieterlukasse: @fedde-s the problem is that when you label an extra sql script as 1.2.0 and someone has already executed the previous "1.2.0" script on his test environment, he will not get the extra change you added. Maybe we should just see this number as something separate from the release as such. It could also just be a sequential number.

fedde-s: @pieterlukasse the automated tests always rebuild the database from scratch using the base schema in `cgds.sql` before running, but indeed, there should be some solution to keep an existing database compatible with the release candidate across pull requests… not sure if there’s a better place to suggest such a solution though.

PR #1077, opened 2016-04-18
Integration of CIVIC database

holtgrewe:
The integration of CIVIC follows the approach taken by integration of the
chang_hotspot list. Static TSV files are added as resources to the WAR
file, with the option to use files from the file system instead through
configuration.

The logo of CIVIC is (c) The Genome Institute, inquiry about using their
logo can be found here:
- https://github.com/genome/civic-client/issues/398

Here is a screenshot of a resulting mutation listing

civic_linkout


holtgrewe:
Hm, I don't know why the build fails.

However, I just saw that I have to remove a number of debug/log print messages. Anyway, please tell me what you think.

obigriffith:
Looks great. Although the actual summary of CIViC content doesn't seem correct. BRAF V600E has 14 predictive evidence statements as well as 8 prognostic and 3 diagnostic. Perhaps something is not being parsed correctly?  Or an old version of the data is being used?

holtgrewe:
@obigriffith Good catch!

It turns out the number of predictive matches was incorporated incorrectly by the Javascript code. The count in the dump from April is 3/15/8 and now displays correctly.

zhx828:
@holtgrewe you may need to update the patient view screenshot to pass the phantomjs check. It's under test/end-to-end/. Just use the link in the travis where the test fails.

jjgao:
@holtgrewe thanks for the pull request! I see a lot of Java backend code to serve the data and the data is committed as a text file. I am wondering if it is possible to ajax data from CIViC directly, so that all the backend components here can be removed and the data can always up to date? 

FYI -- we checked in the MyCancerGenome data in the code because they don't have a public API that we can use to pull data.

obigriffith:
To speak from CIViC perspective we also prefer/recommend using the API directly whenever possible and would be happy to work with you on this. Recently updated docs are here:
https://genome.github.io/civic-api-docs/

If it is important to have complete stable dumps of CIViC for use in cBio this could be still be done via the API. Alternatively you could possibly do gene-by-gene API calls on the fly.

jjgao:
@obigriffith Sounds good. @holtgrewe, would you please try to connect to your API directly in the frontend? There could be performance issues. You may want to use multiple AJAX calls to retrieve the data, e.g. initial call to get whether variants are in CIViC, and when hovered, call to get the content in the tooltip.

It is not very important to have a complete dumps of CIViC for use in cBioPortal.

holtgrewe:
Thanks for your comments. I agree that an integration using real-time information from the CIViC API would be ideal and I and the group that I work for (Core Unit Bioinformatics at Berlin Institute of Health) are willing to contribute such an integration in the future.

However, at the moment, we feel that an integration using a CIViC dump gives better trade-offs, to summarise:

- it works analogously to the hotspot integration was thus relatively easy to do
- works relatively fast and stable as no network communication takes place
- can be tested and guaranteed to work with certain dump versions
- it is already in this PR and available.

One of our collaborators on the clinical side is CIViC contributor Damian Rieke. Having the CIViC data readily available in cBioPortal for discussion in their molecular tumor board would be a nice kickback for him and the other clinicians. I imagine that this would increase the positive feedback loop and more clinician time would be invested into CIViC if they get this data integrated into cBioPortal.

In an ideal world, this would use the CIViC API from AJAX, at the moment, I see the following risks (and involved time requirements) with this path forward:

- I don't have a full understanding of the cBioPortal client/server architecture. It is my impression that the information for all genes is sent in one go (and not paged) to the client data tables. Is this correct? There are cases of colon carcinoma with a large number of mutations and the pages load quite slowly with downloading, parsing, and displaying the JSON. I imagine adding hundreds of asynchronous AJAX calls on page load would frustrate many of our clinicians, or even make their browsers hang/crash (they are on older, office-grade hardware). I agree that my solution is pretty pragmatic, but it appears to work fine and follows the implementation pattern of the preexisting hotspot lists.
- I found the front-end of cBioPortal to be a bit hard to follow and would need relatively much support and consulting in adding annotation through AJAX. @jjgao is there a prior implementation of this in cBioPortal? Could you help me with this if we decide to go down this road for CIViC integration?
- @obigriffith if I understand https://github.com/genome/civic-client/issues/396 correctly then an integration with the API depends on changes in the CIViC client. How stable is your installation and the URL? How stable is the API, I am a bit afraid of changes in APIs here. I've burnt my hands before in other projects and the CIViC dump update that resolved https://github.com/genome/civic-client/issues/395 also changed the column names of the dumps.
- In a world of unstable network connections, I think it would be necessary to get the data from the cBioPortal server to not rely on external services. This is particularly true for software tests as (a) tests should run reliable and not depend on network connections and (b) you want to control what is the test database so an update in the contents of CIViC does not make your tests fail.

Would it be possible to include CIViC via dumps now and later through the API once it is stable and the other risks have been addressed as well?

pieterlukasse:
Hi @holtgrewe , I have a question regarding the variants notation. You seem to store some of the variants in a format that is different from what cBioPortal is using. E.g. 

```
CIVICDB         - ERBB2     DEL 755-759
CBIOPORTAL - ERBB2     L755_T759del
```
Have you written any code in this PR that translates between both formats?

holtgrewe:
Hi @pieterlukasse. No, and here we have an interesting problem ;) Let me start a rant.

Generally, the HGVS (Human Genome Variation Society) has developed a [system ("varnomen")](http://varnomen.hgvs.org/) for describing genome variation on the chromosome, transcript, RNA, and protein level. The most important description the the SNV, e.g., chr1:12345G>T and there is good support for describing deletions, insertions, replacement, e.g. chr1:g.12345_12348delGTAC, chr1:g.12345_12346insT chr1:g.12345_12347delGTACinsTTTTT, or on transcripts NM00001.1:c.123C>T. On the protein level this looks like BRAF:p.V600E, BRAF:L755_T759del, etc.

The HGVS standard is widely spread and the GA4GH variant annotation working group is using it in their [proposed standardised ANN field](http://snpeff.sourceforge.net/VCFannotationformat_v1.0.pdf). I assume that this will also be included in the GA4GH APIs. As a shameless plug, the [Jannovar library](https://github.com/charite/jannovar) that I co-develop has code for converting genomic variants to transcript and protein variants and representing them with a (hopefully) nice Java API and also has parsing support for going in the reverse direction.

Coming back to your point. For point mutations, I would expect Civic to contain HGVS-compatible amino acid exchange notation, e.g. V600E on the BRAF protein. However, some of the literature predates the HGVS standard and even the complete sequence of the human genome or maybe even the full sequence of the gene at hand. Thus, the authors from the literature might use non-HGVS variants for describing variants, e.g. "c.234CT>-" or "DUP of exon 3" or "IVS1 (G =>T)".

In the TSV file in this PR, I used the raw Civic dump which contains this maybe inconsistent variants. These cannot be used correctly for annotating variants but that was OK for this purpose. In a first step, linking to the SNVs in Civic would be nice.

In a later step, one would have to work together with the Civic people to resolve the issues for linking and either adjust Civic to force people to enter standards-compliant stuff there or write glueing code that imports "what is there" into cBioPortal.

But that's something for another day, I think, and a serious undertaking that requires a consistent definition of variant descriptions in both cBioPortal and Civic IMO.

By the way, is there any documentation anywhere on how cBioPortal encodes the non-trivial stuff? In case there is no good solution yet, one could roll something that enforces HGVS-compliant variants, (e.g., using Jannovar :wink:).

pieterlukasse:
This one can probably be closed since it was replaced by #2067 . @holtgrewe you agree?

holtgrewe:
@pieterlukasse yes, indeed. Thanks!

PR #1079, opened 2016-04-18
Display cBioPortal version number in the footer

inodb:
Show version number in the footer inside span. During testing, do not display
version number to prevent having to update the screenshot with every version
change.

## Screenshot
screen shot 2016-04-18 at 6 52 57 pm

### DIFF
--- a/portal/pom.xml
+++ b/portal/pom.xml
@@ -77,6 +77,12 @@
               <filtering>true</filtering>
             </resource>
             <resource>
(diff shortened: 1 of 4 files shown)


inodb:
This is a good first step to showing the version of the portal on the front page. We might eventually want to show actual git commits by using this for example: https://github.com/ktoso/maven-git-commit-id-plugin. It would also be nicer if we could get the info from the REST API.

PR #1080, opened 2016-04-19
Bootstrap styling for data validator HTML output

fedde-s:
This gives the HTML reports of the import data validator a more consistent and accessible styling, by phasing out the mix of Bootstrap and jQuery-Mobile and focussing on Bootstrap.


fedde-s:
screen shot 2016-04-19 at 10 15 13

fedde-s:
screen shot 2016-04-19 at 11 14 39

pieterlukasse:
As just agreed with @jjgao and @n1zea144 : @priti88 will test this. If all is OK, I will merge it since this only concerns a HTML template that is part of the new validator.

priti88:
@pieterlukasse I tested the new HTML template. It looks good.
Thank you

PR #1081, opened 2016-04-19
Fix #855 Show CNA if no mutation data is present

### LINKED ISSUE #855: patient view: bug if no mutation data is present
I have a study which has no mutation data. I went to the patient view and it gives an
Uncaught ReferenceError: mutTableIndices is not defined
The problematic line is "aTargets": [mutTableIndices["annotation"]], in cna.json

What seems to happen is that patient_view.jsp checks whether there is a mutation profile, and if there is not, it does not include mutations.jsp
mutation.jsp has the mutTableIndices definition.

inodb:
Fix #855. Move `var mutTableIndices` to `patient_view.jsp`, such that both `cna.jsp` and
`mutations.jsp` can access it. Previously `mutations.jsp` defined it, which is
a problem when there is not mutation profile, because in that case `cna.jsp`
can't access it.

## Notify reviewers
@aderidder, @pieterlukasse, @zhx828, @jjgao

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/patient_view/cna.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/patient_view/cna.jsp
@@ -34,7 +34,6 @@
 <%@ page import="org.mskcc.cbio.portal.servlet.CnaJSON" %>
 
 <script type="text/javascript">
-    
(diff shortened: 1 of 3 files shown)


zhx828:
@inodb the code looks good to me. I would also suggest to create patient view utils to include all similar initialization parameters/data

inodb:
Thanks @zhx828. Let's put the patient view utils on the to-do list: #1082

aderidder:
Hi @inodb the code looks good to me as well.

pieterlukasse:
tested on our CNA only study and it works. :+1:

PR #1101, opened 2016-04-22
Add screenshot test documentation for devs

inodb:
Currently the screenshot test documentation is lacking.

This PR does the following:

- More verbose output of the screenshot test
- Add documentation to `CONTRIBUTING.md` about how to fix failing screenshot test

## Reviewers
@jjgao @pieterlukasse @fedde-s 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ script:
     - cp $PORTAL_HOME/src/main/resources/log4j.properties.EXAMPLE $PORTAL_HOME/src/main/resources/log4j.properties
     - ~/maven/$MAVEN_VERSION/bin/mvn -e -DPORTAL_HOME=$PORTAL_HOME -Ppublic -DskipTests -Ddb.user=cbio_user -Ddb.password=cbio_pass -Ddb.portal_db_name=public_test -Ddb.connection_string=jdbc:mysql://cbioportal-public.c1xhhbwn8izk.us-east-1.rds.amazonaws.com:3306/ -Ddb.host=cbioportal-public.c1xhhbwn8izk.us-east-1.rds.amazonaws.com clean install
     - java -Ddbconnector=dbcp -jar portal/target/dependency/webapp-runner.jar --expand-war portal/target/cbioportal.war &
-    - sleep 10s
-    - bash -x test/end-to-end/test_make_screenshots.sh
+    - sleep 20s
+    - bash test/end-to-end/test_make_screenshots.sh
 
 notifications:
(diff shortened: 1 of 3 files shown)


test/end-to-end/test_make_screenshots.sh
```diff
@@ -1,25 +1,64 @@
...
-        (echo "screenshot differs see:" && curl -F "clbin=@$1" https://clbin.com && exit 1)
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+compare_image() {
+    local url=${1}
+    local png=${2}
+    local TIMEOUT=5000
```
pieterlukasse: maybe double this timeout?

.travis.yml
```diff
@@ -32,7 +32,7 @@ script:
     - ~/maven/$MAVEN_VERSION/bin/mvn -e -DPORTAL_HOME=$PORTAL_HOME -Ppublic -DskipTests -Ddb.user=cbio_user -Ddb.password=cbio_pass -Ddb.portal_db_name=public_test -Ddb.connection_string=jdbc:mysql://cbioportal-public.c1xhhbwn8izk.us-east-1.rds.amazonaws.com:3306/ -Ddb.host=cbioportal-public.c1xhhbwn8izk.us-east-1.rds.amazonaws.com clean install
     - java -Ddbconnector=dbcp -jar portal/target/dependency/webapp-runner.jar --expand-war portal/target/cbioportal.war &
     - sleep 10s
```
pieterlukasse: should we increase this timeout to avoid rendering issues that are maybe related to high load on the server? Like the one @fedde-s reported last week?

inodb: So was it indeed related to high load? Or was it because of an odd merge issue?

fedde-s: Well, I changed nothing in the git branch yet the Travis build has now somehow succeeded, with no sign of a failing build in the build history.

pieterlukasse: I would bet on the high load. It seems to me it would do no harm to increase this. If it results in less issues with travis, that would be great. If the issues remain, at least we tried and learned that this is perhaps not the problem...

inodb: Fair enough, just checking. Yeah it's a 2*5 + 10 seconds difference per test run so it's not that major, adjusted the timeout and sleep values.

inodb:
Doubled sleep and timeout values to make things a bit more lenient in case of high load @ Travis. Will merge if no further comments, @fedde-s @pieterlukasse

PR #1115, opened 2016-04-26
Added new model classes to use in web, service and persistence.

ersinciftci:
Generated model classes from tables using http://www.mybatis.org/generator/, and
converted foreign key fields to actual classes to create many-to-one relationship for every model.
We can use "association" tag in ResultMaps for many-to-one relationships:
http://www.mybatis.org/mybatis-3/sqlmap-xml.html#Result_Maps


n1zea144:
This generated classes, like User and Authorities that we may never use in the new stack (or not for some time).  Should we remove them?

Issue #1117, opened 2016-04-26
Import no longer works with vcf2maf output


morungos:
This is a problem I just observed attempting to load our GENIE data. the current `vcf2maf` makes the `Entrez_Gene_Id` column always zero. However, the newly validating import logic (cf `ImportExtendedMutationData`) reads as follows.

Essentially, because 0 passes the validity test, a correct Entrez gene identifier is now required. This is a breaking and significant change as the old logic would allow the symbol as a fallback. That no longer happens. 

```
                if (record.getGivenEntrezGeneId().trim().length() > 0) {
                    if (!record.getGivenEntrezGeneId().matches("[0-9]+")) {
                        ProgressMonitor.logWarning("Ignoring line with invalid Entrez_Id " + record.getGivenEntrezGeneId());
                        entriesSkipped++;
                        continue;
                    }   
                    if (entrezGeneId != TabDelimitedFileUtil.NA_LONG) {
                        gene = daoGene.getGene(entrezGeneId);
                        if (gene == null) {
                            //skip
                            ProgressMonitor.logWarning("Entrez_Id " + entrezGeneId + " not found. Record will be skipped for this gene.");
                            entriesSkipped++;
                            continue;
                        }                       
                    }
                }
```

morungos:
There's more to this, too, as I read the logic here, there's a bunch of dead code. Since negative numbers can't be passed in the `TabDelimitedFileUtil.NA_LONG` condition can never be false. And because it can never be false, the following `gene == null` conditions can never be true. The *only* way this logic works is with an Entrez gene id in place.

morungos:
An appropriate fix seems to be to amend

```
                    if (entrezGeneId != TabDelimitedFileUtil.NA_LONG) {
```

to

```
                    if (entrezGeneId != 0) {
```

Since `0` makes sense as an unknown marker, where no negative number is possible due to the preceding patch. `0` is also written by `vcf2maf`. I'll make a PR on this.

PR #1118, opened 2016-04-26
Fix issues with zero for Entrez gene id in MAF import

morungos:
This resolves the issues in #1117, where a zero gene identifier in a MAF import file was breaking the import process, where previously the symbol would be used as a fallback. There's more dead code in this area, but this resolves the core problem.


zheins:
I think there are now 3 pull requests including this addressing the issue. PR #1142 and PR #1140

morungos:
Oh noes. Well, now you can pick the best :-)

pieterlukasse:
#1142 is preferred in this case, but thanks for the proposal anyway :+1:

PR #1124, opened 2016-04-27
Making AD authentication work with rc branch

holtgrewe:
As asked for in #1072, the isolated changes that make authentication work with AD.

However, *authorization* is not working with this patch as - for some reason - the authorities come from the AD server. Further, the email address is not part of the `UserPasswordAuthenticationToken` that is passed around internally. The object of this type contains the user name and also the full LDAP distinguished name but no email address that could be used to access the email table.

Further, I could not find any mention of a user mapper in #117 that @ juleskers mentioned in the discussion in #102.

Thus, so far the only working solution that makes authentication with AD work for me and also offers the possibility to use the cBioPortal authorization via the users and authorities table goes through LDAP.


n1zea144:
Thanks!

Issue #1121, opened 2016-04-27
Submit button is no longer appending "&Action=Submit" in IE10 or IE11


jjgao:
@JiaoJiao123 this is urgent - please prioritize. Thanks.

PR #1127, opened 2016-04-27
Fix Issue #1121 Submit Button on IE

JiaoJiao123:


JiaoJiao123:
@jjgao  I tested it in the VirtualBox IE. The submit action was able to appended and it looks fine for me

Issue #1119, opened 2016-04-26
Update Cancer Hotspots annotation icon mouseover text


schultzn:
Add the cancerhotspots.org link, correct citation, change data from 2015 to 2016.

Change to:

Recurrent Hotspot
This mutated amino acid was identified as a recurrent hotspot (statistical significance, q-value < 0.01) in a set of 11,119 tumor samples of various cancer types (based on Chang et al., Nat Biotechnol. 2016).
Explore all mutations at http://cancerhotspots.org/.

PR #1131, opened 2016-04-27
Change Hotspot Description; OncoKB updates

zhx828:


zhx828:
This's related to #1119

zhx828:
@onursumer  could you do a quick review and merge it?

portal/src/main/webapp/js/src/cbio-util.js
```diff
@@ -478,6 +478,16 @@ cbio.util = (function() {
         return target;
     }
     
+    //Get hotspot description. TODO: add type as parameter for different source of hotspot sources.
```
onursumer: instead of defining a utility function with a parameter, we can also define a template

onursumer:
added a minor comment, other than that looks good to me.

Issue #1109, opened 2016-04-26
sorting by Annotation is not correct in the public portal without oncokb anntoation


jjgao:
![image](https://cloud.githubusercontent.com/assets/840895/14804048/860f9e60-0b2d-11e6-825d-599767a9d36e.png)

PR #1132, opened 2016-04-27
Fix annotation column sorting issue

zhx828:
This is related to #1109


zhx828:
@jjgao could you review? Thank you.

PR #1137, opened 2016-04-28
Added new persistence and service modules

ersinciftci:
This module is using the new model module. Converted ProfileDataMapper.getMutationData to the new
modules as an example (didn't delete the original one in business module).
Didn't add the new model package to typeAliasesPackage because there are classes that has the same name in
org.mskcc.cbio.portal.model package, so we need to use explicit model class names in mapper xmls.
Wanted to turn mapUnderscoreToCamelCase setting on but it breaks the current code. We can make those changes
after we delete old models/modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,10 @@ core/target/*
 annotator/target/*
 importer/target/*
 business/target/*
+model/target/*
+persistence/persistence-api/target/*
+persistence/persistence-mybatis/target/*
+service/target/*
 web/target/*
 mutation-assessor/target/*
 cbioportal-client/target/*
--- a/business/src/main/resources/applicationContext-business.xml
+++ b/business/src/main/resources/applicationContext-business.xml
@@ -25,7 +25,7 @@
 
         <!-- scan for mappers and let them be autowired -->
         <bean class="org.mybatis.spring.mapper.MapperScannerConfigurer">
-            <property name="basePackage" value="org.mskcc.cbio.portal.persistence" />
+            <property name="basePackage" value="org.mskcc.cbio.portal.persistence,org.cbioportal.persistence.mybatis" />
         </bean>
 
         <!-- these are required to get properly autowired mappers into our service classes -->
@@ -45,8 +45,10 @@
             <property name="entityAttributeMapper" ref="entityAttributeMapper" />
         </bean>
 
-        <!-- enable component scanning (beware that this does not enable mapper scanning!) -->    
+        <!-- enable component scanning (beware that this does not enable mapper scanning!) -->
         <context:component-scan base-package="org.mskcc.cbio.portal.service" />
+         <context:component-scan base-package="org.cbioportal.persistence" />
+         <context:component-scan base-package="org.cbioportal.service" />
 
         <!-- enable autowire -->
         <context:annotation-config />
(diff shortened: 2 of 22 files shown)


sheridancbio:
Overall this looks good. I have been trying to hook this stack onto a front end to deploy and do system testing, but to enable PR acceptance, I am adding comments now from code inspection.

"Model" maven module
My memory is that we are currently using the same Model classes for both Data Transfer in/out of DBMS, and also to the service layer. In time we may develop a new model set for the upper (service) interfaces of the persistence layer.
This module has a pom artifact version number (1.1.0) and a <modelVersion>4.0.0</modelVersion> pom setting, which can be a little confusing for .pom novices (such as myself). The "modelVersion" is maven boilerplate .. I suppose specifying the "version" of .pom file this conforms to. Since this is a model module, a comment clarifying this might be a good idea.

Style format is a minor issue .. but the class data members / class variables have two line spacing, which looks unusual.

Model classes have camelCase, which is good. All classes match underlying database fields and have appropriate get/set methods.

cancer_study has field importDate with type Date .. but database field shows datetime type. Are we losing precision here? From mysql docs: The DATETIME type is used for values that contain both date and time parts. MySQL retrieves and displays DATETIME values in 'YYYY-MM-DD HH:MM:SS' format. The supported range is '1000-01-01 00:00:00' to '9999-12-31 23:59:59'

In class GeneticProfile, database field cancer_study_id is represented by name cancerStudy .. and cancerStudy is another model object representing the record linked through this foreign key. This was confusing to me at first, but now I get it and it makes sense.

Also, I hope the framework we are using does something smart regarding the relative lengths of string fields. In GeneticProfile, we have a bunch of varchar(255) fields, but also a mediumtext field. So some fields can hold short strings, and some can hold megabytes of text. Will the framework attempt to validate string lengths .. and can we make decisions between throwing exceptions at time of insert versus auto-truncation, etc. Not sure we have thought about this yet.

In class Mutation, we also drop the _id component of the DB field name for mutation_event_id and genetic_profile_id and sample_id. Field entrez_gene_id becomes simply "gene", and all of these are embedded model objects representing the linked associated records for these foreign keys in external tables. One thought this raises in my mind is: does this mean that when we interact with the database tables, we always are required to join the associate tables in order to get the external related data before returning the basic table data? Hopefully the framework does lazy joins so that simple retrieval/lookup within a table (which does not need to follow a foreign key for related info) does not result in expensive / unnecessary joins at the DB level.

In MutationCount, _id is dropped from fields genetic_profile_id and sample_id. Here is where I comprehended the composition of related records into a common object model. When this involves X-to-1 relationship, it makes sense. [otherwise we might have to model the data with class data members of a collection type]

There are fields in MutationEvent which contain names which are acronyms. There is ambiguity on how to properly convert these to camelCase .. such as ONCOTATOR_DBSNP_RS. DbSnp ? DBSNP? DBSnp? I wonder what the standard is for acronyms, and we should maybe make a list of which acronyms we use if it matters. Another concern with camelCase is the tendency for websites to play creative / silly case games .. like iPhone, but even our own cBioPortal or miRBase. Maybe stuff like this should be "squashed" using the same approach as acronyms. Just a thought.

But this PR is consistent and sensible to provide a separate model package. The maven build completes successfully. I intend to put this into use with some of the existing data paths (WebAPI first, and MutationsJSON later)

Issue #1013, opened 2016-04-05
Possible wrong mapping when using negative Entrez id values in staging files


pieterlukasse:
In some BRCA staging files we find **negative** entrez ids, like in `data_CNA.txt`. The problem with this is that the `ImportTabDelimData.java` process accepts negative entrez ids and looks them up using `daoGene.getGene(Long.parseLong(entrez));` . This would be fine if it were not for the fact that each cBioPortal installation will assign **different** negative entrez ids to different symbols, see `DaoGene.addGeneWithoutEntrezGeneId` which just uses a sequential negative count.

This means that the entry in `data_CNA.txt` with Hugo=MIR4433 and Entrez=-1035 could end up in a completely different "gene"  than MIR4433...

pieterlukasse:
As discussed with @jjgao, the solution will be:
- [ ] in validator, block any negative entrez id values
- [ ] in loader, we should also ignore negative entrez gene IDs and use Hugo symbols (as a temporary backwards compatible solution for now, to support old staging files with negative entrez id values).

fedde-s:
Since thehyve/cbioportal@5b8b04b9f70bdfbe2d5e1f5ac772f70ab690f763, the validator considers Entrez gene IDs that are not positive integers an error, in the sense that it implies the file cannot be loaded.

fedde-s:
On Pieter’s request, thehyve/cbioportal@be7a038858fa5084a0592d926c02f815a2cbd5d5 makes the validator allow non-numeric Entrez identifiers with a warning that says the gene will be skipped. Only negative Entrez IDs will be reported as unloadable.

pieterlukasse:
fixed in #1020 for mutation data (https://github.com/cBioPortal/cbioportal/pull/1020/commits/5052fb61beeb496ae57636300b3f6a551b793240)
and in https://github.com/cBioPortal/cbioportal/commit/9152c07 for CNA, mRNA, Methylation

pieterlukasse:
:warning:  still missing for FUSION data (in `ImportFusionData.java`)

PR #1142, opened 2016-04-29
Add support for Entrez 0 and symbol 'Unknown' in MAF files

fedde-s:
Interpret the combination as indicating an intergenic mutation, and otherwise parse them as missing values. Handle these missing values the same as in other file types with these columns—fall back to the symbol if the Entrez ID is missing.


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -251,38 +251,69 @@ public void importData() throws IOException, DaoException {
...
+                String geneSymbol = record.getHugoGeneSymbol();
+                String entrezIdString = record.getGivenEntrezGeneId();
+
+                CanonicalGene gene = null;
+                if (!(entrezIdString.isEmpty() ||
+                      entrezIdString.equals("0"))) {
+                    Long entrezGeneId;
+                    try {
+                        entrezGeneId = Long.parseLong(entrezIdString);
+                    } catch (NumberFormatException e) {
```
pieterlukasse: this is not correct. We agreed to block negative values as well (see #1013)

aderidder:
Test data used for testing:
[1142Test.tar.gz](https://github.com/cBioPortal/cbioportal/files/259028/1142Test.tar.gz)

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -251,38 +251,69 @@ public void importData() throws IOException, DaoException {
...
+                // what the column Variant_Classification says
+                if (geneSymbol.equals("Unknown") &&
+                        entrezIdString.equals("0") &&
+                        mutationType != null &&
+                        !mutationType.equalsIgnoreCase("IGR")) {
+                    ProgressMonitor.logWarning(
+                            "Treating mutation with gene symbol 'Unknown' " +
+                            "and Entrez gene ID 0 as intergenic ('IGR') " +
+                            "instead of '" + mutationType + "'.");
+                    mutationType = "IGR";
```
pieterlukasse: let's do something like 
```
myMutationFilter.igrRejects++;
entriesSkipped++;
continue;
```
and keep `if` below simple (just `if (gene == null)` )

pieterlukasse:
@n1zea144 , @zheins : I reviewed this one. I will do some tests and update it here. If you can approve this code and tests are OK, we can merge this one.

pieterlukasse:
tested with extended set from @aderidder and all seems fine :+1:

zheins:
I just tested it, looks good! :+1:

PR #1148, opened 2016-04-29
disabled stomach and esophagus study when init cross-cancer expression plots

yichaoS:


jjgao:
seems not working in [the heroku instance](https://cbioportal-pr-1148.herokuapp.com/)

PR #1164, opened 2016-05-02
Update to the latest Mutation Mapper RC

onursumer:
This PR includes the latest release candidate of the mutation mapper.

Main notes:
  - Initialization of the 3D visualizer is delayed until the 3D button is clicked
  - New information panel which shows number of mutations by mutation type
  - Supports genome-nexus connection for annotation data (disabled by default)
  - Supports connection to a separate PDB service (disabled by default)
  - Better dynamic data management


onursumer:
I had to make some changes to make some of the front-end components compatible with the latest version. As far as I tested everything works fine, but we may need few more iterations before the release.

@yichaoS: Can you please test the plots and co-expression pages to make sure that mutation data is handled properly?

@zhx828: Can you please test the annotation column of the mutation table?

onursumer:
Update:
 - Improved information panel
 - Auto-adjusting mutation diagram upon filtering
 - Separated mutation summary view

PR #1165, opened 2016-05-03
Oncoprint polish

adamabeshouse:
1) Set sample and gene order in oncoprinter
2) Render labels at double resolution for retina displays
3) Add "Loading oncoprint" message to loading bar
4) Hide "Case Set: " until message loaded
5) Left-align and increase vertical margin of "Show all samples" button


adamabeshouse:
replaced by https://github.com/cBioPortal/cbioportal/pull/1173

Issue #1111, opened 2016-04-26
link to darwin patient view


n1zea144:
We want to provide a link to the darwin patient view from the cBioPortal patient view when a user is looking at a patient from the MSK-IMPACT study (and when the user is not coming from the MSK-CIS system - see me for more info).

First we need to check if the user has access to the patient data in darwin:

Invalid:
http://darwin.mskcc.org/authCheck/authCheck.jsp?p_userName=XXeubankm&p_dmp_pid=P-0005487

{
darwinAuthResponse: "invalid",
p_userName: "XXeubankm",
p_dmp_pid: "P-0005487"
}

Valid:
http://darwin.mskcc.org/authCheck/authCheck.jsp?p_userName=eubankm&p_dmp_pid=P-0005487

{
darwinAuthResponse: "valid",
p_userName: "eubankm",
p_dmp_pid: "P-0005487"
}

If the user has access, we can present a link on the patient view:

http://darwin.mskcc.org/resources/ptviewer.html?p_DMP_PID=XXX

Note, the checking of user access and the construction of the link should happen in the backend - servlet layer.

PR #1168, opened 2016-05-03
Add link to Darwin patient view (issue #1111)

angelicaochoa:
Darwin link only appears for authorized users to patient's clinical data.

@n1zea144  Current link is indicated by bold green text - is there a logo/format that should be used instead?


core/src/main/java/org/mskcc/cbio/portal/servlet/CheckDarwinAccess.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ *
+ * @author ochoaa
+ */
+public class CheckDarwinAccess {
+    private static String DARWIN_AUTH_URL;
+    private static String DARWIN_RESPONSE_URL;
+    
+    public static String checkAccess(String userName, String patientId){
```
n1zea144: We also want to take into account the cancer study - linking to darwin should only be relevant to mskimpact study/patients (we actually have many merged studies which contain mskimpact, but no easy way to determine that).  Also, we don't want to provide a link to darwin if the user came from the CIS system (see me to explain this).

core/src/main/java/org/mskcc/cbio/portal/servlet/CheckDarwinAccess.java
```diff
@@ -0,0 +1,124 @@
+/*
```
n1zea144: We should move this class to the Util package.

core/src/main/java/org/mskcc/cbio/portal/model/DarwinAccess.java
```diff
@@ -0,0 +1,216 @@
+/*
```
n1zea144: We should make this class a nested class within the CheckDarwinAccess class.

angelicaochoa:
@n1zea144 done and done

morungos:
Wonderful that you're doing this, although I'd really like it to be through a plugin some day. Can we maybe use this as a test case for that?

n1zea144:
@morungos Yes, me too.  In the meantime, if you have any suggestions for insulating this from the general codebase, let us know.

angelicaochoa:
@n1zea144 @jjgao PR should be ready although right now the link to the darwin patient view is indicated by bold green text. Wasn't sure what should be used instead.

morungos:
@n1zea144 I'm now thinking war overlays might do most, if not all of this. http://maven.apache.org/plugins/maven-war-plugin/overlays.html. Although it'll be easier once we've slimmed down a bunch of the configuration. Certainly this will allow the Java and web components to be overridden by an individual deployment without pushing them into core.

n1zea144:
@morungos I'm catching up on your comments about war overlays in issues #1299.  We are certainly open to refactoring this code using overlays to prevent custom code from entering into the core.

PR #1171, opened 2016-05-04
Replace phantomjs with selenium chrome and ff test

inodb:
# What? Why?
We have been running into issues using PhantomJS for making screenshots. ReactJS does not seem to work well, there are also some problems with flexbox support. By testing the browsers we are actually using instead (i.e. firefox & chrome) this will no longer be a problem.

- Use docker-compose to create selenium grid and run tests on firefox and
  chrome instead of phantomjs
- Make screenshots using both firefox and chrome
- Move configurables of screenshot to a screenshot.yml

Pros:
- Instead of testing phantomJS and changing our code to work with PhantomJS we test using firefox and chrome
- Since docker containers are used for making the screenshot on travis in this PR, it is trivial to run the test locally and get the same screenshots

Cons:
- The phantomJS set up was easier to grasp. You install phantomJS and off you go. This testing with a selenium grid is slightly more complicated for those unfamiliar with docker.
- We are storing images in the repo which doesn't scale very well. For our purposes I think it should be fine to add a few more screenshot tests, but we'll have to consider using a service e.g. [sauce labs](https://saucelabs.com/) if we want to do more fancy end to end testing. I do think we should be adding more unit and integration tests first though, see http://googletesting.blogspot.com/2015/04/just-say-no-to-more-end-to-end-tests.html

# Proposed reviewers
@zhx828 @jjgao @pieterlukasse @fedde-s


zhx828:
@inodb great work done. And in addition, since we test our code in Chrome all the time and Firefox sometimes, the most important part is to generate screenshots in IE. Is it easy to be implemented?

inodb:
Thanks! The problem is that IE does not run on docker at the moment: http://stackoverflow.com/questions/32338675/, so we would have to come up with another solution for that. One option would be to use [Sauce labs](https://saucelabs.com/selenium/), which offers IE testing and also uses selenium. I think it would be possible to use their infrastructure instead of using a local docker grid on Travis, see e.g.: https://docs.travis-ci.com/user/sauce-connect/. There's a free tier for open source projects. I think for now though we could stick with using this PR and come up with a solution for IE later. By using selenium we're one step closer to making that possible.

inodb:
As a side note, after speaking to @sheridancbio this morning. We could use a similar approach with a `docker-compose.yml` for scaling cbioportal and the databases instances on dashi2. See also: https://www.brianchristner.io/how-to-scale-a-docker-container-with-docker-compose/. We should have a separate discussion about this, but this PR can perhaps be used as an example.

jjgao:
@inodb is this one ready to go?

inodb:
Yup! Just rebased on rc

PR #1176, opened 2016-05-04
Fix to remove some errors given by validator

pieterlukasse:
2 quick fixes:
- The line end validation is hereby removed. 
- The clinical data file now allows blanks in most of its data columns


pieterlukasse:
@priti88 : can you please test and report results here?

priti88:
@pieterlukasse I tested the validator and it works as expected.
@jjgao can you also merge this to our rc-for-dfci branch?
Thank you.

fedde-s:
@pieterlukasse You made several comments like the one below unnecessary, but only removed one of them. Could you remove the others as well to avoid confusion?

```Python
                 # uncomment to overwrite with the new version
                 #'--error_file', 'test_data/study_maf_test/error_file.txt']
```

pieterlukasse:
@fedde-s : good catch. Done.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1192,6 +1181,7 @@ class ClinicalValidator(Validator):
     REQUIRE_COLUMN_ORDER = False
     PROP_IS_PATIENT_ATTRIBUTE = None
     NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]","[completed]","[null]", ""]
```
n1zea144: I thought in a conversation yesterday you wanted us to replace 'null' with 'na' (at least for expression data).  Is 'na' missing from this list?

pieterlukasse: Yes, only for expression data. See also #1221

For clinical data this list is basically a subset of what is allowed in the java layer. So no, 'na' should not be in this list (since it is also not in https://github.com/cBioPortal/cbioportal/blob/rc/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java#L63)

fedde-s: In other words, the list on the line referred to here is specific to clinical data files.

pieterlukasse:
@n1zea144 I fixed #1221 in #1304 . Could you review that one?

PR #1177, opened 2016-05-04
508 compliance

aderidder:
Added alt property to many of the img tags for 508 compliance


Sjoerd-van-Hagen:
Some of the alts could be more descriptive, e.g. 3d vis -> 3D visualization.

portal/src/main/webapp/WEB-INF/jsp/mutation/mutationMapperTemplates.html
```diff
@@ -37,7 +37,7 @@
 		None of the mutations can be mapped onto this structure.
```
onursumer: Just realized this. Actually this file is not maintained within portal code base. This is generated by mutation-mapper project, and is usually overwritten when we update the mutation mapper library. So we need to fix this in the actual library repository too: https://github.com/cBioPortal/mutation-mapper/issues/37

PR #1192, opened 2016-05-05
Added mixin configurations and an example controller

ersinciftci:


sheridancbio:
I think the Web module may also be dependent on the Model module (Ersin may have already said this)
CustomObjectMapper registers the mixins in the mixinMap and does some policy setting such as super.configure(JsonGenerator.Feature.WRITE_NUMBERS_AS_STRINGS, true);

sheridancbio:
I have not yet deployed / tested this.

ersinciftci:
"I think the Web module may also be dependent on the Model module (Ersin may have already said this)"
- There is a transitive dependency to Model

"Looks good, but one concern is that it appears that returned fieldnames will be camelCase fields"
- You might have missed `super.setPropertyNamingStrategy(PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);` in CustomObjectMapper

Thanks for the fast review.

PR #1194, opened 2016-05-05
Gene Panel data schema, api, and importer

zheins:


jjgao:
@zheins I have a second thought about this. I think we should try to incorporate the germline use case in the data model for the release. Let's discuss.

jjgao:
Could you also update the data format doc in the same pull request?

zheins:
Dead PR, accomplished in PR #1618

PR #1201, opened 2016-05-09
Recommendations against MySQL timeout from Wiki in code

holtgrewe:
Setting the testOnBorrow and validationQuery on the data source
in code and XML as recommended in the cBioPortal wiki.


morungos:
These look sensible to me. I'm using them in production.

holtgrewe:
What's stopping this from going into rc?

inodb:
@holtgrewe Thanks for submitting this. Why we are not adding the other  parameters here as well e.g.: `connectionProperties="zeroDateTimeBehavior=convertToNull;" driverClassName="com.mysql.jdbc.Driver" maxActive="100" maxIdle="30" maxWait="10000"`?
I guess the idea would be that you can overload these values in `context.xml`. Does that make sense @n1zea144 ?

Could be good to update the documentation in the same PR.

jjgao:
closing this since it's so long ago. Please feel free to reopen if still needed.

PR #1202, opened 2016-05-09
Read the SWISSPROT column in MAF files as an accession

fedde-s:
Read the values in the `SWISSPROT` column of mutation data files as UniprotKB accessions instead of names, as this is what VEP and vcf2maf/maf2maf list in this column. Make the dataset validation script check the values in this column and notify the user if these values are invalid or unknown to the portal, by implementing a web API service that returns the entries cached in the database table `uniprot_id_mapping`.


fedde-s:
Can @aderidder still test this?

aderidder:
See screenshot for validator results (relevant: SWISSPROT column, Error column and Warning column)
image

Please note that the loader finds all of them valid. Also, no issue in front-end for e.g. 'Pieter' accession.

pieterlukasse:
code looks good to me :+1:

pieterlukasse:
@fedde-s : as discussed, we need to update this code with some backwards compatibility for legacy  studies. The way to do this would be to introduce a new variable in the mutations meta file. This field would be a simple flag telling the loader how to parse the SWISSPROT column. E.g. `content_swissprot=name` or `content_swissprot=accession`. If the property is missing in the meta file, the default should be the old way `content_swissprot=name` (with a WARNING).

pieterlukasse:
@fedde-s : one more fix that can be made here is: replace the log.error by a log.warning in the validation when the uniprot code cannot be found in DB (issue encountered by @jochemb). The code in `PfamSequenceServlet.java` should be able to handle the situation where the uniprot accession cannot be found in db.

fedde-s:
@pieterlukasse Would it be better to have unusable/invalid accessions with NULL names or names with NULL accessions (depending on the type of data file) in the `mutation_event` table, or to set them both to NULL in that case to treat it sort of as an informal foreign key?

pieterlukasse:
@fedde-s : let's keep it at the validator change for now regarding my last comment.

fedde-s:
For completeness: I’m working on a new pull request incorporating the feedback.

fedde-s:
Replaced by #1366

PR #1203, opened 2016-05-09
WebGL oncoprint changes, including coloring based on recurrence

adamabeshouse:
(1) Color mutations based on recurrence
(2) Fix legend aliasing issue
(3) Fix legend being cut off on tab switch
(4) Only zoom to fit on initialization if at least 200 cases
(5) Show an extra decimal point if less than 3%
(6) Fix oncoprinter file uploading bug


jjgao:
@adamabeshouse could you resolve the conflicts?

PR #1209, opened 2016-05-10
fix for failing CNA data when there is 'new CNA event' in a new study

pieterlukasse:
Fixes remaining issue in #844


pieterlukasse:
@priti88 : can you please verify that it fixes the issue on your side as well?

pieterlukasse:
@n1zea144 @zheins : this was a cause for the MySQLbulkLoader exception. As discussed yesterday, I think you might want to merge this one.

priti88:
I tested this issue and it's fixed.

PR #1211, opened 2016-05-10
Converted geneticprofiledata web_api endpoint to the new architecture

ersinciftci:
There is no logging yet because I want to discuss how to do logging. I'll add an automated end-to-end test script after Stuart's integration test is merged.

### DIFF
--- a/business/pom.xml
+++ b/business/pom.xml
@@ -17,6 +17,11 @@
       <artifactId>commons-collections</artifactId>
       <version>3.2.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.mskcc.cbio</groupId>
+      <artifactId>service</artifactId>
+      <version>${project.version}</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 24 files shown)


business/pom.xml
```diff
@@ -17,6 +17,11 @@
       <artifactId>commons-collections</artifactId>
       <version>3.2.2</version>
     </dependency>
+    <dependency>
```
n1zea144: Should this be:
org.cbioportal.service

ersinciftci: groupId of modules are inherited from the parent pom, so it's still org.mskcc.cbio since we didn't change it. All the modules in a project shares the same groupId.

web/src/main/java/org/mskcc/cbio/portal/web/api/ApiController.java
```diff
@@ -233,7 +234,7 @@
             notes = "")
     @Transactional
     @RequestMapping(value = "/geneticprofiledata", method = {RequestMethod.GET, RequestMethod.POST})
-    public @ResponseBody List<DBProfileData> getGeneticProfileData(
+    public @ResponseBody List<Serializable> getGeneticProfileData(
```
n1zea144: Should we be applying mixins here?

ersinciftci: They are already applied, I think you are confused because I didn't use @View annotation, it’s just to create a different view of models for different services, which I didn’t need for that service, because we don’t need to hide anything for that web_api service specifically or have a different property name just for that web_api service.

PR #1212, opened 2016-05-10
fix stable_id issues, add option for mixed_attributes, allow multiple MAF files in loader

zheins:


zheins:
@pieterlukasse please review and let me know if I need to point this to a different branch.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java
```diff
@@ -496,6 +496,8 @@ public static void main(String[] args) throws Exception {
 	               "meta (description) file" ).withOptionalArg().describedAs( "meta_file.txt" ).ofType( String.class );
 	        OptionSpec<String> study = parser.accepts("study",
 	                "cancer study id").withOptionalArg().describedAs("study").ofType(String.class);
+	        OptionSpec<String> attributeFlag = parser.accepts("a",
+	                "Flag for using MIXED_ATTRIBUTES").withOptionalArg().describedAs("a").ofType(String.class);
```
pieterlukasse: I would further describe this as a "deprecated" option, not to encourage others to start using it since we actually want to go for the split samples / patients format.

pieterlukasse:
I added only 1 minor comment. The rest looks good :+1:

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -75,8 +75,14 @@ public static GeneticProfile loadGeneticProfile(File file) throws IOException, D
...
          // the dbms already contains a GeneticProfile with the file's stable_id. This scenario is not supported
     	 // anymore, so throw error telling user to remove existing profile first:
-         throw new RuntimeException("Error: genetic_profile record found with same Stable ID as the one used in your data:  "
+         if (!existingGeneticProfile.getDatatype().equals("MAF"))
+         {
+             throw new RuntimeException("Error: genetic_profile record found with same Stable ID as the one used in your data:  "
                   + existingGeneticProfile.getStableId() + ". Remove the existing genetic_profile record first.");
+         }
+         else {
+             return existingGeneticProfile;
```
pieterlukasse: just a FYI - this should be fine since the `mutation` table has an active constraint (I introduced in `migration.sql` some time ago - see `UQ_MUTATION_EVENT_ID_GENETIC_PROFILE_ID_SAMPLE_ID` ) to prevent duplicated data that could occur by mistake (one of the possible mistakes with this new code would be to submit the same mutation data file twice - if this mistake is made, MysqlBulkLoader should throw an exception because of the UQ constraint above).

zheins:
Comments made in the code and added a deprecated description to the new attribute.

PR #1222, opened 2016-05-12
Added updating statistics after mutation import

j-hudecek:
Table mutation can get pretty big and without statistics update I often see queries that use it take more than 30 seconds. Running ANALYZE TABLE fixes it.


jjgao:
@j-hudecek Thanks! Could you please send the pr to hotfix branch?

PR #1223, opened 2016-05-12
Created /mutationCounts web api to return mutation counts per sample

ersinciftci:
This web api is the equivalent of MutationJSON.processCountMutationsRequest method.


ersinciftci:
Just realized I forgot to add sample type filtering, fixing now.

morungos:
When this is done, let me know, as I can play with adding an integration test for it. Just to be nice, as there's enough here to deserve it.

ersinciftci:
Ok fixed.

Issue #1113, opened 2016-04-26
Remove calls to System.exit() from command logic


morungos:
We've found a few issues for failing tests in https://github.com/cBioPortal/cbioportal/pull/963. Ultimately, the problem is that some of the `main()` methods call `System.exit()`, which kills the JVM. This makes testing impossible.

We should refactor all the classes that expose a `main()` to be a `Runnable`, where the `main()` invokes the `Runnable.run` and turns any thrown exceptions into a non-zero exit status. The classes can then be tested through their `Runnable` interfaces, and there's no logic left to speak of in `main()`.

At least: that's now up for discussion.

fedde-s:
I don’t think System.exit() strictly makes testing impossible, there’s things like http://stefanbirkner.github.io/system-rules/index.html#ExpectedSystemExit and `SecurityManager`s. But having the main() function as a separate outer layer that basically does nothing but map out which exception types yield which exit statuses for this particular command may be a valid idea.

fedde-s:
It may then make sense to subclass `Runnable` put that mapping logic in a separate function, and have all of the `main` functions do nothing but call that function. It would help standardising the meaning of exit statuses like the BSDs aim to do in their [sysexits](https://www.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3) header, although I don’t know if that’s worth the effort in this case.

ecerami:
@morungos:  @pieterlukasse will fix.

morungos:
Excellent. @fedde-s is right, `System.exit()` doesn't directly make testing impossible. It just much easier to test a very few paths involving `System.exit()` rather than the number we have now. And I would recommend standardising exit status, as this will be the normal pipeliney way to notify failure, which will always be a possible outcome.

pieterlukasse:
hi @morungos, I did a quick change to help in this discussion (not tested yet). Is your proposal something like this https://github.com/thehyve/cbioportal/commit/1bf83decc33fe6c5d23d3d17065b12e21b450188
?

morungos:
Yes, exactly like that! Good one.

morungos:
[...]

PR #1224, opened 2016-05-12
Separate System.exit calls from script functionality

fedde-s:
This is a proposal to solve issue #1113, by defining a superclass that implements `Runnable` to cleanly separate the the core functionality of the utilities in the `scripts` package from common logic specific to command line usage.


morungos:
OK, this is still throwing a `System.exit()` but I think it's from inside `ConsoleUtil`. This is using an exit status of 2 to notify command syntax errors. My `TransactionalScriptRunner` is also doing this, and in more or less the same way. I think maybe we need a new class of `CommandSyntexException` thrown here, and used in `main` to call `System.exit`, and then refactor so nobody calls `main` from within Java. Would this work?

morungos:
Don't know how to add to someone else's PR, but if you have a way you'd like for me to add to this, happy to do so. We need to get `rc` clean soon, and this is a useful part of it.

morungos:
I think the VM exit is in `TestPanCancerStudySummaryImport` -- that would fit, as it's calling main methods.

fedde-s:
Ah, thanks, I’d missed a System.exit(2) in ConsoleUtil, in a function that has a nearly identical name to a function I’d already adjusted. Also, I hadn’t adjusted the tests to call run() instead of main(String[]) yet, only made the run() methods and moved all (but one) of the System.exit()s to a separate shared method.

PR #1229, opened 2016-05-13
Added updating statistics after mutation import

j-hudecek:
Table mutation can get pretty big and without statistics update I often see queries that use it take more than 30 seconds. Running ANALYZE TABLE fixes it.


zheins:
@j-hudecek could you rebase? This looks ok to me. 

Would it make sense to maybe analyze some of the other tables as well? Or does the benefit really only come from running this on the mutation table?

j-hudecek:
The tests are failing so I thought I messed something up during the merging and created a new PR #1764 with a meaningful branch name. I just haven't noticed the hotfix branch doesn't build by itself. Anyway that PR is fresher and looks nicer so I'll close this one.
I guess we could run ANALYZE TABLE on every table we import into but I've only ever seen problems on the mutation

PR #1128, opened 2016-04-27
Use EFDT library for Mutated Gene and CNA tables; Update EFDT library

zhx828:
EFDT: [Enhanced Fixed Data Table](https://github.com/cBioPortal/enhanced-fixed-datatable)

This related to #1002, #850,

PR #1233, opened 2016-05-13
Fix screenshot from PR #1128

inodb:
PR #1128 updated the study view, but the screenshot wasn't updated, which causes the end-to-end test on Travis to fail.


fedde-s:
Thanks! This was interfering with my PRs too.

PR #1235, opened 2016-05-13
Fix patient mixed data; fix for MySQLbulkLoader to fix related unit test

pieterlukasse:
Fix to make the code more tolerant for a previously existing sample (instead of giving an error) for MIXED data type. 

A MySQLbulkLoader fix (discussed in #844) was also necessary to fix the related unit tests in `TestImportClinicalData`

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/MySQLbulkLoader.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/MySQLbulkLoader.java
@@ -73,27 +73,26 @@ public class MySQLbulkLoader {
     * @throws DaoException
     */
    public static int flushAll() throws DaoException {
-        try {
+	   int checks = 0;
+       PreparedStatement stmt = null;
+       boolean executedSetFKChecks = false;
+       try {
             Connection con = JdbcUtil.getDbConnection(MySQLbulkLoader.class);
-            PreparedStatement stmt = con.prepareStatement("SELECT @@foreign_key_checks;");
+            stmt = con.prepareStatement("SELECT @@foreign_key_checks;");
             ResultSet result = stmt.executeQuery();
             
             result.first();
-            int checks = result.getInt(1);
+            checks = result.getInt(1);
 
             stmt = con.prepareStatement("SET foreign_key_checks = ?;");
             stmt.setLong(1, 0);
             stmt.execute();
+            executedSetFKChecks = true;
             
             int n = 0;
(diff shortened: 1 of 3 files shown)


pieterlukasse:
@n1zea144 , @zheins : can you test this one for your clinical data? 

@morungos : this PR also includes the fix you need. I think I got the same errors you reported before while running the unit tests for this class, and with the changes in MySQLbulkLoader  these are now gone. Can you please review and check if it solves your problem as well?

morungos:
There are still a few test issues here, certainly it doesn't yet run clean for me. It's still calling `System.exit()` somewhere. However, I think that the errors reported are actually there, and the stream closed stuff has all gone, so we can focus on the errors themselves.

pieterlukasse:
@morungos : great, thanks for testing. The system.exit() issues are being solved in @fedde-s 's PR #1224 (it contains some of the classes, I expect he will add the remaining refactored classes to this PR)

zheins:
@pieterlukasse - this fixes one of the issues we talked about. I still get this error:

ABORTED! Error:  No data was added.  Please check your file format and try again.

The import code incorrectly counts data added by counting the number of samples added to the database. If no samples are added, it assumes no data was added.

pieterlukasse:
@zheins : thanks for testing. The error you mention should not occur any more unless  `numSamplesProcessed` counter is 0 at the end.... can you check why this is the case when processing your data? I think we are missing a use case here.

pieterlukasse:
The error message is misleading btw...it is not counting the number of samples added, but the number of samples processed now...so 0 would be a strange result indeed. Something wrong with the data?

zheins:
The exception thrown is on line 594. It occurs because a patient specific supplemental file (which we are specifying as "MIXED_ATTRIBUTES" type because of backwards compatibility) is still getting checked for sample specific attributes added. I think we just need to exclude "MIXED_ATTRIBUTES" type files from this check.

pieterlukasse:
@zheins : I divided the validation into a PATIENT/SAMPLE specific set of checks and one specific for MIXED data type regarding "No data added". See my last commit. Can you test this and check if all works fine now? I will run some tests on my side tomorrow when back in the office. Thanks.

zheins:
I think this should be good now. I might have a few more data issues to look into, but I was able to get more than one clinical supplemental file to import. I'll look into it first thing in the morning to make sure there is nothing else we need to address.

PR #1238, opened 2016-05-16
Added stample_list_id to returned Mutation list in Web API

sheridancbio:
- set to be equal to the value supplied in the webAPI call
- new class temporarily added to wrap Mutation and add sample list id attribute


sheridancbio:
This is a fix to make the WebAPI consistent with the previous behavior.
When geneticprofiledata is requeted, one possible parameter is sample_list_id, and if this value is supplied, the webapi inserted this value into a field of all returned objects. The value is not constructed from the database, and this behavior cannot be modeled from the data.

For now, a new class has been added: MutationWithSampleListId, and after the new Mutation Service returns a modeled list of Mutation objects, this adaptor class is used in the service layer to reflect back the supplied sample_list_id to the client (in addition to the fields in the Mutation class).

I recommend that the client (browser code) be rewritten to do this result modification on the client side (if it is needed at all). Then this code can and should be dropped from the backend.

ersinciftci:
You can change sample_list_id property name to camel case, jackson will convert it back to underscore in json anyways. And instead of MutationWithSampleListId extend from Mutation, it could have Mutation as property and we could make it flat with mixin, so you wouldn't need to create that long constructor in MutationWithSampleListId. And since we use MutationWithSampleListId just for viewing purposes, we can put it in web module instead of model, so we wouldn’t need mixin, just annotate the class itself to make it flat.

sheridancbio:
I have incorporated Ersin's suggestions. [model class is now in service module]

ersinciftci:
Rerun the build because it fails randomly sometimes. Looks good to me, thanks Rob.

PR #1246, opened 2016-05-16
Change mutation sort order and legend order

adamabeshouse:


morungos:
Bad problem:  3522ef5 includes a log4j.properties that it shouldn't, and which (for me) gets a templated location for a file system that breaks separation between build and production. This file commit needs to be gone.

jjgao:
@morungos sorry I didn't review this carefully.

@adamabeshouse could you revert the change to log4j.properties? It should not be checked in.

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1251

PR #1256, opened 2016-05-18
More random test fixes

morungos:
Contains a small assortment of test fixes. These are largely data files that were triggering import validation errors and therefore expected exits. In an ideal world, we'd add more tests for these, so that the expected errors for each case are also checked.


core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportCopyNumberSegmentData.java
```diff
@@ -82,7 +82,8 @@ public void testImportSegmentDataNewStudy() throws Exception {
 
         String[] args = {
         		"--data","target/test-classes/segment/data_cna_hg19.seg",
-        		"--meta","target/test-classes/segment/meta_cna_hg19_seg.txt"        		
+        		"--meta","target/test-classes/segment/meta_cna_hg19_seg.txt",
+        		"--loadMode", "bulkLoad"
```
pieterlukasse: @fedde-s has another fix for this one. He will propose something

morungos:
@pieterlukasse @fedde-s OK, let me know when you have input on this.

core/src/test/java/org/mskcc/cbio/portal/util/TestCancerStudyReader.java
```diff
@@ -68,8 +68,25 @@ public void testCancerStudyReader() throws Exception {
...
-                  "brcaxxx is not a supported cancer type."));
+    	 assertEquals("brcaxxx is not a supported cancer type.", e.getMessage());
       }
    }
+
+	@Test
+	   public void testCancerStudyReaderShortName() throws Exception {
+
+	      File file = new File("target/test-classes/cancer_study.txt");
+	      CancerStudy cancerStudy = CancerStudyReader.loadCancerStudy( file );
```
pieterlukasse: @ersinciftci seems to to this in `src/test/resources/` instead of `target/test-classes/`. I believe `src/test/resources/` avoids a compilation step when just changing the test file, so I prefer that.

see https://github.com/cBioPortal/cbioportal/pull/963/files#r58565966

core/src/test/java/org/mskcc/cbio/portal/util/TestCancerStudyReader.java
```diff
@@ -68,8 +68,25 @@ public void testCancerStudyReader() throws Exception {
...
-         assertTrue( e.getMessage().equals( 
-                  "brcaxxx is not a supported cancer type."));
+    	 assertEquals("brcaxxx is not a supported cancer type.", e.getMessage());
       }
    }
+
+	@Test
+	   public void testCancerStudyReaderShortName() throws Exception {
+
+	      File file = new File("target/test-classes/cancer_study.txt");
```
pieterlukasse: @ersinciftci seems to to this in `src/test/resources/` instead of `target/test-classes/`. I believe `src/test/resources/` avoids a compilation step when just changing the test file, so I prefer that.

see https://github.com/cBioPortal/cbioportal/pull/963/files#r58565966

pieterlukasse:
@morungos : nice work. I made some small comments :+1:

core/src/test/java/org/mskcc/cbio/portal/util/TestCancerStudyReader.java
morungos: Good point. The only need for the `target` version is when Maven filtering is in the way, and it shouldn't be except when we really need it. I'll amend

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportCopyNumberSegmentData.java
pieterlukasse: @morungos : never mind. We discussed it here and decided to keep your solution in this case. So, just leave the `loadMode` there

morungos: OK, I'll just make some amendments and update the PR

morungos:
Ok, fixed the paths. There's still so much here I want to tweak, but it's best I get off the critical path and let you get on with stuff.

pieterlukasse:
hi @morungos , sorry for the delay. I still get the following error:
```
org.apache.commons.lang.exception.NestableRuntimeException: Can't call main method in: org.mskcc.cbio.portal.scripts.ImportCancerStudy
```
which is a bit confusing since it is actually trying to call the `run` method. The class `ImportCancerStudy` does have the `run` method so I will check why it is not finding it.

pieterlukasse:
@morungos : I made some changes to your code and got it working now:
```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.66 sec - in org.mskcc.cbio.portal.util.TestTransactionalScripts
```
:)

You can find my changes here: https://github.com/thehyve/cbioportal/commit/7dd218a3d287978e2420285539184c6bc9cf4d7f

Basically, since we have an Interface now, you can use that to simplify your code. I think the error was in that the previous code was looking for a static method called run (? didn't verify this, but my change worked). The method `run` will also not return an exit value but throw RuntimeException instead, so I adjusted that as well. The changes in `demoScript.xml` are based on how we refer to test files in other unit tests as well. This seems to work fine. Somehow the code `systemEnvironment['PORTAL_HOME']` was not working for me, even though I had set PORTAL_HOME with `export PORTAL_HOME=/cbioportal`.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportCopyNumberSegmentData.java
```diff
@@ -81,8 +81,9 @@ public void testImportSegmentDataNewStudy() throws Exception {
         //CancerStudy study = DaoCancerStudy.getCancerStudyByStableId("testnew");
 
         String[] args = {
-        		"--data","target/test-classes/segment/data_cna_hg19.seg",
-        		"--meta","target/test-classes/segment/meta_cna_hg19_seg.txt"        		
+        		"--data","src/test/resources/segment/data_cna_hg19.seg",
+        		"--meta","src/test/resources/segment/meta_cna_hg19_seg.txt",
+        		"--loadMode", "bulkLoad"
         		};
```
pieterlukasse: test worked :+1:

core/src/test/java/org/mskcc/cbio/portal/servlet/TestWebService.java
```diff
@@ -84,6 +84,7 @@ public void setUp() throws DaoException {
 	   List<CancerStudy> mockTrue = new ArrayList<CancerStudy>();
 	   mockTrue.add(null);
 	   expect(control.isAccessibleCancerStudy(isA(String.class))).andStubReturn(mockTrue);
+	   expect(control.getUserDetails()).andStubReturn(null);
        replay(control);
        SpringUtil.setAccessControl(control);
```
pieterlukasse: test class works :+1:

core/src/test/java/org/mskcc/cbio/portal/util/TestCancerStudyReader.java
```diff
@@ -54,22 +54,39 @@
...
+	      assertEquals(expectedCancerStudy, cancerStudy);
+	      // TBD: change this to use getResourceAsStream()
+	      file = new File("src/test/resources/cancer_study_bad_short_name.txt");
+	      try {
+	         cancerStudy = CancerStudyReader.loadCancerStudy( file );
+	         fail( "Should have thrown DaoException." );
+	      } catch (Exception e) {
+	    	 assertEquals("short_name is not specified.", e.getMessage());
+	      }
+	   }
```
pieterlukasse: tests work :+1:

morungos:
This is some very nice decrufting and simplification. I've merged the changes in manually because I can't figure out how to get that commit right from the original source.

It actually didn't make any difference to the behaviour on my systems, but I'm using a standard Oracle JDK8, and reflection might be subtly different elsewhere. In any event, these are the right changes to make.

pieterlukasse:
@morungos : nice. Feel free to merge this PR. As you can see I tested all other parts (and earlier today the fix I proposed). The only real change I noticed is that you chose *not* to add `catch (RuntimeException e)`, which is a choice you can probably judge better since this affects your pipeline. So if all changes work fine on your side as well, I would say this is a go :+1:

morungos:
Good point, I chose not to add the `RuntimeException` mainly because it just rethrows a `RuntimeException` with a different message. This was information loss that could hinder debugging, is all. I'll merge now,

PR #1260, opened 2016-05-18
Remove risk of StringIndexOutOfBoundsException by using endsWith

### LINKED ISSUE #1182: Bad import data throws a string error
I've just loaded some data (note we're not using the validator) but we still managed to get a Java string index error:

```
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
    at java.lang.String.substring(String.java:1927)
    at org.mskcc.cbio.portal.servlet.PortalMetaDataJSON.cancerStudyMap(PortalMetaDataJSON.java:92)
    at org.mskcc.cbio.portal.servlet.PortalMetaDataJSON.doGet(PortalMetaDataJSON.java:244)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:687)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
```

The relevant fragment of `PortalMetaDataJSON` is: 

```
    for (SampleList pl: caseSets) {
        String suffix = pl.getStableId().substring(pl.getStableId().length()-4);
        if (suffix.equals("_all")) {
            numSamples = pl.getSampleList().size();
            break;
        }
    }
```

This is caused by a case_list containing the following:

```
case_list_name: All
stable_id: all
```

Since it seemed like most of the other stable identifiers were, on import, being now implicitly prefixed with the study name, this shouldn't have been an issue, as far as I can see. 

So at the very least, there's a significant inconsistency between the case list of use of stable identifiers, and that in genetic profiles.

morungos:
This should resolve #1182


morungos:
Hmm, too many commits. Guess I need to learn more git about rebasing when pulling from upstream :-(

PR #1262, opened 2016-05-18
Correct exception handling in TransactionalScriptRunner

morungos:
Fix a nasty error, which was always throwing even when the exception was null (!)

This is really only used at UHN now, although that could change, as it's the only transactionally-safe loading system.


inodb:
@morungos If UHN is the only one using this. Might as well merge it. Should this be to hotfix or rc?

morungos:
`rc` is fine, and merging will be a big help

inodb:
Merged in https://github.com/cBioPortal/cbioportal/commit/9a753689c40973d36979b61047110294fdcd9aec

PR #1274, opened 2016-05-23
Replace calls to DAOMutation with MutationRepository

ersinciftci:
Ignored MutationRepositoryTest until I solve the in-memory database naming issue, will try http://www.h2database.com/ instead of hsqldb.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ business/target/*
 model/target/*
 persistence/persistence-api/target/*
 persistence/persistence-mybatis/target/*
+persistence/persistence-mybatis-test/target/*
 service/target/*
 web/target/*
 mutation-assessor/target/*
--- a/business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
+++ b/business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
@@ -214,7 +214,7 @@ public class ApiService {
 		List<Serializable> result = new ArrayList<>();
 
 		if (!mutationProfiles.isEmpty()) {
-			result.addAll(addSampleListIdToMutationList(mutationService.getMutations(mutationProfiles, hugoGeneSymbols, sampleStableIds, sampleListStableId), sampleListStableId));
+			result.addAll(addSampleListIdToMutationList(mutationService.getMutationsDetailed(mutationProfiles, hugoGeneSymbols, sampleStableIds, sampleListStableId), sampleListStableId));
 		}
 		if (!nonMutationProfiles.isEmpty()) {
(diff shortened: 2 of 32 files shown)


ersinciftci:
Fixed MutationRepositoryTest using h2 database.

sheridancbio:
I've read through these changes. Comments:

I'm a little confused about the changes to GeneticAlterationUtil. The class is now annotated @Component which (to my limited spring knowledge) causes the implicit creation of a spring Bean which can then be used for application configuration. Inside the class is an Autowired annotated constructor. I am guessing this is how to give arguments to help with Bean instance construction. But the constructor sets two static properties which are used in member functions. Wouldn't this mean that each time the Bean is instantiated that it is possible these static members would be changing? Perhaps we know that this bean is only ever instantiated once? Something seems strange with this construction -- do we need a safety to only initialize the static members if they are uninitialized? (null)

MutationModelConverter class is very clear. I do wonder why this class is an @component? What circumstances do we imagine wanting to configure in a different Converter implementation? Maybe during testing? Otherwise, couldn't this just be a plain class which is instantiated normally? 

SpringBeanAutowiringSupport.processInjectionBasedOnServletContext used in MutationsJSON, NetworkServlet,... is a capability I did not know about ... glad to learn this. I guess this is how to get spring to resolve Autowired members in classes which are not spring annotated such as @Controller or @Service  (such as in "non-spring" servlet classes)

Replacements of functionality in other classes follows a consistent approach, using the converter as needed.

Adding an application-context for core is necessary for Autowiring the @Components? Is this needed in conjunction with the processInjectionBasedOnServletContext() calls? I guess the component-scan finds the @Components, and the AutowiringSupport wires them using the beans created from the component-scan? (I'm still digesting spring, obviously)

A Sample has a SampleType member now (an Enum rather than a String) : This is better for detecting and characterizing errors sooner.

The multi-mode call getMutations(4 args) has been renamed getMutationsDetailed(4 args) in MutationsRepository. Other calls have been added to support the other Dao functionality needed.

Count classes added to transfer summary objects from the repository / DBMS.

Tests added with .sql support/util scripts. I'm not familiar with hsqldb .. I guess we are using memory-tables for testing. Some older tests are adjusted. Repositories are updated with mybatis support for functionality of new clients. MutationMapper.xml is getting fairly long.

Looks good

n1zea144:
By default Spring manages singleton beans ([singleton scope](http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#beans-factory-scopes-singleton).

MutationModelConverter is being autowired into other classes which is why it is annotated.

In general, anything you want to autowire needs to be identified by Spring and without Spring Boot, you need to put a context file someplace to direct Spring where to look.

PR #1278, opened 2016-05-24
Fix Travis core tests to work with docker services enabled

inodb:
Currently the end-to-end tests are disabled, because the Travis core tests don't work with docker services enabled. This should fix the core tests to continue to work with docker services enabled.


inodb:
Still having some problems with enabling the docker service on Travis and getting the mysql core tests to work. It's either the end-to-end tests work when I enable the docker service, or the core tests work when I enable the mysql service. Tried a bunch of different things:

- Set sudo required to true. This way I can install mysql, but somehow it gives me buffer overflow errors when using openjdk7 and running the core tests. It does seem to work with other oraclejdk8 and some other jdk versions, but then somehow all the core tests fail. Maybe they only work for a specific version of jdk? Is that possible?
- Use mysql docker container. This seems the most reasonable option. Even though this works locally, on Travis I get the error: https://travis-ci.org/cBioPortal/cbioportal/builds/132730182#L199-L216
- Try to start docker services manually using sudo on Travis. (no service found)
- You can run multiple workers if you set multiple environment variables or jdks, but unfortunately I can't get multiple workers with separate services enabled.

If anybody has a suggestion, let me know!

inodb:
Switched to #1308

PR #1230, opened 2016-05-13
Fix1227

aderidder:
Fix for timeline data #1227 

The following needs to be changed in timeline files:
* change genetic_alteration_type to CLINICAL
* STOP_DATE is now a mandatory column (third column). The values should be blank when STOP_DATE is not used.

Issue #1227, opened 2016-05-12
Error in validating timeline file.


priti88:
cancer_study_identifier: xxxxx
genetic_alteration_type: TIMELINE
datatype: TIMELINE
data_filename: data_timeline_specimen.txt

ERROR: meta_timeline_specimen.txt: Could not determine the file type. Please check your meta files for correct configuration.; value encountered: 'genetic_alteration_type: TIMELINE, datatype: TIMELINE'

aderidder:
Hi,
There were some issues with the validator and we've created a fix. 

The following needs to be changed in timeline files:
* change genetic_alteration_type to CLINICAL
* STOP_DATE is now a mandatory column (third column). The values should be blank when STOP_DATE is not used. 

@pieterlukasse the loader code should be improved in how it handles the sometimes optional column STOP_DATE

pieterlukasse:
@priti88 can you confirm that the solution indicated by @aderidder works? We can make further improvements to the loader code at a later stage (e.g. if there is a need to have the columns in any order or if someone really wants to leave out the whole STOP_DATE column....but I think these are not so important now)

priti88:
@pieterlukasse : I added STOP_DATE Column. But, the validator fails with error:
ERROR: data_timeline_specimen.txt: lines [2, 3, 4, (17 more)]: column 3: Blank cell found in column; value encountered: ''' (in column 'STOP_DATE')'

pieterlukasse:
@priti88 : did you use the fix PR #1230 ? Sorry, it was not very clear in Sander's comment.

priti88:
No. I did not 
Can you send a PR for our branch as well ? :)

pieterlukasse:
@priti88 :ok, I'll do that tomorrow

pieterlukasse:
Hi @priti88  : I checked this, and you will need to merge #1277 PR first because @aderidder included some documentation update in #1230 and your custom "rc" branch does not yet have the new `/docs` folder.

pieterlukasse:
@priti88 : see #1279. You can test this one on your branch. Please report back here if it solved the problem.

priti88:
It is solved.

PR #1279, opened 2016-05-24
Fix for timeline data #1227 (but pointing to rc-dfci branch)

pieterlukasse:
Fix for timeline data #1227  (same as #1230, but pointing to rc-dfci branch)

The following needs to be changed in timeline files:
* change genetic_alteration_type to CLINICAL
* STOP_DATE is now a mandatory column (third column). The values should be blank when STOP_DATE is not used.

### DIFF
--- a/core/src/main/scripts/importer/cbioportal_common.py
+++ b/core/src/main/scripts/importer/cbioportal_common.py
@@ -176,6 +176,7 @@ META_FIELD_MAP = {
     MetaFileTypes.TIMELINE: {
         'cancer_study_identifier': True,
         'genetic_alteration_type': True,
+        'datatype': True,
         'data_filename': True
     },
     MetaFileTypes.CASE_LIST: {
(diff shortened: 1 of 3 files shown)


jjgao:
@pieterlukasse will this also be needed for rc branch later? rc-for-dfci was a hack. We need to make sure anything merged to that branch will be merged to rc as well because we are going to delete that branch after rc is released.

pieterlukasse:
@jjgao : yes, thanks for checking. All PRs to this branch are also made to `rc` . This one you can find in #1230 btw

jjgao:
@pieterlukasse: OK. Is this ready to merge?

pieterlukasse:
@jjgao : the idea is that @priti88 will test this first. Once she confirms this works, then we could merge. But we can merge first and patch later if necessary as well.

priti88:
@pieterlukasse : fix works. Thanks

Issue #1162, opened 2016-05-02
Add a JUMP icon in each study in the main query page to jump to a patient


jjgao:
Request from Barry:

"Right now, on the front page, when you hover over the MSK-IMPACT cohort, a little “I” icon appears as does the button to the study summary page. Ordinarily, when I have a DMP ID I want to see the profile for, I go to the study summary page, then click “Select Cases by ID”, add my ID, wait for the page to dynamically find the sample, then I click the highlighted red sample ID in the CNA/Mutations plot to go to the patient view page for that sample. Would be really nice on the front page to have a “Jump to Sample” option. How about when you hover over the study name on the front page, we add another little icon that when clicked, lets you paste in a sample ID right there in a popup field and then “Jumps to Sample”. That would take out a lot of steps from the process and save lots of queries on the slower study page."
- [ ] add a j button after the i button
- [ ] mouse over: text "Click to jump to a specific patient"
- [ ] mouse click: a text field to input patient ID and a button "Just to patient"
- [ ] after submit, open patient view in a new tab, e.g. http://www.cbioportal.org/case.do?cancer_study_id=ucec_tcga_pub&case_id=TCGA-BK-A0CC

hint: try to use [cbio.util.getLinkToPatientView](https://github.com/cBioPortal/cbioportal/blob/master/portal/src/main/webapp/js/src/cbio-util.js#L364)

![image](https://cloud.githubusercontent.com/assets/840895/14963212/98a4a776-1070-11e6-83a2-9b72e99072ce.png)

jjgao:
Since we are redesigning the tree, I am close this pr.

PR #1282, opened 2016-05-24
Patient shortcut, superceeds #1281

pambot:
This is the enhancement listed under #1162. A search icon from Font Awesome was added after the information icon and mouseenter will open a QTip with a form that lets you put in the TCGA patient ID and submit it to get to a new window with that specific patient. The QTip has a 1 second fade delay when inactive.

Closed previous one due to comparing to wrong branch.


jjgao:
@pambot thanks for the pull request. I noticed you put all the code in jstree.js. I forgot to mentioned that we shouldn't touch any code in the js/lib directory. It suppose to hold only 3rd party js libs. Could you please move the code out to the same place as the (i) icon lives?

jjgao:
@adamabeshouse is jstree.js the original file or is it a modified version. @pambot needs to add an icon after the (i) icon. What would you suggest?

adamabeshouse:
@pambot jstree.js is a modified version. I did my best to be clean and follow convention of the rest of the code but there is some gross stuff and hacks... let me know if anything's confusing.

To do that you should look at the method show_node_decorators on line 2902. Specifically, you'll want to add something in the ```if``` block in 2910:2940, which deals with leaves of the jstree, which are studies.

jjgao:
@adamabeshouse could you review the code then?

adamabeshouse:
It wasn't working for me because the qtips weren't being destroyed and so the input value was shadowed by the other existing forms with other patient ids. But adding a line to ```hide_node_decorators``` to destroy the qtips, as shown below, fixed it for me:

```
hide_node_decorators : function(obj) {
    obj = this.get_node(obj, true);
    obj.children('.jstree-external-node-decorator').hide();
    obj.children('.jstree-node-decorator').qtip('destroy'); // <- this line
    obj.children('.jstree-node-decorator').remove();
}
```

adamabeshouse:
Hi @pambot,

The listeners stay because they are attached to ```document``` and not to the node decorator object. Maybe you could add a data attribute to the form element, so something like
```
'<form id="psform" data-study-id="' + node.id + '"><input type="text" ..... 
```
Then load it in the event handler. That way you could add the event somewhere once.

adamabeshouse:
Hi @pambot,

When you add an event listener by doing ```$(document).on("submit", "#psform")```, it adds the event listener to ```#psform``` but does so by "listening on ```document``` and catching it on ```#psform```", so since ```document``` is never destroyed, neither are the event listeners, but they are added each time you mouseover a study.

The solution is to add the event listener to ```document``` as you're doing, but only do it once, perhaps at jstree initialization. Now since the callback needs information about the node, add a ```data-``` attribute to ```#psform-text``` in the qtip content, and read it in the event handler.

Does that make sense?

jjgao:
Closing this. New intercing is being implemented.

PR #1284, opened 2016-05-25
Deleted TestImportExtendedMutationData.testImportExtendedMutationDa…

ersinciftci:
…taWhitelisted2


jjgao:
I think the whitelist function is no longer used. Maybe just remove the code?

ersinciftci:
Deleted the test method.

PR #1294, opened 2016-05-26
properly check for expression commands in OQL using new parser

adamabeshouse:


pieterlukasse:
seems good to me :+1:

pieterlukasse:
I tested on /beta and it seems to work fine :+1:

Issue #1291, opened 2016-05-26
Mutation view hangs when pfam_graphics entry is not present


pieterlukasse:
From my preliminary analysis it seems that the mutation view UI will hang when a gene has a uniprot accession (in `uniprot_id_mapping` table) but this same uniprot entry does not have any entries in the `pfam_graphics` table. Screenshot:

![image](https://cloud.githubusercontent.com/assets/2900303/15578154/aab3350c-235f-11e6-817e-ab221d686f11.png)

Steps to reproduce: 

- Execute this query:
http://www.cbioportal.org/beta/index.do?tab_index=tab_visualize&cancer_study_id=brca_tcga&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=brca_tcga_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=brca_tcga_gistic&Z_SCORE_THRESHOLD=2.0&case_set_id=brca_tcga_cnaseq&case_ids=&gene_list=ADORA3%20CSH1%20CSH2%20&gene_set_choice=user-defined-list&Action=Submit&show_samples=false&
- go to mutations tab, select ADORA3 gene

pieterlukasse:
I think this can be fixed in `PfamSequenceServlet.java`. I will propose a fix for this.

pieterlukasse:
@onursumer : FYI - I'll try to make a fix for this one. Seen this before?

onursumer:
This may also be related to https://github.com/cBioPortal/mutation-mapper/issues/39

onursumer:
In any case we should fix the mutation mapper library. Looks like currently the execution stops if there is no PFAM data is present: https://github.com/cBioPortal/mutation-mapper/blob/rc/src/js/controller/MutationDetailsController.js#L228

pieterlukasse:
@onursumer : I agree there should be a fix in the mutation mapper library as well. I will submit a fix for the java code in `PfamSequenceServlet.java` just to make this scenario less frequent.

pieterlukasse:
@onursumer : I made a PR for the java side. Can you review and merge this if all is OK? 

I also logged a separate ticket for the UI side on your repo: https://github.com/cBioPortal/mutation-mapper/issues/41

Thanks!

PR #1296, opened 2016-05-27
fix to show a dummy domain event when there is no pfam data

pieterlukasse:
This alleviates the issue reported at #1291


onursumer:
@pieterlukasse : this looks good to merge.

PR #1297, opened 2016-05-27
Fix more unit tests

pieterlukasse:
Fixed a couple of tests, making sure cache is cleared correctly before and after test execution. One of the tests was also calling a TRUNCATE table SQL execution, which cannot be rolled back and affects other tests depending on the order of execution of the tests.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
@@ -58,6 +58,7 @@ public class DaoGeneOptimized {
     private static final String GENE_SYMBOL_DISAMBIGUATION_FILE = "/gene_symbol_disambiguation.txt";
         
     private static final DaoGeneOptimized daoGeneOptimized = new DaoGeneOptimized();
+    //nb: make sure any map is also cleared in clearCache() method below:
     private final HashMap<String, CanonicalGene> geneSymbolMap = new HashMap <String, CanonicalGene>();
     private final HashMap<Long, CanonicalGene> entrezIdMap = new HashMap <Long, CanonicalGene>();
     private final HashMap<String, List<CanonicalGene>> geneAliasMap = new HashMap<String, List<CanonicalGene>>();
@@ -70,6 +71,10 @@ public class DaoGeneOptimized {
      * @throws DaoException Database Error.
      */
     private DaoGeneOptimized () {
+        fillCache();
+    }
+    
+    private synchronized void fillCache() {
         try {
             //  Automatically populate hashmap upon init
             ArrayList<CanonicalGene> globalGeneList = DaoGene.getAllGenes();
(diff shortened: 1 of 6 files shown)


pieterlukasse:
@ersinciftci : can you review and merge is all is OK? At my side all tests run fine now.

ersinciftci:
@pieterlukasse looks good to me.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -125,6 +130,26 @@ private DaoGeneOptimized () {
         }
     }
 
+    private void clearCache()
```
jjgao: Why do we need this? Are we updating genes on the fly?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoTypeOfCancer.java
```diff
@@ -147,6 +154,18 @@ public static void deleteAllRecords() throws DaoException {
...
         }
     }
+    
+    /**
+     * Deletes all records from DB using DELETE statement.
+     * 
+     * @param typeOfCancerId
+     * @throws DaoException
+     */
+    public static void deleteAllTypesOfCancer() throws DaoException {
```
jjgao: Where is this method called?

Can we use truncate table instead of loop?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
pieterlukasse: for the unit tests. Different unit tests use different set of genes. Some Import*.java classes could have scenarios where a specific gene is missing, or specific aliases are added or removed, etc

pieterlukasse: This will also make it possible to update your set of genes via an import script (like `ImportGeneData.java` - although this class needs to be improved) and just call recache instead of having to restart the webserver.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoTypeOfCancer.java
pieterlukasse: No...truncate is *exactly* what we are trying to get rid of.

jjgao: why?

pieterlukasse: Tests calling a TRUNCATE table SQL execution, which cannot be rolled back, affect other tests depending on the order of execution of the tests. 

deleteAllTypesOfCancer() is the DELETE alternative to the deleteAllRecords() method above. I also made deleteAllRecords() deprecated so we can remove this. If we want to keep any integrity in this DB, we should not be doing TRUNCATE table anywhere in the code in my opinion. That is why I am so eager in deprecating all these methods. 

Now the real question that remains is: why would we have such "delete all" methods in the Dao layer? That is a good question... Maybe we should remove this all together an implement such tricks in the application layers that need it. I think it is also an interesting discussion to have in the context of the new API/framework @ersinciftci is working on.

PR #1308, opened 2016-06-02
Travis use docker for end-to-end tests and mysql

inodb:
Use docker on Travis, so one can spin up a selenium grid for the visual
regression tests (end-to-end tests). Docker and mysql services cannot be used
in combination on Travis, so the core tests now uses a mysql docker container,
instead of the mysql db running on the host. Maven is still downloaded and run
on the host. An attempt was made to use a maven container instead, but volume
sharing with write permissions does not work (i.e. `docker run -v` does not
give write permissions) and copying over all the data to the container is very
slow. Another problem is that the docker version on Travis is older so 
`docker network` does not work. In short:

- Use docker service on Travis
- Use mysql container
- Use 127.0.0.1 (tcp/ip) instead of localhost (sockets) (necessary for docker
  mysql container connection)
- Make end-to-end visual regression test work again (those uses docker to
  create selenium grid)
- Use oraclejdk7 instead of openjdk7 to prevent buffer overflows to appear
  during testing (https://github.com/travis-ci/travis-ci/issues/5227)

*Proposed reviewers:*
@pieterlukasse @ersinciftci @fedde-s


pieterlukasse:
hi @inodb , this looks good to me :+1: 

I only have one question regarding the mysql: you mention you moved this into a docker container. I can see the docker part is now added. But how where was the previous mysql configuration? Or is mysql a default service in travis? Or was that just line 25 in the old `travis.yml`? Just trying to understand how the setup worked before.

ersinciftci:
@inodb It looks good to me

inodb:
Thanks for reviewing so quickly @pieterlukasse @ersinciftci, much appreciated!

@pieterlukasse Travis has the mysql service enabled by default. That is how it worked before. It is a known issue that enabling docker service removes mysql: https://github.com/travis-ci/travis-ci/issues/4870.

PR #1309, opened 2016-06-03
adding python tests to travis CI

pieterlukasse:
Next to adding the python tests to the travis script, some other important changes were needed in the `travis.yml` file:
- `sudo: required` : needed for the python libraries installation step
- `jdk: oraclejdk7` : needed to circumvent an existing issue with openjdk7 (also mentioned in #1308)


.travis.yml
```diff
@@ -18,6 +18,10 @@ install:
             mkdir -p ~/maven/$MAVEN_VERSION && \
             wget -O - http://www.eu.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz | \
                 tar -C ~/maven/$MAVEN_VERSION --strip-components=1 -xzf -; }
+    - sudo apt-get update && sudo apt-get -y install python-pip
+    - sudo pip install Jinja2
+    - sudo pip install MySQL-python
+    - sudo pip install requests
```
inodb: Maybe move the dependencies to a `requirements.txt` and set version numbers explicitly: https://github.com/inodb/revmut/blob/master/requirements.txt. Then install with: `pip install -r requirements.txt`. I doubt any of these packages will have many backwards incompatible changes, but it's good to be explicit anyway.

.travis.yml
```diff
@@ -26,6 +30,11 @@ before_script:
 
 script:
     - export PORTAL_HOME=$(pwd)
+    - export PYTHONPATH=$PORTAL_HOME/core/src/main/scripts
+    - cd $PORTAL_HOME/core/src/test/scripts/
+    - $PORTAL_HOME/core/src/test/scripts/./unit_tests_validate_data.py
+    - $PORTAL_HOME/core/src/test/scripts/./system_tests_validate_data.py
+    - cd $PORTAL_HOME
```
inodb: As mentioned, this is good for now, but if it's possible to use a test runner such as nosetests w/o too many changes to the code, I think that would be cleaner. Also, a comment here would be great.

inodb:
@pieterlukasse Thanks for adding these tests to the CI. Made some minor comments

pieterlukasse:
@inodb thanks for the comments. I made the changes in a new commit. But somehow the installation of the requests package via the `requirements.txt` gives a different result than using the previous `pip` command directly....I haven't quite figured out why. Maybe you can take a look at it?

Here is a build where the `requests` package installed successfully: https://travis-ci.org/thehyve/cbioportal/builds/134762391

pieterlukasse:
@inodb : I will try once more with the apparently existing 2.2.1 requests library on the server...(so removing it from the `requirements.txt` file)

inodb:
Pretty odd. Haven't seen that before. I'm guessing smth with the version numbers being off. Maybe Travis has some default things installed for Jinja and mysql as well?

I'm not sure if it is worth spending a lot of time debugging this at this stage. Maybe better to just use whatever you had before and was working.

Another option would be to use an Anaconda environment to get around the issue of having Travis default packages installed: https://github.com/inodb/revmut/blob/master/.travis.yml#L11-L30. Worked fine for me in the past. Though I can see how using both Dockerfiles and anaconda to create some form of containerization can be confusing, so perhaps better to not venture there.

pieterlukasse:
hi @inodb : after a lot of trial and error (as you can see here) I finally nailed down the issue and found a solution. Because you had added `services: docker` now the environment changed from ubuntu 12 to 14. This also had some implications for the python environment, which caused the failures. So I had to figure out a solution anyway.  As a solution, I ended up extending PYTHONPATH with hard links to the library folders. 

Furthermore, I implemented the changes you requested (using `requirements.txt` and adding documentation).

inodb:
@pieterlukasse ooh nice find! That looks like a pain to debug. I feel like we're spending a lot of times these days fixing test config environments. We should perhaps start considering either using some cloud service where you can SSH to the test machine (circle ci) or perhaps set up a local CI system.

In any case, looks great to me! Can be merged after squashing the commits.

pieterlukasse:
@inodb : trial and error commits are now squashed. 

Setting up a local CI system could be interesting when making big changes...now that I know travis will sometimes change the system depending on other flags I will pay more attention. We could also just fix it at ubuntu 14 to avoid such surprises (I saw some documentation with `dist: trusty`, maybe this does the job).

PR #1325, opened 2016-06-07
updated to the latest MutationMapper release v1.1.0

onursumer:
# What? Why?
This is a fix for https://github.com/cBioPortal/mutation-mapper/issues/37, https://github.com/cBioPortal/mutation-mapper/issues/43, and https://github.com/cBioPortal/mutation-mapper/issues/44

- Updates the MutationMapper library to the latest release: https://github.com/cBioPortal/mutation-mapper/releases/tag/1.1.0

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers


pieterlukasse:
@jjgao can we merge this one soon? I scanned the 508 related changes and these seem OK.  The other changes seem fine as well (like minor refactoring/fixes and color scheme changes).
I tested the mutations view table header qtips fix and it works fine now, except for the "Annotation" column which shows no qtip. @onursumer maybe you can fix this in a new PR? Then we could merge this one, which is already an improvement. :+1:

portal/src/main/webapp/js/lib/mutationMapper.js
```diff
@@ -1573,18 +1574,26 @@ var PileupUtil = (function()
...
+				count: groupCountMap[group],
+				priority: mutationTypeMap[group].priority});
 		});
 
 		groupArray.sort(function(a, b) {
-			// descending sort
-			return b.count - a.count;
+			if (b.count === a.count) {
+				// tie condition: use mutation type priority
+				return b.priority - a.priority;
```
pieterlukasse: perhaps double check: is `.priority` *always* set?

onursumer:
@pieterlukasse Annotation column was actually _plugged in_ by the portal code base, it is not a part of the MutationMapper library. The initial decision was to have no tool tip for this column.

@zhx828 can you suggest a basic description for the annotation column tool tip?

zhx828:
@pieterlukasse We intended to do that without tooltip in the annotation column. But that was back when we only have very few annotation sources. @jjgao maybe it's time to have a tooltip now? We may have mycancergenome, hotspot, OncoKB, MolecularMatch and Civic in the future, and all will be squeezed into annotation column.

portal/src/main/webapp/js/lib/mutationMapper.js
onursumer: We are setting the priority a few lines above: https://github.com/cBioPortal/cbioportal/pull/1325/files#diff-e64faf74f1b5d548c7cb04ee4bd3aaa2R1585

And the source of this priority is a static map: https://github.com/cBioPortal/mutation-mapper/blob/master/src/js/util/MutationViewsUtil.js#L43

Currently, I don't see a case where the priority can be undefined. But it is always possible that sometime in the future a new entry without a priority value into this static map can break the code :)

portal/src/main/webapp/js/lib/mutationMapper.js
```diff
@@ -1559,12 +1560,12 @@ var PileupUtil = (function()
```
pieterlukasse: you forgot to update the comment 2 lines above

portal/src/main/webapp/js/lib/mutationMapper.js
pieterlukasse: Ok, seems correct :+1:

portal/src/main/webapp/js/lib/mutationMapper.js
onursumer: good catch!

PR #1327, opened 2016-06-07
Rearchitecture Spring 2 - Replace calls to DAOMutation with MutationRepository

ersinciftci:


n1zea144:
A few random comments

- As previously discussed, In Sprint 3 we should move remaining mutations related calls in ApiService.java (like getMutationCounts*) into MutationService.
- When do we want to delete the deprecated methods in DaoMutation?

core/src/main/java/org/mskcc/cbio/portal/dao/GeneticAlterationUtil.java
```diff
@@ -47,9 +51,19 @@
...
  * @author Ethan Cerami.
  */
+@Component
 public class GeneticAlterationUtil {
     private static final String NAN = "NaN";
 
+    private static MutationRepository mutationRepository;
+    private static MutationModelConverter mutationModelConverter;
+
+    @Autowired
```
n1zea144: How will this work when classes not managed by Spring rely on this class (like GetAlterationDataJSON, GetProfileDataJSON, GetGeneticProfileJSON, GetProfileData)?

sheridancbio:
These are comments copied over for posterity from related PR #1274 :

I've read through these changes. Comments:

I'm a little confused about the changes to GeneticAlterationUtil. The class is now annotated @Component which (to my limited spring knowledge) causes the implicit creation of a spring Bean which can then be used for application configuration. Inside the class is an Autowired annotated constructor. I am guessing this is how to give arguments to help with Bean instance construction. But the constructor sets two static properties which are used in member functions. Wouldn't this mean that each time the Bean is instantiated that it is possible these static members would be changing? Perhaps we know that this bean is only ever instantiated once? Something seems strange with this construction -- do we need a safety to only initialize the static members if they are uninitialized? (null)

MutationModelConverter class is very clear. I do wonder why this class is an @component? What circumstances do we imagine wanting to configure in a different Converter implementation? Maybe during testing? Otherwise, couldn't this just be a plain class which is instantiated normally? 

SpringBeanAutowiringSupport.processInjectionBasedOnServletContext used in MutationsJSON, NetworkServlet,... is a capability I did not know about ... glad to learn this. I guess this is how to get spring to resolve Autowired members in classes which are not spring annotated such as @Controller or @Service  (such as in "non-spring" servlet classes)

Replacements of functionality in other classes follows a consistent approach, using the converter as needed.

Adding an application-context for core is necessary for Autowiring the @Components? Is this needed in conjunction with the processInjectionBasedOnServletContext() calls? I guess the component-scan finds the @Components, and the AutowiringSupport wires them using the beans created from the component-scan? (I'm still digesting spring, obviously)

A Sample has a SampleType member now (an Enum rather than a String) : This is better for detecting and characterizing errors sooner.

The multi-mode call getMutations(4 args) has been renamed getMutationsDetailed(4 args) in MutationsRepository. Other calls have been added to support the other Dao functionality needed.

Count classes added to transfer summary objects from the repository / DBMS.

Tests added with .sql support/util scripts. I'm not familiar with hsqldb .. I guess we are using memory-tables for testing. Some older tests are adjusted. Repositories are updated with mybatis support for functionality of new clients. MutationMapper.xml is getting fairly long.

Looks good

core/src/main/java/org/mskcc/cbio/portal/dao/GeneticAlterationUtil.java
ersinciftci: Spring will autowire the static fields during the application startup, so the relying classes don't have to be managed by Spring.

ersinciftci:
@n1zea144 
1- getMutationsCounts is newly introduced, so yeah we will move it into MutationService.
2- After this change is tested, I think. We can do it in Sprint 3.
@sheridancbio 
1- GeneticAlterationUtil is never instantiated in the code because all the methods are static.
2- Because we shouldn't create any class with new keyword except value objects, we should leave the creation of classes to the dependency injection framework, which is Spring core in this case. From wikipedia: "Dependency injection can be applied as a discipline, one that asks that all objects separate construction and behavior. Relying on a DI framework to perform construction can lead to forbidding the use of the new keyword, or, less strictly, only allow direct construction of value objects."
3- It's special to classes extending from HttpServlet.
4- Only for autowiring the @Components. Not needed for processInjectionBasedOnServletContext().

PR #1328, opened 2016-06-07
Fixed bug with importing fusion data correctly.

angelicaochoa:
GeneticProfileReader was not returning the genetic profile meta data loaded from the meta_fusions file. Instead it overwrote the fusions genetic profile with the existing MAF genetic profile, thus overwriting the datatype from 'FUSION' to 'MAF' and causing ImportProfileData to run ImportExtendedMutationData instead of ImportFusionData.


angelicaochoa:
@n1zea144  can you review/merge my fusion import fix?

fedde-s:
Really, ‘datatype’ (the type of file to be parsed) should just not be part of the genetic profile object, as it pertains to the file and how to parse it, and not so much to the actual profile represented by the data. Might it be better to just overwrite the datatype field, and leave all other attributes of the genetic profile object true to the one with this id that will actually be loaded in the database?

Issue #1324, opened 2016-06-07
coexpression tab not working in firefox


jjgao:
somehow it is related to oncoprint code:

![image](https://cloud.githubusercontent.com/assets/840895/15860211/2a428478-2c96-11e6-94a5-51018a9151ba.png)

@yichaoS @adamabeshouse could you look into that?

schultzn:
It seems to affect the Enrichments tab too. Tried Firefox 46.0.1 and 47.0

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1337

Issue #1318, opened 2016-06-06
oncoprint: adding clinical tracks not working


jjgao:
![image](https://cloud.githubusercontent.com/assets/840895/15825618/85dcd104-2bd2-11e6-8366-d959333bbd18.png)

jjgao:
Sometimes, it returns but very slow. The dropdown box is not always working either.

adamabeshouse:
@jjgao Looks like it's working again, but it is still very slow. I'm not sure what could speed the SQL query up, since it looks like the only link from study to clinical attributes is through several joins, including through the clinical data.

jjgao:
@adamabeshouse but I thought the old API was not as slow.

One option: can we add a cache `Map<Study, Set<Attribute>>` to the new implementation?

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1337

Issue #1311, opened 2016-06-03
Use public portal for mutation count endpoint, after next release


adamabeshouse:

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1337

Issue #503, opened 2015-11-17
add sub-menu for each track in oncoprint


jjgao:
For clinical track:
  * Sort A-Z
  * Sort Z-A
  * No sorting
  * Move up
  * Move down
  * Delete

For gene track:
  * Move up
  * Move down
  * Delete

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1337

PR #1337, opened 2016-06-08
A few oncoprint improvements

adamabeshouse:
1) Split up clinical attribute calls on backend to replace costly inner joins and thus speed up the oncoprint clinical track load (#1318)
2) Add 'move up' and 'move down' options to all tracks and add 'remove' option to gene tracks (#503)
3) Fix bug where clinical track would still be in the URL even after removing it
4) Fix oncoprint error in firefox when switching tab (#1324)
5) Use public portal for mutation counts (#1311)


jjgao:
@adamabeshouse could you resolve the conflicts?

PR #1343, opened 2016-06-09
fix to ensure statements are executed in correct order

pieterlukasse:
Fix for DB migration script (SQL statement were not always executed in same order as found in file). 

- Using OrderedDict should solve the problem

@zheins : can you review and test this?


zheins:
@pieterlukasse looks good to me!

PR #1355, opened 2016-06-10
Fix for timeline validation

pieterlukasse:
# What? Why?
Fixes timeline validation problem #1227  .

This is the same as #1230, but now to hotfix.

### DIFF
--- a/core/src/main/scripts/importer/cbioportal_common.py
+++ b/core/src/main/scripts/importer/cbioportal_common.py
@@ -176,6 +176,7 @@ META_FIELD_MAP = {
     MetaFileTypes.TIMELINE: {
         'cancer_study_identifier': True,
         'genetic_alteration_type': True,
+        'datatype': True,
         'data_filename': True
     },
     MetaFileTypes.CASE_LIST: {
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1833,6 +1833,7 @@ class TimelineValidator(Validator):
         'STOP_DATE',
         'EVENT_TYPE']
     REQUIRE_COLUMN_ORDER = True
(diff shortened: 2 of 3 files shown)


jjgao:
@zheins could you review this PR?

zheins:
@pieterlukasse: Just curious, what is the reason/benefit of switching the `genetic_alteration_type` to `CLINICAL`? It shouldn't affect anything going into the database, as only the `cancer_study_identifier` is used in importing from the `ImportTimelineData.java` script.

pieterlukasse:
@zheins : in `cbioportal_common.py` we have the following block:

```python
def get_meta_file_type(metaDictionary, logger, filename):
    """
     Returns one of the metatypes found in MetaFileTypes
    """
    # GENETIC_ALTERATION_TYPE    DATATYPE    meta
    alt_type_datatype_to_meta = {
        # cancer type
        ("CANCER_TYPE", "CANCER_TYPE"): MetaFileTypes.CANCER_TYPE,
        # clinical and timeline
        ("CLINICAL", "PATIENT_ATTRIBUTES"): MetaFileTypes.PATIENT_ATTRIBUTES,
        ("CLINICAL", "SAMPLE_ATTRIBUTES"): MetaFileTypes.SAMPLE_ATTRIBUTES,
        ("CLINICAL", "TIMELINE"): MetaFileTypes.TIMELINE,
```
which is used to define the file type. So the documentation is now following this implementation. We tried to group some `DATATYPE`s by giving them the same `GENETIC_ALTERATION_TYPE`. 

The reasoning is that TIMELINE contains CLINICAL type of data. Do you think we should put it back to `TIMELINE`?

zheins:
I see. That makes sense - it is at its core clinical data. This looks good to me then, I see no problem with changing it.

PR #1356, opened 2016-06-10
Revert 508 compliance changes for the mutation table

onursumer:
# What? Why?
Latest 508 compliance code renders majority of the mutation table columns empty. This is a quick fix to get those columns back. Looks like we need to solve the 508 compliance problem in a different way. 

This partially reverts the PR #1325

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@jjgao @pieterlukasse @fedde-s

### DIFF
--- a/portal/src/main/webapp/js/lib/mutationMapper.js
+++ b/portal/src/main/webapp/js/lib/mutationMapper.js
@@ -14790,11 +14790,6 @@ function MutationDetailsTable(options, gene, mutationUtil, dataProxies, dataMana
 					vars.mutationTypeClass = mutationType.style;
 					vars.mutationTypeText = mutationType.text;
 
-					// 508 compliance to prevent empty label tag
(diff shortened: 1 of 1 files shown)


jjgao:
@pieterlukasse @fedde-s I had to merge this quickly to fix the issue in the released code. Please address 508 compliance with a different approach.

onursumer:
Maybe we should just replace `label`s with `span`s in the mutation mapper template. In that case we don't need to worry about empty `label` tags.

fedde-s:
Ah, I believe those are the `label` tags in the table, below the mutation type legend of the lollipop diagram? I’d ported these diffs from e05e2013afe1eb0064c7663fe936e68233b5247d without help from the original author. But making them `span`s sounds good, I don’t see any control elements for which they’re labels anyway.

PR #1359, opened 2016-06-13
Fix for issue where mvn picks the wrong java version when none is specified

pieterlukasse:
# What? Why?
Apparently the java version used by Maven for the compilation of the code depends on the local environment where the build is running. This fix forces maven to use a *specific* version, making the build process more stable. 

The issue was introduced by the new sub-modules recently added: `persistence` and `service`.

- force maven to use `java 1.7`

@ersinciftci : can you review?

### DIFF
--- a/business/pom.xml
+++ b/business/pom.xml
@@ -26,16 +26,6 @@
 
   <build>
     <plugins>
-      <!-- java 1.7 support -->
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <version>2.0.2</version>
-        <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
-        </configuration>
-      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -282,16 +282,6 @@
 
   <build>
     <plugins>
-      <!-- java 1.7 support -->
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <version>2.0.2</version>
-        <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
-        </configuration>
-      </plugin>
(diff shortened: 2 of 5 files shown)


pieterlukasse:
Btw, this is the error I get on my DEV environment when the java 1.7 directives are not used:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.0.2:compile
 (default-compile) on project persistence-api: Compilation failure
[ERROR] /cbioportal/persistence/persistence-
api/src/main/java/org/cbioportal/persistence/
MutationRepository.java:[9,8] error: generics are not supported in -source 1.3
```

ersinciftci:
Should we add it to `model` module too? It's new as well.

morungos:
I'd be tempted to do this in `pluginManagement` in the primary `pom.xml`, unless there's a good reason for it to be different in subcomponents. We can always override. 

I'm actually thinking we should be using the `dependencyManagement` for component versions a lot more, too.

pieterlukasse:
@morungos can you give an example? It would be nice to have it only once indeed.

morungos:
It'd be almost identical, but the the main `pom.xml`: see: http://stackoverflow.com/a/10483432/2140998. I don't think you need the plugin as well (although you normally might) because it's brought in for compile in Java projects anyway. All this should do with my hazy understanding of Maven strangeness, is get the plugin configuration and version inherited across all subprojects, which is better than defining it in every one. 

Having said that, in this case, it might only be some subprojects that need Java 7+

```
<build>
    <pluginManagement>
       <plugins>
           <!-- build with java 1.7 -->
           <plugin>
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-compiler-plugin</artifactId>
               <version>2.0.2</version>
               <configuration>
                   <source>1.7</source>
                   <target>1.7</target>
               </configuration>
            </plugin>
        </plugins>
    </pluginManagement>
</build>
```

pieterlukasse:
@morungos thanks for the tip. Actually I added it to the main pom.xml and it worked :) even without the  `<pluginManagement>` tag. I will push this up instead.

pieterlukasse:
@morungos I took the liberty of also removing existing `<plugin>` entries in other older `pom.xml` files. Seems to work.

PR #1362, opened 2016-06-14
Adding missing LDAPUserDetailsContextMapper class

holtgrewe:
Adding missing class `LDAPUserDetailsContextMapper`.

Hopefully, the error handling in the constructor is OK now as well.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/authentication/ldap/LDAPUserDetailsContextMapper.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/authentication/ldap/LDAPUserDetailsContextMapper.java
@@ -1,0 +1,217 @@
+package org.mskcc.cbio.portal.authentication.ldap;
+
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.SortedSet;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.mskcc.cbio.portal.authentication.PortalUserDetails;
+import org.mskcc.cbio.portal.dao.PortalUserDAO;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/authentication/ldap/LDAPUserDetailsContextMapper.java
```diff
@@ -0,0 +1,217 @@
...
+        String givenName = null;
+        if (user.getAttributeSortedStringSet(givenNameAttribute) != null)
+            givenName = (String) user.getAttributeSortedStringSet(givenNameAttribute).first();
+        String lastName = null;
+        if (user.getAttributeSortedStringSet(givenNameAttribute) != null)
+            lastName = (String) user.getAttributeSortedStringSet(givenNameAttribute).first();
+        String email;
+        if (user.getAttributeSortedStringSet(givenNameAttribute) == null) {
+            LOG.warn("Retrieving email failed, using username " + username + " for everything");
+            email = username;
```
pieterlukasse: this will not work well with how the authorisation is currently documented (see https://github.com/cBioPortal/cbioportal/blob/master/docs/User-Authorization.md). The email should be there. 

When I mentioned that better exception handling would be nice, I meant only giving a clear exception message instead of NullPointerException. There is no need to implement such extra fallback scenarios (probably not desirable actually, as it complicates documentation).

holtgrewe: OK hopefully fixed this

pieterlukasse: yes, looks better.

Issue #1088, opened 2016-04-21
DB settings confusing


pieterlukasse:
The documentation in https://github.com/cBioPortal/cbioportal/wiki/portal.properties-Reference#database-settings is a bit confusing as it seems that some db properties are only used by the data loading scripts and not by the portal code (which takes config from spring?).

holtgrewe:
I have spent the last two days setting up cBioPortal for private deployment on Flynn (roughly equivalent to Flynn).

This would be much easier if cBioPortal adhered to the [Twelve Factor App Principles](http://12factor.net/), that is giving configuration through environment variables. I have an internal patch that makes GlobalProperties consider the environment variables for each queried property. This works well for the contact email address, for example.

However, in cBioPortal, the connection information appears to be compiled in, so I have to recompile the app everything I change the database settings (which is easy and thus can happen frequently in a Heroku-like deployment).

For me, a "golden path" for database configuration would be to read a database URL environment variable, (cmp. [dj-database-url](https://github.com/kennethreitz/dj-database-url)) by default, and maybe fallback to a portal.properties using this setting. Nothing would be baked in during compilation ;)

pieterlukasse:
I agree with @holtgrewe , just not sure what are the best follow-up steps. @inodb, @morungos  : any ideas? Maybe you guys are already thinking about changing some of this?

holtgrewe:
Another place where the data source appears is in [applicationContext-business.xml](https://github.com/cBioPortal/cbioportal/blob/master/business/src/main/resources/applicationContext-business.xml#L73-82).

Also, this is the place where I attempted to resolve the the mysql timeout issue. This brought me over the timeout of 10 minutes but I'm not certain that the portal will continue to work overnight.

```
@@ -77,10 +77,12 @@
             <property name="password" value="${db.password}"/>
             <property name="minIdle" value="0"/>
             <property name="maxIdle" value="10"/>
             <property name="maxActive" value="100"/>
             <property name="poolPreparedStatements" value="true"/>
+            <property name="testOnBorrow" value="true" />
+            <property name="validationQuery" value="SELECT 1" />
         </bean>
 		<bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
 	    	<property name="staticMethod" value="org.mskcc.cbio.portal.dao.JdbcUtil.setDataSource"/>
 	    	<property name="arguments">
 	        	<list>
```

holtgrewe:
[...]

PR #1364, opened 2016-06-14
Stabilizing Maven packaging and deployment for overlays

### LINKED ISSUE #1299: Remove Maven variables from all configuration of built components
We've had a long-running issue with the way Maven variables are baked into the war file. This means, to all intents and purposes, each user is compiling and building their own version of cBioPortal -- there is no standardization possible. 

There are good historical reasons for this. At the time, it was the easiest way to customize. But newer techniques have emerged, such as [war overlays](http://maven.apache.org/plugins/maven-war-plugin/overlays.html) that make it possible to override local values without affecting the integrity of a standard core component. 

I'd propose: removing, as far as humanly possible, all Maven variable templating for the web application except where it is used for code factoring and/or testing. Putting DB settings into test files is fine, but core should not depend on them. Instead, Java properties and/or environment variables should inject key settings. JNDI is another option. Any approach that moves settings to runtime should be fair game. 

As a consequence, Maven should no longer depend on `PORTAL_HOME`. 

Directly related issues:

 * https://github.com/cBioPortal/cbioportal/issues/938
 * https://github.com/cBioPortal/cbioportal/issues/1088
 * https://github.com/cBioPortal/cbioportal/issues/1169

morungos:
# What? Why?

Change the way we use Maven so that a standard core deployment will run with mvn compile test package, no need for `PORTAL_HOME`, and no baking in of package values. A stub `portal.properties` is packaged in, but can be replaced in a war overlay and/or at runtime. Maven is also now configured to use a `local.properties` (at top level) for Maven variables to override DB settings, etc. 

Fix #1299, #1088, #938 (a bit).

- Removing dependencies on `PORTAL_HOME` from build time
- Adds a new `local.properties` Maven property source, used for DB settings, etc.
- Removes the war filename fiddles, making it much easier to deploy and overlay
- Changed the Maven DB settings key to be something other than the vanilla default
- Runtime properties are now drawn from (a) classpath `default.properties`, classpath `portal.properties`, and `$PORTALHOME/portal.properties`, later values taking precedence, enabling both runtime configuration and easy overlays
- Wired the DB version into Java code, because it shouldn't really be in a properties file, especially one that can be modified at runtime. 
- Added a range of Maven DB test settings so as not to conflict with the live ones
# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.
# Any screenshots or GIFs?

# Notify reviewers

everybody in the core team:
@cBioPortal/core-developers


src/main/resources/default.properties
```diff
@@ -0,0 +1,165 @@
+# app name
```
inodb: Is this file to replace `src/main/resources/portal.properties.EXAMPLE`? I feel that's sort of our default config now.

morungos: That was my idea. Have an explicit `defaults.properties` which is used first. Overwrite that with anything in a classpath `portal.properties`, which folks can bake in as a war overlay. And finally, overwrite that with anything from some external property source. Spring should follow a similar pattern, but I haven't done that yet. In effect, this means the `.EXAMPLE` file can be actually used as fallbacks rather than copied and edited.

inodb: Moving `.EXAMPLE` to `defaults` sounds good to me. Maybe also update the PR to show that.

I'm not too familiar with war overlays. Is there any documentation you can recommend on why that would be better than using environment variables?

portal/pom.xml
```diff
@@ -45,7 +45,6 @@
 
   <build>
     <!-- final name of the app -->
-    <finalName>${final-war-name}</finalName>
```
inodb: Any idea why this was used in the first place?

pom.xml
```diff
@@ -474,7 +479,30 @@
         <artifactId>maven-release-plugin</artifactId>
         <version>2.5.3</version>
       </plugin>
+
+	  <!-- For all subprojects, read a top-level local.properties file for configuration 
+	       information. This can override the properties needed to, e.g., connect to a DB. -->
```
inodb: So is this `local.properties` parsed at runtime? Or is it more to override some of the things in `default.properties`, the advantage being that it is more clear exactly what is unique to the local configuration.

src/main/resources/default.properties
pieterlukasse: @morungos why bake in the file? Why not just refer to it via something like a runtime parameter?

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -713,12 +713,7 @@ public static boolean getRecacheStudyAfterUpdate() {
     }
     
     public static String getDbVersion() {
-        String version = properties.getProperty(DB_VERSION);
-        if (version == null)
-        {
-            return "0";
-        }
-        return version;
+        return DATABASE_VERSION;
```
inodb: :+1:

inodb:
:+1: Really great attempt at sorting all these issues. Thank you! Looks like we're getting somewhere.

For me it's a little hard to fully follow how some of this configuration would work. It would be great if you could update the `docs/` in the same PR so I can get a better idea of what you have in mind.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -52,8 +52,11 @@
  */
 public class GlobalProperties {
 
+	private static final String DATABASE_VERSION = "1.2.1";
```
inodb: Def better than setting as property value, but this is also more something like EXPECTED_DB_VERSION. There should be code elsewhere that decides what to do when the database version does not match up with the expected one.

pom.xml
morungos: Definitely not. It's only for Maven. It is templated into test files (to set DB connection strings), but should not be templated into anything else. It's one of the places we'll want to make it heroku-happy though, so if you can think of a better way to set DB values, make me happy.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
morungos: This is the other part of that. Currently the system reads the version from the DB, and another version from some runtime properties file, and checks they are the same. Which is pointless, as nothing is connected to the code at all. The intent here was to propose moving one of the versions into source, so it's versioned. This then can be checked against the version read from the DB, which is actually done in `DaoInfo`. It's horrible, I admit, because the name here is misleading. Yes, let's change it as you suggest, as it's not too widely referenced.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
morungos: Agreed entirely.

pieterlukasse: I agree it should be named `EXPECTED_DATABASE_VERSION`.

src/main/resources/default.properties
morungos: I'm thinking mainly of security and authentication here. A common use case is to use, say, SAML or OpenID Connect rather than a default setup. There are a many properties here, and Spring Security is the source for the configuration right now. So yes, we could switch to environment variables, but that isn't the current source for Spring configuration. This allows us to drop in a different Spring config for one deployment, without that being the core one supplied. Essentially we'd be taking a vanilla war file from a Maven central repo, and replacing a few files for a local deployment. 

Given where we are now, with Spring Security, that's much easier than rewriting the whole of the current security configuration system to use environment variables.

morungos:
So the core failures here are Travis core MySQL connection errors, which would normally be handled by a portal.properties. There are, therefore, some deployment changes that will be needed alongside this.

portal/pom.xml
n1zea144: There are historical reasons for this, but now, all the maven profiles with exception of heroku can be removed from this pom.

src/main/resources/default.properties
n1zea144: One thing I mentioned in [1299](https://github.com/cBioPortal/cbioportal/issues/1299) that we need to support at MSK currently and perhaps other institutes in the future is running multiple instances of the cBioPortal concurrently.  This could get hairy if we rely on environment variables.

core/src/test/resources/applicationContext-dao.xml
```diff
@@ -9,18 +9,6 @@
...
-	    <property name="ignoreResourceNotFound" value="true" />
-	    <property name="locations">
-		     <list>
-			     <value>file:///#{systemEnvironment['PORTAL_HOME']}/src/main/resources/portal.properties</value>
-		     </list>
-	    </property>
-     </bean>
-     
-
 	<bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
```
n1zea144: Is this duplicated on line 57.

src/main/resources/default.properties
morungos: Agreed. We're moving the same way.

jjgao:

PR #1293, opened 2016-05-26
minor bug fixes and improvements for mutation mapper

onursumer:
See https://github.com/cBioPortal/mutation-mapper/issues/39 for details

PR #1366, opened 2016-06-15
Read SWISSPROT in MAF as accession if specified in meta file

fedde-s:
Based on the feedback on pull request #1202, I restored the default behaviour of reading the SWISSPROT column of mutation data files as the name/id, and added a flag in the corresponding metadata file to optionally specify that it is instead to be read as an accession (which is what VEP and VCF2MAF list in this column when not using their GRCh37-specific cache). Both the dataset validation script and the data loading code have been adjusted, and unit tests are included for both scenarios.

Summarised, this set of patches does the following:
- determine what type of identifier should be validated in the `SWISSPROT` column of MAF files based on the new optional metadata flag `swissprot_identifier`, and validate the format of these identifiers
- adjust automated validator tests that involve MAF files with a SWISSPROT column, and add automated tests for the new scenarios supported
- allow the loader code to parse optional flags in metadata files
- make the mutation data loader parse the SWISSPROT column based on the new metadatum
- add automated tests for the new scenarios supported by the loader
- deprecate a setter method for a now-unnecessary field in the data model
- document the new metadata flag

## Suggested reviewers ##
- @pieterlukasse
- @n1zea144


core/src/main/scripts/importer/validateData.py
```diff
@@ -965,6 +972,12 @@ def checkHeader(self, cols):
                 'that the Uniprot canonical isoform is used when drawing Pfam '
                 'domains in the mutations view',
                 extra={'line_number': self.line_number})
+        elif not 'swissprot_identifier' in self.meta_dict:
+            self.logger.warning(
+                "A SWISSPROT column was found in a file without an "
+                "associated 'swissprot_identifier' metadatum, assuming "
+                "'swissprot_identifier: name'.",
```
pieterlukasse: :information_source:  discussing with @jjgao and @n1zea144 if we still want the default to be *name* instead of *accession* now that they have also upgraded their internal VEP installation.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1179,18 +1192,56 @@ def checkNotBlank(self, value, data):
...
+            # test whether the id is known to the portal, if available
+            if (self.portal.uniprotkb_entry_map is not None and
+                    value not in self.portal.uniprotkb_name_set):
+                self.logger.warning(
+                    'This UniprotKB/Swiss-Prot name is not known to the '
+                    'portal, the portal will attempt to display the protein '
+                    'based on the gene',
+                    extra={'line_number': self.line_number,
+                           'cause': value})
+        # if no reasons to return with a message were found, return valid
```
pieterlukasse: :+1: I like this whole block.

core/src/main/scripts/importer/validateData.py
pieterlukasse: thinking about it, I think we can simplify both the code and the behaviour if we make this meta field mandatory (in the validator). This is probably the best option now that we are in a "transition phase" between old and new VEP installations...having a default value will probably just result in unintended behaviour for one of the two cases (either the people having old VEP or the ones having new VEP - assuming the difference is caused by VEP versios). Having to set a meta property is clear for both cases. 

For the loader, the default can stay. I.e. when circumventing the validation step you can expect to have some sub-optimal results.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1179,18 +1192,56 @@ def checkNotBlank(self, value, data):
...
+                self.extra = 'SWISSPROT value is not a UniprotKB accession'
+                self.extra_exists = True
+                return False
+            # test whether the accession is known to the portal, if available
+            if (self.portal.uniprotkb_entry_map is not None and
+                    value not in self.portal.uniprotkb_entry_map):
+                self.logger.warning(
+                    'This UniprotKB/Swiss-Prot accession is not known to the '
+                    'portal, the portal will attempt to display the protein '
+                    'based on the gene',
```
pieterlukasse: strictly speaking, this is not the case anymore. Since some fixes ago, the portal actually looks up the accession directly from the `pfam_graphics` table (see `PfamSequenceServlet.java`), which is not the table that is used to build up the `uniprotkb_entry_map` here. 

A webservice that selects the list of uniprot accessions from `pfam_graphics` table would suffice for the accession scenario.

pieterlukasse:
Looks good to me. Thoroughly done, with unit tests :+1: 

I added a few comments. The last one is especially important for one of our other projects that will probably rely on the uniprot mapping table being empty as they would like to just use ensembl.

jjgao:
OK... I am a little surprised to see such a big pull request to just support SWISSPROT accession.

I think we also need to discuss whether it is necessary/appropriate to put it into the public web API. Could you explain how the api is used?

fedde-s:
The main reason the pull request is so large is that it also adds validation for the SWISSPROT column, and *all* existing validator unit tests that involve MAF files have been adjusted so their SWISSPROT column does not make them fail.

The API is used so that the validator can request which accessions and names are known to the portal, and inform the user if the ones in their file will not be recognised by the portal.

fedde-s:
@pieterlukasse: As I read your comment https://github.com/cBioPortal/cbioportal/pull/1366#issuecomment-231202668 , these are the scenarios for the value of the column mutation_event.oncotator_uniprot_accession column:
- the importer will load accessions specified in the data file directly, and the frontend then may or may not find it in the `pfam_graphics` table
- the importer will attempt to map names specified in the data file to accessions using the `uniprot_kb_mapping` table (as of this pull request), and either leave it empty or proceed as above
- if the column is left empty (either because nothing was specified in the data file or because the loader failed to map a name to an accession), the frontend will attempt to map the gene to an accession using the `uniprot_kb_mapping` table, and proceed as above

And if the frontend does find a pfam_graphics entry by these means, the mutations will simply be visualised on a bar 1/3 as long as the gene, as if there are no introns or domains.
Is this correct?

pieterlukasse:
hi @jjgao , thanks for your comments. The API point you mention is a good one. At the start, we chose to use the API endpoints to retrieve data as this was easy and the services were already available. It has the added advantages that: 
- the user does not need to configure DB settings or use a properties file when using the python script
- the user can choose different endpoints to validate against. So if he is planning to submit his data to a different server, he can just give the url to that server as a parameter.
- we reuse the existing APIs instead of implementing the same queries again using python DB libraries

The disadvantage is that you cannot use it *yet* when the webserver is configured with some authentication types like saml. Another point to consider is that we might end up adding some web-services that are used only in the validation for now. As long as they are set up as generic web-services that could be useful in other parts of the application, I don't see a problem. Another disadvantage is that you need tomcat to be running while in fact all you want to do is interact with the DB (this could change however if we want to do "real time" data loading in the future). Let me know what you think.

pieterlukasse:
@fedde-s as just discussed, here is the summary of the changes and simplifications:

- [ ] validate only accession (using pfam_graphics accesion list)
- [ ] remove uniprot_mapping related web-services; add new one for pfam_graphics
- [ ] name or empty is a warning
- [ ] keep regex checks for name AND accession
- [ ] data loading code remains the same
- [ ] unit tests can be simplified (for the name scenario)

pieterlukasse:
@fedde-s as just discussed, the summary above can be further reduced to the following:

- [ ] keep regex format checks for name AND accession
- [ ] remove uniprot_mapping related web-services
- [ ] empty is a warning
- [x] data loading code remains the same
- [ ] unit tests can be simplified

fedde-s:
@pieterlukasse, @jjgao: I simplified the whole series of changes as discussed above, by removing all validation of whether names and accessions are known to the portal’s database, only validating whether they do have the format of names or accessions. This means no changes to the web service are necessary, and fewer of the validator’s unit tests have to be adjusted.

fedde-s:
One moment—even though Travis gives a green light, the tests for the validator seem to be failing.

fedde-s:
Fixed.

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -727,15 +728,101 @@ def test_warning_for_missing_SWISSPROT(self):
...
+        self.assertNotIn('portal', record.getMessage().lower())
+        # neither a name nor an accession
+        record = record_iterator.next()
+        self.assertEqual(record.levelno, logging.ERROR)
+        self.assertEqual(record.line_number, 5)
+        self.assertEqual(record.cause, 'A1CF_HUMAN,HBB_YEAST')
+        self.assertNotIn('portal', record.getMessage().lower())
+
+    def test_implicit_name_as_swissprot_identifier(self):
+        """Test if the SWISSPROT column is parsed as a name if unspecified."""
```
pieterlukasse: you could just reuse checks of previous method here

pieterlukasse:
Looks good :+1:  Only 1 minor comment on a unit test.

fedde-s:
Amended the commit to DRY a unit test as Pieter suggested.

jjgao:
@zheins would you please test the code with our pipeline? Not only the pipeline should run, but also we need to verify the data in correctly show up on the page -- e.g. mutation diagram & structure data.

pieterlukasse:
@jjgao @zheins thanks for testing. For your information, the most important fixes to make this work in the frontend were actually recently made in #1293 and #1296

zheins:
@fedde-s @jjgao, I have issues testing this PR that are unrelated to these changes because hotfix appears to be broken.

`Exception in thread "main" org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'apiService'`

`Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [org.cbioportal.service.MutationService] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}`

jjgao:
@zheins something is wrong.. the service module has not been merged to hotfix yet.

zheins:
You're right, git added these files to my repo and I didn't remove them when I switched branches.

zheins:
@fedde-s @jjgao I get this error attempting to import our clinical dataset on the second maf we import. The first maf imports fine:

`Caused by: java.lang.NullPointerException
	at org.mskcc.cbio.portal.model.GeneticProfile.getOtherMetaDataField(GeneticProfile.java:158)
	at org.mskcc.cbio.portal.scripts.ImportProfileData.run(ImportProfileData.java:81)
	... 10 more`

fedde-s:
@zheins Thanks! The branch was last updated with hotfix a before the weekend, but I’ll still rebase to resolve a conflict introduced since then.

zheins:
@fedde-s So the problem is that when multiple MAFs are imported in a single study, `GeneticProfileReader.loadGeneticProfile( descriptorFile );` does not get called for the second file because the genetic profile already exists in the database. This means `setOtherMetaDataField` does not get called, leaving a null for `otherMetaData` in the `GeneticProfile` instance. We need to set this field for the `existingGeneticProfile` in the `GeneticProfileReader` class if one is found.

fedde-s:
@zheins: We should probably add a unit test that imports a mutation data file in a study that already has one, if that’s a supported scenario. Then I can fix your issue by making the test succeed.

zheins:
It should be a supported scenario, we already are doing this in some of our internal clinical studies and in larger projects in which memory issues are encountered if the MAF file is too large.

fedde-s:
@zheins It should be fixed now.

fedde-s:
Sorry, forgot a null check elsewhere. *Now* it should be fixed.

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -84,7 +84,20 @@ public static GeneticProfile loadGeneticProfile(File file) throws IOException, D
...
+                     existingGeneticProfile.getCancerStudyId(),
+                     existingGeneticProfile.getGeneticAlterationType(),
+                     existingGeneticProfile.getDatatype(),
+                     existingGeneticProfile.getProfileName(),
+                     existingGeneticProfile.getProfileDescription(),
+                     existingGeneticProfile.showProfileInAnalysisTab());
+             gp.setGeneticProfileId(existingGeneticProfile.getGeneticProfileId());
+             gp.setTargetLine(geneticProfile.getTargetLine());
+             gp.setOtherMetadataFields(geneticProfile.getAllOtherMetadataFields());
+             return gp;
```
pieterlukasse: so why not return the `existingGeneticProfile` ?

fedde-s: `existingGeneticProfile` is part of a static dictionary in `DaoGeneticProfile`, which forms a naive cache of the database. I think that fields specific to the data loading phase of a particular data file in this profile, which are not in the database, do not belong in the cache either. That is why I make a new object to combine the profile information from the cache with the file-specific information from the new meta file.

fedde-s: Especially since this code is specifically for loading a second data file into the same mutation data profile, splitting the data over multiple files.

fedde-s: Would it be better if I made a new constructor `GeneticProfile(GeneticProfile)` instead, to copy the object? That way this whole list of attributes would stay inside the class.

pieterlukasse: @fedde-s thanks for the explanation. I missed this small detail, but indeed you have 
```
gp.setTargetLine(geneticProfile.getTargetLine());
gp.setOtherMetadataFields(geneticProfile.getAllOtherMetadataFields());
```
which is different from the `existingGeneticProfile`. So ok, this looks good to me :+1: 

@zheins any other comments? If not, could you please ask @jjgao to merge this one?

fedde-s: @pieterlukasse I simplified this code to avoid this confusion in the future.

zheins:
@fedde-s looks good to me. @jjgao this can be merged.

PR #1374, opened 2016-06-15
For heroku profile, copy log4j.properties.EXAMPLE

inodb:
# What? Why?
Heroku review app log shows there is a problem with not being able to locate the log4j.properties file. This copies the example file to the appropriate location when the heroku profile is activated. This makes heroku work again. Only problem is that we still get a few `Error R14 (Memory quota exceeded) Process running mem=593M(114.7%)`. It's mostly the first time you load the page. There's not much to do about that other than either

1. improving memory efficiency
2. paying for bigger dynos (as the workers are called in heroku) or
3. Using a smaller test database

- copy `log4j.properties.EXAMPLE` to `src/main/resources/` when `heroku` profile is active

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@jjgao

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
               </execution>
             </executions>
           </plugin>
-          <!-- copy portal.properties.EXAMPLE -->
+          <!-- copy portal.properties.EXAMPLE and log4j.properties.EXAMPLE -->
           <plugin>
             <artifactId>maven-antrun-plugin</artifactId>
             <!-- only needs to be executed for parent pom -->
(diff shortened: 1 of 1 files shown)


morungos:
I'm looking into this as part of #1364 too. My gut feeling is that there should be a builtin simple log4j.properties, with overlays and/or externalization to override. Anything else means that each user is customizing their build, which is a problem.

inodb:
Ah great! I'll have a look. This is a pretty trivial fix just to get heroku running again. It would be nice if we could get the heroku profile out of the pom completely and just set some environment variables. Let's discuss more on #1364

morungos:
I've usually had to do profile magic to get any kind of CI right, so I wouldn't worry, but I'm aiming in #1364 to cut out the dependency on too much build customization. One option would be a new final submodule which overlays on anything needed for heroku. 

The other profiles, e.g., `ctd2-dashboard` and `genie` in the main `pom.xml` are more worrying. I really can't see them being necessary. But :-) and any input into #1364 or other Maven improvements very happy to collaborate on.

inodb:
Cool. Yeah I have no clue if any of those profiles are even still being used, maybe @n1zea144 or @zheins knows

jjgao:
@inodb could you do this on hotfix?

inodb:
@jjgao opened a new PR to hotfix

PR #1386, opened 2016-06-16
Rearchitecture sprint 2

ersinciftci:
# What? Why?
Rearchitecture sprint 2, replace DaoMutation calls with MyBatis

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.


persistence/persistence-api/src/main/java/org/cbioportal/persistence/dto/SignificantlyMutatedGene.java
```diff
@@ -0,0 +1,34 @@
+package org.cbioportal.persistence.dto;
+
+import java.io.Serializable;
+
+public class SignificantlyMutatedGene implements Serializable {
```
ecerami: Class javadocs...?

ersinciftci: I'm not sure if every class requires class javadoc, as long as package name, class name and property names are meaningful, javadoc explanation is a bit duplicate. It has persistence.dto in package name, which means it's a data transfer object, and the name is SignificantlyMutatedGene, so it means it represents a significantly mutated gene and carries the data from persistence layer. My javadoc comment would be something like /* A class that represents a significantly mutated gene */ but that would be a duplicate information. And for code author information in javadoc, it's better to use git history for that purpose because it's more accurate. Of course some methods and classes requires javadoc, like the one I created here https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcUtil.java#L249, because it's not obvious from the method name and it should be used carefully. So I'll review the new classes and add javadoc to the ones that needs more explanation, in a later PR.

ersinciftci:
This change is already reviewed by @n1zea144 and @sheridancbio in a separate PR to rearchitecture-sprint-2 branch, so it's ready to merge.

PR #1389, opened 2016-06-17
Darwin patient view updates

angelicaochoa:
What? Why?
- Added try/catch for darwin patient view properties in `GlobalProperties` in case properties aren't included in the build. 
- Checked in `darwin_logo.png`
- Replaced darwin text with darwin logo in `patient_view.jsp` 

- a
- b

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers

… NullPointExceptions

2. Added logo for darwin
3. Modified patient_view.jsp to replace darwin text with darwin logo if user has access to patient data


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -722,35 +723,47 @@ public static String getDbVersion() {
...
     public static String getDarwinAuthCheckUrl() {
-        String darwinAuthUrl = properties.getProperty(DARWIN_AUTH_URL);
-        if (darwinAuthUrl == null || darwinAuthUrl.isEmpty()) {
-            return "";
+        String darwinAuthUrl;
+        try{
+            darwinAuthUrl = properties.getProperty(DARWIN_AUTH_URL).trim();            
         }
-        return darwinAuthUrl.trim();
+        catch (NullPointerException e){
```
jjgao: Maybe simplify the code as below -- you can also try to catch any exceptions. Similar for other methods.

```
String darwinAuthUrl = "";
try {
    darwinAuthUrl = properties.getProperty(DARWIN_AUTH_URL).trim(); 
} catch (Exception e) {
}

return darwinAuthUrl;
```

pieterlukasse: Actually I introduced `org.mskcc.cbio.portal.scripts.TrimmedProperties.java` before for this reason. It is used in a number of places already. Please use that.

pieterlukasse: Another remark: it is confusing that `DARWIN_AUTH_URL` means something else here when compared to the same constant in `CheckDarwinAccessMain.java`. I would suggest removing the constant in `CheckDarwinAccessMain.java` and just using this getter method.

angelicaochoa:
thanks! I'll update in a moment @jjgao

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -723,46 +723,42 @@ public static String getDbVersion() {
...
+        
         return darwinAuthority;
     }
     
     public static String getCisUser() {
-        String cisUser;
+        String cisUser = "";
         try{
             cisUser = properties.getProperty(CIS_USER).trim();
         }
```
pieterlukasse: @angelicaochoa maybe you can consider using TrimmedProperties.java class (is a subclass or Properties) to simplify this code?

angelicaochoa: @pieterlukasse do you just mean removing the .trim() at the end or using `TrimmedProperties` as a means to replace the try/catch as well?

angelicaochoa: @pieterlukasse I don't think it would really make sense to be loading the same properties file twice in `GlobalProperties`. 

I guess to avoid code redundancy, if this is the approach we ultimately want to take, `Properties` can be replaced with `TrimmedProperties` when loading the properties file within `GlobalProperties` but that's probably a discussion for another time with the rest of the group, or at least with @jjgao and @n1zea144. 

For now though I'll just use .trim() unless the decision to switch classes is made.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
angelicaochoa: @pieterlukasse constants are modified in `CheckDarwinAccessMain` now

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
n1zea144: I agree with Angelica.  I'd rather have a focused PR to globally replace GlobalProperties with TrimmedProperties.

PR #1392, opened 2016-06-17
Patient view page enhancement

angelicaochoa:
# What? Why?
Function `guessClinicalData` only performs a check on whether the data type is undefined or if the data is null, but doesn't handle other "null values", such as 
'NA', 'unknown', 'undefined', etc. Functions that call `guessClinicalData` then have to add checks for these other "null values". 

I added a function to help standardize how clinical data are determined to be null, empty, NA, etc. for the **displayed** clinical data **ONLY** (not the dropdown tables).

**BEFORE**
before

**AFTER**
after

- a
- b

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers


inodb:
I've actually been working on moving the clinical attribute styling from JS to CSS: https://github.com/cBioPortal/cbioportal/pull/1015. I think for N/A NA we should add something to the `clean()` function in ClinicalAttributesUtil: https://github.com/cBioPortal/cbioportal/pull/1015/files#diff-8504785b87a4049db3deb7b4978ce641R12 which removes such values

angelicaochoa:
@inodb awesome! Definitely should add NA, N/A, and variants of those that we run into

inodb:
Cool! I will try to add this function to PR #1015 then. Let me know if you have other suggestions. What patient is this? Should be a good test case

pieterlukasse:
@angelicaochoa FYI : the list of "standardized null" values can also be found at the enum `ImportClinicalData.MissingAttributeValues` (ImportClinicalData.java) and at `validateData.py` (class `ClinicalValidator`). If you find any new "null  values", please update these places as well.

angelicaochoa:
no problem @pieterlukasse ! I'll get to those later today

PR #1396, opened 2016-06-21
Rearch sprint 1 add json unit tests

sheridancbio:
What? Why?

Fix # .
Adding unit tests / integration tests to check ApiService and JSON generation for Web Api

adding Tests classes in web module
update servlet-api to more recent version to allow use of MockMvc
Checks

 Runs on Heroku.
 Single commit and No merge commits
 Make sure your commit messages end with a Signed-off-by string (this line can be automatically added by git if you run the git-commit command with the -s option)
 If this is a feature, the PR is to rc. If this is a bug fix, the PR is to hotfix.
Any screenshots or GIFs?

here with e.g. GifGrabber.

Notify reviewers

git blame <filename> to determine that and notify them here. Otherwise notify
everybody in the core team:
@n1zea144 @morungos @ersinciftci @jjgao @ecerami


core/src/test/java/org/mskcc/cbio/portal/util/NullHttpServletRequest.java
```diff
@@ -36,16 +36,26 @@
...
+import javax.servlet.DispatcherType;
+import javax.servlet.AsyncContext;
 import javax.servlet.http.Cookie;
+import javax.servlet.http.Part;
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
+import javax.servlet.http.HttpUpgradeHandler;
 
 /**
```
n1zea144: Should this be in servlet package?

morungos: I think it's probably okay in this package, in that it's a stub for testing to inject parameters, not a real class. More of an issue is whether it should exist at all, in that EasyMock or equivalent might be better.

sheridancbio:
tests completed and branch has been rebased into a single commit

PR #1398, opened 2016-06-22
Small validation fixes

pieterlukasse:
# What? Why?
- relaxed validation for quotes (" ). E.g. quotes can occur in some MAF fields, in some of the TCGA studies. This is now a warning instead of an error
- improved the error message when a clinical attribute value should be NUMBER or BOOLEAN and is not.
- fixed and relaxed NCBI_Build validation in MAF files. Allowed values are now: hg19, 37, GRCh
- added test files for scenario where quotes are present in files
- support for NA in clinical data

Notify reviewers:
@n1zea144, @zheins


core/src/main/scripts/importer/validateData.py
```diff
@@ -1195,12 +1205,15 @@ class ClinicalValidator(Validator):
 
     REQUIRE_COLUMN_ORDER = False
     PROP_IS_PATIENT_ATTRIBUTE = None
-    NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]","[completed]","[null]", ""]
+    NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]","[completed]","[null]", "na"]
```
pieterlukasse: "" now missing...need to fix this

PR #1399, opened 2016-06-22
Data validation & data loading integration tests

pieterlukasse:
# What? Why?
- Added TestIntegrationTest.java Integration test using the same data that is used by the `python` validator system test "study_es_0". In the `python` validation test for "study_es_0" it is checked if the study can pass validation without any errors or warnings. Here we submit the same study to the `java` data loading code and check that it also loads correctly and completely into the DB. **This should help keep the data validation and data loading code in sync.**
   - For this, a number of "study_es_0" files were adjusted to simplify this scenario. A number of `.json` files containing basis DB data needed for the tests were also adjusted.

While implementing such a broad test, I took the liberty of also making small fixes or improving comments in related parts:
- Added some extra documentation to ImportClinicalData
- Made a small fix in ImportCopyNumberSegmentData to avoid unnecessary warning messages
- Improved output message of GisticReader
- Fixed documentation link in validation report jinja template
- Improved documentation in demoScript.xml (TransactionalScripts test scenario file)

Proposed reviewers:
@sheridancbio, @ersinciftci


sheridancbio:
Beginning review now.

sheridancbio:
Trying to run unit tests ... set up my normal portal.properties, but having db connection failure. Looking..

sheridancbio:
Ready to merge .. please squash some of these commits if you are able.

sheridancbio:
JJ suggested that it may make sense to separate the validator logic out of the core portal repository (maybe creating a new repository). That might also make development and maintenance easier / faster / more convenient. What do you think @pieterlukasse ?

pieterlukasse:
hi @sheridancbio , thanks for the review. I squashed the commits into topical commits. I made one small extra change in `.gitignore` file to get rid of some irrelevant files that kept appearing in `git status`.

pieterlukasse:
@sheridancbio @jjgao : regarding the idea to move the validation code outside of core: I like the idea. In fact, I think we should also move the data loading code as well. Maybe we can start a new sub-project called `data-loading`? Let me know. We could then move the code there in a next PR.

PR #1402, opened 2016-06-22
Add max_retry for screenshots test

inodb:
# What? Why?
Due to difference in load on Travis, the screenshot can sometimes fail. This
retries taking the screenshot with increasing number of waiting time.  All the
parameters can be set in `screenshots.yml`. Another option would be tor wait for some HTML element to appear instead of doing a sleep, but I haven't been able to find such an element on all pages. It also makes writing the tests harder. Therefore the retry option was chosen.

For example see:

https://travis-ci.org/cBioPortal/cbioportal/jobs/139524886#L1814-L1817

- Add a retry parameter to `screenshots.yml`. On failure the screenshot is taken again with increased timout. The new timeout value is `nr_retry * timout`, thereby increasing the timeout on each retry.

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@fedde-s @ersinciftci 

### DIFF
--- a/test/end-to-end/test_make_screenshots.sh
+++ b/test/end-to-end/test_make_screenshots.sh
@@ -65,19 +65,31 @@ for ((i=0; i < ${#config_screenshot_names[@]}; i++)); do
     name=${config_screenshot_names[$i]}
     url="${config_portal_url}${config_screenshot_urls[$i]}"
 	timeout_in_seconds=${config_screenshot_timeout[$i]}
+    max_retries=${config_screenshot_retry[$i]}
 
     for browser in "${config_browsers[@]}"; do
         png="screenshots/${browser}/${name}.png"
         echo -e "Checking ${DIR}/${png} at $url..."
-        compare_image $config_python_selenium_container $config_selenium_hub_url \
-					  $browser $url $png $timeout_in_seconds
-        # check exit code
-        if [[ $? -gt 0 ]]; then
+        retry=0
+        while [[ $retry -lt $max_retries ]]; do
+            compare_image $config_python_selenium_container $config_selenium_hub_url \
+                          $browser $url $png $timeout_in_seconds
+            compare_exit_code=$?
+            # break if exit code 0
+            if [[ $compare_exit_code -eq 0 ]]; then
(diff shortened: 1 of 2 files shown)


test/end-to-end/test_make_screenshots.sh
```diff
@@ -65,19 +65,31 @@ for ((i=0; i < ${#config_screenshot_names[@]}; i++)); do
...
 
     for browser in "${config_browsers[@]}"; do
         png="screenshots/${browser}/${name}.png"
         echo -e "Checking ${DIR}/${png} at $url..."
-        compare_image $config_python_selenium_container $config_selenium_hub_url \
-					  $browser $url $png $timeout_in_seconds
-        # check exit code
-        if [[ $? -gt 0 ]]; then
+        retry=0
+        while [[ $retries -lt $max_retries ]]; do
```
fedde-s: Should this be `$retry`?

inodb: Yup oops

inodb:
@fedde-s Should work now. I tested with a failing screenshot where it indeed tries multiple times: https://travis-ci.org/cBioPortal/cbioportal/jobs/139524886#L1814-L1817

fedde-s:
Good! And I don’t know a lot about Travis CI configuration and best practices, but the patch looks fine to me, if others agree to the increasing timeouts (after repeated failing tries only) of up to 40 seconds for the patient view and up to 30 for the study view, if I read it correctly.

fedde-s:
I don’t know what the limit is on how much we can run on Travis, but could it potentially increase pending times for other branches and pull requests when a pull request truly fails the test?

It may also slightly increase the time until a contributor can download the screenshot and see what the issue is, but only by a bit, and only in the case of true positives (not unrelated minor fixups/conflict resolutions).

inodb:
I think the build limit is 120 minutes so it should be fine.

Yeah it would increase the time for a true positive. I guess you can still run the tests locally and change the retry in that case, if you want it to be fast.

PR #1410, opened 2016-06-24
replace max value with calculated outlier limit in study view bar chart

JiaoJiao123:
In some study view bar charts, the max value is a very big outlier. And that would make most of the values in the bar chart distributed only in very few bars.
In this pull request, instead of using the direct max value, we use the max value calculated from the box plot quantiles, which will make the values distributed more evenly in the bar chart.


JiaoJiao123:
@zhx828 Could you help with the code review? Since it's based on your work earlier.

JiaoJiao123:
@zhx828  I will add log feature to very small number later since it is not closely related to this pull request.  could you review the code and let me know any other problems? Thank you!

PR #1414, opened 2016-06-27
Allow tabs to be disabled on the visualize.jsp page

zheins:
# What? Why?
Allow tabs to be disabled for performance or other reasons.

# Notify reviewers
@jjgao @zhx828


jjgao:
@zheins could you point to hotfix branch instead so that we can release soon?

PR #1416, opened 2016-06-28
Clean up data handling in datamanager

adamabeshouse:
# What? Why?
Split out OQL filtering from data manager and clean up data handling/return formats.
And some other cleanup that incidentally got looped in on this branch as I went through

- Move all matters OQL to a global object called `OQL`. Filtering is now done on the webservice data level (i.e. on the event level), rather than on the sample-gene datum level, and it turns out much cleaner that way.
- Clean up code in QuerySession and update how data is returned to enable OQL queries which have the same gene but with different parameters in different tracks.
- Update existing code, including mutex, enrichments plots, data download tab, and pancancer study summary, to use the new interface.
- Replace all uses of old data managers PortalDataColl and PortalDataCollManager with the new system.
 
Other cleanup:
- Delete old oncoprint files
- Some small style changes.

@jjgao @yichaoS @pieterlukasse


jjgao:
@yichaoS @kalletlak  would you please review this pull request? Thanks!

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1,1029 +1,982 @@
...
+    var objectKeyIntersection = function(list_of_objs) {
+	var intersection = {};
+	for (var i = 0; i < list_of_objs.length; i++) {
+	    if (i === 0) {
+		var keys = Object.keys(list_of_objs[0]);
+		for (var j = 0; j < keys.length; j++) {
+		    intersection[keys[j]] = true;
+		}
+	    } else {
+		var obj = list_of_objs[1];
```
kalletlak: Is it going to be list_of_objs[1] or list_of_objs[i] ?

adamabeshouse: Ah, good catch!

pieterlukasse:
Hi @adamabeshouse, what do you mean with "Filtering is now done on the webservice data level (i.e. on the event level), rather than on the sample-gene datum level" ? Can you explain a bit more and maybe point to some examples in the code?

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1,1029 +1,982 @@
...
-	};
-	var uniqStrings = function(array_of_strings) {
-	    var seen = {};
-	    var uniq = [];
-	    for (var i=0; i<array_of_strings.length; i++) {
-		var str = array_of_strings[i];
-		if (!seen[str]) {
-		    uniq.push(str);
-		    seen[str] = true;
+	    ],
```
pieterlukasse: why are the query parameters already filled in here?

portal/src/main/webapp/js/src/oql/oqlfilter.js
```diff
@@ -0,0 +1,413 @@
...
+	ret += gene;
+	if (alterations.length > 0) {
+	    ret += ": " + alterations.map(parsedOQLAlterationToSourceOQL).join(" ");
+	}
+	ret += ";";
+	return ret;
+    };
+    
+    /* For the methods isDatumWantedByOQL, ..., the accessors argument is as follows:
+	     * accessors = {
```
pieterlukasse: out of curiosity: why this construction? Why not let each datum have these methods? If this for performance reasons?

portal/src/main/webapp/js/api/cbioportal-datamanager.js
adamabeshouse: Very sorry, this accidentally got into this pull request in pre-finished form (can be ignored, is completed in a different branch). You're right though. This is the only thing that is unfinished in this pull request.

portal/src/main/webapp/js/src/oql/oqlfilter.js
adamabeshouse: Do you mean the conversion back to source OQL? That's for annotating in the portal with the source OQL, for example if there are multiple instances of the same gene in the query and you want to know which track is which set of alterations.

As far as datum methods, the approach I've taken isn't object-oriented.

Does that answer your question?

adamabeshouse:
Hi @pieterlukasse, thanks for asking for the clarification. What I mean by that is that whereas previously we would combine webservice data into sample-gene, "oncoprint" data (then from there into patient-gene data) of the form  {'sample': 'TCGA-0001', 'gene':'BRCA1', 'mut':'MISSENSE', 'cna':'AMPLIFICATION', ... }, and then run that through OQL to filter it, in the new approach, we run webservice data (i.e. single genetic alteration event data) through the OQL filter and then combine those genetic alteration events that make it through into sample-gene and patient-gene data. This gives a more modular layered approach, where you filter the webservice data and then any application that wants to use OQL-filtered data can start with all the information instead of the lossy and oncoprint-application-specific oncoprint format.

PR #1422, opened 2016-06-29
Adding correct UQ constraint to gene table

### LINKED ISSUE #1266: Add missing UQ constraint to `gene` table
Add the constraint as discussed in #805 (see https://github.com/cBioPortal/cbioportal/pull/805#issuecomment-179274725 and https://github.com/cBioPortal/cbioportal/pull/805#issuecomment-180355753)

pieterlukasse:
# What? Why?
Fixes #1266, adding a unique constraint to the gene table.

- gene table gets a unique constraint to ensure we don't get wrong data
- fix for the validation of the db version. Somehow the previous implementation was failing to detect version mismatches. The new implementation works and just redirects to the error page, simplifying the overall implementation of this check.

:warning: if the table gene already contains duplicated entries, the migration script will fail. The script contains some references to documentation on how to fix it in this case. This could be an argument to merge this in `rc` instead of `hotfix`.

# Notify reviewers
@n1zea144 , @zheins


Sjoerd-van-Hagen:
@jjgao Could you merge this one? Thanks!

zheins:
This looks fine to me. This does change the behavior of the db check though, we intentionally had it not popping up the error message unless an actual error is encountered. Can we add that to this?

pieterlukasse:
@zheins not sure if I follow. Can you add more details about this error message you mentioned?

zheins:
Essentially this if statement in `index.jsp`:
`if (dbError != null && userMessage != null)`

The DB error message would only appear if there was some other error on the page as well. This way, even if your DB is out of date you could still use the portal and only have the warning if something actually goes wrong.

pieterlukasse:
@zheins @jjgao is this really what we want? Personally, I think that it will make the code less predictable if we agree on a versioning system that is *not* strictly enforced. Also harder to test and to troubleshoot. 

For example: let's say we can't assume the UQ constraint proposed in this PR exists on a given cBioPortal instance. This means we will have to program specific exceptions in java that still take into account the possibility of ambiguous hugo gene symbols. So we end up with code that is in most cases dead code, but in some cases (i.e. instances that did not keep up with migration.sql) is actually executed.

zheins:
@jjgao can you comment on this? I agree with with your reasoning @pieterlukasse, but I don't remember why that check was put in place.

jjgao:
@pieterlukasse The reason was because of the following scenario:

We have a beta portal running at MSK with the rc code, connecting to the same database as the production code. So there is only one version of the database. But when we change db scheme in the rc, an error message is always displayed on the first page which discourages our beta users from using the beta portal, even though the functionalities are all working. They complained to us a few times -- and I believe a few beta users have given up and decided to use the production protal instead. That's why we decide to display the message only when there is an error connecting to the database.

sheridancbio:
@jjgao @pieterlukasse I think the best solution would be to maintain actual database builds (across version) for all actively deployed portal instances. But if we think we know what we are doing and want to use an older database build with a portal .war file expecting a newer schema version, then perhaps we should introduce a variable such as db.version.suppress_mismatch_warning=true to portal.properties and use it with caution. Appending a note that "database version in use [#} differs from expected version (#)" to relevant error messages can also be done.

pieterlukasse:
@sheridancbio currently the DB check is only done at *data loading* time. The validation is more flexible and only calls some webservices without checking if the DB schema version is up to date. And I think that at the *data loading* step it is good to require the schema to be up to date...although I agree that not all schema changes are relevant for the data loading part. 

Looking at this PR and the recent changes that were made elsewhere to this part of the code, perhaps I should split this PR up. The part that I think is really important here is the UQ constraint that needs to be added to the `gene` table. This will ensure that the DB is not containing legacy erroneous and ambiguous hugo "symbols" (which were in fact not hugo symbols due to an older bug in ImportGeneData.java).

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -267,7 +267,13 @@ protected void doPost(HttpServletRequest httpServletRequest,
             xdebug.logMsg(this, "Got Database Exception:  " + e.getMessage());
             forwardToErrorPage(httpServletRequest, httpServletResponse,
                                DB_CONNECT_ERROR, xdebug);
-        } catch (ProtocolException e) {
+        } catch (DbVersionException e) {
+            String errorMessage = "Mismatch between DB version and portal version. " + e.getMessage();
+            xdebug.logMsg(this, errorMessage);
+            forwardToErrorPage(httpServletRequest, httpServletResponse,
```
fedde-s: Wouldn't this mean that [this documentation](https://github.com/cBioPortal/cbioportal/blob/master/docs/Updating-your-cBioPortal-installation.md#first-time) needs to be updated?

pieterlukasse: yep

pieterlukasse:
I moved the banner part out to #1806

pieterlukasse:
I moved the UQ constraint part to #1807
@sheridancbio :point_up_2:

Issue #1380, opened 2016-06-16
No easy-to-find documentation on configuring Study search options


pieterlukasse:
At one of our installations we got the request to change the list of studies displayed at the search box (see screenshot):

![image](https://cloud.githubusercontent.com/assets/2900303/16119271/0d21dc1c-33db-11e6-9968-671e88df9811.png)

Is this configurable? How to do it? If not, could we add a portal.properties for this, or let it be based on the list of studies in the portal (taking into account authorization settings for the user)?

fedde-s:
This has been solved in the hotfix release now.

PR #1423, opened 2016-06-29
Make study query suggestions configurable

fedde-s:
Allow administrators of cBioPortal installations to configure the example queries listed in the dropdown menu of the study search box, as requested in issue #1380.

screen shot 2016-06-29 at 16 18 57

- parse and document a portal property named `skin.example_study_queries`, a newline-separated list with the previously hard-coded values as the default value
- generate the list of query suggestions to filter studies in the query view based on the new portal property

# Suggested reviewers
@cBioPortal/core-developers

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/step1_json.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/step1_json.jsp
@@ -30,6 +30,9 @@
  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
 --%>
 
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ taglib prefix="fn" uri="http://java.sun.com/jsp/jstl/functions" %>
+
 <%
     String step1ErrorMsg = (String) request.getAttribute(QueryBuilder.STEP1_ERROR_MSG);
 %>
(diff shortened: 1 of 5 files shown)


fedde-s:
This can be useful in installations that do not contain the datasets used as examples by default. Administrators of such installations want to customise the default suggestions and replace them with ones that make sense for the data in the portal.

portal/src/main/webapp/WEB-INF/jsp/step1_json.jsp
```diff
@@ -44,16 +47,8 @@
...
-                    <li role="presentation"><a role="menuitem" tabindex="-1"  href='javascript:void(0)' onclick='$("#jstree_search_input").val("esophageal OR stomach");$("#jstree_search_input").trigger("input");'>esophageal OR stomach</a></li>
-                    <li role="presentation"><a role="menuitem" tabindex="-1"  href='javascript:void(0)' onclick='$("#jstree_search_input").val("serous");$("#jstree_search_input").trigger("input");'>serous</a></li>
-                    <li role="presentation"><a role="menuitem" tabindex="-1"  href='javascript:void(0)' onclick='$("#jstree_search_input").val("breast");$("#jstree_search_input").trigger("input");'>breast</a></li>
+                <ul class="dropdown-menu dropdown-menu-right" role="menu" title="Select from dropdown"><c:forEach var="query" items="${exampleStudyQueries}">
+                    <li role="presentation"><a role="menuitem" tabindex="-1" href='javascript:void(0)' onclick='$("#jstree_search_input").val("${fn:escapeXml(fn:replace(fn:replace(query, "\\", "\\\\"), "\"", "\\\""))}");$("#jstree_search_input").trigger("input");'><c:out value="${query}" /></a></li></c:forEach>
```
adamabeshouse: what's going on here with all the `replace`s?

adamabeshouse:
If it works then it looks good to me!

portal/src/main/webapp/WEB-INF/jsp/step1_json.jsp
fedde-s: > what's going on here with all the replaces?

Escaping of backslashes and double quotes in what is going to be a string
literal in JavaScript inside an html attribute. I should probably add a
comment to clarify that.

adamabeshouse: ahh, I see. A comment would be helpful, especially because it looks like there's a non-commutative situation with the replacement applications, a bit hard to reason about.

fedde-s: > it looks like there's a non-commutative situation with the replacement
applications, a bit hard to reason about.

There is--the backslashes with which quotes are escaped shouldn't be
escaped themselves. Thanks for pointing out that it is hard to read!

fedde-s:
As suggested in @adamabeshouse’s review, I added a comment to the JSP code. And it works perfectly in my tests, both with defaults and with a custom list in `portal.properties`.

fedde-s:
@jjgao can this be merged now?

jjgao:
@fedde-s just to make sure that you would like to merge it rc instead of hotfix?

fedde-s:
Ah, our client would be happy if it were merged to hotfix and released
sooner, so I could rebase the PR by tomorrow morning if that is okay. The
reason I pointed it to hotfix is that it I saw it as technically more of a
(very small) new feature than a bugfix.

fedde-s:
The reason I pointed it to rc*

jjgao:
Small incremental changes can also be merged to hotfix. Please rebase. Thanks.

fedde-s:
Replaced by #1499, which points to the hotfix branch instead.

PR #1424, opened 2016-06-29
Sample count per patient

JiaoJiao123:


JiaoJiao123:
@jjgao Could you check it out and let me know if any problem? Thank you!

PR #1428, opened 2016-06-29
Three new attributes

JiaoJiao123:
1: # of Sample Count Per Patient #813
2: Sequenced
3: Has CNA data


JiaoJiao123:
@jjgao Could you check it out when you get a chance? Thank you!

jjgao:
- [ ] change "Sequenced" to "With mutation data"
- [ ] change "Has CNA data" to "With CNA data"
- [ ] display this three charts by default
- [ ] hide any charts (not only this three charts) if there is only one value for all samples
- [ ] change "# of Sample Count Per Patient" to pie chart

Issue #852, opened 2016-02-17
study view: column plot not working with outliers


jjgao:
Maybe set the min/max (left/right) numbers to 5% and 95% percentiles.

![image](https://cloud.githubusercontent.com/assets/840895/13123207/32c57b50-d589-11e5-8ec3-2031f4ad36e2.png)

zhx828:
@jjgao but how do you want to select the outliers if you shorten the range?

jjgao:
@zhx828 the outliers will be in two groups (left and right). Please prioritize this.

jjgao:
@zhx828 you can use this one to detect ourliers instead of using 5% and 95%: https://onlinecourses.science.psu.edu/stat200/node/135

PR #1429, opened 2016-06-29
replace max value with calculated outlier limit in study view bar chart

JiaoJiao123:
In some study view bar charts, the max value is a very big outlier. And that would make most of the values in the bar chart distributed only in very few bars.
In this pull request, instead of using the direct max value, we use the max value calculated from the box plot quantiles, which will make the values distributed more evenly in the bar chart.


zhx828:
#1410

jjgao:
@JiaoJiao123 please try to use the pull request template next time.

@yichaoS @zhx828 could you please review?

jjgao:
fixed #852

PR #1434, opened 2016-06-30
Adding support for HTTPS in mutation pages 3D view

pieterlukasse:
# What? Why?
Workaround for limitations at **rcsb.org**'s regarding serving content over HTTPS. See also 
https://github.com/3dmol/3Dmol.js/issues/188

- introduced a variation of the already existing `api/getexternalpage.json` API to work as a proxy to retrieve PDB data when cBioPortal is setup as a HTTPS site 
- adjusted the respective *mutation view* pages to use this proxy service when `document.location.protocol == "https"` 

# Notify reviewers
@onursumer


jjgao:
@onursumer would you please review?

jjgao:
@pieterlukasse: it didn't work when testing on heroku:

image

Looks like `document.location.protocol` returns `https:`.

image

onursumer:
Looks like at some point I removed the proxy configuration. The previous configuration was using the server side proxy for that:

`pdbUri: "api/proxy/jsmol/"`

onursumer:
@pieterlukasse @jjgao : Using the jsmol proxy may fix it: https://github.com/cBioPortal/cbioportal/blob/7b9a4537ba056b7c7e7cd02a0e6be4284b5f8b0f/web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java#L172

pieterlukasse:
@jjgao thanks for catching this. I had done some manual tests to work around the fact that I don't have my dev on https and overlooked this typo. I will fix it. 

@onursumer thanks for pointing me to this code! I was unaware of this part. This is really nice. Now I can remove the custom `/api/getexternalpage.plain` service I had added. I think I will have to add `produces = "text/plain", ` to your service to make it work, but I'll test it first.

pieterlukasse:
@jjgao @onursumer : I fixed the typo and was able to use the existing api/proxy/jsmol without any code changes in the backend. So now only 3 js changes :)

pieterlukasse:
Tested this page on heroku and it works: 

image

pieterlukasse:
This one works as well: 

image

pieterlukasse:
@onursumer : is the third scenario related to the stand-alone mutation view?

onursumer:
@pieterlukasse: yes that third page is the standalone view.

Also, a tiny detail but I think the name of the proxy api is a bit confusing. Probably, we should have initially named it `api/proxy/pdb/` instead of `api/proxy/jsmol/`

pieterlukasse:
@onursumer ok, thanks. Do you want to change the api name in this PR?

pieterlukasse:
great news: rcsb.org is now serving with https as well. See e.g. https://files.rcsb.org/view/1hh3.pdb1 :tada: 

this means we can probably simplify and speed up this part

PR #1436, opened 2016-07-01
check case list based on filename instead of path

jochemb:
Case lists are checked for not being text editor temporary files before processing.  This check now works correctly.


fedde-s:
Indeed, that was a bug in importing case list files referenced by relative paths. Looks like a good fix! 👍

Issue #1314, opened 2016-06-06
Finish fix of #1013 for FUSION data


pieterlukasse:
Fix in #1013 needs to be applied to FUSION data importer (in ImportFusionData.java)

pieterlukasse:
this is solved by the validation script already. validateData.py:
```
elif entrez_as_int <= 0:
                self.logger.error(
                    'Entrez gene id is non-positive.',
```

Issue #1102, opened 2016-04-25
FUSION data is not retrieved by new data manager/oncoprint


pieterlukasse:
Fusion data is present in some studies, like prostate cancer studies (e.g. [see this query](http://www.cbioportal.org/index.do?cancer_study_list=prad_su2c_2015&cancer_study_id=prad_su2c_2015&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=prad_su2c_2015_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=prad_su2c_2015_gistic&genetic_profile_ids_PROFILE_MRNA_EXPRESSION=prad_su2c_2015_rna_seq_mrna_median_Zscores&Z_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=prad_su2c_2015_cnaseq&case_ids=&patient_case_select=sample&gene_set_choice=prostate-cancer%3A-ar-signaling-%2810-genes%29&gene_list=TMPRSS2+ERG+SLC45A3+ETV1&clinical_param_selection=null&tab_index=tab_visualize&Action=Submit))

The documentation for the _old_ and _new_ ways of loading data both fail to describe how to make correct fusion data files that will also successfully work in the oncoprint display. The problem seems to be due to the fact that the new data manager layer used by oncoprint expects the **fusion data** to be part of the **mutations** genetic profile. But making a meta file as described in the documentation makes a new separate genetic profile and loads all fusion data into this other genetic profile. 

Possible solutions:
- adjust loader code at `ImportProfileData` to exceptionally load Fusion data into the genetic profile of mutations. This means mutation data should also always be loaded first.
- OR adjust oncoprint/data manager layer to retrieve the data from the Fusion genetic profile. This can be a confusing change as the Fusion genetic profile will have to be "visible" in the query page. Also many other places in the application that retrieve mutation data will need to be adjusted to also consider the extra Fusion genetic profile...

pieterlukasse:
@jjgao, @n1zea144  : can you please comment on the possible solutions? I think the first one is better, and by far the easiest one. I just need to confirm whether the fusion data is indeed loaded into the mutations genetic profile. Can you confirm this?

pieterlukasse:
This issue seems to break the "Plots" tab as well, since apparently the `ImportFusionData` class does not add any samples to the database (but it looks like it expects to reuse the "mutations" profile instead?)

```
[...]

PR #1438, opened 2016-07-01
Update File-Formats.md

jjgao:
# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers

### DIFF
--- a/docs/File-Formats.md
+++ b/docs/File-Formats.md
@@ -667,7 +667,7 @@ The fusion metadata file should contain the following fields:
 1. **cancer_study_identifier**: same value as specified in [study meta file](#cancer-study)
 2. **genetic_alteration_type**: FUSION
 3. **datatype**: FUSION
-4. **stable_id**: fusion
+4. **stable_id**: mutations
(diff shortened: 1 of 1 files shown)


fedde-s:
@jjgao do the validator and cbioportalImporter.py recognise this datatype?

fedde-s:
Ah, no, since commit aee9eb0c3ad9722e6ff450de53dd9d25f7427403 they require the stable id of fusion data to be ‘fusions’. So this documentation change introduces a bug.

pieterlukasse:
@jjgao @fedde-s this will probably not work with the validator. But I will be fixing this soon as I am planning to pick up #1102 and #1314 next week. In the final solution, I think I would prefer the meta file to still point to `fusion`  instead of `mutations` so that the workaround keeps "hidden" in the code for now instead of being visible in the user facing meta files. This makes it also easier to change the implementation later on without having to change the documentation again.

PR #1458, opened 2016-07-06
metaImport needs a -n option

### LINKED ISSUE #1455: metaImport needs a -n option
Discussed with @pieterlukasse and thought about this some more last night.

I think the metaImport script needs a -n (--no_portal_checks) option.  This option already exists in validateData.py, we just need to pass it along.

That way a new user who is following the installation steps on the wiki can successfully import data without having a portal instance already running.  This is important because in the wiki docs, we suggest that people get the entire database set up first, and then deploy the web app.  Without a -n option, new installs are blocked from importing data.

@pieterlukasse suggested that we use cbioportalImporter.py, which provides this functionality.  But, I would much rather introduce the right scripts up-front, e.g. introduce others to the validator and metaImport, so that they can see how it works up-front.

Since @priti88 is out, I will endeavor to try to fix on my own today.  Otherwise, may call upon @ersinciftci.

ersinciftci:
# What? Why?
Fix #1455  

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
everybody in the core team:
@ecerami @pieterlukasse @fedde-s

### DIFF
--- a/core/src/main/scripts/importer/metaImport.py
+++ b/core/src/main/scripts/importer/metaImport.py
@@ -52,6 +52,10 @@ def interface():
                                    help='Path to a directory of cBioPortal '
                                         'info files to be used instead of '
                                         'contacting the web API')
+    portal_mode_group.add_argument('-n', '--no_portal_checks', default=False,
+                                       action='store_true',
+                                       help='Skip tests requiring information '
+                                            'from the cBioPortal installation')
     parser.add_argument('-jar', '--jar_path', type=str, required=False,
                         help='Path to core JAR file')
     parser.add_argument('-html', '--html_table', type=str,
@@ -74,7 +78,6 @@ if __name__ == '__main__':
     args = interface()
     # supply parameters that the validation script expects to have parsed
     args.error_file = False
-    args.no_portal_checks = False
 
     study_dir = args.study_directory
 


fedde-s:
I think adding this option is a bad idea as I mentioned in issue #1455, but if you really do want it, it should be mutually exclusive with `-u` and `-p` as it is in [validateData.py](https://github.com/cBioPortal/cbioportal/blob/rc/core/src/main/scripts/importer/validateData.py#L2579)

fedde-s:
As I mentioned in the issue, I left out the option to load data after only partial validation by design and find adding it a dangerous idea, *especially* if it’s for teaching people what is the right way to do it . If people insist on skipping validation (the old way), there’s still the `cbioportalImporter.py` script that’s used by `metaImport.py` in the background.

core/src/main/scripts/importer/metaImport.py
```diff
@@ -52,6 +52,10 @@ def interface():
                                    help='Path to a directory of cBioPortal '
                                         'info files to be used instead of '
                                         'contacting the web API')
+    portal_mode_group.add_argument('-n', '--no_portal_checks', default=False,
+                                       action='store_true',
+                                       help='Skip tests requiring information '
+                                            'from the cBioPortal installation')
```
pieterlukasse: I would change the description here to `hep='start the loading process with minimal validation, skipping tests that require information from the cBioPortal installation. Please note that the loading process is not guaranteed to succeed in this case'`

ecerami:
This will likely be a temporary fix for ~2-3 weeks until @priti88 is back. I have logged my preferred solution as #1466.

PR #1391, opened 2016-06-17
Fix image links and update validator output in docs

fedde-s:
Fix broken screenshots of the data validation report noticed in issue #1375, and update them along with the textual sample output of the same script to reflect how it currently looks.

### Suggested reviewers ###
@priti88 
@jjgao

Issue #1375, opened 2016-06-15
Fix images in validator documentation


priti88:
The images of HTML screenshots seems to have disappeared from validator document:
https://github.com/cBioPortal/cbioportal/blob/master/docs/Using-the-dataset-validator.md

fedde-s:
The images have been restored and updated as of v1.2.5, as implemented by the pull request that referenced this issue. @priti88, can this be closed?

pieterlukasse:
they even have a nice 3D effect to them now  👍 I'll close it. @priti88 feel free to reopen if you disagree.

PR #1462, opened 2016-07-07
Fix image links and update validator output in docs

fedde-s:
Fix broken screenshots of the data validation report noticed in issue #1375, and update them along with the textual sample output of the same script to reflect how it currently looks.

This pull request replaces #1391, but is directed straight to the master branch as this seems to be the current practice for documentation fixes.

### Suggested reviewers ###
@priti88 
@ecerami 
@n1zea144


fedde-s:
@inodb, do you know if there’s an easy way to see how this documentation update/fix renders on Readthedocs? I don’t expect issues, but I don’t know if Readthedocs handles language specifications in syntax highlighting (```` ```console````) and relative paths to images.

inodb:
Relative paths to images should be fine. You could check some of the existing examples. Not sure about syntax highlighting. We are using a converter in Sphinx to convert markdown to HTML, but it's def not as sophisticated as Shinx's native RST transformer.

One way to test is to sign up for a readthedocs account yourself and point it to this branch.

Faster & easier is to test locally by running the documentation yourself. If you have sphinx installed you should be able to run `make html` in `docs/`. For installation a `pip install sphinx` or `conda install sphinx` should prolly do the trick. Or use this documentation: http://www.sphinx-doc.org/en/stable/install.html.

PR #1464, opened 2016-07-07
Show docs on rtfd.org by using Sphinx to generate HTML

inodb:
# What? Why?
The documentation is pretty hard to navigate at the moment. By using sphinx to generate the HTML documentation from Markdown we can use readthedocs.org to display it. Number one feature is the search feature, see http://cbioportal-inodb.readthedocs.io/en/latest/.

readthedocs is basically a "documentation as a service". It clones your repo, runs sphinx and uploads the HTML result to your readthedocs page. You can choose to generate the docs for multiple branches and different versions in the readthedocs admin panel, so people can easily check the dev documentation or v1.2.2 documentation w/o needing to understand how to navigate github.

- Use Sphinx to generate HTML docs from Markdown
- Set configuration to work with readthedocs.org
- Remove `Home.md` cause I don't think it's used anymore
- Add titles to several Markdown pages

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?
http://cbioportal-inodb.readthedocs.io/en/latest/

# Notify reviewers
@pieterlukasse @ecerami


docs/Data-Loading.md
```diff
@@ -2,7 +2,7 @@
 ## Introduction
 This page is the starting point for data loading. The [General Overview](#general-overview) section below contains all the required steps to get you started. 
 
-If you're interested in getting a quick overview of the changes you will need to make to your existing data files, please check the [Data Loading: What You Need To Change](Data-Loading-%3A-What-You-Need-To-Change.md) page.
+If you're interested in getting a quick overview of the changes you will need to make to your existing data files, please check the [Data Loading: What You Need To Change](Data-Loading-:-What-You-Need-To-Change.md) page.
```
pieterlukasse: this `%3A` was on purpose because `:` apparently doesn't work. I tried it out in the readthedocs page and it also fails there now.

docs/Makefile
```diff
@@ -0,0 +1,230 @@
...
+pseudoxml:
+	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
+	@echo
+	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+
+.PHONY: dummy
+dummy:
+	$(SPHINXBUILD) -b dummy $(ALLSPHINXOPTS) $(BUILDDIR)/dummy
+	@echo
+	@echo "Build finished. Dummy builder generates no files."
```
pieterlukasse: is this file specific to cBioPortal? Seems like a generic file. I would say that if we need some custom configuration here, maybe try to limit it to that. So perhaps remove all latex, xml, etc parts that are not applicable.

docs/conf.py
```diff
@@ -0,0 +1,303 @@
...
+
+# If true, do not generate a @detailmenu in the "Top" node's menu.
+#texinfo_no_detailmenu = False
+
+def setup(app):
+    app.add_config_value('recommonmark_config', {
+            'url_resolver': lambda url: github_doc_root + url,
+            'enable_auto_toc_tree': False,
+            }, False)
+    app.add_transform(AutoStructify)
```
pieterlukasse: same comment as given to the Makefile

pieterlukasse:
Nice :+1: 

I reviewed it and tested your site. It looks very nice! I added a few comments to the code and I have one comment from my tests:

- I tried out the search function and it turns out to be case sensitive for parts that are "code". Example: searching for `cbioportalImporter.py` or `cbioportalimporter.py` gives different results (in this case only the first one gives matches).

ecerami:
@inodb great idea!

I probably won't have time to check this out for a few days.  But, very happy that you have done this!

docs/Data-Loading.md
inodb: Ah well spotted, I'll take a look

docs/Makefile
inodb: Yeah this is just a standard Makefile sphinx generates when you use `sphinx-quickstart`. I've only been using the `html` rule now. If you want to build the docs locally, it's useful. I could just leave the `html` rule in, to make things less verbose.

docs/cBioPortal-ER-Diagram.md
```diff
@@ -1,2 +1,4 @@
+# cBioPortal ER Diagram
```
inodb: It's the first link under 4. Development, right?

I did notice now there is some collapse/expand + sign on the sidebar to the left that doesn't behave entirely as expected. Perhaps not a big deal tho

docs/conf.py
inodb: This is also autogenerated by sphinx. I think it's useful though in this case to have all the config with comments in there, for instance I just spotted the `html_logo` config where we can put the cbioportal logo. Saves me some googling

inodb:
@pieterlukasse Thanks for reviewing so promptly! Really useful comments.
- Good job on spotting the case sensitive searching. Hopefully this is just some config value somewhere
- I found a config for the logo, will have a look at that. Not sure about the styling. You change the theme so it's def possible, but if we can keep the theme and just change the color that would be ideal. Saves us from having to update the theme ourselves.

@ecerami Thanks! I think I'm good with @pieterlukasse's comments for now so in a few days should be fine.

ecerami:
@inodb: sorry for the delay.  I think this looks great!  and, so much better than the current docs.

A few comments / questions:

1.  the left hand side bar seems off, e.g. it doesn't match exactly to the content on the right. e.g. there is a little plus sign next to cBioPortal Deployment, but when you click to open, you get sub-sections which were already displayed.  Seems a little buggy. 
2.  will this get automatically re-generated each time we push to master?
3.  how should we direct people here? e.g. should we change the text here:  https://github.com/cBioPortal/cbioportal/wiki and here:  https://github.com/cBioPortal/cbioportal

inodb:
Thanks for your comments @pieterlukasse and @ecerami. I updated the PR accordingly:

http://cbioportal-inodb.readthedocs.io/en/latest/

- Add custom css style file
- Use cBioPortal-esque theme with logo
- Get rid off collapsable items in side bar
- Rename all markdown files with colon in the name, because URL resolve won't
  work otherwise.
- Include only parts of Makefile and conf.py that are relevant to html output

@pieterlukasse 
- The case insensitive searching is a slightly more complicated issue. Basically `cbioportalImporter` is searched for as two words: `cbioportal` and `importer`, whereas `cbioportalimporter` is one word. Looks like they automatically change camel case searches to multiple words. Couldn't really find much about this, but maybe it's okay to leave it this way.
- As mentioned above I just renamed files with a `:` to get around the issue of the URLs not working. The `%3A` fix didn't work with sphinx.

@ecerami 

1. Removed visibility of `+` symbols in the sidebar with a custom styling sheet.
2. This will indeed get automatically regenerated if we add the readthedocs hook to the cbioportal repo. The nice thing is that we can also generate the docs for multiple branches and versions i.e. not just latest master. There's a button in the bottom left corner that allows you to switch between versions if we were to generate them.
3. Yes, I think we should change the wiki and the README links. Once this is merged I'll update those links to point to the right location.

Final notes:

The version switching button in the lower left corner is a little out of style. I'm trying to move it to the bottom right corner like this site: http://www.writethedocs.org/guide/tools/sphinx-themes/. Just asked on twitter if this is maybe some conf value that can be changed. Otherwise I can change the CSS to do something similar.

pieterlukasse:
@inodb looks nice. 

Regarding the search functionality, I logged a ticket: https://github.com/rtfd/readthedocs.org/issues/2328 Hopefully they will fix this soon. 

Regarding the style, I like the logo. For the font type and background color I would prefer the github style on a white background, which looks a bit more modern I think (but this is just my personal opinion):

image

inodb:
@pieterlukasse Thanks for the quick review and posting that issue!

I actually prefer the rtfd theme's typography over the github one. @zhx828 prefers left as well. I can be converted with more :+1:s on right ;)

pieterlukasse:
@inodb you're welcome. 

So how do you do a poll in a github issue :smile:  I added my :+1: to my comment :wink:

inodb:
@pieterlukasse looks like it's 3 👍 left and 2 👍 right. Still pretty equal so let's leave it as is. @ecerami has no time to review further, but I addressed his previous comments. Ok with you if I merge this @pieterlukasse ?

jjgao:
@inodb good job.
should this be added to hotfix?
how can we point users to rtfd.org?

inodb:
@jjgao Made a pull request to hotfix https://github.com/cBioPortal/cbioportal/pull/1546.

In regards to referring users. Ethan mentioned to update both https://github.com/cBioPortal/cbioportal/wiki and https://github.com/cBioPortal/cbioportal. I believe there's a badge as well that can be added to the cBioPortal README.

inodb:
@ecerami @jjgao @pieterlukasse This has been merged to hotfix instead: https://github.com/cBioPortal/cbioportal/pull/1546.

I changed the links on https://github.com/cBioPortal/cbioportal and https://github.com/cBioPortal/cbioportal/wiki to point to readthedocs.

PR #1468, opened 2016-07-08
Remove otherCase check and error - display warning if record not stored

zheins:
# What? Why?

miRNA genes needs a better solution - for now if a duplicate is found in the DB just skip it and put a warning. Don't error out of import.

- Remove otherCase check and just output a warning if fail to store record for a gene.
# Notify reviewers

@jjgao @pieterlukasse

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
@@ -405,8 +405,7 @@ public class ImportTabDelimData {
 	                            }
                             }                            
                         } else {
-                        	//TODO - review: is this still correct?
-                        	int otherCase = 0;
+							int otherCase = 0;
                             for (CanonicalGene gene : genes) {
                             	if (gene.isMicroRNA() || rppaProfile) { // for micro rna or protein data, duplicate the data
 	                            	boolean result = storeGeneticAlterations(values, daoGeneticAlteration, gene, geneSymbol);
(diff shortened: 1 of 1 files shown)


pieterlukasse:
Hi @zheins, well spotted. I think however that the `otherCase` needs to stay, it can be used to differentiate between the scenario where no miRNA or RPPA could be stored versus where there was an ambiguous gene in some other case. Note: here I am still assuming that in the miRNA and RPPA scenario it is acceptable to have one entry ambiguously pointing to multiple entrez ids (maybe @jjgao can comment on that - I think that is also why there was a "TODO"  comment there...i.e. to check if this is really correct).

So the correct code in my opinion would be something like: 

```java
if (!recordStored) {
    if (otherCase == 0) {
    	//this means that miRNA or RPPA could not be stored
    	ProgressMonitor.logWarning("Could not store miRNA or rppa data"); //TODO detect the type of data and give specific warning
    }
    else if (otherCase > 1) {
    	//this means that genes.size() > 1 and data was not rppa or microRNA, so it is not defined how to deal with 
    	//the ambiguous alias list. Report this:
    	ProgressMonitor.logWarning("Gene symbol " + geneSymbol + " found to be ambiguous. Record will be skipped for this gene.");
    }
    else { 
    	//should not occur. It would mean something is wrong in preceding logic (see else if (genes.size()==1) ) or a configuration problem, e.g. where a symbol maps to both a miRNA and a normal gene:
    	throw new RuntimeException("Unexpected error: unable to process row with gene " + geneSymbol);
    }
}

zheins:
@pieterlukasse that makes sense to me. @jjgao what do you think? What Pieter is suggesting is how I initially fixed our issue, but we didn't understand what `otherCase` was.

jjgao:
@zheins agreed with @pieterlukasse

jjgao:
@zheins could you make the change as @pieterlukasse suggested?

zheins:
@jjgao just rebased.

jjgao:
@pieterlukasse could you review and approve this pr?

jjgao:
@zheins @pieterlukasse is this pull request still relevant?

pieterlukasse:
@jjgao @zheins I see the changes I requested were made, so PR approved 👍

jjgao:
@pieterlukasse thanks. We will do this for hotfix release after 1.5.0.

zheins:
@jjgao I just came across this bug again in a study I was importing. Let's try to get this merged asap.

PR #1474, opened 2016-07-11
Issue 1455 - metaImport needs a -n option

### LINKED ISSUE #1455: metaImport needs a -n option

ersinciftci:
# What? Why?
Fix #1455

# Notify reviewers
@jjgao

### DIFF
--- a/core/src/main/scripts/importer/metaImport.py
+++ b/core/src/main/scripts/importer/metaImport.py
@@ -52,6 +52,10 @@ def interface():
                                    help='Path to a directory of cBioPortal '
                                         'info files to be used instead of '
                                         'contacting the web API')
+    portal_mode_group.add_argument('-n', '--no_portal_checks', default=False,
+                                       action='store_true',
+                                       help='Skip tests requiring information '
+                                            'from the cBioPortal installation')
     parser.add_argument('-jar', '--jar_path', type=str, required=False,
                         help='Path to core JAR file')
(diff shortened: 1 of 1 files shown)


pieterlukasse:
@jjgao This looks good to me, assuming it was successfully tested. :+1: 

@ersinciftci : one remark: according to the comment https://github.com/cBioPortal/cbioportal/pull/1458#issuecomment-231398801, this is to be a *temporary* solution. I would therefore recommend adding something in the code that identifies it as such. Maybe a comment just above the -n option, something like:
```python
#  TODO: workaround to simplify import process when no web-server is running. Will be replaced by solution for #1466
```

ersinciftci:
Added the comment @pieterlukasse suggested.

PR #1481, opened 2016-07-12
Update File-Formats.md

oplantalech:
Changes proposed in this pull request: two updates in the documentation of file formats.
1- To show the Pubmed link in the Data Sets page, the citation field should be filled.
2- Stable_id (from case lists) must start with the study_id followed by an undrescore. If not, the file does not pass the validator.


oplantalech:
@jjgao This PR has not been reviewed yet, it is just a small change in documentation. The code part was on #1486 and it has already been merged. And Travis is failing because the test that was fixed in rc by PR #1657 was never fixed in hotfix.

Issue #1480, opened 2016-07-12
Querying for genes in study view is broken


fedde-s:
On [http://www.cbioportal.org/beta/study.do?cancer_study_id=brca_tcga#summary], while one or more gene symbols are selected in the textbox at the top, clicking the query button next to it does not work in Firefox, Google Chrome or Safari. The former two display a warning that a pop-up was blocked.

Oddly, when I remove the genes from the textbox and click the query button again, it does query for the genes that are no longer in the textbox.

Does anyone have an idea what’s going on and how it can be fixed?

JiaoJiao123:
@fedde-s  yeah right. I noticed that too. Also when you change gene input, it still query for the genes I entered previously. I will look into this.

fedde-s:
Might the change in commit 4d05de6544b5f1a4518a8e5c91b94d22695ffac5 have something to do with this issue?

JiaoJiao123:
@fedde-s yeah maybe, I am not sure. but we want to open a new page upon clicking query.

fedde-s:
I am not an expert, but I can imagine that it might be because `QueryByGeneUtil.addStudyViewFields()` changes the DOM between the user interaction (the click) and the attempt to open the new tab. I have not tested whether this is what’s happening though.

JiaoJiao123:
@fedde-s yeah that's also what I am thinking. It has already been fixed in the pull request https://github.com/cBioPortal/cbioportal/pull/1482

pieterlukasse:
this bug is back in 1.3.0:

![image](https://cloud.githubusercontent.com/assets/2900303/19341509/f0775f96-912d-11e6-90eb-ab20f0401ad9.png)

pieterlukasse:
@JiaoJiao123 @zhx828 :point_up_2:

PR #1482, opened 2016-07-12
Query bug in study page

JiaoJiao123:
The previous hidden input were not removed before we append new ones to the submitted form. That's why it always query on the old gene input.


JiaoJiao123:
Related to Issue #1480 Querying for genes in study view is broken

JiaoJiao123:
@fedde-s  It will query based on current gene input now. And could you help with code review and testing?

pieterlukasse:
@JiaoJiao123 : I tested it here and could not reproduce the problem on hotfix. This means I also cannot check if it works. I tested in Linux/Chrome, Linux/Firefox, Windows/Edge... all on hotfix branch (so not your fix). 

I noticed that other things fail though, like: 
- I cannot open the study view again after doing this query once
- The oncoprint seems to fail in Edge and I only see "Loading...." 
   - after running a second time I get a HTTP 500 with java.lang.OutOfMemoryError: Java heap space, around QueryBuilder.java:row 434

So these are my findings for now. At the moment @fedde-s and I are quite busy with some other topics. Could you maybe consider these findings and ask @zhx828 to review your code and perhaps @adamabeshouse to check why this QueryBuilder step is taking so much memory now?

JiaoJiao123:
@pieterlukasse: hum that's very interesting. I just build a portal instance from hotfix branch and it still has this problem. Same with the query bug in http://www.cbioportal.org/  

Thank you very much for pointing out the other issues, but sorry I didn't see them from hotfix branch. I will work on them if I reproduce them later. 

And hongxin @zhx828 will help with code review and testing.

zhx828:
Code looks good to me.

PR #1486, opened 2016-07-13
Add a warning when pmid exists but citation does not

oplantalech:
A warning has been added to remind that if meta study provides a pmid, then a citation should also be added to be visible in Data Sets. Related to #1481.


pieterlukasse:
code looks good :+1:

PR #1489, opened 2016-07-13
Define Locale in NormalizeExpressionLevels.java

oplantalech:
NormalizeExpressionLevels.java calculates Z-scores from copy number and expression files. However, the values are shown using the default Locale from the computer. This can be a problem if the computer is configured in German, Spanish or other languages where they use commas instead of points as decimal separator.


zheins:
This looks good to me. @oplantalech can you rebase this PR?

oplantalech:
@fedde-s This is ready to merge. Travis is failing because the test that was fixed in rc by PR #1657 was never fixed in hotfix.

PR #1490, opened 2016-07-13
Fix bugs when installing cBioPortal in Windows

oplantalech:
Two Windows-specific bugs that appear when installing cBioPortal have been solved:
1- The `\\` were not correctly interpreted (apparently, Java was escaping and then, MySQL did it too).
2- `newLine()` was not correctly interpreted, so it has been replaced with `.write("\n")`


jjgao:
@n1zea144 could you review?

PR #1499, opened 2016-07-15
Make study query suggestions configurable

fedde-s:
Allow administrators of cBioPortal installations to configure the example queries listed in the dropdown menu of the study search box, as requested in issue #1380.

This can be useful in installations that do not contain the datasets used as examples by default. Administrators of such installations want to customise the default suggestions and replace them with ones that make sense for the data in their portal.

screen shot 2016-06-29 at 16 18 57

- parse and document a portal property named `skin.example_study_queries`, a newline-separated list with the previously hard-coded values as the default value
- generate the list of query suggestions to filter studies in the query view based on the new portal property


fedde-s:
@jjgao Here’s the pull request to hotfix you suggested, replacing #1423.

Issue #1496, opened 2016-07-14
Remove unimplemented OQL features from documentation


adamabeshouse:
In particular, as noticed by @pieterlukasse, the syntactic sugar (e.g. `CNA >= GAIN` or just `CNA` for `AMP HOMDEL`)

pieterlukasse:
@adamabeshouse thanks for adding the ticket. I would like to propose a review of the documentation at http://www.cbioportal.org/onco_query_lang_desc.jsp. So:
- [ ] test each of the examples listed there
- [ ] remove or fix the ones that don't work

New users will probably follow these examples first, before trying their own. @schultzn : can you also help with this one, testing and giving feedback on what can be removed and what should be fixed?

pieterlukasse:
there is also a small bug in the label shown when using OQL. See screenshot below:

![image](https://cloud.githubusercontent.com/assets/2900303/16869639/10e9af8c-4a7e-11e6-804f-5cce9725c931.png)

The expression was on `D96` but the label only shows `96` (I assume it is just a problem with the label - haven't tested further yet). 

One small request from a user was: could the expression be shown on the screen itself instead of only at the pop-up? In this way it is easier to use it in a presentation or document/article. Otherwise you have to edit the figure to add these expressions manually.

schultzn:
I don’t see the same issue in a similar query … see below

I don’t understand the user request. What do you mean by “expression”? mRNA expression level?

pieterlukasse:
@schultzn with "expression" I meant the OQL query, e.g.  `BRCA1:MUT=D96`, sorry for the confusion. 

In your last message it looks like you intended to add a screenshot, but I think that failed somehow. Can you try again?

schultzn:
I responded via email - maybe that does not support including images. I will try again later via the web interface.

I don’t see how an OncoPrint would permanently show the OQL text, i.e. the actual mutation. There is simply no space in this compact visualization.

Niki.

adamabeshouse:
It seems like there may be some miscommunication here.. 

@pieterlukasse I see the issue and I am in the process of fixing it - the issue is that since the amino acid is really redundant information as far as filtering goes, it's not stored during parsing, and so isn't available when printing the query out.

pieterlukasse:
@schultzn : yes, space is quite limited. Maybe we can make it a view option? We already have some options there, so perhaps we could add it there (see screenshot):
[...]

PR #1500, opened 2016-07-15
fix OQL displaying issue where amino acid at position not shown

adamabeshouse:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/1496#issuecomment-232903812

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix ( This is a bug fix to a feature that's in `rc` )

@pieterlukasse


pieterlukasse:
Tested `APC:MUT=R856C MUT=R856H;` on CCLE study and worked fine:

image

Also tested `APC:MUT=R856C` and `APC:MUT=R856H` separately and this gave the correct results (correct sample selected) :+1:

portal/src/main/webapp/js/lib/oql/oql_basic.pegjs
```diff
@@ -80,15 +80,15 @@ ComparisonOp
...
+	/ "NONSENSE"i { return {"type":"class", "value":"NONSENSE", "info":{}}; }
+	/ "NONSTART"i { return {"type":"class", "value":"NONSTART", "info":{}}; }
+	/ "NONSTOP"i { return {"type":"class", "value":"NONSTOP", "info":{}}; }
+	/ "FRAMESHIFT"i { return {"type":"class", "value":"FRAMESHIFT", "info":{}}; }
+	/ "INFRAME"i { return {"type":"class", "value":"INFRAME", "info":{}}; }
+	/ "SPLICE"i { return {"type":"class", "value":"SPLICE", "info":{}}; }
+	/ "TRUNC"i { return {"type":"class", "value":"TRUNC", "info":{}}; }
+        / "FUSION"i { return {"type":"class", "value":"FUSION", "info":{}}; }
+        / letter:AminoAcid position:NaturalNumber string:String { return {"type":"name" , "value":(letter+position+string), "info":{}};}
+        / letter:AminoAcid position:NaturalNumber { return {"type":"position", "value":parseInt(position), "info":{"amino_acid":letter.toUpperCase()}}; }
```
pieterlukasse: why is there this difference between this line and the line above? I.e. in the previous line the `value` is set as `(letter+position....` while here the letter suddenly needs to be in `info`.

portal/src/main/webapp/js/lib/oql/test-oql-parser.js
```diff
@@ -11,7 +11,16 @@ var testAlterationEquality = function(expected, given) {
...
 				if (exp_alt.hasOwnProperty(key)) {
 					same_keys = same_keys && giv_alt.hasOwnProperty(key);
-					same_values = same_values && (giv_alt[key] === exp_alt[key]);
+					var giv_val = giv_alt[key];
+					var exp_val = exp_alt[key];
+					if (typeof giv_val !== typeof exp_val) {
+					    same_values = false;
+					    break;
+					} else if (typeof giv_val === "object") {
+					    same_values = same_values && testAlterationEquality(giv_val, exp_val);
```
pieterlukasse: not sure if I follow this recursion...what is the goal here? Maybe add a documentation line?

portal/src/main/webapp/js/lib/oql/oql_basic.pegjs
adamabeshouse: This is due to how PEG's (parser expression grammars) work, and because I need to handle separately the case where there is stuff after the position (full amino acid change specified) and no stuff after the position (only position specified).

PEG's match greedily, so if those lines are in the opposite order, there will be errors for full amino acid specifications which get partially matched into the position specification. 

Now, if I match the amino acid specification with a general string, then everything will be matched and the position specification will never be reached.

Notice the lines are actually the same except for the addition of `info`, nothing is subtracted from what is already there.

portal/src/main/webapp/js/lib/oql/test-oql-parser.js
adamabeshouse: this is to test that the object is equal, i.e. same keys and same values, and the recursion is to do this if some of the values are themselves objects, which can't be checked with `===`.

jjgao:
@adamabeshouse should this be point to hotfix?

adamabeshouse:
@jjgao Not yet, it's a fix to something that's in rc but not released yet (the new data manager), unless I'm mistaken.

PR #1516, opened 2016-07-20
Add Pull Request Merge Policy & Reviewers Guide

### LINKED ISSUE #1512: Improve Review Procedure
The current review procedure is too slow to keep up with all the open PRs. We need to formalize this procedure. I propose we add two new docs in `docs/`:

1. Pull Request Policy. Outlining what teams need to acknowledge what type of feature/bug fix. This involves creating the teams in the cBioPortal organization based on areas of competence.
2. How to review a PR. Describing in detail what needs to be checked when you review a PR. We currently outline in `CONTRIBUTING` and `PULL_REQUEST_TEMPLATE` what a submitter needs to do, but we don't have any documentation on what a reviewer's responsibilities are.

inodb:
# What? Why?
Fix #1512 

- Add documentation on the pull request merge policy in CONTRIBUTING.md. 
- Update the PULL_REQUEST_TEMPLATE.md to point to this policy.

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
@morungos @ersinciftci @ecerami @jjgao @n1zea144 @sheridancbio @zhx828 @fedde-s @pieterlukasse 

@cBioPortal/core-developers

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,8 +20,14 @@ If this is a new visual feature please add a before/after screenshot or gif
 here with e.g. [GifGrabber](http://www.gifgrabber.com/).
 
 # Notify reviewers
-If your pull request involves changes to an existing file, one or more of the
-people that edited before you might be good candidates to review it. Please use
-`git blame <filename>` to determine that and notify them here. Otherwise notify
-everybody in the core team:
-@cBioPortal/core-developers
+Read our [Pull request merging
+policy](../CONTRIBUTING.md#pull-request-merging-policy). If you are part of the
+cBioPortal organization, notify the approprate team (remove inappropriate):
+
+@cBioPortal/frontend
+@cBioPortal/backend
+@cBioPortal/devops
+
+If you are not part of the cBioPortal organization look at who worked on the
+file before you. Please use `git blame <filename>` to determine that
+and notify them here:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,9 +94,9 @@ Pull Requests (PRs) are reviewed by the
 For each PR the submitter should propose what team(s) is/are appropriate to
 review it. This is the current merging policy:
 
-- A documentation change needs one **+1**
-- A simple bugfix to rc or hotfix requires one **+1**
-- A new feature requires two **+1**. One from someone at MSKCC and one from
+- A documentation change needs one **Approve**
+- A simple bugfix to rc or hotfix requires one **Approve**
+- A new feature requires two **Approve**. One from someone at MSKCC and one from
   another institution.
 
 If these requirements are met, any person with merge rights can merge to rc or
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,67 @@ Once you downloaded the images you do the following for each screenshot:
 - If the change in the screenshot is **desired**, add the screenshot to the
   repo, commit it and push it to your PR's branch.
 
+## Pull Request Merging Policy
+Pull Requests (PRs) are reviewed by the
+[backend](https://github.com/orgs/cBioPortal/teams/backend),
+[frontend](https://github.com/orgs/cBioPortal/teams/frontend) and
+[devops](https://github.com/orgs/cBioPortal/teams/devops) teams of cBioPortal.
+For each PR the submitter should propose what team(s) is/are appropriate to
+review it. This is the current merging policy:
+
+- A documentation change needs one **+1**
+- A simple bugfix to rc or hotfix requires one **+1**
+- A new feature requires two **+1**. One from someone at MSKCC and one from
+  another institution.
+
+If these requirements are met, any person with merge rights can merge to rc or
+hotfix.
+
+## Pull Request Reviewers Guide
+Here we describe the guidelines for the reviewer. Always follow the checks in
+general, then follow the other checks that apply:
+
+### General
+- Double check all the things in the **Checks** section of the Pull Request.
+  Remind the submitter if any of them are not fulfilled
+- Are the test cases spanning a decent amount of scenarios? It is the
+  submitters as well as the reviewers responsibility to not let any errors
+  sneak into the portal.
+
+Bug fixes:
+
+- Should the bug that causes the issue be added as a test case?
+
+New features:
+
+- If this is a new feature make sure the proposed changes are in line with the
+  current planning of cBioPortal e.g. is the right API used, is this in line
+  with current refactoring efforts.
+
+### Backend
+New features:
+
+- Is the new persistence stack used?
+
+### Frontend
+New features:
+
+- What APIs are used to get the data? Is the REST API used?
+- Should this be a separate library in a separate repo or should it be part of cBioPortal?
+- Are dependencies properly listed? Ideally in a package.json
+- How is the package included in cBioPortal?
+
+### Devops
+New features:
+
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -86,6 +86,64 @@ Once you downloaded the images you do the following for each screenshot:
...
+For each PR the submitter should propose what team(s) is/are appropriate to
+review it. This is the current merging policy:
+
+- A documentation change needs one **+1**
+- A simple bugfix to rc or hotfix requires one **+1**
+- A new feature requires two **+1**
+
+If these requirements are met, any person with merge rights can merge to rc or
+hotfix.
+
```
pieterlukasse: I think that for new features it is good to include a reviewer from another institute

CONTRIBUTING.md
```diff
@@ -86,6 +86,64 @@ Once you downloaded the images you do the following for each screenshot:
...
+- What APIs are used to get the data? Is the REST API used?
+- Should this be a separate library in a separate repo or should it be part of cBioPortal?
+- Are dependencies properly listed? Ideally in a package.json
+- How is the package included in cBioPortal?
+
+### Devops
+New features:
+
+- If there are any new configuration properties, do the follow the same
```
pieterlukasse: typo: "do they follow the ..."

CONTRIBUTING.md
```diff
@@ -86,6 +86,64 @@ Once you downloaded the images you do the following for each screenshot:
...
+  structure as existing configuration style?
```
pieterlukasse: what do you mean by structure? Naming conventions? Because we do have a wild range of value types in the properties file itself (which I think is ok btw)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,8 +20,13 @@ If this is a new visual feature please add a before/after screenshot or gif
...
-people that edited before you might be good candidates to review it. Please use
-`git blame <filename>` to determine that and notify them here. Otherwise notify
-everybody in the core team:
-@cBioPortal/core-developers
+Read our [Pull request merging policy](../CONTRIBUTING.md#pull-request-merging-policy). Notify the approprate
+team (remove inappropriate):
+
+@cBioPortal/frontend
+@cBioPortal/backend
+@cBioPortal/devops
```
fedde-s: Are these teams hidden from external developers? They do not show up as links or valid at-mentions for me. Neither did cBioPortal/core-developers.

fedde-s: Probably the same issue: the hyperlinks added to CONTRIBUTING.md (https://github.com/orgs/cBioPortal/teams/backend, etc) give a 404 Not Found error for me.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,8 +20,13 @@ If this is a new visual feature please add a before/after screenshot or gif
...
+
+To get your PR merged faster it can be a good idea to look at whom worked on
```
fedde-s: Typo: “look at who worked on the file”

CONTRIBUTING.md
inodb: Great suggestion! Will update

CONTRIBUTING.md
inodb: Yes, I should clarify that a bit. I meant more in the sense that if we are adding new configuration values, they should ideally be in the `.properties` file. Default values go in `.EXAMPLE` or inside the pom followed by `:default-value` and ideally new config shouldn't break existing `.properties` files. I was hoping some folks of @cBioPortal/devops could extend this.

fedde-s: Aren’t the defaults for most customisation options in https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java ?

inodb: Good point I've actually never updated that file. Well shows how useful it is to have a checklist like this. Is `GlobalProperties.java` the place where we should put the default values? Or do we use the `:default-value` syntax in the pom file. Or do we not care? @n1zea144 @morungos

.github/PULL_REQUEST_TEMPLATE.md
inodb: I see, that's a bummer. Looks like only members of the organization can see the teams. Maybe we should add a document then showing the teams.

CONTRIBUTING.md
```diff
@@ -138,8 +139,8 @@ New features:
 ### Devops
 New features:
 
-- If there are any new configuration properties, do the follow the same
-  structure as existing configuration style?
+- Does the configuration style follow the config guidelines? That is compile
+  (Maven) config goes in `pom.xml`, runtime (Spring) goes in `portal.properties`.
```
fedde-s: I’m not sure if the line about `pom.xml` adds much here, as there’s multiple files with that name in the project, and I expect that Maven has its own documentation on where to configure the build process.

I do think it would be useful to add that defaults for anything in `src/main/resources/portal.properties` (which is really an installation-specific configuration file and therefore not under version control) go in `core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java`.

fedde-s: The lack of clarity on that last point has already led to at least one broken view in the past, see pull request #1389. A newly introduced setting had a suggested value in `portal.properties.EXAMPLE` but no default in `GlobalProperties.java`.

inodb: I was thinking that any `pom.xml` would do, whichever is appropriate for the compile config. I guess if it applies only to `scripts/`. It should be in `scripts/pom.xml` etc. Will try to be more explicit there. It should just be clear that compile parameters should not be in neither `GlobalProperties.java` or `.properties`.

I assume `GlobalProperties.java` is only for accessing runtime config. If that's the case it could make sense to put the defaults in there. Then you can overload those parameters with a `.properties` file and ideally also environment variables. We also have Spring runtime config, so I'm not entirely sure how that fits in with the `GlobalProperties,java`.

Do you have any comments on this @morungos?

jjgao:
@inodb this looks good to me. Should we merge it?

inodb:
@jjgao I made some small wording changes as @fedde-s proposed. You can merge it now. We can update it iteratively.

sheridancbio:
Sorry .. I mean to weigh in earlier on this. I'm in favor of merging as is, and we can continue to improve the process in the future. Some clarification questions though about the "+1" rules though...

In order for a developer to give a "+1" should they have performed a full code review? Or can they just be expressing their intuition and opinion about whether a PR "seems good"? Since most merges will be done based on a single "+1", I think we should be clear that giving a PR a "+1" means that you have fully review the PR (run tests, confirmed that our processes and standards have been followed, etc..) and that you are claiming that you have "signed off" on validating the PR as being "Approved for merging". I think the terminology of "+1" tends to mean "I like this" in social media, so maybe we should use a more formal language.

[more clarifications coming .. submitting comment now, since I see that tests are being rerun on this PR right now]

jjgao:
@sheridancbio very good points. For +1, maybe we should use `Approve` which sounds more formal.

jjgao:
Two more points:

* A pull request should be `Approved by` all `Assignees` before merged. 
* Only the `releaser` can merge, so that s/he knows what have been merged. -- we should find a way to rotate this role.

jjgao:
@inodb agreed on "some smaller fixes for testing or bugs should be merged quickly". For feature release, we need to formalization the decision procedure about what features can/cannot be released and when -- TBD.

Issue #1508, opened 2016-07-19
Duplicate SLF4J component errors


morungos:
I just started getting SLF4J noise in my tests on `rc`:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/stuart/git/cbioportal/business/target/business-1.3.0-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/stuart/.m2/repository/org/apache/logging/log4j/log4j-slf4j-impl/2.5/log4j-slf4j-impl-2.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/stuart/.m2/repository/org/slf4j/slf4j-log4j12/1.6.6/slf4j-log4j12-1.6.6.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
```

morungos:
I think these are a consequence of shading everything. i have no idea why we do this, as it bloats immensely. I removed the business shading and one of these immediately went away. The second looks like we have two binding present.

morungos:
The other part of the problem is the use of dependencies directly in the pom.xml, which means every subproject gets everything. This means subprojects that use a different version will get both. it shouldn't happen, but again, it's not right that this happens.

morungos:
And these are coming from biojava. Dependency tree is:

```
[INFO] +- org.biojava:biojava-core:jar:4.2.0:compile
[INFO] |  +- org.apache.logging.log4j:log4j-slf4j-impl:jar:2.5:runtime
[INFO] |  +- org.apache.logging.log4j:log4j-api:jar:2.5:runtime
[INFO] |  \- org.apache.logging.log4j:log4j-core:jar:2.5:runtime
```

PR #957, opened 2016-03-23
Initial debloating work, currently untested. See #948

morungos:
As commented elsewhere, this is just an initial cut and hasn't yet been tested as a deployed web application. However, the scripts jar file does seem to work OK, and is now a more reasonable size.

PR #1517, opened 2016-07-20
Debloating lite

### LINKED ISSUE #948: Excessive bloat in the war files with shaded jars
The war file built by Maven is immense. I've built a single war which works fine for the main web application that's about 60Mb, but the standard one is 150Mb. Mainly this seems to be because the core and business jars are shaded and contain a lot of the same stuff, and many of these components are also embedded in the war itself. So there are three copies of some dependencies in the final package, at least. 

It's not really clear why these get shaded, or why the shaded versions are deployed in the war.

morungos:
# What? Why?
Fix #948 and #1508.

- Add a new scripts component which is shaded and contains a runnable core
- Removed shading from core and business
- Changed the output war name to include a version
- Resolved the SLF4J duplications

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -167,11 +167,39 @@
       <groupId>org.biojava</groupId>
       <artifactId>biojava-core</artifactId>
       <version>4.2.0</version>
+      <exclusions>
+      	<exclusion>
+      	  <groupId>org.apache.logging.log4j</groupId>
+      	  <artifactId>log4j-slf4j-impl</artifactId>
+      	</exclusion>
+      	<exclusion>
+      	  <groupId>org.apache.logging.log4j</groupId>
+      	  <artifactId>log4j-api</artifactId>
+      	</exclusion>
+      	<exclusion>
+      	  <groupId>org.apache.logging.log4j</groupId>
+      	  <artifactId>log4j-core</artifactId>
+      	</exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.biojava</groupId>
       <artifactId>biojava-structure</artifactId>
       <version>4.2.0</version>
+      <exclusions>
+      	<exclusion>
+      	  <groupId>org.apache.logging.log4j</groupId>
+      	  <artifactId>log4j-slf4j-impl</artifactId>
+      	</exclusion>
+      	<exclusion>
+      	  <groupId>org.apache.logging.log4j</groupId>
+      	  <artifactId>log4j-api</artifactId>
+      	</exclusion>
+      	<exclusion>
+      	  <groupId>org.apache.logging.log4j</groupId>
+      	  <artifactId>log4j-core</artifactId>
+      	</exclusion>
+      </exclusions>
     </dependency>
 
     <!-- Mockito -->
@@ -287,34 +315,6 @@
 
   <build>
     <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-shade-plugin</artifactId>
-        <version>2.3</version>
-        <executions>
-          <execution>
-            <phase>package</phase>
-            <goals>
-              <goal>shade</goal>
-            </goals>
-            <configuration>
-              <artifactSet>
-                <excludes>
-                  <exclude>org.bouncycastle:bcprov-jdk15</exclude>
-                </excludes>
-              </artifactSet>
-              <transformers>
-                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
-                  <resource>META-INF/spring.handlers</resource>
-                </transformer>
-                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
-                  <resource>META-INF/spring.schemas</resource>
-                </transformer>
-              </transformers>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
       <!-- surefire (test) config -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
@@ -355,8 +355,10 @@
             <configuration>
             <!-- username & password gets put in .m2/settings.xml -->
               <settingsKey>settingsKey</settingsKey>
-              <driver>com.mysql.jdbc.Driver</driver>
-              <url>jdbc:mysql://127.0.0.1:3306/cgds_test</url>
+              <driver>${db.test.driver}</driver>
+              <url>${db.test.url}</url>
+              <username>${db.test.username}</username>
+              <password>${db.test.password}</password>
               <sqlCommand>SET storage_engine=INNODB</sqlCommand>
               <sqlCommand>SET SESSION sql_mode = 'ANSI_QUOTES'</sqlCommand>
               <srcFiles>
--- a/pom.xml
+++ b/pom.xml
@@ -26,97 +26,12 @@
 
   <profiles>
     <profile>
-      <id>mskcc</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>mskcc</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>genie</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>genie</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>private</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>private</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>triage</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>triage</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>gdac</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>gdac-portal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>su2c</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>su2c-portal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
       <id>public</id>
       <activation>
         <activeByDefault>true</activeByDefault>
       </activation>
       <properties>
         <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>cbioportal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>prostate</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>prostate-portal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>kras</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>kras</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>target</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>target-portal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>glioma</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>glioma-portal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>acc</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>acc-portal</final-war-name>
-      </properties>
-    </profile>
-    <profile>
-      <id>ctd2-dashboard</id>
-      <properties>
-        <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
-        <final-war-name>ctd2-dashboard-portal</final-war-name>
       </properties>
     </profile>
     <profile>
@@ -212,6 +127,13 @@
     <maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
     <jacoco-maven-plugin.version>0.7.6.201602180812</jacoco-maven-plugin.version>
     <mockito.version>1.10.19</mockito.version>
+    
+    <final.war.name>cbioportal-${project.version}</final.war.name>
+
+	<db.test.driver>com.mysql.jdbc.Driver</db.test.driver>
+	<db.test.url>jdbc:mysql://127.0.0.1:3306/cgds_test</db.test.url>
+    <db.test.username>cbio_user</db.test.username>
+    <db.test.password>somepassword</db.test.password>
 
     <!-- THIS SHOULD BE KEPT IN SYNC TO VERSION IN CGDS.SQL -->
     <db.version>1.2.1</db.version>
@@ -226,6 +148,7 @@
     <module>model</module>
     <module>persistence</module>
     <module>service</module>
+    <module>scripts</module>
   </modules>
 
   <dependencies>
--- a/portal/pom.xml
+++ b/portal/pom.xml
@@ -41,11 +41,40 @@
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.tomcat</groupId>
+      <artifactId>tomcat-catalina</artifactId>
+      <version>7.0.68</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-taglibs</artifactId>
+      <version>${spring.security.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-ldap</artifactId>
+      <version>${spring.security.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-openid</artifactId>
+      <version>${spring.security.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security.extensions</groupId>
+      <artifactId>spring-security-saml2-core</artifactId>
+      <version>${spring.security.saml.version}</version>
+      <scope>runtime</scope>
+    </dependency>
   </dependencies>
 
   <build>
     <!-- final name of the app -->
-    <finalName>${final-war-name}</finalName>
     <plugins>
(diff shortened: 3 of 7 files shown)


pom.xml
```diff
@@ -212,6 +127,13 @@
     <maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
     <jacoco-maven-plugin.version>0.7.6.201602180812</jacoco-maven-plugin.version>
     <mockito.version>1.10.19</mockito.version>
+    
+    <final.war.name>cbioportal-${project.version}</final.war.name>
+
+	<db.test.driver>com.mysql.jdbc.Driver</db.test.driver>
+	<db.test.url>jdbc:mysql://127.0.0.1:3306/cgds_test</db.test.url>
+    <db.test.username>cbioportal</db.test.username>
+    <db.test.password>cbioportal</db.test.password>
```
pieterlukasse: I think the default username and password have been `cbio_user` and `somepassword` until now. See also https://github.com/cBioPortal/cbioportal/blob/master/docs/Pre-Build-Steps.md#create-the-cbioportal-mysql-databases-and-user

core/pom.xml
```diff
@@ -355,8 +355,10 @@
             <configuration>
             <!-- username & password gets put in .m2/settings.xml -->
               <settingsKey>settingsKey</settingsKey>
-              <driver>com.mysql.jdbc.Driver</driver>
-              <url>jdbc:mysql://127.0.0.1:3306/cgds_test</url>
+              <driver>${db.test.driver}</driver>
+              <url>${db.test.url}</url>
+              <username>${db.test.username}</username>
+              <password>${db.test.password}</password>
```
pieterlukasse: nice :+1:

pom.xml
morungos: Excellent, I'll modify. This just makes it that much easier for others to change and override.

inodb: There seems to be some confusion as well where default values should go: https://github.com/cBioPortal/cbioportal/pull/1516#discussion-diff-71653082R142

morungos: I don't think there are any default values for Maven right now. That was kind of the point. At least having a default value tells people how they can override them.

inodb: I understand this adds a parameter to set the test database, but I meant more that there are several ways we specify default values now:

1. specified in the `pom.xml` through `variablename:default-value` syntax. 
2. In `.properties.EXAMPLE`
3. In `GlobalProperties.java`
4. This PR seems yet another way to give a default value directly in the pom

I'm not sure what approach is best, but it would be nice to make some guidelines on this to follow.

morungos: Using the (1) pom.xml is the source for testing and for DB initialization, (2) and (3) are only for runtime. And they're bad ways to do that anyway. (1) is needed for the build process. I'm planning to get to a point where we don't need to use portal.properties for the build *at all*. This is part of that.

morungos: Yes, there are confusions here. For the build process, all settings should be through Maven, somehow. For runtime, all settings should be through externals, either as environment variables or a property file *outside the war*. The portal.properties that gets baked in during a build needs to die.

inodb: Thanks @morungos. That clears things up. I will try to add that to the guidelines for reviewing.

I'm guessing that 90% of current config in `.properties` should actually be runtime.

For runtime, if we are using a property file outside the war, I propose we follow `env variable > properties file > default in pom`.

morungos: That's my plan. I'm actually happy for a properties file to be bundled inside the war file for those runtime properties, but I actually envision that as an overlay process, where a new war is made from a standard one built during a normal build. There'll then be a vanilla standard war, but users can override elements, put in logos, and so on, in a new project that uses this vanilla war as a dependency, and doesn't need to recompile everything to do it. However, for sanity, it'd be nice to allow an external properties for post-packaging configuration, just in case.

The main thing is we need to be careful to distinguish Maven (compile) config from Spring/portal.properties (runtime) config -- with the exception that when testing, we'll probably use Maven to generate files we can use for runtime testing. These are just separate from the actual deployment.

inodb:
Does this supersede #957?

morungos:
@inodb Yes it does. And the errors are in the end to end test picking the output filename without a version (which is still think is bad). We can either override in the version, or I can use the trick I've found for extracting the version in bash and injecting it. Probably be tomorrow when I get that done.

inodb:
What is the problem exactly with not having the version name in the output file?

You can probably just change this line: https://github.com/cBioPortal/cbioportal/blob/master/test/end-to-end/docker-compose.yml#L7 to cbioportal*.war to solve the issue.

morungos:
Well, it's easier to get it built without a version number than it is to script building to put the version number back again. Although it isn't perfect on this right now, the ideal would `-Dfinal.war.name=cbioportal.war` ought to work( (it doesn't but could). I really want the output war to be installable as a Maven artifact so that I can make dependencies on it for war overlays. `cbioportal*.war` won't work because you might (in theory) edit the pom and change the version without cleaning so you get two outputs. 

I generally don't like the idea that the main output component has no visible version identification. That's inviting trouble. I have a one-liner I can use to pull the version from pom.xml and integrate into end to end testing, but it needs a little more string mangling to be perfect.

inodb:
Thanks for the explanation!

Ok I see smth like this works (http://stackoverflow.com/a/26514030/1894184):

```
MVN_VERSION=$(mvn -q \
-Dexec.executable="echo" \
-Dexec.args='${project.version}' \
--non-recursive \
org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)
```

morungos:
@inodb Oh that's simpler than I have now, but similar. The issue is filtering the output from the Maven noise, but I'm almost done without depending on running echo (which I'm not sure will always exist on certain platforms). I'm using...

```
echo 'version=${project.version}' | mvn -N org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate | grep "^version="
```

But I need to hack off the version= prefix I used to filter from the log output. `mvn -q` removes even the actual output needed :-(

morungos:
By the way, I tell a lie, adding `-Dfinal.war.name=cbioportal` does make Maven behave exactly as now, generating w/o a version in the filename.

inodb:
Oh you can pipe through sed: `sed 's/^version=//'`:

```
echo 'version=${project.version}' | mvn -N org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate | grep "^version=" | sed 's/^version=//'
```

But yeah the stackoverflow solution looks a little cleaner

morungos:
Perfect. I'll work that in tomorrow. The `-Dfinal.war.name` trick handles almost everything all the old profiles did, and profiles are a very messy way to set one variable.

inodb:
Great! I'm looking to get this mvn debloating & overlaying merged this week, so let me know if there is anything else that needs reviewing.

morungos:
Ah, I think the very simplest and most reasonable solution to this isn't to pull the version from `pom.xml`, but simply to set `final.war.name` in the `heroku` profile. This bypasses all the need for this, and correctly makes it easy for CI. Edited the PR, squashed and pushed. Let's see if this sorts it.

morungos:
Oh well, that didn't work. I thought the heroku profile was used for CI, but it seems we don't do anything special for CI in Maven. Not sure how to progress this sensibly, because I don't have docker to test out a way to work through the version removal. yes, I could make that happen by default, but it shouldn't. It's real bad to have an unversioned output artifact.

morungos:
I think there's some other issues here, too. Missing the catalina jar, which is needed for `CorsFilter`. I'm gonna make a new branch and test it.

inodb:
For heroku you should just be able to add a `*` here: https://github.com/cBioPortal/cbioportal/blob/master/heroku/Procfile#L1. I've never run heroku locally, nor do I expect people to do that, so that should get around the issue. You can also use the mvn one liner to extract the final war name if you prefer that.

I thought for the `heroku` profile in the pom we just copy the files in `heroku/` to the root directory.

For the end-to-end test, one could extract the version from the maven file on boot in this line: https://github.com/cBioPortal/cbioportal/blob/master/test/end-to-end/docker-compose.yml#L7. Something like:

```
FINAL_WAR_NAME=$(mvn -q -Dexec.executable="echo" -Dexec.args='${final-war-name}' --non-recursive org.codehaus.mojo:exec-maven-plugin:1.3.1:exec) &&
java -Ddbconnector=dbcp -jar portal/target/dependency/webapp-runner.jar --expand-war portal/target/${FINAL_WAR_NAME}
```

morungos:
@inodb But that's Heroku. We need to fix this for Travis. Is that the same thing? If so, it isn't written down anywhere that I can find. 

And secondly, this is a YAML file. I need to inject in the docker-compose (and remember, I don't have docker) but AFAICT docker will only allow a single regular command, not a bash script. It's probably possible to derive an environment variable in the wrapping shell script? Maybe? 

This is an undocumented gloopy config setup which I cannot figure out.

inodb:
@morungos the end-to-end test `docker-compose.yml` file is used on Travis. So that is the same. I can fix the end-to-end test if you want, I got docker set up on my laptop. You can run it as a single command by doing something like `bash -c 'comandaslala && morelala'`. And yes, you're right, that config file should probably be better documented.

Heroku's config is slightly different using the `pom`'s `heroku` profile to copy the `heroku/` folder to root. Those files there are pretty standard for a Heroku set up.

Both the `end-to-end` test and `heroku` use the same `webapp-runner.jar`, which is some Jetty style web server.

inodb:
One thing with adding a `scripts-VERSION.jar` is that we will have to update the documentation on how to use that. It will probably require @zheins to update some of the import pipelines etc to use that `scripts/target/scripts-VERSION.jar` file instead of the `cbioportal.war`. Is that a trivial change @zheins ?

morungos:
Hmm, it's still a terrible mess. docker-compose doesn't allow multiple commands, and there's no way to wrap as it's being run directly. We'll have to do some long bash -c and hope. 

Personally, I'd prefer there being a `travis` profile. That makes sense, as, e.g., we might use a different debugging mode, or minify JS differently.

inodb:
It's not that bad. You can prolly pass the `final.war.name` as an env variable and use that in the command. That's a decently clean solution imo. I'll extend on this PR.

In any case most of this stuff should be moved out at some point anyway. Once we have separate cbioportal frontend and backend repos, there would prolly be one repo that combines them. That's where the end-to-end tests should live.

It could make sense to have a `travis` profile. At this stage though I think the components are to much stitched together. Once we have a cbioportal-backend repo we can add a `travis` profile there. The front-end repo will prolly not even use `mvn`.

inodb:
@morungos One question. Why does it make a `scripts/target/original-scripts-1.3.0-SNAPSHOT.jar` and a `scripts/target/scripts-1.3.0-SNAPSHOT.jar`?

inodb:
This is a fix for the `end-to-end` test: https://github.com/cBioPortal/cbioportal/pull/1550/commits/cb37763005ed3b218b7a98702dfe056bd9164bbb. The env variable solution didn't work out because you can't run bash commands from there. It is possible to pass variables to docker-compose, but I think that would only make it more difficult to follow.

We will probably move the end-to-end tests to the cbioportal repo that combines both cbioportal-frontend and cbioportal-backend, so I don't think it's a huge deal atm that this set-up is a little hard to follow. Nobody's running the end-to-end tests locally anyway afaik.

One weird thing though is that I somehow see the drugs & clinical trial tabs appear in the screenshot. Don't see how this is related to this PR
screen shot 2016-07-27 at 2 52 42 pm

morungos:
OK, done. But I'm still not deploying clean like I used to, although the changes I've faced are more to with runtime jar dependencies. Still testing those for now.

morungos:
OK. Still doesn't work. No idea why. Timeouts aren't the most helpful error messages

inodb:
Yeah it's because the docker-compose is running in the background which spits out most of the log. For that debugging it's easier to run the end-to-end test locally, since you can run docker-compose in one shell to see the output and the screenshot test script in another.

What you can see in the log is the image diffs. Somehow the size is slightly different. No idea why: 

http://rawgit.com/cBioPortal/cbioportal/e470fb18cde93cf9460f063a5839086f6b89e482/test/end-to-end/image-compare/index.html?img1=https://raw.githubusercontent.com/cBioPortal/cbioportal/e470fb18cde93cf9460f063a5839086f6b89e482/test/end-to-end/screenshots/chrome/patient_view_lgg_ucsf_2014_case_id_P04.png&img2=https://clbin.com/AE7p9n.png&label1=rc&label2=e470fb1&screenshot_name=test/end-to-end/screenshots/chrome/patient_view_lgg_ucsf_2014_case_id_P04.png

morungos:
Yeah, I'm on a Mac. No idea how to set up end-to-end testing with docker.

inodb:
Me too, you install the latest version of docker and then all you have to do is compile cbioportal with the public db:

https://github.com/cBioPortal/cbioportal/blob/master/.travis.yml#L108-L116
and run this:
https://github.com/cBioPortal/cbioportal/blob/master/.travis.yml#L122-L128

I'll try to have a look as well what's going on

morungos:
Hmm, finally managed to get that working (I think, this is my first ever experience with docker) but it all looks (a) like noisy errors, and (b) like it works right anyway, except for a false failure.

inodb:
I got it working locally. It is just terribly slow on Travis when running the command to extract the version number. Therefore I changed the code to get the version number using the `mvn` version outside of the container and pass it as a environment variable instead. 

https://github.com/cBioPortal/cbioportal/pull/1550

inodb:
@morungos Yeah all the errors are basically related to the timeout being too short. This is why we added to step to iteratively increase timeouts. Normally you would probe for some element to appear on the page, but it would be a pain to change all the pages to output some final html element.

morungos:
OK, I'll merge these changes -- didn't even know about the `.travis.yml` since Eclipse hides it :-o

morungos:
Actually tried a different approach, now I know about `.travis.yml` I can just set the war name there and no version fudging should be needed. Helpfully, this is only for travis.

morungos:
So this is now clean apart from one popup image position. I'm calling ready to merge

morungos:
The war doesn't yet spin up. I think we're missing some runtime dependencies in Spring security. Testing runtime deployment now.

inodb:
@morungos Nice yeah just setting `-Dfinal.war.name` works as well ofc

How are you testing the runtime deployment? Is there anything I should test?

morungos:
It now works for me. Except not for OpenID Connect, but that's a wider issue, as there's no dependencies for that upstream. However, I do get to the main page. 

If I was feeling particularly brave I'd be able to spin up a full Ansible deploy pointing to the public DB now. As it is, I can get a VirtualBox up and running with about four commands, git clone, checkout, and an ansible one. 

This does change the output side, and that could use a test -- checking that imports work using the new scripts jar.

inodb:
@morungos Ok! Are you going to try and write that test as well or will you leave that for now?

inodb:
I can at least manually try to import something

morungos:
We could use a Travis test for that too, I guess. We should really have one now

inodb:
@morungos: I just heard from @n1zea144 that @angelicaochoa is working on making a separate importer using spring batch, so I'm guessing the whole script jar will be deprecated in a few months. Maybe it's therefore not super important to make a test for that now. Also not entirely sure if it makes sense to move the scripts jar out at this stage & update all the documentation if we are going to replace it in a few months anyway. What do you think? Should we just leave the scripts jar in for now and merge the other changes you made?

morungos:
@inodb I'm not sure that is a good reason not to break it into a new component. It'll need to be by the anyway, so that is a useful transition. Unless importing is going to be through RESTful APIs directly into the web app itself. 

And even so, I do actually need to import stuff, and holding off de-bloating because we're going to be adding a new import process in a few months feels wrong to me.

inodb:
@morungos Ok sounds good. Then for this PR:

1. a test for the import jar is missing. Not sure if you have time to add this or not?
2. we haven't tested whether this configuration actually works at mskcc production
3. the documentation for importing needs to be updated.

Maybe @zheins has a better idea than me about (2), could you have a look?

pieterlukasse:
@inodb trying to catch up with this one. Can you recap how this PR affects the importing documentation (nr 3 in your list)?

inodb:
@pieterlukasse Basically there is a separate scripts jar now, so I assume one has to use that one now instead of the core jar. Therefore I thought the docs need to be updated. Just double check that's correct. We also changed the `cbioportal.war` to have the version in the war name, so there could be some things in the docs that need to be changed accordingly.

@zheins mentioned he would be able to test this PR and subsequent ones for MSKCC production

morungos:
That is indeed the plan. I've been working on a deployment system (see: https://github.com/pughlab/cbioportal-deployment) which uses ansible, and as part of this, I can now (I think, not yet entirely tested) push a new study and import it remotely over ansible. The ansible role `import` does depend on some version files created during initial deployment. 

We'll be using something like this here, as it allows us to manage the system without logging directly onto the systems.

There's also a stub for a war overlay, which I'm polishing now,

n1zea144:
@morungos I'm trying to catch up and understand the work you've been doing here.  I'd be interested in learning more about your deployment system, use of ansible, and how we may be able to adopt some of your work @MSK.  Do you have the time and would you be willing to present some of this work on a core developer meeting or supplemental hangout?  As you know, @angelicaochoa has starting to rework the scripts package using a spring-batch framework which works similar to the current metaimport code (point the importer at a directory with staging files and run), but I haven't thought much about the upstream stage/system which places the desired files into a staging area for import.  Thoughts?

morungos:
I'd be happy to present an overview and talk through the moving parts. I
think the approaches may well be complementary, but the idea of using
Ansible to import wasn't part of the original plan, but it's very helpful
for us, as it will allow us to import remotely much more easily.

-- 
Stuart Watt

zheins:
I just imported our clinical series dataset using our importer pipeline using this branch and the new scripts jar - everything looks fine to me.

inodb:
@zheins @morungos @n1zea144 I misunderstood that this creates a separate `scripts.jar` w/o removing the old importer code from `core.jar`. I thought it removed that code from `core.jar` as well, so this means we don't have to update the docs to merge. That means pretty much everything is done. OK if I go ahead with merging?

pieterlukasse:
@inodb maybe the importer code needs to change at https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/scripts/importer/cbioportalImporter.py#L317 to search for `core.jar` instead of `core-*.jar`?

inodb:
@pieterlukasse The core jar still has the version in its name (e.g. `core-1.2.4-SNAPSHOT.jar`). Just the cbioportal war now also includes the version.

n1zea144:
@morungos can you rebase and then I'll merge?

morungos:
Attempted to rebase. I hope it worked. I'm not a wizard with git, more a street conjuror.

portal/pom.xml
```diff
@@ -41,18 +41,48 @@
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.tomcat</groupId>
+      <artifactId>tomcat-catalina</artifactId>
+      <version>7.0.68</version>
+      <scope>runtime</scope>
```
jjgao: Including tomcat-catalina dependency broke the deployment to tomcat. 

http://stackoverflow.com/questions/14028476/java-web-api-conflict-using-maven

I'll make a quick fix to set the scope to provided.

morungos: Sounds good. It'll break Jetty, but I can handle that. As I hinted, I don't think we should use Tomcat for this anyway, as it locks the war to Tomcat. There are small and simple CORS filters we can use independently of the container.

PR #1473, opened 2016-07-11
Fixed wiki build documents

ecerami:
# What? Why?

Documentation changes to fix the build process.

Fixes #1449, #1447, #1446

Changes proposed in this pull request:

documentation changes

# Notify reviewers
Notifying @jjgao

PR #1524, opened 2016-07-21
Update Deploying.md

oplantalech:
As proposed by @pieterlukasse and me in #1473.

### DIFF
--- a/docs/Deploying.md
+++ b/docs/Deploying.md
@@ -68,7 +68,9 @@ Lastly, open a browser and go to:
 
 ## Important
 
-Each time you add new data or modify any code, you must redeploy the WAR file.
+- Each time you modify any java code, you must recompile and redeploy the WAR file.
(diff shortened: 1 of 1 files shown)


pieterlukasse:
@morungos at some point, once your PRs are merged, I think changing portal.properties will not require a new compilation step anymore, right?

fedde-s:
Would it be simpler to say something like this, putting both reasons to redeploy on one line and hyperlinking to the docs on portal.properties?
```markdown
- Each time you modify any [configuration properties](portal.properties-Reference.md) or Java code, you must recompile and redeploy the WAR file.
```

Although really, portal.properties-Reference.md should link to Customizing-your-instance-of-cBioPortal.md then.

pieterlukasse:
@fedde-s I prefer the separate lines. I like the reference link, let's add that :+1:

Issue #1351, opened 2016-06-10
508 compliance for Study view


pieterlukasse:
Issue: we're trying to make the main Study view 508 compliant. The checkboxes in the CNA and Mutation tables are not compliant due to them not having a label (see screenshot). 
- [x] One easy solution is to give them the title attribute, which suffices for 508

pieterlukasse:
@fedde-s can you add the 508 Wave tool screenshot here showing the errors mentioned above?

fedde-s:
![screenshot_study_view](https://cloud.githubusercontent.com/assets/4929431/15963232/a44a5bac-2f10-11e6-8ec9-fcaa9f1ee5b5.png)

pieterlukasse:
@zhx828 so I am specifically referring to the checkboxes below in my copy of the screenshot:

![image](https://cloud.githubusercontent.com/assets/2900303/15964900/1ea10da6-2f1b-11e6-8a03-79c7ff32517a.png)

But if you could proactively add labels to the other items as well that would be great :smiley:

fedde-s:
These fixes have been merged and solved, and the study view has been further made Section 508-compliant as part of pull requests #1494 and #1497.

PR #1530, opened 2016-07-22
Introduce OncoKB Card for OncoKB integration under annotation column

zhx828:
# What? Why?
Improve OncooKB annotation integration in cBioPortal

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

screen shot 2016-07-22 at 12 13 28 am

screen shot 2016-07-22 at 12 13 39 am

# Notify reviewers
@cBioPortal/frontend


zhx828:
@fedde-s @pieterlukasse I may need to make 508 compliance working well in OncoKB template. Could you tell me which attributes I need to put in and on which HTML tag? Thanks.

pieterlukasse:
@zhx828 you can check some of our PRs and tickets like #1351, and https://github.com/cBioPortal/cbioportal/pulls?q=is%3Apr+508+is%3Aclosed . In #1351 you can also find a screenshot of the WAVE tool that can be used for this. This tool will tell you what needs to be fixed and how. 

@fedde-s can probably give you a summary of most common changes he had to make.

zhx828:
@fedde-s when you have time, could you take a look the 508 compliance commit? Thank you.

fedde-s:
I do not have time to find out how to configure OncoKB integration in a local deployment and run the WAVE tool on it by today, but by eye I do not see any newly introduced input elements without labels or title attributes, nor images missing an alt text that conveys the information in the image, so that looks fine. The column header scope declarations may not even be strictly necessary in such a simple table where the `<th>` tags are in the first row, but being explicit doesn’t hurt 👍

One small question: do you ever expect the variables `title` in `oncokb-card-template.html` and `_datum` in the method `OncoKB.utils.findRegex()` to be empty strings? That could result in links without text.

zhx828:
@fedde-s `OncoKB.utils.findRegex()` handles empty string. And I think it's users' decision whether they want to use empty title.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -121,7 +121,7 @@ tcga_path_report.url=https://tcga-data.nci.nih.gov/tcgafiles/ftp_auth/distro_ftp
 segfile.url=http://cbio.mskcc.org/cancergenomics/gdac-portal/seg/
 
 # Enable OncoKB annotation
-oncokb.url=
+oncokb.url=http://oncokb.org/legacy-api/
```
jjgao: you can also add some code to GloabalProperties.java to use this url by default if oncokb.url is empty of null.

jjgao: maybe also add a new line:
`use_oncokb=true`
Unless it is set to false, we always display oncokb.

PR #1533, opened 2016-07-23
Annotate mutations in oncoprint with prior knowledge, color fusions separate from mutations, and color promoter mutations orange

adamabeshouse:
Allow users to highlight mutations in the oncoprint using prior knowledge:
- cBioPortal count (with user-specified threshold)
- COSMIC count (with user-specified threshold)
- Is it a hotspot, according to cancerhotspots.org
- Is it Likely Oncogenic or Oncogenic according to OncoKB

Additionally, we now handle fusions separately from mutations, so that a fusion and mutation can be shown on the same case.

Finally, we now handle promoter mutations as a separate mutation category in OQL and in the oncoprint.

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Screenshot
image

image

image

Notify: @cBioPortal/frontend @schultzn @jjgao


schultzn:
The screenshot looks great!!

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -195,26 +220,83 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
+	    var query_alterations = Object.keys(oncogenic[query_genes[i]]);
+	    for (var j=0; j<query_alterations.length; j++) {
+		queries.push({'hugoSymbol': query_genes[i], 'alteration': query_alterations[j]});
+	    }
+	}
+	// Execute query
 	var query = {
 	    "geneStatus": "Complete",
 	    "source": "cbioportal",
 	    "evidenceTypes": "GENE_SUMMARY,GENE_BACKGROUND,ONCOGENIC,MUTATION_EFFECT,VUS,STANDARD_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_SENSITIVITY,STANDARD_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_RESISTANCE,INVESTIGATIONAL_THERAPEUTIC_IMPLICATIONS_DRUG_SENSITIVITY",
```
zhx828: Could you change evidenceTypes to 'ONCOGENIC' since you only need oncogenicity info？

jjgao:
@adamabeshouse could you merge with the latest code?

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -86,6 +93,27 @@
 		return cancerTypeMapper.getCancerTypes(cancer_type_ids);
 	}
 
+	@Transactional
+	public List<CosmicCount> getCOSMICCountsByKeywords(List<String> keywords) {
+		return cosmicCountService.getCOSMICCountsByKeywords(keywords);
+		/*
```
adamabeshouse: this is a comment, should be deleted

business/src/main/java/org/mskcc/cbio/portal/model/DBMutationData.java
```diff
@@ -32,4 +34,6 @@
     public String variant_read_count_tumor;
     public String reference_read_count_normal;
     public String variant_read_count_normal;
+    public String cosmic_count;
```
adamabeshouse: this field is not written to server-side, should be deleted

adamabeshouse:
@n1zea144 would you please review the new service and tests?

adamabeshouse:
@kalletlak could you please review the datamanager changes, since you're familiar with it now?

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -245,9 +328,118 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
+		    datum[attribute_name] = count_records.map(function(count_record) { return parseInt(count_record.count, 10); })
+							.reduce(function(x,y) { return x+y; }, 0);
+		}
+	    }
+	    def.resolve(webservice_data);
+	});
+	return def.promise();
+    };
+    
+    var annotateHotSpots = function(webservice_data) {
```
kalletlak: Why don't we make it as an abstract utility function, so that it can be invoked from other views if we require the same functionality.

adamabeshouse: Maybe we can wait until a need arises? Ideally this would handle most use cases since the views are going to pull data from the data manager.

adamabeshouse:
Moved to https://github.com/cBioPortal/cbioportal/pull/1588 which has commits cleaned up.

Issue #1535, opened 2016-07-24
new code of customized sample query broke jsp in tomcat 7


jjgao:
![image](https://cloud.githubusercontent.com/assets/840895/17081371/a6755938-5123-11e6-8f57-76d5d77bc101.png)

![image](https://cloud.githubusercontent.com/assets/840895/17081376/d6863890-5123-11e6-9cb5-59169b45a776.png)

jjgao:
@fedde-s I couldn't find a proper solution, so I revert part of your code in #1536. Please make the proper fix when you have time. Thanks.

fedde-s:
I cannot reproduce the error in Tomcat 7.0.70 as packaged in https://hub.docker.com/_/tomcat/ , either under JDK 8u91 and JDK 7u101.

jjgao:
That's strange. Maybe it's not related to tomcat 7. It worked on Heroku too, but did not on either our server or my Mac.

fedde-s:
Since @jjgao merged the pull request into the hotfix branch over a week ago and I have heard no further complaints, I will assume that the issue has been solved. If not, please do re-open it.

PR #1538, opened 2016-07-25
Simplify JSP EL code quoting in study query list

fedde-s:
The template implementation of the feature introduced in commit 172e1a170fdd086af48a1b853c59a06cb7c84eb6 resulted in exceptions on certain server-side installations and was disabled. This patch reintroduces the feature and cleans up the quoting of JavaScript strings in the new code, which is the likely cause of the issues, by using a tag from the Spring library.


fedde-s:
@jjgao, could you check if this fixes issue #1535 for you?

Issue #1379, opened 2016-06-16
3D view button does not work for some genes, even when PDB data is available


pieterlukasse:
Steps to reproduce:

[...]

PR #1540, opened 2016-07-25
update mutation mapper library

### LINKED ISSUE #1276: Mutations table should also show how many samples are mutated
In the example below, the text above the table should be:

Showing 10 mutation(s) in 8 sample(s)

http://bit.ly/1qFC7yF

onursumer:
# What? Why?
Fix #1276, #1379, and #1617.

See https://github.com/cBioPortal/mutation-mapper/pull/52 for details.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers


portal/src/main/webapp/js/lib/mutationMapper.js
```diff
@@ -16020,6 +16061,24 @@ function MutationDetailsTable(options, gene, mutationUtil, dataProxies, dataMana
 		}
 	}
 
+	function uniqueSamples(rowData)
+	{
+		var samples = {};
+
+		_.each(rowData, function(data, index) {
+			// assuming only the first element contains the datum
+			var mutation = data[0].mutation;
```
zhx828: You also assume data will always be array?

onursumer: Actually assuming that we are passing an array of row data to this function. And the data is retrieved by this API function: https://datatables.net/reference/api/data()

portal/src/main/webapp/WEB-INF/jsp/mutation/mutationMapperTemplates.html
```diff
@@ -351,69 +351,65 @@
...
 </script>
 
 <script type="text/template" id="mutation_table_cna_template">
-	<label alt='{{cnaTip}}' class='simple-tip {{cnaClass}}'>{{cna}}</label>
+	<span alt='{{cnaTip}}' class='simple-tip {{cnaClass}}'>{{cna}}</span>
 </script>
 
 <script type="text/template" id="mutation_table_cosmic_template">
-	<label class='{{cosmicClass}}'>{{cosmicCount}}</label>
+	<span class='{{cosmicClass}}'>{{cosmicCount}}</span>
```
pieterlukasse: do we need to add the `alt` attribute here to keep it 508 compliant? @fedde-s can you comment on this?

jjgao:
@onursumer what's the status of this?

onursumer:
@jjgao : As far as I tested, there is no functional issues. It should be safe to merge.

portal/src/main/webapp/WEB-INF/jsp/mutation/mutationMapperTemplates.html
fedde-s: @pieterlukasse: Assistive technologies will parse `<label>` tags as labels for input elements/form controls. All such inputs require a label tag for accessibility and Section 508 compliance, but I believe that these spans here are not actually labels for inputs. If that’s correct, removing the `<label>` tags actually improves accessibility.

The `alt=` attribute has no meaning for other elements than `` (it’s just sometimes confused with `title=`), and in this file it’s only used as a custom attribute to be read by the qTip library. That’s no issue for accessibility. 👍

fedde-s:
Running a quick accessibility scan shows only five Section 508 violations (accessibility issues) in the views affected by this PR, but these were not introduced by this PR. They are related to the network tab, such as the textless buttons introduced in commit 153f60bd0598c81a72b969f210d8d4aaf9a92419, which need a textual alternative for the icons, for instance in a Bootstrap `<span class="sr-only">`.

PR #1562, opened 2016-08-01
Move MutationDataServlet and CrossCancerMutationDataServlet into service module

sheridancbio:
Part of the rearchitecture sprint 3
MutationDataServlet (getMutationData.json) and CrossCancerMutationDataServlet (crosscancermutation.json) are moved into the service module and provide data for two new Web API endpoints:
- api/mutationsforstudiesandgenes
- api/mutationsforprofilesandgenes

Output from services is converted by Jackson to JSON (rather than older servlets use of org.json.simple library). Escaping of forward slash in json output is now dropped.

javadoc comments and unit tests added
# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.
# Notify reviewers

everybody in the core team:
@n1zea144 @ersinciftci @ecerami


sheridancbio:
This rearchitecture approach was abandoned. Instead of migrating the existing servlets to provide services through the webAPI, the existing servlets and .jsp code was left in place and a strangler-vine approach was adopted, recreating the front end page by page without dependencies on the original servlets, and with expansion to the new webapi as needed to service the new development. At some (unscheduled) future time, any remaining dependencies to the core module are to be cut, and the core module is to be discarded.

Issue #1213, opened 2016-05-10
Validator improvement : Allow only 'LIVING' and 'DECEASED' for OS_STATUS


priti88:
Validator should throw a warning if OS_STATUS is not 'LIVING' or 'DECEASED'.
The survival plot shows up as empty plot otherwise.

pieterlukasse:
@fedde-s an extra requirement: give error if fields known to be PATIENT_ATTRIBUTES (like OS_STATUS) are found in SAMPLE_ATTRIBUTES file and vice-versa.

fedde-s:
This should be fixed now that aa9a48cb1d5601a840267aabc86f5bc6482edd5a is in v1.3.1.

PR #1567, opened 2016-08-03
Add validation for clinical attrs with set meanings

fedde-s:
Certain clinical attributes have meanings coded into the frontend, as documented on [the page on file formats](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#the-patient-file). This pull request adds the assumptions made in the code to the validator, so that loaded studies will not configure the database to contradict them (i.e. by defining sample-level attributes as patient-level ones or using unsupported values for `OS_STATUS` or `DFS_STATUS`, see also issue #1213)

In addition, it corrects the names of three of these attributes both in the validator and the frontend to match what has been seen in data, adds validation that colours defined for cancer types are colours that will be understood in CSS by web browsers, and makes it possible to run the validation script through a symbolic link (a practice relatively common on Unix-like systems) by resolving the link when locating the Jinja template in the same directory.

- Validate that cancer type colors are web colors
- Resolve symlinks in the path to the script when locating the Jinja template for the validator’s output
- Add validation for correct definition of clinical attributes with set meanings in the frontend
- Replace punctuation in three of these attributes to match what is found in [the API output of the main public cBioPortal](http://www.cbioportal.org/api/clinicalattributes/samples)
- Add validation for correct values of clinical attributes with set meanings in the frontend

### Suggested reviewers ###
- @zheins 
- @inodb

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1962,6 +1962,8 @@ class CancerTypeValidator(Validator):
 
     REQUIRED_HEADERS = []
     REQUIRE_COLUMN_ORDER = True
+    # check this in the subclass to avoid emitting an error twice
+    ALLOW_BLANKS = True
 
     COLS = (
         'type_of_cancer',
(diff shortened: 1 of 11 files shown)


fedde-s:
@zheins, @inodb: could you still take a look at whether these additional validations make sense to you?

fedde-s:
Travis is currently failing because of the bug that should be fixed by pull request #1664.

fedde-s:
The screenshot test was broken by unrelated bugs in rc again.

inodb:
Hi @fedde-s. I think the logic implemented is correct. Just wondering if this is the best way to do it. For now it's probably fine, but I'd be surprised if there's no language agnostic way for specifying data vocabularies of csv/tsv files. If something like that exists it could go a long way in defining our own clinical data spec.

fedde-s:
@inodb: I think what I specified in the python code is almost literally the subset of fields that must be returned by the JSON web API for correct data, which I would say is the closest thing to a standard textual representation of the data in cBioPortal. It may make sense to eventually move this to a separate file in true JSON format, as it does not really make the validation script less self-contained than it is now. That would be pretty easy to do after this is released though, as it does not affect the functionality.

inodb:
@fedde-s I guess you are referring to what attributes are patient and which are samples? That part looks fine to me.

I meant the controlled vocabulary for `OS_STATUS` and `DFS_STATUS`. I would expect there are some ways to define these types of checks for a tsv/csv file in a DSL. I see for instance there is csv-schema: https://github.com/digital-preservation/csv-schema. If there is some module that implements csv-schema, all you would have to do is write the schema and handle the validation errors accordingly. I have no clue if csv-schema is any good, but if you find a decent standard for this, there are two advantages compared to hardcoding this logic yourself:

1. DSLs can be more readable, since they were made specifically for that task. If it's easily comprehensible, it might even do as a spec for a file format.
2. If the DSL is common, there are usually implementations in multiple languages. That way I could for instance easily make a JS validator or Java validator.

fedde-s:
Ah, I agree, that would be an improvement! Either way, that sounds like something to discuss for a later pull request, it probably applies to validation of other file formats as well (e.g. segment-wise CNA data and mutation data).

zheins:
@inodb I think @n1zea144 and I discussed using something like what you are describing, but at the time the idea was to keep the validator as simple as possible and to not introduce dependencies that users would need to install. But I do think that would simplify much of the logic in the validator and make it more readable. Maybe the next iteration can use something like that.

This PR looks good to me though @fedde-s.

fedde-s:
Thanks for the reviews! I rebased to update the branch with rc. @jjgao, can this be merged?

jjgao:
@fedde-s sorry missed this on for 1.3.0 release. Will release in 1.4.0.

sandertan:
related to #1015?

fedde-s:
@sandertan: Yes, pull request #1015 did re-implement some of the frontend code that depends on the meaning of these attributes, and it added one attribute to the list (`TUMOR_TISSUE_SITE` as a synonym of `SAMPLE_TYPE`).

Issue #1214, opened 2016-05-11
Validator should give error when no case list option is present


pieterlukasse:
Currently it is possible to load a study without case lists. However this gives an exception in the portal. 
![image002](https://cloud.githubusercontent.com/assets/2900303/15186387/20e54612-179d-11e6-992e-dd3ccf3caf21.jpg)

At the study in the UI (selecting a few genes) to see oncoprint, the following message is given:
![image004](https://cloud.githubusercontent.com/assets/2900303/15186385/1fd69000-179d-11e6-8c8b-db79b42ddeee.jpg)

pieterlukasse:
solution would be to give error if no case lists are found `AND` the study metadata parameter `add_global_case_list` is false or not given.

fedde-s:
Validation for the presence of the case list `<study_id>_all` has been implemented in the referenced pull request and released as part of v1.2.5. @pieterlukasse, does that solve your issue?

fedde-s:
I will assume it does and close this issue.

PR #1571, opened 2016-08-05
Enable validation of required case list `_all`

fedde-s:
A bug led to the validator sometimes failing to emit an error if this
case list is missing in a study, resulting in a server-side exception
(HTTP Status 500) when loading the OncoPrint. See issue #1214.
This commit fixes the bug and adds a unit test to prevent further
regressions.

In addition, fix the exception that occurs if a referenced data file cannot be found, and add a licence notice to the top of the file.

### Reviewers ###
@cBioPortal/core-developers

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -314,7 +314,12 @@ class Validator(object):
 
         self.logger.debug('Starting validation of file')
 
-        with open(self.filename, 'rU') as data_file:
+        try:
+            opened_file = open(self.filename, 'rU')
+        except IOError:
+            self.logger.error('File could not be opened')
+            return
+        with opened_file as data_file:
 
             # parse any block of start-of-file comment lines and the tsv header
             top_comments = []
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1,9 +1,32 @@
 #!/usr/bin/env python2.7
 
-# ------------------------------------------------------------------------------
-# Data validation script - validates files before import into portal.
-# ------------------------------------------------------------------------------
+#
+# Copyright (c) 2016 The Hyve B.V.
+# This code is licensed under the GNU Affero General Public License (AGPL),
+# version 3, or (at your option) any later version.
+#
 
+#
+# This file is part of cBioPortal.
+#
+# cBioPortal is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


fedde-s:
@pieterlukasse and @zheins, could you review this?

core/src/main/scripts/importer/validateData.py
```diff
@@ -2437,21 +2437,30 @@ def processCaseListDirectory(caseListDir, cancerStudyId, logger,
...
+
+    Args:
+        cancer_study_id (str): the study ID to be expected in the stable IDs
+        case_list_fns (Iterable[str]): stable ids of defined case lists
+        file_types (Dict[str, str]): listing of the MetaFileTypes with high-
+            dimensional data in this study--these may imply certain case lists
+        logger: logging.Logger instance to log output to
+    """
+
+    if cancer_study_id + '_all' not in case_list_fns:
```
pieterlukasse: shouldn't this if include a check on `add_global_case_list` as well? I.e. something like "if cancer_study_id + '_all' not in  case_list_fns **and** (`add_global_case_list` not set *or* set to false), then give error.

fedde-s: No, in that case this dictionary will list that the case list with suffix `_all` was defined in the study metadata file.

fedde-s:
@pieterlukasse: I clarified the parts of the code that confused you, added a licence notice to the top of the file, and fixed the unrelated exception that we discussed.

fedde-s:
Travis is currently still failing because the test that was fixed in rc by PR #1657 was never fixed in hotfix.

jjgao:
Since there is no website code, I'll just merge it when it's ready. No worries about the screenshot tests. Please label it as "ready to merge" when ready.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2771,11 +2771,12 @@ def validate_study(study_dir, portal_instance, logger):
...
             validator.validate()
 
-    # finally validate case lists if present
+    # finally validate the case list directory if present
     case_list_dirname = os.path.join(study_dir, 'case_lists')
     if not os.path.isdir(case_list_dirname):
         logger.info("No directory named 'case_lists' found, so assuming no custom case lists.")
     else:
+        # add case lists IDs defined in the directory to any previous ones
         defined_case_list_fns = processCaseListDirectory(
```
pieterlukasse: maybe call this one defined_case_list_ids to keep code consistent?

pieterlukasse:
@fedde-s :+1: only one small comment.

core/src/main/scripts/importer/validateData.py
fedde-s: No, this is a dictionary that maps IDs to filenames.

pieterlukasse:
@jjgao ready to merge :+1:

fedde-s:
Now Travis is failing because of the frontend screenshot issue in pull request #1664.

fedde-s:
The Travis screenshot issues have been fixed upstream.

Issue #1249, opened 2016-05-17
Missing HGVSp_Short validation for mutation data


pieterlukasse:
Validation for column `HGVSp_Short` needs to be implemented to give an error if _more than one_ value is given in this column. 
- [x] part1 :if column value is `p.V600E` then it is OK. If column value contains a list of values like `p.V600E, p.V610E,etc`, then user should get an error. 
  Documentation in https://github.com/cBioPortal/cbioportal/blob/rc/docs/File-Formats.md#data-file-5 should also be adjusted to be a bit more explicit about this. 
- [ ] part 2: check how much work it would be to validate the column value using the hgvs package https://pypi.python.org/pypi/hgvs/  (consider updating installation documentation, Docker file, etc). Alternatively,  could this  also be solved with a simple regex check?

morungos:
@pieterlukasse @fedde-s I had a similar issue in a previous system. I think you'll want probably regex, as there is *huge* variation in compliant HGVS. Not least, you're allow three-letter codes for the amino acids, which I have no idea whether the portal will recognize that p.G12D is exactly the same as p.Gly12Asp (they should be merged statistically, to be strict). Either you'll need to canonicalize the formats on import, or just be stricter and allow only a subset. Genome Nexus could change this, as it's probably more a Genome Nexus problem.

pieterlukasse:
@morungos : thanks for the comments!

I would like to suggest the following approach: we fix the most urgent "part 1" first (this could be submitted as a PR on its own) and leave "part 2" for later. At this moment I'm not into the details of Genome Nexus, but I image we will still like to support the staging files and proper validation of these, regardless of how the storage and backend code of mutation data is implemented.

fedde-s:
@pieterlukasse A ‘value’ would probably be anything starting with the substring ‘p.’, not preceded by a letter. I suspect that the HGVS syntaxes for alleles involving multiple mutations, as in `p.[S73R;N103del]` and `p.[S73R];[N103del]`, may not be supported either.

fedde-s:
As of v1.2.5, the validator will report an error if values contain commas, additional `p.` prefixes, or allele syntaxes, as implemented by the pull request that referenced this issue. @pieterlukasse, can the issue be closed now?

pieterlukasse:
@fedde-s I like the solution in the PR and the extra information you added 👍  

One point that we could work on before closing this is to update the documentation regarding this. 
- Does the documentation clearly state how this field should be filled? 
[...]

PR #1572, opened 2016-08-08
Correct and extend validation of MAF file AA change

fedde-s:
Fix for the main problem in issue #1249.

Since the file format represents mutations one per line, delimited lists
of multiple variants (occasionally generated by mistake and even
supported by HGVS notation with its allele syntax) are not supported in
the amino acid change field. Commas are also used internally in
`GeneticAlterationUtil.getMutationMap`, so are not allowed within values.

In addition, fix validation of blank fields in files that have
both the `HGVSp_Short` and `Amino_Acid_Change` columns.

- Make the validator check if at least one of the columns `HGVSp_Short` or `Amino_Acid_Change` is non-blank, non-NA and non-NULL, to match what the loading code in `ExtendedMutationUtil.getProteinChange` does
- Reject protein mutation specifications with commas in them (these are unsupported by `GeneticAlterationUtil.getMutationMap`) and values that contain the `p.` prefix multiple times or use the HGVS notation for multi-variant alleles, as aggregating multiple mutations onto one line in a mutation data file is not supported.
- Slightly simplify the surrounding code by removing an unnecessary argument from several methods

Example of the new output, from the unit test:
```
WARNING: mutations/data_mutations_wrong_aa_change.maf: line 2: No Amino_Acid_Change or HGVSp_Short value. This mutation record will get a generic "MUTATED" flag
ERROR: mutations/data_mutations_wrong_aa_change.maf: line 3: column 37: Unexpected 'p.' within amino acid change, only one variant can be listed on each line; value encountered: 'p.A195V p.I167I'
ERROR: mutations/data_mutations_wrong_aa_change.maf: line 5: column 37: Comma in amino acid change; value encountered: 'p.N851,Y1055delinsCC'
ERROR: mutations/data_mutations_wrong_aa_change.maf: line 6: column 37: Unexpected ';' or '+' in amino acid change, multi-variant allele notation is not supported; value encountered: 'p.[N851N];[Y1055C]'
WARNING: mutations/data_mutations_wrong_aa_change.maf: line 8: No Amino_Acid_Change or HGVSp_Short value. This mutation record will get a generic "MUTATED" flag
```

### Reviewers ###
@cBioPortal/core-developers


fedde-s:
@onursumer, @n1zea144: could you review if this amendment to the data format validation makes sense to you?

PR #1574, opened 2016-08-09
Added section about configuring groups to Authorization documentation

pieterlukasse:
# What? Why?
Added *missing* section about configuring groups to Authorization documentation. 

Changes proposed in this pull request:
- added documentation that was previously missing for existing documentation

# Notify reviewers
@ecerami, @priti88

PR #1361, opened 2016-06-14
Authentication and authorisation documentation improvements

pieterlukasse:
# What? Why?
this improves existing documentation on SAML authentication method and *adds* documentation on LDAP authentication.

@n1zea144 : maybe you can check this one?

PR #1575, opened 2016-08-09
Fix to make the group authorization configuration case insensitive

pieterlukasse:
# What? Why?
Fix to make the group authorization configuration case insensitive. Similar fix was previously also made for another part of the authorization configuration in #1361. This fix makes another part (the groups configuration) consistent with this behavior. 

This fix was made when I found this issue while improving documentation in #1574

- Fix to make the group authorization configuration case insensitive (so the comparison between `cancer_study.GROUPS` and `authorities.AUTHORITY` configuration values in DB is now case insensitive). 

# Notify reviewers
I suggest the same reviewers of #1574: @ecerami @priti88


core/src/main/java/org/mskcc/cbio/portal/util/CancerStudyPermissionEvaluator.java
```diff
@@ -175,7 +175,12 @@ private boolean hasPermission(CancerStudy cancerStudy, Authentication authentica
         } catch (DaoException e) {
             groups = cancerStudy.getGroups();
         }
-        if (!Collections.disjoint(groups, grantedAuthorities)) {
+        // to upperCase to make the comparison with grantedAuthorities case insensitive:
```
jjgao: Maybe cancerStudy.getGroups(); should return upper case only?

jjgao:
@pieterlukasse since this is a very small fix, could you point it to hofix?

pieterlukasse:
@jjgao is not very urgent for me. So if it is not urgent for you, I would
prefer to leave it as is and just merge in rc.

www.thehyve.nl
E pieter@thehyve.nl
T +31(0)30 700 9713
M +31(0)6 28 18 9540
Skype pieter.lukasse


We empower scientists by building on open source software

2016-08-10 18:24 GMT+02:00 JianJiong Gao <notifications@github.com>:

core/src/main/java/org/mskcc/cbio/portal/util/CancerStudyPermissionEvaluator.java
pieterlukasse: I wanted to avoid changing  cancerStudy.getGroups() as it could be used somewhere else.

pieterlukasse:
moved to #1600

pieterlukasse:
closed #1600 and reopened this one again.

jjgao:
@pieterlukasse could you update?

core/src/main/java/org/mskcc/cbio/portal/util/CancerStudyPermissionEvaluator.java
jjgao: I don't think it's used in other places. Even if it is, they should have the same behavior -- case insensitive. I would suggest modify the SQL to return upper case only.

pieterlukasse:
@jjgao I moved the uppercase work towards the model layer. Not exactly in the SQL statement (since cancer studies can also come out of a study *file*), but pretty much what you intended I think.

PR #1581, opened 2016-08-10
Relaxed validator option (fixes #1563)

### LINKED ISSUE #1563: Relaxed processing of files by validator
Validator currently fails if the clinical files don't have meta data in them. 

The pipeline does that for us, so we are not manually adding them to our clinical files. On running the validator it always fails after the sample clinical file. Is there a way to make this a warning and allow it to finish the validation of all files.
screen shot 2016-08-01 at 3 35 31 pm

angelicaochoa:
# What? Why?
Fix #1563 . Added option to `validateData.py` to relax the validator to prevent fast failing during the validation process. 

- a
- b

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers

# To-Do
- [x] Apply change to `DataFileTestCase.validate()` in `core/src/test/scripts/unit_tests_validate_data.py`
- [x] Add unit test that validates headerless clinical data
- [x] Update doc strings
- [x] Update relaxed mode option name and description


core/src/main/scripts/importer/validateData.py
```diff
@@ -2328,6 +2360,51 @@ def process_metadata_files(directory, portal_instance, logger):
...
+#            
+#        elif 'patient' in f:
+#            meta_file_type = cbioportal_common.MetaFileTypes.PATIENT_ATTRIBUTES
+#            datatype = 'PATIENT_ATTRIBUTES'
+#            data_filename = 'data_clinical_patient.txt'            
+#            
+#        
+    
+    
+
```
fedde-s: Does this comment clarify anything? If not, please remove it.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2601,12 +2678,15 @@ def interface(args=None):
     parser.add_argument('-v', '--verbose', required=False, action='store_true',
                         help='report status info messages in addition '
                              'to errors and warnings')
+    parser.add_argument('-r', '--relaxed', type=str, required=False, 
+                        default='false',
+                        help='option to enable relaxed mode for validator')
```
fedde-s: I would prefer this description to be more specific on what errors in the data it will ignore. Maybe even a different name for this option, to clarify which requirements of the file format are being relaxed. Something like `--no-clinical-definitions`, maybe?

core/src/main/scripts/importer/validateData.py
```diff
@@ -344,12 +351,16 @@ def _validate_file(self):
                                      [header_line],
                                      delimiter='\t',
                                      quoting=csv.QUOTE_NONE,
-                                     strict=True).next()
+                                     strict=True).next()
```
fedde-s: This whitespace change does not look necessary.

core/src/main/scripts/importer/validateData.py
```diff
@@ -249,7 +249,7 @@ class Validator(object):
     REQUIRE_COLUMN_ORDER = True
     ALLOW_BLANKS = False
 
-    def __init__(self, study_dir, meta_dict, portal_instance, logger):
+    def __init__(self, study_dir, meta_dict, portal_instance, logger, relaxed_mode):
```
fedde-s: Please adjust `DataFileTestCase.validate()` in `core/src/test/scripts/unit_tests_validate_data.py` to this change in the constructor’s signature, as currently it makes every single unit test fail.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1312,12 +1331,24 @@ def checkHeader(self, cols):
...
+                self.logger.error(
+                    'Varying numbers of columns in clinical header (%d, %d)',
+                    len(self.attr_defs),
+                    len(self.cols),
+                    extra={'line_number': self.line_number})
+                num_errors += 1
+            
+        # fill in missing attr_defs data if in relaxed mode and clinical 
+        if self.fill_in_attr_defs:
+            self.logger.info('Filling in missing attribute properties for clinical data.')
```
fedde-s: I would strongly prefer if you could use attributes as they are defined in the database. Is it at all possible for the MSKCC pipeline to define whether attributes are strings or numbers, and whether they are attributes of patients or of samples, *before* specific data sets are validated?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1312,12 +1331,24 @@ def checkHeader(self, cols):
...
+            missing_attr_defs = {}
+            for col_index, col_name in enumerate(cols):                    
+                missing_attr_defs[col_index] = {'display_name': col_name,
+                                         'description': col_name,
+                                         'datatype': 'STRING',
+                                         'priority': '1'}
```
fedde-s: I would strongly prefer if you could use attributes as they are defined in the database. Is it at all possible for the MSKCC pipeline to define whether attributes are strings or numbers, and whether they are attributes of patients or of samples, *before* studies are validated?

core/src/main/scripts/importer/validateData.py
```diff
@@ -249,7 +249,7 @@ class Validator(object):
     REQUIRE_COLUMN_ORDER = True
     ALLOW_BLANKS = False
 
-    def __init__(self, study_dir, meta_dict, portal_instance, logger):
+    def __init__(self, study_dir, meta_dict, portal_instance, logger, relaxed_mode):
         """Initialize a validator for a particular data file.
 
         :param study_dir: the path at which the study files can be found
```
fedde-s: Could you document what the new parameter does in the docstring?

core/src/main/scripts/importer/validateData.py
```diff
@@ -2601,12 +2678,15 @@ def interface(args=None):
...
+                        help='option to enable relaxed mode for validator')                             
 
     parser = parser.parse_args(args)
     return parser
 
 
-def validate_study(study_dir, portal_instance, logger):
+def validate_study(study_dir, portal_instance, logger, relaxed_mode):
 
     """Validate the study in `study_dir`, logging messages to `logger`.
```
fedde-s: Could you document what the new parameter does in the docstring?

core/src/main/scripts/importer/validateData.py
fedde-s: The reason for this is that if you do set the datatype that will actually be used when loading, the values in the file can be validated accordingly and numeric attributes really be numeric (or null) values. Certain range sliders in the frontend, for instance, will break otherwise.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1370,15 +1401,16 @@ def checkHeader(self, cols):
...
+                        if not self.relaxed_mode and not self.fill_in_attr_defs:                            
+                            self.logger.warning(
+                                "%s definition for attribute '%s' does not match "
+                                "the portal, and will be loaded as '%s'",
+                                attr_property,
+                                col_name,
+                                srv_attr_properties[attr_property],
+                                extra={'line_number': self.attr_defs[col_index].keys().index(attr_property) + 1,
+                                       'column_number': col_index + 1,
+                                       'cause': value})
```
fedde-s: This will skip over an important validation if a data file does contain the attribute definitions while the flag is given.

core/src/main/scripts/importer/validateData.py
fedde-s: Also, it would be a good idea to add a function that a study without clinical attribute definitions in `core/src/test/scripts/unit_tests_validate_data.py`, as a regression test for the new option.

fedde-s:
I have added several comments on the changes, and I have one main question: what does the loader do if you use an attribute that isn’t in the database in a headerless file? Does it set the display name and the description to the column name, the data type to STRING, the priority to 0 and the type of attribute (patient or sample) to what the meta file specifies?

Should that scenario be reported as an error, because the attributes should have been defined before loading the file? Or should it just be a warning to inform the user of this implicit behaviour?

angelicaochoa:
Thanks for the feedback @fedde-s ! To answer your question about unknown attributes in headerless files, you pretty much nailed it. The attribute properties for relaxed validation are really just placeholders to allow the validation to complete while also flagging attributes that are new sample-level or patient-level attributes. 

I think a warning is sufficient but I understand why an error might be more informative for a user who may not have a pipeline set up to fill in those properties on the fly but may still want to validate their data files in one go. Thoughts @n1zea144 @zheins @ritikakundra ?

pieterlukasse:
@angelicaochoa is the goal to allow for clinical files without the metadata
header lines? Perhaps we can build in native support for this in the
validation without the need of a workaround. Then the code could be quite
simple I think and there would be no need to "modify"/complement some of
the data on the fly.

www.thehyve.nl
E pieter@thehyve.nl
T +31(0)30 700 9713
M +31(0)6 28 18 9540
Skype pieter.lukasse


We empower scientists by building on open source software

2016-08-11 16:58 GMT+02:00 angelicaochoa <notifications@github.com>:

n1zea144:
@pieterlukasse The goal is to allow validation to continue if metadata header lines is missing - to issue a warning instead of failing.

pieterlukasse:
@n1zea144 right, so we could just change the validation code to make this scenario acceptable as well. The only error that would occur is when an attribute is not yet known in the DB. But for existing attributes it could be fine to let the validation just assume the definition found in the DB is the one to use. I haven't checked the code in detail...maybe this is already what you are doing.

n1zea144:
That sounds good to me.  The import point is that from the data curator point-of-view, they would like the validator to continue validating the other files in the study whether or not the clinical attribute definition appears in the database.

core/src/main/scripts/importer/validateData.py
```diff
@@ -313,9 +317,14 @@ def _validate_file(self):
...
-                self.logger.error(
-                    'Invalid header comments, file cannot be parsed')
-                return
+                if not self.relaxed_mode:
+                    self.logger.error(
+                        'Invalid header comments, file cannot be parsed')
+                    return
+                else:   
+                    self.logger.info('Ignoring missing or invalid header comments. '
+                        'Continuing with validation...')
```
fedde-s: I just noticed: this is something you should not need to change, it is more general than clinical data files. If we make a file without the header lines valid, `processTopLines` will not return `False` for those.

fedde-s: And if we do want to disable the fail-fast behaviour and continue validating other files even if no samples are defined, as @n1zea144 asks, that’s the following block in the `validate_study` function:

```Python
    if defined_sample_ids is None:
        logger.error("Sample file could not be parsed. Please fix "
                     "the problems found there first before continuing.")
        return
```

angelicaochoa:
@n1zea144 @fedde-s @pieterlukasse 
This is ready to merge if one of you can review again and see if I've missed anything

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -83,14 +83,16 @@ class DataFileTestCase(LogBufferTestCase):
...
     """
 
-    def validate(self, data_filename, validator_class, extra_meta_fields=None):
+    def validate(self, data_filename, validator_class, extra_meta_fields=None, relaxed_mode=None):
         """Validate a file with a Validator and return the log records."""
         meta_dict = {'data_filename': data_filename}
         if extra_meta_fields is not None:
             meta_dict.update(extra_meta_fields)
+        if relaxed_mode is None:
+            relaxed_mode = False
```
fedde-s: Just a note: the reason that the default for `extra_meta_fields` is `None` instead of the dictionary that will actually be used is that dictionaries are mutable in place, and using mutable objects as default arguments in Python can lead to [confusing situations](http://effbot.org/zone/default-values.htm) if not specifically intended. There is no real need not to use `False` as a default, booleans are not mutable objects.

fedde-s: In other words, this could be as follows:
```Python
def  validate([…], relaxed_mode=False)
```
and then the if-statement is not necessary.

pieterlukasse:
For the record: I reviewed the code on some of the details like: will required columns still be validated? And will duplicated column names give an error? And the changes look good to me :+1: 
One test that maybe should be added is: what happens when a clinical file is missing the required field PATIENT_ID or SAMPLE_ID?

PR #1583, opened 2016-08-11
Updating gene length method

oplantalech:
The current method to calculate gene length uses a bed file to obtain the length of the exons of a particular gene, and then adding the exon lengths in positions that are not overlapping to obtain the length of this particular gene (for example, if a gene have two exons, one that goes from position 1 to 4 and another from position 3 to 7, they will be counted as a single exon going from position 1 to 7). However, the current method has a problem since it only takes the exons of the same gene together if they are in consecutive lines in the bed file. This problem is solved here, using also new methods to update genes created by @pieterlukasse.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
@@ -104,9 +104,6 @@ final class DaoGene {
             if (rows != 1) {
                 ProgressMonitor.logWarning("No change for " + gene.getEntrezGeneId() + " " + gene.getHugoGeneSymbolAllCaps() + "? Code " + rows);
             }
-            else {
-                ProgressMonitor.logWarning("Updated gene"); //should be message, but warnings are nicely summarized. TODO - add message summary to ProgressMonitor
-            }
             //add the current set of aliases:
             rows += addGeneAliases(gene);
 
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
@@ -33,6 +33,7 @@
 package org.mskcc.cbio.portal.dao;
 
 import org.mskcc.cbio.portal.model.CanonicalGene;
+import org.mskcc.cbio.portal.util.ProgressMonitor;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -41,6 +42,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -70,6 +72,57 @@ final class DaoGene {
         return addGene(gene);
     }
 
+    /**
+     * Update Gene Record in the Database. It will also replace this 
+     * gene's aliases with the ones found in the given gene object.
+     */
+    public static int updateGene(CanonicalGene gene) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+        boolean setBulkLoadAtEnd = false;
+        try {
+            //this method only works well with bulk load off, especially 
+            //when it is called in a process that may update a gene more than once
+            //(e.g. the ImportGeneData updates some of the fields based on one 
+            // input file and other fields based on another input file):
+            setBulkLoadAtEnd = MySQLbulkLoader.isBulkLoad();
+            MySQLbulkLoader.bulkLoadOff();
+            //delete aliases first:
+            deleteGeneAlias(gene.getEntrezGeneId());
+            
+            int rows = 0;
+            con = JdbcUtil.getDbConnection(DaoGene.class);
+            pstmt = con.prepareStatement
+                    ("UPDATE gene SET `HUGO_GENE_SYMBOL`=?, `TYPE`=?,`CYTOBAND`=?,`LENGTH`=? WHERE `ENTREZ_GENE_ID`=?");
+            pstmt.setString(1, gene.getHugoGeneSymbolAllCaps());
+            pstmt.setString(2, gene.getType());
+            pstmt.setString(3, gene.getCytoband());
+            pstmt.setInt(4, gene.getLength());
+            pstmt.setLong(5, gene.getEntrezGeneId());
+            rows += pstmt.executeUpdate();
+            if (rows != 1) {
+                ProgressMonitor.logWarning("No change for " + gene.getEntrezGeneId() + " " + gene.getHugoGeneSymbolAllCaps() + "? Code " + rows);
+            }
+            else {
+                ProgressMonitor.logWarning("Updated gene"); //should be message, but warnings are nicely summarized. TODO - add message summary to ProgressMonitor
+            }
+            //add the current set of aliases:
+            rows += addGeneAliases(gene);
+
+            return rows;
+        } catch (SQLException e) {
+            throw new DaoException(e);
+        } finally {
+            if (setBulkLoadAtEnd) {
+                //reset to original state:
+                MySQLbulkLoader.bulkLoadOn();
+            }   
+            JdbcUtil.closeAll(DaoGene.class, con, pstmt, rs);
+        }
+        
+}
+    
     /**
      * Adds a new Gene Record to the Database.
      *
@@ -246,7 +299,7 @@ final class DaoGene {
             JdbcUtil.closeAll(DaoGene.class, con, pstmt, rs);
         }
     }
-
+    
     /**
      * Gets all Genes in the Database.
      *
@@ -396,6 +449,8 @@ final class DaoGene {
      * Deletes all Gene Records in the Database.
      *
      * @throws DaoException Database Error.
+     * 
+     * @deprecated only used by deprecated code, so deprecating this as well.
      */
     public static void deleteAllRecords() throws DaoException {
         Connection con = null;
@@ -415,6 +470,12 @@ final class DaoGene {
         deleteAllAliasRecords();
     }
     
+    /**
+     * 
+     * @throws DaoException
+     * 
+     * @deprecated only used by deprecated code, so deprecating this as well.
+     */
     private static void deleteAllAliasRecords() throws DaoException {
         Connection con = null;
         PreparedStatement pstmt = null;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -248,6 +286,43 @@ private static CanonicalGene getGene(long entrezGeneId) throws DaoException {
     }
 
     /**
+     * Gets all genes where Hugo gene symbol is marked as deprecated.
+     * 
+     * @return list of genes
+     * 
+     * @throws DaoException 
+     */
+    public static List<CanonicalGene> getDeprecatedGenes() throws DaoException {
```
pieterlukasse: @	oplantalech  you can leave this method out

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -169,20 +169,26 @@ public int addGene(CanonicalGene gene) throws DaoException {
...
     }
+
+    /**
+     * Gets all genes where Hugo gene symbol is marked as deprecated.
+     * 
+     * @return list of genes
+     * 
+     * @throws DaoException 
+     */
+    public List<CanonicalGene> getDeprecatedGenes() throws DaoException {
```
pieterlukasse: leave this one out as well

core/src/main/resources/db/migration.sql
```diff
@@ -22,3 +22,8 @@ PREPARE stmt FROM @s;
 EXECUTE stmt;
 DEALLOCATE PREPARE stmt;
 UPDATE info SET DB_SCHEMA_VERSION="1.2.1";
+
+##version: 1.2.2
+
+ALTER TABLE gene ADD COLUMN `PREVIOUS_SYMBOL` VARCHAR(255) NULL;
```
pieterlukasse: leave this one out

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -257,50 +362,107 @@ static void importSuppGeneData(File suppGeneFile) throws IOException, DaoExcepti
...
-            File miRNAFile = new File(args[2]);
-            System.out.println("Reading miRNA data from:  " + miRNAFile.getAbsolutePath());
-            numLines = FileUtil.getNumLines(miRNAFile);
-            System.out.println(" --> total number of lines:  " + numLines);
-            ProgressMonitor.setMaxValue(numLines);
-            ImportMicroRNAIDs.importData(miRNAFile);
+	@Override
+    public void run() {
+		try {
+			SpringUtil.initDataSource();
```
pieterlukasse: here I would add a check to see if DB tables `gene` and `gene_alias` are not empty. If not empty, then throw an error referring the user to a documentation page with information on how to empty these tables in a safe way. So add a new doc with the contents of https://github.com/thehyve/cbioportal/wiki/Updating-gene-and-gene_alias-tables to this PR

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -156,11 +202,51 @@ public static void importData(File geneFile) throws IOException, DaoException {
...
+     * This method will prepare the gene and gene_alias tables for being updated. The gene_alias table 
+     * will just reflect the new data being imported, so we start by making it empty. The gene table
+     * is a bit more complicated since many other tables refer to it. Here we keep track of the previous symbol
+     * for reporting purposes at the end. See TODO - add method ref. 
+     * 
+     * 
+     * @return
+     * @throws DaoException
+     */
+    private static int prepareGeneTablesForUpdate() throws DaoException {
```
pieterlukasse: remove this method

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -138,7 +170,21 @@ public static void importData(File geneFile) throws IOException, DaoException {
 	            if (genes.size()==1) {
 	                CanonicalGene gene = genes.iterator().next();
 	                if (!genesWithSymbolFromNomenClatureAuthority.containsKey(symbol)) {
-	                    daoGene.addGene(gene);
+	                    //if exists
+	                    CanonicalGene dbGene = daoGene.getGene(gene.getEntrezGeneId());
```
pieterlukasse: since we will be only adding genes and not updating them, this check can be removed and only the `else` can stay

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -114,14 +132,28 @@ public static void importData(File geneFile) throws IOException, DaoException {
...
 
         // Add genes with symbol from nomenclature authority
         for (Map.Entry<String, Set<CanonicalGene>> entry : genesWithSymbolFromNomenClatureAuthority.entrySet()) {
             Set<CanonicalGene> genes = entry.getValue();
             if (genes.size()==1) {
-                daoGene.addGene(genes.iterator().next());
+                CanonicalGene gene = genes.iterator().next();
+                //if exists
+                CanonicalGene dbGene = daoGene.getGene(gene.getEntrezGeneId());
+                if (dbGene != null) {
```
pieterlukasse: since we will be only adding genes and not updating them, this check can be removed and only the `else` can stay

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -169,20 +169,26 @@ public int addGene(CanonicalGene gene) throws DaoException {
...
      */
-    public int flushUpdateToDatabase() throws DaoException {
-        DaoGene.deleteAllRecords();
-        MySQLbulkLoader.bulkLoadOn();
-        int ret = 0;
-        for (CanonicalGene gene : getAllGenes()) {
-            ret += DaoGene.addGene(gene);
-        }
-        MySQLbulkLoader.flushAll();
+    public int updateGene(CanonicalGene gene)  throws DaoException {
```
pieterlukasse: since we will be only adding genes and not updating them, this method can be removed

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -71,6 +72,43 @@ public static synchronized int addGeneWithoutEntrezGeneId(CanonicalGene gene) th
     }
 
     /**
+     * Update Gene Record in the Database. It will also replace this 
+     * gene's aliases with the ones found in the given gene object.
+     */
+    public static int updateGene(CanonicalGene gene) throws DaoException {
```
pieterlukasse: since we will be only adding genes and not updating them, this method can be removed

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -174,45 +260,62 @@ private static void logDuplicateGeneSymbolWarning(String symbol, Set<CanonicalGe
...
         
-        daoGeneOptimized.flushUpdateToDatabase();
-    }
+        //Calculate the length for each gene and update genes
+        for (Long key : geneLociMap.keySet()) { //Get the keys and iterate over them
+        	ArrayList<long[]> exons = geneLociMap.get(key); //Get the values
+        	int length = calculateGeneLength(exons);
+        	CanonicalGene gene = daoGeneOptimized.getGene(key);
+        	gene.setLength(length);
+            daoGeneOptimized.updateGene(gene);
```
pieterlukasse: keep only this call to updateGene()

oplantalech:
I have updated the code with the suggestions of @pieterlukasse.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -36,23 +36,45 @@
...
         Map<String, Set<CanonicalGene>> genesWithSymbolFromNomenClatureAuthority = new LinkedHashMap<>();
         Map<String, Set<CanonicalGene>> genesWithoutSymbolFromNomenClatureAuthority = new LinkedHashMap<>();
         try (FileReader reader = new FileReader(geneFile)) {
             BufferedReader buf = new BufferedReader(reader);
             String line;
+            DaoGeneOptimized daoGene = DaoGeneOptimized.getInstance();
+            ArrayList<CanonicalGene> allGenes = daoGene.getAllGenes();
+            for(CanonicalGene geneName : allGenes) { //Loop through all genes
+            	if (geneName != null) {
+                throw new DaoException("Error: the table Gene is not empty.");
```
pieterlukasse: this can be simplified

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -174,45 +203,73 @@ private static void logDuplicateGeneSymbolWarning(String symbol, Set<CanonicalGe
...
+        		}
         }
         
-        daoGeneOptimized.flushUpdateToDatabase();
-    }
+        //Calculate the length for each gene and update genes
+        for (Long key : geneLociMap.keySet()) { //Get the keys and iterate over them
+        	List<long[]> exons = geneLociMap.get(key); //Get the values
+        	int length = calculateGeneLength(exons);
+        	CanonicalGene gene = daoGeneOptimized.getGene(key);
```
pieterlukasse: does this retrieve the aliases?

fedde-s:
I believe all database changes have been removed from the pull request now.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -415,6 +467,12 @@ public static void deleteAllRecords() throws DaoException {
         deleteAllAliasRecords();
     }
     
+    /**
+     * 
+     * @throws DaoException
+     * 
+     * @deprecated only used by deprecated code, so deprecating this as well.
```
fedde-s: Would these comments be sufficient to tell people who are considering using the function what to do instead, and why?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -169,18 +169,13 @@ public int addGene(CanonicalGene gene) throws DaoException {
     }
     
     /**
-     * Update database with gene length
-     * @return number of records updated.
-     * @throws DaoException 
+     * Update Gene Record in the Database. It will also replace this 
+     * gene's aliases with the ones found in the given gene object.
```
fedde-s: Does it the function still return the number of records updated? If so, is there a reason not to document this?

PR #1570, opened 2016-08-04
description for "type" in mutation_count

anthony-cros:
# What? Why?
added necessary description for "type" in mutation_count

I had to dig through the code to figure out what values where acceptable.

Fix # .
added annotation

Changes proposed in this pull request:
- a
- b

# Checks
- [ ] Runs on Heroku.
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Google Style Guide](https://github.com/google/styleguide).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)
- [ ] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to
  hotfix.

# Any screenshots or GIFs?
If this is a new visual feature please add a before/after screenshot or gif
here with e.g. [GifGrabber](http://www.gifgrabber.com/).

# Notify reviewers
If your pull request involves changes to an existing file, one or more of the
people that edited before you might be good candidates to review it. Please use
`git blame <filename>` to determine that and notify them here. Otherwise notify
everybody in the core team:
@cBioPortal/core-developers

PR #1585, opened 2016-08-11
description for "type" in mutation_count

anthony-cros:
see [original PR](https://github.com/cBioPortal/cbioportal/pull/1570)

### DIFF
--- a/business/pom.xml
+++ b/business/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <artifactId>master</artifactId>
     <groupId>org.mskcc.cbio</groupId>
-    <version>1.2.4</version>
+    <version>1.2.5-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
   <artifactId>business</artifactId>
--- a/business/pom.xml
+++ b/business/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <artifactId>master</artifactId>
     <groupId>org.mskcc.cbio</groupId>
-    <version>1.2.4-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


sheridancbio:
getMutationsCounts argument "type" has gone from required to not-required. Checking code now to see if an appropriate default is provided.

sheridancbio:
According to ApiService.java, there is logic which puts into the results hashmap an entry under key "count" or key "frequency" based on the value of this parameter. If the parameter is not supplied, no key will be assigned to the hashmap. It seems like this parameter should be required to me. (Otherwise, no counts are added to the elements retrieved from the repository) Is this a use case ? (return values without count attribute or frequency attribute)?

sheridancbio:
Ah .. I see .. second commit returned parameter to required. ok

sheridancbio:
type 'count' probably means absolute counts. type 'frequency" is probably incorrect terminology. Frequency typically means the average number occurrences per duration interval. I need to speculate as to the true meaning here ... I'm guessing the intention might be 'proportion', 'fraction', or 'prevalence' relative to the complete sample set.  But adjusting this terminology (along with the api description) should probably be a different pull request.

sheridancbio:
Will merge .. but please use git rebase to squash these two commits into one, and to relocate the changes to the front of the hotfix head.

anthony-cros:
I can't seem to rebase, I'll have to look into why on Monday!

sheridancbio:
I am closing this PR ... the submitted change is being incorporated through PR #1607 which I will now merge. This was to assist with rebasing the change set to the head of hotfix

anthony-cros:
Thanks, Robert!

PR #1600, opened 2016-08-17
Authorization fixes

pieterlukasse:
# What? Why?
Combining #1574 and #1575 and pointing to `hotfix` branch instead of `rc`. See original PRs for description.

# Notify reviewers
@jjgao

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/CancerStudyPermissionEvaluator.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/CancerStudyPermissionEvaluator.java
@@ -175,7 +175,12 @@ class CancerStudyPermissionEvaluator implements PermissionEvaluator {
         } catch (DaoException e) {
             groups = cancerStudy.getGroups();
         }
-        if (!Collections.disjoint(groups, grantedAuthorities)) {
+        // to upperCase to make the comparison with grantedAuthorities case insensitive:
+        Set<String> groupsUpper = new HashSet<String>();
+		for (String group: groups) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/CancerStudyPermissionEvaluator.java
```diff
@@ -175,7 +175,12 @@ private boolean hasPermission(CancerStudy cancerStudy, Authentication authentica
         } catch (DaoException e) {
             groups = cancerStudy.getGroups();
         }
-        if (!Collections.disjoint(groups, grantedAuthorities)) {
+        // to upperCase to make the comparison with grantedAuthorities case insensitive:
+        Set<String> groupsUpper = new HashSet<String>();
```
jjgao: Maybe cancerStudy.getGroups() can return uppercase directly -- you could put it in SQL: UPPER(GROUPS) ?

pieterlukasse: hi @jjgao , I wanted to avoid changing this class as it could be used elsewhere. The uppercase here is specific to how the permissions are evaluated in this class, so I wanted to keep this local. 

But now I also see why this PR should be to `rc`....it builds upon a previous fix in getGrantedAuthorities which is not in `hotfix` (see 5e8ed7a9c48e342350e2e7af17236b7294509659)....so maybe it is best to reopen the `rc` PRs...

pieterlukasse:
reopening `rc` based PR...closing this one based on comments above

PR #1614, opened 2016-08-19
Fix #1595: Added selection for mass spec data to enrichments tab

### LINKED ISSUE #1595: support selection of profiles in Protein tab of Enrichments analysis
Now we have both RPPA and mass spec data, we need to support selection of a protein level profile, similar to the mRNA tab.

pambot:
This pull request adds the selection menu to the Protein expression sub-tab in the Enrichments tab, in a similar style to the mRNA sub-tab. This addresses issue #1595. Since stable IDs are still a bit temporary for mass spec, there's still a bit of work on `view.js` when they get finalized. Also, some code could be more efficient, but this first draft should work.


yichaoS:
Could you please the conflicts? This should be due to us moving the scatter plots (under the folding area) to the left side.

Issue #494, opened 2015-11-17
Let the import process generate the standard case lists


pieterlukasse:
Let the import process generate the standard case lists based on the data loaded. For example: if the study loaded contains CNA and mRNA (microarray) data, then the import process should generate the case lists `<study_id>_all`, `<study_id>_cna` and `<study_id>_mrna`. 

This replaces the need for manually providing the import process with these files (see [documentation of Case lists files](https://github.com/cBioPortal/cbioportal/blob/v1.8.2/docs/File-Formats.md#case-lists) for a description of the current process). 

Goals: 
- make case lists optional. Users only provide input files for _custom_ case lists. 
- simplify the import process making it much less error prone

pieterlukasse:
Question1: how to generate "All Complete Tumors" in the case where we have more than 1 data sets for a same genetic profile (e.g. multiple mRNA datasets for mRNA genetic profile)?

Question2: should we generate case lists for all possible combinations of two specific experiments? E.g. CNA + mRNA(RNAseq),  or CNA + mRNA(microarray)

pieterlukasse:
@ecerami : I will be very happy if we can work on this one in next phase ;)

ecerami:
@pieterlukasse:  this is actually a little trickier than it looks :-)  The problem is that if we have a MAF file, and generate a case set of sequenced samples from that MAF file, it's very possible (especially for targeted sequenced panels), that we have many samples that are Wildtype for all genes sequenced.  We therefore need a separate case list to specify all samples sequenced, including the Wildtype cases.

pieterlukasse:
@ecerami : did you mean "excluding the wildtype cases"  in you last sentence?

ecerami:
@pieterlukasse:  no, I mean "including WT cases", e.g. if we have sequenced 100 samples, and 10 are WT, we need a list of 100, not 90.

Sjoerd-van-Hagen:
@ecerami Can we expect a file that lists all the sequenced samples, or all samples that are sequenced but are wild-type?

fedde-s:
@Sjoerd-van-Hagen I think changing the requirements that way would only make it needlessly complex both for the users and for the software, no need to manually figure out which of the sequenced samples are not in the MAF file, only to have cBioPortal make a full list of all sequenced samples again.

pieterlukasse:
@fedde-s , @Sjoerd-van-Hagen   : i discussed it with Ethan in more detail before. I will add the updated information.

[...]

PR #1618, opened 2016-08-24
Gene panel new tech stack

zheins:
# What? Why?
Gene Panel information is needed to determine the difference between no mutations/alterations and not being sequenced/tested for a mutation/alterations.

- New GenePanelRepository to interact with gene panel data
- Changes to db schema, addition of 3 new tables: gene_panel (panel metadata), gene_panel_list (genes associated with the gene panel), and gene_panel_profile_map (maps profile sample pairs to a specific gene panel).
- Unit tests for the gene panel repository
- Temporary import code and functions in the new repository.
- Update the File-Formats documentation

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
everybody in the core team:
@n1zea144 @sheridancbio


zheins:
A core test from rc is failing, not the tests from this PR.

sheridancbio:
I'm reviewing this PR now. From travis, there are 2 persistence unit tests failing due to mismatches to test data (failed asserts) and three errors/exceptions which cause tests to not complete (sqlException: table gene_alias not found)

sheridancbio:
Here are comments from a code readthrough ... these will be revised on completion [about 50% through code now]


- [this note was already addressed in person] we can merge this as is .. but we don't want to be adding anything new to business. Mapper and Repository should be located in persistence-mybatis / persistence-api as appropriate, and services

Very general class ImportTabDelimData.java now takes a genePanel argument to constructor.
ImportFusionData.java also genePanel controlled, as is ImportExtendedMutationData

gene panels have stable IDs .. it is essentially a map from a stable ID to a list of genes (uniprot ID ints)
ImportGenePanel.getGenes should use a Set rather than List for geneIds .. perhaps LinkedHashSet<Integer> geneIds
ImportGenePanel : genes are provided as list of ints, list of Hugo symbol or list of Aliases .. or a mixed bunch. Shouldn't this be specified by the import process? This seems much too accommodating to callers.

ImportGenePanelProfileMap :: requires cancer_study
	GenePanelRepository genePanelRepository = SpringUtil.getGenePanelRepository();  :: no autowiring of @Repository?
	Integer sidIndex = profiles.indexOf("SAMPLE_ID");  .. hacky way of stripping header. Can header really appear anywhere?
	String sid = data.get(sidIndex); .. rename sid to sampleId
	on import .. GenePanels are three-way link 

general:
	all code files must have project copyright and license headers
		core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenePanel.java
style notes:
Avoid tab characters and use 4 space indents: (can be fixed after merge)
	core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
	core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
whitespace at end of lines serves no purpose:
	core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
Too many blank lines inside functions: (in general, I think we throw these around too much and they lose their meaning)
	ImportGenePanelProfileMap.getProfilesLine

persistence/persistence-mybatis-test/src/test/resources/testSql.sql
```diff
@@ -17,6 +17,9 @@ DROP TABLE IF EXISTS "genetic_profile";
 DROP TABLE IF EXISTS "gene";
 DROP TABLE IF EXISTS "cancer_study";
 DROP TABLE IF EXISTS "type_of_cancer";
+DROP TABLE IF EXISTS "gene_panel";
+DROP TABLE IF EXISTS "gene_panel_list";
```
sheridancbio: gene_panel_list has gene_panel and gene as parent tables in foreign key constraints. It must be dropped before any parent table is dropped because of this relationship

gene_panel_profile_map has gene_panel and sample and genetic_profile as parent tables in foreign key constraints. It must be dropped before any parent table is dropped because of this relationship

Move the DROP gene_panel_list line to a position above the Drop gene line
Move the DROP gene_panel_profile_map line to a position above the Drop sample line

sheridancbio:
Note for future ... In discussion above I said:
"Import tests all use a null gene-panel. We should have unit tests with an active gene-panel."
But there were laborious difficulties getting the import scripts in core (which under testing use a memory simulation of an actual DBMS) to make updates through the mybatis update mappings in the persistence module. Import was successfully performed manually outside of unit testing. When persistence modules are (hopefully soon) compartmentalized in an external repository and perform both reading and writing functions, they will have their own unit tests, and the core import scripts may cease to be used.

zheins:
@sheridancbio change has been made

persistence/persistence-mybatis-test/src/test/resources/testSql.sql
sheridancbio: The DROP TABLE line for gene_panel_list is still below the DROP TABLE line for gene
The DROP TABLE line for gene_panel_profile_map is still below the DROP TABLE line for sample.
Move this line upwards by about 6 lines (above sample) :
    DROP TABLE IF EXISTS "gene_panel_profile_map";
Move this line upwards by about 4 lines: (above gene)
    DROP TABLE IF EXISTS "gene_panel_list"

sheridancbio:
I think we also need to adjust web/src/main/resources/applicationContext-web.xml to scan the new controller package (org.cbioportal.weblegacy) as well as the others.

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
```diff
@@ -13,6 +13,7 @@
 import org.cbioportal.model.GenePanel;
 import org.cbioportal.persistence.GenePanelRepository;
 import org.cbioportal.service.GenePanelService;
+import java.util.List;
```
sheridancbio: don't need to import List anymore

web/src/main/java/org/cbioportal/weblegacy/GenePanelController.java
```diff
@@ -50,16 +51,33 @@
 
     @Autowired
     private GenePanelService genePanelService;
-
-    @ApiOperation(value = "Get gene panel information",
+    
+@ApiOperation(value = "Get gene panel information",
```
sheridancbio: this could be shifted over with 4 space indent

web/src/main/java/org/cbioportal/weblegacy/GenePanelController.java
```diff
@@ -50,16 +51,33 @@
...
+            notes = "")
+    @Transactional
+    @RequestMapping(method = RequestMethod.GET, value = "/genepanel/data",  produces="application/json")
+    public String getGenePanelData(@ApiParam(required = true, value = "sample id, such as those returned by /api/samples")
             @RequestParam(required = true) String sampleId,
             @ApiParam(required = true, value = "genetic profile id, such as those returned by /api/geneticprofiles")
             @RequestParam(required = true) String profileId) {
         return genePanelService.getGenePanelBySampleIdAndProfileId(sampleId, profileId);
-    }
+    }
```
sheridancbio: trailing whitespace

zheins:
@sheridancbio returning a list allows the endpoint to encompass multiple use cases - returning multiple or returning just one.

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
zheins: We do. There are `List<GenePanel>` in this class.

core/src/main/resources/db/cgds.sql
```diff
@@ -724,8 +727,34 @@ CREATE TABLE `clinical_event_data` (
...
+    PRIMARY KEY (`INTERNAL_ID`, `GENE_ID`),
+    FOREIGN KEY (`INTERNAL_ID`) REFERENCES `gene_panel` (`INTERNAL_ID`) ON DELETE CASCADE,
+    FOREIGN KEY (`GENE_ID`) REFERENCES `gene` (`ENTREZ_GENE_ID`) ON DELETE CASCADE
+);
+
+CREATE TABLE `gene_panel_profile_map` (
+    `SAMPLE_ID` int(11) NOT NULL,
+    `PROFILE_ID` int(11) NOT NULL,
+    `PANEL_ID` int(11) NOT NULL,
+    PRIMARY KEY (`SAMPLE_ID`, `PROFILE_ID`),
```
fedde-s: @zheins: Why did you not just add a field to `sample_profile`, the existing mapping table between `genetic_profile` and `sample`?

fedde-s: Furthermore: according to File-Formats.md, metadata files for genetic profiles specify one gene panel for the entire profile. Why can `PANEL_ID` not just be added to the `genetic_profile` table then? Will different samples in the same genetic profile (~experiment/data file) be targeted using different gene panels?

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml
```diff
@@ -0,0 +1,302 @@
...
+                    <result property="name" column="NAME"/>
+                    <result property="clinicalTrialKeywords" column="CLINICAL_TRIAL_KEYWORDS"/>
+                    <result property="dedicatedColor" column="DEDICATED_COLOR"/>
+                    <result property="shortName" column="SHORT_NAME"/>
+                    <result property="parent" column="PARENT"/>
+                </association>
+            </association>
+        </association>
+        <association property="typeOfCancer" javaType="org.cbioportal.model.TypeOfCancer">
+            <result property="typeOfCancerId" column="TYPE_OF_CANCER_ID"/>
```
fedde-s: @zheins: could you explain how the framework knows that this one refers to the `sample.TYPE_OF_CANCER_ID` in the result set, and not to one of the other two, `cancer_study.TYPE_OF_CANCER_ID` and `type_of_cancer.TYPE_OF_CANCER_ID`? And likewise for `sample.STABLE_ID` and `patient.STABLE_ID`.

core/src/main/resources/db/cgds.sql
fedde-s: Never mind the latter question, I see that that is an existing use-case catered by the sample-profile matrix file formats. But what about the first question, the `sample_profile` table?

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml
zheins: I didn't make this distinction - this should have been done with aliasing in the sql query. This information probably will not ever be used/exposed (gene panels and the queries associated with them do not need this data), but I think should be fixed just so the code is cleaner.

core/src/main/resources/db/cgds.sql
zheins: Good idea! I didn't know about the other table.  I'll make another PR for this.

pieterlukasse:
Hi @awernibi: FYI - I think this PR is good news for you and aligns well with your interest in being able to see the difference between "no signal" vs "not sequenced" in oncoprint. @adamabeshouse any plans to start using this in oncoprint soon?

pieterlukasse:
@aderidder I think you will also be glad to hear about this one.

core/src/main/resources/db/cgds.sql
pieterlukasse: nb: I believe not all samples are currently found in `sample_profile`, only the ones found in the various data files (e.g. like the maf file). So you using this table in your new PR might entail some extra work...

core/src/main/resources/db/cgds.sql
```diff
@@ -724,8 +727,34 @@ CREATE TABLE `clinical_event_data` (
...
+    FOREIGN KEY (`PROFILE_ID`) REFERENCES `genetic_profile` (`GENETIC_PROFILE_ID`) ON DELETE CASCADE,
+    FOREIGN KEY (`PANEL_ID`) REFERENCES `gene_panel` (`INTERNAL_ID`) ON DELETE CASCADE
+);
+
 CREATE TABLE `info` (
     `DB_SCHEMA_VERSION` varchar(8)
 );
 -- THIS MUST BE KEPT IN SYNC WITH db.version PROPERTY IN pom.xml
-INSERT INTO info VALUES ('1.2.1');
+INSERT INTO info VALUES ('1.2.2');
```
pieterlukasse: @zheins you forgot to update the pom.xml!

core/src/main/resources/db/cgds.sql
sheridancbio: I am trying to understand the issue just raised by Pieter. Is this the scenario? : a sample is processed as part of a batch of samples using a particular (limited) gene panel; then the results come back as invalid for the sample or showing no mutation events at all ; the sample_profile table does not show the sample even though it was processed with a gene panel so we are unable to register the gene_panel ; finally in the oncoprint we cannot distinguish for such a sample whether a gene was "tested but not mutated" versus "untested" ?
Secondary question: can a sample be sent for sequencing but then be excluded from the generated results because of detection of problems during processing? If this does happen, perhaps we should not consider the gene panel genes as being "tested" because the test itself was invalidated.

fedde-s: This has been partially discussed following the proposal in issue #494.

pieterlukasse: @sheridancbio you understanding is correct. 

And yes, filling `sample_profile` with all samples of a profile would be a potential solution for #494. However, this does not remove the need for case lists as they could serve another use case (e.g. custom case lists). You should also be careful to make sure the other parts of the code, that use `sample_profile` in its current form, keep working as expected.

zheins: @pieterlukasse @fedde-s There is a new file format that is documented to associate samples/profiles/panels. ImportGenePanelProfileMap.java handles this new file.

PR #1619, opened 2016-08-24
Fix unit tests for MySQL 5.1

mandawilson:
# What? Why?
Fix unit tests for MySQL 5.1.  InnoDB is required for unit tests and was not successfully set as default storage engine.

- By default sets MySQL system variable default_storage_engine=InnoDB, but for MySQL 5.1 run with 'mvn -DstorageEngineVariable=storage_engine test' to set storage_engine=InnoDB
- Updated documentation for using bitly on localhost.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers


core/src/main/resources/db/cgds.sql
```diff
@@ -282,7 +282,7 @@ CREATE TABLE `genetic_profile_samples` (
 CREATE TABLE `sample_profile` (
   `SAMPLE_ID` int(11) NOT NULL,
   `GENETIC_PROFILE_ID` int(11) NOT NULL,
-  UNIQUE KEY `UQ_SAMPLE_ID_GENETIC_PROFILE_ID` (`SAMPLE_ID`,`GENETIC_PROFILE_ID`) COMMENT 'Constraint to allow each sample only once in each profile.',
+  UNIQUE KEY `UQ_SAMPLE_ID_GENETIC_PROFILE_ID` (`SAMPLE_ID`,`GENETIC_PROFILE_ID`), -- Constraint to allow each sample only once in each profile
```
sheridancbio: Good catch. I think this has been fixed in mysql releases after 5.4, but until we upgrade our production system I think we should continue to comment out our comments on indices. (Before we switch back to actual SQL comments, we should be ready to upgrade our specification of a minimal supported deployment platform to be mysql 5.4 or newer (or whatever the appropriate version is)

sheridancbio:
After discussion on this PR, the configuration question has been handled through a separate PR. But we still intend to comment out the "COMMENT" clauses on the KEY fields, because we expect some current users to continue to use mysql 5.1 for a while which does not support this. When the minimum development platform specifies mysql 5.7 or newer, we can put the COMMENT clauses into force.

Issue #1413, opened 2016-06-27
Add heatmap by adding new tracks based on oncoprint.js


pambot:
This is the first step to adding a heatmap to the OncoPrint tab. It can be initially added as new tracks but with different visual parameters (i.e. no whitespace in between the blocks).

pambot:
**Progress report:** Heatmap tracks and some sensible defaults have been added, now it's time to tweak the Clinical Toolbar to allow the user to customize the view.

**Changes suggested by @jjgao:**
1) Do not show heat map by default
2) Add heatmap submenu to show heatmap
3) Allow user to select a z-score profile
4) Allow user to add additional genes
5) Allow user to change the color scale

fedde-s:
I believe this issue has been solved by #1847 and #1991. They do not include a way for the user to select the colour scheme, but the latter pull request does change the default based on reviews/feedback.

PR #1625, opened 2016-08-26
PR for OncoPrint heatmap tracks, issue #1413

pambot:
This PR contains the code for the heatmap tracks addition to the OncoPrint tab (as seen during the 8/22 lab meeting). It's not ready for merging yet, but I'm tying this fork branch to #1413 to keep track of suggested improvements (often communicated on Slack) and their respective commits.


portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -583,6 +583,25 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	'getGeneticProfileIds': function () {
 	    return this.genetic_profile_ids;
 	},
+	'getDefaultGeneticProfileId': function () {
+			var gp_ids = this.getGeneticProfileIds();
+			var chosen_ids = [];
+			for (var i = 0; i < gp_ids.length; i++) {
+				if (~gp_ids[i].indexOf('Zscores') && ~gp_ids[i].indexOf('mrna')) {
```
adamabeshouse: All this `~` should be `!`

`~` is bitwise NOT, `!` is boolean. You may get the same result but it's confusing.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -583,6 +583,25 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	'getGeneticProfileIds': function () {
 	    return this.genetic_profile_ids;
 	},
+	'getDefaultGeneticProfileId': function () {
+			var gp_ids = this.getGeneticProfileIds();
```
adamabeshouse: There's a method `getGeneticProfiles` on the data manager which includes other metadata about the genetic profiles besides the ids, including the data type. You should use that instead of using `indexOf` on the id.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -583,6 +583,25 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	'getGeneticProfileIds': function () {
 	    return this.genetic_profile_ids;
 	},
+	'getDefaultGeneticProfileId': function () {
```
adamabeshouse: I think this should be renamed to make it clear this is specifically for the heatmap. Outside the heatmap there's no notion of a default genetic profile id.

adamabeshouse:
Looks great! Just a few comments.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
fedde-s: I think it tests if it’s unequal to `-1`, as `~(-1) == 0`. But I believe that this is not a common way to express this, and it already seems to be confusing people. I suggest rephrasing it to `gp_ids[i].indexOf('Zscores') != -1 && gp_ids[i].indexOf('mrna') != -1`, even if it is a bit more verbose.

adamabeshouse: oh interesting, good call. Yes I wholeheartedly agree with @fedde-s, dealing with -1 more explicitly is definitely easier for people in general (definitely me) to read.

pieterlukasse:
FYI @fedde-s is rebasing this and making some extra fixes on issues we found while testing this. He will submit this in a new PR soon.

jjgao:
superseded by #1847

PR #1628, opened 2016-08-29
OQL editor

shrumit:
This PR contains all the code for the new OQL editor, developed as part of GSoC 2016.


portal/src/main/webapp/WEB-INF/jsp/global/css_include.jsp
```diff
@@ -39,7 +39,7 @@
 <link href="css/responsiveslides.css?<%=GlobalProperties.getAppVersion()%>" type="text/css" rel="stylesheet" />
 <link href="css/tipTip.css?<%=GlobalProperties.getAppVersion()%>" type="text/css" rel="stylesheet" />
 <link href="css/jquery.qtip.min.css?<%=GlobalProperties.getAppVersion()%>" type="text/css" rel="stylesheet" />
-<link href="css/smoothness/jquery-ui-1.10.3.custom.css?<%=GlobalProperties.getAppVersion()%>" type="text/css" rel="stylesheet" />
+<link href="css/smoothness/jquery-ui-1.10.3.custom.css?20160703-1926" type="text/css" rel="stylesheet" />
```
adamabeshouse: what happened here?

core/src/main/java/org/mskcc/cbio/portal/util/StringUtilsLevenshtein.java
```diff
@@ -0,0 +1,145 @@
+
```
adamabeshouse: is Apache Commons not already accessible from the project? Can you not just import this method directly from Apache Commons, instead of adding this file in?

portal/src/main/webapp/WEB-INF/jsp/global/js_include.jsp
```diff
@@ -65,6 +65,12 @@
 <script type="text/javascript" src="js/lib/plotly.min.js?<%=GlobalProperties.getAppVersion()%>"></script>
 <script type="text/javascript" src="js/lib/igv_webstart.js?<%=GlobalProperties.getAppVersion()%>"></script>
 <script type="text/javascript" src="js/lib/genomespace.js"></script>
+<script type="text/javascript" src="js/lib/handlebars.min.js"></script>
+<script type="text/javascript" src="js/lib/rangy-core.js"></script>
+<script type="text/javascript" src="js/lib/rangy-selectionsaverestore.js"></script>
+<script type="text/javascript" src="js/lib/rangy-textrange.js"></script>
+<script type="text/javascript" src="js/lib/jquery.caret.js?<%=GlobalProperties.getAppVersion()%>"></script>
```
adamabeshouse: are all of these libraries used and not replaceable by home-rolled code? It's a lot to add to the global include, if we can avoid it.

portal/src/main/webapp/WEB-INF/jsp/step4_json.jsp
```diff
@@ -59,27 +59,36 @@
...
         $(document).ready(function() {
             GeneSymbolValidator.initialize();
+            OqlMenu.initialize();
         });
     </script>
+    
+    <div contenteditable="true" id="gene_list" placeholder="Enter HUGO Gene Symbols or Gene Aliases" title="Enter HUGO Gene Symbols or Gene Aliases" spellcheck="false"><%
+        if (localGeneList != null && localGeneList.length() > 0) {
+	        String geneListWithSemis =
+			        org.mskcc.cbio.portal.oncoPrintSpecLanguage.Utilities.appendSemis(localGeneList);
```
adamabeshouse: What are Semis? We're aiming to phase out all backend OQL code so it would be better to do this on the frontend.

portal/src/main/webapp/css/oql-editor.css
```diff
@@ -0,0 +1,195 @@
...
+
+#gene_list:empty:not(:focus):before{
+	color: #898989;
+	content:attr(placeholder);
+}
+
+
+/* OQL Menu */
+
+.hidden-menu {
```
adamabeshouse: All of these class names seem to me too general, I think they should all be prefixed by `oql-editor` e.g. `.oql-editor-hidden-menu`

portal/src/main/webapp/js/src/dynamicQuery.js
```diff
@@ -781,16 +781,17 @@ function caseSetSelected() {
 function geneSetSelected() {
     //  Get the selected ID from the pull-down menu
     var geneSetId = $("#select_gene_set").val();
-
-    if (window.metaDataJson.gene_sets[geneSetId].gene_list == "") {
+	console.log('geneSetSelected');
```
adamabeshouse: delete debugging code

portal/src/main/webapp/js/src/oql-menu.js
```diff
@@ -0,0 +1,304 @@
...
+var OqlMenu = (function($){
+	var $TA, $ME, lineHeight, gene_list;
+	var inState = false;	// whether menu should be relaunched on space
+	
+	// Generates OQL menu based on model and inserts into page
+	var initialize = function(){
+		gene_list = document.getElementById('gene_list');
+		rangy.init();
+		
+		console.log('oql init');
```
adamabeshouse: take out debugging code

adamabeshouse:
Looks great @shrumit. Just a few comments
(1) As I put in a few places where I saw it, make sure to take out any console logging you were using for debugging.

(2) Please implement tests, you can even make one file that goes in the same folder as oql-editor (if it's not in its own folder, it should be) that we can run using command line `node`.

(3) I may be wrong but it looks like you didn't change the way that gene replacements are handled?

Related to (3), could you please deploy this on Heroku - just ask someone on Slack to do it for you, then it will get automatically redeployed whenever you update the branch.

Great work!

adamabeshouse:
image

Bug: if the OQL editor is open then mouseover a bad gene doesn't show the gene replacement panel.

adamabeshouse:
Feature: when you select a mutation type, you don't need `MUT =` so it can just be, for example, `TRUNC` instead of `MUT = TRUNC`

adamabeshouse:
Bug: the default EXP setting needs to be made explicit: `EXP <= 2 EXP >= -2`. Just `EXP` is as of now not valid OQL

jjgao:
@shrumit thanks for the pull request. Would you please create another one and point to rc instead of master?

shrumit:
@jjgao Sure I'll do that. I'm still working out some final details.

PR #1629, opened 2016-08-30
Ensure that exit code is set properly

sambrightman:
Currently both tools return 0 if problems occur but they do not
ultimately have an uncaught exception. For example, if metaImport.py
calls main_validate() and gets return code != 0, it will exit with
success despite printing an error message. validateData.py does the same
when run standalone.

Without this PR, it is hard to programmatically verify success, as
mentioned in #848.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -2898,3 +2898,4 @@ if __name__ == '__main__':
                 1: 'failed',
                 2: 'not performed as problems occurred',
                 3: 'succeeded with warnings'}.get(exit_status, 'unknown')))
+    sys.exit(exit_status)
--- a/core/src/main/scripts/importer/metaImport.py
+++ b/core/src/main/scripts/importer/metaImport.py
@@ -113,6 +113,7 @@ if __name__ == '__main__':
                 print >> sys.stderr, Color.BOLD + "Overriding Warnings. Importing study now" + Color.END
                 print >> sys.stderr, "#" * 71 + "\n"
                 cbioportalImporter.main(args)
+                exitcode = 0
             else:
                 print >> sys.stderr, Color.BOLD + "Warnings. Please fix your files or import with override warning option" + Color.END
                 print >> sys.stderr, "#" * 71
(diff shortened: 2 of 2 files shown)


sambrightman:
I didn't actually check this yet with successful import, will do so now.

sambrightman:
Yes, so with warnings ignored of course this still produced a non-zero exit code. Although the client could check this, it seems like [poor style](http://programmers.stackexchange.com/questions/153926/non-zero-exit-status-for-clean-exit) to claim error status when client explicitly requested a re-definition of success. Will update the exit code   for `metaImport.py` and leave it for `validateData.py` - that way clients wishing to use a two-step process can discover the warnings before choosing to override them.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2898,3 +2898,4 @@ def main_validate(args):
                 1: 'failed',
                 2: 'not performed as problems occurred',
                 3: 'succeeded with warnings'}.get(exit_status, 'unknown')))
+    sys.exit(exit_status)
```
pieterlukasse: the metaImport calls the validateData, so in this case we would like to *not* have a sys.exit (especially in case of status == 0). Maybe only do this if status != 0 ?

pieterlukasse:
@sambrightman thanks for the details. The code looks OK and makes sense to me :+1: 

I made one small comment. Can you please check this?

core/src/main/scripts/importer/validateData.py
sambrightman: `metaImport.py` imports and calls the function directly. The code above is only run when `validateData.py` is run as a standalone script.

pieterlukasse: ah yes, please ignore my comment ;)

pieterlukasse:
@priti88 you might want to test this as well. Can you please check this and then ask @jjgao to merge if all is OK?

fedde-s:
Good! Looks like those exit statements were mostly pretty obvious and simply not considered or forgotten. As for non-zero exit statuses on successful use being poor style: I wouldn’t say that in general given that it occurs in tools as central to Unix as grep (1 if no selected lines were found) and diff (1 if the inputs are different), but since metaImport.py is about importing, I agree with your choice there!

PR #1513, opened 2016-07-20
Extra unit tests and data loader integration tests

pieterlukasse:
# What? Why?
This PR includes more checks in the `TestIntegrationTest.java` increasing its coverage of the different data types supported by the data model (continuing what was started in #1399). It also includes more unit tests for clinical data loading and mRNA expression data loading. 

This PR also includes some small fixes in two other test classes to try to prevent failures due to execution order of the unit tests. 

Code and/or comments clean-up was done where applicable in `ImportClinicalData.java`. Because we now make test assertions on warnings given by the loader, it was important to add the option to reset the warnings list in `ProgressMonitor.java`

Changes proposed in this pull request:
- see commit list

# Notify reviewers
@ersinciftci @morungos @ecerami @sheridancbio

PR #1633, opened 2016-09-01
Created the web layer for the new REST API design

ersinciftci:
Sorry for the huge PR.

# What? Why?
Created the Spring MVC web layer for the new REST API design.

- Java version changed to 8
- Added controllers for all core resource types.
- Added new models to support new controllers.
- Added `summary` parent models to show the summary view in response types in Swagger UI, instead of showing the detailed response, because the default projection is `SUMMARY`.
- Converted back some changes we made in Sprint 2 of Rearchitecture effort, so `/geneticprofiledata` API is using the old stack now, because it's a lot easier to design the new REST API/models without thinking about backward compatibility.
- Added new exception types under `service` module as an example, will add more when needed.
- Added exception handling for controllers.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
everybody in the core team:
@cBioPortal/core-developers


sheridancbio:
I'm beginning a review of this PR now. From the description, I'm wondering how rolling back the Rearchitecture Sprint2 work helps make it easier to move forward. There will be backwards compatibility issues either way I would think. The main changes in Sprint 2 were to make the servlets which depended on DaoMutations call the new persistence stack instead of DaoMutations. I thought that with the strangler vine approach we would not be worrying about rewriting servlets, but would link new frontend code through a new webapi to new backend handlers (initially similar to the mutation mybatis approach). I guess one reason why might be that with the servlets dependent on the new persistence layer, we must keep the equivalent functions in the mybatis code to support the needs of the servlets ... and that clashes with a desired new approach to the persistence layer? (Maybe optimizations for projections or some other general new feature)  I'll have more to say when I get into the code.

inodb:
@ersinciftci Thanks for this!
- [ ] Please change java version in `docker-compose.yml` to 8
- [ ] This page is failing as well with some Null pointer exception in `getGeneSymbol`: http://localhost:8080/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04

    ```
SEVERE: Servlet.service() for servlet [MutationsJSON] in context with path [] threw exception
        at org.mskcc.cbio.portal.model.ExtendedMutation.getGeneSymbol(ExtendedMutation.java:910)
        at org.mskcc.cbio.portal.servlet.MutationsJSON.processGetMutationsRequest(MutationsJSON.java:300)
        at org.mskcc.cbio.portal.servlet.MutationsJSON.processRequest(MutationsJSON.java:126)
        at org.mskcc.cbio.portal.servlet.MutationsJSON.doPost(MutationsJSON.java:786)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:648)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:729)
```

**Comment by ersinciftci** [comment]
@sheridancbio Sorry I didn't mean the whole Sprint 2, just the part that we connected `ApiController`'s `/geneticprofiledata` API to the new stack, so I disconnected it because it forces our mixins to flat all the JSON, and also caused some model property names/types not match with DB column names/types, because different names/types are used in front-end code.

File: model/src/main/java/org/cbioportal/model/ClinicalAttribute.java [hunk 1]
```diff
@@ -0,0 +1,61 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+
+public class ClinicalAttribute implements Serializable {
```
anthony-cros: Would you consider using lombok's [@Data](https://projectlombok.org/features/Data.html)? lombok adds some overhead to the build process but overall it does a great job a cutting down the java boilerplate. (PS: there's also a immutable counterpart to `@Data` called `@Value`)

ersinciftci: I like lombok but every technology/library we introduce makes the project a little bit more complex, so I prefer keeping it simple/pure java for now, but yeah maybe later.

ersinciftci:
Added a few more commits; updated end-to-end test screenshots because Java 8's collection sort algorithm is different from Java 7's, so there were some sorting differences in tables. And I realized that I can make summary model classes abstract since they are not supposed to be used directly,
they are there just to make swagger-ui response types simpler, since default projection is summary.

model/src/main/java/org/cbioportal/model/ClinicalAttribute.java
anthony-cros: Fair enough!

sheridancbio: I need to read up on lombok (first time I have heard about this) .. but in general I agree that we should be slow to add more frameworks that are not widely known to the development team. If we want to adopt a framework we should also not apply it in just one place, but across the codebase .. to keep the code approach consistent. (some frameworks might only apply to one area of course)

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/MutationMyBatisRepositoryTest.java
```diff
@@ -145,7 +145,6 @@ public void getMutationsWithAll() throws Exception {
 
 		List<Mutation> result = mutationMyBatisRepository.getMutations(sampleIds, entrezGeneIds, 6);
 		Assert.assertEquals(1, result.size());
-		Assert.assertNotNull(result.get(0).getMutationEvent().getGene());
```
sheridancbio: I'm guessing the rationale for dropping this is that a mutation must be associated with a mutationEvent and a mutationEvent must have a Gene (not null) ?  so this test cannot possibly fail?

sheridancbio:
I have read through half of the PR at this point and need to get back later. But here are my  accumulated comments up to this point (two early points were just about code style, and so I moved them to the end since they are less important):

In ApiService, the function call addSampleListIdToMutationList() has been dropped. That is fine, but we should test that the oncoprint still displays mutations properly because this was added to echo back the Sample List argument back in the response even though it was not part of the model. Looking at the rest of the code in this class, it seems like a revert, which should be fine (test needed though)

In ProfileDataMapper.xml it has been changed to allow a return of all mutations/copy number changes (before only those matching the supplied gene list were displayed). This is fine if the service layer call supplying /api/geneticprofiledata returns a null response when there is no supplied gene list. If the intention is to change the behavior of the API endpoint, there should be a discussion about whether a gene list should be a required parameter. Perhaps with paging results it does not need to be required, but some thought should be given. Making a parameter required later when not required now will be a non-backwards compatible api change (major change revision)

Embedded types are removed from model classes. Model classes are split up to correspond to profile levels (summary, full) : seems reasonable, but this organization is relevant mostly to the presentation layer (services) rather than the persistence layer. It is true that the non-summary classes are still first class objects that be used everywhere, but I continue to suspect that we will eventually want separate models for persistence interface and for presentation layer interface.

model/src/main/java/org/cbioportal/model/CopyNumberSegment.java  copy number segments are being modeled .. I hope this does not clash with the work Jiaojiao has been doing on this.

Gene class contains a List<String> of gene aliases. This is an embedded class object ... maybe this pulls out all aliases from an alias table which match the gene id? Reading/Writing may work differently or require special code since this does not match database tables? Do we want to stick to POJOs?
Other classes with non-flat embedded objects: PatientSummary.java has List<ClinicalData> member

SampleType enum is moved inside of SampleSummary class (not necessary, but ok. downside: wind up with complex class filenames)

MutationMapper.xml switches from Sample.stable_id to Sample.internal_id for "sampleId" ... but do we want to expose our internal ids to the outside world?

CODE STYLE points: (these are based on the belief that we want to use google style guide for java code)

Java argument names should be in camelcase (ApiService.java has public List<DBProfileData> getGeneticProfileDataBySample(List<String> genetic_profile_ids,
In standard Java style the argument name would be geneticProfileIds.
Maybe we want to make a formal exception to the standard java code style when the arguments are Api parameters which have underscore separators?

Google java style forbids import statements with wildcards (https://google.github.io/styleguide/javaguide.html) See the change in ApiService.java includes

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -341,33 +326,107 @@ else if(type.equals("frequency"))
...
+
+	@Transactional
+	public List<DBProfileData> getGeneticProfileData(List<String> genetic_profile_ids, List<String> genes) {
+		return getGeneticProfileData(genetic_profile_ids, genes, null, null);
+	}
+
 	@Transactional
-	public List<Serializable> getGeneticProfileData(List<String> geneticProfileStableIds, List<String> hugoGeneSymbols,
-													List<String> sampleStableIds, String sampleListStableId) {
+	public List<DBProfileData> getGeneticProfileDataBySampleList(List<String> genetic_profile_ids, List<String> genes, String sample_list_id) {
```
anthony-cros: just a style suggestion: I find putting one parameter per line really helps with readability (unless said parameters are very short/basic)

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -341,33 +326,107 @@ else if(type.equals("frequency"))
...
+		return getGeneticProfileData(genetic_profile_ids, genes, null, sample_list_id);
```
anthony-cros: no way to turn that null into an `Optional`?

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -341,33 +326,107 @@ else if(type.equals("frequency"))
...
+		return getGeneticProfileData(genetic_profile_ids, genes, sample_ids, null);
+	}
+
+	@Transactional
+	public List<DBProfileData> getGeneticProfileData(List<String> genetic_profile_ids, List<String> genes, List<String> sample_ids, String sample_list_id) {
+		List<DBGeneticProfile> profiles = getGeneticProfiles(genetic_profile_ids);
+		List<String> mutation_profiles = new ArrayList<>();
+		List<String> non_mutation_profiles = new ArrayList<>();
+		for (DBGeneticProfile p : profiles) {
+			if (p.genetic_alteration_type.equals("MUTATION_EXTENDED")) {
```
anthony-cros: if "MUTATION_EXTENDED" is used throughout the code base (I haven't check), I'd recommend making it part of an enum

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -341,33 +326,107 @@ else if(type.equals("frequency"))
...
+			ret.addAll(to_add);
 		}
-		if (!nonMutationProfiles.isEmpty()) {
-			result.addAll(getNonMutationGeneticProfileData(nonMutationProfiles, hugoGeneSymbols, sampleStableIds,
-					sampleListStableId));
+		if (!non_mutation_profiles.isEmpty()) {
+			List<DBGeneticAltRow> genetic_alt_rows = profileDataMapper.getGeneticAlterationRow(non_mutation_profiles, genes);
+			List<DBProfileDataCaseList> ordered_sample_lists = profileDataMapper.getProfileCaseLists(non_mutation_profiles);
+
+			Set<String> desired_samples = new HashSet<>();
```
anthony-cros: i'd recommend breaking down the rest of the method into more fine-grained methods, not so much for reusability but to make the intent clearer (comments would do too but I prefer re-using the language's modularization constructs whenever possible)

core/src/main/java/org/mskcc/cbio/portal/model/converter/MutationModelConverter.java
```diff
@@ -45,8 +45,8 @@ public ExtendedMutation convert(Mutation mutation) {
         if (mutationEvent.getEndPosition() != null) {
             event.setEndPosition(mutationEvent.getEndPosition());
         }
-        if (mutationEvent.getAminoAcidChange() != null) {
-            event.setProteinChange(mutationEvent.getAminoAcidChange());
+        if (mutationEvent.getProteinChange() != null) {
```
anthony-cros: all these if-nulls feel very anti-patterny, maybe mutation events should be modelled differently? Could it have several subclasses, or maybe more of an EAV-like lookup like the one for clinical attributes?

I understand this may not be the focus of this PR though, it's just to put it out there!

model/src/main/java/org/cbioportal/model/ClinicalAttribute.java
anthony-cros: it really paid of for us at ICGC but it is definitely not without headaches at time. Still, it has a few annotations that truly make developing in java less painful (notably the `@Value`, `val` and the `@XXXConstructor` ones)

model/src/main/java/org/cbioportal/model/summary/MutationSummary.java
```diff
@@ -0,0 +1,259 @@
+package org.cbioportal.model.summary;
+
+import java.io.Serializable;
+
+public abstract class MutationSummary implements Serializable {
```
anthony-cros: is the idea to move some of the logic from the services to this model objects eventually? In their current form they are pretty [anemic](http://www.martinfowler.com/bliki/AnemicDomainModel.html)

web/src/test/java/org/cbioportal/web/api/ApiControllerTest.java
```diff
@@ -543,7 +544,7 @@ public void geneticProfileDataTest4() throws Exception {
         GeneticProfile geneticProfile_brca_tcga_mutations = new GeneticProfile();
```
anthony-cros: These feel like they'd be more easily populated external data (like JSON files)

anthony-cros:
I made a few comments, mostly suggestions. Sorry I can't speak too much for the semantics as I'd need to be more familiar with the inner details of the application. Some comments are more a matter of style so feel free to pick and choose!

core/src/main/java/org/mskcc/cbio/portal/model/converter/MutationModelConverter.java
sheridancbio: Unless the constructor for MutationEvent is doing something unexpected, then the "event" object has all fields starting out as null values. It seems harmless to not test for null return values from getEndPosition(), etc .. just assign the value without a test. If you are assigning a null, you are not changing anything. If these objects are not supposed to have null values in data members, then that's another story --- but if that is true then we *still* should not need to test for null because we would have assured the fields are properly initialized in the constructor and were not null.

model/src/main/java/org/cbioportal/model/summary/MutationSummary.java
sheridancbio: I think a fundamental problem we have is that we currently have only a single model which is used both for the persistence layer transactions and also for the service/presentation layer functions. If we put too much domain knowledge into the model classes, there are inappropriate / distracting functionality there when using the model classes during development of the persistence layer. For mocking and unit testing, I think it is best to keep the lower level behaviors (if any) separate from the higher level behaviors. That way functionality can be easily tested in isolation. Also, embedded objects may be more appropriate to model at the higher level (perhaps a model of a genetic profile would contain mapping of mutation objects for example) but such an embedded collection might not be desired at the persistence level. If the added behaviors (more complex ones than getters and setters) have assumptions about which processing level will be calling them and what required initialization is needed, it is possible that these functions will be called prematurely.

web/src/test/java/org/cbioportal/web/api/ApiControllerTest.java
sheridancbio: Definitely yes. I created this ugliness (first time using mockito) ... and there must be a better way of specifying your mock objects than hand constructing them like this.

model/src/main/java/org/cbioportal/model/summary/MutationSummary.java
anthony-cros: Indeed, there's already the ORM impedance mismatch problem to deal with for the persistence model. The actual domain model needs to be completely decoupled from the shortcomings of the persistence layer (and I would recommend a Repository layer in between, so the backend can be abstracted away), this ought to help turning the very fat services like `ApiService` into the thinest layer possible, for the very limited logic that truly does not belong in the objects themselves (if any).

core/src/main/java/org/mskcc/cbio/portal/model/converter/MutationModelConverter.java
anthony-cros: +1 on checking it as early as possible (constructor) and not checking later. The few fields that may indeed be null can be turned into Optionals so it can be properly handled by the type system

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/MutationMyBatisRepositoryTest.java
ersinciftci: Since mutationEvent itself isn't a resource that we would expose by itself through an API, it's always part of mutation resource, I removed the Gene property from mutationEvent since mutation class already has it.

model/src/main/java/org/cbioportal/model/summary/MutationSummary.java
ersinciftci: I read Martin Fowler's `Patterns of Enterprise Application Architecture` book a few weeks ago, and I find Transaction Scripts in Service Layer is a good fit for cBioPortal because the domain logic is not that complicated. Martin Fowler says he doesn't like it but he admits that there are other experts that like it, he says he is an `Object Bigot` but I prefer being pragmatic. There is always a trade-off when you choose a technique over another one, and I try not to be a perfectionist and keep the project simple, easy to understand by everyone in the group. And that's why I'm trying to use the same models for every layer, because I worked in lots of projects that had different models for different layers, those models were very similar but you needed deal with lots of boilerplate code to convert one model to another (I know there are tools like http://mapstruct.org/, but still). Since we have Mixins that we can use to modify the presentation, and since we use MyBatis and manually map the columns to properties(and it's better to have embedded objects if we want to join tables and get extra data), I don't see a problem with using the same models everywhere for now.

ersinciftci:
Thanks for the comments, I replied the ones that involves the new code, but the changes in classes in `org.mskcc.cbio.portal` package (like `ApiService`) is just a revert to the old code, I agree with all your comments on those ones, but most of the code is in that way anyway so I'm not planning to  improve those parts.

ersinciftci:
@sheridancbio I'll reply your long comment in the day.

ersinciftci:
@sheridancbio 
Firstly, yes we should test the part that I reverted back. I did a basic UI test and it seemed fine but more testing is needed.

The change I made in `ProfileDataMapper.xml` was just to make Pieter's integration test happy, ideally every `foreach` in MyBatis scripts should have empty and null check because `IN` clauses throw SQL error if the list is empty or null, but none of the queries in `business` module has it. So I applied a best practice just to one place to make the integration test happy, but it would be a waste of time to fix them all since they are working fine for now and we are not going to reuse those XML mappings in `business` module anyway.

Those summary models are just needed to show the summary response in swagger-ui instead of full embedded response, because SpringFox doesn't support `@JsonView` yet https://github.com/springfox/springfox/issues/563 . So it's not actually required by presentation layer itself, just needed for presentation layer's documentation. In presentation layer we already have `super.setSerializationInclusion(JsonInclude.Include.NON_NULL)` in our `CustomObjectMapper`, so we don't need those summary models for projections, it's just for swagger-ui until SpringFox adds @JsonView support. And since those summary models are abstract, we won't actually use them anywhere in the code.

For some models I decided to add List type properties, like Patient and Gene, because those data wouldn't worth creating an API, like gene aliases, instead of having a gene alias API, I just return aliases for a gene if the projection is DETAILED. And return Patient and Sample clinical data if the projection is DETAILED, instead of having separate API. Because I see those resource as an extra information of another resource, instead of an individual resource.

All internal ids are `@JsonIngore`d in mixins, so we are not exposing them.

You are right about the styling issues but it's in the code that I reverted back, so they are not actually new code, and most of the old code is in that way anyway, so I'm not touching it for now.

pieterlukasse:
hi @sheridancbio the PR #1513 has some extra tests that maybe need some refactoring based on this code. Shall we merge that one first and update this one, or the other way around?

sheridancbio:
@pieterlukasse My first thought is to try to merge outstanding PR's first, and make adjustments and new tests during the merge of this PR. (i.e. not make other PRs that are ready now wait for this PR to be merged) .. But any changes to current PR's which move them in the right direction to be ready for the changes in this PR may save work later. But maybe @ersinciftci has comments about the best ordering of PRs (we really should push to clear a bunch of them this week I hope)

adamabeshouse:
Is it on purpose that there's no way to query clinical data by clinical attribute? That is a feature we use in the oncoprint now that seems to be missing.

ersinciftci:
@sheridancbio I generally prefer the least recently modified PR to get merged first, so @pieterlukasse 's PR can get merged and I can update this one. @adamabeshouse you are right, I just didn't think of a use case for it but "Clinical Tracks" in OncoPrint actually needs it, so I'll add it to the next iteration with other APIs I'm going to add, because this PR is pretty big already.

pieterlukasse:
@sheridancbio @ersinciftci thanks. I agree it would be preferable to merge older PRs first (and actually have this as a general policy - unless it is an urgent fix to hotfix).

kalletlak:
@ersinciftci, I was getting the following error while querying for http://www.cbioportal.org/api-design/api/studies?study_ids=hnsc_tcga

0a6b5744-76a4-11e6-8139-918cb5bc08a9

ersinciftci:
@kalletlak thanks, I didn’t see that old API uses the same path `studies` too, need to change mine then, probably to `cancer-studies` :)

ersinciftci:
We have the same problem with `/genes` too. I'll think of a way.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
```diff
@@ -28,6 +80,5 @@ public CustomObjectMapper() {
         mixinMap.put(Sample.class, SampleMixin.class);
         mixinMap.put(TypeOfCancer.class, TypeOfCancerMixin.class);
         super.setMixInAnnotations(mixinMap);
-        super.configure(JsonGenerator.Feature.WRITE_NUMBERS_AS_STRINGS, true);
```
JiaoJiao123: @ersinciftci @sheridancbio I think this might cause another issue besides the data type mismatch in unit test. Did you check is there anyone who is using the API and expect values to be string? After we remove WRITE_NUMBERS_AS_STRINGS setting, it will return number, which might make break some current implementation.

model/src/main/java/org/cbioportal/model/PatientClinicalData.java
```diff
@@ -0,0 +1,25 @@
+package org.cbioportal.model;
+
+import org.cbioportal.model.summary.ClinicalDataSummary;
+
+public class PatientClinicalData extends ClinicalDataSummary {
+
+    private Patient patient;
+    private ClinicalAttribute clinicalAttribute;
```
n1zea144: Do you have the potential for duplicating a lot of ClinicalAttribute data if you return a collection of PatientClinicalData?

model/src/main/java/org/cbioportal/model/SampleClinicalData.java
```diff
@@ -0,0 +1,25 @@
+package org.cbioportal.model;
+
+import org.cbioportal.model.summary.ClinicalDataSummary;
+
+public class SampleClinicalData extends ClinicalDataSummary {
+
+    private Sample sample;
+    private ClinicalAttribute clinicalAttribute;
```
n1zea144: Do you have the potential for duplicating a lot of ClinicalAttribute data if you return a collection of SampleClinicalData?

model/src/main/java/org/cbioportal/model/PatientClinicalData.java
ersinciftci: This non-Summary model classes are for DETAILED projection, and the default projection is SUMMARY, so it will be duplicated only if users want to, they will probably use DETAILED projection in some special cases. But it's good to have that option.

model/src/main/java/org/cbioportal/model/SampleClinicalData.java
ersinciftci: This non-Summary model classes are for DETAILED projection, and the default projection is SUMMARY, so it will be duplicated only if users want to, they will probably use DETAILED projection in some special cases. But it's good to have that option.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
ersinciftci: I tested, it doesn't seem to break, and I think think anyone else is using it.

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -134,6 +134,20 @@
...
+        <servlet-name>rest</servlet-name>
+        <servlet-class>org.springframework.web.servlet.DispatcherServlet</servlet-class>
+        <init-param>
+            <param-name>contextConfigLocation</param-name>
+            <param-value>classpath:applicationContext-rest.xml</param-value>
+        </init-param>
+        <load-on-startup>1</load-on-startup>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>rest</servlet-name>
```
n1zea144: As noted, we will do away with rest root endpoint when code is merged from api-design to rc.

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -134,6 +134,20 @@
...
+        <url-pattern>/rest/*</url-pattern>
```
n1zea144: As noted, we will do away with rest root endpoint when code is merged from api-design to rc.

n1zea144:
Note - code merged with Visual Regression failures - due to changes in OncoKB annotation and minor "pixel shift" discrepancies.

model/src/main/java/org/cbioportal/model/summary/MutationSummary.java
anthony-cros: Hey Ersin, sorry I missed that response of yours somehow!

I would contend that while the domain logic is not overly complex here indeed, it's not trivial either (for instance consider the EAV-like `clinical_attribute` table). I would really recommend shielding a core domain logic from that kind of low-level persistence considerations*. Right now these transaction scripts are highly coupled and display poor cohesion because they need to worry about both duties (though I only looked at a small subset of the codebase, so this is to be taken with a grain of salt).

I'm not familiar with mapstruct, but I know dozer which I believe plays a similar role. This tools help indeed, though really they just compensate for the weaknesses of the language. Indeed Java beans are notoriously painful to transform. I would however not let that stand in the way having a properly layered architecture... Admittedly these are the kind of issues that made me move away from java altogether though.

Just my two cents :)

*: IMO this is really more indicative of the need to use relation/document-store hybrid like postgres. Jumping into a full-blown document store like mongo would be overkill however, as the input file format is very relational in nature.

Issue #1541, opened 2016-07-26
SVG of survival plot won't open in Illustrator


schultzn:
This applies to the survival curve in query results.

error: The operation cannot complete because of an unknown error. [CANT].

This appears to be due to an Illustrator bug, but the solution can be found here:
https://github.com/CIRDLES/topsoil/issues/26

Replace sans-serif with Verdana in the SVG file and the file opens fine.

Can this be changed in the original SVG?

I am not sure if other SVG files are also affected

PR #1634, opened 2016-09-01
fixed issue #1541

yichaoS:
Onur fixed it.


onursumer:
Looks good to me. Maybe we should also add a one line comment above to state that this is a workaround for Adobe illustrator?

jjgao:
Good point on adding a comment. @yichaoS could you do that?

Issue #1622, opened 2016-08-25
Bar chart outlier issues


zhx828:
Please see screenshots. This is found in  http://www.cbioportal.org/study.do?cancer_study_id=hnsc_tcga#summary

![screen shot 2016-08-25 at 1 24 15 pm](https://cloud.githubusercontent.com/assets/5400599/17979177/561af994-6ac7-11e6-83b9-3ddf433192f8.png)
![screen shot 2016-08-25 at 1 24 18 pm](https://cloud.githubusercontent.com/assets/5400599/17979178/57e8826e-6ac7-11e6-8ae8-286cd15cea4c.png)
![screen shot 2016-08-25 at 1 26 25 pm](https://cloud.githubusercontent.com/assets/5400599/17979242/9c384b66-6ac7-11e6-8f61-8b6371827d07.png)

PR #1635, opened 2016-09-01
outlier bug fix

JiaoJiao123:
Bar chart outlier issues #1622


zhx828:
Please see screenshot study: hnsc_tcga
screen shot 2016-09-03 at 1 47 53 pm

zhx828:
Why show >2 if there is no case fall into this category? study: hnsc_tcga
screen shot 2016-09-03 at 1 49 16 pm

zhx828:
There is a potential issue about this bar chart. Please see screenshot, we probably need to find a way either remove NA or use log scale for y-axis. @jjgao what do you think? This is from mskimpact study.

screen shot 2016-09-03 at 9 10 03 pm

jjgao:
@zhx828 it is a good idea to provide an option to remove NA columns. Supporting log scale of y-axis for all bar chart is also a good idea. Another option would be supporting truncated y-axis.

JiaoJiao123:
@zhx828 remember the discussion we had earlier? if we want to make >max and NA optional, we have to loop through data twice. then we decided keep both of them.

zhx828:
@JiaoJiao123 no need to fix in this pull request. We should add this feature in iViz instead. I will add an issue to keep tracking of this idea.

zhx828:
screen shot 2016-09-12 at 11 20 06 am
 study: hnsc_tcga

PR #1640, opened 2016-09-06
New documentation about updating pfam_graphics table

oplantalech:
This is a new document about how to update the pfam_graphics table, which is not explained in the documentation. @pieterlukasse

### DIFF
--- a/docs/Updating-pfam_graphics-table.md
+++ b/docs/Updating-pfam_graphics-table.md
@@ -1,0 +1,18 @@
+# Updating pfam_graphics table
+
+Updating the pfam_graphics table is easy using the scripts provided in cBioPortal. There is no need to download any file previously.
+
+1- Empty the table `pfam_graphics` (the scripts raise an error if one of the domains that we want to add is already in the table). 
+```sql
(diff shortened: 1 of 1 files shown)


docs/Updating-pfam_graphics-table.md
```diff
@@ -0,0 +1,18 @@
+# Updating pfam_graphics table
+
+Updating the pfam_graphics table is easy using the scripts provided in cBioPortal. There is no need to download any file previously.
+
+1- Empty the table `pfam_graphics` (the scripts raise an error if one of the domains that we want to add is already in the table). 
+```sql
+TRUNCATE TABLE pfam_graphics;
+```
+
+2- Run the script `FetchPfamGraphics.java` with the file path of the output file and its name as an argument. This function first searches for all the proteins from mouse that have been reviewed in UniProtKB, then tries to retrieve their Pfam domains, and finally saves the information in a file. This script can also be run in the command line by going into  `<your_cbioportal_folder>/core/src/main/scripts/` and typing:
```
fedde-s: I don’t think should be about mice?

Issue #1637, opened 2016-09-06
Documentation refers to outdated, broken datasets only


fedde-s:
The documentation contains links to public datasets, on https://github.com/cBioPortal/cbioportal/blob/master/docs/Public-datasets.md .  These datasets seem to be outdated however, and do not follow the file formats required to load them into cBioPortal using the documented commands. I understand that correct, updated data is available on https://github.com/cBioPortal/datahub .  If that’s right, the documentation should refer there instead.

fedde-s:
@zheins, could you take a look at this?

fedde-s:
This has been solved by pull request #1650.

PR #1643, opened 2016-09-06
Update Public-datasets.md

zheins:
# What? Why?
Fixes Issue #1637 
Update the documentation for public available datasets to point to datahub.

# Notify reviewers
@fedde-s


zheins:
Potentially moving the tcga tarballs, need to wait for datahub pr to be reviewed and merged

jjgao:
Do we need to list all of them here? Maybe just point to datahub is enough so that we don't have to update this page for every datahub release.

fedde-s:
@jjgao: sounds good! And then maybe we can even find some other place for that link, such as at the end of the steps to get a portal installation up and running. But if not, we can just keep this ‘Public datasets’ page with a one-sentence paragraph that links to the datahub.

jjgao:
superseded by @1650

PR #1650, opened 2016-09-08
Updated documentation URL to point to new datahub location

pieterlukasse:
# What? Why?

- Updated to point to new datahub location. 
- The other details that I removed here are not necessary anymore since datahub project is easy to browse and github UI provides a clear "download" button.

Notify: @zheins


fedde-s:
I believe this to be a duplicate of pull request #1643.

fedde-s:
Although this one does exactly what @jjgao suggested in that PR, so we could merge this one and close the other.

PR #1657, opened 2016-09-09
Fix chrome/ff screenshot fail, break local test

inodb:
The new chrome image is slightly bigger on Travis (1px). Unfortunately this breaks
the end-to-end test locally, but better to have CI working at this stage. We should come up with a better solution.

Also replace FF screenshot, since IDH1 oncokb annotation went from level 3 to 1. 

# Reviewers
@adamabeshouse @jjgao @sheridancbio


adamabeshouse:
lgtm 👍

inodb:
Looks like the annotation for one of the oncokb mutations changed. The IDH1 went from Level 3 to Level 1. Is that correct? @zhx828  
screen shot 2016-09-11 at 8 42 05 pm

inodb:
Seems like it is a bug:
screen shot 2016-09-11 at 8 59 12 pm

zhx828:
@inodb thanks so much for pointing out. It should be related to the changes I made recently. I will take a look ASAP. You can merge this pull to make the local test working. I will update the screenshots once I fixed the issue.

inodb:
Thanks @zhx828 ! Merged it

sheridancbio:
This worked for my branch ... but I did need to run the tests twice, because in one run of the test, a tooltip/popup window location was shifted by about 80 pixels. On rerun, all tests passed.

inodb:
@sheridancbio :+1:

Yeah I'm still not sure why that tooltip comes up in the first place. But glad it worked.

PR #1664, opened 2016-09-12
Update rc screenshots 

zhx828:

PR #1665, opened 2016-09-12
API to fetch gene by hugo symbol

JiaoJiao123:
# What? Why?
This API is used to fetch gene entities by hugo symbol list, returned values including entrez_gene_id, hugo_gene_symbol, type, cytoband, and length about input hugo symbol.

#why?
In another API to fetch copy number segment data, we need to get chromosome info from hugo symbol. But there is no such services we can use to achieve this in current API list. 

- service to return all gene info from hugo symbol
- related unit tests

# Notify reviewers
@sheridancbio


zhx828:
No need to update screenshots. They will be updated in this pull https://github.com/cBioPortal/cbioportal/pull/1664

sheridancbio:
reviewing now

model/src/main/java/org/cbioportal/model/Gene.java
```diff
@@ -49,4 +51,26 @@ public Integer getLength() {
     public void setLength(Integer length) {
         this.length = length;
     }
+    
+    public String getChromosome() {
```
sheridancbio: Almost all of the new Model classes have no functionality aside from getters, setters, and maybe constructors.
Since we area using the model classes to serve the needs across all layers (persistence, service, web) I think we need to be careful about building in functionality which is not appropriate for all layers. But I think that these simple utility functions are not layer-specific, are generally helpful, and it would be worse to create some kind of "GeneChromosomeParser" class. One question though: some genes are located on both the X and Y chromosomes (see below). Should we distinguish between genes only on X, genes only on Y, and genes on both X and Y ?
mysql> select * from gene where cytoband like "%;%" and TYPE="protein-coding";
+----------------+------------------+----------------+------------------+--------+
| ENTREZ_GENE_ID | HUGO_GENE_SYMBOL | TYPE           | CYTOBAND         | LENGTH |
+----------------+------------------+----------------+------------------+--------+
|         286530 | P2RY8            | protein-coding | Xp22.33; Yp11.3  |   4459 |
|           6473 | SHOX             | protein-coding | Xp22.33;Yp11.3   |   6114 |
|           9189 | ZBED1            | protein-coding | Xp22.33;Yp11     |   4755 |
|           8225 | GTPBP6           | protein-coding | Xp22.33; Yp11.32 |   2363 |
|           8623 | ASMTL            | protein-coding | Xp22.3; Yp11.3   |   3506 |
|          28227 | PPP2R3B          | protein-coding | Xp22.33; Yp11.3  |   9015 |
|          55344 | PLCXD1           | protein-coding | Xp22.33; Yp11.32 |   6191 |
|          64109 | CRLF2            | protein-coding | Xp22.3; Yp11.3   |   1205 |
|         207063 | DHRSX            | protein-coding | Xp22.33; Yp11.2  |   3620 |
+----------------+------------------+----------------+------------------+--------+

sheridancbio: I'm guessing that if we don't want to distinguish this, it is better to mark these as on X rather than Y ... since annotated it as Y might indicate that females would not have the gene, whereas marking it on X correctly indicates that all humans have the gene.

JiaoJiao123: @sheridancbio Thank you for bring this up. I didn't know there are genes on both X and Y. But I think in current get chromosome function, it will  recognize it as X since "Xp22.33; Yp11.3" only meet the condition that cytoband.toUpperCase().startsWith("X")

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+ * @author jiaojiao
+ */
+@Service
+public class GeneServiceImpl implements GeneService{
+    @Autowired
+    private GeneRepository geneRepository;
+    
+    @Override
+    public List<Gene> getGeneListByHugoSymbols(List<String> hugo_gene_symbol){
+        if(hugo_gene_symbol == null || hugo_gene_symbol.size() == 0)
```
sheridancbio: Style comments: add a space after keywords (like "if") .. and keep the opening  "{" on the same line as the if statement. add spaces around "else", and add a space between ")" and "{" in the function declaration line.

web/src/main/java/org/cbioportal/webLegacy/GeneController.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @author jiaojiao
+ */
+@RestController
+public class GeneController {
+    @Autowired
+    private GeneService geneService;
+
+    @RequestMapping(method = {RequestMethod.GET, RequestMethod.POST}, value = "/genes/fetchByHugoGeneSymbol")
+    public List<Gene> getGeneListByHugoSymbols(@RequestParam(required = true) List<String> hugo_gene_symbols){
```
sheridancbio: space before "{"

web/src/main/resources/applicationContext-web.xml
```diff
@@ -25,10 +25,10 @@
       <bean name="/applicationSwaggerConfig" class="org.mskcc.cbio.portal.swagger.ApiServiceSwaggerConfig">
       </bean>
 
-    <context:component-scan base-package="org.mskcc.cbio.portal.web"/>
-    <context:component-scan base-package="org.mskcc.cbio.portal.documentation"/>
-    <context:component-scan base-package="org.cbioportal.web"/>
-    <context:component-scan base-package="org.cbioportal.weblegacy"/>
+    <context:component-scan base-package="org.mskcc.cbio.portal.web" />
```
sheridancbio: I guess these additional spaces are harmless .. but why not be consistent? the following lines below with "<mvc:resources ... " do not have an extra space before the closing of the tag.

sheridancbio:
We decided to change the endpoint name from fetchByHugoGeneSymbol to /api/gene/fetch-by-hugo. Also some renaming in Business module was needed to make room for the new GeneMapper mybatis interface.

JiaoJiao123:
@sheridancbio do we need to do the renaming in business module in this PR? If so, could you give an instance about what to change?

model/src/main/java/org/cbioportal/model/Gene.java
sheridancbio: Ok, I'm putting in an issue into the issue tracker for discussion. for now, these will be on "X"

sheridancbio:
squash of commits was performed during merge

Issue #1160, opened 2016-05-02
importGeneData gene multiple lenghts error


aderidder:
I was looking into importing gene data using importGeneData.java. I took the NCBI genes file and I'm using gencode v24 annotation for the gene length (ftp://ftp.sanger.ac.uk/pub/gencode/Gencode_human/release_24/gencode.v24.annotation.gtf.gz).
When I run importGeneData, I'm getting errors that a gene has multiple lengths. I've looked into this and the reason seems to be that if a gene does not have its exons together, the code gives this error. For example:

```
X   6061643 6061749 SNORA31 .   -
X   38318828    38318990    RPGR    .   -
X   38317245    38317465    RPGR    .   -
X   38308111    38308241    SNORA31 .   -
```

RPGR is ok; SNORA31 gives the error. 

We're wondering why this is an error? Can't the length also be based on both exons in this case?

jjgao:
I looks like we assumed that exon annotations of the same gene are together in the file.

https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java#L184

We need to fix this.

pieterlukasse:
@jjgao I'm looking into this now

pieterlukasse:
@jjgao we checked this in more detail now. Gencode data has some strange stuff in it like: SNORA31 actually "has exons" in virtually every chromosome...and ends up with a total length of 3000 if we consider all of these, while in Entrez it is reported to have 1 exon and a length of ~80.... (see http://www.ncbi.nlm.nih.gov/gene/677814). Here are SNORA31 exons according to Gencode:
```
1       67102644        67102779        SNORA31 .       -
1       153012481       153012614       SNORA31 .       -
2       4827000 4827084 SNORA31 .       -
3       100646120       100646197       SNORA31 .       +
4       81928312        81928439        SNORA31 .       +
4       105105986       105106094       SNORA31 .       +
4       186130020       186130155       SNORA31 .       +
5       80306234        80306368        SNORA31 .       -
5       106546664       106546803       SNORA31 .       +
7       32791813        32791946        SNORA31 .       +
7       34892416        34892549        SNORA31 .       +
7       52269436        52269569        SNORA31 .       -
8       117313527       117313661       SNORA31 .       +
9       26796172        26796304        SNORA31 .       -
9       101322476       101322533       SNORA31 .       +
9       131021167       131021306       SNORA31 .       -
[...]

PR #1678, opened 2016-09-19
Improve calculation of gene length

oplantalech:
Following #1583, the calculation of gene length has been revised. Currently, a Perl command is necessary to parse the Gencode file (gtf) into a bed file, while the proposed code directly parses the Gencode file. Regarding the method itself, the proposed code uses Ensembl IDs to differentiate between different genes, which is more accurate than using the symbols (example #1160, where there are two different genes called SNORA31 in the same chromosome; in the Gencode file they have different Ensembl IDs) . 

If different Ensembl IDs point to the same symbol, the method only updates the length of the gene if the chromosome given in the gtf file matches the cytoband from the gene in the database. In the case multiple Ensembl IDs with the same symbol are on the same chromosome (like SNORA31 in #1160) or no cytoband information is available, the length of the last Ensembl ID is taken. Finally, a test for the method has been added, and also the code documentation has been improved. @pieterlukasse @sandertan


oplantalech:
Also, another improvement in the length calculation is that the end and start bases from the Gencode file are part of the sequence. For example, if for one gene the start base is 2 and the end is 10, the length of the gene is not 8, but 9.

pieterlukasse:
please also add the respective documentation on how to use the new script

oplantalech:
@pieterlukasse documentation added

pieterlukasse:
@oplantalech code looks good. Just some minor code documentation missing.

pieterlukasse:
I did some manual validation using `gencode.v24.annotation.gtf` and the new seed data that has gene lengths based on this fix and this looks good :+1:

Sjoerd-van-Hagen:
@jjgao 

Hi JJ,

This PR has been open for a long time now. Could you make sure it gets reviewed and merged?

Best,

Sjoerd.

pieterlukasse:
@Sjoerd-van-Hagen @jjgao I had reviewed it, but by mistake never sent the review comments somehow. Fortunately github stored them and they are now sent to @oplantalech

oplantalech:
@pieterlukasse I've included your suggestions in the new commit. Can you check again if all you requested has been taken into account?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -36,95 +36,133 @@
...
-                    }
-                    if (!strAliases.equals("-")) {
-                        aliases.addAll(Arrays.asList(strAliases.split("\\|")));
-                    }
-                    
-                    if (geneSymbol.startsWith("MIR") && type.equalsIgnoreCase("miscRNA")) {
-                        line = buf.readLine();
-                        continue; // ignore miRNA; process seperately
+                String parts[] = line.split("\t");
+                int taxonimy = Integer.parseInt(parts[0]);
```
n1zea144: This should be spelled "taxonomy"

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -137,34 +175,49 @@ public static void importData(File geneFile) throws IOException, DaoException {
...
+       
     }
-    
-    private static void logDuplicateGeneSymbolWarning(String symbol, Set<CanonicalGene> genes) {
+
+    private static void logDuplicateGeneSymbolWarning(String symbol, Set<CanonicalGene> genes, boolean isOfficialSymbol) {
         StringBuilder sb = new StringBuilder();
-        sb.append("More than 1 gene has the same symbol ")
-                .append(symbol)
+        if (isOfficialSymbol)
```
n1zea144: For consistency in style, I would put braces around this statement.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -222,6 +397,8 @@ private static int calculateGeneLength(List<long[]> loci) {
                 max = l[1];
             }
         }
+        if (max < min)
```
n1zea144: For style consistency I would add braces here.

pieterlukasse:
approved changes based on assumption changes were made. Hard to check since the whole PR has been squashed again...

pieterlukasse:
Were minor comments anyway

jjgao:
@oplantalech: there are some conflicts of this pull request with the current rc branch. Could you please create a pull request to rc with the same change set? Thanks.

oplantalech:
@jjgao The new PR based on RC is here: #2149

PR #1679, opened 2016-09-19
visualize copy number segment data with IGV javascript version

JiaoJiao123:
# What? 
Replace IGV Java version with IGV.js in the query result page IGV tab.
#why?
Current IGV java version is accessing copy number segment file from url which is unsecure and inconvenient.

- get copy number data through API 
- create an igv viewer for each gene in a seperate tag

# Notify reviewers
@zhx828

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/igv.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/igv.jsp
@@ -29,66 +29,68 @@
  - You should have received a copy of the GNU Affero General Public License
  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
 --%>
-
-<%@ page import="java.util.List" %>
-<%@ page import="java.util.ArrayList" %>
-<%@ page import="java.net.URLEncoder" %>
-<%@ page import="com.google.common.base.Joiner" %>
-<%@ page import="java.io.UnsupportedEncodingException" %>
-<%@ page import="org.mskcc.cbio.portal.model.GeneWithScore" %>
-<%@ page import="org.mskcc.cbio.portal.servlet.QueryBuilder" %>
-<%@ page import="org.mskcc.cbio.portal.util.IGVLinking" %>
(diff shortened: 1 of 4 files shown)


portal/src/main/webapp/WEB-INF/jsp/igv.jsp
```diff
@@ -35,60 +35,66 @@
...
+                + '<script type="text/javascript" src="js/lib/igv.js"><\/script></head>';
+            var bodyContent1 = '<body><div id="igvDiv" style="padding-top: 10px;padding-bottom: 10px; border:1px solid lightgray;width:100%"></div><script type="text/javascript">  $(document).ready(function () {    var div = $("#igvDiv")[0],   options = {'
+                + 'showNavigation: true, showRuler: true, genome: "hg19", locus: "' + hugoSymbol + '", tracks: [  { url: "api/copynumbersegments", indexed: false, name: "Segmented CN", type:"seg", json: true, method: "POST", ';
+            var bodyContent2 = 'cancerStudyId: "' + cancerStudyId + '", hugoSymbol: "' + hugoSymbol +'", sampleIds: "' + sampleIds + '"},{name: "Genes", url: "https://s3.amazonaws.com/igv.broadinstitute.org/annotations/hg19/genes/gencode.v18.collapsed.bed", order: Number.MAX_VALUE,  displayMode: "EXPANDED"}]};igv.createBrowser(div, options);});<\/script><\/body>';    
+            var fullContent = headerContent + bodyContent1 + bodyContent2;
+            return fullContent;
+        }
+        
+        $("#igv-result-tab").click(function(){
+            var genes = window.QuerySession.getQueryGenes();
```
zhx828: Is there possibility that 'genes' will be an empty array, null or undefined?

JiaoJiao123: I don't think so. User need to input gene to be able to get to the query page. But it is good to add a condition check just in case.

PR #1648, opened 2016-09-07
Structural Variant model / persistence / controller + unit tests & refactoring

sheridancbio:
Structural_Variant model / persistence / controller / unit tests
Other refactoring of core done and style standardization.
This is a correction / completion of PR #1609 

Changes proposed in this pull request:
    -set api endpoint to "structuralvariant" from "sv"
    -put controller into subpackage : org.cbioportal.web.legacy
    -added adjustments to cgds.sql, migration.sql, seed_mini.sql, testSql.sql [appropriate schema/test data]
    -dropped insert functionality from mybatis mapper and service layer
    -altered sv table sample field mapping to use internal_id (not stable). gene still uses hugo gene name
    -renamed classes with "SV" to have "StructuralVariant"
    -in sv table, changed some data types to be sensible (not int(255))
    -adjusted foreign key in cgds.sql (to internal)
    -reordered table drop statements to drop child tables before parents
    -controller test class moved to a parallel package path
    -added additional unit tests for repository
    -corrected logical operator precidence problem in mapper
    -style changes: drop tab characters and standardize indents (4 - java, 2 - xml)
    -added additional records to test data database construction script
    -reduce controller level tests : one "full" mock object and one empty (to test proper json formatting)
    -changed GeneticProfileType to enum class (with lookup by string value)
    -add copyright/license to top of all involved files

# Checks
- [ ] Runs on Heroku.
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Google Style Guide](https://github.com/google/styleguide).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)
- [ ] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to
  hotfix.

# Notify reviewers
If your pull request involves changes to an existing file, one or more of the
people that edited before you might be good candidates to review it. Please use
`git blame <filename>` to determine that and notify them here. Otherwise notify
everybody in the core team:
@n1zea144 @zheins @jjgao @jake-rose

PR #1683, opened 2016-09-19
Copy number segment data API and its unit test

JiaoJiao123:
# What? Why?
The basic usage of this API is to fetch copy number segment data and the details is specified in this [RFC ](https://docs.google.com/document/d/1mqqnyDeaWN0uTDuBU2bzUBFwetUWHX_PtHstCeCpnLk/edit#heading=h.pm7gdme3529l)
It is designed for [PR#1679 implementing IGV tab with its javascript version](https://github.com/cBioPortal/cbioportal/pull/1679). 

- query copy number segment data by cancer study, sample list and hugo symbol
- related unit test

# Notify reviewers
@sheridancbio 
@cBioPortal/backend


ecerami:
Do these API change mesh with the recent API work that @ersinciftci, @sheridancbio and others have been working on?  I just want to make sure that we are only pushing out API additions that mesh with the larger API strategy.

JiaoJiao123:
@ecerami It should be. I discussed with @sheridancbio about the API structure to use and etc. I think we are using sprint 2/3.

sheridancbio:
@ecerami @ersinciftci We are separating the api controllers into two packages .. org.cbioportal.web will contain controllers which have the REST design approach which @ersinciftci is leading the design on. Older controllers and controllers which are written initially in a non-REST idiom will go in org.cbioportal.weblegacy. This endpoint that @JiaoJiao123 has developed is needed now, and it is being deployed in the org.cbioportal.weblegacy package. There are two other data services that also being deployed at this time: https://github.com/cBioPortal/cbioportal/pull/1618 and https://github.com/cBioPortal/cbioportal/pull/1648/files. When the new API design effort has stabilized, we can convert these data services into the REST based interaction. The REST Api design effort is currently focussed on the patient-view functionality. It does have an api endpoint for genomic segment data defined. Possibly when it comes time to develop the backend implementation for the new REST endpoint we can simply convert the persistence layer and service layer code in this PR to provide the data services for the new endpoint.

pieterlukasse:
maybe good to add an integration test as well (once #1513 is merged...almost there)

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/CNASegmentMapper.xml
```diff
@@ -0,0 +1,26 @@
...
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
+<mapper namespace="org.cbioportal.persistence.mybatis.CNASegmentMapper">
+    <cache/>
+
+    <select id="getCNASegmentData" resultType="org.cbioportal.model.CNASegmentData">
+        select s.stable_id as sample, cns.chr as chr, cns.start as start, cns.end as end, cns.num_probes as numProbes, cns.segment_mean as value from sample s, copy_number_seg cns, cancer_study cs 
+        where cs.cancer_study_identifier = #{cancerStudyId} 
+        
+        <if test="sampleIds != null and !sampleIds.isEmpty()">
```
sheridancbio: Do we really want to allow a user to get the *complete* set of copy number segments for a study?  I guess it depends on how big this return of data could be. If it is many megabytes for a big study, perhaps we want to require a gene list or a sample list?

service/src/main/java/org/cbioportal/service/impl/CNASegmentServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Autowired
+    private CNASegmentRepository cnaSegmentRepository;
+    
+    @Autowired
+    private GeneRepository geneRepository;
+
+    @Override
+    public List<CNASegmentData> getCNASegmentData(String cancerStudyId, List<String> hugoGeneSymbols, List<String> sampleIds) {
+        List<String> chromosomes = new ArrayList<String>();
```
sheridancbio: wouldn't we expect duplicates here?  Shouldn't those duplicates be merged into a unique list?  I think Set<String> chromosomes = new Set<String>()  might be what we want.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
```diff
@@ -28,6 +26,5 @@ public CustomObjectMapper() {
         mixinMap.put(Sample.class, SampleMixin.class);
         mixinMap.put(TypeOfCancer.class, TypeOfCancerMixin.class);
         super.setMixInAnnotations(mixinMap);
-        super.configure(JsonGenerator.Feature.WRITE_NUMBERS_AS_STRINGS, true);
```
sheridancbio: we must test the built warfile and make sure all visualizations work after making this change.

web/src/main/java/org/cbioportal/weblegacy/CNASegmentController.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+import org.cbioportal.model.CNASegmentData;
+import org.cbioportal.service.CNASegmentService;
+
+@RestController
+public class CNASegmentController {
+
+    @Autowired
+    private CNASegmentService cnaSegmentService;
+
```
sheridancbio: Add some human readable descriptions with @ApiOperation and @ApiParam annotations -- see GenePanelController.java for examples

web/src/test/java/org/cbioportal/web/api/ApiControllerTest.java
```diff
@@ -32,11 +32,11 @@
 import org.springframework.test.context.web.WebAppConfiguration;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.request.MockMvcRequestBuilders;
-import org.springframework.test.web.servlet.result.MockMvcResultHandlers;
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers;
 import org.springframework.test.web.servlet.setup.MockMvcBuilders;
 import org.springframework.web.context.WebApplicationContext;
 
+@Ignore
```
sheridancbio: I had the impression that this went by the test methods .. not the entire controller test class

web/src/test/java/org/cbioportal/weblegacy/CNASegmentControllerTest.java
```diff
@@ -0,0 +1,118 @@
...
+        cNASegmentData1.setSample("TCGA-AG-3732-01");
+        
+        CNASegmentData cNASegmentData2 = new CNASegmentData(); 
+        cNASegmentData2.setStart(148347132);
+        cNASegmentData2.setEnd(148348416);
+        cNASegmentData2.setChr("7");
+        cNASegmentData2.setNumProbes(3);
+        cNASegmentData2.setValue((float) -1.5009);
+        cNASegmentData2.setSample("TCGA-AG-3732-01");
+
```
sheridancbio: two blank lines here is excessive --- consider having no blank line here or between other statements within this function.

web/src/test/java/org/cbioportal/weblegacy/CNASegmentControllerTest.java
```diff
@@ -0,0 +1,118 @@
...
+        CNASegmentData cNASegmentData2 = new CNASegmentData(); 
+        cNASegmentData2.setStart(148347132);
+        cNASegmentData2.setEnd(148348416);
+        cNASegmentData2.setChr("7");
+        cNASegmentData2.setNumProbes(3);
+        cNASegmentData2.setValue((float) -1.5009);
+        cNASegmentData2.setSample("TCGA-AG-3732-01");
+       
+        
+      	mockResponse.add(cNASegmentData1);
```
sheridancbio: this line should be indented the same as others

JiaoJiao123:
@pieterlukasse that's a good suggestion. But I might add the integration test later.

web/src/main/java/org/cbioportal/weblegacy/GeneController.java
```diff
@@ -49,8 +52,13 @@
     @Autowired
     private GeneService geneService;
 
+    @ApiOperation(value = "Get all information about genes from hugo symbols including entrezGeneId, hugoGeneSymbol, cytoband, length and chromosome",
+            nickname = "getGeneListByHugoSymbols",
+            notes = "")
+    @Transactional
     @RequestMapping(method = {RequestMethod.GET, RequestMethod.POST}, value = "/gene/fetch-by-hugo")
-    public List<Gene> getGeneListByHugoSymbols(@RequestParam(required = true) List<String> hugo_gene_symbols) {
+    public List<Gene> getGeneListByHugoSymbols(@ApiParam(value = "a list of gene hugo symbols that you want to query on")
```
sheridancbio: An example case .. or directions to the /api/gene endpoint would help the user understand better

web/src/test/java/org/cbioportal/weblegacy/CNSegmentControllerTest.java
```diff
@@ -0,0 +1,116 @@
...
+        cnSegmentData1.setSample("TCGA-AG-3732-01");
+        CNSegmentData cnSegmentData2 = new CNSegmentData(); 
+        cnSegmentData2.setStart(148347132);
+        cnSegmentData2.setEnd(148348416);
+        cnSegmentData2.setChr("7");
+        cnSegmentData2.setNumProbes(3);
+        cnSegmentData2.setValue((float) -1.5009);
+        cnSegmentData2.setSample("TCGA-AG-3732-01");
+       
+      	mockResponse.add(cnSegmentData1);
```
sheridancbio: indent

PR #1682, opened 2016-09-19
merge hotfix to rc

jjgao:

PR #1684, opened 2016-09-20
Fix exception for swissprot_identifier: name

fedde-s:
Due to an error in the code, `ImportProfileData` would crash when loading
mutation data files of which the metadata file explicitly specifies that
the SWISSPROT column contains names (which is how this column is parsed
by default if unspecified). This commit fixes the bug.

# Suggested reviewers
@jjgao


fedde-s:
@jjgao, this fixes one of the issues you noticed in #1682, it was actually a bug in hotfix that was covered by a test in rc, so it is also relevant to hotfix. Can you merge it? Then after that we can merge hotfix into your rc branch again.

PR #1687, opened 2016-09-20
setting default value of empty string for property session service url

### LINKED ISSUE #1685: No default for sessions.service.url in GlobalProperties.java
There is no default for sessions.service.url in `GlobalProperties.java` which causes the portal to crash on current rc.

This should in the future be caught in review process:

https://github.com/cBioPortal/cbioportal/blob/hotfix/CONTRIBUTING.md#devops

```
SEVERE: Exception sending context initialized event to listener instance of class org.springframework.web.context.ContextLoaderListener
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'proxyController': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire method: public void org.mskcc.cbio.portal.web.ProxyController.setSessionServiceURL(java.lang.String); nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'session.service.url' in string value "${session.service.url}"
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessPropertyValues(AutowiredAnnotationBeanPostProcessor.java:292)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1185)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:537)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:475)
        at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:304)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:228)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:300)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:195)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:703)
        at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:760)
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:482)
        at org.springframework.web.context.ContextLoader.configureAndRefreshWebApplicationContext(ContextLoader.java:403)
        at org.springframework.web.context.ContextLoader.initWebApplicationContext(ContextLoader.java:306)
        at org.springframework.web.context.ContextLoaderListener.contextInitialized(ContextLoaderListener.java:106)
        at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4729)
        at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5167)
        at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
        at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1408)
        at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1398)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: org.springframework.beans.factory.BeanCreationException: Could not autowire method: public void org.mskcc.cbio.portal.web.ProxyController.setSessionServiceURL(java.lang.String); nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'session.service.url' in string value "${session.service.url}"
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredMethodElement.inject(AutowiredAnnotationBeanPostProcessor.java:596)
        at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:87)
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessPropertyValues(AutowiredAnnotationBeanPostProcessor.java:289)
        ... 22 more
Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'session.service.url' in string value "${session.service.url}"
        at org.springframework.util.PropertyPlaceholderHelper.parseStringValue(PropertyPlaceholderHelper.java:174)
        at org.springframework.util.PropertyPlaceholderHelper.replacePlaceholders(PropertyPlaceholderHelper.java:126)
        at org.springframework.beans.factory.config.PropertyPlaceholderConfigurer$PlaceholderResolvingStringValueResolver.resolveStringValue(PropertyPlaceholderConfigurer.java:259)
        at org.springframework.beans.factory.support.AbstractBeanFactory.resolveEmbeddedValue(AbstractBeanFactory.java:802)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:871)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:858)
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredMethodElement.inject(AutowiredAnnotationBeanPostProcessor.java:553)
        ... 24 more
```

mandawilson:
This sets the default value for sessions.service.url to be an empty string to fix #1685

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
+++ b/web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
@@ -54,7 +54,7 @@ public class ProxyController
   public void setBitlyURL(String property) { this.bitlyURL = property; }
 
   private String sessionServiceURL;
-  @Value("${session.service.url}")
+  @Value("${session.service.url:''}") // default is empty string
   public void setSessionServiceURL(String property) { this.sessionServiceURL = property; }
 
   private String pdbDatabaseURL;
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -159,8 +159,11 @@ genomespace=true
 # set this to true if you update cancer studies in your production database without shutting the web server
 recache_study_after_update=false
 
-# session-service URL 
-session.service.url=http://localhost:8080/session_service/api/sessions/msk_portal/
+# session-service url: http://[host]:[port]/[session_service_app]/api/sessions/[portal_instance]/ 
+# example session-service url: http://localhost:8080/session_service/api/sessions/public_portal/
+# see: https://github.com/cBioPortal/session-service
+# excluding this value or setting it to an empty string will revert to the previous bookmarking method
+session.service.url=
 
 # disabled tabs, | delimited
 # possible values: cancer_types_summary, mutual_exclusivity, plots, mutations, co_expression, enrichments, survival, network, download, bookmark, IGV


web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
```diff
@@ -54,7 +54,7 @@
   public void setBitlyURL(String property) { this.bitlyURL = property; }
 
   private String sessionServiceURL;
-  @Value("${session.service.url}")
+  @Value("${session.service.url:}")
```
JiaoJiao123: just wondering, what will the colon do?

inodb: After colon it sets a default value. In this case empty string. I'm just wondering if this is the way we should be setting default values for spring or if we should do it through `GlobalProperties.java`.

mandawilson: I think having a goal of getting properties from one place where default values can be set is good and we should do that.  Making that happen should probably be a separate issue, since this solution matches the rest of the code in ProxyController which is not using GlobalProperties but accessing properties directly from portal.properties.

ersinciftci: This might be easier to understand than a single colon: http://stackoverflow.com/a/33687265/2464448 . I think GlobalProperties class is an anti-pattern, so this looks good to me, but maybe we can change it to @Value("${session.service.url:#{null}}") like in the answer that I gave the link to.

mandawilson: @ersinciftci What do you think?

ersinciftci: @mandawilson `${session.service.url:''}` with a comment works too.

inodb: @ersinciftci In what way is using GlobalProperties an anti pattern? Might be good to send a PR and update the CONTRIBUTING doc, explaining this anti pattern & what a better solution is: https://github.com/cBioPortal/cbioportal/blob/master/CONTRIBUTING.md#devops.

@mandawilson @ersinciftci What's slightly odd to me now is that when using the default `EXAMPLE.properties` you get a different set up then when using no `EXAMPLE.properties`. I think the behavior should be the same. So maybe better to have the sessions service running by default?

mandawilson: @inodb Definitely the session service should not be running by default.  I would like to have an example session service URL in the portal.properties.EXAMPLE file, but we could do this:

inodb: @mandawilson  Yeah I think having the URL commented out is a good idea. It might be good to point to some documentation of how to use the session service, unless it will be an MSKCC internal thing for now. In that case it might be better to add `(mskcc internal)` in the comment

sheridancbio:
I plan to merge this change now. My general comments are that I agree that we should move towards a more centralized / integrated method of doing configuration. I also like the idea of having the configuration be the same if all user-set properties files / environment variables are removed as compared to having all of the default properties files and variables in-place. And I am in favor of dropping our homemade static GlobalProperties approach and moving to a flexible framework / non-baked in approach. But that will be a significant effort and should not delay this fix.

web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
ersinciftci: @inodb I think GlobalProperties is created for JSP pages mainly, if we are not going to use JSP anymore, we can easily use @Value("${property.name}") in whatever class we want to use that property, so no need to have a central class for managing properties.

ersinciftci: @mandawilson so how do you turn the session service on? if the url is defined in properties file, it's turned on by default? if that's not the case, we don't need to comment out the url in portal.properties.EXAMPLE file.

mandawilson: @ersinciftci The session service is a separate Spring Boot application that requires a Mongo database.  If you have it running and set the session.service.url in the portal application then the portal application uses the session service to store the information the portal needs for bookmarking.  Most people will not have the session service running, so it should be turned off by default.

ersinciftci: @mandawilson I see, ok makes sense then, thanks.

PR #1653, opened 2016-09-09
Change of PFAM url

oplantalech:
PFAM has been moved to a new location.

PR #1696, opened 2016-09-21
New seed DB

pieterlukasse:
# What? Why?

This PR documents the location of the new seed DB and how to load it. 

These files contain the updated seed database for cBioPortal. We included:
1. Updated genes (60072) and gene aliases according to latest NCBI file
2. Updated gene lengths in `gene` table according to latest GRCh37 data from Gencode, using the fix proposed in #1678
3. Updated `pfam_graphics` table with latest data, according to #1653

Compared to the previous seed database, the following number of values are removed from these tables because their associated Entrez Gene IDs do not exist anymore in the latest NCBI Gene info file:
- cosmic_mutation: 247
- uniprot_id_mapping: 1375
- interaction: 56
- drug_interaction: 6

These tables were dropped, since they were not found in cgds.sql anymore:
- micro_rna
- micro_rna_alteration
- mutation_frequency
- patient_list
- patient_list_list
# Notify reviewers

@cBioPortal/frontend
@cBioPortal/backend
@cBioPortal/devops

In particular: 
@aderidder @sheridancbio @ersinciftci @sandertan

### DIFF
--- a/docs/Downloads.md
+++ b/docs/Downloads.md
@@ -8,10 +8,7 @@ This page describes the various files available for download.
 
 # [Seed Database](seed-database)
 
-The files are MySQL database dump for seeding a new instance of the cBioPortal.  They contain all the necessary background data for a properly functioning cBioPortal website, including cancer types, gene, uniprot-mappings, drug, pdb and network data.  The files are available for download here:
-
-- [cbioportal-seed SQL (.gz) file - part1 (no pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_no-pdb_hg19.sql.gz)
-- [cbioportal-seed SQL (.gz) file - part2 (only pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_only-pdb.sql.gz)
+The files are MySQL database dump for seeding a new instance of the cBioPortal. For instructions for downloading it [see our datahub page](https://github.com/cbioportal/datahub/blob/master/seedDB/README.md). 
 
 # [MAF Example](maf-example)
 
--- a/docs/Downloads.md
+++ b/docs/Downloads.md
@@ -10,7 +10,8 @@ This page describes the various files available for download.
 
 This file is a MySQL database dump for seeding a new instance of the cBioPortal.  It contains all the necessary background data for a properly functioning cBioPortal website, including cancer types, gene, uniprot-mappings, drug and network data.  The file is available for download here:
 
-[cbioportal-seed.sql.gz](https://storage.googleapis.com/cbiostorage-1/cbioportal-seed.sql.gz)
+[cbioportal-seed SQL (.gz) file - part1 (no pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_no-pdb_hg19.sql.gz)
+[cbioportal-seed SQL (.gz) file - part2 (only pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_only-pdb.sql.gz)
 
 # [MAF Example](maf-example)
 
--- a/docs/Downloads.md
+++ b/docs/Downloads.md
@@ -8,10 +8,10 @@ This page describes the various files available for download.
 
 # [Seed Database](seed-database)
 
-This file is a MySQL database dump for seeding a new instance of the cBioPortal.  It contains all the necessary background data for a properly functioning cBioPortal website, including cancer types, gene, uniprot-mappings, drug and network data.  The file is available for download here:
+The files are MySQL database dump for seeding a new instance of the cBioPortal.  They contain all the necessary background data for a properly functioning cBioPortal website, including cancer types, gene, uniprot-mappings, drug, pdb and network data.  The files are available for download here:
 
-[cbioportal-seed SQL (.gz) file - part1 (no pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_no-pdb_hg19.sql.gz)
-[cbioportal-seed SQL (.gz) file - part2 (only pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_only-pdb.sql.gz)
+- [cbioportal-seed SQL (.gz) file - part1 (no pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_no-pdb_hg19.sql.gz)
+- [cbioportal-seed SQL (.gz) file - part2 (only pdb_ tables)](https://storage.googleapis.com/cbiostorage-1/seedDB/v1/cbioportal-seed_only-pdb.sql.gz)
 
 # [MAF Example](maf-example)
 
--- a/docs/Import-the-Seed-Database.md
+++ b/docs/Import-the-Seed-Database.md
@@ -26,6 +26,8 @@ and (this command takes a bit longer to import PDB data that will enable the vis
 
     > mysql --user=cbio_user --password=somepassword cbioportal  < seed-cbioportal_only-pdb.sql
 
+:information_source: please be aware of the version of the seed DB. If it is different from what the cBioPortal system is expecting, the system will at some point ask you to run a migration step. The system will automatically give you a clear message about this (with instructions) if a migration is needed. 
+
 ## Drug-target and Clinical Trial Data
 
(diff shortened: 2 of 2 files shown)


pieterlukasse:
@aderidder can you help testing this?

pieterlukasse:
thanks to @sandertan for helping making the summary above and the files available! 🎉

aderidder:
Hi @pieterlukasse 
Great! I've just downloaded the new files and I'm importing them into my cbioportal at the moment. I'll also import one or two studies, one of which previously failed due to the gene table issues. I'll keep you posted.

sandertan:
Hi,

I installed and tested the new seed-db with cBioPortal 1.2.2 and successfully loaded the BRCA, KIRC, PRAD, UCEC and UCS (all provisional) TCGA studies. For the installation I used Docker, where it's possible to give both `*.sql.gz` files with `-v`. 

Removed error from SKCM study:
- "Gene symbol maps to multiple Entrez ids...."

Warnings that do no appear anymore or that have been decreased in number: 
- "Entrez gene id not known to the cBioPortal instance. This gene will not be loaded"
- "Gene symbol and Entrez identifier do not match, the symbol will be ignored"

Also tested the AAC and BRCA study with cBioPortal 1.2.5. The only (new) error is caused by the new version of validateData.py, which gives an error for [Not Available] and 'Multiple comma separated values' in the SWISSPROT column of `data_mutations_extended.txt`, but when we change this to a warning (or remove the column), study can be loaded successfully.

@aderidder is it working for you as well?

aderidder:
Yes, seems ok. The error I encoutered before does not occur anymore

pieterlukasse:
@sheridancbio can you also review and test this one? 
from what I read above, all seems OK 👍

sheridancbio:
I hope to merge this in a few hours. Can you rebase to the head of hotfix and squash?

sheridancbio:
I am testing this new seed database now.

The first issue to raise is that I tried to load this into a database which I created called cgds_seedtest, but the database name is hardcoded into the creation script in cbioportal-seed_no-pdb_hg19.sql (hardcoded to be `cbioportal`). I think that for flexibility we want users of the seed databases to load them into whatever named database they want to create (they may want to create several separate databases). The instructions for loading the databases contain a database name specification but it is being ignored due to the CREATE and USE commands at the top of the .sql file. In the command (mysql --user=cbio_user --password=somepassword cbioportal  < cbioportal-seed.sql) the `cbioportal` database name is not used.

sheridancbio:
In order to bring cgds.sql into accordance with the seed database, I recommend these changes to both cgds.sql and the new seed database:

--> 6 database tables have a mysql comment specified (not in cgds.sql):
  attribute_metadata
  clinical_attribute
  entity
  sanger_cancer_census
  mutation
  mutation_event
> Fix: These comments can be added to cgds.sql

--> 3 tables have DEFAULT CHARSET set to utf8 (not in cgds.sql):
  clinical_trials
  info
  users
> Fix: These tables can have DEFAULT CHARSET set to utf8 in cgds.sql

--> 4 tables have ENTREZ_GENE_ID as type bigint(20) (cgds uses int(255))
  cna_event
  gistic_to_gene
  mut_sig
  sanger_cancer_census
> Fix: from the current gene2ensembl.gz download, GeneID has a max value of 108594174. type int can hold values up to 20 times as high. It is debatable whether we will run out of gene numbering space in the next 10 years. I suggest we continue to use integer fields rather than bigint because begint is mysql specific (for example, oracle calls the 16 bit wide integer a "LONGINTEGER" instead). int(255) is silly however. The maximum display with for a normal integer is 11, so I suggest we change cgds and the seed database to use int(11) field types.

--> field widths differ in table clinical_trial_keywords
    PROTOCOLID is char(50) (instead of 20 in cgds)
    KEYWORD is varchar(256) (instead of 160 in cgds)
> Fix: The longer field widths can be used in cgds.sql

--> field KEYWORD in table clinical_trial_keywords has a default value (not in cgds)
> Fix: add the default value to cgds.sql

--> table cancer_study does not have field IMPORT_DATE (cgds does)
> Fix: add the field in the seed database: `IMPORT_DATE` datetime DEFAULT NULL,

--> table mutation has index on field MUTATION_EVENT_ID (cgds does not)
> Fix: add index to cgds.sql

--> table sample_profile has index on field SAMPLE_ID (cgds does not)
> Fix: add index to cgds.sql

--> table sample_list_list has index on field SAMPLE_ID but calls the index `PATIENT ID`
> Fix: use the normal name of SAMPLE_ID for the index on SAMPLE_ID (to match cgds)

--> table genetic_profile has property AUTO_INCREMENT=2 (cgds does not)
> Fix: unless there is a reason for this, rebuild seed database without the faster increment

--> table sample_list_list has default '0' for SAMPLE_ID (cgds does not)
> Fix: drop this default value from the seed database (we should not default to 0)

--> table uniprot_id_mapping has default value of 0 for field ENTREZ_GENE_ID (cgds does not)
> Fix: drop this default value from the seed database (we should not default to 0)

sheridancbio:
correction to above comments : it was decided that we can drop the table clinical_trial_keywords (no longer used). I've edited my previous comments to follow this.

pieterlukasse:
@sheridancbio thanks for the detailed list! I think this will make it possible to reach our next goal: remove schema from the seed files and use cgds.sql for that part 👍  

The questions I have for this specific release of the seed DB are:
- does it introduce any new issues? E.g. the hardcoded database name `cbioportal` : was this introduced now or was it also present in the previous seed? 
- are there any problems with the content of this new seed DB? 
- are there any incompatibilities that would break the code (e.g. missing fields) in hotfix?

If the answer to all the above is "No" then I would like to propose we make these extra improvements in the next release of the seed DB. What do you think?

sheridancbio:
@pieterlukasse I agree that improvements can be made in a subsequent PR (hopefully quickly).

Answering your questions: I looked at what I think is the current seed database (correct me if I am wrong). The link on our github instructions page is broken here: https://cbioportal.readthedocs.io/en/latest/Import-the-Seed-Database.html
but I assume that this is the correct link to the downloads page is this:
https://cbioportal.readthedocs.io/en/latest/Downloads.html
which let me download cbioportal-seed.sql.gz
This database did not have the hardcoded "CREATE DATABASE `cbioportal`" and "USE DATABASE `cbioportal`" links up top, so the new seed database does introduce a new difficulty for users who want to create databases with a different name. I checked mysql documentation and there is no easy way "rename" a database once it is created, so I think an installer would need to hand edit both files (they both have the CREATE / USE commands) in order to easily import into differently named databases.

For checking the correctness of the seed database content, it sounds like it has worked for others. But I'll do a little more checking myself and reply again soon. I don't think there are any fields that are both missing and needed (IMPORT_DATE is not used anywhere in the portal)

The fixes for harmonizing the seed database to cgds can be done in a separate PR. Maybe I'll work on that.

sheridancbio:
The COMMENT to tables attribute_metadata, clinical_attribute, and entity would be more proper as a comment to the field DATATYPE or ENTITY_TYPE rather than to to entire table. I am moving them there in cgds.sql

sheridancbio:
There was some discussion this morning at our stand up meeting about the seed database. We were thinking that it makes sense to make some small fixes (from above comments), to make the improvement of separating the create table statements from the insert statements, to update the seed database to use the new db_schema_version which is about to be released to public portal, and to deploy it through the github data-portal lsf based storage so that it will be on the same platform as the public portal databases. Do you think that would make sense? We can do a new PR here based on your approach and your PR changes to the docs if it sounds good. CC: @n1zea144

pieterlukasse:
@sheridancbio ok, so removing the hardcoded db name "cbioportal" is the only fix we need for this PR, right? 

Regarding `db_schema_version`: can you explain what this is? Is this a variable name, table name, or a typo?

Regarding git lfs: 👍 yes please! This would be great! I proposed this before in slack as well. Shall I move the files there and update the links in this PR?

sheridancbio:
db_schema_version is the version number of the database schema. We are moving to schema 1.3.0 with the next public release, so I was just referring to the changes needed to make the seed database follow the 1.3.0 schema (such as adding the gene_panel table, structural_variants, etc..

What I was proposing is that we alter the seed database you have deployed to make it follow 1.3.0 schema, remove the create table statements (so that cgds.sql can be used instead) and relocate the files to github data-portal. So the seed database would be altered (also dropping the hardcoded database name), and the instructions would refer to github. Maybe we should package or bundle somehow the three pieces users will need (schema file, insert_pdb, insert_non_pdb). That's my suggestion.

pieterlukasse:
@sheridancbio that is a great proposal. Unfortunately @sandertan is on holidays for the next 2 weeks. Will you have some time to pick this up? If yes, feel free to do so. 
I made a dump of the new seed DB and started uploading this to a branch to make a new PR to /datahub. Then maybe you can take it from there? I enabled commits from maintainers, so you should be able to commit to this PR.

adamabeshouse:
@pieterlukasse 
@zheins changed the tables `patient_list` and `patient_list_list` to `sample_list` and `sample_list_list` - the old name was a vestigial misnomer. Those tables are still in use.

pieterlukasse:
@adamabeshouse what do you mean? Where are they used?

pieterlukasse:
@sheridancbio I would love to get this one into the seed DB as well: #1422

sheridancbio:
#1806 and #1807 are now connected to this PR (via #1422 which is closed)

sheridancbio:
This looks good. 
Once we have https://github.com/cbioportal/datahub/blob/master/seedDB/README.md in place we can merge this right away.

PR #1700, opened 2016-09-22
Small changes to gene panel implementation

zheins:
# What? Why?
Use existing sample_profile table for gene panel mapping instead of making the new table. Add controller tests for gene panel.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@fedde-s 
@sheridancbio


core/src/main/resources/db/cgds.sql
```diff
@@ -279,15 +278,31 @@ CREATE TABLE `genetic_profile_samples` (
...
 --
 CREATE TABLE `sample_profile` (
   `SAMPLE_ID` int(11) NOT NULL,
   `GENETIC_PROFILE_ID` int(11) NOT NULL,
+  `PANEL_ID` int(11) DEFAULT NULL,
   UNIQUE KEY `UQ_SAMPLE_ID_GENETIC_PROFILE_ID` (`SAMPLE_ID`,`GENETIC_PROFILE_ID`), -- Constraint to allow each sample only once in each profile
   FOREIGN KEY (`GENETIC_PROFILE_ID`) REFERENCES `genetic_profile` (`GENETIC_PROFILE_ID`) ON DELETE CASCADE,
-  FOREIGN KEY (`SAMPLE_ID`) REFERENCES `sample` (`INTERNAL_ID`) ON DELETE CASCADE
+  FOREIGN KEY (`SAMPLE_ID`) REFERENCES `sample` (`INTERNAL_ID`) ON DELETE CASCADE,
+  FOREIGN KEY (`PANEL_ID`) REFERENCES `gene_panel` (`INTERNAL_ID`) ON DELETE CASCADE
```
sheridancbio: Here we are allowing this field to be null in this table ... how does that affect / interact with the parent table?
I think it will not be a problem .. you would not require a matching null record in the parent table because usually you are mapping to a primary key field. And since one option is to cascade a delete in parent to cause a null in child, I think all is well.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoSampleProfile.java
```diff
@@ -57,10 +57,16 @@ public static int addSampleProfile(int sampleId, int geneticProfileId) throws Da
...
                 pstmt = con.prepareStatement
-                        ("INSERT INTO sample_profile (`SAMPLE_ID`, `GENETIC_PROFILE_ID`) "
-                                + "VALUES (?,?)");
+                        ("INSERT INTO sample_profile (`SAMPLE_ID`, `GENETIC_PROFILE_ID`, `PANEL_ID`) "
+                                + "VALUES (?,?,?)");
                 pstmt.setInt(1, sampleId);
                 pstmt.setInt(2, geneticProfileId);
+                if (panelId != null) {
+                    pstmt.setInt(3, panelId);
+                }
```
sheridancbio: my style preference: "} else {" on one line (several repeats of this below)

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -169,11 +169,13 @@ public void importData() throws IOException, DaoException {
 		        	continue;
 		        }
 		        
-				if( !DaoSampleProfile.sampleExistsInGeneticProfile(sample.getInternalId(), geneticProfileId)) {
-					DaoSampleProfile.addSampleProfile(sample.getInternalId(), geneticProfileId);
+				if( !DaoSampleProfile.sampleExistsInGeneticProfile(sample.getInternalId(), geneticProfileId)) {
```
sheridancbio: trailing whitespace, literal tab characters, messed up indenting on following lines

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml
```diff
@@ -296,7 +298,6 @@
     </insert>
 
     <insert id="insertGenePanelSampleProfileMap" parameterType="map" useGeneratedKeys="false">
-        insert into gene_panel_profile_map (SAMPLE_ID, PROFILE_ID, PANEL_ID)
-            values(#{sampleId}, #{profileId}, #{panelId})
+        update sample_profile set PANEL_ID = #{panelId} where SAMPLE_ID = #{sampleId} and GENETIC_PROFILE_ID = #{profileId}
```
sheridancbio: Hopefully in the import process we first create all the sample_profile entries, before we attempt to load the gene panel cases and update those existing records.

web/src/test/java/org/cbioportal/web/api/GenePanelControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+@WebAppConfiguration
+@ContextConfiguration(classes = {GenePanelControllerConfig.class, CustomObjectMapper.class})
+public class GenePanelControllerTest {
+    @Autowired
+    private WebApplicationContext webApplicationContext;
+    @Autowired
+    private GenePanelService genePanelServiceMock;
+    private MockMvc mockMvc;
+    private GenePanel genePanel1;
+        private GenePanel genePanel2;
```
sheridancbio: indent is off

web/src/test/java/org/cbioportal/web/api/GenePanelControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+        .accept(MediaType.parseMediaType("application/json;charset=UTF-8")))
+        .andExpect(MockMvcResultMatchers.status().isOk())
+        .andExpect(MockMvcResultMatchers.content().contentType("application/json;charset=UTF-8"))
+        .andExpect(MockMvcResultMatchers.jsonPath("$", Matchers.hasSize(2)))
+        .andExpect(MockMvcResultMatchers.jsonPath("$[0].stable_id").value("GENEPANEL2"))
+        .andExpect(MockMvcResultMatchers.jsonPath("$[0].description").value("2 genes tested"))
+        .andExpect(MockMvcResultMatchers.jsonPath("$[1].stable_id").value("GENEPANEL3"))
+        .andExpect(MockMvcResultMatchers.jsonPath("$[1].description").value("3 genes tested"));         
+    }    
+}
```
sheridancbio: already discussed this .. need a test for getPanelBySampleIdProfileId

core/src/main/resources/db/cgds.sql
fedde-s: @sheridancbio Not all genetic profiles are annotated with gene panel information. In fact, in existing data files, no genetic profile is. In those cases, a sample-profile relation (which is what a record in the `sample_profile` table represents) does not have an associated gene panel.

As an alternative to the nullable foreign key, this could be represented as a link table between the `sample_profile` table and the `gene_panel` table: a table with a foreign key to the primary key of each of the two tables. That would require adding a primary key constraint to the existing table `sample_profile`, and a unique constraint to the corresponding foreign key in the new link table to disallow many-to-many relations.

fedde-s: I do wonder if `ON DELETE SET NULL` might be preferable to `ON DELETE CASCADE`. The relation between sample and profile should be deleted if either the sample or the study is, but should it also be deleted if the gene panel is no longer supported? Might not matter too much, but I am not experienced enough to assess that.

sheridancbio:
It is a good tag to make not of PRs which alter the db schema. I think a better name might be "db schema change", since there may be confusion if the tag should be applied to database to code which manipulates actual data in the database.  Also the "includes" part seems inherent.

fedde-s:
@sheridancbio I agree, but it is applicable to this PR, even though it has no migration code from the previous version to the current one yet.

core/src/main/resources/db/cgds.sql
sheridancbio: @fedde-s we talked about your point of ON DELETE ... clauses.  It started a good conversation about what cases may come up where we want to alter or remove a previously existing gene panel. Changes like this would happen during the import / db update process.

After talking about it a bit we think that we prefer to prohibit all cascades --- so that if any profile/sample is linked to a gene panel identifier, then the gene panel cannot be deleted from the gene_panel table. First the importer/updater will need to resolve "properly" (whatever that means in the case) the old references to the gene panel identifier.

So we think we will remove the entire cascade clause at this time, and talk more about what should happen if a gene panel needs to change and how to support it in the import code.

zheins: @fedde-s I also changed the gene panel import script to check if a gene panel is in use by the sample_profile table. If it is, it stops the import of the gene panel as to not corrupt existing study data.

Thanks for pointing that out. As @sheridancbio stated, we need to talk more about this.

Issue #1646, opened 2016-09-07
frontend selection of mutation genetic profiles may need improvements


sheridancbio:
Originally, this issue was a bug report about unpopulated mutation lists in the results page.

The bugs have been fixed, but there is still possible ambiguity about which genetic profile should be used for frontend visualization (such as the mutationmapper) when there are more than one mutation profile to choose from.

----------------------------------------------
Below Here Is The Original Bug Report
----------------------------------------------
Accessing views through /beta deployment (rc?) show unpopulated mutation lists:

(suspect malfunctioning mutation data servlets - oncoprint looks ok)
@ersinciftci @n1zea144 
===============================

PatientVew (accessed through sample list in study view):
http://www.cbioportal.org/beta/case.do?cancer_study_id=brca_tcga&case_id=TCGA-3C-AAAU

Local debugging shows some exceptions:
SEVERE: Servlet.service() for servlet [MutationsJSON] in context with path [/cbioportaltest] threw exception
java.lang.NumberFormatException: For input string: ""
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:592)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.mskcc.cbio.portal.servlet.MutationsJSON.getDrugs(MutationsJSON.java:425)
	at org.mskcc.cbio.portal.servlet.MutationsJSON.processGetMutationsRequest(MutationsJSON.java:276)
	at org.mskcc.cbio.portal.servlet.MutationsJSON.processRequest(MutationsJSON.java:126)
	at org.mskcc.cbio.portal.servlet.MutationsJSON.doPost(MutationsJSON.java:786)

===========================

Through Query Page: (query study brca_tcga  on genes TP53 BRCA1)
There are many mutations in this query, but the "Mutations" Tab is empty for both genes.

[...]

PR #1705, opened 2016-09-23
Fixes to avoid exceptions for malformed OMA links in mutation_events

sheridancbio:
# What? Why?
Problem described here: #1646 

- In addition to "NA", check mutation_event link fields for the value [Not Available]
- also added additional error logging, a notification to user on failed redirect, and unit tests

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
@cBioPortal/backend

- MutationDataUtils now calls OmaLinkUtil function to see if link is valid
- three values are checked as empty field marker: "", "NA", "[Not Available]"
- invalid links (null, empty, empty field marker) automatically send "NA"
- redirect servlet now catches errors and logs .. also presents a fail message to user

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/servlet/OmaRedirectServlet.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/servlet/OmaRedirectServlet.java
@@ -32,14 +32,17 @@
 
 package org.mskcc.cbio.portal.servlet;
 
-import org.mskcc.cbio.portal.util.OmaLinkUtil;
-
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.MalformedURLException;
+import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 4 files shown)


pieterlukasse:
@sheridancbio should the validator also be updated?

sheridancbio:
@pieterlukasse I don't think so. This fix allows the backend to handle (without exception) the three possible placeholders for when there is no link to mutation-assessor. So it should now be valid to use "NA", "[Not Available]", or "" (empty string) for the three link fields in mutation event. When reporting these through the Api, the Web Server now reports "NA" for any field which contains one of these values. I am not sure what test the validator currently runs, but I suppose the valid values for the three link fields should be either (1) a string which is validly parsed by the Java URL class, or (2) one of these three placeholder strings. Perhaps a Null value in the field is also properly handled by the Webserver too (response to query would be "NA" I think).

sheridancbio:
But I think these fields will soon be removed from the data model anyway ... since we are going to move towards dynamic queries to the running mutationassessor web server (rather than getting them through static database tables). So any work on the validator for these fields would soon be irrelevant.

PR #1710, opened 2016-09-26
Created the web layer for the new REST API design

ersinciftci:
Sorry for the huge PR.

Created the Spring MVC web layer for the new REST API design.

Added controllers for all core resource types.
Added new models to support new controllers.
Added summary parent models to show the summary view in response types in Swagger UI, instead of showing the detailed response, because the default projection is SUMMARY.
Converted back some changes we made in Sprint 2 of Rearchitecture effort, so /geneticprofiledata API is using the old stack now, because it's a lot easier to design the new REST API/models without thinking about backward compatibility.
Added new exception types under service module as an example, will add more when needed.
Added exception handling for controllers.
Moved the old /api resources to /api-legacy, only the APIs proposed in this PR is going to be under /api.


sheridancbio:
I just read through the changes quickly. Initial impression : looks good.

sheridancbio:
I've completed code review.
Summary -- I think we can merge now if tests go smoothly and a couple of questions are answered.

Comments:

- PR returns ApiService to output DBMutationData
- PR drops addSampleListIdToMutationList() and keeps the original code in ApiService
- ProfileDataMapper now allows genes argument to be empty and returns all... do we want this? Could returning all profile data be too costly, or are we going to rely on paging to prevent too-expensive call servicing?
- do we now have redundant model classes?: (we can resolve in future if that is best)
	CNSegmentData
	CopyNumberSegment
- does MutationMapper.xml output the sample internal ID, cancer study internal ID, and genetic_profile internal ID to the end user? No -- I see @JSonIgnore in CancerStudySummaryMixin SampleSummaryMixin and GeneticProfileSummaryMixin
- recommend that we soon add @ApiParam human understandable descriptions to new API controllers (describing examples, what happens when optional argument is empty, etc..
- great .. ApiController is moved to weblegacy package
- but: this change is needed --- we must have only a package called org.cbioportal.weblegacy ... and must not have a package named org.cbioportal.webLegacy. The two git moves to webLegacy (ApiController and CosmicCountController) must be "re-moved" to weblegacy. This gets confusing on mac platforms which do not have case sensitive filesystems)
- ApiController reverts to original call logic
- If we are switching output of CNSegmentController to camelcase property names, should input parameters be camelcase? we need a standard soon to know what is appropriate here and in other legacy controllers.

I still need to run system tests on the api-legacy endpoints before merge.

sheridancbio:
System testing went pretty well. Some bugs exist in rc branch already, and some Api behavior too. But in terms of malfunctions which are introduced in this PR which are not already in rc, the main new problem seems to be that the igv.js CopyNumberSegment display is no longer working. We are investigating this, but suspect the Api call which detects chromosome from gene.

sheridancbio:
Correction ... old tech stack for /api/genetprofiles is not Dao, but the persistence classes in business module.

pieterlukasse:
:warning:  @ersinciftci this PR breaks the validation script for when the  `-u` parameter is used. To fix it, we could update validateData.py to use /api-legacy, but ideally would use the new api. 

We were using: 

cancertypes
clinicalattributes/patients
clinicalattributes/samples
genes
genesaliases

What do you think is the best solution now? Are all these implemented (or are you planning to)?

ersinciftci:
@pieterlukasse I'm planning to implement all the APIs one by one, but it will take some time, so I think for now you can change url to `/api-legacy` to not block anything, and then we can change it to use the new API once it's ready.

ersinciftci:
Ok I just noticed I already changed validateData.py to `api-legacy` but @angelicaochoa changed it back to `api` in `rc`, that's why it's broken. I'm talking to her, we will probably revert it to `api-legacy`.

Issue #1681, opened 2016-09-19
support priority studies in the query interface


jjgao:
Currently we have hard code mskimpact for priority category. We need to make it into properties file.

e.g.

```
# settings for priority studies
# e.g. priority_studies=cancer_study_1#category1;cancer_study_2#category1;cancer_study_3#category2
priority_studies=
```

jjgao:
@adamabeshouse could you prioritize this? It'd be great if we can put it out in this release (v1.3.0). Thanks!

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1718

PR #1718, opened 2016-09-26
Prioritize studies in study selector

adamabeshouse:
Implements https://github.com/cBioPortal/cbioportal/issues/1681

Format of the specifying portal properties variable is `priority_studies=category1#study1a,study1b,study1c;category2#study1`
Implementation-wise, we could just as easily replace the hashtags with commas, but for user-friendliness I left it a different character delineating the category from the studies.

# Checks
  the `-s` option)
  hotfix.

# Notify reviewers
@jjgao @ritikakundra 
@cBioPortal/backend @cBioPortal/frontend

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -32,7 +32,6 @@
 
 package org.mskcc.cbio.portal.util;
 
-import com.mysql.jdbc.StringUtils;
 import org.mskcc.cbio.portal.servlet.QueryBuilder;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -770,6 +772,14 @@ public static String getDarwinAuthority() {
         return darwinAuthority;
     }
     
+    public static String getPriorityStudies() {
```
jjgao: Maybe either
  `Map<String, String> //study to category`
or
   `Map<String, Set<String>> //category to studies`
can be returned instead of `String`?

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -168,3 +168,7 @@ session.service.url=
 # disabled tabs, | delimited
 # possible values: cancer_types_summary, mutual_exclusivity, plots, mutations, co_expression, enrichments, survival, network, download, bookmark, IGV
 disabled_tabs=
+
+# study ids and categories to force to top of study selector
+# format is category1#study1a,study1b,study1c;category2#study2
+priority_studies=MSKCC DMP#mskimpact
```
jjgao: Let's just put an empty property here, ie.

`priority_studies=`

PR #1720, opened 2016-09-27
Npm install cbioportal frontend

inodb:
Add maven-frontend-plugin to npm install separate cbioportal-frontend repo from
github. Replace clinical information tab as proof of concept.

To facilitate development, one can use http://localhost:3000 for react frontend
instead if localdev is set to true in localStorage.
- Install maven-frontend-plugin to get npm install
- Add package.json that refers to a cbioportal-frontend repo
- Copy the dist folder of the front end repo to the war
- add script tag to patient_view jsp


inodb:
Going for #1844 approach instead

Issue #1712, opened 2016-09-26
Beta-Portal: IGV should load data when a user switches to a different chromosome


schultzn:
IGV should load data when a user switches to a different chromosome

sheridancbio:
@JiaoJiao123 If this is working, I think this issue should be closed

PR #1721, opened 2016-09-27
update igv js library and calling functions

JiaoJiao123:
Related to issue #1712 
1) add the function to allow switch between different chromosomes 
2) adjust gene function call because of the segment api parameters change due to PR #1719

### DIFF
--- a/portal/src/main/webapp/js/lib/igv.js
+++ b/portal/src/main/webapp/js/lib/igv.js
@@ -22,6 +22,10 @@
  * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/js/lib/igv.js
```diff
@@ -5122,10 +5122,41 @@ var igv = (function (igv) {
...
         }
+        var div = $("#igvDiv")[0],
+            options = {
+                showNavigation: true,
+                showRuler: true,
+                genome: "hg19",
+                locus: feature,
+                tracks: [
+                    {
+                        url: "http://localhost:8080/cbioportal/api/copynumbersegments",
```
jjgao: Should localhost... be here?

Ideally, igv.js should not worry about the sementics of the URL. Maybe a call back function should be pass in to igv instead of url?

JiaoJiao123: Good catch! I just updated the url link by removing localhost.
@sheridancbio  just brought up a very good point that we might need to update their liscence since we modified their library. I will update it once we find out how to do it.

PR #1729, opened 2016-09-29
Validator gives error for certain values in Swiss-Prot column

sandertan:
# Validator gives error for certain values in Swiss-Prot column

The latest version of cBioPortal (1.2.5) gives errors when importing a TCGA (provisional) study from cBioPortal Datahub, because of invalid values in the Swiss-Prot column of data_mutations_extended.txt.

Fix: Changing certain scenerios from errors to warnings. This causes the loader to be unable to find an accession number, and the HUGO symbol will be used to gather an accession in order to get Pfam domains.

- When value contains `[Not Available]` or `NA`, give a warning instead of an error.
- When value contains multiple values seperated by column, give a specific warning instead of an error.
- In addition, a scenario appeared where the multiple values were incapsulated by double quotes. This now also gives a warning instead of an error.
- The first part of a UniProt name is 1 to 5 characters. In previous code, 2-5 characters was allowed, which gave an error at UnitProt name `P_HUMAN`. 
- I changed the unittest to expect a warning instead of an error at a test for the validator.

Actually, even when the Swiss-Prot name is completely wrong, it is not a big problem because the HUGO symbol can be used to get a UniProt accession. This might not be accurate, so a warning should be given, but loading should continue because cBioPortal uses the HUGO symbol. So maybe in future versions, we might want to remove all scenarios in which an error is given for this column completely, and make it a warning.
# Notify

@pieterlukasse @zheins


core/src/main/scripts/importer/validateData.py
```diff
@@ -1248,8 +1248,19 @@ def checkSwissProt(self, value):
         else:
             # format described on http://www.uniprot.org/help/entry_name
             if not re.match(
-                        r'^[A-Z0-9]{2,5}_[A-Z0-9]{2,5}$',
+                        r'^[A-Z0-9]{1,5}_[A-Z0-9]{1,5}$',
                         value):
+                value_list = value.strip('"').split(',')
```
pieterlukasse: we can simplify this whole part by just letting the whole scenario return a warning. So if the regular expression does not match, then there should be a warning about invalid name.

pieterlukasse:
@sandertan code looks good 👍  

I will test it.

pieterlukasse:
tested and looks good :+1: 

I added a small improvement for the scenario where the SWISSPROT column contains multiple values. A better warning message is given in this case.

PR #1733, opened 2016-09-29
Fix bug in datamanager.js getMutationProfileIds

sheridancbio:
datamanager.js has a function 'getMutationProfileIds' which was returning all retrieved profile ids rather than just the ids of mutation profiles. This was causing mutationmapper.js to not get mutation events if the first profile id returned was a copy number alternation profile (for example)

- return list is now filtered to only return mutation profiles

note: we still need to add logic in the handler which starts mutationmapper to choose the "correct" mutation profile if there are more than one.
# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@onursumer @adamabeshouse @jjgao 

- instead of returning all profile ids, add filter to only return mutation profiles

### DIFF
--- a/portal/src/main/webapp/js/api/cbioportal-datamanager.js
+++ b/portal/src/main/webapp/js/api/cbioportal-datamanager.js
@@ -1453,9 +1453,11 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	'getMutationProfileIds': makeCachedPromiseFunction(
 		function(self, fetch_promise) {
 		    self.getGeneticProfiles().then(function (profiles) {
-			fetch_promise.resolve(profiles.map(function (p) {
-			    return p.id;
-			}));
+			fetch_promise.resolve(
+                            profiles
+                            .filter(function (p) {return p.genetic_alteration_type === "MUTATION_EXTENDED";})
+                            .map(function (p) {return p.id;})
+                        );
 		    }).fail(function () {
 			fetch_promise.reject();
 		    });


sheridancbio:
An example query which shows the problem / failure: (this query returns 2 mutations, but click the mutations tab and it shows "there are no mutations")

http://www.cbioportal.org/beta/index.do?cancer_study_list=brca_tcga&cancer_study_id=brca_tcga&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=brca_tcga_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=brca_tcga_gistic&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=brca_tcga_cnaseq&case_ids=&patient_case_select=sample&gene_set_choice=user-defined-list&gene_list=AKT1&clinical_param_selection=null&tab_index=tab_visualize&Action=Submit&show_samples=false

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1453,9 +1453,11 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	'getMutationProfileIds': makeCachedPromiseFunction(
 		function(self, fetch_promise) {
 		    self.getGeneticProfiles().then(function (profiles) {
-			fetch_promise.resolve(profiles.map(function (p) {
-			    return p.id;
-			}));
+			fetch_promise.resolve(
+                            profiles
+                            .filter(function (p) {return p.genetic_alteration_type == "MUTATION_EXTENDED";})
```
adamabeshouse: In general it's better to use triple equalities in javascript, double equality does more fuzzy equalities. In this case I don't think it makes a difference but just as a rule I like to use triple equality.

sheridancbio: fixing now

sheridancbio:
Corrected equality operator

PR #1734, opened 2016-09-29
Gene panel api expansion for oncoprint

zheins:
# What? Why?
Adds functionality to the gene panel api for the oncoprint.

- Gene panel api can be queried with a profile and list of genes
- Returns sample that have gene panel data and the genes from the genes queried which are part of that panel.
- If no data provided for sample that is in a profile queried, it should be assumed that whole exome was done for that genetic profile.

Behavior:

- Endpoint is `genepanel/data`
- Takes a genomic profile and a gene list
- Returns a list of gene panels, each of which contains the genes from the query which are in the gene panel and the samples to which the gene panel applies
- If no gene panels exist for the profile queried, empty list returned.
- If some samples in the profile have no gene panel data associated with them (the sample is not in one of the sample lists inside the gene panel response objects) they are assumed to be whole exome.
- Example query: api-legacy/genepanel/data?profile_id=<PROFILE_STABLE_ID>&genes=TP53,AKT1,...

-Example output:
`[
  {
    "internalId": null,
    "stableId": "IMPACT341",
    "description": "Targeted (341 cancer genes) sequencing of various tumor types via MSK-IMPACT on Illumina HiSeq sequencers.",
    "genes": [
      {
        "entrezGeneId": null,
        "hugoGeneSymbol": "TP53",
        "type": null,
        "cytoband": null,
        "length": null,
        "aliases": null,
        "chromosome": null
      },
      {
        "entrezGeneId": null,
        "hugoGeneSymbol": "AKT1",
        "type": null,
        "cytoband": null,
        "length": null,
        "aliases": null,
        "chromosome": null
      }
    ],
    "samples": [
      "SampleId1",
      "SampleId2",
      "SampleId3"
    ]
  }
]`

@adamabeshouse does this sound ok to you?

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@adamabeshouse @sheridancbio

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/GenePanelRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/GenePanelRepository.java
@@ -42,7 +42,7 @@ import java.util.*;
 
 public interface GenePanelRepository {
     
-        String getGenePanelBySampleIdAndProfileId(String sampleId, String profileId);
+        List<GenePanelWithSamples> getGenePanelsByProfile(String profileId);
         // TODO: All of the below methods are for importing purposes only. They should be
         // removed once a proper import solution is put in place.
         List<GenePanel> getGenePanelByStableId(String stableId);
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml
@@ -21,6 +21,16 @@
         <result property="description" column="description"/>
     </resultMap>
     
+    <resultMap id="genePanelSampleResultMap" type="org.cbioportal.model.GenePanelWithSamples">
+        <result property="stableId" column="stableId" />
+        <result property="description" column="description"/>
+        <collection property = "genes" ofType="org.cbioportal.model.Gene">
+            <result property="hugoGeneSymbol" column="hugoGeneSymbol"/>
+        </collection>
+        <collection property="samples" ofType="String">
+            <result property="stableId" column="sampleId"/>
+        </collection>
+    </resultMap>
 
    <resultMap id="sampleResultMap" type="org.cbioportal.model.Sample">
         <result property="internalId" column="INTERNAL_ID"/>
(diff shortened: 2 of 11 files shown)


adamabeshouse:
👍

web/src/test/java/org/cbioportal/weblegacy/GenePanelControllerTest.java
```diff
@@ -121,16 +121,30 @@ public void genePanelByStableIdTest() throws Exception {
     }
 
     @Test
```
sheridancbio: (optional)  For list formatting testing, I would recommend adding another sample and another gene to the mocked service return (2 samples, 2 genes). Or maybe add a second test which returns results with more than one member in the list.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GenePanelRepository.java
```diff
@@ -42,7 +42,7 @@
 
 public interface GenePanelRepository {
     
-        String getGenePanelBySampleIdAndProfileId(String sampleId, String profileId);
```
zheins: This is not needed. We are able to get the samples in a profile, and for now the only use case is given a profile and set of genes, what are the sequenced genes.

sheridancbio:
@zheins and I discussed the response type and we thought that there was room for improving performance, since there is likely to be a great deal of repeated gene list data in each query response. On a query with 1000 samples, and 100 genes, if the genes are all present in the gene panel which was used to sequence the samples, and all (or almost all) samples used the same gene panel, then the same list of 100 genes will be returned 1000 times in the response.

An alternative which has one level of indirection / data compression would be to return two lists:

list 1 = a list of integer index numbers into list 2 (one index number per sample)
list 2 = a unique list of gene lists. If there are multiple gene panels used, there might be multiple gene lists showing how each panel intersected with the set of genes queried to produce the oncoprint, but there would be no more members in list 2 than the number of different gene panels used in the profile.
A fuller / flatter / more streamlined backend/frontend interface could be a two step process --- first query the profile and get a mapping of samples to the gene panel used (per sample). For each (unique) used gene panel, a second set of queries would determine which of the genes used to construct the oncoprint intersect with the genes in the gene-panel. Finally the frontend could construct a full data structure from the two steps above to display the oncoprint markings.

The problem with the implemented data exchange is that it is over-tailored to suit the needs of the oncoprint visualization (rather than providing a reusable/flexible data service)

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml
```diff
@@ -103,17 +113,21 @@
...
         select
-            gene_panel.STABLE_ID as stableId
-        from sample_profile
+            gene_panel.STABLE_ID as stableId,
+            gene_panel.DESCRIPTION as description,
+            gene.HUGO_GENE_SYMBOL as hugoGeneSymbol,
+            sample.STABLE_ID as sampleId
+        from genetic_profile
+            inner join sample_profile on genetic_profile.GENETIC_PROFILE_ID = sample_profile.GENETIC_PROFILE_ID
             inner join gene_panel on sample_profile.PANEL_ID = gene_panel.INTERNAL_ID
```
sheridancbio: If we wanted to get out records which show samples that have no assigned gene panel (whole exon sequenced) we could make this inner join into a left join.  (LEFT JOIN gene_panel on ...)

Issue #1695, opened 2016-09-21
Absent Select Patient/Case Set of  "Tumor samples with Mass spec data"


sandertan:
Hi,

On the Query page, when Protein Levels (Mass spectrometry) is selected under Select Genomic Profiles, there is no "Tumor samples with Mass Spec data" under Select Patient/Case Set.

For example, this applies when using the dataset: Breast Invasive Carcinoma, TCGA, provisional.

Kind regards,
Sander

sandertan:
Seems to be fixed 👍

Issue #1694, opened 2016-09-21
Absent "Mass spec"-column under Datasets


sandertan:
Hi,

Now that it's possible to load mass spectrometry data, it would be nice if there would be a column under "Datasets" to easily see which studies contain samples with this data.

Kind regards,
Sander

fedde-s:
See [the comment I just made in the pull request that addresses this](https://github.com/cBioPortal/cbioportal/pull/1738#issuecomment-260949758).

sandertan:
Design choice to not have it in. In future we might want to replace "RPPA" column with "Protein level" column.

PR #1738, opened 2016-09-30
Mass spectrometry: allow import, add to datasets tab & add to query selection

sandertan:
# Added functionality for mass spectrometry data

With the addition of mass spectrometry data from CPTAC for the TCGA studies, it will be useful if allow this data to be imported. Also some functionality was missing to easily select samples with this data.

- Add ms_abundance to allowed_data_types.txt
- Add mass spec variable to CancerStudyStats.java
- Add mass spec case to SampleListCategory.java
- Add mass spec variable to DataSetsUtil.java and data_sets.jsp to add column to Datasets tab #1694 and option to select mass spec samples on Query page #1695
# Notify reviewers

@pieterlukasse @zheins


core/src/main/scripts/importer/allowed_data_types.txt
```diff
@@ -32,3 +32,4 @@ METHYLATION	CONTINUOUS	methylation_hm450
 FUSION	FUSION	fusion
 PROTEIN_LEVEL	LOG2-VALUE	rppa
 PROTEIN_LEVEL	Z-SCORE	rppa_Zscores
+PROTEIN_LEVEL	Z-SCORE	ms_abundance
```
pieterlukasse: Next to Z-SCORE, we will need another one here for normal protein levels, similar to LOG2-VALUE rppa on row 33 above.

portal/src/main/webapp/data_sets.jsp
```diff
@@ -85,6 +85,7 @@
 				 out.println("<th scope='col'>Tumor miRNA</th>");
 				 out.println("<th scope='col'>Methylation (HM27)</th>");
 				 out.println("<th scope='col'>RPPA</th>");
+         out.println("<th scope='col'>Mass Spectrometry</th>");
```
pieterlukasse: please fix identation

core/src/main/java/org/mskcc/cbio/portal/util/DataSetsUtil.java
```diff
@@ -132,14 +132,15 @@ public DataSetsUtil() {
 			int tumorMIRNA = getCount(cancerStudy, "_microrna");
 			int methylationHM27 = getCount(cancerStudy, "_methylation_hm27");
 			int rppa = getCount(cancerStudy, "_rppa");
+			int massSpec = getCount(cancerStudy, "_ms_abundance");
```
pieterlukasse: note, once you update the list of allowed types (according to my first comment) you may need to update this one...

jjgao:
I am not sure if we add a new data type based on technology. Another way to do it is to model it as PROTEIN_EXPRESSION data, as what Pam has done for her GSoC project.

fedde-s:
Note that 1.4.x adds a list of stable ID suffixes to the new web API for sample lists used to be used on the datasets page, in [this commit](https://github.com/cBioPortal/cbioportal/commit/3ff802493c3ec453a95664ecc1aab4c63da5fae5#diff-ffcae3c86b808d3056122367db0f3312R27). `_ms_abundance` should probably be added there, as well as under [the relevant heading in the documentation on data files](https://github.com/cBioPortal/cbioportal/blob/5d909b86dd3119803848825dd515c049f406a3d5/docs/File-Formats.md#case-lists).

sandertan:
Hi @jjgao , I think it could be useful to have it as a separate column because it's the only way to get a quick overview which of the studies have mass spec data. mRNA is also split based on technology.

If you're confident that a combined PROTEIN_EXPRESSION column is better, I can make the adjustments for it. The question remains how to sum up the samples when both tech is available: a sample can only be counted once, or twice (and thereby the total can exceed the total nr of samples).

Related to this, we're working on the GSVA project, and wondering if a GSVA column would conceptually fit in the datasets page. Could be yes (how else to know which samples have GSVA data), or no because it's based on gene expression data and increases complexity of the table.

sandertan:
The functionality for the validator of this PR is included in https://github.com/cBioPortal/cbioportal/pull/2310. Adding a "Mass Spec"-column to the datasets tab or combining it with RPPA can be done in a future PR.

PR #1743, opened 2016-09-30
Implemented 4 APIs in Clinical Data with persistence, service and web tests

ersinciftci:
Implemented 4 APIs in Clinical Data:

GET /studies/{studyId}/samples/{sampleId}/clinical-data
GET /studies/{studyId}/patients/{patientId}/clinical-data
GET /studies/{studyId}/clinical-data
POST /clinical-data/fetch

with persistence, service and web tests.

### DIFF
--- a/model/src/main/java/org/cbioportal/model/Sample.java
+++ b/model/src/main/java/org/cbioportal/model/Sample.java
@@ -14,14 +14,31 @@ public class Sample extends SampleSummary {
         BLOOD_NORMAL("Blood Derived Normal"),
         SOLID_NORMAL("Solid Tissues Normal");
 
-        private String name;
+        private String value;
 
-        SampleType(String name) {
-            this.name = name;
+        SampleType(String value) {
+            this.value = value;
         }
 
-        public String getName() {
-            return name;
+        public String getValue() {
+            return value;
+        }
+
+        public static SampleType fromString(String value) {
+
+            if (value != null) {
+                for (SampleType sampleType : SampleType.values()) {
+                    if (value.equalsIgnoreCase(sampleType.value)) {
+                        return sampleType;
+                    }
+                }
+            }
+            return null;
+        }
+
+        @Override
+        public String toString() {
+            return value;
         }
     }
 
--- a/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepositoryTest.java
+++ b/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepositoryTest.java
@@ -1,0 +1,589 @@
+package org.cbioportal.persistence.mybatis;
+
+import org.cbioportal.model.*;
+import org.cbioportal.model.meta.BaseMeta;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Configurable;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration("/testContextDatabase.xml")
+@Configurable
+public class ClinicalDataMyBatisRepositoryTest {
+
+    @Autowired
+    private ClinicalDataMyBatisRepository clinicalDataMyBatisRepository;
+
+    @Test
+    public void getAllClinicalDataOfSampleInStudySingleStudyNullSampleNullAttributeIdProjection() throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("study_tcga_pub");
+        List<SampleClinicalData> result =
+                clinicalDataMyBatisRepository.getAllClinicalDataOfSampleInStudy(studyIds, null, null, "ID", null, null,
+                        null, null);
+
+        Assert.assertEquals(6, result.size());
+        SampleClinicalData data = result.get(0);
+        Assert.assertEquals("DAYS_TO_COLLECTION", data.getAttrId());
+        Assert.assertEquals((Integer) 1, data.getInternalId());
+        Assert.assertNull(data.getClinicalAttribute());
+        Assert.assertNull(data.getSample());
+        Assert.assertNull(data.getAttrValue());
+    }
+
+    @Test
+    public void getAllClinicalDataOfSampleInStudyEmptyResult() throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("invalid_study");
+        List<SampleClinicalData> result =
+                clinicalDataMyBatisRepository.getAllClinicalDataOfSampleInStudy(studyIds, null, null, "ID", null, null,
+                        null, null);
+
+        Assert.assertEquals(0, result.size());
+    }
+
+    @Test
+    public void getAllClinicalDataOfSampleInStudySingleStudyNullSampleNullAttributeSummaryProjection()
+            throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("study_tcga_pub");
+        List<SampleClinicalData> result =
+                clinicalDataMyBatisRepository.getAllClinicalDataOfSampleInStudy(studyIds, null, null, "SUMMARY", null,
+                        null, null, null);
+
+        Assert.assertEquals(6, result.size());
+        SampleClinicalData data = result.get(0);
+        Assert.assertEquals("DAYS_TO_COLLECTION", data.getAttrId());
+        Assert.assertEquals("276", data.getAttrValue());
+        Assert.assertEquals((Integer) 1, data.getInternalId());
+        Assert.assertNull(data.getClinicalAttribute());
+        Assert.assertNull(data.getSample());
+    }
+
+    @Test
+    public void getAllClinicalDataOfSampleInStudySingleStudyNullSampleNullAttributeDetailedProjection()
+            throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("study_tcga_pub");
+        List<SampleClinicalData> result =
+                clinicalDataMyBatisRepository.getAllClinicalDataOfSampleInStudy(studyIds, null, null, "DETAILED", null,
+                        null, null, null);
+
+        Assert.assertEquals(6, result.size());
+        SampleClinicalData data = result.get(0);
+        Assert.assertEquals("OTHER_SAMPLE_ID", data.getAttrId());
+        Assert.assertEquals("5C631CE8-F96A-4C35-A459-556FC4AB21E1", data.getAttrValue());
+        Assert.assertEquals((Integer) 1, data.getInternalId());
+        Sample sample = data.getSample();
+        Assert.assertEquals((Integer) 1, sample.getInternalId());
+        Assert.assertEquals((Integer) 1, sample.getPatientId());
+        Assert.assertEquals(Sample.SampleType.PRIMARY_SOLID_TUMOR, sample.getSampleType());
+        Assert.assertEquals("TCGA-A1-A0SB-01", sample.getStableId());
+        Assert.assertEquals("brca", sample.getTypeOfCancerId());
+        Patient patient = sample.getPatient();
+        Assert.assertEquals((Integer) 1, patient.getInternalId());
+        Assert.assertEquals("TCGA-A1-A0SB", patient.getStableId());
+        Assert.assertEquals((Integer) 1, patient.getCancerStudyId());
+        CancerStudy cancerStudy = patient.getCancerStudy();
+        Assert.assertEquals((Integer) 1, cancerStudy.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", cancerStudy.getCancerStudyIdentifier());
+        Assert.assertEquals("TCGA, Nature 2012", cancerStudy.getCitation());
+        Assert.assertEquals("<a href=\\\"http://cancergenome.nih.gov/\\\">The Cancer Genome Atlas (TCGA)</a> Breast " +
+                "Invasive Carcinoma project. 825 cases.<br><i>Nature 2012.</i> <a href=\\\"http://tcga-data.nci.nih." +
+                "gov/tcga/\\\">Raw data via the TCGA Data Portal</a>.", cancerStudy.getDescription());
+        Assert.assertEquals("SU2C-PI3K;PUBLIC;GDAC", cancerStudy.getGroups());
+        Assert.assertEquals("Breast Invasive Carcinoma (TCGA, Nature 2012)", cancerStudy.getName());
+        Assert.assertEquals("brca", cancerStudy.getTypeOfCancerId());
+        Assert.assertEquals("23000897", cancerStudy.getPmid());
+        TypeOfCancer typeOfCancer = cancerStudy.getTypeOfCancer();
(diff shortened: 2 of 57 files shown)


ersinciftci:
Removed `refactoring` label since this is a new API and new code, not refactoring old code.

service/src/test/java/org/cbioportal/service/impl/ClinicalDataServiceImplTest.java
```diff
@@ -0,0 +1,249 @@
...
+                .thenReturn(expectedBaseMeta);
+
+        BaseMeta result = clinicalDataService.fetchMetaClinicalData(studyIds, sampleIds, ATTRIBUTE_ID,
+                ClinicalDataServiceImpl.SAMPLE_CLINICAL_DATA_TYPE);
+
+        Assert.assertEquals((Integer) 5, result.getTotalCount());
+    }
+
+    @Test
+    public void fetchMetaClinicalDataPatientClinicalDataType() throws Exception {
```
sheridancbio: These tests really cannot fail, but they are good placeholders for testing any added domain logic in the service layer code later.

web/src/main/java/org/cbioportal/web/ClinicalDataController.java
```diff
@@ -13,54 +24,125 @@
...
-                                                                                                 @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_NUMBER) Integer pageNumber) {
+    @Autowired
+    private ClinicalDataService clinicalDataService;
 
-        throw new UnsupportedOperationException();
+    @RequestMapping(value = "/studies/{studyId}/samples/{sampleId}/clinical-data", method = RequestMethod.GET,
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ApiOperation("Get all clinical data of a sample in a study")
+    public ResponseEntity<List<? extends ClinicalDataSummary>> getAllClinicalDataOfSampleInStudy(
+            @PathVariable String studyId,
```
sheridancbio: this is the style of declaration of the new rest-style inline parameters

web/src/main/java/org/cbioportal/web/ClinicalDataController.java
```diff
@@ -13,54 +24,125 @@
...
+            @PathVariable String sampleId,
+            @RequestParam(required = false) String attributeId,
+            @RequestParam(defaultValue = "SUMMARY") Projection projection,
+            @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_SIZE) Integer pageSize,
```
sheridancbio: These parameters could benefit from @ApiParam human-understandable explanations of the behavior of arguments. (Unless they are explained in the overall ApiOperation comment). For instance, if there is an absolute upper limit on the user specifiable page size, it could be commented on here, as well as what will happen on an out-of-range error (parameter forced in to bounds .. or error code ... or null result?)
Adding these comments could be a future improvement.

web/src/main/java/org/cbioportal/web/ClinicalDataController.java
```diff
@@ -13,54 +24,125 @@
...
+            @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_NUMBER) Integer pageNumber,
+            @RequestParam(required = false) ClinicalDataSort sort,
+            @RequestParam(defaultValue = "ASC") Direction direction) {
```
sheridancbio: It looks like this may be SQL specific. If this parameter gets inlined into the mybatis select statement, I think it would be wise to insert one level of indirection and not make it a SQL keyword. (open to discussion obviously)
I see below that this maps to a Direction Enum rather than direct to the mybatis level, so ok.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
```diff
@@ -43,6 +44,7 @@
     public CustomObjectMapper() {
 
         super.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        super.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
```
sheridancbio: good. But since the names of the enum values are used (if I understand correctly), we should make a note in the enum that any change to the enum names will be an alteration in the output of the REST controller. [to prevent unintended API alteration by someone who thinks they can improve the enum constant names].

web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
```diff
@@ -38,4 +43,60 @@
         return new ResponseEntity<>(new ErrorResponse("Sample not found: " + ex.getSampleId()),
                 HttpStatus.NOT_FOUND);
     }
+
+    @ExceptionHandler(MissingServletRequestParameterException.class)
```
sheridancbio: This functionality is super-handy

web/src/main/java/org/cbioportal/web/interceptor/PagingValidationInterceptor.java
```diff
@@ -0,0 +1,49 @@
+package org.cbioportal.web.interceptor;
```
sheridancbio: This was discussed for other interface aspect uses, such as security and permission checking. Another candidate: cacheing.

web/src/main/java/org/cbioportal/web/parameter/PagingConstants.java
```diff
@@ -3,8 +3,8 @@
 public class PagingConstants {
```
sheridancbio: Probably should be combined with other staticly defined adjustable constants, or be made into properties that can be configured eternally to the code (portal.properties, etc). But this seems ok for a start. We should think whether or not people would actually want to tune these .

web/src/test/java/org/cbioportal/weblegacy/ApiControllerTest.java
```diff
@@ -1,4 +1,4 @@
-package org.cbioportal.web.api;
+package org.cbioportal.weblegacy;
```
sheridancbio: good move

sheridancbio:
In hand testing the api I have run into an exception.
The table clinical_attribute no longer exists but it is used in ClinicalDataMapper.
This table was reconfigured into clinical_attribute_meta by @zheins in the migration script for database version 1.3.0. I am not sure if the query should now be against clinical_attribute_meta or not. Maybe @ersinciftci and @zheins can comment about why the table was changed and whether it still meets the needs of the new REST api endpoint "clincal-data/fetch"

model/src/main/java/org/cbioportal/model/Sample.java
```diff
@@ -14,14 +14,31 @@
         BLOOD_NORMAL("Blood Derived Normal"),
         SOLID_NORMAL("Solid Tissues Normal");
```
ersinciftci: We can access it anywhere actually http://stackoverflow.com/a/6280981 . For example I access it in `org.cbioportal.persistence.mybatis.typehandler.SampleTypeTypeHandler`.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,508 @@
...
+    public void getAllClinicalDataOfPatientInStudySingleStudyNullPatientNullAttributeDetailedProjection()
+            throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("study_tcga_pub");
+        List<PatientClinicalData> result =
+                clinicalDataMyBatisRepository.getAllClinicalDataOfPatientInStudy(studyIds, null, null, "DETAILED", null,
+                        null, null, null);
+
+        Assert.assertEquals(4, result.size());
```
ersinciftci: Since we are testing the correctness of a SQL select query, I think it's enough to check the size and first element, because you can't really return a different kind of second or third element from a select query.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,508 @@
...
+        studyIds.add("study_tcga_pub");
+        studyIds.add("acc_tcga");
+        List<String> patientIds = new ArrayList<>();
+        patientIds.add("TCGA-A1-A0SB");
+        patientIds.add("TCGA-A1-B0SO");
+        BaseMeta result = clinicalDataMyBatisRepository.getMetaPatientClinicalData(studyIds, patientIds, null);
+
+        Assert.assertEquals((Integer) 7, result.getTotalCount());
+    }
+}
```
ersinciftci: You are right, I'll add a few negative tests now.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/CancerTypeMapper.java
```diff
@@ -0,0 +1,4 @@
+package org.cbioportal.persistence.mybatis;
+
+public interface CancerTypeMapper {
```
ersinciftci: In the `DETAILED` projection of Clinical Data, we return patient,study and cancer type information too, and instead of writing select statement for those tables in every MyBatis script, I reuse them from their corresponding script using the `include` tag. But MyBatis doesn't see those scripts unless I create an interface for them. So they are empty for now but they are required to use sql fragments from other scripts.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -0,0 +1,186 @@
...
+            ,
+            <include refid="org.cbioportal.persistence.mybatis.ClinicalAttributeMapper.select">
+                <property name="prefix" value="${prefix}clinicalAttribute."/>
+            </include>
+        </if>
+
+    </sql>
+
+    <sql id="fromSample">
+
```
ersinciftci: Right, deleting them now.

web/src/main/java/org/cbioportal/web/ClinicalDataController.java
ersinciftci: Right, I'll add them.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
ersinciftci: No, Enum names aren't used, it's still the property name like before, this line just changes ObjectMapper to use .toString() method of enum instead of .name() method, because .name() returns PRIMARY_SOLID_TUMOR, but my overriden toString() returns `Primary Solid Tumor`, and we want the second one.

ersinciftci:
Yeah I need to change the table name to clinical_attribute_meta in ClinicalAttributeMapper.xml, changing now.

model/src/main/java/org/cbioportal/model/Sample.java
sheridancbio: ok

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,508 @@
...
+    @Test
+    public void getAllClinicalDataOfSampleInStudySingleStudySingleSampleNullAttributeSummaryProjection2PageSize()
+            throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("study_tcga_pub");
+        List<String> sampleIds = new ArrayList<>();
+        sampleIds.add("TCGA-A1-A0SB-01");
+        List<SampleClinicalData> result =
+                clinicalDataMyBatisRepository.getAllClinicalDataOfSampleInStudy(studyIds, sampleIds, null, "SUMMARY",
```
sheridancbio: ok

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepositoryTest.java
sheridancbio: right .. I was mixing this layer up with the controller

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/CancerTypeMapper.java
sheridancbio: ok

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
sheridancbio: ok

sheridancbio:
I've scanned for other uses of clinical_attribute table, and there are references in ./persistence/persistence-mybatis-test/src/test/resources/testSql.sql
@zheins says that these were missed when clinical_attribute_meta was added.

If you are making your updates and have time, you could throw in that fix too. (change table name to clinical_attribute_meta, add a field in the field list, and add a reference to a single study for the extra field). I'll add details to this comment.

ersinciftci:
Ok I already did that change.

sheridancbio:
I don't see the updates to testSql.sql yet ... it still refers to "clinical_attribute".
Let me know when you have updated the PR

ersinciftci:
Yeah writing the @ApiParam parts, will let you know when I push.

sheridancbio:
@zheins and I discussed the response type and we thought that there was room for improving performance, since there is likely to be a great deal of repeated gene list data in each query. On a query with 1000 samples, and 100 genes, if the genes are all present in the gene panel which was used to sequence the samples, and all (or almost all) samples used the same gene panel, then the same list of 100 genes will be returned 1000 times in the response.

An alternative which has one level of indirection / data compression would be to return two lists:
- list 1 = a list of integer index numbers into list 2 (one index number per sample)
- list 2 = a unique list of gene lists. If there are multiple gene panels used, there might be multiple gene lists showing how each panel intersected with the set of genes queried to produce the oncoprint, but there would be no more members in list 2 than the number of different gene panels used in the profile.

A fuller / flatter / more streamlined backend/frontend interface could be a two step process --- first query the profile and get a mapping of samples to the gene panel used (per sample). For each (unique) used gene panel, a second set of queries would determine which of the genes used to construct the oncoprint intersect with the genes in the gene-panel.  Finally the frontend could construct a full data structure from the two steps above to display the oncoprint markings.

ersinciftci:
I did all the changes and pushed. Not sure if @sheridancbio 's last comment is related to this PR.

sheridancbio:
Sorry. My previous comment here was related to the gene panel api .. I've moved the comment to the correct place.

inodb:
@ersinciftci might be good to specify some example request in the generated swagger docs. That will also help to show which API endpoints are implemented

sheridancbio:
Tested API on local deployment.. merging

ersinciftci:
@inodb since we are generating the swagger-ui docs from the code, I think it's not a good idea to hardcode some default values in the code since cbioportal can be used with any kind of data, not only TCGA. And I don't think swagger-ui supports example values for path variables, it supports default values but it's not what we want, we want just an example but not make it default if it doesn't exist in an API call.

inodb:
@ersinciftci OK. I guess you don't want to fill it in with some random data plucked from the db, because you don't know if it will look any good.

Is there a way to provide mock data?

This is all not super critical, but would be useful when we eventually make the API publicly available.

ersinciftci:
@inodb The best I can do is provide some TCGA values in `Description` column of the parameter, so it will be like `Study ID e.g. acc_tcga`, so people can at least copy paste from there, and it wouldn't change any functionality in the code, sounds good?

inodb:
Sounds good!

PR #1746, opened 2016-10-03
Associate clinical attributes with cancer study

zheins:
# What? Why?
For performance reasons and to allow greater flexibility of studies to have their own clinical attribute namespace, this PR associates clinical attributes to cancer studies.

- Add column to `clinical_attribute` table corresponding to the `CANCER_STUDY_ID` column in the `cancer_study` table.
- Update usages of this table in importer, Dao objects, and api calls
- Update the study view to use the api appropriately by only making one call to get the clinical attributes.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@zhx828 
@jjgao 
@n1zea144


zheins:
@jjgao @zhx828 this is ready to review

kalletlak:
@zheins I think clinical attribute data would be duplicated if we have same attribute for different studies. Instead why dont we have new table mapping primary keys from 'clinical_attribute' and 'cancer_study' tables

zheins:
@kalletlak The idea is that you can customize the attribute for a study, so they won't necessarily be duplicates. Every study will have its potentially unique version of its attributes. If migrating an old database, there will definitely be duplicate rows though.We could do this mapping still and only put in rows in the clinical_attribute table if it is actually unique. @jjgao what do you think?

jjgao:
@kalletlak: another purpose of this pull request is also try NOT to enforce up-front normalization for all attributes, because that created a lot of issues and burden for our curators. As @zheins said, we wanted to enable study-specific meta data.

We do need modify the validator to make sure the key attributes (age, gender, oncotree_code, etc) are defined consistently across studies. -- Another pull request.

zhx828:
JS changes look good to me. I created a branch to track the changes back to iViz in above reference.

fedde-s:
@jjgao:
> We do need modify the validator to make sure the key attributes (age, gender, oncotree_code, etc) are defined consistently across studies. -- Another pull request.

I have identified the key attributes explicitly used for visualisations in the frontend and added them to the validator in pull request #1567.

business/src/main/resources/org/mskcc/cbio/portal/persistence/ClinicalFieldMapper.xml
```diff
@@ -13,67 +13,66 @@
...
-    select clinical_attribute.ATTR_ID,clinical_attribute.DISPLAY_NAME,clinical_attribute.DESCRIPTION,clinical_attribute.DATATYPE,clinical_attribute.PATIENT_ATTRIBUTE,clinical_attribute.PRIORITY from clinical_attribute
-                inner join clinical_sample on clinical_sample.ATTR_ID = clinical_attribute.ATTR_ID
-                inner join sample on clinical_sample.INTERNAL_ID = sample.INTERNAL_ID
-                inner join patient on sample.PATIENT_ID = patient.INTERNAL_ID
-                inner join cancer_study on patient.CANCER_STUDY_ID=cancer_study.CANCER_STUDY_ID
-    where cancer_study.CANCER_STUDY_IDENTIFIER=#{study_id}
-    group by clinical_attribute.ATTR_ID 
+    select clinical_attribute_meta.ATTR_ID,clinical_attribute_meta.DISPLAY_NAME,clinical_attribute_meta.DESCRIPTION,clinical_attribute_meta.DATATYPE,clinical_attribute_meta.PATIENT_ATTRIBUTE,clinical_attribute_meta.PRIORITY from clinical_attribute_meta
+    where cancer_study.CANCER_STUDY_ID=#{study_id} and clinical_attribute_meta.PATIENT_ATTRIBUTE=0    
+    group by clinical_attribute_meta.ATTR_ID
```
jjgao: Why do we need "group by"?

PR #1753, opened 2016-10-05
Fixes for cbioportalImporter script

pieterlukasse:
# What? Why?

The metaImport/cbioportalImporter.py script was not working in `rc` anymore because the core .jar file has been debloated in one of the previous PR (probably in #1517). The new .jar file to use is called `scripts...jar` instead of `core...jar`. Here we update the importer code to look for this new .jar

- let importer code look for the new "scripts" .jar
- fix to scripts/pom.xml to make sure the `portal.properties` file that gets into the scripts .jar also has its variables correctly replaced during the maven build step

# Notify reviewers
@morungos @zheins


zheins:
This looks good to me @pieterlukasse. I've been using the scripts package in all of our MSK imports for the last few months, this is a good change.

PR #1754, opened 2016-10-05
Fix for when "relaxed" parameter is not set

pieterlukasse:
# What? Why?
When "relaxed mode" parameter is not set via the command line, the code was failing. 

- fix to let the validation script work, also when "relaxed mode" is not set

# Notify reviewers
@angelicaochoa @zheins


n1zea144:
@pieterlukasse Just merged 1753.  Can you update this branch with rc now?

pieterlukasse:
@n1zea144 would be good if @zheins can confirm this still works well with the relaxed mode. I only tested the other mode ;)

PR #1755, opened 2016-10-05
DB schema version fixes

pieterlukasse:
# What? Why?
The PR #1618 changed `migration.sql` but forgot to update pom.xml `db.version`. Here we fix this and also add some extra info to the console that helps troubleshooting db schema version mismatches. 

- correct db schema version in pom.xml
- better reporting on what the schema version check found (specially useful when running data loading from the command line)

# Notify reviewers
@zheins @sheridancbio

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoInfo.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoInfo.java
@@ -54,6 +54,7 @@ public class DaoInfo {
             rs = pstmt.executeQuery();
             if (rs.next()) {
                 version = rs.getString("DB_SCHEMA_VERSION");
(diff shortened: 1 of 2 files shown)


sheridancbio:
Ah .. I see. The versionCheck() function is only called by VersionUtil.java which is an application (presumably used by the importer or validator).  Ok.. I will add System.out.println functionality in VersionUtil.java too.

jjgao:
Will this pr superseded by https://github.com/cBioPortal/cbioportal/pull/1763?

sheridancbio:
yes, when PR #1763 is merged, this functionality will be in the codebase.

pieterlukasse:
Ok, #1763 seems to cover this. Although having the info in catalina.out could also be nice ;)  Anyway, I'll close this one. Thanks!

Issue #1758, opened 2016-10-05
"Nucleotide changes of point mutations" not working in oncoprint


jjgao:
Not sure if this has been fixed.

![image](https://cloud.githubusercontent.com/assets/840895/19113989/0efbf792-8ada-11e6-91da-395809e9ee9c.png)

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1772

PR #1772, opened 2016-10-07
Fix "Nucleotide change of point mutations" clinical track

adamabeshouse:
# What? Why?

Fix https://github.com/cBioPortal/cbioportal/issues/1758
# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.
# Any screenshots or GIFs?

# Notify reviewers

@jjgao


jjgao:
@adamabeshouse there is significant backend change in this pull request. Maybe we can postpone this for the next release. 

@sheridancbio Could you review the code? Thanks.

sheridancbio:
sure. I think there was code for countSNP in the rearchitecture efforts on the mutation tech stack, so I can compare. @ersinciftci  may be interested to look at this implementation too, and consider approaches / endpoints for the Rest API

adamabeshouse:
@jjgao somehow it's still running out of memory too..

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -118,25 +119,10 @@
...
-			}
-			mutationsBySample.get(id).add(mutation);
-		}
-		List<MutationSignature> signatures = new LinkedList<>();
-		if (context_size_on_each_side_of_snp == 0) {
-			for (Map.Entry kv: mutationsBySample.entrySet()) {
-				signatures.add(MutationSignatureFactory.NoContextMutationSignature((String)kv.getKey(), (List<Mutation>)kv.getValue()));
-			}
-		}
+		List<SNPCount> snps = mutationService.getSNPCounts(genetic_profile_id, sample_ids);
```
sheridancbio: This is a new model type (SNPCount) to hold a SNP with a count.

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -118,25 +119,10 @@
 	}
 
 	public List<MutationSignature> getSampleMutationSignatures(String genetic_profile_id, List<String> sample_ids, int context_size_on_each_side_of_snp) {
```
sheridancbio: This is an expansion of the older web api, in which all services were supplied through a single service class. We should consider implementing this as a separate standalone stack of classes (in modules "persistence-*", "service", and "web"). It could become a controller in the weblegacy (api-legacy) location.

The reason for considering this is that we are requiring newly developed API code to have unit tests at all three levels, and also because the business module is about to be broken up, and this functionality will be relocated anyway. The ApiController and ApiService classes combine multiple diverse endpoints and should probably be broken up into smaller related components (by related data types) ... or we can wait for them to migrate into the new REST api arrangement.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MutationMapper.xml
```diff
@@ -362,5 +362,25 @@
...
+            count(*) as count
+        from sample,mutation_event,mutation,genetic_profile
+        where mutation_event.MUTATION_EVENT_ID=mutation.MUTATION_EVENT_ID and
+            mutation.SAMPLE_ID=sample.INTERNAL_ID and
+            mutation.GENETIC_PROFILE_ID=genetic_profile.GENETIC_PROFILE_ID and
+            genetic_profile.STABLE_ID=#{geneticProfileStableId} and
+            mutation_event.REFERENCE_ALLELE in ("A","T","C","G") and
+            mutation_event.TUMOR_SEQ_ALLELE in ("A","T","C","G") and
+            sample.STABLE_ID in
+            <foreach item="item" collection="sampleStableIds" open="(" separator="," close=")">
```
adamabeshouse: check if empty with <if>

sheridancbio:
This PR was discussed with @adamabeshouse and @ersinciftci yesterday.

**Web Layer**
Our recommendation was to create a new controller in the org.cbioportal.weblegacy package (with endpoint in {appname}/web-legacy/{endpointname} pattern. For example see:
https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/weblegacy/GenePanelController.java
(there are 2 endpoints under this controller)
Web layer tests should verify that calls to the api endpoint succeed (or fail when appropriate) and that the JSON formatting of the response is basically correct. Tests for the GenePanelController use mockMVC to simulate a running web server and mock the service layer functions with mockito.
https://github.com/cBioPortal/cbioportal/blob/master/web/src/test/java/org/cbioportal/weblegacy/GenePanelControllerTest.java
(also, there is a configuration class in the same package to accomplish the mocking)

**Service Layer**
If all functionality can be accomplished in SQL statements, the service layer often becomes a set of passthrough functions calling the relevant methods in the persistence layer. But some functionality and transformations may be better implemented here.
https://github.com/cBioPortal/cbioportal/blob/master/service/src/main/java/org/cbioportal/service/GenePanelService.java
Unit testing may not be meaningful at this level, unless there is domain coding in the service methods.

**Persistence Layer**
We have abstracted an API to the persistence layer, represented by Repository interfaces residing in the persistence-api maven module. Example:
https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-api/src/main/java/org/cbioportal/persistence/GenePanelRepository.java
Implementations of these interfaces is done (because we have been using mybatis framework for implementations) in persistence-mybatis module, using a repository class and a mybatis mapper class linked to a .xml 
See:
(mybatis repository class) https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenePanelMyBatisRepository.java
(mybatis mapper class) https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenePanelMapper.java
(mybatis xml specification) https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenePanelMapper.xml

Unit testing at the persistence layer use a testing database created from:
https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis-test/src/test/resources/testSql.sql
this file probably will need to be adjusted to include cases of mutation events which cover representative nucleotide substitutions.
tests are written in the persistence-mybatis-test module. Example:
https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GenePanelMyBatisRepositoryTest.java
In these unit tests, functions of the Repository are called, and the expected return results (based on the test data in testSql.sql) are asserted.

** Models **
Model classes belong in the model module. They should be POJO classes with only getters and setters.
Example:
https://github.com/cBioPortal/cbioportal/blob/master/model/src/main/java/org/cbioportal/model/GenePanel.java

Those are the main layers/modules to use to organize the code. Sometimes utility classes/functions are created as helpers.
(service layer custom exceptions) https://github.com/cBioPortal/cbioportal/tree/master/service/src/main/java/org/cbioportal/service/exception
(web layer custom argument processing) https://github.com/cBioPortal/cbioportal/tree/master/web/src/main/java/org/cbioportal/web/parameter

sheridancbio:
I just reran the travis tests. compile seems to be failing based on failure to find this method: getOrDefault(java.lang.String,<nulltype>)
Maybe a dependency is missing.

adamabeshouse:
@sheridancbio I guess we're not using the latest version of Java, that's a standard library Map feature in Java 8. I just replaced it.

jjgao:
@adamabeshouse @sheridancbio any remaining items for this pr?

sheridancbio:
This is one weakness of github -- when you squash / --amend your commits, the date of commit winds up being earlier than it should be. So I cannot tell when when the PR was last updated, or what changes have occurred since the review.

We are going to talk about this on Wednesday .. so design my change based on that conversation. But if this is ready now and we want to push it out before the Engineering Review meeting, I guess we could (since this api is in web-legacy).  Is it ready for another review @adamabeshouse ?

jjgao:
@sheridancbio maybe we don't have to squash every time? 

Any update on this?

sheridancbio:
@jjgao I just slacked @adamabeshouse ... I can review if all changes are in. Also I requested rebase to rc, so that end-to-end tests will pass.

sheridancbio:
review in progress

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/MutationalSignatureMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private MutationalSignatureMyBatisRepository mutationalSignatureMyBatisRepository;
+	
+	@Test
+	public void getSNPCountsByGeneticProfile() {
+		List<SNPCount> snpCounts = mutationalSignatureMyBatisRepository.getSNPCounts("study_tcga_pub_mutations");
+		
+		Assert.assertEquals(8, snpCounts.size());
+	}
+	
+	@Test
```
sheridancbio: It appears that none of the tests check the case where a query is made for a sample which does not exist. We should add that test to make sure no exception is thrown .. and it would be better to add a check where some of the requested sampleIds do match and some do not. (a mixed query) .. and make sure that the output is of the correct size.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/MutationMyBatisRepository.java
```diff
@@ -12,6 +12,7 @@
 
 import java.util.Arrays;
 import java.util.List;
+import org.cbioportal.model.SNPCount;
```
sheridancbio: why is this imported but not used?

adamabeshouse: deleted

web/src/main/java/org/cbioportal/weblegacy/MutationalSignatureController.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import org.cbioportal.model.MutationalSignature;
+import org.cbioportal.service.MutationalSignatureService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.bind.annotation.*;
+
+/**
+ *
+ * @author heinsz
```
sheridancbio: seems like this is not right .. but I don't like @author tags anyway

web/src/main/java/org/cbioportal/weblegacy/MutationalSignatureController.java
```diff
@@ -0,0 +1,71 @@
...
+public class MutationalSignatureController {
+
+    @Autowired
+    private MutationalSignatureService mutationalSignatureService;
+
+    @ApiOperation(value = "Get mutation signatures for given samples in a study gene panel information",
+            nickname = "getGenePanel",
+            notes = "")
+    @Transactional
+    @RequestMapping(method = {RequestMethod.GET, RequestMethod.POST}, value = "/mutational-signature")
```
sheridancbio: we messaged a bit about GET v. POST and when to use one or the other. This is in weblegacy, and the "handle both request types" was normal in ApiController. We should have a bigger discussion as a development team about how to use request types in the REST API

web/src/test/java/org/cbioportal/weblegacy/MutationalSignatureControllerTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.springframework.test.context.web.WebAppConfiguration;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders;
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.WebApplicationContext;
+
+/**
+ *
+ * @author heinsz
```
sheridancbio: not updated .. but again, @author tags are not required

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -791,34 +791,28 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
 		fetch_promises.push(mutation_signatures_promise.promise());
-		self.getGeneticProfiles().then(function (genetic_profiles) {
-		    var mutation_profiles = genetic_profiles.filter(function (x) {
-			return x.genetic_alteration_type === "MUTATION_EXTENDED";
-		    });
-		    $.when.apply($, mutation_profiles.map(function (mutation_profile) {
-			var _def = new $.Deferred();
-			$.ajax({
-			    type: 'POST',
-			    url: 'api-legacy/mutationsignatures',
```
adamabeshouse: This feature has already existed, but it was temporarily disabled because it was broken, running into the out of memory error, which this pull request is intended to fix. At that time, the endpoint was `mutationsignatures`. Does that answer your question?

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1399,15 +1393,15 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
 				    is_patient_attribute: "0"
 				};
-				/*sample_clinical_attributes_set['NO_CONTEXT_MUTATION_SIGNATURE'] = {
+				sample_clinical_attributes_set['NO_CONTEXT_MUTATION_SIGNATURE'] = {
 				 attr_id: "NO_CONTEXT_MUTATION_SIGNATURE",
 				 datatype: "COUNTS_MAP",
 				 description: "Number of point mutations in the sample counted by different types of nucleotide changes.",
 				 display_name: "Nucleotide change of point mutations",
 				 is_patient_attribute: "0",
 				 categories: ["C>A", "C>G", "C>T", "T>A", "T>C", "T>G"],
```
adamabeshouse: This is confusing, but I think the plus side is this decouples the display order from the order the data is served in.

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+				referenceAllele = COMPLEMENTARY_NUCLEOTIDE.get(referenceAllele);
+				tumorAllele = COMPLEMENTARY_NUCLEOTIDE.get(tumorAllele);
+				if (referenceAllele == null || tumorAllele == null) {
+					// Something's wrong
+					return -1;
+				}
+			}
+			Integer index = CANONICAL_SNP_TYPES_INDEX.get(referenceAllele+">"+tumorAllele);
+			if (index == null) {
+				index = -1;
```
adamabeshouse: That's good to note, but I don't think it's relevant here, because it refers to a definitive mutation event, which wouldn't have the ambiguous nucleotide?

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+			}
+			
+			for (SNPCount snp: snpCounts) {
+				String sampleId = snp.getSampleId();
+				int[] signature = signatureCounts.get(sampleId);
+				if (signature != null) {
+					// Increment appropriate index
+					int snpIndex = getCanonicalSNPIndex(snp.getReferenceAllele(), snp.getTumorAllele());
+					if (snpIndex > -1) {
+						signature[snpIndex] += snp.getCount();
```
adamabeshouse: See comment above

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+				mutationProfile = gp;
+				break;
+			}
+		}
+		return mutationProfile;
+	}
+	public List<MutationalSignature> getMutationalSignaturesBySampleIds(String study_id, List<String> sample_ids) {
+		GeneticProfile mutationProfile = getMutationProfile(study_id);
+		if (mutationProfile == null) {
+			return new ArrayList<>();
```
adamabeshouse: good idea. We can do that here, or should we wait to discuss it?

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MutationalSignatureMapper.xml
```diff
@@ -0,0 +1,28 @@
...
+<mapper namespace="org.cbioportal.persistence.mybatis.MutationalSignatureMapper">
+    <cache/>
+
+    <select id="getSNPCountsBySampleId" resultType="org.cbioportal.model.SNPCount">
+        select
+            sample.STABLE_ID as sampleId,
+            mutation_event.REFERENCE_ALLELE as referenceAllele,
+            mutation_event.TUMOR_SEQ_ALLELE as tumorAllele,
+            count(*) as count
+        from sample,mutation_event,mutation,genetic_profile
```
adamabeshouse: done

service/src/test/java/org/cbioportal/service/impl/MutationalSignatureServiceImplTest.java
```diff
@@ -0,0 +1,44 @@
...
+	    sampleIds.add("sample1");
+	    
+	    List<MutationalSignature> signatures = MutationalSignatureServiceImpl.MutationalSignatureFactory.NoContextSignatures(testSNPCounts, sampleIds);
+	    Assert.assertTrue(signatures.size() == 1);
+	    MutationalSignature signature = signatures.get(0);
+	    Assert.assertTrue(signature.getSample().equals("sample1"));
+	    Assert.assertTrue(Arrays.equals(signature.getMutationTypes(), new String[]{"C>A", "C>G", "C>T", "T>A", "T>C", "T>G"}));
+	    Assert.assertTrue(Arrays.equals(signature.getCounts(), new int[]{20, 0, 45, 15, 1, 0}));
+    }
+}
```
adamabeshouse: done

service/src/test/java/org/cbioportal/service/impl/MutationalSignatureServiceImplTest.java
```diff
@@ -0,0 +1,44 @@
...
+	    testSNPCounts.add(new SNPCount("sample1", "A", "T", 15));
+	    testSNPCounts.add(new SNPCount("sample1", "C", "T", 5));
+	    testSNPCounts.add(new SNPCount("sample1", "G", "T", 20));
+	    testSNPCounts.add(new SNPCount("sample1", "C", "T", 40));
+	    testSNPCounts.add(new SNPCount("sample1", "T", "C", 1));
+	    
+	    List<String> sampleIds = new LinkedList<>();
+	    sampleIds.add("sample1");
+	    
+	    List<MutationalSignature> signatures = MutationalSignatureServiceImpl.MutationalSignatureFactory.NoContextSignatures(testSNPCounts, sampleIds);
```
adamabeshouse: is this bad?

web/src/main/java/org/cbioportal/weblegacy/MutationalSignatureController.java
adamabeshouse: deleted

portal/src/main/webapp/js/api/cbioportal-datamanager.js
sheridancbio: yes, got it

portal/src/main/webapp/js/api/cbioportal-datamanager.js
sheridancbio: ok

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+@Service
+public class MutationalSignatureServiceImpl implements MutationalSignatureService {
+
+	public static class MutationalSignatureFactory {
+		private static final Map<String, String> COMPLEMENTARY_NUCLEOTIDE = new HashMap<>();
+		private static final String[] CANONICAL_SNP_TYPES = new String[]{"C>A", "C>G", "C>T", "T>A", "T>C", "T>G"};
+		private static final Map<String, Integer> CANONICAL_SNP_TYPES_INDEX;
+		
+		static {
+			COMPLEMENTARY_NUCLEOTIDE.put("A","T");
```
sheridancbio: sure .. let's wait. There may be some excellent third party libraries too for representing nucleotide sequences (with operations like reverse complement), amino acid sequences, codon mappings and non-synonymous mutation detection, etc. We should probably prefer an established library to crafting our own.

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
sheridancbio: I didn't understand the use cases. I think I was imagining that our mutation event profiles might involve calls such as A -> Y (where Y is an ambiguous code for T or C). dbSNP can also contain ambiguous nucleotide codes (according to ucsc browser documentation). Maybe when we construct mutation profiles we don't import cases like this ... or don't get them from the platforms which do the calling.

sheridancbio: also .. maybe what matters is not the mutant nucleotide but the reference genome nucleotide ... and maybe we don't make mutation event calls at the positions in the reference genome which are ambiguous.

adamabeshouse: @jjgao can you weigh in on this?

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
sheridancbio: I think we should wait .. but remember to address it when migrating this endpoint to the public /api/ area.

service/src/test/java/org/cbioportal/service/impl/MutationalSignatureServiceImplTest.java
sheridancbio: I was just making a comment to myself ... so I remember what the test function is testing (during a second review or readthrough of the pull request)

web/src/main/java/org/cbioportal/weblegacy/MutationalSignatureController.java
sheridancbio: I think it is ok to keep it as both (POST + GET) for now. We should have a bigger design discussion about HTTP request types and how we want to use them in the REST API ... once that is clear we can adjust mutational-profiles when it moves to the public area.

service/src/main/java/org/cbioportal/service/impl/MutationalSignatureServiceImpl.java
adamabeshouse: In dashi-dev, there are no ambiguous nucleotide codes for the reference allele, but the tumor seq allele I see L, E, R, and N cases. It appears here: http://www.bioinformatics.org/sms/iupac.html that L and E are not nucleotide codes, which makes me wonder whether N and R are the nucleotide codes or whether these are all legacy code which originally referred to amino acids and can now be safely ignored.

adamabeshouse: Talked to @jjgao, we're going to ignore them.

jjgao:
@adamabeshouse how much work do you think it would be if we want to move this forward?

adamabeshouse:
@jjgao not much, I think the only remaining work is to switch to querying by entrezGeneId and to return a not found status code when querying with nonexistent profile id.

jjgao:
@adamabeshouse in that case, let's try to finish it and release.

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/2150

Issue #1757, opened 2016-10-05
Add a button to download segment data in CN Segments tab


jjgao:
to download segment data of all selected samples in all chromosomes.

Format: text
Location: botton right
Text: Download segments

PR #1795, opened 2016-10-11
download segment file for selected samples

JiaoJiao123:
related to issue #1757

#what?
1) add end point in copy number segment api to return file content.
2) call the segment file api in the front end and download .seg file

# Notify reviewers
@jjgao


jjgao:
@JiaoJiao123 could you point this to hotfix since this is a small improvment?

Issue #873, opened 2016-02-29
Make releases available to maven


pieterlukasse:
It would be good to have the releases (now recorded at https://github.com/cBioPortal/cbioportal/releases) and their respective artifacts available as a "repository" for maven. This simplifies building new external Java projects that include cBioPortal as a dependency. 

Ideally we let `travis` do this.

pieterlukasse:
Shared by @jjgao :  https://malalanayake.wordpress.com/2014/03/10/create-simple-maven-repository-on-github/

Maybe this could work if we let travis push up the artifacts to github?

pieterlukasse:
http://stackoverflow.com/questions/12343452/how-to-publish-artifacts-in-travis-ci
and
https://docs.travis-ci.com/user/deployment/releases/
and use use `maven-scm-plugin` ? Or is `maven-scm-plugin` maybe not necessary after all?

pieterlukasse:
@aderidder can help test it as he would like to build another project using cBioPortal as dependency.

inodb:
@pieterlukasse / @aderidder  Is this still something you want to do? Just looked at it today. Excuse the delay

You can actually already use cBioPortal as a dependency if you build it separately first since the artifacts are stored in `~/.m2/org.mskcc/something` (as far as I understand). See for instance @morungos solution: https://github.com/pughlab/cbioportal-octane-overlay

Now if you want to resolve the dependency automatically, instead of building it yourself, we could maybe use jitpack: https://jitpack.io/. Would require some trivial changes to make the building work (use i.e. `portal.properties.EXAMPLE` by default).

inodb:
Ok actually solved this using jitpack, further discussion on #1798

pieterlukasse:
Hi @inodb , the original purpose was to have .jar files available for download, so that people working on preparing the datasets would not need to go through the steps of building cbioportal just to use some of the helper classes that are part of the scripts package. See #810

But having the build artifacts available for download will also simplify other steps, such as installation. 

The jitpack idea might solve this. Where can I find the artifacts?

pieterlukasse:
Ah, found it: https://jitpack.io/com/github/inodb/cbioportal/scripts/v1.3.0-test-jitpack/scripts-v1.3.0-test-jitpack.jar

Nice! 👍

pieterlukasse:
@inodb your solution solves the problem! We can close this issue once we merge your PR 👍

PR #1798, opened 2016-10-11
Make {portal,log4j}.properties.EXAMPLE default fallback

inodb:
Copy the .EXAMPLE property files to their non-EXAMPLE postfixed counterparts (log4j.properties and portal.properties) if they don't exist.

This change allows one to serve the artifacts through [jitpack](https://jitpack.io/) (jitpack runs with default profile), thereby solving #873. See an example with @morungos war overlays here: https://github.com/pughlab/cbioportal-octane-overlay/pull/1/files. 

It also simplifies Travis & heroku specific build process.

Another option would have been to update all references to `portal.properties` to refer to `portal.properties.EXAMPLE` if it did not exist, but this seemed easier.

Also as @pieterlukasse pointed out you can get the artifacts like this (the scripts jar in this case):
https://jitpack.io/com/github/inodb/cbioportal/scripts/v1.3.0-test-jitpack/scripts-v1.3.0-test-jitpack.jar

or using mvn

```
# this points to the latest version of my branch use-env-vars and gets the war artifact
mvn dependency:get -DremoteRepositories=https://jitpack.io -Dartifact=com.github.inodb.cbioportal:cbioportal:use-env-vars-SNAPSHOT:war -Ddest=cbioportal.war
```
 or you can specify it as a dependency in your pom file, as shown in the mvn overlay example further up ☝️ .

## Reviewers
@cBioPortal/devops @n1zea144 @jjgao @pieterlukasse @angelicaochoa

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,68 @@
       <properties>
         <importer-main-class>org.mskcc.cbio.importer.Admin</importer-main-class>
       </properties>
+    <build>
+    <plugins>
+          <!-- copy portal.properties.EXAMPLE and log4j.properties.EXAMPLE if they don't exist -->
+          <plugin>
+            <artifactId>maven-antrun-plugin</artifactId>
+            <version>1.8</version>
+            <!-- only needs to be executed for parent pom -->
+            <inherited>false</inherited>
+            <executions>
+              <execution>
+                <phase>validate</phase>
+                <goals>
+                  <goal>run</goal>
+                </goals>
+                <configuration>
+                  <target>
+                    <taskdef resource="net/sf/antcontrib/antlib.xml" classpathref="maven.dependency.classpath" />
+                    <if>
+                      <not>
+                        <available file="${basedir}/src/main/resources/portal.properties" />
+                      </not>
+                      <then>
+                        <copy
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -131,9 +169,9 @@
 
     <final.war.name>cbioportal-${project.version}</final.war.name>
 
-	<db.test.driver>com.mysql.jdbc.Driver</db.test.driver>
+  <db.test.driver>com.mysql.jdbc.Driver</db.test.driver>
```
inodb: Disregard formatting change

pom.xml
```diff
@@ -131,9 +169,9 @@
...
     <!-- For MySQL < 5.5 change 'default_storage_engine' to 'storage_engine' -->
-	<db.test.url>jdbc:mysql://127.0.0.1:3306/cgds_test?sessionVariables=default_storage_engine=InnoDB</db.test.url>
+  <db.test.url>jdbc:mysql://127.0.0.1:3306/cgds_test?sessionVariables=default_storage_engine=InnoDB</db.test.url>
```
inodb: Disregard formatting change

PR #1803, opened 2016-10-13
Bugfix (1799) : migration.sql

### LINKED ISSUE #1799: Failed to run mysql migration script
screen shot 2016-10-12 at 11 15 23 am

sheridancbio:
Move line in migration.sql

Fixes #1799 
@n1zea144 @jjgao @kalletlak 

and notify them here:

- move ALTER TABLE line beneath a needed CREATE TABLE line


sheridancbio:
@fedde-s you were completely right. I think when I read your comment originally, I had experienced just what you suggested -- under MyISAM there is no attempt to create the FOREIGN KEY at all, so there is no error.  This means that we should make sure to always test migration.sql under both MyISAM and INNODB engines.

fedde-s:
I am not fully aware of what exactly the current status of the seed database reforms is, but it might be useful to keep that idea in mind in designing it. So far, the seed database explicitly defined (nearly) all tables as MyISAM.

PR #1805, opened 2016-10-14
fixed command to reflect versioned .war artifact

pieterlukasse:
# What? Why?

 .war artifact is now versioned, e.g. cbioportal-1.3.1-SNAPSHOT.war instead of just cbioportal.war

Small fix in documentation. 

@inodb


pieterlukasse:
@n1zea144 could you merge this one?

PR #1806, opened 2016-10-14
fixed and improved way db version mismatch is displayed

### LINKED ISSUE #1790: No banner on incompatible database schema
As documented on [Updating your cBioPortal installation](https://github.com/cBioPortal/cbioportal/blob/master/docs/Updating-your-cBioPortal-installation.md#first-time), along with the migration script we added a banner to notify administrators when their database schema version is incompatible with the deployed version of cBioPortal and needs to be updated. The current release (1.3.0) does not display any warnings when running against the previous version of the database (1.2.1), but instead hangs when opening any view that uses clinical attributes.

pieterlukasse:
# What? Why?

Fixed and improved way db version mismatch is displayed. Fixes #1790

- fix for the validation of the db version. Somehow the previous implementation was failing to detect version mismatches. The new implementation works and just redirects to the error page, simplifying the overall implementation of this check.

Notify 

@fedde-s @zheins


zheins:
Do we always want to redirect to the error page without even attempting the query? @jjgao I think this would cause an issue similar to the one you described in #1422. It might make sense to merge this and set up a different database for the beta portals so we can apply database version updates without affecting prod.

pieterlukasse:
Yes, I think we should want this. See also https://github.com/cBioPortal/cbioportal/pull/1422#issuecomment-240654934

fedde-s:
Also note that in 1.3.0, only the home (query) page works fine with the 1.2.x schema, but most other views (which depend on other API calls and servlets) hang because of exceptions in JavaScript code. That's why this is a critical issue, nothing tells administrators about the problem.

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -274,7 +275,13 @@ protected void doPost(HttpServletRequest httpServletRequest,
             xdebug.logMsg(this, "Got Database Exception:  " + e.getMessage());
             forwardToErrorPage(httpServletRequest, httpServletResponse,
                                DB_CONNECT_ERROR, xdebug);
-        } catch (ProtocolException e) {
+        } catch (DbVersionException e) {
+            String errorMessage = "Mismatch between DB version and portal version. " + e.getMessage();
```
sheridancbio: There are at least three versions involved. The portal release version is displayed at the bottom of the portal footer ... the current public portal reports "cBioPortal Version 1.3.2-SNAPSHOT". This comes from pom.xml.

This is completely independent of the version numbers of the database schemas. To keep this clear, I think we should skew the version numbering now so that the numbers are not so easy to confuse.

There is a version number for the database stored in the 'info' table. For the public portal, the current "database version" is stored in a field called DB_SCHEMA_VERSION. The current value there is "1.3.0".

The warfile embeds properties which set an expected database version number.  The currently running war file for the public tomcat sets "db.version=1.3.0" in portal.properties. This value is compared against the string in the connected database. In this case, the expected database version number is 1.3.0, and the actual datbase version number matches, so there is no error reported. But neither of these match the "portal version".

I recommend we change the property name in portal.properties to "db.expected_schema_version". Some databases are updated with new data nightly or more frequently. We do not increment the version number of the database on data updates ... only schema updates, so I recommend that we talk in terms of db schema versions rather than db versions.

For this PR though, it is sufficient to change this message to not refer to the 'portal version'. maybe "DB Version Exception : " + e.getMessage();

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -239,9 +239,10 @@ protected void doPost(HttpServletRequest httpServletRequest,
             {
             	String extraMessage = "";
             	//extra message for the cases where property is missing (will happen often in transition period to this new versioning model):
-            	if (dbPortalVersion.equals("0"))
+            	if (dbPortalVersion.equals("0")) {
             		extraMessage = "The db.version property also not found in your portal.properties file. This new property needs to be added by the administrator.";
-                httpServletRequest.setAttribute(DB_ERROR, "Current DB Version: " + dbVersion + "<br/>" + "DB version expected by Portal: " + dbPortalVersion + "<br/>" + extraMessage);
+            	}
+                throw new DbVersionException("Current DB Version: " + dbVersion + "<br/>" + "DB version expected by Portal: " + dbPortalVersion + "<br/>" + extraMessage);
```
sheridancbio: It seems a little odd to throw and catch an exception within a single function call, but I suppose it is better than a "goto" style statement. It seems like the block of catch statements below create a priority order however .. where only the first applicable catch phrase will be executed, and other (possibly detectable) conditions are not reported. Since that seems to be the case, it would be equivalent to just inline the catch block in place of the throw statement. Then a new custom exception class would not be needed.

sheridancbio:
I spoke just now with @jjgao about this issue. He pointed out that here we have a nightly update process running on our database, and we deploy both a main and a beta warfille which are widely used. Both the main and the beta warfile are backed by a single database which receives the frequent updates.

I pointed out that in time we will want to make a change to the database which is *actually* an incompatible change to schema and we will not be able to use the same database for two different warfiles with different expectations. When times like that come we will no longer be able to run one database supplying data to two different warfiles (master / rc). We will need a solution then, but JJ hopes those times are infrequent.

For now, he wants us to support the current configuration ... he would be in favor of an approach where there is a properties setting which overrides the "normal" exception detection. Something like "db.suppress_schema_version_mismatch_errors=true". Setting this property would perhaps log the detection of a schema version mismatch, but would not display warnings to the user or throw exceptions during the processing of requests. When this setting is false [the default] then the generation of exceptions and the display of error messages would happen as suggested in this PR.

Personally, I think that if we are going to throw exceptions for QueryBuilder, we probably should throw exceptions for all functions which interact with the database, such as the WebAPI endpoints, the Data Sets page, Patient View, etc...  to make the site non-functional (except for displaying the error message and perhaps hosting of static content)

jjgao:
@sheridancbio Thanks for the summary.

@pieterlukasse maybe we don't have to throw the exception in the backend as long as we display the message on the web page. Some configuration that allows disable the error message would be helpful as @sheridancbio pointed out.

jjgao:
@pieterlukasse would you like to make the suggested change to this pr?

sheridancbio:
Also, we have recently made the full transition off of the "bridge" database schema we were using to support both master and rc branches at once. We are back to using a single database, but I think we will have this need again in the future [to use the same database for two different branches of the code ... [master &  rc (or release) branch] which expect different version numbers but can both function with the same database. A "suppress error" configuration flag will definitely be useful.

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
pieterlukasse: I actually like the fact that all `forward...()`  logic is placed together. I think the exception should normally be thrown as an exception. If the code that handles the exception is placed there instead, it will make the code harder to read (in my opinion) since you would now suddenly be mixing up the normal workflow of the method with the exception handling logic.

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
pieterlukasse: The version number has been updated. I also improved the messages, adding the word "schema" to make it more clear. I did not update the property name itself though.

pieterlukasse:
I made the behaviour configurable now as @sheridancbio suggested. If the property `db.suppress_schema_version_mismatch_errors` is not set , or not set to `true`, then the banner will appear in case of mismatch:

image

If  `db.suppress_schema_version_mismatch_errors = true`, then the portal will continue:

image

If a query error is found (`DaoException` - nb: this will not always happen as much of the data is also cached in the Dao layer), then the banner will display the error. In this case, the DB version is also checked and, if different, an extra warning message is appended to the normal error message:

image

In all cases (both with `db.suppress_schema_version_mismatch_errors` or not) the mismatch is reported in the log files as a `WARNING`:

image

fedde-s:
Nice, an update for this issue! Great idea to also inform the person who starts the server at least via the logs: that is the primarily person who should be aware of the potentially broken setup, more so than other users looking at the web pages afterwards.

sheridancbio:
Looks great. Thanks for setting up this configuration option for us .. I'll add the option into our configuration properties files.

sheridancbio:
Oh no; cannot merge yet. Could you rebase to the head of hotfix?
(and maybe squash commits into a single while doing the rebase?)

PR #1819, opened 2016-10-17
Include fixes from iViz milestone 1.0.3

zhx828:


zhx828:
Need to improve cbioportal-client.js performance or bypass the keyCombinations function

zhx828:
We chose to bypass cbioportal-client clinical data call by calling API directly. This is a temporary fix.

PR #1825, opened 2016-10-19
Implemented 5 endpoints in Cancer Type API and Genetic Profile API

ersinciftci:
Implemented 5 endpoints in Cancer Type API and Genetic Profile API with persistence, service and web tests

GET /cancer-types
GET /cancer-types/{cancerTypeId}
GET /genetic-profiles
GET /genetic-profiles/{geneticProfileId}
GET /studies/{studyId}/genetic-profiles

Also added sampleCount to the Study APIs `DETAILED` projection.

### DIFF
--- a/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/CancerTypeMyBatisRepositoryTest.java
+++ b/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/CancerTypeMyBatisRepositoryTest.java
@@ -1,0 +1,111 @@
+package org.cbioportal.persistence.mybatis;
+
+import org.cbioportal.model.TypeOfCancer;
+import org.cbioportal.model.meta.BaseMeta;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Configurable;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration("/testContextDatabase.xml")
+@Configurable
+public class CancerTypeMyBatisRepositoryTest {
+
+    @Autowired
+    private CancerTypeMyBatisRepository cancerTypeMyBatisRepository;
+
+    @Test
+    public void getAllCancerTypesIdProjection() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("ID", null, null, null, null);
+
+        Assert.assertEquals(2, result.size());
+        TypeOfCancer typeOfCancer = result.get(0);
+        Assert.assertEquals("acc", typeOfCancer.getTypeOfCancerId());
+    }
+
+    @Test
+    public void getAllCancerTypesSummaryProjection() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("SUMMARY", null, null, null, null);
+
+        Assert.assertEquals(2, result.size());
+        TypeOfCancer typeOfCancer = result.get(0);
+        Assert.assertEquals("brca", typeOfCancer.getTypeOfCancerId());
+        Assert.assertEquals("Breast Invasive Carcinoma", typeOfCancer.getName());
+        Assert.assertEquals("breast,breast invasive", typeOfCancer.getClinicalTrialKeywords());
+        Assert.assertEquals("HotPink", typeOfCancer.getDedicatedColor());
+        Assert.assertEquals("Breast", typeOfCancer.getShortName());
+        Assert.assertEquals("tissue", typeOfCancer.getParent());
+    }
+
+    @Test
+    public void getAllCancerTypesDetailedProjection() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("DETAILED", null, null, null, null);
+
+        Assert.assertEquals(2, result.size());
+        TypeOfCancer typeOfCancer = result.get(0);
+        Assert.assertEquals("brca", typeOfCancer.getTypeOfCancerId());
+        Assert.assertEquals("Breast Invasive Carcinoma", typeOfCancer.getName());
+        Assert.assertEquals("breast,breast invasive", typeOfCancer.getClinicalTrialKeywords());
+        Assert.assertEquals("HotPink", typeOfCancer.getDedicatedColor());
+        Assert.assertEquals("Breast", typeOfCancer.getShortName());
+        Assert.assertEquals("tissue", typeOfCancer.getParent());
+    }
+
+    @Test
+    public void getAllCancerTypesSummaryProjection1PageSize() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("SUMMARY", 1, 0, null, null);
+
+        Assert.assertEquals(1, result.size());
+    }
+
+    @Test
+    public void getAllCancerTypesSummaryProjectionNameSort() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("SUMMARY", null, null, "name", "ASC");
+
+        Assert.assertEquals(2, result.size());
+        Assert.assertEquals("Adrenocortical Carcinoma", result.get(0).getName());
+        Assert.assertEquals("Breast Invasive Carcinoma", result.get(1).getName());
+    }
+
+    @Test
+    public void getMetaCancerTypes() throws Exception {
+
+        BaseMeta result = cancerTypeMyBatisRepository.getMetaCancerTypes();
+
+        Assert.assertEquals((Integer)2, result.getTotalCount());
+    }
+
+    @Test
+    public void getCancerTypeNullResult() throws Exception {
+
+        TypeOfCancer result = cancerTypeMyBatisRepository.getCancerType("invalid_cancer_type");
+
+        Assert.assertNull(result);
+    }
+
+    @Test
+    public void getCancerType() throws Exception {
+
+        TypeOfCancer result = cancerTypeMyBatisRepository.getCancerType("acc");
+
+        Assert.assertEquals("acc", result.getTypeOfCancerId());
+        Assert.assertEquals("Adrenocortical Carcinoma", result.getName());
+        Assert.assertEquals("adrenocortical carcinoma", result.getClinicalTrialKeywords());
+        Assert.assertEquals("Purple", result.getDedicatedColor());
+        Assert.assertEquals("ACC", result.getShortName());
+        Assert.assertEquals("adrenal_gland", result.getParent());
+    }
+}
--- a/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
+++ b/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
@@ -1,0 +1,196 @@
+package org.cbioportal.persistence.mybatis;
+
+import org.cbioportal.model.CancerStudy;
+import org.cbioportal.model.GeneticProfile;
+import org.cbioportal.model.meta.BaseMeta;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Configurable;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration("/testContextDatabase.xml")
+@Configurable
+public class GeneticProfileMyBatisRepositoryTest {
+
+    @Autowired
+    private GeneticProfileMyBatisRepository geneticProfileMyBatisRepository;
+
+    @Test
+    public void getAllGeneticProfilesIdProjection() throws Exception {
+
+        List<GeneticProfile> result = geneticProfileMyBatisRepository.getAllGeneticProfiles("ID", null, null, null,
+                null);
+
+        Assert.assertEquals(6, result.size());
+        GeneticProfile geneticProfile = result.get(0);
+        Assert.assertEquals((Integer) 2, geneticProfile.getGeneticProfileId());
+        Assert.assertEquals("study_tcga_pub_gistic", geneticProfile.getStableId());
+        Assert.assertNull(geneticProfile.getCancerStudy());
+    }
+
+    @Test
+    public void getAllGeneticProfilesSummaryProjection() throws Exception {
+
+        List<GeneticProfile> result = geneticProfileMyBatisRepository.getAllGeneticProfiles("SUMMARY", null, null, null,
+                null);
+
+        Assert.assertEquals(6, result.size());
+        GeneticProfile geneticProfile = result.get(0);
+        Assert.assertEquals((Integer) 2, geneticProfile.getGeneticProfileId());
+        Assert.assertEquals("study_tcga_pub_gistic", geneticProfile.getStableId());
+        Assert.assertEquals((Integer) 1, geneticProfile.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", geneticProfile.getCancerStudyIdentifier());
+        Assert.assertEquals("COPY_NUMBER_ALTERATION", geneticProfile.getGeneticAlterationType());
+        Assert.assertEquals("DISCRETE", geneticProfile.getDatatype());
+        Assert.assertEquals("Putative copy-number alterations from GISTIC", geneticProfile.getName());
+        Assert.assertEquals("Putative copy-number from GISTIC 2.0. Values: -2 = homozygous deletion; -1 = hemizygous " +
+                "deletion; 0 = neutral / no change; 1 = gain; 2 = high level amplification.",
+                geneticProfile.getDescription());
+        Assert.assertEquals(true, geneticProfile.getShowProfileInAnalysisTab());
+        Assert.assertNull(geneticProfile.getCancerStudy());
+    }
+
+    @Test
+    public void getAllGeneticProfilesDetailedProjection() throws Exception {
+
+        List<GeneticProfile> result = geneticProfileMyBatisRepository.getAllGeneticProfiles("DETAILED", null, null,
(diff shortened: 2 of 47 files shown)


sheridancbio:
Reviewing this PR now

web/src/main/java/org/cbioportal/web/CancerTypeController.java
```diff
@@ -18,19 +28,42 @@
...
+    @Autowired
+    private CancerTypeService cancerTypeService;
+
+    @RequestMapping(value = "/cancer-types", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
     @ApiOperation("Get all cancer types")
-    public ResponseEntity<List<TypeOfCancer>> getAllCancerTypes(@RequestParam(defaultValue = "SUMMARY") Projection projection,
-                                                                @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_SIZE) Integer pageSize,
-                                                                @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_NUMBER) Integer pageNumber) {
+    public ResponseEntity<List<TypeOfCancer>> getAllCancerTypes(
+            @ApiParam("Level of detail of the response")
```
sheridancbio: Hopefully swagger UI will show the choices for Projection enum

web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
```diff
@@ -1,8 +1,6 @@
 package org.cbioportal.web.error;
 
-import org.cbioportal.service.exception.PatientNotFoundException;
-import org.cbioportal.service.exception.SampleNotFoundException;
-import org.cbioportal.service.exception.StudyNotFoundException;
+import org.cbioportal.service.exception.*;
```
sheridancbio: this change is against google java style ... although I understand the preference. I think our style should allow wildcards for standard java packages like java.util.*, but for our own packages I think explicit listings are better.
We can discuss this as part of the development guidelines/practices document.

web/src/test/java/org/cbioportal/web/StudyControllerTest.java
```diff
@@ -35,35 +35,36 @@
...
-    public static final String TEST_CITATION_2 = "test_citation_2";
-    public static final String TEST_GROUPS_2 = "test_groups_2";
-    public static final int TEST_STATUS_2 = 0;
-    public static final String TEST_DATE_2 = "2013-10-12 11:11:15";
-    public static final String TEST_TYPE_OF_CANCER_NAME = "test_type_of_cancer_name";
-    public static final String TEST_CLINICAL_TRIAL_KEYWORDS = "test_clinical_trial_keywords";
-    public static final String TEST_DEDICATED_COLOR = "test_dedicated_color";
-    public static final String TEST_TYPE_OF_CANCER_SHORT_NAME = "test_type_of_cancer_short_name";
-    public static final String TEST_PARENT = "test_parent";
+    private static final int TEST_CANCER_STUDY_ID_1 = 1;
```
sheridancbio: This is proper ... to prevent the development of unwanted dependencies from other test classes

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/CancerTypeMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Test
+    public void getAllCancerTypesDetailedProjection() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("DETAILED", null, null, null, null);
+
+        Assert.assertEquals(2, result.size());
+        TypeOfCancer typeOfCancer = result.get(0);
+        Assert.assertEquals("brca", typeOfCancer.getTypeOfCancerId());
```
sheridancbio: I see .. there is no summary class -- just one representation for both levels.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -91,7 +91,7 @@
             INNER JOIN type_of_cancer ON cancer_study.TYPE_OF_CANCER_ID = type_of_cancer.TYPE_OF_CANCER_ID
         </if>
         <include refid="whereSample"/>
-        <if test="sortBy != null">
+        <if test="sortBy != null and projection != 'ID'">
```
sheridancbio: Maybe an alternative approach is to set the sort by field to the id field when the projection is ID.

sheridancbio:
Deployed locally and hand tested. Looks good.

ersinciftci:
I'll respond to comments and make improvements probably next week.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/CancerTypeMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Autowired
+    private CancerTypeMyBatisRepository cancerTypeMyBatisRepository;
+
+    @Test
+    public void getAllCancerTypesIdProjection() throws Exception {
+
+        List<TypeOfCancer> result = cancerTypeMyBatisRepository.getAllCancerTypes("ID", null, null, null, null);
+
+        Assert.assertEquals(2, result.size());
+        TypeOfCancer typeOfCancer = result.get(0);
```
pieterlukasse: I think it is not so minor. Small issues like this caused quite some problems in the past when we were trying to get all tests working...

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+        List<GeneticProfile> result = geneticProfileMyBatisRepository.getAllGeneticProfiles("SUMMARY", 1, 0, null, null);
+
+        Assert.assertEquals(1, result.size());
+    }
+
+    @Test
+    public void getAllGeneticProfilesSummaryProjectionStableIdSort() throws Exception {
+
+        List<GeneticProfile> result = geneticProfileMyBatisRepository.getAllGeneticProfiles("SUMMARY", null, null,
```
pieterlukasse: I think we can (and should) avoid a "global" test database as much as possible. In my opinion each test should take care of its own required data and add it via the setup method. And having each test running in rollback mode (which is already the case) will avoid them from interfering with each other.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+        Assert.assertNull(result);
+    }
+
+    @Test
+    public void getGeneticProfile() throws Exception {
+
+        GeneticProfile geneticProfile = geneticProfileMyBatisRepository.getGeneticProfile("study_tcga_pub_gistic");
+
+        Assert.assertEquals((Integer) 2, geneticProfile.getGeneticProfileId());
```
pieterlukasse: why should we suppress this? Doesn't the frontend also need to know the id so that it can call related web-services that need it to find the correct genetic profile data?

pieterlukasse:
I added some small comments as well. Good review by @sheridancbio ! 

As a side note: I need a few extra endpoints implemented for RFC 28 (https://github.com/cBioPortal/cbioportal/projects). Shall I start on this? These are the ones I need:

Genetic Data :
- GET /genetic-profiles/{geneticProfileId}/genetic-data : Get all genetic data in a genetic profile

Genetic Profiles :
- GET /genetic-profiles/{geneticProfileId} : Get genetic profile
- GET /studies/{studyId}/genetic-profiles : Get all genetic profiles in a study

pieterlukasse:
Another question is: for RFC 28 we will also need to add data to the database (write some data loader code). Are we going to reuse part of this code in the https://github.com/cBioPortal/importer module? Maybe @n1zea144 and @angelicaochoa have some ideas about this?

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
sheridancbio: The internal id for a profile (the auto-increment integer value) is not stable .. external users should be using stable_ids which should remain valid across different releases of the portal (unless there is a very good reason why using the internal id gives a major benefit)

sheridancbio:
About @pieterlukasse 's question about service layer calls that are "pass-through" calls to the persistence layer:
We should maybe have a bigger conversation about this. I think that more often than not we have been delegating the "work" of joining tables, filtering, sorting, etc... to the Persistence layer through clever mysql constructions and mybatis. There are some cases where processing is written into the service layer, but it is the minority. So should we have a service layer call which merely delegates to the persistence layer? I think there are pros/cons to having or not having a service layer pass-through function. It is not strictly necessary, so it can be considered code bloat. On the other hand, it sets a pattern which is followed by other tech stack development efforts ... so it makes us aware that there *is* a service layer in which domain knowledge/transformations/logic can live. We don't need to locate all data manipulation code in complex mysql constructions. In cases where domain logic is implemented in the service layer, we have a natural place for expanding the unit tests (in a "pure" programming space which has no worries about data source representations (tables) or output formatting (json)). I think we need to think more about how to make better use of the service layer. Maybe we should be taking more of the load off of the mysql server where appropriate and moving (where appropriate) expensive table joining or selection (such as the cases where we pass in a big comma separated list of identifiers which become part of an "fieldname in (,,,)" clause) to the service layer instead (perhaps making multiple simpler calls which will each be binary-tree searchable and then aggregating results). But in general, I think it is good to establish that we have a service layer, and that the presentation code can more insulated from the persistence interface. We have the flexibility to decouple the API calls from the services, and from the Repository interfaces. I think we should try to use that flexibility more. Also, each service does not have to be a "do the work of constructing the API output" general function. There can be layers of service calls (one service calling another service function) ... and we have a little of that with a few utilities. But we could have more substantial data transformation functions which are reused too.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/CancerTypeMyBatisRepositoryTest.java
ersinciftci: Since we are using an in-memory H2 database for these tests, not a 3rd party MySql that can have different versions or different storage engines, the tests should be deterministic. So yes you can't guarantee a particular order, it's up to the database implementation, but you won't get a random order every time you call it, it will be the same every time if the inserted data is the same. And when we change the inserted data, we need to change some tests in any case.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,196 @@
...
+        Assert.assertEquals("TCGA, Nature 2012", cancerStudy.getCitation());
+        Assert.assertEquals("SU2C-PI3K;PUBLIC;GDAC", cancerStudy.getGroups());
+        Assert.assertEquals((Integer)0 , cancerStudy.getStatus());
+    }
+
+    @Test
+    public void getAllGeneticProfilesSummaryProjection1PageSize() throws Exception {
+
+        List<GeneticProfile> result = geneticProfileMyBatisRepository.getAllGeneticProfiles("SUMMARY", 1, 0, null, null);
+
```
ersinciftci: I just prefer leaving a blank line before starting assertions, because they are a different logical group than the data preparation part. And I also like having a blank line after the method definition. Since this method is short, looks like a lot of blank lines, but just tried to be consistent with other methods.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
ersinciftci: Since these tests are integration tests, not unit tests (because we connect to a database even though it's in-memory), I prefer testing as a whole, interfering with each other, just like in production. And we won't have delete/update statements in the new stack anyway, so it's easier to achieve.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
ersinciftci: Yes, we are suppressing using mixins.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepository.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public GeneticProfile getGeneticProfile(String geneticProfileId) {
+        return geneticProfileMapper.getGeneticProfile(geneticProfileId, Constants.DETAILED_PROJECTION);
+    }
+
+    @Override
+    public List<GeneticProfile> getAllGeneticProfilesInStudy(String studyId, String projection, Integer pageSize,
+                                                             Integer pageNumber, String sortBy, String direction) {
+
+        return geneticProfileMapper.getAllGeneticProfiles(studyId, projection, pageSize,
```
ersinciftci: I prefer having more verbose methods in persistence-api interfaces, since they are planned to be used by 3rd parties, but we don't need that verbosity level with Mapper methods, since they are internal. So I merge the different methods in the same method in Mapper, if you provide null value to property, I ignore that property in Mapper XML file, so we have more flexible queries and less boilerplate code. But I agree that the name of the method shouldn't be `getAllGeneticProfiles` in the Mapper, it should be `getGeneticProfiles`, and if we used Java 8, we could use https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html for studyId profile, so it would be more clear that it's an optional parameter. We can change it when we switch to Java 8.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/tool/Constants.java
```diff
@@ -0,0 +1,6 @@
+package org.cbioportal.persistence.mybatis.tool;
+
+public class Constants {
```
ersinciftci: Right, I'll change it to PersistenceConstants and will move it to persistence-api.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
ersinciftci: Right, I'll set the sort by field of the id field.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneticProfileMapper.xml
```diff
@@ -0,0 +1,65 @@
...
+    </sql>
+
+    <select id="getAllGeneticProfiles" resultType="org.cbioportal.model.GeneticProfile">
+        SELECT
+        <include refid="select">
+            <property name="prefix" value=""/>
+        </include>
+        FROM genetic_profile
+        INNER JOIN cancer_study ON genetic_profile.CANCER_STUDY_ID = cancer_study.CANCER_STUDY_ID
+        <if test="studyId != null">
```
ersinciftci: We either do this or will have duplicate queries for the cases we get by cancerId and we get all of them. I preferred this because I think those parameters as criteria, and if a criteria is null, it means don't apply that criteria, which means return everything.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneticProfileMapper.xml
```diff
@@ -0,0 +1,65 @@
...
+        </if>
+    </select>
+
+    <select id="getMetaGeneticProfiles" resultType="org.cbioportal.model.meta.BaseMeta">
+        SELECT
+        COUNT(*) AS totalCount
+        FROM genetic_profile
+        <if test="studyId != null">
+            INNER JOIN cancer_study ON genetic_profile.CANCER_STUDY_ID = cancer_study.CANCER_STUDY_ID
+            WHERE cancer_study.CANCER_STUDY_IDENTIFIER = #{studyId}
```
ersinciftci: We don't get 0, we get the count of all profiles, because the SQL block in if doesn't apply if it's null, which means the count of all profiles.

service/src/main/java/org/cbioportal/service/GeneticProfileService.java
```diff
@@ -0,0 +1,23 @@
...
+import org.cbioportal.model.meta.BaseMeta;
+import org.cbioportal.service.exception.GeneticProfileNotFoundException;
+
+import java.util.List;
+
+public interface GeneticProfileService {
+
+    List<GeneticProfile> getAllGeneticProfiles(String projection, Integer pageSize, Integer pageNumber, String sortBy,
+                                               String direction);
+
```
ersinciftci: Yes, deleted.

web/src/main/java/org/cbioportal/web/CancerTypeController.java
ersinciftci: Yes it does.

web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
ersinciftci: Right, missed it, will fix in next PR.

ersinciftci:
@sheridancbio replied the comments. I didn't get what you mean by `why are we not using persistence-api module` though.

sheridancbio:
I think it was a reference to the PersistenceConstants class location .. which you have already moved.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/CancerTypeMyBatisRepositoryTest.java
pieterlukasse: I think that is too much implicit information. In this case, I would like to recommend adding a comment line in the code explaining this or implementing something like a sorting to make things explicit. Otherwise the next developer that looks at this code would have to figure this out all over again when in doubt...

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneticProfileMyBatisRepositoryTest.java
pieterlukasse: @sheridancbio are you saying we should not have `geneticProfile.getGeneticProfileId()` available?

Issue #1788, opened 2016-10-10
Loading "Mutation count" is slow


jjgao:
Loading "Mutation count" seems to be slow. We should try to improve the performance. 

![image](https://cloud.githubusercontent.com/assets/840895/19253114/de667f62-8f16-11e6-8926-020f94184e6c.png)

adamabeshouse:
Separate call for missense by position equality, then can use group by and count to do all in single query

adamabeshouse:
https://github.com/cBioPortal/cbioportal/pull/1842

PR #1842, opened 2016-10-24
Optimize mutation count fetching

adamabeshouse:
# What? Why?

Fix https://github.com/cBioPortal/cbioportal/issues/1788

Implemented new (legacy) API endpoint for getting mutation counts from single positions,
which is faster to do than with position inequalities.

Use this in data manager instead of old mutation count endpoint.
# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.
# Notify reviewers

@cBioPortal/frontend
@jjgao

### DIFF
--- a/business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
+++ b/business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
@@ -48,6 +48,7 @@ import org.springframework.transaction.annotation.Transactional;
 import org.cbioportal.service.CosmicCountService;
 import org.cbioportal.model.MutationSignature;
 import org.cbioportal.model.CosmicCount;
+import org.cbioportal.persistence.dto.PositionMutationCount;
 
 /**
  *
@@ -137,6 +138,11 @@ public class ApiService {
 		}
 		// TODO: implement other contexts
 		return signatures;
+	}
+
+	@Transactional
+	public List<PositionMutationCount> getPositionMutationCounts(Map<String, List<Integer>> hugoGeneSymbolToPositions) {
+		return mutationService.getPositionMutationCounts(hugoGeneSymbolToPositions);
 	}
 
         @Transactional
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/MutationRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/MutationRepository.java
@@ -8,6 +8,8 @@ import org.cbioportal.persistence.dto.MutatedGeneSampleCount;
 import org.cbioportal.persistence.dto.SignificantlyMutatedGene;
 
 import java.util.List;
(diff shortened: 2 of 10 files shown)


sheridancbio:
I looked over this quickly. New mutation data functions are added and connected in persistence-mybatis and 
service layers for counting mutations at a number of positions within a number of genes within a number of studies. The web layer presentation is through the older ApiController (expanding it).

I wonder what @ersinciftci thinks about accepting "new tech stack" functions to serve data requests to legacy / not yet REST style designed API endpoints. This has already be done for GenePanelRepository, for instance, serving requests to /api-legacy/genepanel

More detailed review is needed, but overall my first impression is that the PR looks good.

sheridancbio:
One thing where I wonder if there is a better way to do this, is the passing of maps between gene symbol strings list of integer offsets. This seems like an overly crafted argument type. Perhaps we should consider multiple calls rather than bundling a "multi-request" into one hit to the endpoint. (Just thinking out loud)

adamabeshouse:
@sheridancbio I hear you but this way we save on the cost of making multiple ajax requests. And is it that contrived? It seems to be something people do and it sort of makes sense to me. 

Another option would be to pass two arguments: list of genes, and list of lists of integers.

jjgao:
@sheridancbio @adamabeshouse do we have any further comments? What's the next action on this pull request?

sheridancbio:
I've been focussing on other PR's. I was also hoping to get input from @ersinciftci regarding the overall approach here. Lastly, I don't understand who the client for this is ... is this mainly a performance increase for the DataManager?

But if this is needed quickly, I would say that my personal belief is that we should stop adding endpoints to ApiController; so let's move the endpoint out of ApiController .. set up a new endpoint in api-legacy (maybe something like /api-legacy/mutation-count-by-gene-positions)
@ersinciftci .. any comment about naming?

The other issue here is that the return type is a class from the persistence layer ... I would move that to the model module, but this is not like other model types and represents more of a list packaging than a natural data relationship. So this model should probably be distinguished somehow (separate package within model? class name with "legacy" not sure what is right). But this also brings me back to thinking that we are not being smart trying to use the same model for database interactions and web server API endpoint interactions. Should we have separate models? Or should we allow additional model classes for data transfer out of the Api endpoints? Either way, I think it is "wrong" for us to be using a persistence level class to specify the return type from the web layer. (this is a bigger conversation)

sheridancbio:
Also, this introduces a new endpoint .. should this be in hotfix? it seems like a new feature.

sheridancbio:
Another need is to add unit tests to the Web layer .. and possibly other layers, similar to the testing done in the mutation-signature endpoint work.

adamabeshouse:
Taking @sheridancbio 's comments into account, I've reimplemented this branch off of `rc`:https://github.com/cBioPortal/cbioportal/pull/1923

PR #1844, opened 2016-10-26
Pull cbioportal-frontend code from jitpack

inodb:
Add jitpack.io to install cbioportal-frontend repo from github. Replace
clinical information tab as proof of concept.
- use jitpack.io to download cbioportal-frontend repo jar
- unpack & copy reactapp folder to war
- add script tag to patient_view jsp
- To facilitate development, one can use http://localhost:3000 in the script
  tag instead if localdev is set to true in localStorage.
- Similarly, to use a heroku instance for serving scripts and css, set the
  subdomain name in `localStorage['heroku']` e.g. `cbioportal-frontend`

Example (click on clinical information tab):

http://cbioportal-pr-1844.herokuapp.com/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04

## Screenshot
From:
screen shot 2016-11-03 at 8 37 31 pm

To:
screen shot 2016-11-03 at 8 36 43 pm

As can be seen in the screenshots, some functionality such as copy/csv/search are lost. I don't think this is a big deal, since you can still just select the columns and copy (this is why HTML tables are great :)). Eventually once we settle on a better React table, we can swap this one out for another.


portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
```diff
@@ -1155,5 +1155,36 @@ window["<%=PatientView.CANCER_STUDY_META_DATA_KEY_STRING%>"]
...
+    document.write('<script src="http://localhost:3000/reactapp/js/common.bundle.js"></scr'+'ipt>');
+    document.write('<script src="http://localhost:3000/reactapp/js/main.app.js"></scr'+'ipt>');
+    // Show alert
+    document.write('<div style="position: fixed; top: 0; left: 0; width: 100%;">'+
+                       '<div class="alert alert-warning">' +
+                            '<button type="button" class="close" data-dismiss="alert">&times;</button>' +
+                            'cbioportal-frontend dev mode' +
+                       '</div>' +
+                   '</div>');
+} else {
```
inodb: @alisman Seems like https://github.com/cBioPortal/cbioportal/pull/1846 has been merged, so I think we should just go ahead and show the replaced clinical information tab. Even though there is slight regression in terms of user features I think it's a good idea to get something replaced in the next release. Just so we can get used to the process of adding new features from the frontend repo to the parent project.

alisman:
I just noticed a problem with csv download format.  It's not same as old one.  I will fix.

PR #1851, opened 2016-10-27
Fixes

yichaoS:
- fixes based on feedback
- add maven plugin for installing npm and node
- removed ghost "specRunner.html"


portal/src/main/webapp/js/src/cbio-stat.js
```diff
@@ -1,14 +1,12 @@
-if (cbio === undefined) {
```
zhx828: We shouldn't remove initialization here. I notice you declare the cbio in cbio-stat.spec.js but we will not include cbio-stat.spec.js in the final js build file, right?

portal/src/main/webapp/js/SpecRunner.html
```diff
@@ -1,99 +0,0 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
```
zhx828: Is it no longer needed?

yichaoS: No. This is an independent page showing some old test results based on jasmine but even those tests are not working anymore. Here is the screenshot. 
image

portal/src/main/webapp/js/src/cbio-stat.js
yichaoS: No it's not. Added it back.

PR #1852, opened 2016-10-28
Validator: fix requirement for OS_MONTHS

sandertan:
Remove the requirement: `OS_MONTHS has to be specified when OS_STATUS is DECEASED` in validateData.py

The requirement was added in 1.3.1 in commit aa9a48cb1d5601a840267aabc86f5bc6482edd5a . It was added because @aderidder noticed in early 2016 that the front end could not handle a missing value for OS_MONTHS when OS_STATUS was deceased, which caused issues with the patient information page.

We cannot reproduce this error anymore, but this requirement does lead to failed verifications of studies that are presumed to be good:

TCGA studies from datahub (such as lgg_tcga) that could be loaded prior to 1.3.1, now fail to load, because studies can contain [Not Available] in the OS_MONTHS column of `data_bcr_clinical_data_patient.txt`. We loaded this data in prior versions (1.2.5) of cBioPortal, and did not find any issues with the patient information page. 

We also loaded time-line data which contained [Not Available] in OS_MONTHS for a patient and this also seems to work fine. The timeline cannot show months in this case, but everything else on the page functions correctly.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1798,15 +1798,6 @@ class PatientClinicalValidator(ClinicalValidator):
             elif col_name == 'OS_MONTHS':
                 osmonths_value = value
 
-        if osstatus_is_deceased and (
-                    osmonths_value is None or
-                    osmonths_value.lower() in self.NULL_VALUES):
-            if osmonths_value is None or osmonths_value == '':
-                osmonths_value = '<none>'
-            self.logger.error(
-                "OS_STATUS is 'DECEASED', but OS_MONTHS is not specified",
-                extra={'line_number': self.line_number,
-                       'cause': osmonths_value})
 
     def onComplete(self):
         """Perform final validations based on the data parsed."""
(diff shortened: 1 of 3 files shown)


fedde-s:
A relevant unit test is failing. @sandertan, could you check it?

fedde-s:
Otherwise, looks good! Did you test whether it breaks the survival curve for all the patients in the study that do have the attribute? If it does, I think it should be a warning.

fedde-s:
...or a bug report for the frontend people.

sandertan:
I checked the survival curve with @pulyakhina and it still works!

pieterlukasse:
I'm suspecting this line might still fail: https://github.com/cBioPortal/cbioportal/blob/master/portal/src/main/webapp/WEB-INF/jsp/patient_view/clinical_timeline.jsp#L164

pieterlukasse:
We debugged `clinical_timeline.jsp` and this is not an issue. The whole clinical attribute is skipped because it has the correct '[Not Available]` flag which is now (since not too long ago) correctly parsed by `ImportClinicalData.java`  :thumbsup:

pieterlukasse:
@jjgao this is ready to merge :+1:

PR #1855, opened 2016-10-31
Datasets API

JiaoJiao123:
# What? Why?
This API is designed for the [cBioPortal DataSets page](http://www.cbioportal.org/data_sets.jsp).

The current dataset page is served in the jsp file. We refactored this page under the cBioPortal front end structure. In the new file, we want to retrieve datasets info through API directly instead of through java code, and that's why we are developing this API.

- get public datasets info
- corresponding unit test

# Notify reviewers
@cBioPortal/backend

### DIFF
--- a/model/src/main/java/org/cbioportal/model/DataSet.java
+++ b/model/src/main/java/org/cbioportal/model/DataSet.java
@@ -1,0 +1,69 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.cbioportal.model;
+import java.io.Serializable;
+/**
+ *
+ * @author jiaojiao
+ */
+public class DataSet implements Serializable{
+    private String name;
+    private String citation;
+    private String cancer_study_identifier;
+    private String PMID;
+    private String stable_id;
+    private int count;
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public String getCitation() {
+        return citation;
+    }
+
+    public void setCitation(String citation) {
+        this.citation = citation;
+    }
+
+    public String getCancer_study_identifier() {
+        return cancer_study_identifier;
+    }
+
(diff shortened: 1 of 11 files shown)


ersinciftci:
Isn't list of datasets actually list of cancer studies with a little bit more information? If that's the case, I think it's a good candidate for Study API's DETAILED projection instead of creating a new API.

ecerami:
I agree with @ersinciftci's assessment.  We also discussed with @sheridancbio, @jjgao and Ben at last week's #mini-moonshot call.  Can we not proceed with this pull-request, and roll into the existing API instead?

jjgao:
@JiaoJiao123 let's define an API that can accommodate both the Data Sets page and the query page. could you work with @ersinciftci, @sheridancbio and @andy?

ersinciftci:
@JiaoJiao123 Since query page needs less data, it should suffice to call `SUMMARY` projection, and `DETAILED` projection should return the extra fields for data sets page. In order to do that, you need to move `sampleCount` property to `SUMMARY` projection (it's currently in `DETAILED`), and add the data sets related extra properties and joins to the `DETAILED` projection. You can always ping me on slack if you have questions/need help.

ersinciftci:
@jjgao @sheridancbio and in the long term we can create columns for all those values in cancer_study table and precompute the data during import, so no calculation and join needed while retrieving data.

ersinciftci:
Actually I just realized it won't make much difference since the result will be cached after the first call anyway.

n1zea144:
On a related note, we have plans to do away with case lists altogether.  The case list and its corresponding size can be computed/cached at runtime.

JiaoJiao123:
@ersinciftci @jjgao I didn't know the new Study API end point has already been implemented. I just checked it out and it is really nice! I really like the projection option. I would look into it more and see how to add the extra info for datasets page, will keep you updated.

sheridancbio:
I'm also sorry I did not communicate with you yesterday. I'm glad we think we can accommodate the needs of both pages with the single endpoint. I hope I can help with the adjustments to the study projections.

PR #1865, opened 2016-11-02
Update warning to fix unit-test

sandertan:
# What? Why?
PR1729 included a last minute change to the warning, which caused the unit-test to fail. This PR modifies the warning message so that the unit-test succeeds, and also moved the warning message from `cause`-field to `message`-field.


sandertan:
Adding this to PR1861

PR #1809, opened 2016-10-14
Add validation for clinical attributes with set meanings

fedde-s:
This brings the commits from pull request #1567 to the hotfix branch for inclusion in release v1.3.1.

PR #1808, opened 2016-10-14
Remove database consistency validation of clinical attributes

fedde-s:
Clinical attributes used to share a single namespace across all studies, meaning that once the meaning of an attribute name was defined (with a display name, mouseover description, data type, sample or patient level and priority), all subsequent studies would have to adhere to that. Commit b6dcfc082116293178b0cd762b0be7aba3152d51 made clinical attributes study-specific. Because of that, validating whether the definitions of attributes in a study are consistent with those configured in the portal is now no longer applicable. Removing the dependency on these APIs fixes the problems caused by duplicate entries introduced into this API's output, reported in issue #1804.

Changes proposed in this pull request:
- Remove database consistency validation for clinical attributes
- Remove clinical attribute API calls from the validator
- Clarify the message emitted when attributes do break requirements by adding the line and column numbers of the offending value

PR #1866, opened 2016-11-02
Relaxed the validation for some sample level attributes

pieterlukasse:
# What? Why?
Relaxed the validation for some sample level attributes. This was done for a few attributes...
- TUMOR_SITE
- TUMOR_TISSUE_SITE 
- PRIMARY_SITE
- METASTATIC_SITE
- GLEASON_SCORE

...for which the impact on the UI is expected to be minimal. E.g. METASTATIC_SITE will still appear in clinical data if set in patient clinical file. The only effect expected is it being omitted from a header label in the patient/sample view (according to some logic found in ClinicalAttributesUtil.js). See for example http://www.cbioportal.org/case.do?cancer_study_id=brca_tcga&case_id=TCGA-BH-A18V

- Relaxed the validation for some sample level attributes. This will also help validating some of the current https://github.com/cBioPortal/datahub studies that are now failing. 

# Checks
- [x] Checked the METASTATIC_SITE attribute in some detail in the code and debugging using example link provided above. For other attributes, assumption was made based on previous findings in #1567, #1809 and a new and more flexible clinical attributes data model implemented in #1746 and #1808

# Notify reviewers
@zheins (because it relates to validation of some datahub studies) @inodb (since this relates to some of the clinical attributes that have been reported to be mainly used in the patient/sample view)


pieterlukasse:
tests are reported at https://github.com/cBioPortal/datahub/pull/11

Issue #1858, opened 2016-11-01
Data loading fails for mutation data when using InnoDB


pieterlukasse:
Data loading will fail with a message like : 

```
Reading data from:  /datasets/tests/acc/tcga/data_mutations_extended.txt
Recaching...
Finished recaching...
--> profile id:  8
--> profile name:  Mutations
--> genetic alteration type:  MUTATION_EXTENDED
--> OMA Scores Column Number:  32
--> records inserted into `mutation` table: 0

Warnings / Errors:
-------------------
...

ABORTED! java.lang.RuntimeException: org.mskcc.cbio.portal.dao.DaoException: DB Error: only 0 of the 13431 records were inserted in `mutation`. More error/warning details: Row 1 doesn't contain data for all columns
Traceback (most recent call last):
  File "./cbioportalImporter.py", line 344, in <module>
```

pieterlukasse:
The cause is with the `MySQLbulkLoader.java` code which iterates over `mySQLbulkLoaders` in an varying order and sometimes tries to insert data into `mutation` table before inserting into `mutation_event` table in the method `MySQLbulkLoader.flushAll()`

This, apparently combined with the fact that the statements to temporarily disable FK constraint checks like `stmt = con.prepareStatement("SET foreign_key_checks = ?;");` are probably ignored by MySQL as this is sent via JDBC....see http://stackoverflow.com/questions/19665943/java-jdbc-issue-row-1-doesnt-contain-data-for-all-columns

ericblanc20:
Same problem using thehyve's hotfix branch on `study_es_0`, following the doc. (The branch uses `LinkedHashMap` in `MySQLbulkLoader` & doesn't use bulk loader in `DaoGene.addGene()` ):

```
Reading data from:  ../../../test/scripts/test_data/study_es_0/data_fusions.txt
Recaching...
Finished recaching...
--> profile id:  3
--> profile name:  Fusions
--> genetic alteration type:  FUSION
--> records inserted into `mutation` table: 0

Warnings / Errors:
-------------------
...
ABORTED! java.lang.RuntimeException: org.mskcc.cbio.portal.dao.DaoException: DB Error: only 0 of the 3 records were inserted in `mutation`. More error/warning details: Row 1 doesn't contain data for all columns
```

mysql version 5.7.13, all tables use InnoDB.
macosx 10.10.5, Darwin Kernel Version 14.5.0

Is there a workaround, should I use other versions, and if yes, which one?

[...]

PR #1868, opened 2016-11-03
Removed bulk load option from addGene() method

pieterlukasse:
# What? Why?
Removed bulk load option from addGene() method. This was resulting in a similar issue to #1858, but now for RPPA data (where "dummy" genes need to be added on the fly during import - see `ImportTabDelimData.java`, `parseRPPAGenes()` method). Because many tables depend on gene, we want to add gene directly to DB, and	never through MySQLbulkLoader to avoid any Foreign key constraint violations.

- let DaoGene.addGene() insert any genes directly into DB instead of using a delayed MySQLbulkLoader process

# Checks
- [x] Successfully imported 2 studies from datahub after this fix.

# Notify reviewers
@n1zea144 @zheins

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
@@ -78,15 +78,8 @@ final class DaoGene {
      * @throws DaoException Database Error.
      */
     public static int addGene(CanonicalGene gene) throws DaoException {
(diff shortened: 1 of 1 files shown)


pieterlukasse:
tests are reported at https://github.com/cBioPortal/datahub/pull/11

pieterlukasse:
this one can be closed if #1902 gets merged

pieterlukasse:
ah, wait, this is more urgent...to hotfix. #1902 goes to rc

PR #1873, opened 2016-11-04
Fix bug in validator regarding scientific notation of int

sandertan:
# What? Why?
Fix small bug in validator regarding values in scientific notation (1e+05) in the segment files. We checked the java code, and found it can correctly parse such values to a long.


sandertan:
We also tested this in the portal, and the correct value is displayed.

PR #1884, opened 2016-11-08
Delete summary models

ersinciftci:
Frontend team is generating the Typescript API client from the Swagger JSON spec file, so we can't use summary classes in the spec, so we will show detailed models in the swagger-ui page, which is OK.

Other changes:
Change Clinical Data APIs persistence layer so the interface becomes more verbose
Move chromosome logic from Gene model to GeneServiceImpl

https://docs.google.com/document/d/1HjOXWMq4z1JjeZLkw4eq3aG_o0y3IhZa47bGUtdoJ3Y/edit


persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepository.java
```diff
@@ -21,32 +20,94 @@
...
     @Override
-    public BaseMeta getMetaPatientClinicalData(List<String> studyIds, List<String> patientIds, String attributeId) {
-        return clinicalDataMapper.getMetaPatientClinicalData(studyIds, patientIds, attributeId);
+    public BaseMeta getMetaPatientClinicalData(String studyId, String patientId, String attributeId) {
+        return clinicalDataMapper.getMetaPatientClinicalData(Arrays.asList(studyId), Arrays.asList(patientId),
+                attributeId);
+    }
+
+    @Override
+    public List<ClinicalData> getAllClinicalDataInStudy(String studyId, String attributeId,
```
sheridancbio: passing in a clinicalDataType argument as a way to select between Sample and Patient makes sense if the code differences between the two modes of operation are minor, but below it seems like there is not much sharing of code between modes in the called methods --- the argument is tested and a different subroutine call is selected. If this is the case, maybe we should consider having two different methods here and dropping the clinicalDataType argument.

I guess it also depends on the callers of this function -- if there are very many calls, then making this change would push the "if" statement up into the calling code, which could clutter up the calling code in many places.

service/src/main/java/org/cbioportal/service/impl/ClinicalDataServiceImpl.java
```diff
@@ -15,103 +13,69 @@
...
                                                                          String sortBy, String direction) {
 
-        if (clinicalDataType.equals(SAMPLE_CLINICAL_DATA_TYPE)) {
-            return clinicalDataRepository.getAllClinicalDataOfSampleInStudy(Arrays.asList(studyId), null, attributeId,
-                    projection, pageSize, pageNumber, sortBy, direction);
-        } else {
-            return clinicalDataRepository.getAllClinicalDataOfPatientInStudy(Arrays.asList(studyId), null, attributeId,
-                    projection, pageSize, pageNumber, sortBy, direction);
-        }
+        return clinicalDataRepository.getAllClinicalDataInStudy(studyId, attributeId, clinicalDataType, projection,
```
sheridancbio: maybe here is a case where the "if" logic was eliminated at this level and pushed down into clinicalDataRepository. The alternative would be to push the "if" logic up to callers of the service --- making them choose between two different service level functions. It seems like there is not much shared code here between the two modes here because this is mostly a passthrough call.

web/src/main/java/org/cbioportal/web/ClinicalDataController.java
```diff
@@ -38,7 +38,7 @@
     @RequestMapping(value = "/studies/{studyId}/samples/{sampleId}/clinical-data", method = RequestMethod.GET,
             produces = MediaType.APPLICATION_JSON_VALUE)
     @ApiOperation("Get all clinical data of a sample in a study")
-    public ResponseEntity<List<? extends ClinicalDataSummary>> getAllClinicalDataOfSampleInStudy(
+    public ResponseEntity<List<ClinicalData>> getAllClinicalDataOfSampleInStudy(
             @ApiParam(required = true, value = "Study ID e.g. acc_tcga")
```
sheridancbio: I guess we are relying on this required specification to prevent null values in the studyId and sampleId. But maybe tests for null at the service layer .. and more importantly the persistence layer .. could make sense, because these layers may be called by other code (validator? performance testing code?) where the programmer passes in a null value (maybe thinking that null means "return all" or something). Thoughts?

For now, I'll approve without null argument checks since these are required arguments at the controller level.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
```diff
@@ -37,58 +37,44 @@
 import com.fasterxml.jackson.databind.SerializationFeature;
 import java.util.HashMap;
 import java.util.Map;
-import org.cbioportal.model.*;
```
sheridancbio: I'm surprised I didn't notice this duplication before .. this is a good reason to sort import statements alphabetically.

sheridancbio:
I am merging this now. There is at least one case where a repository function prototype was changed (to accept a string rather than a list of strings), but it seems like the Web API interface has not changed, so front-end callers should not be affected.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/ClinicalDataMyBatisRepository.java
```diff
@@ -21,32 +20,94 @@
     private OffsetCalculator offsetCalculator;
 
     @Override
-    public List<SampleClinicalData> getAllClinicalDataOfSampleInStudy(List<String> studyIds, List<String> sampleIds,
+    public List<ClinicalData> getAllClinicalDataOfSampleInStudy(String studyId, String sampleId,
```
ersinciftci: Web API itself didn't change, I just changed the persistence interface of Clinical Data, because it was not consistent with other APIs, I made it more verbose through having more methods.

web/src/main/java/org/cbioportal/web/ClinicalDataController.java
ersinciftci: I agree that we need to put more effort to make service and persistence layers foolproof, I'm currently more focused on Web API level so there are some parts needs to be improved. We can revisit these parts later when we get feedback from 3rd party users of those layers.

PR #1885, opened 2016-11-08
get sample count by category

JiaoJiao123:
update DETAILED projection to return sample count by sample analysis type

### DIFF
--- a/model/src/main/java/org/cbioportal/model/CancerStudy.java
+++ b/model/src/main/java/org/cbioportal/model/CancerStudy.java
@@ -2,13 +2,101 @@ package org.cbioportal.model;
 
 import org.cbioportal.model.summary.CancerStudySummary;
 
-import java.io.Serializable;
-import java.util.Date;
 
 public class CancerStudy extends CancerStudySummary {
 
     private TypeOfCancer typeOfCancer;
     private Integer sampleCount;
+    private String stableID;
+    private Integer allCount;
+    private Integer sequencedCount;
+    private Integer cnaCount;
+    private Integer rna_seq_v2_mrna_count;
+    private Integer mrna_count;
+    private Integer microrna_count;
+    private Integer methylation_hm27_count;
+    private Integer rppa_count;
+    private Integer complete_count;
+    
+    public Integer getAllCount() {
+        return allCount;
+    }
+
+    public void setAllCount(Integer allCount) {
+        this.allCount = allCount;
+    }
+
+    public Integer getSequencedCount() {
+        return sequencedCount;
+    }
+
+    public void setSequencedCount(Integer sequencedCount) {
+        this.sequencedCount = sequencedCount;
+    }
+
+    public Integer getCnaCount() {
+        return cnaCount;
+    }
+
+    public void setCnaCount(Integer cnaCount) {
+        this.cnaCount = cnaCount;
+    }
+
+    public Integer getRna_seq_v2_mrna_count() {
+        return rna_seq_v2_mrna_count;
+    }
+
+    public void setRna_seq_v2_mrna_count(Integer rna_seq_v2_mrna_count) {
+        this.rna_seq_v2_mrna_count = rna_seq_v2_mrna_count;
+    }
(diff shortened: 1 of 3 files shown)


model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
 
-import java.io.Serializable;
-import java.util.Date;
 
 public class CancerStudy extends CancerStudySummary {
 
     private TypeOfCancer typeOfCancer;
     private Integer sampleCount;
+    private String stableID;
+    private Integer allCount;
```
ersinciftci: allCount is the same thing with sampleCount, you can delete allCount and use sampleCount, because the name is more clear, allCount isn't self explaining, all count of what?

ersinciftci: allCount is the same thing with sampleCount, you can delete allCount and use sampleCount, because the name is more clear, allCount isn't self explaining, all count of what?

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer sequencedCount;
+    private Integer cnaCount;
+    private Integer rna_seq_v2_mrna_count;
```
ersinciftci: no underscore in property names

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer mrna_count;
```
ersinciftci: no underscore in property names

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer microrna_count;
```
ersinciftci: no underscore in property names

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer methylation_hm27_count;
```
ersinciftci: no underscore in property names

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer rppa_count;
```
ersinciftci: no underscore in property names

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer complete_count;
```
ersinciftci: no underscore in property names

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
 import org.cbioportal.model.summary.CancerStudySummary;
 
-import java.io.Serializable;
-import java.util.Date;
 
 public class CancerStudy extends CancerStudySummary {
 
     private TypeOfCancer typeOfCancer;
     private Integer sampleCount;
+    private String stableID;
```
ersinciftci: this is a class representing a cancer study, and cancer study doesn't have a stable id, so stable id of what?

model/src/main/java/org/cbioportal/model/CancerStudy.java
```diff
@@ -2,13 +2,101 @@
...
+    private Integer allCount;
+    private Integer sequencedCount;
```
ersinciftci: can you make the names more clear, like sequencedSampleCount

ersinciftci:
Opened a new PR for the same task https://github.com/cBioPortal/cbioportal/pull/1889

PR #1887, opened 2016-11-09
Potential solution to force increase groupConcatMaxLen

### LINKED ISSUE #1880: Study page: mutation count incorrect sometimes
It happens sometimes that the numbers of mutations are way lower than it should be in the mutations table on study page. It seems more often on large studies.

zhx828:
# What? Why?
Fix #1880 and https://github.com/cBioPortal/iViz/issues/293 .

Make function ```groupConcatMaxLenSet``` in ```MutationMapper``` returns Integer.
We think the reason we randomly have the issue like #1880 is because the script to increase ```group_concat_max_len``` is not executed before the database query sometimes. By doing the fixes in the pull request, the database query will wait until groupConcatMaxLenSet returns the value.
 
# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@ersinciftci @jjgao 
@cBioPortal/backend


ersinciftci:
This is just a little hack we want to try and see if it solves the concurrency problem, so it looks good to me.

PR #1889, opened 2016-11-11
Add datasets related counts to Study API's DETAILED projection

ersinciftci:
Added 8 new properties to CancerStudy model to be used in DETAILED projection, and changed sampleCount property to allSampleCount, and moved it to SUMMARY projection from DETAILED projection.

So the new CancerStudy model looks like this now, lines starting with * are only returned in `DETAILED` projection:
```
{
    "allSampleCount": 0,
    "cancerStudyIdentifier": "string",
    "citation": "string",
    *"cnaSampleCount": 0,
    *"completeSampleCount": 0,
    "description": "string",
    "groups": "string",
    "importDate": "2016-11-10T18:03:22.183Z",
    *"methylationHm27SampleCount": 0,
    *"miRnaSampleCount": 0,
    *"mrnaMicroarraySampleCount": 0,
    *"mrnaRnaSeqV2SampleCount": 0,
    "name": "string",
    "pmid": "string",
    "publicStudy": true,
    *"rppaSampleCount": 0,
    *"sequencedSampleCount": 0,
    "shortName": "string",
    "status": 0,
    *"typeOfCancer": {
      *"clinicalTrialKeywords": "string",
      *"dedicatedColor": "string",
      *"name": "string",
      *"parent": "string",
      *"shortName": "string",
      *"typeOfCancerId": "string"
    *},
    "typeOfCancerId": "string"
  }
```

Performance is not much of a concern for this particular Datasets case, because after the first call, it will be cached by MyBatis indefinitely, and the same static Datasets list will be returned from cache for subsequent calls.

### DIFF
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/StudyMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/StudyMapper.xml
@@ -22,12 +22,27 @@
         </if>
     </sql>
 
+    <sql id="selectDetailed">
+        ,
+        COUNT(CASE WHEN sample_list.STABLE_ID = CONCAT(cancer_study.CANCER_STUDY_IDENTIFIER,'_sequenced') THEN 1 ELSE NULL END) AS sequencedSampleCount,
+        COUNT(CASE WHEN sample_list.STABLE_ID = CONCAT(cancer_study.CANCER_STUDY_IDENTIFIER,'_cna') THEN 1 ELSE NULL END) AS cnaSampleCount,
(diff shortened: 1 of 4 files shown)


JiaoJiao123:
I tested it and doesn't look like the count sample queries brought any performance issue. Looks great!

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/StudyMapper.xml
```diff
@@ -22,12 +22,27 @@
         </if>
     </sql>
 
+    <sql id="selectDetailed">
+        ,
+        COUNT(CASE WHEN sample_list.STABLE_ID = CONCAT(cancer_study.CANCER_STUDY_IDENTIFIER,'_sequenced') THEN 1 ELSE NULL END) AS sequencedSampleCount,
```
sheridancbio: Two concerns .. first, I still worry about performance here due to the additional joins, but I was persuaded that this should be manageable because we use similar joining logic elsewhere. second, I think we may be moving away from sample_list and sample_list_list ... in which case we will need a new implementation of this functionality. But for now this is ok.

PR #1893, opened 2016-11-14
Increase varchar size for allele fields in `MUTATION_EVENT`

angelicaochoa:
# What? Why?
Import was throwing a Data Truncation exception for allele values that were larger than 255 chars. 

* `migration.sql`: increased sizes of `REFERENCE_ALLELE` and `TUMOR_SEQ_ALLELE` to 400 chars from 255 chars to prevent MySQL data truncation exceptions
* updated `cgds.sql` to reflect these changes

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/backend

and notify them here:

### DIFF
--- a/core/src/main/resources/db/cgds.sql
+++ b/core/src/main/resources/db/cgds.sql
@@ -335,8 +335,8 @@ CREATE TABLE `mutation_event` (
   `CHR` varchar(5),
   `START_POSITION` bigint(20),
   `END_POSITION` bigint(20),
-  `REFERENCE_ALLELE` varchar(255),
-  `TUMOR_SEQ_ALLELE` varchar(255),
+  `REFERENCE_ALLELE` varchar(400),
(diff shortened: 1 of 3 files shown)


sheridancbio:
Good fix. I wonder how long the maximum allele can be though .. I am imagining that these longer entries are due to insertions or deletions. (so if 300 residues are inserted, the allele may list 300 residues which were detected at a position ... ?)  If I'm understanding it correctly, maybe we need to search for (or set) an upper limit on the number of residues which are added or removed to be called a "deletion" mutation rather than a fusion or structural variant or some other type of event. I'd say a deletion of 50000 residues is probably not appropriate to represent as a "mutation" but maybe as a fusion or structural variant?

PR #1807, opened 2016-10-14
Adding correct UQ constraint to gene table

pieterlukasse:
# What? Why?

Fixes #1266, adding a unique constraint to the gene table. Replaces #1422. 

Changes proposed in this pull request:
- gene table gets a unique constraint to ensure we don't get wrong data

:warning: if the table gene already contains duplicated entries, the migration script will fail. The script contains some references to documentation on how to fix it in this case. 
# Notify reviewers

@sheridancbio  , @zheins, @jjgao

PR #1899, opened 2016-11-15
Fix to the DB_SCHEMA_VERSION setting in cgds.sql

sheridancbio:
In PR #1818 the DB_SCHEMA_VERSION setting in cgds.sql should have been updated to 1.3.1, but it was not. (migration.sql and pom.xml were updated to 1.3.1, and the table schema were adjusted in cgds.sql)

This change fixes the DB_SCHEMA_VERSION to match (bringing cdgs.sql into sync with migration.sql for the rc branch.)

Also, the version number for rc is skewed upwards to 100.0.0
This is to help distinguish the database schema version from the cbioportal web application release version number. (these numbers should not be expected to match, so this avoids confusion and makes clear that there are two separate things being versioned (the web application itself, and the database version)

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers

@cBioPortal/backend

- migration.sql and pom.xml refer to 1.3.1, and cgds.sql reflects that schema
- 1.3.1 was used "reused" during adjustment PR#1850
- cgds.sql value had not been updated as it should have been in PR#1818 (was still 1.3.0)

### DIFF
--- a/core/src/main/resources/db/cgds.sql
+++ b/core/src/main/resources/db/cgds.sql
@@ -692,7 +692,7 @@ CREATE TABLE `clinical_event_data` (
 
 -- --------------------------------------------------------
 CREATE TABLE `info` (
-  `DB_SCHEMA_VERSION` varchar(8)
+  `DB_SCHEMA_VERSION` varchar(24)
 ) DEFAULT CHARSET=utf8;
 -- THIS MUST BE KEPT IN SYNC WITH db.version PROPERTY IN pom.xml
-INSERT INTO info VALUES ('1.3.0');
+INSERT INTO info VALUES ('1.4.0');
--- a/core/src/main/resources/db/migration.sql
+++ b/core/src/main/resources/db/migration.sql
@@ -191,3 +191,8 @@ CREATE TABLE `structural_variant` (
   FOREIGN KEY (`GENETIC_PROFILE_ID`) REFERENCES `genetic_profile` (`GENETIC_PROFILE_ID`) ON DELETE CASCADE
 );
 UPDATE info SET DB_SCHEMA_VERSION="1.3.1";
+
+##version: 1.4.0
+-- alter version number to distinguish from cbioportal web application version numbering
+ALTER TABLE info MODIFY COLUMN DB_SCHEMA_VERSION VARCHAR(24);
+UPDATE info SET DB_SCHEMA_VERSION="1.4.0";
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -163,7 +163,13 @@ def run_migration(db_version, sql_filename, connection, cursor):
             continue
 
         # skip blank lines
-        if len(line.strip()) < 1 or line.startswith('#'):
+        if len(line.strip()) < 1:
+            continue
+        # skip comments
+        if line.startswith('#'):
+            continue
+        # skip sql comments
+        if line.startswith('--') and len(line) > 2 and line[2].isspace():
             continue
         # only execute sql line if the last version seen in the file is greater than the db_version
         if run_line:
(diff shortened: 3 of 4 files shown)


jjgao:
since the code will be 1.4.0, should we upgrade the database schema version to 1.4.0?

sheridancbio:
I think it is confusing to try to "sync" the DB_SCHEMA_VERSION with the cbioportal release version. They will be continually drifting apart, since there is not a new schema at every public release, and sometimes we will go through several schema updates between public releases. So I would strongly recommend that we actually "skew" the database version number to make it clearly different than the portal codebase version number.

On Friday I hope to discuss a proposal to be more formal about changing the database schema version ... for instance, @angelicaochoa wanted to adjust the database schema in hotfix, but rc has its own changes to the database schema. These will need to be merged in the future, but to keep things clear, about what database is expected by what code base, and what historical database schemas we have released (clarity needed for migration.sql) I think we should always issue a new version number to the schema any time modification-to-schema code is merged to our managed branches (master, rc, hotfix). migration.sql in master should take old versions of the database through the history of the versions which were incorporated into master (perhaps skipping over development versions which were in rc or hotfix if these are not part of the history of master).

So I think at the next merge of rc or hotfix to master we should change to some new numbering which is clearly different -- like 7.1.0 or something, and then continue from there.

jjgao:
@sheridancbio Agreed. Good idea to change to a different numbering, e.g. 3.14, 3.141, ... :)

sheridancbio:
The DB_SCHEMA_VERSION is now skewed away from the range used for the cbioportal web application codebase. (the two should not be expected to match, and should not be compared to each other.) This skew makes it clear that there are two separate version identifiers : one for the database schema, and one for the web application build.

sheridancbio:
DB_SCHEMA_VERSION is set to 3.0.0
Going forward, I will recommend that we either move away from the ordered incrementing (3.0.0 .. 3.0.1 .. 3.0.2 .. 3.1.0 ...) and switch to using checksums on cgds.sql, or that we increment the center number for rc schema adjustments (3.0.0 .. 3.1.0 .. 3.2.0 .. 3.3.0 ...) and increment the final number for hotfix schema adjustments (3.2.0 .. 3.2.1 .. 3.2.2). Doing it this way means we can continue to use migration.sql for updating as is .. but it means that we will need to "hand-apply" hotfix changes to in-development databases which have been advanced in version number without a merged hotfix update.

jjgao:
@sheridancbio maybe we should start with a bigger number, e.g. 100.0.0? Or can we support db1.3.1?

sheridancbio:
I have added a fix to migrate_db.py ... previously it was not ignoring sql-style comments. I have added code to skip over sql-style comments on their own line. (notification : @zheins )

sheridancbio:
@jjgao I don't think we can break the #.#.# pattern without breaking migrate_db.py .. since it compares numbers to see if sections of migration.sql are higher or lower than the current database version. Maybe we should reconsider this incrementing approach. (We can talk at Friday meeting)

Also, the width of DB_SCHEMA_VERSION is limited to 8 characters currently.  I'm going to up that to 256 characters for the field width. I don't think we will be updating major numbers very often .. but who knows. Ok, I'll go with 100.0.0

sheridancbio:
@n1zea144 argued that there is not a good case for skewing the version numbers and that we should allow the cbioportal build version number naturally drift away from the db_schema_version number. If we now increment the major version number of db_schema_version for incompatible changes to the schema, it will soon become clear that the numbers are distinct and so there is no need to artificially skew. I have altered the code in this PR to move to version number 1.4.0, but still to widen the field in the info table to varchar(24)

sheridancbio:
Note : the fix to the migrate_db.py script in this PR is also present in #1807 .. when merging, the two should become one.

PR #1902, opened 2016-11-16
Introducing Genetic entity

pieterlukasse:
# What? Why?
In RFC 28 we discussed the need for a table called `genetic_entity` to support genetic profiles of other entities that are not genes. This PR makes the necessary changes to support this (no other entities are added here yet).  

- table `genetic_alteration` is linked to `genetic_entity` instead of to `gene` and table `gene` now has a reference to `genetic_entity` (see model in [RFC 28 - Appendix 5](https://docs.google.com/document/d/1g5HPPXrLqJQmfxP1RcWGA6-QdcHCTLr_JAuQklQdfMs/edit#heading=h.h5c0v3jwn3i5))
- respective SQL code added to `migration.sql` and `cgds.sql`
- updated Dao layer accordingly 
- updated java unit tests accordingly
- not directly related: update `commons-dbcp` version in 2 pom.xml files

# Notify reviewers
@jjgao @ersinciftci @n1zea144 @angelicaochoa


pieterlukasse:
@inodb the end-to-end test is failing here. But I think it cannot work unless this test also uses an updated version of the DB. Is migration.sql executed in this step?

core/src/main/resources/db/cgds.sql
```diff
@@ -177,16 +178,34 @@ CREATE TABLE `sample_list_list` (
 );
 
 -- --------------------------------------------------------
+CREATE TABLE `genetic_entity` (
+  `INTERNAL_ID` int(11) NOT NULL AUTO_INCREMENT,
+  `ENTITY_TYPE` varchar(45) NOT NULL,
```
ersinciftci: I thought we decided `genetic_entity` table to have just one column?

core/src/main/resources/db/cgds.sql
```diff
@@ -177,16 +178,34 @@ CREATE TABLE `sample_list_list` (
 );
 
 -- --------------------------------------------------------
+CREATE TABLE `genetic_entity` (
+  `INTERNAL_ID` int(11) NOT NULL AUTO_INCREMENT,
```
ersinciftci: Can we call this `ID`? I know we have lots of tables using the name `INTERNAL_ID`, but it's because they have other kinds of IDs in the same table, like STABLE_ID. But this table doesn't have that problem.

pieterlukasse:
@inodb the changes are not backwards compatible. This is a change in `genetic_alteration` table (a.o.) which affects most of the portal.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -67,45 +67,58 @@ private static synchronized int getNextFakeEntrezId() throws DaoException {
     public static synchronized int addGeneWithoutEntrezGeneId(CanonicalGene gene) throws DaoException {
         CanonicalGene existingGene = getGene(gene.getHugoGeneSymbolAllCaps());
         gene.setEntrezGeneId(existingGene==null?getNextFakeEntrezId():existingGene.getEntrezGeneId());
-        return addGene(gene);
+        return addOrUpdateGene(gene);
```
sheridancbio: It is a bit troubling to see updates/development to DaoGene. Since the Dao code is still active, we must make adjustments here I guess, but for each major effort to a Dao class we may want to consider rebuilding the class in a new persistence / service / web manner after the new API under development. We do want to drop the Dao classes (and the core module in general), so these efforts may be wasteful.
Perhaps more important, there is persistence level code in development now (a spring batch importer) based on the existing schemas and relationships between table. Changes such as this will need to be mirrored in that codebase to keep the codebases compatible, so some coordination is needed I think. Also, there is code in the new API for extracting / reporting genes, genetic profiles, etc. That code needs to be updated / coordinated as well. (I expect those change are below in this PR, since there are unit tests written for the new web API endpoints)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -67,45 +67,58 @@ private static synchronized int getNextFakeEntrezId() throws DaoException {
...
-                pstmt = con.prepareStatement
-                        ("INSERT INTO gene (`ENTREZ_GENE_ID`,`HUGO_GENE_SYMBOL`,`TYPE`,`CYTOBAND`,`LENGTH`) "
-                                + "VALUES (?,?,?,?,?)");
-                pstmt.setLong(1, gene.getEntrezGeneId());
-                pstmt.setString(2, gene.getHugoGeneSymbolAllCaps());
-                pstmt.setString(3, gene.getType());
-                pstmt.setString(4, gene.getCytoband());
-                pstmt.setInt(5, gene.getLength());
+            	//new gene, so add genetic entity first:
+            	int geneticEntityId = DaoGeneticEntity.addNewGeneticEntity(DaoGeneticEntity.EntityTypes.GENE);
```
sheridancbio: note: mysql autoincrement is what prohibits reuse of a newly generated geneticEntityId number.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -67,45 +67,58 @@ private static synchronized int getNextFakeEntrezId() throws DaoException {
...
+            	pstmt.setInt(1, geneticEntityId);
+                pstmt.setLong(2, gene.getEntrezGeneId());
+                pstmt.setString(3, gene.getHugoGeneSymbolAllCaps());
+                pstmt.setString(4, gene.getType());
+                pstmt.setString(5, gene.getCytoband());
+                pstmt.setInt(6, gene.getLength());
                 rows += pstmt.executeUpdate();
 
-            }
+            }
```
sheridancbio: my style preference:
    } else {

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
```diff
@@ -265,9 +278,10 @@ private static CanonicalGene getGene(long entrezGeneId) throws DaoException {
                     ("SELECT * FROM gene");
             rs = pstmt.executeQuery();
             while (rs.next()) {
+            	int geneticEntityId = rs.getInt("GENETIC_ENTITY_ID");
                 long entrezGeneId = rs.getInt("ENTREZ_GENE_ID");
                 Set<String> aliases = mapAliases.get(entrezGeneId);
-                CanonicalGene gene = new CanonicalGene(entrezGeneId,
+                CanonicalGene gene = new CanonicalGene(geneticEntityId, entrezGeneId,
```
sheridancbio: I expect this to be considered an internal id .. and not be reported through a web API, since it is arbitrary and subject to change between database rebuilds.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -61,6 +61,7 @@
     //nb: make sure any map is also cleared in clearCache() method below:
     private final HashMap<String, CanonicalGene> geneSymbolMap = new HashMap <String, CanonicalGene>();
     private final HashMap<Long, CanonicalGene> entrezIdMap = new HashMap <Long, CanonicalGene>();
+    private final HashMap<Integer, CanonicalGene> geneticEntityMap = new HashMap<Integer, CanonicalGene>();
```
sheridancbio: note: we are caching these in ram .. like other gene info

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
```diff
@@ -221,6 +225,60 @@ private void cacheGene(CanonicalGene gene) {
...
+    	CanonicalGene gene = daoGeneOptimized.getGene(entrezGeneId);
+		if (gene != null) {
+			return gene.getGeneticEntityId();
+		}
+		else {
+			throw new RuntimeException("Invalid entrezGeneId symbol. Not found in cache: " + entrezGeneId);
+		}    			
+    }
+    
+    public static int getGeneticEntityId(long entrezGeneId, String hugoGeneSymbol) {
```
sheridancbio: If caller passes in both entrezGeneId *and* hugoGeneSymbol, then hugoGeneSymbol is ignored. With this setup, it is possible that entrezGeneId and hugoGeneSymbol arguments will refer to different genes, and this function will not notice the error. One improvement would be to throw an exception if both arguments are not "empty". Another approach would be to resolve the geneticEntityId for both arguments and then throw an exception if the geneticEntityIds for the two arguments differ. We could also have two separate functions ... one which takes entrezGeneId only, and one which takes hugoGeneSymbol only.
None of these changes are required ... I'm just raising the issue for possible consideration.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java
```diff
@@ -81,7 +82,7 @@ ObjectNode process(
             ArrayList<Integer> orderedSampleList
         );
     }
-
+
```
sheridancbio: unneeded whitespace at end of line

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java
```diff
@@ -104,7 +111,7 @@ public int addGeneticAlterations(int geneticProfileId, long entrezGeneId, String
        if (MySQLbulkLoader.isBulkLoad() ) {
           //  write to the temp file maintained by the MySQLbulkLoader
           MySQLbulkLoader.getMySQLbulkLoader("genetic_alteration").insertRecord(Integer.toString( geneticProfileId ),
-                  Long.toString( entrezGeneId ), valueBuffer.toString());
```
sheridancbio: yes .. entrezGeneId should not be Long anyway ... it has been set to an int(11) mysql field everywhere now
a 4 space indent would be standard, although fixing style is not necessary for this PR. (another example are the extra spaces around argument in the function call)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class DaoGeneticEntity {
+
+	/**
+     * Private Constructor to enforce Singleton Pattern.
+     */
+    private DaoGeneticEntity() {
+    }
+    
+    public static enum EntityTypes
```
sheridancbio: @ersinciftci said these embedded enum classes are still globally usable (constants not restricted in scope to the contained methods).
I do wonder whether the enum constant "GENE" is likely to clash with enum constants in other locations. This may lead to confusion. I would suggest constants such as GENETIC_ENTITY_TYPE_GENE, and GENETIC_ENTITY_TYPE_GENE_SET. This is my gut feeling .. perhaps it is not needed.

core/src/main/java/org/mskcc/cbio/portal/dao/MySQLbulkLoader.java
```diff
@@ -244,7 +244,8 @@ private int loadDataFromTempFileIntoDBMS() throws DaoException, IOException {
         	 if (stmt.getWarnings() != null) {
         		 otherDetails = "More error/warning details: " + stmt.getWarnings().getMessage();
              }
-             throw new DaoException("DB Error: only "+updateCount+" of the "+nLines+" records were inserted in `" + tableName + "`. " + otherDetails);
+             throw new DaoException("DB Error: only "+updateCount+" of the "+nLines+" records were inserted in `" + tableName + "`. " + otherDetails + 
+            		 " See tmp file for more details: " + tempFileName);
```
sheridancbio: Extra enhancement?

core/src/main/java/org/mskcc/cbio/portal/model/CanonicalGene.java
```diff
@@ -56,19 +57,39 @@
     private String type;
 
     public CanonicalGene(String hugoGeneSymbol) {
-        this(-1, hugoGeneSymbol);
```
sheridancbio: this seems convoluted. I suppose this is why there is the prior code for resolving by entrezgeneid *or* hugo using a single function call.

core/src/main/java/org/mskcc/cbio/portal/model/CanonicalGene.java
```diff
@@ -56,19 +57,39 @@
...
-    }
-
-    public CanonicalGene(long entrezGeneId, String hugoGeneSymbol) {
-        this(entrezGeneId, hugoGeneSymbol, null);
+        this(hugoGeneSymbol, null);
     }
 
     public CanonicalGene(String hugoGeneSymbol, Set<String> aliases) {
-        this(-1, hugoGeneSymbol, aliases);
```
sheridancbio: these overloaded constructors are not easy to comprehend on first reading .. I would recommend some javadoc comments to explain the logic of these constructors (use cases). Alternatively, we could try to make the constructors simpler / fewer.
postscript: I see there are javadoc comments below here, and those are helpful

core/src/main/resources/db/migration.sql
```diff
@@ -191,3 +191,83 @@ CREATE TABLE `structural_variant` (
...
+-- ========================== start of genetic_entity related migration =============================================
+-- add genetic_entity table 
+CREATE TABLE `genetic_entity` (
+  `INTERNAL_ID` int(11) NOT NULL AUTO_INCREMENT,
+  `ENTITY_TYPE` varchar(45) NOT NULL,
+  PRIMARY KEY (`INTERNAL_ID`)
+);
+-- update gene table to use genetic_element:
+ALTER TABLE `gene` 
+ADD COLUMN `GENETIC_ENTITY_ID` INT NULL FIRST;
```
sheridancbio: will ENTREZ_GENE_ID still be the primary key for the table **gene** ?  I think it is usually good practice to keep the primary key fields first in the column order for a table. I'm not sure if that is recommended, but if GENETIC_ENTITY_ID is the first column I personally might expect it to be the PRIMARY KEY unless I look closer.

core/src/main/resources/db/migration.sql
```diff
@@ -191,3 +191,83 @@ CREATE TABLE `structural_variant` (
   FOREIGN KEY (`GENETIC_PROFILE_ID`) REFERENCES `genetic_profile` (`GENETIC_PROFILE_ID`) ON DELETE CASCADE
 );
 UPDATE info SET DB_SCHEMA_VERSION="1.3.1";
+
+
+##version: 1.3.2
```
sheridancbio: 2.0.0 ?  (See above)

core/src/main/resources/db/migration.sql
```diff
@@ -191,3 +191,83 @@ CREATE TABLE `structural_variant` (
...
+ALTER TABLE `genetic_entity` 
+ADD COLUMN `TMP_GENE_ID` INT NOT NULL AFTER `ENTITY_TYPE`,
+ADD UNIQUE INDEX `TMP_GENE_ID_UNIQUE` (`TMP_GENE_ID` ASC);
+
+-- populate genetic_entity
+insert into genetic_entity (entity_type, tmp_gene_id)
+(Select 'GENE', ENTREZ_GENE_ID from gene);
+
+-- update gene table to have GENETIC_ENTITY_ID point to the correct one:
+SET SQL_SAFE_UPDATES = 0;
```
sheridancbio: mysql specific setting ... I guess we are dependent on this platform in so many ways.

core/src/main/resources/db/migration.sql
```diff
@@ -191,3 +191,83 @@ CREATE TABLE `structural_variant` (
...
+CREATE DEFINER = CURRENT_USER TRIGGER `gene_BEFORE_DELETE` BEFORE DELETE ON `gene` FOR EACH ROW BEGIN  DELETE FROM genetic_entity where genetic_entity.INTERNAL_ID = OLD.GENETIC_ENTITY_ID; END;
+-- if you want to run above directly in DB, use 2 lines below:
+-- DELIMITER $$ 
+-- CREATE DEFINER = CURRENT_USER TRIGGER `gene_BEFORE_DELETE` BEFORE DELETE ON `gene` FOR EACH ROW BEGIN  DELETE FROM genetic_entity where genetic_entity.INTERNAL_ID = OLD.GENETIC_ENTITY_ID; END$$ DELIMITER ;
+
+
+
+-- ========================== end of genetic_entity related migration =============================================
+
+UPDATE info SET DB_SCHEMA_VERSION="1.3.2";
```
sheridancbio: 2.0.0

core/src/main/resources/db/migration.sql
```diff
@@ -191,3 +191,83 @@ CREATE TABLE `structural_variant` (
...
+-- BEFORE delete query to avoid the scenario where a gene is deleted and a genetic_entity is left "orphan"
+-- Definition needs to be on 1 row... (? test in python )
+CREATE DEFINER = CURRENT_USER TRIGGER `gene_BEFORE_DELETE` BEFORE DELETE ON `gene` FOR EACH ROW BEGIN  DELETE FROM genetic_entity where genetic_entity.INTERNAL_ID = OLD.GENETIC_ENTITY_ID; END;
+-- if you want to run above directly in DB, use 2 lines below:
+-- DELIMITER $$ 
+-- CREATE DEFINER = CURRENT_USER TRIGGER `gene_BEFORE_DELETE` BEFORE DELETE ON `gene` FOR EACH ROW BEGIN  DELETE FROM genetic_entity where genetic_entity.INTERNAL_ID = OLD.GENETIC_ENTITY_ID; END$$ DELIMITER ;
+
+
+
+-- ========================== end of genetic_entity related migration =============================================
```
sheridancbio: note .. comment skipping in migrate_db.sql depends on either applying #1807 or applying #1899

core/src/test/resources/seed_mini.sql
```diff
@@ -99,18 +100,40 @@ INSERT INTO "type_of_cancer" ("TYPE_OF_CANCER_ID","NAME","CLINICAL_TRIAL_KEYWORD
...
+INSERT INTO "gene" ("GENETIC_ENTITY_ID","ENTREZ_GENE_ID","HUGO_GENE_SYMBOL","TYPE","CYTOBAND","LENGTH") VALUES (@max_entity_id,675,'BRCA2','protein-coding','13q12.3',11269);
+INSERT INTO "genetic_entity" ("ENTITY_TYPE") VALUES ('GENE');
+SET @max_entity_id = (Select MAX(INTERNAL_ID) from genetic_entity);
+INSERT INTO "gene" ("GENETIC_ENTITY_ID","ENTREZ_GENE_ID","HUGO_GENE_SYMBOL","TYPE","CYTOBAND","LENGTH") VALUES (@max_entity_id,3265,'HRAS','protein-coding','11p15.5',1854);
+INSERT INTO "genetic_entity" ("ENTITY_TYPE") VALUES ('GENE');
+SET @max_entity_id = (Select MAX(INTERNAL_ID) from genetic_entity);
+INSERT INTO "gene" ("GENETIC_ENTITY_ID","ENTREZ_GENE_ID","HUGO_GENE_SYMBOL","TYPE","CYTOBAND","LENGTH") VALUES (@max_entity_id,3845,'KRAS','protein-coding','12p12.1',7302);
+INSERT INTO "genetic_entity" ("ENTITY_TYPE") VALUES ('GENE');
+SET @max_entity_id = (Select MAX(INTERNAL_ID) from genetic_entity);
+INSERT INTO "gene" ("GENETIC_ENTITY_ID","ENTREZ_GENE_ID","HUGO_GENE_SYMBOL","TYPE","CYTOBAND","LENGTH") VALUES (@max_entity_id,4893,'NRAS','protein-coding','1p13.2',4449);
```
sheridancbio: For testing, we probably need to add some genetic_entity records which refer to non-gene entities (e.g. GSVA)
I guess these can be added at the time that the unit tests are written. We do want unit tests which make sure profiles can be queried properly (without cross-type mixing) across all entity types.

sheridancbio:
@pieterlukasse .. I think you can shift your section to the end of migration.sql and increase the db version number in three places to 2.0.0 and that will resolve the conflicts introduced in #1899 . I can talk with you if there is confusion.

core/src/main/resources/db/cgds.sql
pieterlukasse: sure, I'll rename to ID

core/src/main/resources/db/cgds.sql
pieterlukasse: Yes, that is correct. But @n1zea144 would also like to have the entity type represented, and I thought it could make sense to have it here. Could also be in a separate table with a foreign key here pointing to it, but that is perhaps not really needed now. This model can easily evolve in subsequent iterations....this would also be my response to the idea of changing this to int now.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
pieterlukasse: @sheridancbio yes, but there is no way around this yet. There are many many parts of the code still using the Dao layer. See for example all the methods that use `DaoGeneticAlteration.getGeneticAlterationMap(int, Collection<Long>)` directly or indirectly.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
pieterlukasse: correct

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
pieterlukasse: I'll keep that in mind ;)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
pieterlukasse: But these are internal api's right?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
pieterlukasse: correct

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java
pieterlukasse: good point. I removed this method now. It is not used.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java
pieterlukasse: fixed

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
pieterlukasse: I think this is fine. I would like to keep it as it is for now (also related to my previous comments about this field).

core/src/main/java/org/mskcc/cbio/portal/dao/MySQLbulkLoader.java
pieterlukasse: yep, can be quite handy. I believe I used this to troubleshoot one of the unit tests that was failing after the refactoring.

core/src/main/java/org/mskcc/cbio/portal/model/CanonicalGene.java
pieterlukasse: this is the part of the code that I don't like so much...I spent quite some time thinking of how to fix this part of the code...but perhaps we need to do this in a subsequent PR. For now, I focused on getting it to work with the new data model. One of the "problems" we have now is that genetic entity id is not always know, needed or used, depending on the data type or the process that is using it. E.g. 
- when importing new genes batch wise (not so often) we don't have a genetic entity yet
- when making a new "gene" on the fly for some entities like "phosphoproteins" (somewhat more often) we don't have a genetic entity yet
- when retrieving mutation data we don't need genetic entity id since all mutation data is still linked directly to entrez gene id. This also applies to discrete CNA data (via `cna_event` table)
- when retrieving genetic alteration data (such as RPPA, mRNA, etc) then genetic entity is needed

I'm therefore, a.o., wondering whether genetic entity id should be part of `CanonicalGene`. Maybe we should later on split this up in different classes to make such distinctions more explicit...

core/src/main/java/org/mskcc/cbio/portal/model/CanonicalGene.java
pieterlukasse: I agree. I added some comments now and deprecated one of them.

core/src/main/resources/db/cgds.sql
```diff
@@ -177,16 +178,34 @@ CREATE TABLE `sample_list_list` (
...
+  `GENETIC_ENTITY_ID` int(11) NOT NULL,
   `ENTREZ_GENE_ID` int(11) NOT NULL,
   `HUGO_GENE_SYMBOL` varchar(255) NOT NULL,
   `TYPE` varchar(50),
   `CYTOBAND` varchar(50),
   `LENGTH` int(11),
   PRIMARY KEY (`ENTREZ_GENE_ID`),
-  KEY `HUGO_GENE_SYMBOL` (`HUGO_GENE_SYMBOL`)
+  UNIQUE KEY `GENETIC_ENTITY_ID_UNIQUE` (`GENETIC_ENTITY_ID`),  
+  KEY `HUGO_GENE_SYMBOL` (`HUGO_GENE_SYMBOL`),
```
pieterlukasse: note: here I am *not* changing anything in the KEY `HUGO_GENE_SYMBOL`, it just moved down one line.

core/src/main/resources/db/cgds.sql
```diff
@@ -177,16 +178,34 @@ CREATE TABLE `sample_list_list` (
...
+  FOREIGN KEY (`GENETIC_ENTITY_ID`) REFERENCES `genetic_entity` (`INTERNAL_ID`) ON DELETE CASCADE
 );
 
+DROP TRIGGER IF EXISTS `gene_BEFORE_DELETE`;
+-- BEFORE delete trigger to avoid the scenario where a gene is deleted and a genetic_entity is left "orphan"
+-- Definition needs to be on 1 row..otherwise mvn test fails here...
+CREATE DEFINER = CURRENT_USER TRIGGER `gene_BEFORE_DELETE` BEFORE DELETE ON `gene` FOR EACH ROW BEGIN  DELETE FROM genetic_entity where genetic_entity.INTERNAL_ID = OLD.GENETIC_ENTITY_ID; END;
```
pieterlukasse: good point. We could do without this trigger. I will remove it.

core/src/main/resources/db/cgds.sql
```diff
@@ -695,4 +713,4 @@ CREATE TABLE `info` (
   `DB_SCHEMA_VERSION` varchar(8)
 ) DEFAULT CHARSET=utf8;
 -- THIS MUST BE KEPT IN SYNC WITH db.version PROPERTY IN pom.xml
-INSERT INTO info VALUES ('1.3.0');
+INSERT INTO info VALUES ('1.3.2');
```
pieterlukasse: I agree, it will help older portals to run on a newer DB (just wondering if this will happen often). I will update the number.

core/src/main/resources/db/migration.sql
pieterlukasse: :+1:

core/src/main/resources/db/migration.sql
pieterlukasse: agree, I will update this

core/src/main/resources/db/migration.sql
pieterlukasse: yeah...although I think this might only be needed when running in mysql directly. I still need to test running this via the python script. Will do.

core/src/main/resources/db/migration.sql
pieterlukasse: thanks. I see #1899 is merged now.

core/src/main/resources/db/migration.sql
pieterlukasse: :+1:

core/src/test/resources/seed_mini.sql
pieterlukasse: I am leaving GSVA out of this PR on purpose. This PR is really just to introduce genetic entity.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java
sheridancbio: DaoGene may be internal, but we have endpoints like this: https://github.com/cBioPortal/cbioportal/blob/rc/web/src/main/java/org/cbioportal/web/GeneController.java which return a Gene object or list of Gene objects. If the Gene model adds geneticEntitiyId field, we need to suppress it before the response is delivered from the endpoint (maybe here: https://github.com/cBioPortal/cbioportal/blob/rc/web/src/main/java/org/cbioportal/web/mixin/GeneMixin.java)

core/src/main/resources/db/cgds.sql
sheridancbio: I see. It should be fine .. but will take a little manual code combination to make sure both changes are integrated into the new schema.

sheridancbio:
Looks good. Some testing needed here, but then I plan to approve. Plus, we need to prepare a second database, and a pipeline for keeping the second database up to date. (to support both master and rc deployments)

Also: set of commits includes a merge commit, which causes difficulty in squashing .. but I will try to get around that. (in future, please try to avoid adding merge commits into the branch)

sheridancbio:
@jjgao @n1zea144 @inodb After this PR is merged to rc, the rc code and the master code will not be able to use the same database for a time. We have a few options on how to manage that, such as:
1. discontinue the public deployment of rc until master uses the new schema
1. create a "snapshot" database for rc with the new schema which will not be updated (at least, not nightly)
1. create a process which will run after the nightly update and will clone the master database and then apply the migration script (to make it usable by rc)
1. add code to this PR to support a "backwards compatibility" mode flag which will let us configure rc to *not* require the genetic_entity table (perhaps simulating or sidestepping it)

We should talk/plan what happens after this PR gets merged (it is close to ready, in my opinion)
(my opinion is that it would be good if we can figure out a way to do the "clone and migrate" solution ... since that would be useful for future major number increases to the database schema too)

sheridancbio:
After conversation with @jjgao, we are thinking that we will move forward like this:
asap:
  -  partially migrate the amazon database which supplies heroku so that it has the new database features but does not destroy the old links and fields. This database should function with the old code base and the new codebase which relies on (merges) this PR
when portal release 1.4.0 is ready (next week?):
 - migrate databases (msk/public) to what is appropriate for 1.4.0 codebase (will not include this PR)
 - build 1.4.0 code and release
after release:
 - merge this PR to rc
- dump the cgds_public database and import a clone on dashi-dev
- migrate fully the dashi-dev database to be compatible with the (new) rc branch
- release public/beta, which is built on the rc branch and points to the dashi-dev DB
- do not update and stop updating msk/beta --- it will stay identical to root until 1.5.0
during development of 1.5.0:
 - new beta releases go to public/beta (not msk/beta)
when portal release 1.5.0 is ready (December?):
- migrate databases (msk/public) fully to the latest schema (will include this PR)
- build 1.5.0 code and release (public/beta will now return to using the standard database)
- dump the new database and import to amazon for use with heroku going forward

sheridancbio:
Since we think we will wait until 1.4.0 is released before merging this to RC, I think work could begin on the next PR which goes forward from this point (the GSVA profiles, I suppose).

pieterlukasse:
@sheridancbio thanks. I will fix the conflict in DaoGene that is being reported here by github and then continue with GSVA.

pieterlukasse:
@sheridancbio added extra merge with rc to solve conflict in `DaoGene.java` and also fix an error introduced in `migration.sql` at 0857751260c815698f0856bfae0c051b43a78f3c

pieterlukasse:
@sheridancbio it is all squashed in 1 commit now.

pieterlukasse:
all checks have passed :tada:

sheridancbio:
@pieterlukasse ... if you can rebase to the head of RC, I will merge now.

pieterlukasse:
@sheridancbio rebase done.

sheridancbio:
running quick local test, then plan to merge

Issue #1699, opened 2016-09-22
chromosome of gene may be both X and Y


sheridancbio:
During the development of a Gene Repository and Api, a function for resolving the chromosome of a gene based on the filed CYTOBAND in table gene was needed. Most genes can be resolved to one of the numbered chromosomes or to X or to Y. But there are a handful of genes which are present on both X and Y chromosomes.

For now, we are going to report these genes as on "X". This will be used initially for selecting chromosome segment data for display in IGV.js which is able to display chromosome Y. For a small set of genes which are present on both X and Y, segment data for these genes may be absent when viewing chromosome Y.

Is this an issue which we need to correct? How should this be properly handled. When asking the chromosome of a gene, should there be a "bothXandY" return value?

sheridancbio:
linked PR is #1665 
mysql> select * from gene where cytoband like "%;%" and TYPE="protein-coding";
+----------------+------------------+----------------+------------------+--------+
| ENTREZ_GENE_ID | HUGO_GENE_SYMBOL | TYPE | CYTOBAND | LENGTH |
+----------------+------------------+----------------+------------------+--------+
| 286530 | P2RY8 | protein-coding | Xp22.33; Yp11.3 | 4459 |
| 6473 | SHOX | protein-coding | Xp22.33;Yp11.3 | 6114 |
| 9189 | ZBED1 | protein-coding | Xp22.33;Yp11 | 4755 |
| 8225 | GTPBP6 | protein-coding | Xp22.33; Yp11.32 | 2363 |
| 8623 | ASMTL | protein-coding | Xp22.3; Yp11.3 | 3506 |
| 28227 | PPP2R3B | protein-coding | Xp22.33; Yp11.3 | 9015 |
| 55344 | PLCXD1 | protein-coding | Xp22.33; Yp11.32 | 6191 |
| 64109 | CRLF2 | protein-coding | Xp22.3; Yp11.3 | 1205 |
| 207063 | DHRSX | protein-coding | Xp22.33; Yp11.2 | 3620 |
+----------------+------------------+----------------+------------------+--------+

jjgao:
It looks like the current API cannot handle it when we are requesting data for X or Y chromosome.

E.g. http://www.cbioportal.org/index.do?session_id=5b353242498eb8b3d566c301

request:
![image](https://user-images.githubusercontent.com/840895/42055494-511363fc-7ae5-11e8-8aa4-0601f1673ec4.png)

response:
[...]

PR #1906, opened 2016-11-16
Add 4 endpoints in Genes API

ersinciftci:
GET /genes
POST /genes/fetch
GET /genes/{geneId}
GET /genes/{geneId}/aliases

geneId can be entrezGeneId or hugoGeneSymbol, it's treated as entrezGeneId if it's integer.

Also updated igv.jsp to use the new API.


ersinciftci:
@JiaoJiao123 I updated igv.jsp to use the new API in this PR, but don't know in what page to test it. Can you test it using the heroku app in this PR?

JiaoJiao123:
@ersinciftci yeah sure I will test it and let you know.

sheridancbio:
starting to review code

sheridancbio:
I've verified that there should be no Hugo Gene Symbols which are parseable as integers. See http://www.genenames.org/about/guidelines (all symbols start with a letter)

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
```diff
@@ -4,16 +4,115 @@
...
+    <select id="getGenes" resultType="org.cbioportal.model.Gene">
+        SELECT
+        <include refid="select">
+            <property name="prefix" value=""/>
+        </include>
+        FROM gene
+        <if test="sortBy != null and projection != 'ID'">
+            ORDER BY ${sortBy} ${direction}
+        </if>
+        <if test="projection == 'ID'">
```
sheridancbio: In other endpoints do we order by the returned ID when using projection ID?  I think it is a good idea ... if we don't do it elsewhere, it would be good to carry it over to other endpoints. We could respect the direction of sort argument though .. here we hardcode it to be ASC (not a required change)

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMyBatisRepository.java
```diff
@@ -31,29 +31,84 @@
...
+    }
+
+    @Override
+    public Gene getGeneByHugoGeneSymbol(String hugoGeneSymbol) {
+
+        return geneMapper.getGeneByHugoGeneSymbol(hugoGeneSymbol, PersistenceConstants.DETAILED_PROJECTION);
+    }
+
+    @Override
+    public List<String> getAliasesOfGeneByEntrezGeneId(Integer entrezGeneId) {
```
sheridancbio: I'm not positive about how this is used .. but I suspect that the main use case for aliases is to do the reverse of what is supported here. I think the reason we need the aliases is so that when a user enters a gene name into the query box we can go from alias string to the gene object. That functionality is not provided here. Perhaps the use case for this "alias to gene" request is buried in servlets that we have not yet tried to replace. We should seek comments from others ( @jjgao @n1zea144 ?) . 
It is ok to go forward with this as is and add the needed functionality later .. but good to contemplate the need now, while we are adjusting the model.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
```diff
@@ -31,13 +31,30 @@
...
+
+import java.util.List;
 
-/**
- *
- * @author jiaojiao
- */
 public interface GeneRepository {
-    List<Gene> getGeneListByHugoSymbols(List<String> hugo_gene_symbol);
+
```
ersinciftci: I'll add javadocs for all interfaces after I finish implementing APIs in my design list.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneMyBatisRepositoryTest.java
```diff
@@ -43,35 +10,219 @@
...
+
+        List<Gene> result = geneMyBatisRepository.getAllGenes("SUMMARY", null, null, "hugoGeneSymbol", "ASC");
+
+        Assert.assertEquals(16, result.size());
+        Assert.assertEquals("AKT1", result.get(0).getHugoGeneSymbol());
+        Assert.assertEquals("AKT2", result.get(1).getHugoGeneSymbol());
+        Assert.assertEquals("AKT3", result.get(2).getHugoGeneSymbol());
+        Assert.assertEquals("ARAF", result.get(3).getHugoGeneSymbol());
+        Assert.assertEquals("ATM", result.get(4).getHugoGeneSymbol());
+        Assert.assertEquals("BRAF", result.get(5).getHugoGeneSymbol());
```
ersinciftci: Done.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/GeneMyBatisRepositoryTest.java
```diff
@@ -1,40 +1,7 @@
-/*
- * Copyright (c) 2016 Memorial Sloan Kettering Cancer Center.
- *
- * This library is distributed in the hope that it will be useful, but WITHOUT
```
ersinciftci: Done.

model/src/main/java/org/cbioportal/model/GeneticProfile.java
```diff
@@ -4,11 +4,28 @@
 
 public class GeneticProfile implements Serializable {
 
+    public enum GeneticAlterationType {
```
ersinciftci: Done.

sheridancbio:
@JiaoJiao123 : if you do any manual testing of this PR and see that igv_js is working correctly with the new API, please let me know. If the testing looks good, I can do the merging of this faster.

service/src/main/java/org/cbioportal/service/GeneService.java
```diff
@@ -31,13 +31,23 @@
...
 
-/**
- *
- * @author jiaojiao
- */
 public interface GeneService {
-    List<Gene> getGeneListByHugoSymbols(List<String> hugo_gene_symbol);
+
+    List<Gene> getAllGenes(String projection, Integer pageSize, Integer pageNumber, String sortBy, String direction);
+
```
sheridancbio: want javadoc here eventually

JiaoJiao123:
@ersinciftci @sheridancbi 
Ersin Thank you for making the change in the IGV file. I just checked out the CN Segment tab from the heroku link. Nothing breaks. And the files looks good.

service/src/main/java/org/cbioportal/service/exception/GeneNotFoundException.java
```diff
@@ -0,0 +1,19 @@
+package org.cbioportal.service.exception;
```
sheridancbio: creating service layer exceptions here is good. I just wonder if we plan to use exception throwing from the persistence layer too? If so, it might be good to keep all of our custom exception classes together in a lower module. But this is fine for now.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -31,50 +31,138 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 @Service
-public class GeneServiceImpl implements GeneService{
+public class GeneServiceImpl implements GeneService {
+
     @Autowired
     private GeneRepository geneRepository;
-
```
sheridancbio: good style improvements

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -31,50 +31,138 @@
...
+    public List<Gene> getAllGenes(String projection, Integer pageSize, Integer pageNumber, String sortBy,
+                                  String direction) {
 
-        List<Gene> genes = geneRepository.getGeneListByHugoSymbols(hugo_gene_symbol);
-        for (Gene gene : genes) {
-            gene.setChromosome(getChromosome(gene));
+        List<Gene> geneList = geneRepository.getAllGenes(projection, pageSize, pageNumber, sortBy, direction);
+
+        for (Gene gene : geneList) {
+            gene.setChromosome(getChromosome(gene.getCytoband()));
```
sheridancbio: post-processing the returns from persistence in service. This should be fine.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -31,50 +31,138 @@
...
-        if (gene.getCytoband() == null) {
+        return geneRepository.getMetaGenes();
+    }
+
+    @Override
+    public Gene getGene(String geneId) throws GeneNotFoundException {
+
+        Gene gene;
+
+        if (isInteger(geneId)) {
```
sheridancbio: logic for detecting Integer v. Hugo argument. Since this is argument parsing, it might have gone in the web layer, but it is fine here too I think.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -31,50 +31,138 @@
...
+        }
+    }
+
+    @Override
+    public List<Gene> fetchGenes(List<String> geneIds, String projection) {
+
+        List<Integer> entrezGeneIds = new ArrayList<>();
+        List<String> hugoGeneSymbols = new ArrayList<>();
+
+        splitIdsByType(geneIds, entrezGeneIds, hugoGeneSymbols);
```
sheridancbio: nice. Handles a mixed-type list more abstractly.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -31,50 +31,138 @@
...
+            if (isInteger(geneId)) {
+                entrezGeneIds.add(Integer.valueOf(geneId));
+            } else {
+                hugoGeneSymbols.add(geneId);
+            }
+        }
+    }
+
+    private boolean isInteger(String geneId) {
+        return geneId.matches("^-?\\d+$");
```
sheridancbio: There must be a built-in library call for this -- we should not have to write this function. Then again, maybe the built in function detects too much (like octal coding or some other oddity). I guess this matches "simple" integers.
I know that we have negative gene identifiers in our gene table, so I guess the minus sign is needed.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -31,50 +31,138 @@
...
             return null;
         }
-        if (gene.getCytoband().toUpperCase().startsWith("X")) {
+
+        cytoband = cytoband.toUpperCase();
+        if (cytoband.startsWith("X")) {
             return "X";
-        }
-        if (gene.getCytoband().toUpperCase().startsWith("Y")) {
+        } else if (cytoband.startsWith("Y")) {
```
sheridancbio: I'll point out again that some Cytoband entries contain a combination of X and Y .. for genes located on both X and Y. We may want to model this properly .. although for now I think all actual cases start with X and nearly all individuals do have an X chromosome so in a simplified model this is pretty close to "true". See https://github.com/cBioPortal/cbioportal/issues/1699

service/src/test/java/org/cbioportal/service/impl/GeneServiceImplTest.java
```diff
@@ -0,0 +1,169 @@
...
+        Gene gene1 = new Gene();
+        gene1.setCytoband("12q13.13");
+        expectedGeneList1.add(gene1);
+        List<Gene> expectedGeneList2 = new ArrayList<>();
+        Gene gene2 = new Gene();
+        expectedGeneList2.add(gene2);
+
+        List<String> geneIds = new ArrayList<>();
+        geneIds.add(ENTREZ_GENE_ID.toString());
+        geneIds.add(HUGO_GENE_SYMBOL);
```
sheridancbio: Mixed type gene list test. Good

service/src/test/java/org/cbioportal/service/impl/GeneServiceImplTest.java
```diff
@@ -0,0 +1,169 @@
...
+    public void fetchMetaGenes() throws Exception {
+
+        BaseMeta expectedBaseMeta1 = new BaseMeta();
+        expectedBaseMeta1.setTotalCount(1);
+        BaseMeta expectedBaseMeta2 = new BaseMeta();
+        expectedBaseMeta2.setTotalCount(1);
+
+        List<String> geneIds = new ArrayList<>();
+        geneIds.add(ENTREZ_GENE_ID.toString());
+        geneIds.add(HUGO_GENE_SYMBOL);
```
sheridancbio: mixed type fetch for meta

service/src/test/java/org/cbioportal/service/impl/GeneServiceImplTest.java
```diff
@@ -0,0 +1,169 @@
...
+                .thenReturn(expectedBaseMeta1);
+
+        Mockito.when(geneRepository.fetchMetaGenesByHugoGeneSymbols(Arrays.asList(HUGO_GENE_SYMBOL)))
+                .thenReturn(expectedBaseMeta2);
+
+        BaseMeta result = geneService.fetchMetaGenes(geneIds);
+
+        Assert.assertEquals((Integer) 2, result.getTotalCount());
+    }
+}
```
sheridancbio: What should happen when fetchGenes is passed an empty list?  Empty response? Exception?  Maybe we should add a "negative case" unit test.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
ersinciftci: yeah Apache Commons detects too much. Couldn't find one for simple integers with minus sign.

service/src/test/java/org/cbioportal/service/impl/GeneServiceImplTest.java
sheridancbio: Maybe the "empty list" case is prevented because of a required argument at the web layer. Still can't hurt to have these tests eventually.

web/src/main/java/org/cbioportal/web/GeneController.java
```diff
@@ -19,37 +30,75 @@
...
+            return new ResponseEntity<>(responseHeaders, HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(
+                    geneService.getAllGenes(projection.name(), pageSize, pageNumber,
+                            sortBy == null ? null : sortBy.name(), direction.name()), HttpStatus.OK);
+        }
     }
 
-    @RequestMapping(value = "/genes/{entrezGeneId}", method = RequestMethod.GET)
+    @RequestMapping(value = "/genes/{geneId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
sheridancbio: this can now be hugo too

web/src/main/java/org/cbioportal/web/GeneController.java
```diff
@@ -19,37 +30,75 @@
...
 
-        throw new UnsupportedOperationException();
+        return new ResponseEntity<>(geneService.getGene(geneId), HttpStatus.OK);
     }
 
-    @RequestMapping(value = "/genes/fetch", method = RequestMethod.POST)
-    @ApiOperation("Fetch genes by ID")
-    public ResponseEntity<List<Gene>> fetchGenes(@RequestParam(defaultValue = "SUMMARY") Projection projection,
-                                                                  @RequestBody List<String> entrezGeneIds) {
+    @RequestMapping(value = "/genes/{geneId}/aliases", method = RequestMethod.GET,
```
sheridancbio: as mentioned above .. I think we will eventually need a "fetch by alias" function .. either as a separate endpoint or perhaps as additional functionality in the /genes/{geneId} endpoint ... making that work to also do alias resolution. But it is probably better to have a separate "alias-included" query from an "exact match" query

web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
```diff
@@ -1,6 +1,7 @@
 package org.cbioportal.web.error;
 
 import org.cbioportal.service.exception.CancerTypeNotFoundException;
+import org.cbioportal.service.exception.GeneNotFoundException;
```
sheridancbio: When we do a /genes/fetch query and supply a list of genes, I wonder what happens if one gene in a long list is not found. I'm guessing we throw the exception and there is no return value. I'm not sure if we need the use case of supplying  a list of gene identifiers and the system returns all matching genes, dropping those which fail to match. Probably not. Maybe this can be handled by an "ID" projection query? But the exception generation logic would not be projection-type sensitive so I don't think this is covered (if we even need it).

web/src/main/resources/applicationContext-web.xml
```diff
@@ -37,7 +37,6 @@
     <mvc:interceptors>
         <mvc:interceptor>
             <mvc:mapping path="/**"/>
-            <mvc:exclude-mapping path="/*"/>
```
sheridancbio: is this a bug fix?

web/src/test/java/org/cbioportal/web/GeneControllerTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+    @Test
+    public void getGeneNotFound() throws Exception {
+
+        Mockito.when(geneService.getGene(Mockito.anyString())).thenThrow(new GeneNotFoundException("test_gene_id"));
+
+        mockMvc.perform(MockMvcRequestBuilders.get("/genes/test_gene_id")
+                .accept(MediaType.APPLICATION_JSON))
+                .andExpect(MockMvcResultMatchers.status().isNotFound())
+                .andExpect(MockMvcResultMatchers.jsonPath("$.message").value("Gene not found: test_gene_id"));
```
sheridancbio: single gene id not found (uses exception)

web/src/main/resources/applicationContext-web.xml
ersinciftci: yes, it was my mistake

sheridancbio:
Ok, merging now, relying on hand testing of igv from @JiaoJiao123

PR #1923, opened 2016-11-22
Position mutation count endpoint

adamabeshouse:
Implements a new endpoint in the legacy API which gives the total number of mutations in the database for a given codon position. This special case allows us to considerably speed up as opposed to the general case of counting mutations which fall in position ranges.

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@sheridancbio @ersinciftci @jjgao

### DIFF
--- a/portal/src/main/webapp/js/api/cbioportal-datamanager.js
+++ b/portal/src/main/webapp/js/api/cbioportal-datamanager.js
@@ -170,11 +170,10 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 
     var getCBioPortalMutationCounts = function (webservice_data) {
 	/* In: - webservice_data, a list of data obtained from the webservice API
-	 * Out: Promise which resolves with map from gene+","+start_pos+","+end_pos to cbioportal mutation count for that position range and gene
+	 * Out: Promise which resolves with map of type [gene]=>[codon position]=>mutation count
 	 */
-	var counts_map = {};
 	var def = new $.Deferred();
-	var to_query = {};
+	var to_query = {}; // [gene]=>[position]=>true, iff it should be queried
 	for (var i = 0; i < webservice_data.length; i++) {
 	    var datum = webservice_data[i];
 	    if (datum.genetic_alteration_type !== "MUTATION_EXTENDED" || datum.simplified_mutation_type !== "missense") {
@@ -182,47 +181,37 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	    }
 	    var gene = datum.hugo_gene_symbol;
 	    var start_pos = datum.protein_start_position;
-	    var end_pos = datum.protein_end_position;
-	    if (gene && start_pos && end_pos && !isNaN(start_pos) && !isNaN(end_pos)) {
-		to_query[gene + ',' + parseInt(start_pos, 10) + ',' + parseInt(end_pos, 10)] = true;
+	    if (gene && start_pos && !isNaN(start_pos)) {
+		to_query[gene] = to_query[gene] || {};
+		to_query[gene][parseInt(start_pos, 10)] = true;
 	    }
 	}
-	var queries = Object.keys(to_query).map(function (x) {
-	    var splitx = x.split(',');
-	    return {
-		gene: splitx[0],
-		start_pos: splitx[1],
-		end_pos: splitx[2]
-	    };
-	});
-	var genes = queries.map(function (q) {
-	    return q.gene;
-	});
-	var starts = queries.map(function (q) {
-	    return q.start_pos;
-	});
-	var ends = queries.map(function (q) {
-	    return q.end_pos;
-	});
-
-	if (queries.length > 0) {
-	    window.cbioportal_client.getMutationCounts({
-		'type': 'count',
-		'per_study': false,
-		'gene': genes,
-		'start': starts,
-		'end': ends,
-		'echo': ['gene', 'start', 'end']
-	    }).then(function (counts) {
-		for (var i = 0; i < counts.length; i++) {
-		    var gene = counts[i].gene;
-		    var start = parseInt(counts[i].start, 10);
-		    var end = parseInt(counts[i].end, 10);
-		    counts_map[gene + ',' + start + ',' + end] = parseInt(counts[i].count, 10);
-		}
-		def.resolve(counts_map);
-	    }).fail(function () {
+	if (Object.keys(to_query).length > 0) {
+	    var query_genes = Object.keys(to_query);
+	    var query_positions = query_genes.map(function(gene) {
+		return Object.keys(to_query[gene]).map(function(pos) { return parseInt(pos, 10); });
+	    });
+	    $.ajax({
+		type: "POST",
+		url: "api-legacy/position-mutation-count",
+		data: {
+		    genes: query_genes,
+		    positions: query_positions
+		},
+		dataType: "json",
+		traditional: true
+	    }).fail(function() {
 		def.reject();
+	    }).then(function(position_mutation_counts) {
+		var gene_to_position_to_count = {};
+		for (var i=0; i<position_mutation_counts.length; i++) {
+		    var gene = position_mutation_counts[i].hugoGeneSymbol;
+		    var position = position_mutation_counts[i].codonPosition;
+		    var count = position_mutation_counts[i].mutationCount;
+		    gene_to_position_to_count[gene] = gene_to_position_to_count[gene] || {};
+		    gene_to_position_to_count[gene][position] = count;
+		}
+		def.resolve(gene_to_position_to_count);
 	    });
 	} else {
 	    def.resolve({});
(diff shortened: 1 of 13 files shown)


adamabeshouse:
@sheridancbio what do you think of this improvement on https://github.com/cBioPortal/cbioportal/pull/1842 ?

jjgao:
What's the status of this pr?

service/src/main/java/org/cbioportal/service/impl/MutationCountServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+ * @author abeshoua
+ */
+@Service
+public class MutationCountServiceImpl implements MutationCountService {
+	
+	@Autowired
+	private MutationCountRepository mutationCountRepository;
+	
+	@Override
+	public List<PositionMutationCount> getPositionMutationCounts(List<String> genes, List<List<Integer>> positions) {
```
sheridancbio: here we are bridging the gap between the api getting a call for a list of genes, and the repository providing responses for a single gene. The API would be simpler if this logic was moved to the frontend .. but I understand that we may want to make a "batch" call with a bunch of genes rather than making many calls with single genes. (because of internet request latency and performance)  But from a DMBS point of view, the cost in backend processing is the same, since the call is "serialized" here.  In fact .. multiple parallel calls might execute faster on the backend if the DBMS is serviced by many threads.

service/src/main/java/org/cbioportal/service/impl/MutationCountServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+		List<PositionMutationCount> counts = new LinkedList<>();
+		Iterator<String> geneIt = genes.iterator();
+		Iterator<List<Integer>> positionIt = positions.iterator();
+		while (geneIt.hasNext()) {
+			if (!positionIt.hasNext()) {
```
sheridancbio: I see .. if we don't have the same count in the gene list and the list of position lists, then we are going to stop processing excess genes. But maybe it would be best to do a simple test and if size(gene array) != size(positions list array) then the call should return with an error status (400 Bad Request) and message.

If the gene array is too large, we probably should not be ignoring the "surplus" genes .. or if the gene list is too short we probably should not assume that the surplus position lists which were passed in are meaningless.

Or maybe I am confused here. I would expect that for every gene, we would get a position list (perhaps an empty one) .. so that they have a 1 to 1 match

service/src/test/java/org/cbioportal/service/impl/MutationCountServiceImplTest.java
```diff
@@ -0,0 +1,187 @@
...
+    public void getPositionMutationCountsOnePositionListMissing() throws Exception {
+
+        ArrayList<String> testGenes = new ArrayList<>();
+	testGenes.add("BRAF");
+	testGenes.add("KRAS");
+        ArrayList<List<Integer>> testPositions = new ArrayList<>();
+	ArrayList<Integer> brafPositions = new ArrayList<>();
+	brafPositions.add(1);
+	brafPositions.add(2);
+	testPositions.add(brafPositions);
```
sheridancbio: so we are calling without supplying the kras position list? Then we expect only the braf result coming out.
I don't understand why this functionality should be supported. I think there is a concept I am missing here

service/src/test/java/org/cbioportal/service/impl/MutationCountServiceImplTest.java
```diff
@@ -0,0 +1,187 @@
...
+        List<PositionMutationCount> brafResult = new ArrayList<>();
+	List<PositionMutationCount> krasResult = new ArrayList<>();
+	brafResult.add(pmc1);
+	krasResult.add(pmc2);
+
+        Mockito.when(mutationCountRepository.getPositionMutationCounts(org.mockito.Matchers.eq("BRAF"), org.mockito.Matchers.anyList())).thenReturn(brafResult);
+	Mockito.when(mutationCountRepository.getPositionMutationCounts(org.mockito.Matchers.eq("KRAS"), org.mockito.Matchers.anyList())).thenReturn(krasResult);
+	
+	List<PositionMutationCount> expectedResult = new ArrayList<>();
+	List<PositionMutationCount> result = mutationCountService.getPositionMutationCounts(testGenes, testPositions);
```
sheridancbio: Here, we passed in no genes, but two position lists .. and we expect nothing out.
This also confuses me.

sheridancbio:
Just to add a quick comment .. overall, I think this looks good. It has the right layers, good unit tests, code looks basically good etc..

So not many changes may be needed.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -170,59 +170,48 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
-		    counts_map[gene + ',' + start + ',' + end] = parseInt(counts[i].count, 10);
-		}
-		def.resolve(counts_map);
-	    }).fail(function () {
+	if (Object.keys(to_query).length > 0) {
+	    var query_genes = Object.keys(to_query);
+	    var query_positions = query_genes.map(function(gene) {
+		return Object.keys(to_query[gene]).map(function(pos) { return parseInt(pos, 10); });
+	    });
+	    $.ajax({
```
adamabeshouse: The datamanager makes other ajax calls, for mutation annotation (oncokb, cosmic, ..), so I don't see this as out of scope of its responsibility.

service/src/main/java/org/cbioportal/service/impl/MutationCountServiceImpl.java
adamabeshouse: I don't think we're talking about a significant difference either way, but it's easier to manage and listen to a single ajax call on the frontend.

service/src/main/java/org/cbioportal/service/impl/MutationCountServiceImpl.java
adamabeshouse: good idea

service/src/test/java/org/cbioportal/service/impl/MutationCountServiceImplTest.java
adamabeshouse: good point

portal/src/main/webapp/js/api/cbioportal-datamanager.js
sheridancbio: ok

service/src/main/java/org/cbioportal/service/impl/MutationCountServiceImpl.java
sheridancbio: Ok .. if nobody else has comments then we can avoid the request latency of multiple calls by batching here. It would be interesting to do some actual field testing to compare the batch request to sequential asynchronous requests and see what runtime differences there are for a large gene set .. this general principle might help guide design of other APIs.

web/src/test/java/org/cbioportal/weblegacy/MutationCountControllerConfig.java
```diff
@@ -0,0 +1,64 @@
...
+ * @author abeshoua
+ */
+@Configuration
+@EnableWebMvc
+@ComponentScan(basePackages = {"org.cbioportal.weblegacy"}, resourcePattern = "**/*MutationCountController.class")
+public class MutationCountControllerConfig extends WebMvcConfigurerAdapter {
+    @Override
+    public void configureMessageConverters(List<HttpMessageConverter<?>> converters) {
+        MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter = new MappingJackson2HttpMessageConverter();
+        mappingJackson2HttpMessageConverter.setObjectMapper(new CustomObjectMapper());
```
sheridancbio: @ersinciftci pointed out in #1996 out that this hardwiring of ObjectMappers in test classes is not correct, since the weblegacy package controllers are not using the CustomObjectMapper configurations for the actual controller instances in production.  He would delete lines 54-59 here.

This issue will be addressed/fixed in a separate PR

web/src/test/java/org/cbioportal/weblegacy/MutationCountControllerTest.java
```diff
@@ -0,0 +1,105 @@
...
+	private MutationCountService mutationCountServiceMock;
+	private MockMvc mockMvc;
+	
+	private PositionMutationCount pmc1;
+	
+	@Before
+	public void setup() {
+		Mockito.reset(mutationCountServiceMock);
+		mockMvc = MockMvcBuilders.webAppContextSetup(webApplicationContext).build();
+		pmc1 = new PositionMutationCount();
```
sheridancbio: if more mock test data is added later, it would make sense to separate this construction of mock objects into a separate member function.

jjgao:
@adamabeshouse update on this pr?

adamabeshouse:
@jjgao now that we are annotating all mutations in the oncoprint, not just missense, maybe it would make most sense to make the queries here a subroutine of the mutation count endpoint, rather than a separate endpoint?

jjgao:
@adamabeshouse that makes sense. Do we still want to modify and merge this PR? @ersinciftci

ersinciftci:
@jjgao @adamabeshouse I already created a Mutation Count by Position endpoint in the new API a few weeks ago http://cbioportal-rc.herokuapp.com/api/swagger-ui.html#!/Mutations/fetchMutationCountsByPositionUsingPOST , is it the same?

adamabeshouse:
@ersinciftci the contribution of this PR is to make an optimized query when only counting at single positions. Does your API do that? If not maybe we could copy over the query from this PR and use it as a subroutine of yours.

jjgao:
@adamabeshouse should we close this?

adamabeshouse:
yes I suppose its completely obsolete at this point :D

PR #1936, opened 2016-11-28
Update screenshot after db migration

inodb:
The clinical attribute order on study view changed after applying the db migration. This updates the screenshot.


inodb:
@sheridancbio Agreed. I think both backend & frontend don't enforce a specific ordering

PR #1941, opened 2016-11-28
Change oncokb.url property to show.oncokb in property.EXAMPLE 

zhx828:
# What? Why?
Fix # 1940.

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@jjgao 
@cBioPortal/backend

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -665,7 +665,7 @@ public class GlobalProperties {
         if(showOncokb == null || showOncokb.isEmpty()) {
             showOncokb = "true";
         }
-        // This only applies if there is no oncokb.url property in the portal.properties file.
+        // This only applies if there is no oncokb.api.url property in the portal.properties file.
         // Empty string should be used if you want to disable the OncoKB annotation.
         if(oncokbApiUrl == null) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -123,6 +123,7 @@
     public static final String PATIENT_VIEW_DIGITAL_SLIDE_META_URL = "digitalslidearchive.meta.url";
     public static final String PATIENT_VIEW_TCGA_PATH_REPORT_URL = "tcga_path_report.url";
     public static final String ONCOKB_URL = "oncokb.url";
```
jjgao: maybe also rename it to oncokb.api.url so that we can reset it for the instances that have set a oncokb.url already.

web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
```diff
@@ -64,8 +68,24 @@
 
   private String oncokbURL;
   @Value("${oncokb.url:http://oncokb.org/legacy-api/}")
```
jjgao: same as above.

zhx828:
@jjgao I've made the changes you suggested above. Please take a look when you have time. Thank you.

Issue #1620, opened 2016-08-25
Markup errors in ReadtheDocs documentation


fedde-s:
The markup on https://cbioportal.readthedocs.io/en/hotfix/File-Formats.html was written for GitHub, and copied to ReadtheDocs without adjustments in pull request #1546. This has caused several syntax elements to show up verbatim in the document rather than being typeset:
- [x] HTML escape sequences in the explanation of the `data_filename` fields of most metadata files  
  ![screen shot 2016-08-25 at 12 22 42](https://cloud.githubusercontent.com/assets/4929431/17965814/c40a3266-6abe-11e6-97ca-1e38ad099660.png)
- [x] level-five headers under [Clinical Data → Data Files](https://cbioportal.readthedocs.io/en/hotfix/File-Formats.html#data-files)  
  ![screen shot 2016-08-25 at 12 24 17](https://cloud.githubusercontent.com/assets/4929431/17965835/ebf3867e-6abe-11e6-9536-6eee53611241.png)
- [x] a hyperlink in the description of patient attribute files
  ![screen shot 2016-08-25 at 12 46 36](https://cloud.githubusercontent.com/assets/4929431/17966489/17d06fde-6ac2-11e6-853f-700b79afe7c4.png)
- [ ] the table under [Expression Data → Supported stable_id values for MRNA_EXPRESSION](https://cbioportal.readthedocs.io/en/hotfix/File-Formats.html#supported-stable-id-values-for-mrna-expression)  
  ![screen shot 2016-08-25 at 12 25 40](https://cloud.githubusercontent.com/assets/4929431/17965875/1323e6bc-6abf-11e6-84e1-fd6681ca7a18.png)
- [x] the emoji escape sequences `:warning:`, `:information_source`, `:arrow_up:` and `:arrow_down:` in various places
  ![screen shot 2016-08-25 at 12 31 51](https://cloud.githubusercontent.com/assets/4929431/17966056/f349daee-6abf-11e6-8848-0595bfef4f46.png)
- [x] and finally, an internal link in the description of the `groups` metadatum of studies, which was probably already broken when we moved away from GitHub’s wiki feature  
  ![screen shot 2016-08-25 at 12 51 28](https://cloud.githubusercontent.com/assets/4929431/17966596/b2f0cc98-6ac2-11e6-85dd-1c63fc9ecf95.png)

fedde-s:
@inodb, could you look at this?

inodb:
[...]

PR #1944, opened 2016-11-29
Fix docs MRNA table

inodb:
Change markdown to HTML style table. Overlooked item in issue #1620.

Also fix UTF8 character in about_us page (Marie-José). Problem is that readthedocs does not work well with html5 entities, whereas showdown used in `about_us.jsp` does not work well with UTF8. Since the about_us.jsp is prolly used more often I picked the solution that makes that work. CC: @aderidder

### DIFF
--- a/docs/File-Formats.md
+++ b/docs/File-Formats.md
@@ -390,7 +390,7 @@ For historical reasons, cBioPortal expects the `stable_id` to be one of those li
 
 <table>
 <thead>
-<tr><th>datatype</th><th>stable_id</th><th>description</th></th>
+<tr><th>datatype</th><th>stable_id</th><th>description</th></tr>
 </thead>
 <tr><td>CONTINUOUS</td><td>mrna_U133</td><td>Affymetrix U133 Array</td></tr>
 <tr><td>Z-SCORE</td><td>mrna_U133_Zscores</td><td>Affymetrix U133 Array</td></tr>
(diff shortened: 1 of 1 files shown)


docs/File-Formats.md
```diff
@@ -391,23 +391,24 @@ The expression metadata file should contain the following fields:
...
-CONTINUOUS|mirna|MicroRNA data
-Z-SCORE|mirna_median_Zscores|MicroRNA data
-Z-SCORE|mrna_merged_median_Zscores|?
-CONTINUOUS|mrna|mRNA data
-DISCRETE|mrna_outliers|mRNA data of outliers
-Z-SCORE|mrna_zbynorm|?
-CONTINUOUS|rna_seq_mrna_capture|data from Roche mRNA Capture Kit
-Z-SCORE|rna_seq_mrna_capture_Zscores|data from Roche mRNA Capture Kit
+<table>
+<tr><td>datatype </td><td> stable_id </td><td> description</td></tr>
```
fedde-s: I would suggest wrapping this row in a `<thead>`, and using `<th>`s for the cells instead of `<td>`. Or is there a reason not to? I do not have a local installation if Sphinx to test how it will look on Readthedocs, but the CommoMark spec explicitly mentions support for the keywords `thead` and `th`.

sandertan:
Pieter helped me fix the UTF8 problem in https://github.com/cBioPortal/cbioportal/pull/2332/files  , which is in master since 1.5.1 I believe.

fedde-s:
So I understand that that would mean that we should reduce this to just the patch to fix the mRNA table, right @sandertan? It doesn't look like the other part still matches any relevant paragraph of `About-Us.md` anyway.

sandertan:
@fedde-s yes :)

jjgao:
@inodb I pointed it to hotfix. Could you rebase?

docs/File-Formats.md
inodb: Done

inodb:
@jjgao done

inodb:
@jjgao also updated gsva tables, because they weren't in html format either

PR #1948, opened 2016-11-29
Add 3 endpoints in Patients API

ersinciftci:
POST /patients/fetch
GET /studies/{studyId}/patients
GET /studies/{studyId}/patients/{patientId}

### DIFF
--- a/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
+++ b/persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
@@ -1,0 +1,177 @@
+package org.cbioportal.persistence.mybatis;
+
+import org.cbioportal.model.CancerStudy;
+import org.cbioportal.model.Patient;
+import org.cbioportal.model.meta.BaseMeta;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Configurable;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration("/testContextDatabase.xml")
+@Configurable
+public class PatientMyBatisRepositoryTest {
+    
+    @Autowired
+    private PatientMyBatisRepository patientMyBatisRepository;
+
+    @Test
+    public void getAllPatientsInStudyIdProjection() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "ID", null, null, null,
+            null);
+
+        Assert.assertEquals(14, result.size());
+        Patient patient = result.get(0);
+        Assert.assertEquals((Integer) 1, patient.getInternalId());
+        Assert.assertEquals("TCGA-A1-A0SB", patient.getStableId());
+        Assert.assertNull(patient.getCancerStudy());
+    }
+
+    @Test
+    public void getAllPatientsInStudySummaryProjection() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "SUMMARY", null, null,
+            null, null);
+
+        Assert.assertEquals(14, result.size());
+        Patient patient = result.get(0);
+        Assert.assertEquals((Integer) 1, patient.getInternalId());
+        Assert.assertEquals("TCGA-A1-A0SB", patient.getStableId());
+        Assert.assertEquals((Integer) 1, patient.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", patient.getCancerStudyIdentifier());
+        Assert.assertNull(patient.getCancerStudy());
+    }
+
+    @Test
+    public void getAllPatientsInStudyDetailedProjection() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "DETAILED", null, null,
+            null, null);
+
+        Assert.assertEquals(14, result.size());
+        Patient patient = result.get(0);
+        Assert.assertEquals((Integer) 1, patient.getInternalId());
+        Assert.assertEquals("TCGA-A1-A0SB", patient.getStableId());
+        Assert.assertEquals((Integer) 1, patient.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", patient.getCancerStudyIdentifier());
+        CancerStudy cancerStudy = patient.getCancerStudy();
+        Assert.assertEquals((Integer) 1, cancerStudy.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", cancerStudy.getCancerStudyIdentifier());
+        Assert.assertEquals("brca", cancerStudy.getTypeOfCancerId());
+        Assert.assertEquals("Breast Invasive Carcinoma (TCGA, Nature 2012)", cancerStudy.getName());
+        Assert.assertEquals("BRCA (TCGA)", cancerStudy.getShortName());
+        Assert.assertEquals("<a href=\\\"http://cancergenome.nih.gov/\\\">The Cancer Genome Atlas (TCGA)</a> Breast" +
+            " Invasive Carcinoma project. 825 cases.<br><i>Nature 2012.</i> <a href=\\\"http://tcga-data.nci." +
+            "nih.gov/tcga/\\\">Raw data via the TCGA Data Portal</a>.", cancerStudy.getDescription());
+        Assert.assertEquals(true, cancerStudy.getPublicStudy());
+        Assert.assertEquals("23000897", cancerStudy.getPmid());
+        Assert.assertEquals("TCGA, Nature 2012", cancerStudy.getCitation());
+        Assert.assertEquals("SU2C-PI3K;PUBLIC;GDAC", cancerStudy.getGroups());
+        Assert.assertEquals((Integer)0 , cancerStudy.getStatus());
+    }
+
+    @Test
+    public void getAllPatientsInStudySummaryProjection1PageSize() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "SUMMARY", 1, 0, null,
+            null);
+
+        Assert.assertEquals(1, result.size());
+    }
+
+    @Test
+    public void getAllPatientsInStudySummaryProjectionStableIdSort() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "SUMMARY", null, null,
+            "stableId", "ASC");
+
+        Assert.assertEquals(14, result.size());
+        Assert.assertEquals("TCGA-A1-A0SB", result.get(0).getStableId());
+        Assert.assertEquals("TCGA-A1-A0SD", result.get(1).getStableId());
+        Assert.assertEquals("TCGA-A1-A0SE", result.get(2).getStableId());
+        Assert.assertEquals("TCGA-A1-A0SF", result.get(3).getStableId());
+        Assert.assertEquals("TCGA-A1-A0SG", result.get(4).getStableId());
+        Assert.assertEquals("TCGA-A1-A0SH", result.get(5).getStableId());
+    }
+
+    @Test
+    public void getMetaPatientsInStudy() throws Exception {
+
+        BaseMeta result = patientMyBatisRepository.getMetaPatientsInStudy("study_tcga_pub");
+
+        Assert.assertEquals((Integer) 14, result.getTotalCount());
+    }
+
+    @Test
+    public void getPatientInStudyNullResult() throws Exception {
+
+        Patient result = patientMyBatisRepository.getPatientInStudy("study_tcga_pub", "invalid_patient");
+
+        Assert.assertNull(result);
+    }
+
+    @Test
+    public void getPatientInStudy() throws Exception {
+
+        Patient patient = patientMyBatisRepository.getPatientInStudy("study_tcga_pub", "TCGA-A1-A0SI");
+
+        Assert.assertEquals((Integer) 7, patient.getInternalId());
+        Assert.assertEquals("TCGA-A1-A0SI", patient.getStableId());
+        Assert.assertEquals((Integer) 1, patient.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", patient.getCancerStudyIdentifier());
+        CancerStudy cancerStudy = patient.getCancerStudy();
+        Assert.assertEquals((Integer) 1, cancerStudy.getCancerStudyId());
+        Assert.assertEquals("study_tcga_pub", cancerStudy.getCancerStudyIdentifier());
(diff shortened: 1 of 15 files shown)


persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,177 @@
...
+    public void fetchPatients() throws Exception {
+
+        List<String> studyIds = new ArrayList<>();
+        studyIds.add("study_tcga_pub");
+        studyIds.add("study_tcga_pub");
+        List<String> patientIds = new ArrayList<>();
+        patientIds.add("TCGA-A1-A0SB");
+        patientIds.add("TCGA-A1-A0SE");
+
+        List<Patient> result = patientMyBatisRepository.fetchPatients(studyIds, patientIds, "SUMMARY");
```
sheridancbio: This pairing of arguments by array index feels wrong .. but I guess doing it this way allows mapping to the same handler in the mybatis mapper, rather than having one handler for a list of (study,patient) pairing objects .. and a different handler for a simple list of study identifiers (patientIds=null).

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/PatientMapper.xml
```diff
@@ -19,4 +19,57 @@
...
+
+    <sql id="from">
+        FROM patient
+        INNER JOIN cancer_study ON patient.CANCER_STUDY_ID = cancer_study.CANCER_STUDY_ID
+    </sql>
+
+    <sql id="where">
+        <where>
+            <if test="patientIds == null">
+                cancer_study.CANCER_STUDY_IDENTIFIER = #{studyIds[0]}
```
sheridancbio: so .. without patientIds, we will only look at the first study identifier in the list?  Intuitively, if a caller to the repository passes in a list of cancer studies, wouldn't they expect the union of all patients from all of the studies passed in?

perhaps the repository should throw an exception when called with a list of studies, but no patient Ids. (or at least, we should put comments by this repository functions in the interface explaining that such a call will ignore all but the first study)

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/PatientMapper.xml
```diff
@@ -19,4 +19,57 @@
...
+        SELECT
+        <include refid="select">
+            <property name="prefix" value=""/>
+        </include>
+        <include refid="from"/>
+        <include refid="where"/>
+        <if test="sortBy != null and projection != 'ID'">
+            ORDER BY ${sortBy} ${direction}
+        </if>
+        <if test="projection == 'ID'">
```
sheridancbio: note to self: we do order by STABLE_ID on an ID projection here. (hardcode order ASC)

service/src/main/java/org/cbioportal/service/exception/PatientNotFoundException.java
```diff
@@ -2,13 +2,23 @@
 
 public class PatientNotFoundException extends Exception {
```
sheridancbio: this now functions as "Patient Not Found In Study" exception (rather than global search) .. or maybe it handles both cases.

web/src/main/java/org/cbioportal/web/PatientController.java
```diff
@@ -14,45 +21,86 @@
...
-    @RequestMapping(value = "/studies/{studyId}/patients/{patientId}", method = RequestMethod.GET)
+    @RequestMapping(value = "/studies/{studyId}/patients/{patientId}", method = RequestMethod.GET,
+        produces = MediaType.APPLICATION_JSON_VALUE)
     @ApiOperation("Get a patient in a study")
-    public ResponseEntity<Patient> getPatientInStudy(@PathVariable String studyId,
-                                                            @PathVariable String patientId) {
+    public ResponseEntity<Patient> getPatientInStudy(
+        @ApiParam(required = true, value = "Study ID e.g. acc_tcga")
+        @PathVariable String studyId,
+        @ApiParam(required = true, value = "Patient ID e.g. TCGA-OR-A5J2")
```
sheridancbio: Good guidance with examples

web/src/main/java/org/cbioportal/web/PatientController.java
```diff
@@ -14,45 +21,86 @@
...
+        @ApiParam(required = true, value = "List of patient identifiers")
+        @RequestBody List<PatientIdentifier> patientIdentifiers) throws PageSizeTooBigException {
 
-        throw new UnsupportedOperationException();
-    }
+        if (patientIdentifiers.size() > PagingConstants.MAX_PAGE_SIZE) {
+            throw new PageSizeTooBigException(patientIdentifiers.size());
+        }
+
+        List<String> studyIds = new ArrayList<>();
```
sheridancbio: argument processing inside controller : ok
the list construction code is reasonably short ; when the logic gets heavier I would start suggesting a util class

web/src/main/java/org/cbioportal/web/PatientController.java
```diff
@@ -14,45 +21,86 @@
...
+        List<String> patientIds = new ArrayList<>();
 
-    @RequestMapping(value = "/patients/query", method = RequestMethod.POST)
-    @ApiOperation("Query patients by example")
-    public ResponseEntity<List<Patient>> queryPatientsByExample(@RequestParam(defaultValue = "SUMMARY") Projection projection,
-                                                                                 @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_SIZE) Integer pageSize,
-                                                                                 @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_NUMBER) Integer pageNumber,
-                                                                                 @RequestBody Patient examplePatient) {
+        for (PatientIdentifier patientIdentifier : patientIdentifiers) {
```
sheridancbio: I see .. this is where the "parallel array" comes from .. the actual argument through the api *is* a pairing of (study,patient)

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,177 @@
...
+    private PatientMyBatisRepository patientMyBatisRepository;
+
+    @Test
+    public void getAllPatientsInStudyIdProjection() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "ID", null, null, null,
+            null);
+
+        Assert.assertEquals(14, result.size());
+        Patient patient = result.get(0);
```
ersinciftci: This was my reply https://github.com/cBioPortal/cbioportal/pull/1825#discussion_r86157639

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,177 @@
...
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "SUMMARY", 1, 0, null,
+            null);
+
+        Assert.assertEquals(1, result.size());
+    }
+
+    @Test
+    public void getAllPatientsInStudySummaryProjectionStableIdSort() throws Exception {
+
+        List<Patient> result = patientMyBatisRepository.getAllPatientsInStudy("study_tcga_pub", "SUMMARY", null, null,
```
ersinciftci: I'll add more test cases eventually (more negative and non-default scenarios).

web/src/main/java/org/cbioportal/web/parameter/sort/PatientSortBy.java
```diff
@@ -0,0 +1,7 @@
+package org.cbioportal.web.parameter.sort;
+
+public enum PatientSortBy {
+
+    stableId,
```
sheridancbio: I think it is convention to have enum constants be all upper case. But maybe for the framework to work correctly they must match our data member names.

persistence/persistence-mybatis-test/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
ersinciftci: Yes, it allows us to reuse the queries, and we also get the same pairings from the web layer from fetch endpoint, so makes things easier.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/PatientMapper.xml
ersinciftci: Yes it should throw an error, there should be lots of other error checks, but currently I implement it only for the web layer inputs, since persistence module it's not going to be used by other 3rd party software in a foreseeable future. So my priority is to release these basic APIs in a timely manner, instead of making them impossible to break for 3rd party use. So I say we better leave it for later as well.

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+    public void setUp() throws Exception {
+
+        Mockito.reset(patientService);
+        mockMvc = MockMvcBuilders.webAppContextSetup(wac).build();
+    }
+    
+    @Test
+    public void getAllPatientsInStudyDefaultProjection() throws Exception {
+
+        List<Patient> patientList = createExamplePatients();
```
sheridancbio: patientList has 2 patients in return value

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+        mockMvc.perform(MockMvcRequestBuilders.get("/studies/test_study_id/patients")
+            .param("projection", "META"))
+            .andExpect(MockMvcResultMatchers.status().isOk())
+            .andExpect(MockMvcResultMatchers.header().string(HeaderKeyConstants.TOTAL_COUNT, "2"));
+    }
+
+    @Test
+    public void getPatientInStudyNotFound() throws Exception {
+
+        Mockito.when(patientService.getPatientInStudy(Mockito.anyString(), Mockito.anyString())).thenThrow(
```
sheridancbio: test of global handler output format

service/src/main/java/org/cbioportal/service/exception/PatientNotFoundException.java
ersinciftci: In the current system, Patient id is only meaningful in a Study, so 2 different studies can have Patient id `15` referring to different patients. So I realized that a Patient id without Study id doesn't mean anything, so I added Study id in the exception. Because whenever we query a Patient from database, we do so with Study id.

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+        CancerStudy cancerStudy = new CancerStudy();
+        cancerStudy.setCancerStudyId(TEST_CANCER_STUDY_ID_1);
+        cancerStudy.setCancerStudyIdentifier(TEST_CANCER_STUDY_IDENTIFIER_1);
+        cancerStudy.setTypeOfCancerId(TEST_TYPE_OF_CANCER_ID_1);
+        cancerStudy.setName(TEST_NAME_1);
+        cancerStudy.setShortName(TEST_SHORT_NAME_1);
+        cancerStudy.setDescription(TEST_DESCRIPTION_1);
+        patient.setCancerStudy(cancerStudy);
+
+        Mockito.when(patientService.getPatientInStudy(Mockito.anyString(), Mockito.anyString())).thenReturn(patient);
```
sheridancbio: one patient return value .. construction could also be externalized to be side-by-side with the two patient constructor (encourage reuse) but it is fine here too

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+        mockMvc.perform(MockMvcRequestBuilders.get("/studies/test_study_id/patients/test_patient_id")
```
sheridancbio: single patient fetch test

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+    public void fetchPatientsDefaultProjection() throws Exception {
+
+        List<Patient> patientList = createExamplePatients();
+
+        Mockito.when(patientService.fetchPatients(Mockito.anyListOf(String.class), Mockito.anyListOf(String.class),
+            Mockito.anyString())).thenReturn(patientList);
+
+        List<PatientIdentifier> patientIdentifiers = new ArrayList<>();
+        PatientIdentifier patientIdentifier1 = new PatientIdentifier();
+        patientIdentifier1.setStudyId(TEST_CANCER_STUDY_IDENTIFIER_1);
```
sheridancbio: construction of the patientIdentifiers argument (study,patient) pairs.  Construction could also be externalized (previous comment)

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+        patientIdentifiers.add(patientIdentifier1);
+        PatientIdentifier patientIdentifier2 = new PatientIdentifier();
+        patientIdentifier2.setStudyId(TEST_CANCER_STUDY_IDENTIFIER_2);
+        patientIdentifier2.setPatientId(TEST_STABLE_ID_2);
+        patientIdentifiers.add(patientIdentifier2);
+
+        mockMvc.perform(MockMvcRequestBuilders.post("/patients/fetch")
+            .accept(MediaType.APPLICATION_JSON)
+            .contentType(MediaType.APPLICATION_JSON)
+            .content(objectMapper.writeValueAsString(patientIdentifiers)))
```
sheridancbio: pass of constructed parameter into POST request body

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+            .andExpect(MockMvcResultMatchers.status().isOk())
+            .andExpect(MockMvcResultMatchers.content().contentType(MediaType.APPLICATION_JSON))
+            .andExpect(MockMvcResultMatchers.jsonPath("$", Matchers.hasSize(2)))
+            .andExpect(MockMvcResultMatchers.jsonPath("$[0].internalId").doesNotExist())
```
sheridancbio: good .. checks that internal ids don't leak out later (@JsonIgnore)

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+    @Test
+    public void fetchPatientsMetaProjection() throws Exception {
+
+        BaseMeta baseMeta = new BaseMeta();
+        baseMeta.setTotalCount(2);
+
+        Mockito.when(patientService.fetchMetaPatients(Mockito.anyListOf(String.class),
+            Mockito.anyListOf(String.class))).thenReturn(baseMeta);
+
+        List<PatientIdentifier> patientIdentifiers = new ArrayList<>();
```
sheridancbio: here is reuse opportunity for construction code .. why not externalize it? (not necessary .. up to you)

ersinciftci:
I think it's ready to be merged.

PR #1965, opened 2016-12-02
Upgrade to maven 3.3.9 and Java 8

n1zea144:
Upgrade to maven 3.3.9 and Java 8

Travis configuration:

- update maven URL (previous mirror broke a lot)
- use maven version 3.3.9

Travis end-to-end tests:

- use docker image maven:3.3.9-jdk-8
- Update firefox/chrome screenshot of study view. Sorting of java 8 is slightly
  different, so the mutated genes on the study view have a different ordering.
  A separate issue is filed to change the sorting on frontend.


n1zea144:
@inodb Do we have to do something to get Java8 working with travis?

inodb:
@n1zea144 change `.travis.yml` to use `oraclejdk8` instead of `oraclejdk7`

fedde-s:
Ohh, great, once `rc` gets updated as well, we will be able to use Travis in our forks again, as has been impossible for a few months now! Thank you!

PR #1966, opened 2016-12-02
Fix travis db access denied

inodb:
Fix access denied error for docker mysql core db: 

https://travis-ci.org/cBioPortal/cbioportal/jobs/180756291#L360

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,21 +34,15 @@ install:
         fi
 
 before_script:
-    # run mysql db for core tests
+    # set up empty database for core test on Travis
     - |
         if [[ "${TEST}" == core ]]
         then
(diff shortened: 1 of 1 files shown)


inodb:
@lloydwatkin Works now. What I think what happened: the new infrastructure enables both mysql & docker service. This causes the docker container not to bind to the port anymore, because there is already a mysql db running. Changed it now to use the mysql on Travis instead (no more docker container).

lloydwatkin:
Cool, I haven't actually hit that yet in my builds (I'm using travis-ci.com though). Will keep in mind if things start failing, I guess it won't be a problem unless I specify the mysql service though?

PR #1968, opened 2016-12-02
Travis - quickfix run on deprecated infrastructure

inodb:
The MySQL upgrade on Travis causes a problem where the mysql client can't
connect to the docker mysql anymore. This change makes Travis run on the
deprecated infrastructure instead.


inodb:
Found a better solution: https://github.com/cBioPortal/cbioportal/pull/1966

PR #1972, opened 2016-12-06
Adding API implementation for GeneticData

pieterlukasse:
# What? Why?
This adds new endpoints and respective implementations for retrieving genetic data (aka genetic alteration data). It uses the new genetic entity data model introduced in #1902. 

- implementation of existing endpoint: `/api/genetic-profiles/{geneticProfileId}/genetic-data`
- created new endpoint with respective implementation: ` /genetic-profiles/{geneticProfileId}/fetch`

See also [proposal document](https://docs.google.com/document/d/12T0gTzArDdjxn74xcdXfdbxeeo_yDK6Kp24S9UWVFjs/edit). 

# Checks

# Notify reviewers
@ersinciftci @sheridancbio

### DIFF
--- a/model/src/main/java/org/cbioportal/model/GeneticEntity.java
+++ b/model/src/main/java/org/cbioportal/model/GeneticEntity.java
@@ -30,7 +30,7 @@ public abstract class GeneticEntity implements Serializable  {
 	public static enum EntityType
     {
         GENE,
-        GENE_SET,
+        GENESET,
         PHOSPHOPROTEIN;
     }
 	
--- a/model/src/main/java/org/cbioportal/model/GeneticEntity.java
+++ b/model/src/main/java/org/cbioportal/model/GeneticEntity.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2016 The Hyve B.V.
+ * This code is licensed under the GNU Affero General Public License (AGPL),
+ * version 3, or (at your option) any later version.
+ */
+
+/*
+ * This file is part of cBioPortal.
+ *
+ * cBioPortal is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package org.cbioportal.model;
+
+import java.io.Serializable;
+
+public class GeneticEntity implements Serializable  {
+
+	public static enum EntityType
+    {
+        GENE,
+        GENE_SET,
+        PHOSPHOPROTEIN;
+    }
+	
+	private Integer entityId;
+	private String entityStableId;
+    private EntityType entityType;
+
+    public Integer getEntityId() {
+        return entityId;
+    }
+
+    public void setEntityId(Integer entityId) {
+        this.entityId = entityId;
+    }
+    
+    public String getEntityStableId() {
+        return entityStableId;
+    }
+
+    public void setEntityStableId(String entityStableId) {
+        this.entityStableId = entityStableId;
+    }
+    
+    public EntityType getEntityType() {
+        return entityType;
+    }
+
+    public void setEntityType(EntityType entityType) {
+        this.entityType = entityType;
+    }
+    
+}
--- a/model/src/main/java/org/cbioportal/model/GeneticEntity.java
+++ b/model/src/main/java/org/cbioportal/model/GeneticEntity.java
@@ -25,7 +25,7 @@ package org.cbioportal.model;
 
 import java.io.Serializable;
 
-public class GeneticEntity implements Serializable  {
+public abstract class GeneticEntity implements Serializable  {
 
 	public static enum EntityType
     {
@@ -35,8 +35,6 @@ public class GeneticEntity implements Serializable  {
     }
 	
 	private Integer entityId;
-	private String entityStableId;
-    private EntityType entityType;
 
     public Integer getEntityId() {
         return entityId;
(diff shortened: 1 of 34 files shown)


persistence/persistence-mybatis-test/src/test/resources/testSql.sql
```diff
@@ -46,6 +46,7 @@ CREATE TABLE "type_of_cancer" (
 CREATE TABLE "gene" (
   "ENTREZ_GENE_ID" INTEGER NOT NULL,
   "HUGO_GENE_SYMBOL" VARCHAR(255) NOT NULL,
+  "GENETIC_ENTITY_ID" INTEGER, -- just a quick fix for GeneMapper.xml to work. TODO this part of the file should not exist...use cgds.sql instead. TODO 2 - add genetic_entity insert statements when making tests that involve this table
```
pieterlukasse: note: I think I know what line of cgds.sql is failing...we could give it one more try later...

ersinciftci:
I just realized one thing, if we use inheritance in the models, like GeneticEntity, client api that automatically generated by swagger.json won't be able to deserialize the response, because it will expect a GeneticEntity(abstract parent) object but it will be a Gene or GeneSet object. So we can remove the geneticEntity property from GeneticData model, since it's just an extra information and won't be used by frontend anyways. So you can delete GeneticEntity class entirely.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneticEntityRepository.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+package org.cbioportal.persistence;
+
+import org.cbioportal.model.GeneticEntity;
+import org.cbioportal.model.GeneticEntity.EntityType;
+
+public interface GeneticEntityRepository {
```
ersinciftci: Not sure why we need this.

web/src/main/java/org/cbioportal/web/GeneController.java
```diff
@@ -88,7 +88,7 @@
             @RequestBody List<String> geneIds) throws PageSizeTooBigException {
 
         if (geneIds.size() > PagingConstants.MAX_PAGE_SIZE) {
-            throw new PageSizeTooBigException(geneIds.size());
+            throw new PageSizeTooBigException(geneIds.size()); //TODO - is this correct? Missing API documentation on how one can specify page size parameter
```
ersinciftci: My initial API design didn't have any Paging and Sorting capabilities, but I added them to the APIs I implemented during implementation, so we need to add the same capabilities to this API as well. Sorry that I forgot to mention.

ersinciftci: My initial API design didn't have any Paging and Sorting capabilities, but I added them to the APIs I implemented during implementation, so we need to add the same capabilities to this API as well. Sorry that I forgot to mention.

model/src/main/java/org/cbioportal/model/GeneticEntity.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package org.cbioportal.model;
+
+import java.io.Serializable;
+
+public class GeneticEntity implements Serializable  {
```
ersinciftci: I think this should be an abstract class, since we shouldn't use it anywhere directly.

pieterlukasse: good idea. Done. getEntityStableId and getEntityType are now abstract methods.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneticEntityRepository.java
pieterlukasse: I agree that it should at least not be called *Mybatis because it does not interact with mybatis directly. I changed the name of the class and documented its purpose. We could also move it to the persistence package directly instead of having it in persistence-mybatis.

pieterlukasse:
Hi @ersinciftci thanks for the review so far. I made the changes based on your review and pushed this up as a separate commit. I also removed `GeneticEntity` from the webservice response.

pieterlukasse:
@ersinciftci I also added geneset support in my last commit. This helps clarify GeneticEntity and we need it now in our development of other features ;)

The result of a geneset query is this:
```json
[
  {
    "value": "0.949333333333333",
    "geneticProfileId": "brca_tcga_gsva_oncogenic_sets_scores",
    "geneticEntityId": "MORF_ATRX",
    "sampleId": "TCGA-A1-A0SD-01"
  },
  {
    "value": "0.279833333333333",
    "geneticProfileId": "brca_tcga_gsva_oncogenic_sets_scores",
    "geneticEntityId": "MORF_ATRX",
    "sampleId": "TCGA-A1-A0SE-01"
  },
```

I will probably add the geneset endpoints to this commit as well, to make that part complete...now that I added GenesetRepository, mybatis, etc already (and we need them as well)

ersinciftci:
Can you remove GeneticEntity from GeneticData model entirely, instead of JsonIgnore'ing in the mixin, because we won't ever need it, and when you do that, you won't need GeneticEntityRepository and GeneticEntityVirtualRepository classes, so you can remove them too. So after you delete them, it means we are not going to directly use GeneticEntity abstract class anywhere, so you don't need those 2 abstract methods in GeneticEntity either, so you can also delete EntityType enum in GeneticEntity as well.

sheridancbio:
#1902 has been merged. I am writing up proposal now for discussed changes to data model (including @ersinciftci 's comments above) and breaking endpoint into one specific to gene-set (GSVA) profiles, and one for gene (mRNA, methylation, ...) profiles.

pieterlukasse:
hi @sheridancbio I am back to this part now. Shall we resume the discussions regarding the API? Last time we had some questions for the group regarding the unsuitability of **entrez ids** in the API for some of the current "genes" we have (like phosphoproteins and microRNAs). The idea was to split up in specific endpoints for these as well, next to having a specific set of endpoints for gene sets.

pieterlukasse:
this PR will soon be replaced by a new one

pieterlukasse:
replaced by simplified PR #2158

PR #1990, opened 2016-12-13
Add 4 endpoints in Sample Lists API

ersinciftci:
GET /sample-lists
GET /sample-lists/{sampleListId}
GET /sample-lists/{sampleListId}/sample-ids
GET /studies/{studyId}/sample-lists

### DIFF
--- a/web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
+++ b/web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
@@ -4,6 +4,7 @@ import org.cbioportal.service.exception.CancerTypeNotFoundException;
 import org.cbioportal.service.exception.GeneNotFoundException;
 import org.cbioportal.service.exception.GeneticProfileNotFoundException;
 import org.cbioportal.service.exception.PatientNotFoundException;
+import org.cbioportal.service.exception.SampleListNotFoundException;
 import org.cbioportal.service.exception.SampleNotFoundException;
 import org.cbioportal.service.exception.StudyNotFoundException;
 import org.cbioportal.web.exception.PageNumberInvalidFormatException;
@@ -70,6 +71,13 @@ public class GlobalExceptionHandler {
 
         return new ResponseEntity<>(new ErrorResponse("Gene not found: " + ex.getGeneId()),
                 HttpStatus.NOT_FOUND);
+    }
+
+    @ExceptionHandler(SampleListNotFoundException.class)
+    public ResponseEntity<ErrorResponse> handleCaseListNotFound(SampleListNotFoundException ex) {
+
+        return new ResponseEntity<>(new ErrorResponse("Sample list not found: " + ex.getSampleListId()),
+            HttpStatus.NOT_FOUND);
     }
 
     @ExceptionHandler(MissingServletRequestParameterException.class)
--- a/web/src/main/java/org/cbioportal/web/mixin/SampleListMixin.java
+++ b/web/src/main/java/org/cbioportal/web/mixin/SampleListMixin.java
@@ -1,0 +1,21 @@
+package org.cbioportal.web.mixin;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.cbioportal.model.CancerStudy;
+
+public class SampleListMixin {
+
+    @JsonIgnore
+    private Integer listId;
+    @JsonProperty("sampleListId")
+    private String stableId;
+    private String category;
+    @JsonIgnore
+    private Integer cancerStudyId;
+    @JsonProperty("cancerStudyId")
+    private String cancerStudyIdentifier;
+    private CancerStudy cancerStudy;
+    private String name;
+    private String description;
+}
--- a/web/src/test/java/org/cbioportal/web/SampleListControllerTest.java
+++ b/web/src/test/java/org/cbioportal/web/SampleListControllerTest.java
@@ -1,0 +1,323 @@
+package org.cbioportal.web;
+
+import org.cbioportal.model.CancerStudy;
+import org.cbioportal.model.SampleList;
+import org.cbioportal.model.meta.BaseMeta;
+import org.cbioportal.service.SampleListService;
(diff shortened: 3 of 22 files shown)


sheridancbio:
review in progress

web/src/main/java/org/cbioportal/web/mixin/SampleListMixin.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.cbioportal.model.CancerStudy;
+
+public class SampleListMixin {
+
+    @JsonIgnore
+    private Integer listId;
+    @JsonProperty("sampleListId")
```
sheridancbio: renaming of "stableId" in accordance with recently adopted standard on API return type field names

web/src/main/java/org/cbioportal/web/mixin/SampleListMixin.java
```diff
@@ -0,0 +1,21 @@
...
+    private String stableId;
+    private String category;
+    @JsonIgnore
+    private Integer cancerStudyId;
+    @JsonProperty("cancerStudyId")
```
sheridancbio: This will be a little confusing until database schema are updated to reflect api field naming. cancerStudyId refers to the (internal) integer id in the database tables, but cancerStudyIdentifier to the stable name string in the API. Let's not forget to change the database fields too, as discussed during the evaluation of the standard.

web/src/test/java/org/cbioportal/web/SampleListControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+
+        Mockito.reset(sampleListService);
+        mockMvc = MockMvcBuilders.webAppContextSetup(wac).build();
+    }
+    
+    @Test
+    public void getAllSampleListsDefaultProjection() throws Exception {
```
sheridancbio: Add web layer test for getAllSampleListsDetailedProjection() and getAllSampleListsInStudyDetailedProjection()

web/src/test/java/org/cbioportal/web/SampleListControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+            .andExpect(MockMvcResultMatchers.status().isNotFound())
+            .andExpect(MockMvcResultMatchers.jsonPath("$.message")
+                .value("Sample list not found: test_sample_list_id"));
+    }
+
+    @Test
+    public void getSampleList() throws Exception {
+
+        SampleList sampleList = new SampleList();
+        sampleList.setListId(TEST_LIST_ID_1);
```
sheridancbio: I'm sorry for setting this pattern of manual code-based mock object construction. I'm pretty sure that there is a more compact / cleaner way of creating mock objects.

web/src/test/java/org/cbioportal/web/SampleListControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+                .value(TEST_CANCER_STUDY_IDENTIFIER_1))
+            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudy.description").value(TEST_STUDY_DESCRIPTION_1))
+            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudy.name").value(TEST_STUDY_NAME_1))
+            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudy.shortName").value(TEST_SHORT_NAME_1))
+            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudy.typeOfCancerId")
+                .value(TEST_TYPE_OF_CANCER_ID_1));
+    }
+
+    @Test
+    public void getAllSampleListsInStudyDefaultProjection() throws Exception {
```
sheridancbio: Add web layer test for getAllSampleListsDetailedProjection() and getAllSampleListsInStudyDetailedProjection()

sheridancbio:
Also, I did hand testing of the API through swagger and results looked good. (including detailed and summary projections of /studies/{studyId}/sample-lists and /sample-lists)

Queries of unknown list identifiers and study identifiers were properly reported with error messages and status codes in /sample-lists/{sampleListId},  however in /sample-lists/{sampleListId}/sample-ids, an unknown sampleListId returned status 200 and an empty body (I would have expected a 404.) Also in /studies/{studyId}/sample-lists an unknown studyId argument returned status 200 with an empty body. Maybe you can test these cases too and see if exception handling is set up for these cases.

web/src/test/java/org/cbioportal/web/SampleListControllerTest.java
ersinciftci: Actually this is how I normally do it too. I think manual code-based mocks are good, I haven't seen a cleaner way of doing this, this is quite clear.

ersinciftci:
It's the way I intended it to be, if we are getting single resource by id, it returns 404 if the resource is not found, if we are getting a list of resources, it returns empty list if resource is not found. We can discuss it, but it's the way I have been doing so far for other APIs as well.

ersinciftci:
Added requested tests.

sheridancbio:
OK .. I understand about the list return type and using null for not found values. I had only queries with a single study / listid, so was not thinking that way.

web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
```diff
@@ -72,6 +73,13 @@
                 HttpStatus.NOT_FOUND);
     }
 
+    @ExceptionHandler(SampleListNotFoundException.class)
+    public ResponseEntity<ErrorResponse> handleCaseListNotFound(SampleListNotFoundException ex) {
```
sheridancbio: should this function name have been "handleSampleListNotFoundException" ?  (sorry .. just noticed this now)

ersinciftci: Oh right, fixing it in the next PR, thanks

PR #1996, opened 2016-12-16
Make changes in some API response property names

ersinciftci:
Also removed some nested properties, which was causing lots of nested huge responses, and wasn't really useful. We can discuss this PR as a whole next Wednesday if required.

Also removed unannotated properties from mixins, as they have no effect on the output and it's unnecessary to keep models and mixins in sync for unannotated (means unchanged) properties.

All implemented APIs are impacted. You can compare the difference between swagger.json's using this PR's deployment http://cbioportal-pr-1996.herokuapp.com/api/v2/api-docs and rc deployment http://cbioportal-rc.herokuapp.com/api/v2/api-docs .


sheridancbio:
Since this PR changes the format / field names of the return type from the public API, I think we do need to approve these changes during a review meeting before merging to RC. Is there a summary of the changes? Let's try to put together a proposal showing what was changed and why. I can announce this for this week's meeting if we can get the details together today.

ersinciftci:
@sheridancbio We already approved most of the changes in last week's review meeting. The only ones we didn't discuss is the removal of nested `cancerStudy` properties from Patient and SampleList responses, and removal of `patient` from Sample response, because I found them overcomplicating the model doesn't provide much information since you query patients always with cancer study id, so all the patients in the response list will have the same cancer study details. You can query patients from different studies when using `/fetch`, but I'm sure before doing that, the API user already knew the details about cancer study anyway, since he choose specific studies to fetch from.

sheridancbio:
Last week's hangouts call was an informal discussion of the issues. We did not distribute a concrete proposal ahead of the hangout and did not give other's a chance to participate, so no formal decision was made. The process for modifying the API is to submit proposals, distribute them with enough time ahead of the meeting to let interested developers set aside time to participate in the meeting, and then to make the decision to Accept/Delegate/Reject during the Wednesday meeting. We can distribute a proposal now so that it can be formally accepted during the meeting on 12/21.

ersinciftci:
@sheridancbio Then it was 2 weeks ago, it was a formal discussion, because I remember I added the proposal to the document before the meeting.

sheridancbio:
Oh ... maybe I am mixed up. I'll check what was accepted on 12/7.  I thought we talked informally about the field names "stableXXX" and GeneticProfileId last week during the informal call.
Maybe I am not remembering properly.

sheridancbio:
@ersinciftci I see ... yes, we accepted name changes on 12/7  .. sorry for the confusion.

sheridancbio:
reviewing in detail now

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
```diff
@@ -51,7 +51,6 @@
 import org.cbioportal.model.Patient;
 import org.cbioportal.model.Sample;
 import org.cbioportal.model.SampleList;
-import org.cbioportal.model.StructuralVariant;
```
sheridancbio: The StructuralVariant mixin is removed from the CustomObjectMapper in the web package .. but it does not seem to be connected to any object mapper in weblegacy. (the mixin class is moved below). Should we have a separation of weblegacy controller mixins and web controller mixins?  (for example, why is CopyNumberSegmentMixin still connected here?)

We can go forward, but if we are relocating the mixins for legacy controllers, we should set that up soon.

web/src/main/java/org/cbioportal/web/parameter/sort/CancerTypeSortBy.java
```diff
@@ -2,10 +2,20 @@
...
-    shortName,
-    parent
+    cancerTypeId("typeOfCancerId"),
+    name("name"),
+    clinicalTrialKeywords("clinicalTrialKeywords"),
+    dedicatedColor("dedicatedColor"),
+    shortName("shortName"),
+    parent("parent");
+
+    private String originalValue;
```
sheridancbio: added level of encapsulation for sort-by keywords

web/src/main/java/org/cbioportal/weblegacy/mixin/StructuralVariantMixin.java
```diff
@@ -30,7 +30,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-package org.cbioportal.web.mixin;
+package org.cbioportal.weblegacy.mixin;
```
sheridancbio: mixin is relocated, but not registered with a CustomObjectMapper. We may need to configure an object mapper to be used with the weblegacy controllers

web/src/test/java/org/cbioportal/web/PatientControllerTest.java
```diff
@@ -145,18 +145,9 @@ public void getPatientInStudy() throws Exception {
             .andExpect(MockMvcResultMatchers.status().isOk())
             .andExpect(MockMvcResultMatchers.content().contentType(MediaType.APPLICATION_JSON))
             .andExpect(MockMvcResultMatchers.jsonPath("$.internalId").doesNotExist())
-            .andExpect(MockMvcResultMatchers.jsonPath("$.stableId").value(TEST_STABLE_ID_1))
-            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudyId").value(TEST_CANCER_STUDY_IDENTIFIER_1))
-            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudyIdentifier").doesNotExist())
-            .andExpect(MockMvcResultMatchers.jsonPath("$.cancerStudy.cancerStudyId").doesNotExist())
```
sheridancbio: embedded object now gone

web/src/test/java/org/cbioportal/weblegacy/ApiControllerConfig.java
```diff
@@ -61,12 +61,7 @@
 @EnableWebMvc
 @ComponentScan(basePackages = {"org.mskcc.cbio.portal.web.api", "org.mskcc.cbio.portal.persistence", "org.mskcc.cbio.portal.service"})
 public class ApiControllerConfig extends WebMvcConfigurerAdapter {
-    @Override
-    public void configureMessageConverters(List<HttpMessageConverter<?>> converters) {
-        MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter = new MappingJackson2HttpMessageConverter();
-        mappingJackson2HttpMessageConverter.setObjectMapper(new CustomObjectMapper());
```
sheridancbio: Remove attempt to link to a CustomObjectMapper ... not needed? And what about the connected mixins? Perhaps this is not needed for test code?

web/src/test/java/org/cbioportal/weblegacy/CNSegmentControllerConfig.java
```diff
@@ -47,12 +47,7 @@
 @EnableWebMvc
 @ComponentScan(basePackages = {"org.cbioportal.weblegacy"}, resourcePattern = "**/*CNSegmentController.class")
 public class CNSegmentControllerConfig extends WebMvcConfigurerAdapter {
-    @Override
-    public void configureMessageConverters(List<HttpMessageConverter<?>> converters) {
-        MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter = new MappingJackson2HttpMessageConverter();
-        mappingJackson2HttpMessageConverter.setObjectMapper(new CustomObjectMapper());
```
sheridancbio: Remove attempt to link to a CustomObjectMapper ... not needed? And what about the connected mixins? In this case there is a mixin defined for CopyNumberSegments configured through the web package. Perhaps this is not needed for test code?

web/src/test/java/org/cbioportal/weblegacy/GenePanelControllerConfig.java
```diff
@@ -52,12 +52,7 @@
 @EnableWebMvc
 @ComponentScan(basePackages = {"org.cbioportal.weblegacy"}, resourcePattern = "**/*GenePanelController.class")
 public class GenePanelControllerConfig extends WebMvcConfigurerAdapter {
-    @Override
-    public void configureMessageConverters(List<HttpMessageConverter<?>> converters) {
-        MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter = new MappingJackson2HttpMessageConverter();
-        mappingJackson2HttpMessageConverter.setObjectMapper(new CustomObjectMapper());
```
sheridancbio: Don't we need this?

ersinciftci:
@sheridancbio The reason of your confusion is, `CopyNumberSegmentMixin` my designed but not implemented Copy Number Segment API's mixin, not the one in legacy, none of the legacy API use mixin, that's why I removed CustomObjectMapper from the test config classes. If legacy APIs want to use a mixin, there should be another ObjectMapper configured for legacy APIs because I have things like `super.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);` in the CustomObjectMapper which would impact and break some of legacy APIs. I didn't create an ObjectMapper for legacy apis in the PR because as I said none of them uses mixins. There is a mixin class `StructuralVariantMixin.java` which is intended to be used with Structural Variant API, but not really used as I said a few weeks ago because CustomObjectMapper is tied only to `web` package. It works in `StructuralVariantControllerTest` because CustomObjectMapper is manually tied in the `StructuralVariantControllerTestConfig`. So instead of manually wiring beans in test config classes, web layer tests should use their respective Spring context xmls( `@ContextConfiguration("/applicationContext-web.xml")` for new APIs, `@ContextConfiguration("/applicationContext-weblegacy.xml")` for legacy ones) so they work exactly how they would work in production.

sheridancbio:
Ok. I think we do need to set up a CustomObjectMapper for legacy controllers, and use mixins. But that should be in a separate PR. I'm surprised this has not caused problems yet; I think legacy would be sensitive to some of the Mapper configurations (I think CNSegmentController cared about number to string conversion or something similar) but it will be good to decouple the mapper settings for web from those of weblegacy in any case.

PR #1998, opened 2016-12-20
Clustering functionality for oncoprint heatmap

pieterlukasse:
# What? Why?
This PR adds hierarchical clustering feature to the oncoprint heatmaps. 

- New "cluster heatmap" button in oncoprint "Heatmap" menu
- New javascript clustering library. Clustering is done in 2 Worker threads in the background. 

# Notify reviewers
@adamabeshouse @jjgao


pieterlukasse:
Screenshot of how it looks like: 
image

portal/src/main/webapp/js/src/oncoprint/webgl/setup.js
```diff
@@ -1734,6 +1734,46 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
...
+		    evt.preventDefault();
+		    evt.stopPropagation();
+
+		    // cluster heatmap for selected profile:
+		    var genetic_profile_id = $(toolbar_selector + ' #oncoprint_diagram_heatmap_menu #oncoprint_diagram_heatmap_profiles').val();
+		    if (genetic_profile_id) {
+				//sort according to order found in the clustering results: 
+				State.sorting_by_given_order = true;
+		    	var grp = State.heatmap_track_groups[genetic_profile_id];
+				var heatmap_track_group_id = grp.track_group_id;
```
adamabeshouse: what you're calling `track_uid_map` and computing in `getTrackUIDMap` is actually available identically as `grp.gene_to_track_id`, as far as I can tell.

adamabeshouse: what you're calling `track_uid_map` and computing in `getTrackUIDMap` is actually available identically as `grp.gene_to_track_id`, as far as I can tell.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1247,6 +1246,25 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	    });
 	    return def.promise();
 	},
+	'getTrackUIDMap': function(groupId, trackDataType) {
```
adamabeshouse: See my comment below, this method is unnecessary.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1437,6 +1455,49 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	'getPatientHeatmapData': function(genetic_profile_id, genes) {
 	    return getHeatmapDataCached(this, genetic_profile_id, genes, 'patient');
 	},
+	'getClusteringOrder': function (case_ui_map, track_uid_map, heatmapData, case_ids) {
```
adamabeshouse: Since you're returning a promise, I'd typically fetch `case_ui_map` again from `this.getCaseUIDMap` instead of passing it in.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1437,6 +1455,49 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
+					if (!entityId) {
+						throw new Error("Unexpected error during getClusteringOrder: attribute 'gene' not found");
+					}
+					var value = heatmapData[j].oncoprint_data[i].profile_data;
+					cluster_input[case_ids[i]][entityId] = value;
+				}					
+			}
+			//do hierarchical clustering in background:
+			$.when(cbio.stat.hclusterCases(cluster_input), cbio.stat.hclusterGeneticEntities(cluster_input)).then(
+					function (sampleClusterOrder, entityClusterOrder) {
```
adamabeshouse: I just want to confirm entity is the best choice of word here, because it seems pretty vague to me. Why not "gene"?

portal/src/main/webapp/js/src/oncoprint/webgl/oncoprint-bundle.js
```diff
@@ -6476,7 +6476,6 @@ var OncoprintWebGLCellView = (function () {
 	renderAllTracks(this, model);
     }
     OncoprintWebGLCellView.prototype.setTrackGroupOrder = function(model) {
-	clearZoneBuffers(this, model);
```
adamabeshouse: sorry about this, I thought I removed it, maybe it got mixed up in a rebase or something.

portal/src/main/webapp/js/src/oncoprint/webgl/setup.js
```diff
@@ -1734,6 +1734,46 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
...
+									});
+							//show progress bar:
+							LoadingBar.show();
+							LoadingBar.update(0.2);
+						    LoadingBar.msg("Clustering...");
+						    //update progress bar every 0.2s:
+						    window.setInterval(function() {
+						    	var d = new Date();
+						    	var n = d.getMilliseconds();
+							    LoadingBar.update(n/1000);
```
adamabeshouse: This is a nitpick but I'm not crazy about this because normally the loading bar size actually signifies progress percentage. If you're going to do this, I'd rather show a different color than green.

jjgao:
@pieterlukasse maybe the "Cluster heatmap" button should be checkbox so that you can turn it on and off.

When there are multiple heatmaps, which one is used to cluster?

jjgao:
@pieterlukasse @adamabeshouse I'd love to release it this Thursday or Friday. Do you think that's doable? If so, please create a pr to release-1.4.0 branch after resolve the comments above.

jjgao:
@pieterlukasse I realized that you would have to rebase to release-1.4.0. Don't worry if it not easy to do. We can release in the next round.

pieterlukasse:
other comment I received: do clustering using 1-spearman's correlation for distance measure instead of 
1-pearson's correlation.

pieterlukasse:
...and sort the oncoprint rows in the same order as they appear in the cluster. This could be a harder one since the tracks in both parts don't need to be the same.

pieterlukasse:
Hi @jjgao, @adamabeshouse , thanks for the review. Regarding the checkbox @jjgao proposed, how about this:
image

And regarding @adamabeshouse 's comment about the missing "interface for switching back to regular genetic sorting after you've clustered the heatmap", what about having an extra (read only/disabled) radio button in the sort menu that gets selected when clustering is selected? Something like this:

image
 
Finally, answering @jjgao 's question regarding which heatmap gets clustered: it is the one currently selected in the heatmap drop down shown below:

image

pieterlukasse:
Here are the 1-spearman *vs* 1-pearson (as distance metric) results (for a query with 30 genes on brca study). 

Pearson: takes around **2 seconds:**

image

Spearman: takes around **10 seconds:**

image

@jjgao given the performance difference, do you think it is still worthwhile using spearman? Or should this be an option for the user (like a radio button to let him choose)?

portal/src/main/webapp/js/src/oncoprint/webgl/setup.js
pieterlukasse: done. using yellow now

portal/src/main/webapp/js/api/cbioportal-datamanager.js
pieterlukasse: long story. But here it makes sense as we will have heatmaps for other non-gene entities (like proteins or genesets). To keep this code generic, I chose to use a generic naming here.

portal/src/main/webapp/js/src/oncoprint/webgl/setup.js
pieterlukasse: you are right. I now removed  getTrackUIDMap

portal/src/main/webapp/js/api/cbioportal-datamanager.js
pieterlukasse: you are right. I now removed getTrackUIDMap

portal/src/main/webapp/js/api/cbioportal-datamanager.js
pieterlukasse: done

pieterlukasse:
@jjgao @adamabeshouse  I have finished the implementation of according to the design shared in previous comments. I also updated the code based on @adamabeshouse's comments. It is fully tested :thumbsup:  (I discovered that there were quite some scenarios where the checkbox should actually be disabled). Please check it again.

Clustering distance measure has also been changed to 1-spearman. It is slower now. Not sure if it is better. Results are shared in some of the previous comments.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
adamabeshouse: cool

jjgao:
@pieterlukasse good idea to make spearman an option -- by default, it is still pearson for performance reason.

There is a small bug: After you check `Cluster Heatmap` and go to `Sort` and click `Sort by Data`, `Cluster Heatmap` is not unchecked.

jan-18-2017 16-11-22

adamabeshouse:
I also see the bug JJ mentions above.

I think it would also be good to use the word 'heatmap' in the Sort menu. Like "Heatmap clustering". That would make it more explicit where to go to explore that option further.

pieterlukasse:
@adamabeshouse @jjgao thanks for testing. It was actually designed to stay checked since the sorting option only changes the column sorting...so the rows would remain clustered until the checkbox was unchecked. Anyway, I agree your solution is less confusing, so I changed the code to reset also the rows/tracks clustering if any other sorting option is selected. 

I also changed the menu option name in the sort menu, as @adamabeshouse proposed.

Finally, I restored 1-Pearson as the default distance metric. Spearman remains in the code. If we find enough interest in having this as an option, we could implement it in the future (so I would like to leave the option out for now). Hopefully we can merge now :)

adamabeshouse:
Looks good to me!

What do you think should happen when it's clustered by heatmap and then we remove the heatmap? Currently it stays as clustered by heatmap, but I wonder if it should switch automatically back to sort by data.

pieterlukasse:
@adamabeshouse it would indeed be more consistent to switch back to sort by data. Let me do that. I have also discussed the spearman option again with the client and this is high prio. I think I can find a way to speed it up (currently the spearman ranking step is repeated too often, I could improve that).

adamabeshouse:
@pieterlukasse I was looking through the spearman code and I noticed that the jStat calculation includes converting the input arrays to ranks. It might give considerable speedup to perform the conversion of all the vectors to ranks up front, and make your own routine which skips that step in the calculation. This brings the number of vector-to-rankvector conversions down from n^2 to n, which is extremely significant when you're talking about a lot of samples.

pieterlukasse:
thanks @adamabeshouse ! That is what I was planning to do...and then do pearson correlation on the ranks (according to what I understand from the theory...although the jstat implementation seems to indicate otherwise)

pieterlukasse:
clustering using spearman based distance metric now takes ~1s ! :+1: 

image

adamabeshouse:
@pieterlukasse nice!! Since this is something which is immediately useful with huge scale, I think it's definitely worth the effort to see just how fast you can make it by precomputing the vector deviations too, and pasting the rest of the math from the jStat code.

Where the jStat spearman code is:
``
jStat.spearmancoeff =  function (arr1, arr2) {
  arr1 = jStat.rank(arr1);
  arr2 = jStat.rank(arr2);
  var arr1dev = jStat.deviation(arr1);
  var arr2dev = jStat.deviation(arr2);
  return jStat.sum(arr1dev.map(function (x, i) {
    return x * arr2dev[i];
  })) /
  Math.sqrt(jStat.sum(arr1dev.map(function (x) {
    return Math.pow(x, 2);
    })) * jStat.sum(arr2dev.map(function (x) {
      return Math.pow(x, 2);
  }))
  );
}
``

the optimized would be

``
var spearmanoptimized =  function (arr1, arr2) {
  var arr1dev = arr1.dev;
  var arr2dev = arr2.dev;
  return jStat.sum(arr1dev.map(function (x, i) {
    return x * arr2dev[i];
  })) /
  Math.sqrt(jStat.sum(arr1dev.map(function (x) {
    return Math.pow(x, 2);
    })) * jStat.sum(arr2dev.map(function (x) {
      return Math.pow(x, 2);
  }))
  );
}
``

What do you think?

pieterlukasse:
@adamabeshouse that is what I did first, but it was still not fast enough. Then I simplified the code even further, as discussed in https://github.com/jstat/jstat/issues/200. This resulted in the excellent performance we have now!

adamabeshouse:
@pieterlukasse that is very surprising to me, since it looks to me like corrcoeff does essentially the same thing. Are you 100% sure of this?

pieterlukasse:
@adamabeshouse yes, it is somewhere in the small details I think. We also quickly looked and found that the pearson implementation had a simple while loop at one of the functions it called while the other code uses `map` function...maybe `map` is not so efficient? 

Anyway, shall we merge now?

adamabeshouse:
@pieterlukasse go ahead. Let me know if you need help resolving a rebasing conflict.

pieterlukasse:
@adamabeshouse rebased and merged. Resolved conflicts in 
```
both modified:   portal/src/main/webapp/js/api/cbioportal-datamanager.js
both modified:   portal/src/main/webapp/js/src/oncoprint/setup.js
```

jjgao:
@pieterlukasse: one last comment, when there is no heatmap, is it possible not to display "Sorted by heatmap clustering order" in sort menu? It's confusing especially when there are no expression profiles in a study.

portal/src/main/webapp/js/src/clustering/clustering-worker.js
```diff
@@ -0,0 +1,173 @@
...
+	var r = jStat.corrcoeff(ranks1, ranks2); 
+	if (isNaN(r)) {
+		r = 0; //will result in same distance as no correlation //TODO - calculate correlation only on items where there is data...?
+	}
+	return 1 - r;
+}
+
+/**
+ * Prepares the data for using spearman method in the distance function. 
+ * It will pre-calculate ranks and deviation and store this in inputItems[x].preProcessedValueList.
```
pieterlukasse: actually it is only ranks now. Need to update this comment

PR #2013, opened 2016-12-27
Add 2 endpoints in Mutations API

ersinciftci:
GET /genetic-profiles/{geneticProfileId}/mutations
POST /genetic-profiles/{geneticProfileId}/mutations/fetch

Also made some other changes:
*Added `-parameter` Java 8 compiler arg to preserve parameter names in compiled classes
*Removed `@Param`s from MyBatis mapper classes because parameter names are preserved now
*Changed controller parameter validations to JSR-303 annotations
*Moved mutation related methods that was created in new layered structure but related to old
frontend to `business` module
*Moved `api` context's Swagger configuration to the `web` package
*Moved MyBatis tests to `persistence-mybatis` module and deleted `persistence-mybatis-test`
module as these tests are fast and we always run them, so there is no point having a
separate module


ersinciftci:
[Finalize Mutations API Implementation](https://trello.com/c/SAqXZ1eZ/23-finalize-mutations-api-implementation)

business/src/main/java/org/mskcc/cbio/portal/model/Mutation.java
```diff
@@ -0,0 +1,349 @@
...
+
+package org.mskcc.cbio.portal.model;
+
+import org.cbioportal.model.Gene;
+import org.cbioportal.model.GeneticProfile;
+import org.cbioportal.model.Sample;
+
+import java.io.Serializable;
+
+public class Mutation implements Serializable {
```
ecerami: @onursumer @ersinciftci:  what makes this legacy?  isn't this the latest mutation model?

ecerami: oops, adding this here instead.  what makes this legacy?  isn't this the latest mutation model?

onursumer: @ecerami : I think this is the latest: https://github.com/cBioPortal/cbioportal/pull/2013/files#diff-73a035faa8446bd537e6c061008a3246

ersinciftci: @ecerami it's not, it's the one that has mutation event and all the other properties, I didn't delete it because if you remember in the first 2 sprints of first rearchitecturing effort with Ben and Rob, we changed mutation related `Dao`s to use MyBatis, so this model is used by those `Dao`s, so I moved it to a legacy module. @onursumer since everything in `business` module is legacy, I didn't change the name, all those name changes adding `Legacy` to the class name is because MyBatis throws error if there are multiple mappers in the system with the same name, but this class doesn't have that problem.

PR #2020, opened 2016-12-28
Oncoprint and survival curve optimizations and an oncoprint bug fix

adamabeshouse:
Speed up query loading
(1) Optimization to survival curve calculation
(2) Query data by sample list when possible to reduce cache accesses
(3) Optimization in getOncoprintXGenomicEventData to run through data
only once instead of a few times
(4) Fix off-by-one error in oncoprint zoom to fit
(5) No longer deepCopyObject in QuerySession - onus is now on caller
to make a copy if modifications to be made (currently not an issue).

Speed up oncoprint and reduce memory usage
(1) Cache shape objects and return references to them
instead of saving new computed shape object every time.
(2) Cache computed webGL vertexes for a shape instead of
recomputing each time.
(3) Use position bit packing and referencing color from texture for about a 3.5-fold savings in webGL data memory usage!

UI polish
(1) Fix track highlight behavior on mouseout of cell area

# Effects
On 3 gene MSKIMPACT query:
The current MSK portal loads after 16 seconds
The new implementation loads after 6 seconds

On 55 gene MSKIMPACT query:
The current MSK portal hangs for 1:30 then crashes
The new implementation loads successfully after 1:12

There's more to be done but this is a good start.

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@jjgao @inodb 
@cBioPortal/frontend

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
@@ -171,18 +171,8 @@
 //           $('.dropdown-menu #select_clinical_attributes').click(function(){$('#clinical_dropdown').dropdown('toggle');});
         </script>
 
-        <!--<script type="text/javascript" charset="utf-8" src="js/src/oncoprint/new/events.js?<%=GlobalProperties.getAppVersion()%>"></script>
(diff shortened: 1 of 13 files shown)


jjgao:
Great work! Thank, @adamabeshouse!

PR #2032, opened 2017-01-05
Add 1 endpoint in Significantly Mutated Genes API

ersinciftci:
GET /studies/{studyId}/significantly-mutated-genes

Also changed:
*Persistence tests now are using cgds.sql, so they always run with up to date schema
In order to do that, I had to break the circular dependency between business,core,
service,persistence modules, so I moved Gene Panel related classes to business module,
because they are used by classes in the core module.
*Added a new `internal` group for swagger-ui, so Significantly Mutated Genes API is not
visible in the default API documantation page, you can see it by changing the combobox
on top right corner in swagger-ui to `internal`
*Changed the type of `SHOW_PROFILE_IN_ANALYSIS_TAB` column in `genetic_profile` table to `BOOLEAN` in cgds.sql, increased the minor version and added migration script.


ersinciftci:
[Implement MutSig API](https://trello.com/c/uejdKzGX/2-implement-mutsig-api)

core/src/main/resources/db/migration.sql
```diff
@@ -270,3 +270,7 @@ ALTER TABLE `genetic_entity` DROP COLUMN `TMP_GENE_ID`;
 -- ========================== end of genetic_entity related migration =============================================
 
 UPDATE info SET DB_SCHEMA_VERSION="2.0.0";
```
onursumer: Is there a reason why we still keep this line?

web/src/main/java/org/cbioportal/web/mixin/MutSigMixin.java
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class MutSigMixin {
+
+    @JsonIgnore
+    private Integer cancerStudyId;
+    @JsonProperty("studyId")
+    private String cancerStudyIdentifier;
+    @JsonIgnore
+    private Integer numbasescovered;
```
onursumer: Why do we use all lowercase variable names here?

ersinciftci: Because in the database the column name doesn't use underscore between words for this column, and I prefer to keep model property names in a way that it would match with database columns automatically if we use an ORM tool in the future.

core/src/main/resources/db/migration.sql
ersinciftci: because every migration in this file is independent from each other, so every block should end with setting db schema to respective version.

PR #2037, opened 2017-01-06
Restore zoom slider and dont save minimap open state

adamabeshouse:
Restore zoom slider controls
Always hide minimap by default
Hide unsightly whitespace while oncoprint loads

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@schultzn @jjgao


jjgao:
Now the menu gets really long, it moves the oncoprint up and down. It also mess up with legend. Maybe we should not hide the menu any more. @schultzn

schultzn:
It was always useful for quick screenshots to have the menu bar hidden.

Why do we use “Download” instead of the download icon we use elsewhere?

jjgao:
@schultzn We changed everything to text for this menu. We could put some icon, e.g. download, to save some space, but it'll still be a problem when the screen is small.

@adamabeshouse in the current code, there is some space between "Altered in..." line and OncoPrint. Could you bring the space back? Ideally, show/hiding menu should not move the OncoPrint.

adamabeshouse:
@jjgao @schultzn I think someone asked me once to put "Altered cases" close to the oncoprint - maybe I could put the toolbar above, and try to position it absolute so it doesnt shift things up and down?

schultzn:
@adamabeshouse I think that is a good compromise.

PR #2038, opened 2017-01-06
Don't show oncokb in end-to-end test

inodb:
Since we have regular updates to oncokb annotations we prefer not to test this with the screenshot test.


jjgao:
@inodb could you do it on hotfix instead? We'll merge hotfix to rc.

PR #2057, opened 2017-01-12
Add 1 endpoint in mRNA Percentile API

ersinciftci:
POST /genetic-profiles/{geneticProfileId}/mrna-percentile/fetch

Also changed:
Added `alias` optional parameter to `GET /genes` for gene suggestion functionality


ersinciftci:
[Implement mRNAExpressionRank API](https://trello.com/c/0rNkdqpx/4-implement-mrnaexpressionrank-api)

model/src/main/java/org/cbioportal/model/MrnaPercentile.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public BigDecimal getPercentile() {
+        return percentile;
+    }
+
+    public void setPercentile(BigDecimal percentile) {
+        this.percentile = percentile;
+    }
+
+    public BigDecimal getzScore() {
```
jjgao: In the future, we may remove all zscores profiles in the database and implement frontend zscore calculation. Maybe for the mRNA percentile feature, it's OK not to send back zScores. The percentile is probably enough.

@ecerami are you OK if we remove the `mRNA z-score` line below and also remove it from the api?

image

ecerami: @jjgao:  I vote to keep it for now, e.g. to make sure we have enough info to replicate the current front end.

web/src/main/java/org/cbioportal/web/MrnaPercentileController.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @RequestMapping(value = "/genetic-profiles/{geneticProfileId}/mrna-percentile/fetch", method = RequestMethod.POST,
+        produces = MediaType.APPLICATION_JSON_VALUE)
+    @ApiOperation("Get mutations in a genetic profile")
+    public ResponseEntity<List<MrnaPercentile>> getMrnaPercentile(
+        @ApiParam(required = true, value = "Genetic Profile ID e.g. acc_tcga_mutations")
+        @PathVariable String geneticProfileId,
+        @ApiParam(required = true, value = "Sample ID e.g. TCGA-OR-A5J2-01")
+        @RequestParam String sampleId,
+        @ApiParam(required = true, value = "List of Entrez Gene IDs")
```
pieterlukasse: in our portal we have some micro RNA's that do not have a stable entrez id (but gets auto generated in the backend - see `ImportMicroRNAIDs.java` calling `CanonicalGene(geneSymbol,aliases)`). This API will therefore not support them?  And will this support then be part of another endpoint?

web/src/main/java/org/cbioportal/web/MrnaPercentileController.java
```diff
@@ -0,0 +1,52 @@
...
+        @Size(min = 1, max = PagingConstants.MAX_PAGE_SIZE)
```
pieterlukasse: why only the first 1000 records and no paging options? Samples can have expression data for more than 1000 genes, right? Or will the service expect the UI to send the genes in batches and give an error if the list is larger than 1000?

ersinciftci: We noticed this yesterday, will be changed to 50000 in the next PR.

web/src/main/java/org/cbioportal/web/MrnaPercentileController.java
ersinciftci: I return whatever is in the database, I don't check if it's negative, or not stable.

pieterlukasse: ok, so in this case only the semantics are wrong (as negative values are *not* entrez ids)

web/src/main/java/org/cbioportal/web/MrnaPercentileController.java
```diff
@@ -0,0 +1,52 @@
...
+@Validated
+@Api(tags = "mRNA Percentile", description = " ")
+public class MrnaPercentileController {
+
+    @Autowired
+    private MrnaPercentileService mrnaPercentileService;
+
+    @RequestMapping(value = "/genetic-profiles/{geneticProfileId}/mrna-percentile/fetch", method = RequestMethod.POST,
+        produces = MediaType.APPLICATION_JSON_VALUE)
+    @ApiOperation("Get mutations in a genetic profile")
```
pieterlukasse: just noticed this description is wrong

PR #2062, opened 2017-01-12
Add 1 endpoint in Copy Number Segments API

ersinciftci:
GET /studies/{studyId}/samples/{sampleId}/copy-number-segments


ersinciftci:
[Implement CNA Segments API [API Spec Ready]](https://trello.com/c/JOqqaQjm/48-implement-cna-segments-api-api-spec-ready)

model/src/main/java/org/cbioportal/model/CopyNumberSeg.java
```diff
@@ -1,17 +1,20 @@
 package org.cbioportal.model;
 
 import java.io.Serializable;
+import java.math.BigDecimal;
 
-public class CopyNumberSegment implements Serializable {
+public class CopyNumberSeg implements Serializable {
```
onursumer: Why do we need to rename this class?

ersinciftci: As I said in a previous PR, I'm trying to match model names to database table names just in case we use an ORM library in the future.

PR #2087, opened 2017-01-22
Add 1 endpoint in Copy Number Segments API

ersinciftci:
POST /copy-number-segments/fetch


ersinciftci:
[CNA segments API for multiple samples](https://trello.com/c/wtxxu2Gz/70-cna-segments-api-for-multiple-samples)

PR #2099, opened 2017-01-25
Added validation to block commas and other special characters in sample id

pieterlukasse:
# What? Why?

Added validation to block commas and some other special characters in sample id (`,;+/=*`) since these can cause problems in the portal. 

E.g. commas in sample ids give problems in old parts of the java code such as CnaJSON.processCnaFractionsRequest(). The other characters have been verified to give problems in Clinical tab. The `;` was also added as a precaution to this list of disallowed characters.

- small new check on sample id in clinical file

# Notify reviewers
@zheins

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1700,6 +1700,14 @@ class SampleClinicalValidator(ClinicalValidator):
                         extra={'line_number': self.line_number,
                                'column_number': col_index + 1,
                                'cause': value})
+                # this one gives problems in old parts of the java code such as CnaJSON.processCnaFractionsRequest(),
+                # so block commas in sample id:
+                if ',' in value:
+                    self.logger.error(
+                        'Comma (,) in SAMPLE_ID is not supported',
+                        extra={'line_number': self.line_number,
+                               'column_number': col_index + 1,
+                               'cause': value})
                 if value in self.sample_id_lines:
                     if value.startswith('TCGA-'):
                         self.logger.warning(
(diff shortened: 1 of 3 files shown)


pieterlukasse:
according to @pulyakhina the + and / also seem to be a problem...maybe we should add this here as well...?

pieterlukasse:
@sandertan when reviewing, can you test + and / to see if these indeed give problems (and where)?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1700,6 +1700,14 @@ def checkLine(self, data):
...
                                'column_number': col_index + 1,
                                'cause': value})
+                # this one gives problems in old parts of the java code such as CnaJSON.processCnaFractionsRequest(),
+                # so block commas in sample id:
+                if ',' in value:
+                    self.logger.error(
+                        'Comma (,) in SAMPLE_ID is not supported',
+                        extra={'line_number': self.line_number,
+                               'column_number': col_index + 1,
+                               'cause': value})
```
sandertan: I tested this and it works well. I've tested some other symbols, `+ / = *` and the `/ = *` hide the whole sample in the Clinical Data tab. The sample with `+` was displayed correctly without the `+`.

I didn't experience issues with the CNA table.

sandertan: We might want to block more symbols in this column. Also, many symbols in the attribute name field will cause issues.

pieterlukasse: updated the code to block these other characters as well

sandertan:
Thanks, changes are working correctly!

screen shot 2017-05-26 at 17 52 56

jjgao:
@zheins @ritikakundra do you know if we have any sample ids with characters `,;+/=*`.

PR #2102, opened 2017-01-27
Define species and genome version in portal.properties

oplantalech:
Here, 'species', 'genome build' and 'NCBI build' are established as global variables (in portal.properties). The default values from these variables are the ones cBioPortal currently uses (from human). However, when these variables refer to mouse, mouse studies can be loaded (if the database contains genes, gene aliases, and Pfam domains from mouse). This PR also incorporates the improvements in calculating gene length that have been proposed in #1678.

The portal.properties file should incorporate these new variables. Examples:
For mouse:
`species=mouse`
`ncbi.build=38`
`ucsc.build=mm10` 

For human (or leave them empty):
`species=human`
`ncbi.build=37`
`ucsc.build=hg19` 

Specifying other species in portal.properties will raise an error message in the home page (similar to when the database is out of sync).

@pieterlukasse


Sjoerd-van-Hagen:
@jjgao Hi JJ, do you think we could get this merged next week?

PR #2103, opened 2017-01-30
fix to avoid parsing error message in console when using IE11 or Chrome

pieterlukasse:
# What? Why?
Fix a small issue that makes IE11 show an error message in the browsers console. 

# Notify reviewers
@yichaoS can you please review this one? It's just a small change.


pieterlukasse:
It actually also gives an error in Chrome:

image

Issue #2063, opened 2017-01-13
Use gene panel info when calculating the percentages in study view


jjgao:
For `Mutated Genes` and `CN-altered Genes` tables, the percentages are not correct if different gene panels were used.

PR #2130, opened 2017-02-07
fix issue #2063 (using gene panel in study view)

yichaoS:
Fix https://github.com/cBioPortal/cbioportal/issues/2063
Based on changes under https://github.com/cBioPortal/iViz/pull/361/files#diff-684aebcce37ce500e90bf6891b895ec4


jjgao:
@yichaoS 

In the new screenshot, why all frequency column for mutations table are NA? Is it because they don't have gene panel data?

In case that gene panel data is not available, we assume all genes are sequenced. The previous logic should be used, ie. the denominator should be the selected sequenced samples.

PR #2134, opened 2017-02-10
Create a new maven module for db scripts

ersinciftci:
In order to break the dependency from persistence-mybatis to core,
moved db scripts from core to a new module. persistence-mybatis is
using db scripts for testing purposes.


ersinciftci:
@inodb @sheridancbio In order to run `sql-maven-plugin`, I need to unpack cgds.sql from the new `db-scripts` module using `maven-dependency-plugin`, and you can unpack only after the `package` phase of maven. So these `core` tests connecting to a real db have to run after `package` phase, hence they became integration tests. And they were actually integration tests anyway, this change just forced us to do the correct thing :) `mvn clean test` doesn't run integration tests, so I changed it to `mvn clean integration-test`, which also runs unit tests. 

Here is the order of maven lifecycle phases:
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference

inodb:
Thanks @ersinciftci, makes sense!

ersinciftci:
Please note that I could use `<srcFile>${project.parent.basedir}/db-scripts/src/main/resources/cgds.sql</srcFile>` in `sql-maven-plugin` for now, but in the long term I prefer db scripts to exist in their own repos, so implemented in a way that would work in that case as well.

ersinciftci:
@pieterlukasse I don't see a hardcoded path in `migrate_db.py`, user sends the path as parameter, what do I need to change there?

ersinciftci:
Adjusted `Updating-your-cBioPortal-installation.md`

PR #2137, opened 2017-02-14
Add Proxy API

ersinciftci:
Proxy API will be used to call 3rd party APIs when it's not possible
to call them from frontend.


ersinciftci:
[(5) Create generic proxy API](https://trello.com/c/H8spiJeH/110-5-create-generic-proxy-api)

PR #2138, opened 2017-02-14
Security rearchitecture

mandawilson:
# What? Why?
Security rearchitecture.

- New security module
- Secured the new API endpoints

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers

@n1zea144 @ersinciftci 

and notify them here:


mandawilson:
@n1zea144 Remember that we discussed how the CancerStudyPermissionEvaluator class used to get "fresh" groups to check permissions and this one doesn't?  We should review that and see if group caching is still a problem.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -51,24 +56,28 @@ public Sample getSampleInStudy(String studyId, String sampleId) throws SampleNot
...
 
     @Override
+    @PreAuthorize("hasPermission(#studyIds, 'List<CancerStudyId>', 'read')")
     public BaseMeta fetchMetaSamples(List<String> studyIds, List<String> sampleIds) {
 
         return sampleRepository.fetchMetaSamples(studyIds, sampleIds);
     }
 
     @Override
+    @PreAuthorize("hasPermission(#internalIds, 'List<InternalSampleId>', 'read')")
```
ersinciftci: we don't need to secure this one as it's used internally and don't correspond to an API (and it will never as we will never get internal ids from users), it's public because it's used by another service but it's internal. So you can delete this, and the part handles `List<InternalSampleId>` in evaluator, and also the new `getSampleByInternalId` from sample mapper and repository.

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -0,0 +1,393 @@
...
+            return true;
+        } else if ("List<GeneticProfileId>".equals(targetType)) {
+            List<String> geneticProfileIds = (List<String>) targetId;
+            for (String geneticProfileId : geneticProfileIds) {
+                GeneticProfile geneticProfile = geneticProfileRepository.getGeneticProfile(geneticProfileId);
+                if (geneticProfile == null || !hasPermission(authentication, geneticProfile, permission)) {
+                    return false;
+                }
+            }
+        } else if ("List<InternalSampleId>".equals(targetType)) {
```
ersinciftci: As I said in the other comment, we don't need this part, and since it gets samples one by one from db for every internal id in `genetic_profile_samples` table (can be thousands of them), it will slow things down.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/PortalUserMapper.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
+<mapper namespace="org.cbioportal.persistence.mybatis.PortalUserMapper">
+    <cache/>
```
mandawilson: Remove cache.

mandawilson:
@n1zea144, @ersinciftci and I talked about getting the fresh cancer groups and we decided to add a query to the PortalUserMapper.xml. He correctly pointed out that that file should not cache queries anyway.  I will also rename it SecurityMapper.xml.  So I will do that next week, and also make the other change Ersin suggested if you agree we should.

ersinciftci:
The reason it would fit better in PortalUserMapper.xml instead of StudyMapper.xml is because `GROUPS` shouldn't be part of `cancer_study` table, there should be a separate table that matches studies to groups, because groups is not a property of a standalone study, it's something we assign to a study. But for now we can query it in PortalUserMapper.xml since it won't have caching, in the long term we should create a table for it.

mandawilson:
We get the uncached groups for permission checking now.  Waiting to hear from @n1zea144 on the fetchMetaSamples question.

jjgao:
[Create security module](https://trello.com/c/UilTjBHR/73-create-security-module)

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
mandawilson: We do need permission checking on "List<InternalSampleId>" for ApiService.java even if we don't for SampleServiceImpl.java.

pom.xml
```diff
@@ -193,6 +193,7 @@
     <module>service</module>
     <module>scripts</module>
     <module>db-scripts</module>
+    <module>security</module>
```
n1zea144: Should we be removing references to spring-security dependencies below?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
n1zea144: It seems redundant to preauthorize in multiple places on the same execution thread.  Also, we will probably take a performance hit validating on every sample ID.  Of course, code evolves and assumptions we make today are not necessarily true tomorrow.  My fear would be that at some point, code is introduced which is not properly annotated, that makes calls into this method.  For the performance reasons, I would do what @ersinciftci suggests, but I would also document the method explaining why it is not annotated.

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -0,0 +1,396 @@
...
+    private SampleListRepository sampleListRepository;
+
+    @Value("${authenticate}")
+    private String AUTHENTICATE;
+
+    @Value("${app.name}")
+    private String APP_NAME;
+    private String DEFAULT_APP_NAME = "public_portal";
+
+    @Value("${filter_groups_by_appname}")
```
n1zea144: As it stands, code cannot deploy propertly without a defined property.  I would set a default value here.

n1zea144:
BTW, @mandawilson I failed to mention - this is a big PR.  Lots of nice improvements.  Great work!

security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/saml/MSKCCPortalUserDetailsService.java
```diff
@@ -30,14 +30,16 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-package org.mskcc.cbio.portal.authentication.saml;
+package org.cbioportal.security.spring.authentication.saml;
```
n1zea144: When we get around to merging @pieterlukasse SAML updates, I think we should remove this and the restful authentication handler from the codebase.  We can discuss more when you get to it.

ersinciftci:
Looks great, thanks!

n1zea144:
The default behavior has changed.  If I don't define filter_groups_by_appname=true, the MSK portal no longer work (because we run multiple instances of the portal that share a common database).  Can you see what's changed?

Also, we are not properly handling when a user has access to zero studies.

image

n1zea144:
I'm not going to hold up the merge because of this, but we now have a situation where if the user has access to no studies (I know it shouldn't happen), they get to the query page with nothing to do.  I think a message should be displayed that lets the user know they don't have access to any studies:

image

PR #2139, opened 2017-02-15
Correct documented path to migration.sql

fedde-s:
Commit b6f8d00d2e07d6346b7d3c9a170d924520765c52 broke the documented
command to upgrade the database schema to the version expected by
the portal during installation, by making a mistake in the path to
one of the required files. This fixes the path, so that the command
runs as documented again.


fedde-s:
@ersinciftci, does this look good to you?

ersinciftci:
@fedde-s Oh, I changed the first part but missed the `db` folder, nice catch, thanks!

PR #2140, opened 2017-02-15
enabled virtual study in mutation tab

yichaoS:
# What? Why?
Fix https://github.com/cBioPortal/iViz/issues/347


portal/src/main/webapp/js/src/DataProxyFactory.js
```diff
@@ -41,7 +41,7 @@ var DataProxyFactory = (function()
 	 *
 	 * @return singleton instance of MutationDataProxy class
 	 */
-	function getDefaultMutationDataProxy()
+	function getDefaultMutationDataProxy(_isVirtualStudy)
```
onursumer: Just a minor styling note: For function parameter names, it might be better not to start with underscore

PR #2149, opened 2017-02-17
Improve calculation of gene length (rc)

oplantalech:
Create PR based on #1678 for rc. @jjgao


jjgao:
@oplantalech thank you!

PR #2155, opened 2017-02-24
Gene set related tables

pieterlukasse:
# What? Why?
This PR introduces the DB `geneset*` tables needed to support the GSVA part. 

- new `geneset*` tables (5 in total)
- new `genetic_profile_link` table to keep track of the relation between GSVA-SCORE and GSVA P-VALUE profiles (score and p-value are stored in separate profiles)
- small related change in DaoCancerStudy.deleteCancerStudy

# Notify reviewers
@n1zea144 : part of this has been reviewed and tested by @angelicaochoa before. Maybe @angelicaochoa can review this one as well?

### DIFF
--- a/db-scripts/src/main/resources/cgds.sql
+++ b/db-scripts/src/main/resources/cgds.sql
@@ -78,9 +78,11 @@ DROP TABLE IF EXISTS `gene_panel_list`;
 DROP TABLE IF EXISTS `gene_panel`;
 DROP TABLE IF EXISTS `genetic_profile_samples`;
 DROP TABLE IF EXISTS `genetic_alteration`;
+DROP TABLE IF EXISTS `genetic_profile_link`;
 DROP TABLE IF EXISTS `genetic_profile`;
 DROP TABLE IF EXISTS `uniprot_id_mapping`;
 DROP TABLE IF EXISTS `gene_alias`;
+DROP TABLE IF EXISTS `geneset_gene`;
 DROP TABLE IF EXISTS `gene`;
 DROP TABLE IF EXISTS `sample_list_list`;
 DROP TABLE IF EXISTS `sample_list`;
@@ -90,6 +92,9 @@ DROP TABLE IF EXISTS `authorities`;
 DROP TABLE IF EXISTS `users`;
 DROP TABLE IF EXISTS `cancer_study`;
 DROP TABLE IF EXISTS `type_of_cancer`;
+DROP TABLE IF EXISTS `geneset_hierarchy_leaf`;
+DROP TABLE IF EXISTS `geneset_hierarchy_node`;
+DROP TABLE IF EXISTS `geneset`;
 DROP TABLE IF EXISTS `genetic_entity`;
 
 -- --------------------------------------------------------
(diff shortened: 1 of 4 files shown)


db-scripts/src/main/resources/cgds.sql
```diff
@@ -207,6 +213,53 @@ CREATE TABLE `gene_alias` (
...
+CREATE TABLE `geneset_hierarchy_leaf` (
+  `NODE_ID` BIGINT NOT NULL,
+  `GENESET_ID` INT NOT NULL,
+  PRIMARY KEY (`NODE_ID`, `GENESET_ID`),
+  FOREIGN KEY (`NODE_ID`) REFERENCES `geneset_hierarchy` (`NODE_ID`) ON DELETE CASCADE,
+  FOREIGN KEY (`GENESET_ID`) REFERENCES `geneset` (`ID`) ON DELETE CASCADE
+);
+
+-- --------------------------------------------------------
+CREATE TABLE `geneset_info` (
```
ersinciftci: Instead of adding a new table for this, maybe we can add another column to `info` table, called `GENESET_VERSION`?

pieterlukasse: Sounds good. Will do.

db-scripts/src/main/resources/cgds.sql
```diff
@@ -207,6 +213,53 @@ CREATE TABLE `gene_alias` (
...
+  `NODE_ID` BIGINT(20) NOT NULL auto_increment,
+  `NODE_NAME` VARCHAR(200) NOT NULL,
+  `PARENT_ID` BIGINT NULL DEFAULT NULL,
+  PRIMARY KEY (`NODE_ID`),
+  UNIQUE INDEX `NODE_NAME_UNIQUE` (`NODE_NAME` ASC, `PARENT_ID` ASC)
+);
+
+-- --------------------------------------------------------
+CREATE TABLE `geneset_hierarchy_leaf` (
+  `NODE_ID` BIGINT NOT NULL,
```
fedde-s: So this node ID does not actually refer to a node associated with the gene set, but rather to the *parent* node of the gene set?

pieterlukasse: Yes, it is a foreign key to the table that contains nodes. We could have named it PARENT_NODE_ID, but I think the context of this table and the direction of the foreign key make clear what its purpose is. The table `geneset_gene` also follows this "convention" of just repeating the field name of where it is pointing to. Other tables in cgds do the same. It is a matter of taste IMO, and I would rather keep it like this.

db-scripts/src/main/resources/cgds.sql
pieterlukasse: pushed up this change now.

db-scripts/src/main/resources/cgds.sql
fedde-s: It's clear to me now that I read it as a link table that links gene sets to categories in the hierarchy in a many-to-many relation, thanks for the clarification.

ersinciftci: I think `PARENT_NODE_ID` would be better because normally in a tree a leaf is also a node, so it can have it's own NODE_ID, in our case it's not necessary but I thought you gave it a NODE_ID just in case when I first looked at the columns of this table.

ersinciftci: `geneset_gene` table is a mapping between `gene_set` and `gene`, so there is no need to name it `PARENT_GENESET_ID`, but hierarchy case is different, there is a parent table and the leaf table, not a direct mapping.

pieterlukasse: I understand...but I still think `PARENT_NODE_ID` is not adding much more clarity in this context and is potentially confusing because the other table `geneset_hierarchy` also has a field called `PARENT_ID`....so I would like to suggest we keep it as is. 

One thing that I realized now is that we could (maybe should) provide an updated ER diagram in this PR. This is anyway the best way to ready DB models as such relations become immediately clear. We have one hanging in our corner and actually have used it a lot:

ersinciftci: It would only make sense if you change the `geneset_hierarchy` table name to `geneset_hierarchy_node`.

pieterlukasse: `geneset_hierarchy_node`...I like that. Let's do that.

pieterlukasse: pushed up this change as well

angelicaochoa:
👍  looks good, I had no issues with building and didn't find any discrepancies with field or table names in PR #2166

pieterlukasse:
@ersinciftci @angelicaochoa thanks! Merging now.

PR #2158, opened 2017-02-24
Gene set APIs

pieterlukasse:
# What? Why?
This PR adds both public and some internal endpoints necessary for the gene set functionality in both data validation and front end. 

- New public endpoints for `geneset` alteration data 
- New internal endpoints to support specific gene set features in the frontend (see #2164 )


model/src/main/java/org/cbioportal/model/GenesetAlteration.java
```diff
@@ -0,0 +1,40 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+
+public class GenesetAlteration implements Serializable {
```
ersinciftci: This is the `genetic_alteration` joined with `geneset` table, and current `GeneticAlteration.java` is the `genetic_alteration` joined with `gene` table, can you rename the `GeneticAlteration.java` to `GeneAlteration.java` to be consistent with this one, and then create an abstract parent class for both of them named `GeneticAlteration.java` (won't be used anywhere, just to indicate those classes are related, uses the same table), which will have `geneticEntityId` and `values`, and then remove `values` from both of those classes?

model/src/main/java/org/cbioportal/model/GenesetData.java
```diff
@@ -0,0 +1,44 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+import java.math.BigDecimal;
+
+public class GenesetData implements Serializable {
```
ersinciftci: Same thing goes for this as well, we can change the current `GeneticData.java` to `GeneData.java` and create an abstract parent for these two called `GeneticData.java`, it can have `geneticProfileId`, `sampleId` and `value`, and we can remove those fields from these two classes.

model/src/main/java/org/cbioportal/model/GeneticProfile.java
```diff
@@ -20,7 +20,16 @@
         PHOSPHORYLATION,
         PROTEIN_LEVEL,
         PROTEIN_ARRAY_PROTEIN_LEVEL,
-        PROTEIN_ARRAY_PHOSPHORYLATION;
+        PROTEIN_ARRAY_PHOSPHORYLATION, 
+        GENESET_SCORE;
+    }
+    
+    public static class DataType {
```
ersinciftci: I guess this is supposed to be an enum?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
```diff
@@ -58,4 +58,6 @@
     BaseMeta fetchMetaGenesByEntrezGeneIds(List<Integer> entrezGeneIds);
 
     BaseMeta fetchMetaGenesByHugoGeneSymbols(List<String> hugoGeneSymbols);
+    
+	List<Gene> getGenesByGenesetId(String genesetId);
```
ersinciftci: I think this should be in `GenesetRepository`.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMapper.java
```diff
@@ -58,4 +58,6 @@
     BaseMeta getMetaGenesByEntrezGeneIds(List<Integer> entrezGeneIds);
 
     BaseMeta getMetaGenesByHugoGeneSymbols(List<String> hugoGeneSymbols);
+    
+	List<Gene> getGenesByGenesetId(String genesetId, String projection);
```
ersinciftci: Same here, I think this should be in GenesetMapper.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMyBatisRepository.java
```diff
@@ -111,4 +111,10 @@ public BaseMeta fetchMetaGenesByHugoGeneSymbols(List<String> hugoGeneSymbols) {
 
         return geneMapper.getMetaGenesByHugoGeneSymbols(hugoGeneSymbols);
     }
+
+	@Override
+	public List<Gene> getGenesByGenesetId(String genesetId) {
```
ersinciftci: Same here, I think this should be in GenesetMyBatisRepository.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
```diff
@@ -113,6 +113,18 @@
         </where>
     </select>
 
+    <select id="getGenesByGenesetId" resultType="org.cbioportal.model.Gene">
```
ersinciftci: Same here, I think this should be in GenesetMapper.xml.

service/src/main/java/org/cbioportal/service/GeneService.java
```diff
@@ -51,4 +51,6 @@
     List<Gene> fetchGenes(List<String> geneIds, String geneIdType, String projection);
 
     BaseMeta fetchMetaGenes(List<String> geneIds, String geneIdType);
+    
+	List<Gene> getGenesByGenesetId(String genesetId);
```
ersinciftci: Same here, I think this should be in GenesetService.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -135,4 +135,10 @@ public BaseMeta fetchMetaGenes(List<String> geneIds, String geneIdType) {
     private boolean isInteger(String geneId) {
         return geneId.matches("^-?\\d+$");
     }
+
+	@Override
+	public List<Gene> getGenesByGenesetId(String genesetId) {
```
ersinciftci: Same here, I think this should be in GenesetServiceImpl.

model/src/main/java/org/cbioportal/model/GenesetAlteration.java
pieterlukasse: What would be stored in ` geneticEntityId`? Would this be the external id (so sometimes `Entrez_gene_id`, sometime geneset external id)? Or would it just be the internal `geneticEntityId` which is now present in both `gene` and `geneset` tables? 

As a side note: I thought you wanted to avoid the geneticEntityId as much as possible. Could you ellaborate a bit on the reason for having it in `model` now?

model/src/main/java/org/cbioportal/model/GenesetData.java
pieterlukasse: Ok, makes sense. Will do.

model/src/main/java/org/cbioportal/model/GeneticProfile.java
pieterlukasse: I found enum to be a bit annoying when it comes to values with `-` . These statics are used (I think in all cases) for guiding logic that is specific for one particular data type at a time. Anyway, do you have tips for how to deal with this if we would really want to use enum here?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
pieterlukasse: Ok, we can move it there

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMapper.java
pieterlukasse: ok

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMyBatisRepository.java
pieterlukasse: ok

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
pieterlukasse: ok

service/src/main/java/org/cbioportal/service/GeneService.java
pieterlukasse: ok

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
pieterlukasse: ok

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        
+        for (Sample sample : samples) {
+            int indexOfSampleId = internalSampleIds.indexOf(sample.getInternalId());
+            if (indexOfSampleId != -1) {
+                for (GenesetAlteration genesetAlteration : genesetAlterations) {
+                    GenesetData genesetData = new GenesetData();
+                    genesetData.setGeneticProfileId(geneticProfileId);
+                    genesetData.setSampleId(sample.getStableId());
+                    genesetData.setGenesetId(genesetAlteration.getGenesetId());
+                    genesetData.setValue(genesetAlteration.getSplitValues()[indexOfSampleId]); //TODO introduce this "splitValues cache" in genetic data API as well...since .split(",") turned out to be bottleneck here...service went from 4s to 165ms after this improvement!
```
pieterlukasse: performance improvement! @ersinciftci you may want to check this out and do the same in geneticAlteration data api

model/src/main/java/org/cbioportal/model/GenesetAlteration.java
ersinciftci: Actually yeah you can leave `geneticEntityId` out for now as we don't need it, just `values` would be enough in the parent class.

model/src/main/java/org/cbioportal/model/GeneticProfile.java
ersinciftci: You can take a look at SampleType enum in Sample model, it doesn't have `-` but it has white space in it, can be solved the same way I think. You will need a MyBatis type handler like `org.cbioportal.persistence.mybatis.typehandler.SampleTypeTypeHandler` for conversion. And `super.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);` in CustomObjectMapper will use the toString method of enum, so in JSON response you will get `P-VALUE` instead of `P_VALUE`. (You need to return the value from toString method in your enum, just like SampleType enum).

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
ersinciftci: @pieterlukasse that's right! I refactored the method a few times, didn't realize that I was calling the `split` on the same alteration object for every sample, nice catch!

model/src/main/java/org/cbioportal/model/GenesetAlteration.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public String getValues() {
+        return values;
+    }
+
+    public void setValues(String values) {
+        this.values = values;
+    }
+    
+    private String[] splitValues = null;
```
ersinciftci: Can you put this on the top with other fields?

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
pieterlukasse: Ok, will do.

model/src/main/java/org/cbioportal/model/GenesetAlteration.java
pieterlukasse: sure

model/src/main/java/org/cbioportal/model/GenesetAlteration.java
pieterlukasse: ok

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
pieterlukasse: done

model/src/main/java/org/cbioportal/model/GenesetAlteration.java
pieterlukasse: done

model/src/main/java/org/cbioportal/model/GenesetData.java
pieterlukasse: done

model/src/main/java/org/cbioportal/model/GeneticProfile.java
pieterlukasse: ok, done

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
pieterlukasse: done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMapper.java
pieterlukasse: done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneMyBatisRepository.java
pieterlukasse: done

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
pieterlukasse: done

service/src/main/java/org/cbioportal/service/GeneService.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
```diff
@@ -0,0 +1,264 @@
...
+    @Autowired
+    private GenesetDataService genesetDataService;
+    @Autowired
+    private GeneticProfileService geneticProfileService;
+    @Autowired
+    private GenesetService genesetService;
+    @Autowired
+    private SampleService sampleService;
+    @Autowired
+    private SampleListRepository sampleListRepository;
```
ersinciftci: Can we use SampleListService here instead of using SampleListRepository directly?

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
```diff
@@ -0,0 +1,264 @@
...
+	@Override
+	public List<GenesetCorrelation> fetchCorrelatedGenes(String genesetId, String geneticProfileId, List<String> sampleIds,
+			double correlationThreshold) throws GeneticProfileNotFoundException {
+		SpearmansCorrelation spearmansCorrelation = new SpearmansCorrelation();
+		List<GenesetCorrelation> result = new ArrayList<GenesetCorrelation>();
+		
+		// find the genes in the geneset
+		List<Gene> genes = genesetService.getGenesByGenesetId(genesetId);
+		
+        // the geneset data:
```
ersinciftci: I think there are some indentation issues in the rest of this method.

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
```diff
@@ -0,0 +1,264 @@
...
+		// return sorted
+        sortResult(result);
+		return result;
+	}
+
+
+	private GeneticProfile getLinkedZscoreProfile(GeneticProfile expressionProfile) {
+		
+		//Find the related z-score profile via the genetic_profile_link table:
+    	List<GeneticProfile> referringProfiles = geneticProfileService.getGeneticProfilesReferringTo(expressionProfile.getStableId());
```
ersinciftci: I think there are some indentation issues in the rest of this method.

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
```diff
@@ -0,0 +1,264 @@
...
+			if (value != null) {
+				result[i] = value;
+			} else {
+				result[i] = Double.NaN; //set to NaN when value is not available for this sample
+			}
+		}
+		return result;
+	}
+	
+	class FilteredGeneAndGenesetValues {
```
ersinciftci: Can we move this to `model`? Even if it's only used in this class, I would still make it a model.

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+import org.cbioportal.service.SampleService;
+import org.cbioportal.service.exception.GeneticProfileNotFoundException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class GenesetDataServiceImpl implements GenesetDataService {
+
+    @Autowired
+    private GeneticDataRepository geneticDataRepository;
```
ersinciftci: Can we use GeneticDataService instead of GeneticDataRepository here? You will need to add getGenesetGeneticAlterations to GeneticDataService too.

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    @Autowired
+    private SampleService sampleService;
+    @Autowired
+    private GeneticProfileService geneticProfileService;
+    @Autowired
+    private SampleListRepository sampleListRepository;
```
ersinciftci: Can we use SampleListService here instead of using SampleListRepository directly?

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
```diff
@@ -0,0 +1,385 @@
...
+    @Autowired
+    private GeneticProfileService geneticProfileService;
+    @Autowired
+    private GenesetHierarchyRepository genesetHierarchyRepository;
+    @Autowired
+    private SampleService sampleService;
+    @Autowired
+    private SampleListRepository sampleListRepository;
+	@Autowired
+    private GeneticProfileRepository geneticProfileRepository;
```
ersinciftci: GeneticProfileService is already autowired here, you can remove this one.

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
```diff
@@ -0,0 +1,385 @@
...
+    @Autowired
+    private GenesetDataService genesetDataService;
+    @Autowired
+    private GeneticProfileService geneticProfileService;
+    @Autowired
+    private GenesetHierarchyRepository genesetHierarchyRepository;
+    @Autowired
+    private SampleService sampleService;
+    @Autowired
+    private SampleListRepository sampleListRepository;
```
ersinciftci: Can we use SampleListService here instead of using SampleListRepository directly?

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
```diff
@@ -0,0 +1,385 @@
...
+		}
+	}
+
+	/**
+	 * Inner class to hold score and p-value tuple. 
+	 * 
+	 * @author pieter
+	 *
+	 */
+	static class ScoreAndPvalue {
```
ersinciftci: Can you make this a proper model in `model` module (with a more clear name), even if it's only used here.

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
```diff
@@ -0,0 +1,385 @@
...
+    			pvalueOfMax = pvalue;
+    		}
+		}
+		
+		if (percentile == null) {
+			geneset.setRepresentativeScore(max);
+			geneset.setRepresentativePvalue(pvalueOfMax);
+		} else {
+			//sort scores:
+			Collections.sort(positiveScoresAndPvalues, ScoreAndPvalue.comparator);
```
ersinciftci: Can you use lambda expression here using `positiveScoresAndPvalues.sort` method, instead of Comparator?

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
```diff
@@ -0,0 +1,385 @@
...
+			Collections.sort(negativeScoresAndPvalues, ScoreAndPvalue.comparatorDesc);
```
ersinciftci: Can you use lambda expression here using `positiveScoresAndPvalues.sort` method, instead of Comparator?

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
```diff
@@ -0,0 +1,264 @@
...
+        		
+        		if (o1.getCorrelationValue() < o2.getCorrelationValue())
+        			return 1;
+        		if (o1.getCorrelationValue() > o2.getCorrelationValue())
+        			return -1;
+        		
+        		return 0;
+            }
+        };
+		Collections.sort(result, comparator);
```
ersinciftci: Can you use lambda expression here using `result.sort` method, instead of Comparator?

web/src/main/java/org/cbioportal/web/CopyNumberCountController.java
```diff
@@ -39,7 +39,7 @@
         produces = MediaType.APPLICATION_JSON_VALUE)
     @ApiOperation("Get counts of specific genes and alterations within a CNA genetic profile")
     public ResponseEntity<List<CopyNumberCount>> fetchCopyNumberCounts(
-        @ApiParam(required = true, value = "Genetic Profile ID e.g. acc_tcga_mutations")
+        @ApiParam(required = true, value = "Genetic Profile ID e.g. acc_tcga_gistic")
```
ersinciftci: I thought I fixed this, but seems I forgot, thanks!

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneticDataRepository.java
```diff
@@ -1,13 +1,16 @@
...
 
+import org.cbioportal.model.GenesetGeneticAlteration;
+import org.cbioportal.model.GeneGeneticAlteration;
+
 public interface GeneticDataRepository {
 
     String getCommaSeparatedSampleIdsOfGeneticProfile(String geneticProfileId);
 
-    List<GeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
+    List<GeneGeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds,
```
pieterlukasse: done

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneticDataRepository.java
```diff
@@ -1,13 +1,16 @@
...
                                                   String projection);
+
+	List<GenesetGeneticAlteration> getGenesetAlterations(String geneticProfileId, List<String> genesetIds, String projection);
```
pieterlukasse: done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneticDataMapper.java
```diff
@@ -1,13 +1,17 @@
...
+import org.cbioportal.model.GenesetGeneticAlteration;
 
 import java.util.List;
 
 public interface GeneticDataMapper {
 
     String getCommaSeparatedSampleIdsOfGeneticProfile(String geneticProfileId);
 
-    List<GeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
+    List<GeneGeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds,
```
pieterlukasse: done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneticDataMapper.java
```diff
@@ -1,13 +1,17 @@
...
 
 public interface GeneticDataMapper {
 
     String getCommaSeparatedSampleIdsOfGeneticProfile(String geneticProfileId);
 
-    List<GeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
+    List<GeneGeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
                                                   String projection);
+
+	List<GenesetGeneticAlteration> getGenesetAlterations(String geneticProfileId, List<String> genesetIds,
```
pieterlukasse: done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneticDataMyBatisRepository.java
```diff
@@ -21,9 +21,15 @@ public String getCommaSeparatedSampleIdsOfGeneticProfile(String geneticProfileId
     }
 
     @Override
-    public List<GeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
+    public List<GeneGeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds,
```
pieterlukasse: done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GeneticDataMyBatisRepository.java
```diff
@@ -21,9 +21,15 @@ public String getCommaSeparatedSampleIdsOfGeneticProfile(String geneticProfileId
...
     @Override
-    public List<GeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
+    public List<GeneGeneticAlteration> getGeneticAlterations(String geneticProfileId, List<Integer> entrezGeneIds, 
                                                          String projection) {
 
         return geneticDataMapper.getGeneticAlterations(geneticProfileId, entrezGeneIds, projection);
     }
+
+	@Override
+	public List<GenesetGeneticAlteration> getGenesetAlterations(String geneticProfileId, List<String> genesetIds,
```
pieterlukasse: done

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneticDataMapper.xml
```diff
@@ -11,7 +11,7 @@
         WHERE genetic_profile.STABLE_ID = #{geneticProfileId}
     </select>
     
-    <select id="getGeneticAlterations" resultType="org.cbioportal.model.GeneticAlteration">
+    <select id="getGeneticAlterations" resultType="org.cbioportal.model.GeneGeneticAlteration">
```
pieterlukasse: done

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneticDataMapper.xml
```diff
@@ -35,4 +35,29 @@
         </where>
     </select>
 
+    <select id="getGenesetAlterations" resultType="org.cbioportal.model.GenesetGeneticAlteration">
```
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
pieterlukasse: Done. Not sure why I used SampleListRepository...maybe SampleListService was not there yet....anyway nice catch.

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
pieterlukasse: what about the method getCommaSeparatedSampleIdsOfGeneticProfile being used here?

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
pieterlukasse: nice tip. Didn't know that one. Done now.

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
pieterlukasse: I prefer not. I think it will be confusing. This is just a "transient" type to keep score and p-value together while we sort on score. At the end, the values are set in the geneset item: 
```java
geneset.setRepresentativeScore(representativeNegativeScoreAndPvalue.score);
geneset.setRepresentativePvalue(representativeNegativeScoreAndPvalue.pvalue);
```
Or do you know a better way that does not need this class perhaps?

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
pieterlukasse: I prefer not to. I am afraid we will add too many things to model. This class is just a transient type to allow a method to return 2 lists. I could also change the method to return a list with two lists...although I always find this to be less readable...or maybe you have another idea?

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
pieterlukasse: should be fixed now

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
pieterlukasse: should be fixed now

service/src/main/java/org/cbioportal/service/impl/GenesetCorrelationServiceImpl.java
ersinciftci: Yeah, you are right, we wouldn't want `model` to be too crowded either. I think a Map would be better here, as this class is not really part of the domain and is just a temporary value holder.

ersinciftci: Or you can call `spearmansCorrelation.correlation` in the same method you create the lists, so you wouldn't need to return 2 lists, up to you.

pieterlukasse: ah, good point. I'll do that last one.

pieterlukasse: done

service/src/main/java/org/cbioportal/service/impl/GenesetHierarchyServiceImpl.java
pieterlukasse: Done. I had to add a new dependency to commons-lang3 though. And I'm not sure if the code is much prettier now...but yeah...one custom class less :)

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
pieterlukasse: Looking at it again, I think this could lead to some undesired features. Here are some thoughts:
- `geneticDataRepository.getGenesetGeneticAlterations` and `geneticDataRepository.getCommaSeparatedSampleIdsOfGeneticProfile` return CSV string with the values, something that is a bit ugly about our data model and that we would like to keep hidden from outer layers. Exposing these in the service layer would defeat this goal. 
- When I first wrote this code, I intended `GeneticDataRepository` to be a "shared" repository layer for all genetic data. That is why I added `getGenesetGeneticAlterations` there as well. As such, it should be fine to use it here in `GenesetDataServiceImpl` as well as in `GeneticDataServiceImpl` (maybe this last one should be renamed to `GeneGeneticDataServiceImpl` to keep it more consistent). Alternatively, we could introduce `GenesetDataRepository` instead of having both gene and geneset services share the same `GeneticDataRepository`.

ersinciftci:
@pieterlukasse from this line `@Api(tags = "Gene sets", description = " ")`, I just realized that `Gene set` is two separate words, I always thought it was `Geneset` (I googled and this shows up https://en.wikipedia.org/wiki/Gene_set_enrichment), so in your PR, I guess you need to change lots of class and variable names from `Geneset` to `GeneSet` and from `geneset` to `geneSet`. And also `/genesets` to `/gene-sets`.

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
ersinciftci: I see, then the current way is fine, not a big deal.

web/src/main/java/org/cbioportal/web/GenesetDataController.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+
+@PublicApi
+@RestController
+@Validated
+@Api(tags = "Gene set Data", description = " ")
```
ersinciftci: Can we change this to `Gene Set Genetic Data`?

web/src/main/java/org/cbioportal/web/GenesetController.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+
+@PublicApi
+@RestController
+@Validated
+@Api(tags = "Gene sets", description = " ")
```
ersinciftci: Can you make all words start with uppercase; `Gene Sets`.

web/src/main/java/org/cbioportal/web/GenesetCorrelationController.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+
+@InternalApi
+@RestController
+@Validated
+@Api(tags = "Gene set correlation", description = " ")
```
ersinciftci: Can you make all words start with uppercase; `Gene Set Correlation`.

web/src/main/java/org/cbioportal/web/GenesetDataController.java
```diff
@@ -0,0 +1,59 @@
...
+public class GenesetDataController {
+
+
+	@Autowired
+    private GenesetDataService genesetDataService;
+    
+    @RequestMapping(value = "/genetic-profiles/{geneticProfileId}/geneset-data/fetch", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE,
```
ersinciftci: and this to `/geneset-genetic-data/`

web/src/main/java/org/cbioportal/web/GenesetHierarchyController.java
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+
+@InternalApi
+@RestController
+@Validated
+@Api(tags = "Gene set hierarchy", description = " ")
```
ersinciftci: Can you make all words start with uppercase; `Gene Set Hierarchy`.

web/src/main/java/org/cbioportal/web/GenesetHierarchyController.java
```diff
@@ -0,0 +1,100 @@
...
+public class GenesetHierarchyController {
+
+    @Autowired
+    private GenesetHierarchyService genesetHierarchyService;
+
+    @RequestMapping(value = "/genesets/hierarchy/fetch", method = RequestMethod.POST,
```
ersinciftci: Can you change the path to `/geneset-hierarchy/fetch`, because we you say `/genesets/`, you are either getting a list of genesets, or a specific geneset resource, so having something other than a geneset id after `/genesets/` wouldn't make sense. `/geneset-hierarchy` looks like a different resource than `/genesets`.

web/src/main/java/org/cbioportal/web/GenesetDataController.java
pieterlukasse: done

web/src/main/java/org/cbioportal/web/GenesetDataController.java
pieterlukasse: done

web/src/main/java/org/cbioportal/web/GenesetController.java
pieterlukasse: done

web/src/main/java/org/cbioportal/web/GenesetCorrelationController.java
pieterlukasse: done

web/src/main/java/org/cbioportal/web/GenesetHierarchyController.java
pieterlukasse: done

web/src/main/java/org/cbioportal/web/GenesetHierarchyController.java
pieterlukasse: Ok, done.

web/src/main/java/org/cbioportal/web/GenesetDataController.java
pieterlukasse: As a side note: I think genetic data sounds a bit strange for what it is now in the API. It is actually "alteration data" (not only in the gene set scenario but also for genes). I.e. Currently `GeneticDataController.getAllGeneticDataInSampleInStudy` would be clearer if it would be named `getAllAlterationDataInSampleInStudy`. But this could be a topic of another PR. I agree we should make it at least uniform now. Then it is easier to make the change to another word later.

ersinciftci: It was alteration data before, but @jjgao wanted it to be Genetic Data.

pieterlukasse:
rebased on latest rc

pieterlukasse:
@ersinciftci thanks for the review. Last change I made here was to try to comply to your 404 changes in #2294. Maybe you can take a quick look at the controller layer (just the exceptions it is throwing now)?

pieterlukasse:
@ersinciftci merging now, given our discussion on slack.

PR #2159, opened 2017-02-24
Support for gene panel & gene panel matrix data in validateData.py & cbioportalImporter.py

n1zea144:
Support for gene panel & gene panel matrix data in validateData.py & cbioportalImporter.py.

Gene panel data consists of two files, a gene panel file which contains meta data about the panel, including a gene list, and a matrix file which maps a gene panel to a sample-genetic profile pair.  Some notes:

- Since the gene panel file is very similar to any meta file, it was decided that gene panel data files should have a meta_gene_panel prefix in its name so that it can leverage existing meta file validation support already provide in validateData.py.  Unfortunately, because the gene panel file is being treated as a meta file, it is not possible to leverage the existing checkGeneIdentification code which is buried with the datafile - Validator class design.
- Since a gene panel stable id is not tied to a particular cancer study, code that checks for a valid genetic_alteration_type and stable_id was relaxed when the gene panel meta file is processed.
- A minimal check for required headers is implemented in GenePanelMatrixValidator.  Additional checks as described in the class can be implemented in a future effort.
- Since a gene panel file exists outside of a study, metaImport.py was not updated as its current implementation is study centric data importing.  Providing this support is better left in the hands of its author.

### DIFF
--- a/core/src/main/scripts/importer/cbioportal_common.py
+++ b/core/src/main/scripts/importer/cbioportal_common.py
@@ -46,6 +46,7 @@ class MetaFileTypes(object):
     TIMELINE = 'meta_timeline'
     CASE_LIST = 'case_list'
     MUTATION_SIGNIFICANCE = 'meta_mutsig'
+    GENE_PANEL_MATRIX = 'meta_gene_panel_matrix'
 
 # fields allowed in each meta file type, maps to True if required
 META_FIELD_MAP = {
@@ -85,7 +86,8 @@ META_FIELD_MAP = {
         'show_profile_in_analysis_tab': True,
         'profile_name': True,
         'profile_description': True,
-        'data_filename': True
+        'data_filename': True,
+        'gene_panel': False
     },
     MetaFileTypes.CNA_LOG2: {
         'cancer_study_identifier': True,
@@ -125,7 +127,8 @@ META_FIELD_MAP = {
         'profile_description': True,
         'data_filename': True,
         'normal_samples_list': False,
-        'swissprot_identifier': False
+        'swissprot_identifier': False,
+        'gene_panel': False
     },
     MetaFileTypes.EXPRESSION: {
         'cancer_study_identifier': True,
@@ -155,7 +158,8 @@ META_FIELD_MAP = {
         'show_profile_in_analysis_tab': True,
         'profile_name': True,
         'profile_description': True,
-        'data_filename': True
+        'data_filename': True,
+        'gene_panel': False
     },
     MetaFileTypes.FUSION: {
         'cancer_study_identifier': True,
@@ -165,7 +169,8 @@ META_FIELD_MAP = {
         'show_profile_in_analysis_tab': True,
         'profile_name': True,
         'profile_description': True,
-        'data_filename': True
+        'data_filename': True,
+        'gene_panel': False
     },
     MetaFileTypes.GISTIC_GENES: {
         'cancer_study_identifier': True,
@@ -193,6 +198,12 @@ META_FIELD_MAP = {
         'genetic_alteration_type': True,
         'datatype': True,
         'data_filename': True
+    },
+    MetaFileTypes.GENE_PANEL_MATRIX: {
+        'cancer_study_identifier': True,
+        'genetic_alteration_type': True,
+        'datatype': True,
+        'data_filename': True
     }
 }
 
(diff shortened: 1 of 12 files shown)


n1zea144:
Hi @pieterlukasse @fedde-s @zheins I noticed this is still outstanding.  Can one of you take a look?

core/src/test/scripts/test_data/study_es_0/data_gene_panel_matrix.txt
```diff
@@ -0,0 +1,2 @@
+SAMPLE_ID	study_es_0_mutations
+TCGA-A1-A0SB-01	example_gene_panel_stable_id
```
fedde-s: In addition to a matrix file, I think it could be useful to add a `gene_panel` field to in one of the other `meta_` files in this test data set. The paragraph in `docs/File-Formats.md` after the one that explains gene panel matrix files seems to suggests that it's supported as an optional field in most data types.

docs/Data-Loading.md
```diff
@@ -45,6 +45,7 @@ There are just a few rules to follow:
 - cancer type files can be mandatory if the study is referring to a cancer type that does not yet exist in the DB.
 - meta files can be named anything, as long as it starts or ends with name 'meta'. E.g. meta_test, meta.test, test.meta are all fine; metal_test and metastudy are wrong.
 - data files can be named anything and are referenced by a property `data_filename` set in the meta file. 
+- gene panel filenames should start with meta_gene_panel. E.g. meta_gene_panel_impact410.
```
n1zea144: If I remember , I had to have the file start with meta_ so that it would be processed by the validator.  I think I added _gene_panel for consistency, similar to how we have adopted meta_CNA, meta_log2CNA, etc. Technically, there is no reason for the added restriction.

core/src/test/scripts/test_data/study_es_0/data_gene_panel_matrix.txt
fedde-s: I think it should at least be included in the automated test, and added to `cbioportal_common.META_FIELD_MAP`. The validator warns about keys in meta_* files that it does not recognise at all, so that you notice if, say, a file is being recognised as the wrong data type or there's a typo in one of the keys. Or are a matrix file and a property (and clinical attribute) not supported in the same study?

fedde-s:
I noticed an issue in cbioportalImporter.py changes too: the gene panel matrix file creates foreign keys to data profiles such as the mutation data set, so has to be loaded after these profiles are. By default, cbioportalImporter loads data files in arbitrary order after the cancer type, study and samples are defined. To make it work, we'd need to add a special case like Sander did in [his pull request that adds gene set scores to the loader](https://github.com/cBioPortal/cbioportal/pull/2166/files#diff-6dc4df435d4621d77be7f633aa13036a), and load it after all ‘regular’ file pairs.

This is also causing problems (even an exception before any part of the study loads) for gene panel definitions themselves, but as we agreed that these should not be loaded as part of a study we can remove its constant from the `MetaFileTypes` class and the `get_meta_file_type()` function that recognises them in a study folder.

fedde-s:
Thank you for all the effort to support them so far, though!

Sjoerd-van-Hagen:
@n1zea144 hi Ben,

Do you think you have time to implement the changes suggested by Fedde? Maybe we can be of help?

Best,

n1zea144:
Hi @Sjoerd-van-Hagen @fedde-s This fell of my radar.  Yes, I can make changes based on Fedde's suggestions.  I just need a few days.

n1zea144:
Hi @Sjoerd-van-Hagen @fedde-s It took longer for me to get to this than I thought, but I've made most/all of the updates suggested.  I hope they are done to your liking and that this PR can be merged.

Best,

core/src/main/scripts/importer/cbioportal_common.py
```diff
@@ -222,7 +234,8 @@ class MetaFileTypes(object):
     "org.mskcc.cbio.portal.scripts.ImportGisticData" : False,
     "org.mskcc.cbio.portal.scripts.ImportMutSigData" : False,
     "org.mskcc.cbio.portal.scripts.ImportProfileData" : True,
-    "org.mskcc.cbio.portal.scripts.ImportTimelineData" : True
+    "org.mskcc.cbio.portal.scripts.ImportTimelineData" : True,
+    "org.mskcc.cbio.portal.scripts.ImportGenePanelProfileMap" : True
```
n1zea144: I'll do this.  I'm replying without looking closer at the code, but I'm wondering why I haven't encountered these issue myself.   Perhaps its because I've been running cbioportalImporter.py directly?  

./core/src/main/scripts/importer/cbioportalImporter.py -jar ./scripts/target/scripts-*.jar -c import-study-data -m core/src/test/scripts/test_data/study_es_0/meta_gene_panel_matrix.txt -d core/src/test/scripts/test_data/study_es_0/data_gene_panel_matrix.txt

In any event, I'll fix the code as you suggest.

Thanks @fedde-s

core/src/main/scripts/importer/validateData.py
```diff
@@ -2646,6 +2653,14 @@ def process_metadata_files(directory, portal_instance, logger, relaxed_mode):
...
                 'Citation is required when giving a pubmed id (pmid).')
 
+        # TBD: if we are processing a gene panel meta file,
+        # we should validate the gene list.  Currently checkGeneIdentification
+        # validation is buried in Validator class which makes this difficult.
+        # Another issue to consider is whether or not to raise a flag when a
+        # gene panel getting validated possesses the same stable id as an existing
+        # gene panel within the database.
+        #if meta_file_type == cbioportal_common.MetaFileTypes.GENE_PANEL:
+
```
n1zea144: I agree with the proper fix.  The Java code is going away in the future as we will be replacing it with the spring-batch implementation of the scripts package, so I don't want to prioritize this.  I will go ahead and remove the comment.

core/src/main/scripts/importer/cbioportal_common.py
fedde-s: Ah, it looks like this logic is not being tested when using the deprecated `-c import-study-data -m <metadata_file> -d <data_file>` form instead of the `-s <study_directory>` option.

fedde-s:
Superseded by #2427.

PR #2163, opened 2017-02-27
Securing api legacy

mandawilson:
# What? Why?
Securing the legacy API.

- Adding PreAuthorize and Post filters to service methods to secure API.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend
@cBioPortal/backend

and notify them here:

### DIFF
--- a/business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
+++ b/business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
@@ -44,6 +44,8 @@ import org.mskcc.cbio.portal.persistence.SampleListMapperLegacy;
 import org.mskcc.cbio.portal.persistence.SampleMapperLegacy;
 import org.mskcc.cbio.portal.persistence.StudyMapperLegacy;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.security.access.prepost.PostFilter;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -95,7 +97,8 @@ public class ApiService {
 		return cosmicCountMapperLegacy.getCOSMICCountsByKeywords(keywords);
 	}
 
-        @Transactional
+  @Transactional
+  @PostFilter("filterObject['studyID'] == null || hasPermission(filterObject['studyID'], 'CancerStudy', 'read')")
 	public List<Map<String, String>> getMutationsCounts(Map<String,String[]> customizedAttrs, String type, Boolean per_study, List<String> studyIds, List<String> genes, List<Integer> starts, List<Integer> ends, List<String> echo) {
 
             List<Map<String, String>> results = new ArrayList<Map<String, String>>();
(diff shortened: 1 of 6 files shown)


security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -181,6 +182,15 @@ public boolean hasPermission(Authentication authentication, Serializable targetI
                 }
             }
             return true;
+        } else if ("List<SampleListId>".equals(targetType)) {
```
n1zea144: The implementation of this loop and other permission checking of various lists results in behavior such that if a single ID in a list is unauthorized, then nothing is returned.  Will this ever happen and if so, is this the behavior we want?  The alternative is to silently drop unauthorized ids.  I just wanted to raise the issue for discussion-consensus.

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -146,68 +149,25 @@ else if(type.equals("frequency"))
             
 		return results;
 	}
-        @Transactional
```
mandawilson: @jjgao @JiaoJiao123 and @adamabeshouse all confirmed the `getMutationsCountsJSON` method can be removed.

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -244,13 +206,15 @@ else if(type.equals("frequency"))
...
+  @PreAuthorize("hasPermission(#study_id, 'CancerStudy', 'read')")
 	public List<DBClinicalField> getSampleClinicalAttributes(String study_id, List<String> sample_ids) {
 		List<Integer> internal_sample_ids = sampleMapperLegacy.getSampleInternalIdsBySample(study_id, sample_ids);
 		return getSampleClinicalAttributesByInternalIds(internal_sample_ids);
 	}
 
+  // if we make this method public again, secure it
 	@Transactional
-	public List<DBClinicalField> getSampleClinicalAttributesByInternalIds(List<Integer> sample_ids) {
+	private List<DBClinicalField> getSampleClinicalAttributesByInternalIds(List<Integer> sample_ids) {
```
mandawilson: @jjgao agreed it could be made private.

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -272,13 +237,15 @@ else if(type.equals("frequency"))
...
+  @PreAuthorize("hasPermission(#study_id, 'CancerStudy', 'read')")
 	public List<DBClinicalField> getPatientClinicalAttributes(String study_id, List<String> patient_ids) {
 		List<Integer> internal_patient_ids = patientMapperLegacy.getPatientInternalIdsByPatient(study_id, patient_ids);
 		return clinicalFieldMapper.getPatientClinicalFieldsByPatientInternalIds(internal_patient_ids);
 	}
 
+  // if we make this method public again, secure it
 	@Transactional
-	public List<DBClinicalField> getPatientClinicalAttributesByInternalIds(List<Integer> patient_ids) {
+	private List<DBClinicalField> getPatientClinicalAttributesByInternalIds(List<Integer> patient_ids) {
```
mandawilson: @jjgao agreed it could be made private.

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
mandawilson: @ersinciftci and I agreed it should be that way for CancerStudy lists, because when they are passed as a parameter they should have the correct cancer study list already (they should first get the filtered list, then use that filtered list to get other information).  In general that *should* be true with other lists as well, but we definitely should think carefully about each case to make sure it is true.

mandawilson: I should make clear that the above loop is not for a PostFilter, it is for when you are checking the permission of a parameter that is a list.  So it will be when a person has submitted a list of IDs and they do not have permission to access one or more of the elements.  If we are filtering we should always use the other hasPermission that is checking individual objects, not lists.

business/src/main/java/org/mskcc/cbio/portal/service/ApiService.java
```diff
@@ -546,26 +528,31 @@ else if(type.equals("frequency"))
...
 	}
 
 	@Transactional
+  @PostFilter("hasPermission(filterObject.id, 'CancerStudy', 'read')")
 	public List<DBStudy> getStudies() {
 		return studyMapperLegacy.getAllStudies();
 	}
 
 	@Transactional
+  @PreAuthorize("hasPermission(#study_ids, 'List<CancerStudyId>', 'read')")
```
mandawilson: See my response to your comment in CancerStudyPermissionEvaluator.  It results in 403 error.  Not an empty list.  We are assuming that it is malicious in this case.

mandawilson:
Regarding the getClinicalAttributes, I agree but don't really have anything to PostFilter on.  The response is an array that has elements like this: 
  `{
    "attr_id": "MY_ATTR",
    "display_name": "My Attribute",
    "description": "My attribute description",
    "datatype": "STRING",
    "is_patient_attribute": "1",
    "priority": "1"
  }`

mandawilson:
Regarding "In some cases a 403 (forbidden) is returned when a 404 (not found) is more appropriate" you are completely right.  The hasPermission is always supposed to return false if the object it is checking is null, so the correct way to handle that is supposed to be by checking for null/empty list/whatever in the `@PreAuthorize`.

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
n1zea144: I think since we are only supporting access to web service from website, that there should not be any incorrect id's past to the endpoints.  We may want to re-think this if we make the web service programmatically accessible.

PR #2164, opened 2017-02-28
Gene set frontend

oplantalech:
# What? Why?
This PR is part of the Gene Sets implementation done by The Hyve.

* Query page:
       - Enable GSVA profile selection in the query page.
       - Enable gene set selection in the query page.
       - Improved and extend validation triggered by submit button on query  page
       - Implement pop-up to select gene sets from hierarchical tree
       - Allow to query only gene sets
* Plots tab:
      - Enable option to select GSVA profile data in the Plots tab
      - Respective changes in GetProfileDataJSON.java backend
      - Respective changes in GetGeneticProfileDataJSON.java backend
      - *Summary*: 12 files in `webapp/js/src/plots-tab` and some backend files
* Co-expression tab:
      - Enable option to select GSVa profile data in the Co-expression tab
      - Respective changes in GetCoExpressionJSON.java backend
      - Respective changes in GetGeneticProfileDataJSON.java backend
      - *Summary*: 4 files in `webapp/js/src/co-exp` and some backend files
* Datasets page:
      - Add GSVA statistics to the Datasets page (how many samples have GSVA score)

## How to test?
* In [`docs/Import-Gene-Sets.md`](https://github.com/cBioPortal/cbioportal/blob/geneset_importer/docs/Import-Gene-Sets.md) we have included both a [quick example](https://github.com/cBioPortal/cbioportal/blob/geneset_importer/docs/Import-Gene-Sets.md#quick-example) to import gene sets, gene set hierarchy, gene set test study, as well as a complete description of the import process.
* An updated Seed-DB to 2.1.0, including all 18025 MSigDB gene sets, a hierarchical tree yaml file and a loadable test study (`gbm_tcga`) with GSVA scores and pvalues can be found in [this PR to Datahub](https://github.com/cBioPortal/datahub/pull/25).
* At a later time we will provide a test server that includes all gene set PRs


Sjoerd-van-Hagen:
@jjgao is this one being looked at?

pieterlukasse:
@yichaoS you can also test things here: https://cbioportal.thehyve.net/cbioportal/

pieterlukasse:
@yichaoS here are screenshots of the new functionality:

Extra "Genetic Profile" for GSVA scores. When checked, an extra "Select Gene Sets" box appears on the bottom of the query page:
image

Gene sets selection pop-up, showing a hierarchy of gene sets for which GSVA scores were calculated in this study (this is a way to select significant gene sets, similar to MUTSIG / GISTIC buttons in Genes box):
image

:information_source:  In the query page we also added a possibility to query only gene sets, so it is allowed to leave Genes box empty. 

Selected Gene sets appear as heatmap tracks:
image

The heatmap can be clustered (just like any other expression heatmap):
image

Gene sets can be selected in the "Plots" tab:
image

Correlation of expression data vs Gene set scores can be found in the "Co-Expression" tab (make sure you select the U133 microarray profile - which is the one that gives results here). 
image

There is also one sub-tab per queried gene set:
image

GSVA scores are available on the download tab:
image

And on the download section in the query page:
image

yichaoS:
apr 17 2017 13 56
Table width change once gene set is selected

yichaoS:
screen shot 2017-04-17 at 1 46 00 pm
Axis and "?" icon overlaps sometimes.

oplantalech:
@yichaoS I have corrected the issues you pointed out, and also others, the most important is that now the legend in the plots of the Plots and Co-expression tabs only refer to genes, and not gene sets.

yichaoS:
@oplantalech A lot of comments are still pending, including the missing "cytoband" column, which is quiet important.

core/src/main/java/org/mskcc/cbio/portal/servlet/GetProfileDataJSON.java
```diff
@@ -169,10 +172,21 @@ protected void doPost(HttpServletRequest httpServletRequest,
                 //Get raw data (plain text) for each profile
                 for (String geneticProfileId: geneticProfileIds) {
                     try {
-                        ArrayList<String> tmpProfileDataArr = GeneticAlterationUtil.getGeneticAlterationDataRow(
-                                gene,
-                                internalSampleIds,
-                                DaoGeneticProfile.getGeneticProfileByStableId(geneticProfileId));
+                    	ArrayList<String> tmpProfileDataArr;
+                    	GeneticProfile geneticProfile = DaoGeneticProfile.getGeneticProfileByStableId(geneticProfileId);
+                    	if (geneticProfile.getGeneticAlterationType().equals(GeneticAlterationType.GENESET_SCORE)) {
```
pieterlukasse: I think that the assumption here was that geneticProfileIds is always of length 1, but the code does not guarantee this, so yes...it is a bit hacky. There is a type for entities, see `/core/src/main/java/org/mskcc/cbio/portal/model/EntityType.java` (actually used here as well, see EntityType.GENESET). Perhaps we can use this. So the web-service would need an extra entity type argument and fetch the data based on that rather than trying to infer via the genetic profile.

portal/src/main/webapp/js/src/cbio-util.js
```diff
@@ -339,15 +339,27 @@ cbio.util = (function() {
...
     var safeProperty = function(property) {
-        return property.replace(/[^a-zA-Z0-9-]/g, '_');
+        return uidProperty(property);
+    };
+
+    var uidPropertyMap = {};
+    var uidCount = 0;
+    var uidProperty = function(property) {
+        if (!uidPropertyMap[property]) {
+            uidPropertyMap[property] = "uid"+uidCount;
```
oplantalech: Maybe the description needs to be revised. Gene set names contain characters like dots, which cause problems if they are used as ids for HTML elements. By using the original function, the problematic characters are replaced, but this does not necessarily make them unique (for example, "GSVA/1" and "GSVA.1" would both be transformed to "GSVA_1").  Therefore, we decided to change the function to explicitly make a new unique id each time the function is called.

portal/src/main/webapp/js/src/dynamicQuery.js
```diff
@@ -790,6 +841,9 @@ function geneSetSelected() {
 
         //  Set the gene list text area
         $("#gene_list").val(gene_set.gene_list);
+
+        // Validate the genes in the gene list
+        GeneSymbolValidator.initialize();
```
oplantalech: This validator is only initialized when the user selects genes from the list "User-defined List". A gene that belongs to these pre-defined lists might not be in the database, and this case is not currently validated. That's why we're initializing the validator here, to check for these specific cases.

portal/src/main/webapp/js/src/co-exp/coExpression.js
```diff
@@ -275,52 +334,179 @@ var CoExpView = (function() {
...
                     $(event.target.parentNode).addClass('row_selected');
                     //Get the gene name of the selected row
-                    var aData = coExpTableInstance.fnGetData(this);
+                    var aData = $("#" + Names.tableId).dataTable().fnGetData(this);
                     if (null !== aData) {
                         $("#" + Names.plotId).empty();
                         $("#" + Names.plotId).append("<img style='padding:220px;' src='images/ajax-loader.gif' alt='loading' />");
                         var coexpPlots = new CoexpPlots();
-                        coexpPlots.init(Names.plotId, geneId, aData[0], aData[2], aData[3], $("#coexp-profile-selector :selected").val());
+                        var profile1Id = null;
```
oplantalech: Since this leads to the building of the co-expression plots, I renamed the variables `entity_x_profile` and `entity_y_profile`, and in `GetGeneticAlterationDataJSON` I also renamed `entity1` and `entity2` to `entity_x` and `entity_y`. This names make more sense.

oplantalech:
@yichaoS Please check that all your suggestions have been implemented. For the frontend testing, I show you a co-expression screenshot with the same study in rc:
captura de pantalla 2017-04-24 a les 18 20 35

And with the new implementation proposed in this PR:
captura de pantalla 2017-04-24 a les 18 26 43

portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
```diff
@@ -85,47 +85,52 @@
...
+                    	
+                    	Boolean genesetQuery = false;
+		               	if (localGeneSetList.length() > 0 && localGeneList.length() == 0) {
+                    		genesetQuery = true;
+                    	}
+                    	
                         String currentValue = values[i].trim();
 
-                        if (currentValue.contains("mutation") && !disabledTabs.contains("mutations"))
+                        if ((currentValue.contains("mutation") && !disabledTabs.contains("mutations")) && !genesetQuery)
```
zhx828: Looks like `genesetQuery ` will control lots of tabs visibility. In this case, shouldn't we just update disabledTabs array?

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1040,7 +1041,16 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	    return this.oql_query;
 	},
 	'getQueryGenes': function () {
-	    return OQL.genes(this.oql_query);
+		if (this.oql_query.trim().length > 0) {
```
oplantalech: If you look at [line 179 from `global_variables.js`](https://github.com/thehyve/cbioportal/blob/geneset_frontend/portal/src/main/webapp/WEB-INF/jsp/global/global_variables.jsp#L179), `converted_oql` is generated by trimming another string. This `converted_oql` is what is passed to `initDatamanager` as `oql_query`. So it is not possible that `this.oql_query` is null:
```js
var converted_oql = oql_html_conversion_vessel.textContent.trim();
```

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1040,7 +1041,16 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
+		    return OQL.genes(this.oql_query);
+		} 
+		return null;
+	},
+	'getQueryGenesets': function () {
+		if (this.geneset_ids.trim().length > 0) {
```
oplantalech: Similarly to `this.oql_query`, `geneset_ids` comes also from trimming another string ([line 180 from `global_variables.js`](https://github.com/thehyve/cbioportal/blob/geneset_frontend/portal/src/main/webapp/WEB-INF/jsp/global/global_variables.jsp#L180)):
```js
var genesets = '<%=WebserviceParserUtils.getGenesetIds(request)%>'.trim();
```

portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
oplantalech: well, `disabledTabs` and `genesetQuery` are just pointing to different things (the tabs that are disabled and whether we have gene sets in the query or not), so I would keep it like this. How do you suggest to update `disableTabs`?

oplantalech: I pushed a commit which adds some information on why `genesetQuery` is created.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
zhx828: Good to know. Thanks.

portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
zhx828: I totally see your point. But the sole purpose of `disabledTabs` here is to control which tab to show.  Instead of putting `genesetQuery` to each if-else condition I would prefer to do something like this
```
if(genesetQuery) {
    var array = ['IGV', ...];
    disabledTabs = _.union(disabledTabs, array);
}
```
What do you think?

oplantalech: @zhx828 I have rewritten the code. I think now looks nicer, thank you for your suggestion.

PR #2187, opened 2017-03-08
Solution for hiding bitly option when bitly.user is not set

pieterlukasse:
# What? Why?
Solution for hiding bitly option when bitly.user is not set.

- new getter for existing portal.property `bitly.user`. This new getter is used to check whether the property is set, and if the property is not set, the bitly option is removed from the UIs accordingly. 

# Notify reviewers
@jjgao @n1zea144


sandertan:
I tested this and it does what it's supposed to do. Great solution because it fixes an ugly scenario when bitly.user is not set.

PR #2186, opened 2017-03-08
Gene set oncoprint tracks

fedde-s:
This PR is part of the Gene Sets implementation done by The Hyve.

Changes proposed in this pull request:
* Add gene set score heatmap tracks to the Oncoprint

  Based on the queried gene set score profiles, add a special track group
  underneath all other tracks, that displays the scores in a heatmap with
  a distinct colour scheme.

* Enable Oncoprint queries with only gene sets (and no genes)

## Screenshot ##
Screenshot of an Oncoprint view containing a clustered heatmap of gene set scores

## How to test ##
* In [`docs/Import-Gene-Sets.md`](https://github.com/cBioPortal/cbioportal/blob/geneset_importer/docs/Import-Gene-Sets.md) we have included both a [quick example](https://github.com/cBioPortal/cbioportal/blob/geneset_importer/docs/Import-Gene-Sets.md#quick-example) to import gene sets, gene set hierarchy, gene set test study, as well as a complete description of the import process.
* An updated Seed-DB to 2.1.0, including all 18025 MSigDB gene sets, a hierarchical tree yaml file and a loadable test study (`gbm_tcga`) with GSVA scores and pvalues can be found in [this PR to Datahub](https://github.com/cBioPortal/datahub/pull/25).
* At a later time we will provide a test server that includes all gene set PRs

PR #2190, opened 2017-03-09
Gene set oncoprint track expansion

fedde-s:
This PR is part of the Gene Sets implementation done by The Hyve.

- Add track expansion support to  Oncoprint classes

  Make the classes of the Oncoprint aware of tracks that are added to the
  Oncoprint as the expansion of other tracks.

- Generalise rule sets for heatmap track groups

  Separate the selection of a rule set and legend from the logic for
  adding heatmap tracks selected by profile from the menu bar, so
  that the legend can be reused for expression heatmaps added by
  expanding gene set tracks.

- Enable expansion functionality for gene set tracks

  Register callbacks to request genes in the gene set by correlation and
  insert their expression heatmap tracks underneath the gene set track,
  when the expand button is pressed for the gene set track.

## Screenshot ##

screen shot 2017-03-09 at 17 24 50 0100

## How to test ##
* In [`docs/Import-Gene-Sets.md`](https://github.com/cBioPortal/cbioportal/blob/geneset_importer/docs/Import-Gene-Sets.md) we have included both a [quick example](https://github.com/cBioPortal/cbioportal/blob/geneset_importer/docs/Import-Gene-Sets.md#quick-example) to import gene sets, gene set hierarchy, gene set test study, as well as a complete description of the import process.
* An updated Seed-DB to 2.1.0, including all 18025 MSigDB gene sets, a hierarchical tree yaml file and a loadable test study (`gbm_tcga`) with GSVA scores and pvalues can be found in [this PR to Datahub](https://github.com/cBioPortal/datahub/pull/25).
* At a later time we will provide a test server that includes all gene set PRs

### DIFF
--- a/portal/src/main/webapp/js/src/oncoprint/setup.js
+++ b/portal/src/main/webapp/js/src/oncoprint/setup.js
@@ -1,3 +1,5 @@
+/* globals cbio, QuerySession */
+/* jshint devel: true, laxbreak: true */
 var stringListUnique = function(list) {
     var seen = {};
     var ret = [];
@@ -886,6 +888,44 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 	    return done.promise();
 	};
 	
+	var HEATMAP_RULE_SET_PARAMS = {
+		'type': 'gradient',
+		'legend_label': "Expression heatmap",
+		'value_key': 'profile_data',
+		'value_range': [-3, 3],
+		'colors': [[0,0,255,1],[0,0,0,1],[255,0,0,1]],
+		'value_stop_points': [-3,0,3],
+		'null_color': 'rgba(224,224,224,1)'
+	};
+	
+	/**
+	 * Makes a track reuse the legend of any previous heatmap tracks.
+	 *
+	 * If any tracks previously given to this function still
+	 * exist, make the new track share the rule set and legend
+	 * of the existing ones.
+	 *
+	 * @param {number} track_id - ID of a track to use the shared legend
+	 */
+	var useExistingHeatmapRuleSet = (function() {
+	    var previous_hmtrack_ids = [];
+	    return function(track_id) {
+		var existing_track_ids = oncoprint.model.getTracks(),
+		    existing_hmtrack_ids;
+		// make an array of heatmap track handles that still exist
+		existing_hmtrack_ids = previous_hmtrack_ids.filter(
+		    function (track_id) {
+			return existing_track_ids.indexOf(track_id) !== -1;
+		    });
+		// reuse the ruleset of an existing track if any
+		if (existing_hmtrack_ids.length > 0) {
+		    oncoprint.shareRuleSet(existing_hmtrack_ids[0], track_id);
+		}
+		existing_hmtrack_ids.push(track_id);
+		previous_hmtrack_ids = existing_hmtrack_ids;
+	    };
+	})();
+	
 	var makeRemoveGenesetTrackHandler = function(geneset_id) {
 	    return function(track_id) {
 		delete State.geneset_to_track_id[geneset_id];
@@ -1037,20 +1077,7 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 	    
 	    'is_minimap_shown': false,
 	    
-	    'getAnyHeatmapTrackId': function() {
-		var ret = null;
-		var heatmap_profile_ids = Object.keys(this.heatmap_track_groups);
-		for (var i=0; i<heatmap_profile_ids.length; i++) {
-		    var gene_to_track_id = this.heatmap_track_groups[heatmap_profile_ids[i]].gene_to_track_id;
-		    var genes_with_tracks = Object.keys(gene_to_track_id);
-		    if (genes_with_tracks.length > 0) {
-			ret = gene_to_track_id[genes_with_tracks[0]];
-			break;
-		    }
-		}
-		return ret;
-	    },
-	    'getNewTrackGroupId': function() {
+	    'getNewTrackGroupId': function () {
 		// Return 1 more than the max heatmap track group id, minimum 2
 		var heatmap_genetic_profiles = Object.keys(this.heatmap_track_groups);
 		return Math.max(Math.max.apply(null, heatmap_genetic_profiles.map(function(id) { return State.heatmap_track_groups[id].track_group_id; })) + 1, 2);
@@ -1163,38 +1190,29 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 	    },
 	    'addHeatmapTrack': function (genetic_profile_id, gene) {
 		oncoprint.suppressRendering();
-		State.heatmap_track_groups[genetic_profile_id] = State.heatmap_track_groups[genetic_profile_id] || {
-		    genetic_profile_id: genetic_profile_id,
-		    track_group_id: State.getNewTrackGroupId(),
-		    gene_to_track_id: {}
-		};
 		var heatmap_track_group = State.heatmap_track_groups[genetic_profile_id];
-		    var track_params = {
-			'rule_set_params': {
-			    'type': 'gradient',
-			    'legend_label': "Heatmap",
-			    'value_key': 'profile_data',
-			    'value_range': [-3, 3],
-			    'colors': [[0,0,255,1],[0,0,0,1],[255,0,0,1]],
-			    'value_stop_points': [-3,0,3],
-			    'null_color': 'rgba(224,224,224,1)'
-			},
-			'has_column_spacing': false,
-			'track_padding': 0,
-			'label': gene,
-			'target_group': heatmap_track_group.track_group_id,
-			'removable': true,
-			'removeCallback': makeRemoveHeatmapHandler(genetic_profile_id, gene),
-			//'sortCmpFn': function(d1, d2) {return 0;},
-			'description': gene + ' data from ' + genetic_profile_id,
-			//'track_group_header': genetic_profile_id
+		if (typeof heatmap_track_group === 'undefined') {
+		    heatmap_track_group = State.heatmap_track_groups[genetic_profile_id] = {
+			    genetic_profile_id: genetic_profile_id,
+			    track_group_id: State.getNewTrackGroupId(),
+			    gene_to_track_id: {}
 		    };
-		    var any_hm_id = State.getAnyHeatmapTrackId();
-		    var new_hm_id = oncoprint.addTracks([track_params])[0];
-		    heatmap_track_group.gene_to_track_id[gene] = new_hm_id;
-		    if (any_hm_id !== null) {
-			oncoprint.shareRuleSet(any_hm_id, new_hm_id);
-		    }
+		}
+		var track_params = {
+		    'rule_set_params': HEATMAP_RULE_SET_PARAMS,
+		    'has_column_spacing': false,
+		    'track_padding': 0,
+		    'label': gene,
+		    'target_group': heatmap_track_group.track_group_id,
+		    'removable': true,
+		    'removeCallback': makeRemoveHeatmapHandler(genetic_profile_id, gene),
+		    //'sortCmpFn': function(d1, d2) {return 0;},
+		    'description': gene + ' data from ' + genetic_profile_id,
+		    //'track_group_header': genetic_profile_id
+		};
+		var new_hm_id = oncoprint.addTracks([track_params])[0];
+		useExistingHeatmapRuleSet(new_hm_id);
+		heatmap_track_group.gene_to_track_id[gene] = new_hm_id;
 		URL.update();
 		oncoprint.releaseRendering();
 		return new_hm_id;
--- a/portal/src/main/webapp/js/src/oncoprint/setup.js
+++ b/portal/src/main/webapp/js/src/oncoprint/setup.js
@@ -2022,6 +2022,8 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 		//reset all tracks of heatmaps that were clustered to their original order:
 		for (var clusteredHeatmapGroup in State.trackIdsInOriginalOrder) {
 			if (State.trackIdsInOriginalOrder.hasOwnProperty(clusteredHeatmapGroup)) {
+				// first remove any expansion tracks, as these are not part of the saved order
+				oncoprint.removeAllExpansionTracksInGroup(clusteredHeatmapGroup);
 				oncoprint.setTrackGroupOrder(clusteredHeatmapGroup, State.trackIdsInOriginalOrder[clusteredHeatmapGroup]);
 			}
 		}
@@ -2134,6 +2136,8 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 					State.clustered_by_profile_id = genetic_profile_id;
 					//sort according to order found in the clustering results:
 					State.sorting_by_given_order = true;
+					// remove any expansion tracks, as these are not part of the clustering order
+					oncoprint.removeAllExpansionTracksInGroup(heatmap_track_group_id);
 					//store original order before clustering:
 					var trackIdsInOriginalOrder = oncoprint.model.getTrackGroups()[heatmap_track_group_id];
 					State.trackIdsInOriginalOrder[heatmap_track_group_id] = trackIdsInOriginalOrder;
--- a/portal/src/main/webapp/js/src/oncoprint/setup.js
+++ b/portal/src/main/webapp/js/src/oncoprint/setup.js
@@ -1393,7 +1393,6 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 			'track_padding': 0,
 			'label': track_geneset_id,
 			'target_group': this.GENESET_HEATMAP_TRACK_GROUP_INDEX,
-			'removable': true,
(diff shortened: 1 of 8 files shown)


fedde-s:
This functionality still contains a bug when clustering or consequently un-sorting the gene set tracks while expanded, I'm working with Pieter to fix this.

fedde-s:
I added two commits to fix an issue with expansion tracks and clustering. The first of the commits adjusts the implementation in the Oncoprint classes in this same pull request, so I could squash it before merging to simplify the commit history.

fedde-s:
@adamabeshouse, for clarity, this pull request builds on the one that implements gene set score tracks themselves, which you already reviewed. I would appreciate your opinion on this one as well, as it adds a feature to the Oncoprint model class.

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
```diff
@@ -144,6 +144,11 @@ var OncoprintModel = (function () {
 	this.track_active_rules = {}; // from track id to active rule map (map with rule ids as keys)
 	this.track_info = {};
 	this.track_has_column_spacing = {}; // track id -> boolean
+	this.track_expansion_init_callback = {}; // track id -> function that initializes expansion data if applicable
+	this.track_expansion_genes = {}; // track id -> array of gene data to expand if initialized
+	this.track_expansion_index = {}; // track id -> index in track_expansion_genes[track_id] if initialized
```
adamabeshouse: I don't understand the need for this field `track_expansion_index` - it seems extremely hacky to me.

fedde-s:
Thanks, @adamabeshouse. I agree with the overly specific ‘gene’ phrasing: I put it in there while prototyping and wondered whether it would be worth rephrasing. Will change.

The ‘gene’ entries in the array `oncoprintModel.track_expansion_genes` are already opaque objects to the oncoprint code, initialised and parsed only by callbacks set from `CreateCBioPortalOncoprintWithToolbar()` in `setup.py`. The knowledge I did put in the Oncoprint model is that each object in the array will cause the callback to add one track, and that giving it the first one again (by resetting the index if all of a track's expansion tracks have been removed) will cause it to expand the same first track. Do you think that the callbacks should address this array and index directly in the `State` object in `CreateCBioPortalOncoprintWithToolbar()`, moving the logic to reset (the index) to the first gene to the expansion tracks’ `removeCallback`s?

adamabeshouse:
Thanks @fedde-s, I think I understand now. Everything makes more sense and I think the only thing that I object to is that I think the track expansion callback should not take any arguments (i.e. the gene metadata) - the oncoprint should call it with no arguments and its arguments should be baked into the passed-in callback via closure. This allows you to get rid of the (in my opinion) confusing `track_expansion_genes` and `track_expansion_index`, and move all that related logic out of the oncoprint.

These are just aesthetic preferences I have:
- A clearer name for `track_expansion_init_callback` might be `track_expand_callback` since it looks like we run it every time the track is expanded, not just on init.
- A clearer name for `track_expansion_callback` might be `track_init_callback` since it's being run for the tracks which are shown as part of the expansion, as opposed to the track that is being expanded.

fedde-s:
I have also been thinking about moving`track_expansion_genes` and `track_expansion_index` out of the `OncoprintModel`, but it would mean that the logic of restarting from the top of the list after removing all expansion tracks would move to a `removeCallback` defined in the same closure: any potential future uses of the expansion callback infrastructure would have to provide their own logic. Do you think it would be an improvement if I did that?

`track_expansion_init_callback` should only called if it has not been called before for the same track ID if I have not made a mistake; `expandTrack` only calls if it has not already created a Promise onto which to chain `.then`-calls for consecutive expansion steps to be run after the initialisation promise has been resolved. In that light, I think the `init` in its name is appropriate. Renaming the second callback to its purpose of adding expansion tracks to the Oncoprint more apparent could still be useful.

And about removing the word ‘gene’ from the expansion API as we discussed before: I did call the menu items in the `OncoprintTrackOptionsView` ‘Show genes’/‘More genes’ and ‘Remove genes’, as the user needs to know that this is about the expression of individual gene sets here. Should this be configurable via track parameters, or would that prematurely add complexity just to generalise things that we may not actually end up reusing for other purposes?

adamabeshouse:
What I mean is instead of holding onto the `track_expansion_genes` structure and referencing it, you up top create all the tracks that could be expanded, but don't put anything in them, you mark them as expansion tracks so they start hidden. Then in their onExpand callback, which you pass in from the same place that you have the track metadata, you use the track metadata which is looped into the callback's closure. That way there's no need for any complicated indexing anywhere.

I definitely think the word gene should be removed and parametrized.

fedde-s:
Listing which (and how many) genes are correlated with a gene set for the queried samples is a back-end request that takes several seconds. We cannot do this until the user has indicated that they want to expand the genes for a particular gene set, especially since a single Oncoprint heatmap may reasonably contain over a hundred gene sets with the new minimap view and clustering functionality.

adamabeshouse:
Then can we put all that logic in the onTrackExpand callback? It fetches the genes and data, then adds tracks for them, marking those as subtracks or expansion tracks or w/e. I think ALL of that should be done outside the oncoprint - managing/caching that data, etc. Oncoprint is just a visualization library.

fedde-s:
Yes, makes sense, I'll see what I can do!

fedde-s:
I added two more commits: one to make gene set tracks non-removable (see the commit message for the reason and why it's related to expansion) and one to move the handling of metadata necessary for expanding gene set tracks to a single callback in `setup.js`. @adamabeshouse, could you review if this resolves your objections?

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprinttrackoptionsview.js
```diff
@@ -192,6 +193,27 @@ var OncoprintTrackOptionsView = (function () {
 	    $dropdown.append($sort_dec_li);
 	    $dropdown.append($dont_sort_li);
 	}
+	if (model.isTrackExpandable(track_id)) {
+	    $dropdown.append($makeDropdownOption(
+		    (model.isTrackExpanded(track_id) ? 'More' : 'Show') + ' genes',
```
adamabeshouse: Please find a way to not hardcode the word 'gene'

adamabeshouse:
Looks much better at first glance. Will do a more detailed review shortly.

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprinttrackoptionsview.js
fedde-s: Good point, missed that one. It has to be clear to the user what the biological meaning of the tracks is, but it's now pretty obvious that that information should be defined in the same place as the callback that encodes this meaning.

fedde-s:
@adamabeshouse, do you think that the callback I added as a track param in the last commit is a good solution to make the word ‘gene’ configurable, as you asked?

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -886,6 +888,44 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
...
+	/**
+	 * Makes a track reuse the legend of any previous heatmap tracks.
+	 *
+	 * If any tracks previously given to this function still
+	 * exist, make the new track share the rule set and legend
+	 * of the existing ones.
+	 *
+	 * @param {number} track_id - ID of a track to use the shared legend
+	 */
+	var useExistingHeatmapRuleSet = (function() {
```
adamabeshouse: why are you doing this instead of the approach I had below with `getAnyHeatmapTrackId`?

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -1258,9 +1393,12 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 			'track_padding': 0,
 			'label': track_geneset_id,
 			'target_group': this.GENESET_HEATMAP_TRACK_GROUP_INDEX,
-			'removable': true,
 			'description': track_geneset_id + ' gene set scores from ' + genetic_profile_id,
 			'removeCallback': makeRemoveGenesetTrackHandler(track_geneset_id),
+			'expandCallback': makeGenesetExpandHandler(track_geneset_id),
+			'expandButtonTextCallback': function (is_expanded) {
```
adamabeshouse: this is good, but it's more idiomatic in this case to call it `Getter` rather than `Callback`

adamabeshouse: e.g. https://facebook.github.io/fixed-data-table/api-table.html `rowHeightGetter`

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
```diff
@@ -144,6 +145,10 @@ var OncoprintModel = (function () {
 	this.track_active_rules = {}; // from track id to active rule map (map with rule ids as keys)
 	this.track_info = {};
 	this.track_has_column_spacing = {}; // track id -> boolean
+	this.track_expandable = {}; // track id -> boolean
```
adamabeshouse: can this be inferred from the existence of `track_expansion_tracks`?

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
```diff
@@ -144,6 +145,10 @@ var OncoprintModel = (function () {
...
+	this.track_expand_callback = {}; // track id -> function that adds expansion tracks for its track if set
+	this.track_expand_button_callback = {}; // track id -> function from boolean to string if customized
```
adamabeshouse: as mentioned above, I'd call this `track_expand_button_text_getter`

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
```diff
@@ -786,12 +810,25 @@ var OncoprintModel = (function () {
 	delete this.track_sort_direction[track_id];
 	delete this.track_info[track_id];
 	delete this.track_has_column_spacing[track_id];
+	delete this.track_expandable[track_id];
+	delete this.track_expand_callback[track_id];
+	delete this.track_expansion_tracks[track_id];
```
adamabeshouse: don't forget `delete this.track_expand_button_text_getter[track_id]`

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
```diff
@@ -1012,7 +1049,54 @@ var OncoprintModel = (function () {
     OncoprintModel.prototype.isTrackSortDirectionChangeable = function (track_id) {
 	return this.track_sort_direction_changeable[track_id];
     }
+    
+    OncoprintModel.prototype.isTrackExpandable = function (track_id) {
+	// return true if the flag is defined and true
+	return Boolean(this.track_expandable[track_id]);
```
adamabeshouse: how does `Boolean` work? typically in this case I use ``!!``

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
adamabeshouse: oh I understand, it can have tracks to expand but not be allowed to expand

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
adamabeshouse: nevermind `Boolean` seems fine

portal/src/main/webapp/js/src/oncoprint/setup.js
fedde-s: Makes sense, it's a pure function rather than a handler.

portal/src/main/webapp/js/src/oncoprint/setup.js
fedde-s: `getAnyHeatmapTrackId` would not have to return *any* heatmap track ID, but specifically one which shares the same legend. That is, an RNA or protein expression heatmap track either added via the Heatmap menu (tracked in the list it used before) or one added by expanding gene set tracks (not tracked in the same list, as they share no other functionality with the former), but not a gene set heatmap track (different legend). Because of all this, I thought it would be confusing not to make the function explicitly about sharing the ruleset/legend between them.

portal/src/main/webapp/js/src/oncoprint/setup.js
fedde-s: Although it is not a getter either—it takes an argument.

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
fedde-s: `track_expansion_tracks` links a track to its already (visibly) existing tracks. This boolean, on the other hand, states whether a track option should be displayed to (run the expandCallback and) expand the track with more tracks.

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
fedde-s: I consider using the `Boolean()` function more explicit and simple than `!!`, even if it is also a constructor (when used as `new Boolean()`) for a class of objects that are truthy (because all objects are).

portal/src/main/webapp/js/src/oncoprint/setup.js
fedde-s: Do you have a cleaner idea to achieve this?

portal/src/main/webapp/js/src/oncoprint/setup.js
fedde-s: Might `expandButtonTextFn` be even more idiomatic, then? Or shall I just make it `Getter`?

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
fedde-s: Would renaming it to `track_expansion_enabled` clarify this?

adamabeshouse: yes, that sounds good.

portal/src/main/webapp/js/src/oncoprint/setup.js
adamabeshouse: oh, I see. No this is a good idea.

portal/src/main/webapp/js/src/oncoprint/setup.js
adamabeshouse: I'd say `Getter`, more specific

fedde-s: Renamed it to `Getter` now :+1:

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
fedde-s: Renamed to `_enabled`!

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
fedde-s: As mentioned above, done 😉

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintmodel.js
fedde-s: Added a delete statement.

pieterlukasse:
@adamabeshouse thanks for the review! I deployed the latest version to https://cbioportal.thehyve.net/cbioportal/ so stress testing could be done there. Maybe you can do some tests there as an example of what you mean with stress testing?

inodb:
Seems like heroku build is working

adamabeshouse:
what's going on here? chrome viewing the link @pieterlukasse just posted:
image

fedde-s:
@adamabeshouse, that is a feature introduced in #2186.

fedde-s:
@adamabeshouse, you can add gene set tracks by querying a gene set profile from the query page, just like you would query genetic alterations per gene. This allows querying gene sets without querying genes, which was an explicit requirement of the RFC. The query page functionality was introduced in #2164.

fedde-s:
I squashed the amendments I committed based on review; the resulting commits explain disjoint parts of the implementation.

adamabeshouse:
oh ok, this looks really cool.

found some issues(?, maybe they are features)
(1) if you expand a gene set, then drag to move it, the expanded genes stay where they are. But they still stay linked, so that if you remove expanded tracks they disappear without issue, so that's good.
(2) you can drag an expansion track away from its expansion region, i.e. you can move it anywhere not just among the 5 expanded tracks
(3) should there be some visual way of associating / differentiating expanded tracks? Like a pipe `|` or carat `>` at the beginning of the name?

fedde-s:
Thanks for reviewing the functionality too, @adamabeshouse! The behaviours you mentioned are not specifically designed features, but effects of adding the requested functionality to the existing code using the simplest change I could think of. Since people are waiting to use this functionality, I'd suggest to release it and track these enhancements as follow-up issues.

PR #2198, opened 2017-03-09
Delete Cancer Study By Cascade

angelicaochoa:
# What? Why?
* Added new delete cancer study method that deletes a study by cascade from DB with foreign key constraints enforced.
* Fixed a bug in `DaoCancerStudy.deleteCancerStudy(...)` method - array of SQL statements did not contain clean up statement for `cna_event` or delete `copy_number_seg_file` records referencing the cancer study being deleted

* Deprecated DaoCancerStudy.deleteCancerStudy(...) methods.
* Fixed a bug in `DaoCancerStudy.deleteCancerStudy(...)`
* Added new method that deletes a study by cascade from DB with foreign key constraints enforced.
* Added method that removes unused references in `cna_event`, `mutation_event`, and `gistic` that may have been unique to a single study that no longer exists in the database.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers

@n1zea144 
@sheridancbio 
@zheins 

and notify them here:

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
@@ -449,6 +449,12 @@ public final class DaoCancerStudy {
         }
     }
 
+    /**
+     * 
+     * @param cancerStudyStableId
+     * @throws DaoException 
+     * @deprecated
+     */
     public static void deleteCancerStudy(String cancerStudyStableId) throws DaoException
     {
         CancerStudy study = getCancerStudyByStableId(cancerStudyStableId);
@@ -481,14 +487,55 @@ public final class DaoCancerStudy {
             JdbcUtil.closeAll(DaoCancerStudy.class, con, pstmt, rs);
         }
     }
+    
+    /**
+     * Calls deleteCancerStudyByCascade() if cancer study exists by stable id.
+     * @param cancerStudyStableId
+     * @throws DaoException 
+     */
+    public static void deleteCancerStudyByCascade(String cancerStudyStableId) throws DaoException {
+        CancerStudy study = getCancerStudyByStableId(cancerStudyStableId);
+        if (study != null) {
+            deleteCancerStudyByCascade(study.getInternalId());
+        }
+    }
 
+    /**
+     * Deletes a cancer study by internal id from db with foreign key constraints enforced.
+     * @param internalCancerStudyId
+     * @throws DaoException 
+     */
+    public static void deleteCancerStudyByCascade(int internalCancerStudyId) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            pstmt = con.prepareStatement("DELETE FROM cancer_study WHERE cancer_study_id = ?");
+            pstmt.setInt(1, internalCancerStudyId);
+            int rows = pstmt.executeUpdate();
+            // throw dao exception if no rows affected by update
+            if (rows == 0) {
+                throw new DaoException("deleteCancerStudyByCascade() failed: No rows affected");
+            }
+            removeCancerStudyFromCache(internalCancerStudyId);
+        } catch (SQLException e) {
+            throw new DaoException(e);
+        } finally {
+            JdbcUtil.closeAll(DaoCancerStudy.class, con, pstmt, rs);
+        }
+        reCacheAll();
+        System.out.println("deleted study:\nID: "+internalCancerStudyId);
+    }
+    
     /**
      * Deletes the Specified Cancer Study.
      *
      * @param internalCancerStudyId Internal Cancer Study ID.
      * @throws DaoException Database Error.
+     * @deprecated
      */
-    static void deleteCancerStudy(int internalCancerStudyId) throws DaoException {
+    public static void deleteCancerStudy(int internalCancerStudyId) throws DaoException {
         Connection con = null;
         PreparedStatement pstmt = null;
         ResultSet rs = null;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -449,6 +449,12 @@ public static void deleteAllRecords() throws DaoException {
         }
     }
 
+    /**
+     * 
+     * @param cancerStudyStableId
+     * @throws DaoException 
+     * @deprecated
```
sheridancbio: right

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -481,12 +487,49 @@ public static void deleteCancerStudy(String cancerStudyStableId) throws DaoExcep
...
+            deleteCancerStudyByCascade(study.getInternalId());
+        }
+    }
 
     /**
+     * Deletes a cancer study by internal id from db with foreign key constraints enforced.
+     * @param internalCancerStudyId
+     * @throws DaoException 
+     */
+    static void deleteCancerStudyByCascade(int internalCancerStudyId) throws DaoException {
```
sheridancbio: this probably should have a visibility specifier (public or private). Public is probably good.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -481,12 +487,49 @@ public static void deleteCancerStudy(String cancerStudyStableId) throws DaoExcep
...
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            pstmt = con.prepareStatement("DELETE FROM cancer_study WHERE cancer_study_id = ?");
+            pstmt.setInt(1, internalCancerStudyId);
+            pstmt.executeUpdate();
```
sheridancbio: I think executeUpdate() is going to return the number of records altered. If this returns zero, then the deletion did not happen, and we probably should throw a DaoException to let the caller know that the call failed.
("Could not find cancer_study_id in table cancer_study")
An interesting question is --- when it succeeds, I wonder if the returned value is 1 (for the one record deleted from cancer_study) or a bigger number (for all the cascaded deletes that happen across all tables). I have no idea what the behavior is, but would guess we will get the bigger number.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -481,12 +487,49 @@ public static void deleteCancerStudy(String cancerStudyStableId) throws DaoExcep
...
+            removeCancerStudyFromCache(internalCancerStudyId);
```
sheridancbio: do we need both an explicit removeCancerStudyFromCache(id), and also the more general reCacheAll() below? I'm not familiar with reCacheAll .. but if it deletes the whole cache and then recaches maybe this is not needed here?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -515,6 +558,7 @@ static void deleteCancerStudy(int internalCancerStudyId) throws DaoException {
                 "delete from clinical_patient where INTERNAL_ID IN (select INTERNAL_ID from patient where CANCER_STUDY_ID=?);",
                 "delete from patient where CANCER_STUDY_ID=?;",
                 "delete from copy_number_seg where CANCER_STUDY_ID=?;",
+                "delete from copy_number_seg_file where CANCER_STUDY_ID=?;",
```
sheridancbio: bugfix for old function? good.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
zheins: I think recacheAll needs to be called to update the other DaoObjects if subsequent operations are going to be performed in this transaction. The front end cache doesn't get updated, but at least this way the other maps will be for any further dao calls.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
zheins: Checking if this is greater than 0 would be a good idea, but not a huge deal. Small detail.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
zheins: Are there any other tables that have gotten missed over time? I have a list I can share with you, maybe we should check if any other tables need deleting.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -481,14 +487,91 @@ public static void deleteCancerStudy(String cancerStudyStableId) throws DaoExcep
...
+    private static void deleteUnusedReferencesFromDb() throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            
+            // sql statements to remove unused references in mutation_event, cna_event, 
+            // and gistic after a study is deleted. 
+            String[] sqls = {
```
zheins: I thought we decided to leave this out for now and research a better solution? It's fine to include though.

PR #2209, opened 2017-03-10
Add natural sort for sample label ordering

inodb:
Make the sample label of e.g. SAMPLE11 higher than SAMPLE2 by using natural
sort instead of lexicographic sort.

@jjgao


core/src/main/java/org/mskcc/cbio/portal/util/NaturalOrderComparator.java
```diff
@@ -0,0 +1,187 @@
...
+        {
+            return 0;
+        }
+        else
+        {
+            return s.charAt(i);
+        }
+    }
+
+    public static void main(String[] args)
```
jjgao: maybe put it in unit test?

inodb:
@jjgao Moved the main function to a test. Seems to work well.

I've added a card on Trello for the natural sort on the new frontend (https://github.com/cBioPortal/cbioportal-frontend/issues/210). There's indeed a library there that does the same thing: https://www.npmjs.com/package/javascript-natural-sort

Issue #2173, opened 2017-03-06
Small issues in previous migration script


pieterlukasse:
The genetic entity `migration.sql` script has 2 minor issues which should be solved.
1- genetic_entity_id field is added to the `gene` table in a different order than found in `cgds.sql`. This can lead to problems when dumping a migrated DB and trying to load it again in another fresh DB schema based just on `cgds.sql`
2- The foreign key added to the `gene` table is named `fk_gene_1` while in `cgds.sql` this foreign key constraint is not named. Would be nice to align this (although it does not really lead to issues now). Another difference is that this same FK is CASCADE on delete in `cgds.sql`, while in `migration.sql` it is not. It is important to align this one to CASCADE. 

The fix will involve a new migration step, something like this (except not all DBs have `fk_gene_1`, so this step needs to be conditional):
```
-- update gene table to use genetic_element (after HUGO_GENE_SYMBOL, reflecting same order as in cgds.sql...this is important for exporting/importing DBs later on):
ALTER TABLE `gene` 
CHANGE COLUMN `GENETIC_ENTITY_ID` `GENETIC_ENTITY_ID` INT(11) NOT NULL AFTER `HUGO_GENE_SYMBOL`;

ALTER TABLE `gene` 
DROP FOREIGN KEY `fk_gene_1`;

ALTER TABLE `gene` 
ADD FOREIGN KEY (`GENETIC_ENTITY_ID`)
  REFERENCES `genetic_entity` (`ID`)
  ON DELETE CASCADE
  ON UPDATE NO ACTION;
```

Notify: @sandertan

sheridancbio:
On issue 1, my concept would be to use cgds.sql always as the definition of the schema. If the migration script does not properly construct a matching schema when being applied to an earlier database schema then that should be considered a bug in migration.sql and can be directly fixed without creating a new db_schema_version. I noticed this difference too, and I had intended to add the "AFTER `HUGO_GENE_SYMBOL`" clause you indicated above in a new PR. (I also wanted to adjust couple of other small inconsistencies).

[...]

PR #2210, opened 2017-03-12
Adjust migration.sql to match cgds.sql history

sheridancbio:
Partial Fixes for issue #2173

- order of fields in gene table corrected
- foreign key cascade clause made to match
- also, cgds.sql does not name constraints, so that has been removed. We should investigate the naming of constraints and decide to apply this globally or not.

In this PR, migration.sql is made consistent with cgds.sql (taking cgds.sql as the definition of the table schemas).
In a future PR, some changes to cgds.sql and migration.sql will likely be added as well, potentially adding custom names to the foreign key constraints. Note that some installers who have previously used the erroneous migration.sql script will have non-matching database schemas in the databases .. this will include a different ordering of columns in the gene table. Installers will need to correct these differences manually if they cause problems. (Probably no problems will occur for normal use)

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@cBioPortal/backend
@pieterlukasse 
@jjgao

### DIFF
--- a/db-scripts/src/main/resources/migration.sql
+++ b/db-scripts/src/main/resources/migration.sql
@@ -213,7 +213,7 @@ CREATE TABLE `genetic_entity` (
 );
 -- update gene table to use genetic_element:
 ALTER TABLE `gene` 
-ADD COLUMN `GENETIC_ENTITY_ID` INT NULL;
+ADD COLUMN `GENETIC_ENTITY_ID` INT NULL AFTER `HUGO_GENE_SYMBOL`;
 
 -- add temporary column to support migration:
 ALTER TABLE `genetic_entity` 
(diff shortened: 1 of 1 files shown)


sheridancbio:
I have changed the base branch of the PR from branch 'rc' to branch 'release-1.5.0'

sheridancbio:
I see this line:
 ADD COLUMN `GENETIC_ENTITY_ID` INT NULL;
Just to verify, the line means that "NULL" is the default value for the field?
I wonder if it could / should be [for clarity]:
ADD COLUMN `GENETIC_ENTITY_ID` INT DEFAULT NULL;
would this be appropriate?

pieterlukasse:
@sheridancbio the way I read it is that NULL is just making it explicit that the field is nullable. This is syntax generated by mysql workbench. But in this case it is not so relevant since the field is made NOT NULL later in the migration script:
```sql
-- add UQ and FK constraint for GENETIC_ENTITY_ID in gene table:
ALTER TABLE `gene` 
CHANGE COLUMN `GENETIC_ENTITY_ID` `GENETIC_ENTITY_ID` INT NOT NULL,
ADD UNIQUE INDEX `GENETIC_ENTITY_ID_UNIQUE` (`GENETIC_ENTITY_ID` ASC);
```

PR #2214, opened 2017-03-14
Add docker support

inodb:
Add docker support. Update documentation

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.


inodb:
@jjgao @n1zea144 @fedde-s @pieterlukasse  LGTM :+1:

fedde-s:
At The Hyve we're internally already using a variant that builds at build time instead of at runtime, maintained at https://github.com/thehyve/cbioportal-docker . But that has some limitations for which we haven't integrated solutions yet. Among other things, all configuration including database passwords for the data loading scripts (`portal.properties`), the webserver (`context.xml`) *and* the integration tests of the `core` package (`applicationContext-dao.xml`) have to be specified at build time.

I see that this pull request includes some of the lessons that we learned both in our version and the version that has been partially documented in the official docs for a while now though, good job! I have not tested it, but overall it looks like an improvement.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    && rm -rf $CATALINA_HOME/webapps/examples \
+    && rm -rf /var/lib/apt/lists/*
+#======== Configure cBioPortal ===========================#
+COPY . $PORTAL_HOME
+WORKDIR $PORTAL_HOME
+EXPOSE 8080
+#======== Build cBioPortal on Startup ===============#
+CMD mvn -DskipTests clean install \
+     && cp $PORTAL_HOME/portal/target/cbioportal*.war $CATALINA_HOME/webapps/cbioportal.war \
+     && find $PWD/core/src/main/scripts/ -type f -executable \! -name '*.pl'  -print0 | xargs -0 -- ln -st /usr/local/bin \
```
fedde-s: I think there's no need to link these scripts to PATH in a default command that only starts the Tomcat web server afterwards. The reason we included this line in some other Dockerfiles is to run scripts such as the study loading script `metaImport.py` in a container *instead* of the web server.

README.html
```diff
@@ -0,0 +1,1028 @@
+<!DOCTYPE html><html><head><meta charset="utf-8"><style>body {
```
fedde-s: What is the purpose of this file? Is it just a copy of `README.md` with the addition of a large custom style sheet?

Dockerfile
fedde-s: Would it be an improvement to use an `ENTRYPOINT` line to specify that cBioPortal always has to be compiled and deployed before running another command, and a `CMD` to specify Tomcat as the default command to run afterwards?

README.html
inodb: yeah this should prolly be removed

Dockerfile
inodb: @alexsigaras should we remove this line?

inodb:
Thanks for the comments @fedde-s ! Yes @alexsigaras has spent some time going through your cBioPortal docker version.

I also think this is p much as good as it gets with the current `properties` file containing both runtime & build time parameters and the `GlobalProperties.java` parsing `properties` at run time. To get more benefits of distributing a binary (or a docker image) we'll have to tackle those issues first. This is: here's a docker image with all the dependencies installed to compile & run cBioPortal. We'd like to get to: here's a docker image of cBioPortal that you can pass some properties/env variables and run w/o compiling further down the line.

Dockerfile
fedde-s: Ah, no, it looks like the last section of `docs/Import-Study-Using-Docker.md` in this pull request suggests a quick-and-dirty solution of running the database migration script as a side-process in the same container as the webserver. That depends on this line.

docs/Deploy-Using-Docker.md
```diff
@@ -0,0 +1,169 @@
...
+
+Update the seeded database schema to match the cBioPortal version in the image.
+
+:warning: Running the following docker command will most likely make your database irreversibly incompatible with older versions of the portal code.
+
+##### Template
+
+```bash
+docker exec -it ${CONTAINER_NAME} \
+migrate_db.py -p /cbioportal/src/main/resources/portal.properties -s /cbioportal/db-scripts/src/main/resources/migration.sql
```
fedde-s: I prefer to run scripts in their own, clean containers separate from the Catalina process like we do in the version we use at The Hyve, but this will do for now.

fedde-s:
Cool, thanks @inodb (and @alexsigaras)! Even the `exec` thing seems like the pragmatic way to go given your comments above. I will approve if the HTML file is either removed or explained.

inodb:
Thanks @fedde-s ! Removed the HTML

docs/Deploy-Using-Docker.md
alexsigaras: @fedde-s nothing against taking this approach however you still need to build cbioportal at every instance that you spin. Running an exec instead minimizes that build time.

Dockerfile
alexsigaras: @fedde-s I would favor removing the find line if there is a cleaner way to import studies. We can update the import study with docker as well.

```ENTRYPOINT``` and ```CMD``` commands are very similar with the ```ENTRYPOINT``` simply requiring a different escape. 
- When running the container in detached (-d) the commands have no real difference.
- When running interactively with tty (-it) for testing or development, I would favor simplicity or allowing the user to build under different options i.e don't skip tests.

Following discussions in the past I suggest compiling cBioPortal before the ```CMD``` and ```ENTRYPOINT```. Such approach significantly limits the time it takes to run a docker container and allows the CMD to be limited to running apache tomcat.

Thoughts?

alexsigaras: @fedde-s we have retained all good practices you followed on the variant you mentioned.

This dockerfile aims to support docker automated builds regardless of the git tag or release and is meant to blend in with the code.
In short, every time there is a tag release in github there will be a docker equivalent version build automatically in docker hub using github triggers.

This allows for full transparency on what the image contains, vetted by Docker itself as it is build in their cloud infrastructure with zero interventions.

jjgao:
@fedde-s please approve if it looks good now.

fedde-s:
Sorry, I meant to click approve but I was travelling from work without access to a computer.

PR #2221, opened 2017-03-16
Validator: Scientific notation and NA values

sandertan:
# What? Why?
Add some possible scenarios that should not error. When recreating staging files for the CCLE study, I noticed some values are causing errors in the validator. The importer does not complain about them. As far as I noticed, these values do not break anything.

The first fix is for when a chromosomal location is written in scientific notation, such as `115,000,000` written as `1.15e+08` in the CNA segmented data:

screen shot 2017-03-16 at 10 32 44

The second fix affects a couple of optional columns in the mutations data, after running Maf2Maf:

screen shot 2017-03-16 at 10 32 55

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1155,34 +1155,34 @@ class MutationsExtendedValidator(Validator):
     
     def checkVerificationStatus(self, value):
         # if value is not blank, then it should be one of these:
-        if self.checkNotBlank(value) and value.lower() not in ('verified', 'unknown'):
+        if self.checkNotBlank(value) and value.lower() not in ('verified', 'unknown', 'na'):
             return False
         return True
     
     def checkValidationStatus(self, value):
         # if value is not blank, then it should be one of these:
         if self.checkNotBlank(value) and value.lower() not in ('untested', 'inconclusive',
-                                 'valid', 'invalid'):
+                                 'valid', 'invalid', 'na'):
             return False
         return True
     
     def check_t_alt_count(self, value):
-        if not self.checkInt(value) and value != '':
+        if not self.checkInt(value) and value not in ('', '.'):
             return False
         return True
     
     def check_t_ref_count(self, value):
-        if not self.checkInt(value) and value != '':
+        if not self.checkInt(value) and value not in ('', '.'):
             return False
         return True
     
     def check_n_alt_count(self, value):
-        if not self.checkInt(value) and value != '':
+        if not self.checkInt(value) and value not in ('', '.'):
             return False
         return True
 
     def check_n_ref_count(self, value):
-        if not self.checkInt(value) and value != '':
+        if not self.checkInt(value) and value not in ('', '.'):
             return False
         return True
 
@@ -1912,6 +1912,12 @@ class SegValidator(Validator):
                                'cause': value})
             elif col_name in ('loc.start', 'loc.end'):
                 try:
+                    # convert possible scientific notation to python scientific notation
+                    if "e+" in value:
+                        value = float(value.replace("e+", "e"))
+                        if not value.is_integer():
+                            # raise value error 'Genomic position is not an integer'
+                            raise ValueError()
                     parsed_coords[col_name] = int(value)
                 except ValueError:
                     self.logger.error(


fedde-s:
@sandertan, did you test if the importer does indeed

- [ ] insert the same value into the database when the `[nt]_(alt|ref)_count` columns of mutation data files contain a `.` as when they are blank,
- [ ] insert the same value when the columns`Validation_Status` and `Verification_Status` contain the string `NA` as when they are blank
- [ ] insert the expected integer value when  a scientific-notation floating point string with a `+` sign occurs as a genomic start or end coordinate in segment-wise CNA data files, and
- [ ] do the same for genomic coordinates in GISTIC_GENES data files?

Especially for the mutation data, I don't think we can make too many assumptions on whether servlets, API logic and front-end code will post-harmonise these values for us if we load invalid values into the database.

fedde-s:
If these values are not treated identically but they definitely should in all cases be equivalent, we should add the same logic to the importer code before merging this.

sandertan:
Thanks Fedde, I'll test it.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1895,6 +1895,9 @@ def checkLine(self, data):
                                'column_number': col_index + 1,
                                'cause': value})
             elif col_name in ('loc.start', 'loc.end'):
+                # convert possible scientific notation to python scientific notation
+                if "e+" in value:
+                    value = float(value.replace("e+", "e"))
                 try:
                     parsed_coords[col_name] = int(value)
                 except ValueError:
```
fedde-s: Another issue here: please log the appropriate message about the validity of the data file format if this field is not a floating-point number instead of letting the validation script crash with an exception, and also log a message if the floating-point number that does not make sense as a genomic coordinate.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2435,6 +2438,9 @@ def parse_genomic_coord(self, value, column_number):
         Return the parsed value if valid, None otherwise.
         """
         parsed_value = None
+        # convert possible scientific notation to python scientific notation
+        if "e+" in value:
+            value = float(value.replace("e+", "e"))
         try:
             parsed_value = int(value)
         except ValueError:
```
fedde-s: The same issue as in segment-wise CNA data files here: please log the appropriate message about the validity of the data file format if this field is not a floating-point number instead of letting the validation script crash with an exception, and also log a message if the floating-point number that does not make sense as a genomic coordinate.

sandertan:
As @fedde-s suggested, I tested with importing with different values in the `[nt]_(alt|ref)_count` columns. The importer writes `-1` when they are blank or `.`, so this validator change is okay.

When `Validation_Status` and `Verification_Status` are blank or `NA`, the importer writes `NA` to database. 

screen shot 2017-03-16 at 14 09 46

I also tested the segment-wise CNA data files, the database seems to import `1.15e+08` correctly as `115,000,000`. 

screen shot 2017-03-16 at 12 06 26

I don't have test data for GISTIC peaks, so I removed this change from the commit.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1896,6 +1896,11 @@ def checkLine(self, data):
                                'cause': value})
             elif col_name in ('loc.start', 'loc.end'):
                 try:
+                    # convert possible scientific notation to python scientific notation
+                    if "e+" in value:
+                        value = float(value.replace("e+", "e"))
+                        if int(value) == value:
+                            pass
```
fedde-s: This conditional does not seem to do anything in either case. Could be useful raise a ValueError or log something if [value.is_integer()](https://docs.python.org/2/library/stdtypes.html#float.is_integer) returns false?

sandertan: Done

sandertan:
Before merging, I see if I can add '---' as empty value for `Verification_Status`, which is causing an error in the skcm_tcga study, reported in https://github.com/cBioPortal/datahub/issues/4

sandertan:
Addition to this PR:

In `skcm_tcga` the `Validation_Status` contains the value `---` 2677 times. The importer imports this as `---`:

screen shot 2017-04-04 at 13 20 02


This screenshot shows the frontend does not have troubles with this value:

screen shot 2017-04-04 at 13 21 18


[Link](http://www.cbioportal.org/index.do?cancer_study_list=skcm_tcga&cancer_study_id=skcm_tcga&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=skcm_tcga_mutations&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=skcm_tcga_sequenced&case_ids=&patient_case_select=sample&gene_set_choice=user-defined-list&gene_list=CACNA1D&clinical_param_selection=null&tab_index=tab_visualize&Action=Submit&show_samples=false&) to the public portal for this study - gene - samples (sort by AA change to get samples with `Validations Status ---` on top).

So I added `---` as valid value for `Validation_Status`. However, this value is not desribed by cBioPortal [FileFormat](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md) (`--` sort of is). In addition, this is not a valid value according to the [NCI MAF specification](https://wiki.nci.nih.gov/display/TCGA/Mutation+Annotation+Format+%28MAF%29+Specification). Last, I'm not sure how this value got in this datafile and therefor it might be better to rerun maf2maf to regenerate this file..

fedde-s:
I did  not review the latest addition to this pull request. @pieterlukasse, do you have time for that?

pieterlukasse:
The changes proposed here look fine to me. Only one concern regarding the `---` value being added to the DB: I would like to avoid that. Can this be converted to `NA` by the loader? Is `---` the same as `NA`? I would propose we store both as `NA` in the DB in this case.
@zheins @n1zea144 Can you check if it would be correct to interpret `Validation_Status=---` as being `Validation_Status=NA`?

n1zea144:
@pieterlukasse The importing code in the cBioPortal looks at validation status to filter out events ([MutationFilter](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/MutationFilter.java#L128)). It will filter out events if the validation status is "Redacted".  This was a convention worked among some groups at MSK, it is not a convention that came out of TCGA.  According to the TCGA MAF Spec the following values are permitted:

- Untested
- Inconclusive
- Valid
- Invalid

If you see a '---' as a validation status, it was probably a bug in the firehose output.  I would update the MAF and replace the values with NA.

sandertan:
@n1zea144 Thanks for the comment. 

I've removed the commit that would allow `---` and removed these values from the Validation_Status column in the MAF file of `skcm_tcga` in https://github.com/cBioPortal/datahub/pull/39 .

PR #2246, opened 2017-03-24
Update the URL for documentation links

oplantalech:
The URL for the documentation links is outdated. This PR solves that issue.


fedde-s:
@jjgao, this is a fix for a long-standing issue in the default properties file, breaking the ‘FAQ’, ‘About Us’ and ‘News’ pages for instances based on the example properties file. The default for instances that don't was previously fixed in 211ad5c8f6493dae87acf4715c0ecc09b0068b80. Would this be good for release in v1.5.0?

PR #2249, opened 2017-03-24
Add util to place link with icon

zhx828:


zhx828:
# What? Why?
OncoKB starts to support TCGA variant type terms. In order to keep the consistency with OncoPrint, remove consequence mapping in OncoKBConnector.js;
Add link in variant description

# Notify reviewers
@onursumer @jjgao

PR #2252, opened 2017-03-27
Omit CIViC spinners in mutations tab if disabled

fedde-s:
Due to a missed conditional, the Annotation column of the mutations
tab showed loading spinner icons for CIViC even if the integration
was disabled, both in the single-study and cross-cancer versions
of the query result views. This adds the conditional and omits the
tags as appropriate.

Tested both with integration enabled and disabled; behaves as I
would expect in both cases.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
@@ -406,9 +406,11 @@ if (sessionError != null) {  %>
     <span class='annotation-item chang_hotspot' alt='{{changHotspotAlt}}'>
         <img width='{{hotspotsImgWidth}}' height='{{hotspotsImgHeight}}' src='{{hotspotsImgSrc}}' alt='Recurrent Hotspot Symbol'>
     </span>
+    <% if (showCivic) { %>
     <span class='annotation-item civic' proteinChange='{{proteinChange}}' geneSymbol='{{geneSymbol}}'>
         <img width='14' height='14' src='images/ajax-loader.gif' alt='Civic Variant Entry'>
     </span>
+    <% } %>
 </script>
 
 <script type="text/template" id="studies-with-no-data-tmpl">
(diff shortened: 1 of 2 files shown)


fedde-s:
@zhx828, could you review this in relation to the issue you reported?

zhx828:
This could work, but I'm thinking whether you should inject the variable when calling for the template. More specifically, in the `AnnotationColumn.js`, you should add variable `showCivic` in the `vars` before calling for `BackboneTemplateCache.getTemplateFn("mutation_table_annotation_template")` @onursumer what do you think?

fedde-s:
On the JSP side, `showCivic` is in scope from [line 72](https://github.com/cBioPortal/cbioportal/blob/919795fab5e45ae45609da1afedd447a23a46be7/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp#L72) of `cross_cancer_results.jsp` and [line 37](https://github.com/cBioPortal/cbioportal/blob/919795fab5e45ae45609da1afedd447a23a46be7/portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp#L37) of `mutation_details.jsp`, respectively.

portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
```diff
@@ -406,9 +406,11 @@ if (sessionError != null) {  %>
     <span class='annotation-item chang_hotspot' alt='{{changHotspotAlt}}'>
         <img width='{{hotspotsImgWidth}}' height='{{hotspotsImgHeight}}' src='{{hotspotsImgSrc}}' alt='Recurrent Hotspot Symbol'>
     </span>
+    <% if (showCivic) { %>
```
onursumer: The underscore template definition would be server-side independent if we use an underscore condition check instead of JSP condition check.

Having said that we won't be using underscore templates after refactoring, so we can handle this while rewriting with React.

fedde-s: @onursumer, for clarity: do you mean we can leave it like this so that v1.5.0 can be released, and move it to the front end in a later iteration?

PR #2254, opened 2017-03-28
Inframe and truncating driver annotation in oncoprint

adamabeshouse:
Inframe annotation with all data sources (hotspot, oncokb, cbioportal count, cosmic count)
Truncating annotation with oncokb only

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

image
index.do?cancer_study_list=nsclc_tcga_broad_2016&cancer_study_id=nsclc_tcga_broad_2016&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=nsclc_tcga_broad_2016_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=nsclc_tcga_broad_2016_cna&data_priority=0&case_set_id=nsclc_tcga_broad_2016_cnaseq&case_ids=&patient_case_select=sample&gene_set_choice=user-defined-list&gene_list=EGFR+MET&clinical_param_selection=null&tab_index=tab_visualize&Action=Submit&show_samples=false&

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -144,6 +144,7 @@ show.oncokb=true
 
 # Enable Chang's hotspot list (true, false)
 show.hotspot=true
+hotspots.url=http://cancerhotspots.org/api/
 
 # Enable Civic variant annotation (true, false)
 show.civic=false
--- a/web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
+++ b/web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
@@ -54,6 +54,11 @@ import java.util.Map;
 @RequestMapping("/proxy")
 public class ProxyController
 {
+	
+  private String hotspotsURL;
+  @Value("${hotspots.url:http://cancerhotspots.org/api/}")
+  public void setHotspotsURL(String property) { this.hotspotsURL = property; }
+  
   private String bitlyURL;
   @Value("${bitly.url}")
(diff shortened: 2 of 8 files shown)


web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
```diff
@@ -98,7 +103,7 @@ public void setEnableOncokb(Boolean property) {
       Map<String, String> pathToUrl = new HashMap<>();
       
       pathToUrl.put("bitly", bitlyURL);
-      pathToUrl.put("cancerHotSpots", "http://cancerhotspots.org/api/hotspots/single/");
+      pathToUrl.put("cancerHotSpots", hotspotsURL);
```
adamabeshouse: done

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -144,6 +144,7 @@ show.oncokb=true
 
 # Enable Chang's hotspot list (true, false)
 show.hotspot=true
+hotspots.url=http://cancerhotspots.org/api/hotspots/single
```
onursumer: Isn't this supposed to be `http://cancerhotspots.org/api/` as opposed to `http://cancerhotspots.org/api/hotspots/single`?
Seems like we are already appending `hotspots/single/` in the `ProxyController` class?

adamabeshouse: good catch, I forgot to change it in the example

adamabeshouse:
@jjgao @onursumer changes made and pushed

PR #2256, opened 2017-03-28
Hide unsequenced samples/patients in oncoprint

adamabeshouse:
Also small polish where instead of "Show unaltered columns" in the View menu, now it says "Show unaltered samples/patients"

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

### DIFF
--- a/portal/src/main/webapp/js/api/cbioportal-datamanager.js
+++ b/portal/src/main/webapp/js/api/cbioportal-datamanager.js
@@ -1733,6 +1733,91 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	    });
 	    return fetch_promise.promise();
 	}),
+	'getUnsequencedSamplesBasedOnSequencedSampleList': makeCachedPromiseFunction(
+		function(self, fetch_promise) {
+		    var sample_list_ids = self.getCancerStudyIds().map(function(cancer_study_id) {
+			return cancer_study_id + "_sequenced";
+		    });
+		    var sample_ids = self.getSampleIds();
+		    window.cbioportal_client.getSampleLists({'sample_list_ids':sample_list_ids}).then(function(sample_lists) {
+			fetch_promise.resolve(sample_lists.reduce(function(map, next) {
+			    var sequenced_samples = next.sample_ids.reduce(function(map, next) { map[next] = true; return map;}, {});
+			    var unsequenced_samples = sample_ids.reduce(function(map, next) {
+				if (!sequenced_samples[next]) {
+				    map[next] = true;
+				}
+				return map;
+			    }, {});
+			    map[next.study_id] = unsequenced_samples;
+			    return map;
+			}, {}));
+		    }).fail(function() {
+			fetch_promise.reject();
+		    });
+		}),
+	'getUnsequencedPatientsBasedOnSequencedSampleList': makeCachedPromiseFunction(
+		function(self, fetch_promise) {
+		    var sample_list_ids = self.getCancerStudyIds().map(function(cancer_study_id) {
+			return cancer_study_id + "_sequenced";
+		    });
+		    $.when(self.getPatientIds(), self.getPatientSampleIdMap(), window.cbioportal_client.getSampleLists({'sample_list_ids':sample_list_ids})).then(function(patient_ids, sample_to_patient, sample_lists) {
+			fetch_promise.resolve(sample_lists.reduce(function(map, next) {
+			    var sequenced_patients = next.sample_ids.reduce(function(map, next) {
+				var patient = sample_to_patient[next];
+				if (typeof patient !== "undefined") {
+				    map[patient] = true;
+				}
+				return map;
+			    }, {});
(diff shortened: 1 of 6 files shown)


fedde-s:
Mainly, what I would hope to be more clearly documented is that leaving out the case lists affects more than just the Data Sets page.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1703,6 +1703,91 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 	    });
 	    return fetch_promise.promise();
 	}),
+	'getUnsequencedSamplesBasedOnSequencedSampleList': makeCachedPromiseFunction(
+		function(self, fetch_promise) {
+		    var sample_list_ids = self.getCancerStudyIds().map(function(cancer_study_id) {
+			return cancer_study_id + "_sequenced";
```
adamabeshouse: @jjgao not sure how to address this

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1703,6 +1703,91 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
+	    }),
+	    'getUnsequencedSampleUIDsBasedOnSequencedSampleList': makeCachedPromiseFunction(
+		    function(self, fetch_promise) {
+			$.when(self.getUnsequencedSamplesBasedOnSequencedSampleList(), self.getCaseUIDMap())
+				.then(function(unsequenced_samples, case_uid_map) {
+				    var ret = [];
+				    var studies = Object.keys(unsequenced_samples);
+				    for (var i=0; i<studies.length; i++) {
+					ret = ret.concat(
+						Object.keys(unsequenced_samples[studies[i]]).map(function(sample) {
```
adamabeshouse: good idea

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -871,6 +868,14 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 	    oncoprint.setSortConfig({'type':'order', 'order':order});
 	};
 	
+	var updateViewMenuLabels = function(state) {
```
adamabeshouse: this doesn't bother me because it's still clear which function is being called. all that may be unclear to the reader who doesn't know about hoisting is how it's possible that its being called before its declared. but I think anyone reading it would just Control+F and find `updateViewMenuLabels`

fedde-s: Sure, if the pattern used here prevents people from considering to call functions before this block has defined them all, it's not important.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
pieterlukasse: Good point...In my opinion, using the stable id of case lists to drive business logic in the application should be avoided. However, this is spread by now in cBioPortal...for both case list stable ids and genetic profile stable ids. What @fedde-s suggests here is a good idea. I think this could be the first step for solving the issue on the longer run. Basically have a central place that can be the "place to query" when one has to find specific sample lists. For example, a utils file with a function like: 
```
function getSampleListIdsForDataTypes(dataTypes) {
    //business logic that returns one or more sample list ids based on given data types
    //...or perhaps just a wrapper here, with business logic implemented in backend
}
```
This business logic would still use the stable id matching to find the correct sample lists, but having it implemented in one place would make it easier to maintain or change it in the future. 

@jjgao @adamabeshouse what do you think of this idea?

adamabeshouse: @pieterlukasse @fedde-s I agree with what you're saying.
I'm looking at the `sample_list` table now, and it looks like the sample lists we're interested in can be more robustly picked out as having the category `all_cases_with_mutation_data`. What do you all think of that? @jjgao

fedde-s:
You seem to have squashed a lot of unrelated changes into the same commit, which would make it a bit hard if you want me to review them. I would appreciate it if you could push amendments in separate commits once review has started, and postpone squashing until after they have been reviewed.

adamabeshouse:
@fedde-s Sorry about that. All the changes are still related to this pull request.

jjgao:
@pieterlukasse @fedde-s any more review?

fedde-s:
@jjgao, I was mostly still waiting for any follow-up on the discussion following [this comment](https://github.com/cBioPortal/cbioportal/pull/2256#discussion_r108867231), as I'd prefer not to add hard-coded dependencies on (poorly documented) conventions in the data file formats duplicated in multiple frontend functions. The discussion seems to have ended with a request for your input on it, could you comment?

pieterlukasse:
Is this PR still being pursued?

adamabeshouse:
even if this feature is still desired, this code is now completely obsolete

PR #2262, opened 2017-03-29
Add 3 endpoints in Clinical Attributes API

ersinciftci:
GET /clinical-attributes
GET /studies/{studyId}/clinical-attributes
GET /studies/{studyId}/clinical-attributes/{clinicalAttributeId}

Also changed the max limit of entrez gene id parameters in Discrete Copy Number API

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/ClinicalAttributeServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/ClinicalAttributeServiceImpl.java
@@ -1,0 +1,67 @@
+package org.cbioportal.service.impl;
+
+import org.cbioportal.model.ClinicalAttribute;
+import org.cbioportal.model.meta.BaseMeta;
+import org.cbioportal.persistence.ClinicalAttributeRepository;
+import org.cbioportal.service.ClinicalAttributeService;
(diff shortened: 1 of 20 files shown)


ersinciftci:
[(3) Querying discrete CNA data for many rows leads to 400 error, and querying variant count data almost always leads to 503 error](https://trello.com/c/PkMEQ4q1/177-3-querying-discrete-cna-data-for-many-rows-leads-to-400-error-and-querying-variant-count-data-almost-always-leads-to-503-error)

service/src/main/java/org/cbioportal/service/impl/ClinicalAttributeServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+
+@Service
+public class ClinicalAttributeServiceImpl implements ClinicalAttributeService {
+
+    @Autowired
+    private ClinicalAttributeRepository clinicalAttributeRepository;
+
+    @Override
+    @PostFilter("hasPermission(filterObject.cancerStudyIdentifier, 'CancerStudy', 'read')")
```
onursumer: Here, it would be nicer to be able to pass a lambda function or something similar instead of string.

PR #2268, opened 2017-03-31
Calculate mutation variant counts during import

ersinciftci:
Created a new table `mutation_variant_count`, which holds both keyword
and entrez gene id counts. Entrez gene id counts are denormalized, meaning rows
that has the same genetic profile id and entrez gene id will have the same gene
count. I implemented this way because whenever we need keyword count, we need
gene count too, so easier to read both in one query.


db-scripts/src/main/resources/cgds.sql
```diff
@@ -407,6 +408,18 @@ CREATE TABLE `mutation_count` (
 );
 
 -- --------------------------------------------------------
+CREATE TABLE `mutation_variant_count` (
```
jjgao: I am not sure if the meaning of the table name is clear.

db-scripts/src/main/resources/cgds.sql
```diff
@@ -407,6 +408,18 @@ CREATE TABLE `mutation_count` (
...
+    `GENETIC_PROFILE_ID` int(11) NOT NULL,
+    `KEYWORD` varchar(50) DEFAULT NULL,
```
jjgao: Thinking of matching to position as well, maybe instead of using `KEYWORD`, we could use `MUTATION_EVENT_ID`? And we can put `COUNT_BY_PROTEIN_POSITION` as a placeholder now. Later we can even extended it to `COUNT_BY_NT_POSITION` (which is actually asked by our users).

jjgao:
@ersinciftci I am wondering how much time it takes to run the code for a large study?

ersinciftci:
@jjgao I can test if you can point me to a large study in datahub.

PR #2285, opened 2017-04-03
Added `IMPORT_DATE` usage for importing cancer studies

angelicaochoa:
# What? Why?
Cancer study field `IMPORT_DATE` was not getting used but would be useful info to have. 

This PR makes the necessary changes in `DaoCancerStudy.java` and `CancerStudy.java` to start using the `IMPORT_DATE` field again.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers

@cBioPortal/backend

and notify them here:

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
@@ -34,6 +34,7 @@ package org.mskcc.cbio.portal.dao;
 
 import java.sql.*;
 import java.text.*;
+import java.time.LocalDate;
 import java.util.*;
 import org.apache.commons.lang.StringUtils;
 import org.mskcc.cbio.portal.model.*;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -317,6 +318,7 @@ public static void addCancerStudy(CancerStudy cancerStudy, boolean overwrite) th
             //data loading process can set this to AVAILABLE:
             //TODO - use this field in parts of the system that build up the list of studies to display in home page:
             pstmt.setInt(10, Status.UNAVAILABLE.ordinal());
+            pstmt.setDate(11, java.sql.Date.valueOf(LocalDate.now()));
```
zheins: sql.Date to match datetime datatype in cgds.sql

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -34,6 +34,7 @@
 
 import java.sql.*;
 import java.text.*;
+import java.time.LocalDate;
```
sheridancbio: LocalDate is a very new class in the java SDK .. it requires a java 1.8 compiler. Do we have other code in the codebase which requires java 1.8? I think we do, so ok -- but if there is an easy 1.7 compatible class to accomplish the same thing, that might be wiser (in case other centers are using 1.7)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
sheridancbio: I think our standard pattern is to use import statements instead of full class package specifiers in function code blocks. I guess in this case we have a collision of class names? .. I see - with java.util.Date. It is good that we are using full package specifiers consistently for both classes. I'd probably drop the import "java.util.*" statement too, but it is ok so long we don't rely on it. (Maybe this is why google forbids importing with wildcards)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
sheridancbio: Confirmed that we are already committed to 1.8 -- portal will no longer build with 1.7 compiler.

sheridancbio:
Merging now. Tested deployed portal briefly on all primary pages [query / result pages, patient view, study summary, cross cancer query] to make sure no obvious broken frontend elements now that import date may be reported in /api/studies

Issue #2247, opened 2017-03-24
Empty survival plots in some TCGA studies


oplantalech:
Due to a change of the data model (PR #1746), now all the clinical variables defined per study are defined in the clinical_attribute_meta, even the ones with contain no data. Therefore, since some TCGA studies contain variables in their clinical patient data files with only '[Not Available]' values, in the study view some empty plots appear:
![captura de pantalla 2017-03-24 a les 15 23 48](https://cloud.githubusercontent.com/assets/20278013/24299224/bf6c3b24-10a8-11e7-835d-0a8f8942af9f.png)

If we do not want to display empty plots in the study view, then I suggest to reupload the affected studies in datahub with the empty clinical variables removed from the data files. However, this needs **discussion**: if clinical variables are defined in the files, then we want the plots related to them to be displayed, even if they do not contain data (empty plots can be hidden by default in the study view). We might also raise a warning if the validator finds clinical attributes with no data.

Also, if empty plots appear, it would be nice to display "No data available" in them.

@pieterlukasse @n1zea144 what is your opinion?

zhx828:
Thanks for pointing out this issue. Just want to let you know that I will add `No data available` for survival plots. It will be also used for the new rainbow survival feature.

PR #2292, opened 2017-04-05
Study view bug fixes; Enable rainbow survival 

### LINKED ISSUE #2291: Study view hangs with ldap authentication and no sequenced or cna case list
Study view hangs with ldap authentication and no sequenced or cna case list:
captura de pantalla 2017-04-05 a les 12 16 47

This happens because cBioPortal is trying to retrieve non-existing case lists (sequenced and/or cna), and a 403 error is returned. However, in a system with no authentication, non-existing case lists return an empty list: http://www.cbioportal.org/api/sample-lists/prad_mskcc_2014_dummy/sample-ids.

These calls are made because the study view always shows the pie charts for 'With mutation data' and 'With CNA data'. For example, if there are no samples with mutation data, the pie chart shows all samples with mutation data = NO. 

This issue is probably related with the PR #2138 that @mandawilson did.

zhx828:
# What? Why?
Fix #2291, related to #2247 and https://github.com/cBioPortal/iViz/issues/450

- Fix issue that the study without certain sample list hanged due to authentication (API will return 403). Solution: Get sample lists first then decide which sample list should be queried
- Give `No data available` message when study contains clinical attributes meta related to survival plot, but without actual data. More details are here #2247
- Give message `Failed to load data, refresh the page may help` when the API calls of Mutated Gene table and CNA table failed
- Silent release rainbow survival since this is an old feature

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?
Screenshot (Survival analysis based on ucec_tcga, Diagnosis Age):
screen shot 2017-03-21 at 10 12 24 pm

screen shot 2017-04-05 at 3 52 51 pm
screen shot 2017-04-05 at 3 52 54 pm

# Notify reviewers
No need to review. They have been reviewed by @yichaoS  @kalletlak and @adamabeshouse


pieterlukasse:
@zhx828 thanks for quickly taking action here. I would like however to discuss the 403 issue in more detail. I think it needs fixing in the backend, and not only in the frontend. The real issue is that when one uses the API to request a non-existing sample list, e.g. http://www.cbioportal.org/api/sample-lists/prad_mskcc_2014_dummy/sample-ids, the result is an empty list, while the **same** request returns 403 when the portal is running with authentication and the user is already logged in. I think that in both cases the result should be 404 (and not 403) or an empty list. At least it should be consistent.  
Adding @ersinciftci @mandawilson to this discussion as well.

oplantalech:
@zhx828 It would also be nice to include the key you have introduced in the survival plots in the 'Mutation count vs. CNA' plot.

oplantalech:
@zhx828 While revising your fixes, I found a similar bug for 'Mutation count vs. CNA' plot: when there is no gene panel data, a 403 error is raised instead of returning an empty list. The plot in the study view hangs indefinitely.
captura de pantalla 2017-04-06 a les 14 01 05

mandawilson:
The security on the API always returns a 403 error on an object that is not found.  Intentionally.  I do understand that others might not agree, and am fine with it being changed after the right people are notified of the consequences.  Note I have done a lot of work with PHI so I am probably approaching this overly cautiously.  Say I make the change proposed (returning a 404 when the object is not found), and then a user does the following (with this made up API):

http://site.com/api/STUDY_THAT_DOES_NOT_EXIST

And they get a 404.  Next time I do:

http://site.com/api/STUDY_THEY_ARE_NOT_ALLOWED_TO_SEE

And they get a 403 error.  Now my API has leaked information about a study that the user is not allowed to see (and some might encode a bit of information into the study name).  So now they know a study exists.  That probably doesn't matter.

But we are talking about having sample and patient level security.  So then someone can do:

http://site.com/api/STUDY_THEY_ARE_ALLOWED_TO_SEE/PATIENT_THEY_ARE_NOT_ALLOWED_TO_SEE

Or:

http://site.com/api/STUDY_THEY_ARE_ALLOWED_TO_SEE/SAMPLE_THEY_ARE_NOT_ALLOWED_TO_SEE

And they get 404 errors.  Again, we have leaked information.  Now the patient ids *should not* be identifying, and really how much information can you get from a non-identifying patient id and a sample name anyway?  But to me security on an object means you cannot learn anything about this object.  Including that it exists.  And I would insist that a decision to change that come from higher up than anyone on this thread.

ersinciftci:
I agree with @mandawilson

zhx828:
@oplantalech I fixed the issue in the latest commit. I also have an issue to keep tracking of the rest of error handling progress.(https://github.com/cBioPortal/iViz/issues/450) But the items haven't been finished should be separated into anther pull request since this is more urgent to merge.

ersinciftci:
Created a PR https://github.com/cBioPortal/cbioportal/pull/2294 so that http://www.cbioportal.org/api/sample-lists/prad_mskcc_2014_dummy/sample-ids will return a 404 instead of empty list.

ersinciftci:
I guess I need to change the frontend error handling code in the same PR, because the study view screenshot test is failing. Can you guide me what file to change @zhx828 ?

zhx828:
I dont think you need to change anything. Once this pull request is merged,
you can do a rebase, the screenshot test will pass

ersinciftci:
@zhx828 I was going to rebase but this PR is merged to hotfix, can we merge it to rc as well?

zhx828:
@jjgao can we merge hotfix branch to rc branch?

PR #2294, opened 2017-04-06
Return 404 when a path variable can't be found

ersinciftci:
We were returning an empty list if a list-returning API has a path parameter and
that parameter can't be found in DB. Now we will return 404 in those cases.
This only applies to path parameters, nothing changed about query or body parameters.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/ClinicalAttributeServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/ClinicalAttributeServiceImpl.java
@@ -1,10 +1,13 @@
 package org.cbioportal.service.impl;
 
+import org.cbioportal.model.CancerStudy;
 import org.cbioportal.model.ClinicalAttribute;
 import org.cbioportal.model.meta.BaseMeta;
 import org.cbioportal.persistence.ClinicalAttributeRepository;
 import org.cbioportal.service.ClinicalAttributeService;
+import org.cbioportal.service.StudyService;
 import org.cbioportal.service.exception.ClinicalAttributeNotFoundException;
+import org.cbioportal.service.exception.StudyNotFoundException;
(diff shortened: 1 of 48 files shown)


service/src/main/java/org/cbioportal/service/impl/ClinicalAttributeServiceImpl.java
```diff
@@ -52,15 +59,19 @@ public ClinicalAttribute getClinicalAttribute(String studyId, String clinicalAtt
     @PreAuthorize("hasPermission(#studyId, 'CancerStudy', 'read')")
     public List<ClinicalAttribute> getAllClinicalAttributesInStudy(String studyId, String projection, Integer pageSize,
                                                                    Integer pageNumber, String sortBy,
-                                                                   String direction) {
+                                                                   String direction) throws StudyNotFoundException {
+
+        studyService.getStudy(studyId);
```
onursumer: Why don't we use the return value here? Is it only called to trigger the exception condition?

ersinciftci: Yes, we don't need the return value, as long as doesn't throw an exception, we continue.

pieterlukasse: maybe add a comment on top of each place like this? The unaware developer might be tempted to remove it at the first sight, since it looks like it is serving no purpose while the contrary is true (i.e. it is absolutely critical that this line is there).

ersinciftci: Yeah it sucks that there is no way to indicate my purpose in this case other than comments (comments are my last resort, I try to make the code more clear by changing the code itself before adding comments), but yeah in this case it seems we need it, and Onur's question proves it. I'll add them.

PR #2310, opened 2017-04-11
Validator: Mass Spectrometry data

sandertan:
# What? Why?
This PR adds changes to validator to allow the validation of mass spectrometry data (added in https://github.com/cBioPortal/datahub/pull/31) 

- Add validation of mass spectrometry data. The data and meta files are in the same format as RPPA files, so validation is done using the same method. 
- Rename the RPPA validation method to Protein level validation.


core/src/main/scripts/importer/allowed_data_types.txt
```diff
@@ -32,3 +32,5 @@ METHYLATION	CONTINUOUS	methylation_hm450
 FUSION	FUSION	fusion
 PROTEIN_LEVEL	LOG2-VALUE	rppa
 PROTEIN_LEVEL	Z-SCORE	rppa_Zscores
+PROTEIN_LEVEL	CONTINUOUS	protein_quantification
+PROTEIN_LEVEL	Z-SCORE	protein_quantification_zscores
```
pieterlukasse: should we keep the pattern and make this `protein_quantification_Zscores` with capital Z?

core/src/main/scripts/importer/validateData.py
```diff
@@ -2094,7 +2094,7 @@ def parseFeatureColumns(self, nonsample_col_vals):
         # products if the antibody name has a particular format.
         value = nonsample_col_vals[0].strip()
         if '|' not in value:
-            self.logger.error('No pipe symbol in RPPA probe column',
+            self.logger.error('No pipe symbol in Composite.Element.REF probe column',
```
pieterlukasse: maybe the word "probe" is strange in the context of mass spec...just remove the word probe in this sentence. I was also wondering whether it makes sense to have `Composite.Element.REF` as column name in case of mass spec. Is there any probe information in `Composite.Element.REF` column for CPTAC data?

docs/File-Formats.md
```diff
@@ -661,7 +661,7 @@ profile_name: Protein expression Z-scores (RPPA)
 
 #### Data file
 
-An RPPA data file is a two dimensional matrix with an antibody per row and a sample per column.  For each antibody-sample pair, a real number represents the RPPA level for that sample.  The antibody information can contain one or more HUGO gene symbols and/or entrez gene identifiers, separated by a space, and an antibody ID pair separated by the "|" symbol. 
+An protein level data file is a two dimensional matrix with a RPPA antibody per row and a sample per column. For each antibody-sample pair, a real number represents the protein level for that sample.  The antibody information can contain one or more HUGO gene symbols and/or entrez gene identifiers, separated by a space, and an antibody ID pair separated by the "|" symbol.
```
pieterlukasse: "A protein level..." (replace An by A)

docs/File-Formats.md
```diff
@@ -616,18 +616,18 @@ data_filename: data_methylation_hm27.txt
 The methylation data file follows the same format as expression data files. See [Expression Data](#expression-data) for a description of the expression data file format. The Portal expects a single value for each gene in each sample, usually a beta-value from the Infinium methylation array platform.
 
 
-## RPPA Data
+## Protein level Data
 
-Protein expression measured by reverse-phase protein array. Antibody-sample pairs, with a real number representing the RPPA level for that sample. 
+Protein expression measured by reverse-phase protein array or mass spectrometry (CPTAC). Antibody-sample pairs, with a real number representing the protein level for that sample.
```
pieterlukasse: please remove "(CPTAC)"

core/src/main/scripts/importer/allowed_data_types.txt
sandertan: I'll add this to list of issues with the current CPTAC data, so we can consider changing it when we update the data. To me it makes more sense to keep it lower case, because 
1. some other terms are also lower which usually are in upper case (e.g. rna and rrpa)
2. it looks like this whole column should be lower case
3. bit of a hassle to go through the pipeline to regenerate this data only for 1 small change.

core/src/main/scripts/importer/validateData.py
sandertan: I'll add it to the list of things to change in an updated version of the CPTAC data.

docs/File-Formats.md
sandertan: Done

docs/File-Formats.md
sandertan: Done

core/src/main/scripts/importer/allowed_data_types.txt
pieterlukasse: I also prefer lower case. Let's keep it as is :+1:

pieterlukasse:
Changes are good and I tested it as well :+1:

PR #2314, opened 2017-04-11
Use new external db for travis & heroku

inodb:
Use Amazon db as provided through MSKCC instead of personal account


jjgao:
this code has merged in pr #2311 and #2318

PR #2319, opened 2017-04-12
Make OncoKB and CIViC asynchronous

### LINKED ISSUE #2204: Difference resources in annotation column should stay asynchronous
For a big query, especially under cross cancer page, different resources(CIViC, OncoKB) response time are different. We shouldn't wait for all resources respond.

In this case, http://www.cbioportal.org/beta/index.do?tab_index=tab_visualize&cancer_study_list=coadread_tcga_pub&cancer_study_id=coadread_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&Z_SCORE_THRESHOLD=2.0&case_set_id=coadread_tcga_pub_nonhypermut&case_ids=&gene_list=KRAS+NRAS+BRAF&gene_set_choice=user-defined-list&Action=Submit&show_samples=false& CIViC needs to take a look the response time.

oplantalech:
# What? Why?
Fix #2204. Note that it has been fixed for the mutation tab in the results, but I am not sure if they are completely asynchronous in the patient view, too (however, I believe this is less important since the patient view is going to be changed in the future).

# Notify reviewers
@zhx828

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
@@ -475,6 +475,15 @@ function initOncoKB(instanceId, ids, mapData, type, indicatorCallback) {
                     indicatorCallback(instance);
                 }
             });
+            if (showCivic) {
+                instance.getCivicIndicator().then(function () {
+                    if(_.isFunction(indicatorCallback)) {
+                        indicatorCallback(instance);
+                    }
+                }).fail(function() {
+                    console.log('getCivicIndicator failed.');
+                });
+            }
         }
     }
     return instance;
--- a/portal/src/main/webapp/js/src/mutation/column/AnnotationColumn.js
+++ b/portal/src/main/webapp/js/src/mutation/column/AnnotationColumn.js
@@ -172,6 +172,29 @@ function AnnotationColumn(oncokbInstanceManager, showHotspot, enableMyCancerGeno
                     callback(params);
                 }
             });
+            if (showCivic) {
+                oncokbInstance.getCivicIndicator().done(function () {
+                    var tableData = dataTable.fnGetData();
+                    if (tableData.length > 0) {
+                        _.each(tableData, function (ele, i) {
+                            if (oncokbInstance.getVariant(ele[indexMap['datum']].mutation.get("mutationSid"))) {
+                                if (oncokbInstance.getVariant(ele[indexMap['datum']].mutation.get("mutationSid")).hasOwnProperty('evidence')) {
+                                    ele[indexMap["datum"]].oncokb = oncokbInstance.getVariant(ele[indexMap['datum']].mutation.get("mutationSid"));
+                                    ele[indexMap['datum']].mutation.set({oncokb: true});
+                                    //dataTable.fnUpdate(null, i, indexMap["annotation"], false, false);
+                                }
+                            }
+                        });
+                        //dataTable.fnUpdate(null, 0, indexMap['annotation']);
+                    }
+
+                    if (_.isFunction(callback)) {
+                        callback(params);
+                    }
+                }).fail(function() {
+                    console.log('getCivicIndicator failed.');
+                });
+            }
         }
     }
 
--- a/portal/src/main/webapp/js/src/oncokb/OncoKBConnector.js
+++ b/portal/src/main/webapp/js/src/oncokb/OncoKBConnector.js
@@ -157,6 +157,7 @@ var OncoKB = (function(_, $) {
     function Instance(id) {
         this.initialized = false;
         this.dataReady = false;
+        this.civicDataReady = false;
         this.tumorType = ''; //Global tumor types
         this.source = 'cbioportal';
         this.geneStatus = 'Complete';
@@ -863,7 +864,6 @@ OncoKB.Instance.prototype = {
 
         self.setVariantUniqueIds();
 
-        var geneSymbols = []
         for (var key in this.variants) {
             var variant = this.variants[key];
             var uniqueStr = variant.gene + variant.alteration + variant.tumorType + variant.consequence;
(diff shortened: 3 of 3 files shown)


portal/src/main/webapp/js/src/oncokb/OncoKBConnector.js
```diff
@@ -157,6 +157,7 @@ var OncoKB = (function(_, $) {
     function Instance(id) {
         this.initialized = false;
         this.dataReady = false;
+        this.civicDataReady = false;
```
jjgao: I don't understand why all the civic code resides in OncoKBConnector.js? Is it going to be different in the new version of PV? cc'ing @zhx828 @onursumer

onursumer: @jjgao We don't have the `OncoKBConnector` in the new refactored code. 

Does civic annotation depend on the data coming from OncoKB? If not, then I don't think it is a good idea to couple civic with OncoKB this way.

zhx828: I think in the future, each resource under annotation column will have its own dataProxy and related code into a library.

portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
```diff
@@ -475,6 +475,11 @@ function initOncoKB(instanceId, ids, mapData, type, indicatorCallback) {
                     indicatorCallback(instance);
                 }
             });
+            instance.getCivicIndicator().then(function () {
+                if(_.isFunction(indicatorCallback)) {
+                    indicatorCallback(instance);
+                }
+            });
```
zhx828: Do you want to handle failed to call `instance.getCivicIndicator()`?

portal/src/main/webapp/js/src/mutation/column/AnnotationColumn.js
```diff
@@ -111,6 +111,8 @@ function AnnotationColumn(oncokbInstanceManager, showHotspot, enableMyCancerGeno
         $(selector).find('span.chang_hotspot[alt=""]').empty();
         oncokbInstanceManager.getInstance(helper.gene).addEvents(selector, 'column');
         oncokbInstanceManager.getInstance(helper.gene).addEvents(selector, 'alteration');
+        oncokbInstanceManager.getInstance(helper.gene).addEvents(selector, 'column');
```
zhx828: Why do we duplicate the code here?

portal/src/main/webapp/js/src/mutation/column/AnnotationColumn.js
```diff
@@ -172,6 +174,25 @@ function AnnotationColumn(oncokbInstanceManager, showHotspot, enableMyCancerGeno
                     callback(params);
                 }
             });
+            oncokbInstance.getCivicIndicator().done(function () {
```
zhx828: Same here, you should handle the fail() as well

portal/src/main/webapp/js/src/oncokb/OncoKBConnector.js
zhx828: @jjgao it's ok to let civic code live inside OncoKBConnector.js for now. But no more new features should be developed, OncoKBConnector.js will soon be deprecated.

oplantalech: CIViC does not rely on data coming from OncoKB, but we did it inside OncoKBConnector.js to get advantage of mainly the OncoKB event handlers without having to duplicate the code.

portal/src/main/webapp/js/src/mutation/column/AnnotationColumn.js
oplantalech: Thanks for pointing this out! It is an error.

portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
oplantalech: I corrected that.

portal/src/main/webapp/js/src/mutation/column/AnnotationColumn.js
oplantalech: I corrected that.

zhx828:
@oplantalech Please try to disable CIViC and test. `instance.getCivicIndicator().then()` will throw exception since getCicivIndicator inside of OncoKBConnector.js will return `undefined`. And base on current structure, if you want to use CIViC annotation, you have to enable OncoKB annotation. Maybe this is something you also want to consider.

portal/src/main/webapp/WEB-INF/jsp/patient_view/patient_view.jsp
```diff
@@ -475,13 +475,15 @@ function initOncoKB(instanceId, ids, mapData, type, indicatorCallback) {
...
                 }
             });
-            instance.getCivicIndicator().then(function () {
-                if(_.isFunction(indicatorCallback)) {
-                    indicatorCallback(instance);
-                }
-            }).fail(function() {
-                console.log('getCivicIndicator failed.');
-            });
+            if (showCivic) {
```
zhx828: This works. But getCivicIndicator itself already have `if (showCivic)` condition check. I would rather reuse that. In order to do so, you probably want to return a promise inside of `getCivicIndicator` and return def.reject() if the CIViC is disabled.

oplantalech: Thank you for pointing this out. I just corrected it.

oplantalech:
@jj Can you merge? I just squashed all the commits into one after @zhx828 approved the PR.

oplantalech:
@jjgao Can you merge? I just squashed all the commits into one after @zhx828 approved the PR.

portal/src/main/webapp/js/src/oncokb/OncoKBConnector.js
pieterlukasse: I also would prefer something like `CivicConnector.js` instead. But if there are refactoring plans anyway, perhaps it is OK for now.

PR #2369, opened 2017-04-18
Validator: Add cBio version to log and html-report

sandertan:
# What? Why?
Add cBioPortal version to log and html-report. This is useful when reviewing reports from older versions of cBioPortal, since many warnings/errors have changed in recent versions.

- Add version to `logger.info` at start of validation:

```INFO: -: Running validation from cBioPortal version 1.5.2-SNAPSHOT```

- Add version to html report:

screen shot 2017-04-18 at 14 26 18

- Because the validator can be run independently from cBioPortal, the default value for this version is `unknown`.

- Modify unit tests to compare the new html-report, using a `test.xml` that includes a mock cBioPortal version

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -39,6 +39,7 @@ import csv
 import itertools
 import requests
 import json
+import xml.etree.ElementTree as ET
 
 import cbioportal_common
 
@@ -136,10 +137,11 @@ class Jinja2HtmlHandler(logging.handlers.BufferingHandler):
 
     """Logging handler that formats aggregated HTML reports using Jinja2."""
 
-    def __init__(self, study_dir, output_filename, *args, **kwargs):
+    def __init__(self, study_dir, output_filename, cbio_version, *args, **kwargs):
         """Set study directory name, output filename and buffer size."""
         self.study_dir = study_dir
         self.output_filename = output_filename
+        self.cbio_version = cbio_version
         self.max_level = logging.NOTSET
         self.closed = False
         # get the directory name of the currently running script,
@@ -175,6 +177,7 @@ class Jinja2HtmlHandler(logging.handlers.BufferingHandler):
         template = j_env.get_template('validation_report_template.html.jinja')
         doc = template.render(
             study_dir=self.study_dir,
+            cbio_version=self.cbio_version,
             record_list=self.buffer,
             max_level=logging.getLevelName(self.max_level))
         with open(self.output_filename, 'w') as f:
@@ -3065,6 +3068,14 @@ def validate_study(study_dir, portal_instance, logger, relaxed_mode):
     logger.info('Validation complete')
 
 
+def get_pom_path():
+    """
+    Get location of pom.xml. In system and integration test this is mocked.
+    """
+    pom_path = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))))) + "/pom.xml"
+    return pom_path
+
+
 def main_validate(args):
 
     """Main function: process parsed arguments and validate the study."""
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by sandertan** [comment]
Thanks @fedde-s for review. I refactored and `git commit ammend`-ed it, which will make it more clear for review I think.

File: core/src/main/scripts/importer/validateData.py [hunk 1]
```diff
@@ -3105,6 +3117,24 @@ def main_validate(args):
...
+        xml_root = ET.parse(pom_path).getroot()
+        for xml_child in xml_root:
+
+            # to circumvent the default namespace (possibly varying apache url) split on '}'
+            if xml_child.tag.split("}")[1] == "version":
+                cbio_version = xml_child.text
+
+                # output cBioPortal version
+                logger.info("Running validation from cBioPortal version %s" % cbio_version)
+    except Exception:
```
fedde-s: This should be more specific, and only catch the exceptions you actually want to ignore by passing through with a default `cbio_version` (I/O errors opening and reading the file). It should not, for instance, ignore manual interrupts by the user and actual runtime errors in logging handlers. If the `ET.parse()` call is the only part that performs file I/O, false positives could be limited even further by putting only that line in the `try` block and moving the handling of the XML contents to [an `else` clause after the `except`](https://docs.python.org/2/reference/compound_stmts.html#index-12).

core/src/test/scripts/system_tests_validate_data.py
```diff
@@ -10,6 +10,7 @@
 import logging
 import os
 from importer import validateData
+from importer.validateData import get_pom_path
```
fedde-s: Is there a reason against accessing this function via the import on the line above, as `validateData.get_pom_path`? I think that would more explicitly show you're overriding (mocking) a global in another module.

core/src/test/scripts/system_tests_validate_data.py
```diff
@@ -26,8 +27,17 @@ class ValidateDataSystemTester(unittest.TestCase):
...
+    def setUp(self):
+        def dummy_pom_path():
+            return "test_data/test.xml"
+        self.orig_defined_sample_ids = get_pom_path
+        validateData.get_pom_path = dummy_pom_path
+
     def tearDown(self):
         """Close logging handlers after running validator and remove tmpdir."""
+        # restore original function
+        validateData.get_pom_path = self.orig_defined_sample_ids
```
fedde-s: Since this doesn't have anything to do with sample IDs, could you give this attribute a different name? Something like `self.orig_pom_path_function`, I mean.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3065,6 +3068,14 @@ def validate_study(study_dir, portal_instance, logger, relaxed_mode):
     logger.info('Validation complete')
 
 
+def get_pom_path():
+    """
+    Get location of pom.xml. In system and integration test this is mocked.
+    """
+    pom_path = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))))) + "/pom.xml"
```
fedde-s: Nice, it seems like this also just ‘bottoms out’ without errors even if the script path is the root directory. But please do canonicalise the script's path using `os.path.realpath()` here, or it will break situations where the script is installed into `$PATH` via a symlink, such as the Dockerised setups.

sandertan: Done.

core/src/main/scripts/importer/validateData.py
sandertan: Excellent. thanks.

core/src/test/scripts/system_tests_validate_data.py
sandertan: Changed it.

fedde-s:
Looks good! @sandertan, do you want to squash the fixup commit into the main commit before this is merged, or shall I mark it as approved?

sandertan:
I've squashed it!

Issue #436, opened 2015-11-11
Sporadic NullPointerException in the public portal


jjgao:

jjgao:
It was because of the recaching.

ecerami:
hurray!  thank you.

can you explain the cause in more detail?  just curious.


Sent from mobile phone.  Please excuse typos.

-------- Original message --------

PR #2375, opened 2017-04-18
Fix iViz issue #436 (accept user customized case list)

yichaoS:
Fixed issue: https://github.com/cBioPortal/iViz/issues/436
Demo: https://cbioportal-pr-2375.herokuapp.com/


core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -383,15 +384,22 @@ private void processData(CohortDetails cohortDetails,
...
-				else if (sampleIds != null && sampleIds.length()>0) {
-                    // TODO: temporary solution
-				    String[] _keys = sampleIds.split("\\+");
-                    List<String> _arr = new ArrayList<String>();
-                    for (String _key: _keys) {
-                        String[] _tmp = _key.split("\\|");
-                        _arr.add(_tmp[1]);
+				else if (sampleIdsStr != null && sampleIdsStr.length() > 0) {
+                    List<String> sampleIds = new ArrayList<>();
+				    if (sampleIdsStr.contains("|")) { //TODO: a better way to distinguish input from "Case Ids" box or "Summary" page
```
kalletlak: any specific reason to distinguish the requests?

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -330,7 +330,8 @@ protected void doPost(HttpServletRequest httpServletRequest,
     private void processData(CohortDetails cohortDetails,
                              String geneList,
 							 HashSet<String> geneticProfileIdSet,
-							 String sampleSetId, String sampleIds,
+							 String sampleSetId, 
+                             String sampleIdsStr, //raw string from "user custom case ids" box
```
zhx828: I'm confused. Why don't we process the sampleIdsStr outside of processData and make processData only accept sample arrary?

portal/src/main/webapp/js/lib/vc-session.js
```diff
@@ -146,6 +146,17 @@ window.vcSession = window.vcSession ? window.vcSession : {};
       _selectedCases.push(_selectedCase);
       return _selectedCases;
     };
+    
+    var _generateCohortDescription = function(_cases) {
```
zhx828: Where is this coming from? There is already a function called `generateCohortDescription_`

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
yichaoS: yes. The format of samples+studies from the summary page is different from the main query form. There has been a problem using the same format. I marked TODO for this so to find a better solution in the future.

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -400,27 +408,48 @@ else if (sampleIds != null && sampleIds.length()>0) {
...
-					SampleList defaultSampleSet = annotatedSampleSets.getDefaultSampleList();
-					if (defaultSampleSet == null) {
-						continue;
-					}
-					List<String> sampleList = defaultSampleSet.getSampleList();
-					if(inputStudySampleMap.get(cancerStudyId).size()>0){
-						sampleList.retainAll(inputStudySampleMap.get(cancerStudyId));
-					}
-					studySampleMap.put(cancerStudyId, sampleList);
+                    if (sampleSetId.equals("-1") && sampleIdsStr != null && sampleIdsStr.length() > 0) { //using user customized case list
```
yichaoS: No this doesn't get executed for every cohort.

portal/src/main/webapp/js/lib/vc-session.js
yichaoS: A merging problem. I'll remove this.

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
yichaoS: It would be the same. My thinking was to have process data to accept all raw parameters from the query form or summary page query. And there's also dispatch these parameters back to the result page, so I thought it would be confusing to process it outside "processData". Please let me know what you think.

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
kalletlak: could you give an example that's causing problem?

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
kalletlak: its under a for loop. `for (String cancerStudyId : inputStudySampleMap.keySet()) {...`

yichaoS: Is this the block you are referring to? sampleSetId.equals("-1") && sampleIdsStr != null && sampleIdsStr.length() > 0

core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
yichaoS: It's related to the scale of code change. I'll create another issue/PR for this.

PR #2377, opened 2017-04-19
User authorities are grouped by email and grouping is CASE SENSITIVE

mandawilson:
# What? Why?
Our users and authorities tables are set up to be case insensitive.  We can have one user in the authorities table with one email saved with different cases.  MyBatis does a **case sensitive** grouping on the user authorities email so one user in the database can be grouped into two separate sets of user authorities.  This was throwing an exception in MSK specific code.  Distinct users with emails of different cases in the authorities table can be found with this query:

	SELECT email, GROUP_CONCAT(DISTINCT case_sen_email), COUNT(DISTINCT case_sen_email) AS replicates
	FROM (SELECT CONVERT(email USING latin1) COLLATE latin1_bin AS case_sen_email, email FROM authorities) AS emails 
	GROUP BY email 
	HAVING replicates > 1;

- Convert all emails to lower case in SELECT queries.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers

@cBioPortal/backend

and notify them here:

### DIFF
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/SecurityMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/SecurityMapper.xml
@@ -4,6 +4,7 @@
 <mapper namespace="org.cbioportal.persistence.mybatis.SecurityMapper">
     <!-- do not cache these queries -->
 
+    <!-- user authorities are grouped by email and grouping is CASE SENSITIVE -->
     <resultMap id="userAuthoritiesResultMap" type="org.cbioportal.model.UserAuthorities">
         <result property="email" column="email"/>
         <collection property="authorities" ofType="String">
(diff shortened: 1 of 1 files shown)


ersinciftci:
It seems H2 database supports `LOWER` function, so it won't be a problem when we start writing unit tests for security module.

mandawilson:
@ersinciftci, @sheridancbio and I talked and we think the inserts should convert the email to lower case as well.  That is what core/src/main/scripts/importUsers.py does.  If we are worried about platforms not supporting SQL LOWER we could of course do this in a Java class instead.

mandawilson:
I think the retrieval conversion should always be kept just in case someone manually adds authorities or introduces a third way of programmatically adding them.  As you say @sheridancbio it is not harmful and there is a low cost to it.

mandawilson:
We should not merge this until we have also reviewed my changes to MSKCCPortalUserDetailsService.java (in MSK's Mercurial).

PR #2386, opened 2017-04-21
Do null check on data; Add spinning bar while downloading data;

### LINKED ISSUE #2343: Cytoband == NULL crashes CNA table
According to the data model, cytoband can be NULL. However, in 552306093e1dcf875b495042a43265c0c626bb7c code was added which gives an error when cytoband is NULL. This causes the CNA table in the Study View to crash.

screen shot 2017-04-14 at 14 36 40

zhx828:
Minor performance improvement on bigger study in study view clinical tab

# What? Why?
Fix #2343 

# Notify reviewers
@sandertan


zhx828:
Thanks, @sandertan.  @onursumer could you help me to review?

PR #2390, opened 2017-04-24
Gene set importer improvements

sandertan:
# What? Why?
Fix bug with study importer and improved perforance of importing geneset

- In study importer, z-score expression is now always imported after expression. This is important because the `source_stable_id` of the z-score expression file is the `stable_id` of the expression file.
- Improved performance of gene set importer by using the bulk loader (18x faster).

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneset.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneset.java
@@ -58,13 +58,11 @@ public class DaoGeneset {
      * @return number of records successfully added
      * @throws DaoException 
      */
-    public static int addGeneset(Geneset geneset) throws DaoException {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneset.java
```diff
@@ -78,17 +76,14 @@ public static int addGeneset(Geneset geneset) throws DaoException {
...
             rs = pstmt.getGeneratedKeys();
             rs.next();
             int newId = rs.getInt(1);
             geneset.setId(newId);
             
-            // add geneset genes
-            rows += addGenesetGenes(geneset);
-            
-            return rows;
+            return geneset;
```
angelicaochoa: 👍

PR #2397, opened 2017-04-27
sync portal and iviz repo

yichaoS:


portal/src/main/webapp/js/lib/vc-session.js
```diff
@@ -147,24 +147,23 @@ window.vcSession = window.vcSession ? window.vcSession : {};
       return _selectedCases;
     };
 
-    var generateCohortDescription_ = function(_cases) {
```
zhx828: line `150-167` was changed from the new changes. Why do we change back?

zhx828:
Please ignore my comments. This is just a synchronization pull. If any changes needed, we will update them next time specially those formats

yichaoS:
@zhx828 That particular line is my mistake. just fixed it.

PR #2398, opened 2017-04-30
Fix #2306 add _uncalled genetic mutations profile

### LINKED ISSUE #2306: handling uncalled events in the patient view page (supporting cfDNA projects)
- [x] [data format] decide whether a mutation is called or uncalled mutation in MAF -- check the current MAF to see if there is indication about uncalled events, if not, contact Nick.
- [x] [database] import uncalled mutations with `_uncalled` profile
- [x] [web api] get both called and uncalled mutation profiles on patient view page
- [x] [UI] use light color for the sample label for uncalled events in the mutations table
- [ ] Change pipeline to divide mutations in a uncalled and called txt. Or change importer to do it based on a specific column.

inodb:
Fix #2306. Allow import of uncalled mutations for uncalled mutations in ctDNA. This allows
import of genetic profiles postfixed with `_uncalled`.

Looking for some suggestions here i.e. whether it is necessary to create the `MUTATION_EXTENDED_UNCALLED` genetic data type or if it would be better to use `MUTATION_EXTENDED`. The idea is that we want to show uncalled mutations in the patient view, so we can show the number of supporting reads in the sequencing data, even though we can't call the mutation with confidence. At the moment we only want to show these on the patient view, everywhere else the mutations should be ignored.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/model/GeneticAlterationType.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/model/GeneticAlterationType.java
@@ -36,6 +36,7 @@ package org.mskcc.cbio.portal.model;
 // don't forget to change the other one too
 public enum GeneticAlterationType {
     MUTATION_EXTENDED,
+    MUTATION_EXTENDED_UNCALLED, // uncalled mutations for cfDNA
     FUSION,
     STRUCTURAL_VARIANT,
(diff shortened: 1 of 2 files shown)


inodb:
OK, never mind. Just using `show_profile_in_analysis_tab: false` seems to accomplish what we want

PR #2411, opened 2017-05-04
Add 1 endpoint in Mutation Enrichments API

ersinciftci:
POST /genetic-profiles/{geneticProfileId}/mutation-enrichments/fetch

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/MutationEnrichmentServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/MutationEnrichmentServiceImpl.java
@@ -1,0 +1,129 @@
+package org.cbioportal.service.impl;
+
+import org.cbioportal.model.AlterationEnrichment;
+import org.cbioportal.model.Gene;
+import org.cbioportal.model.Mutation;
+import org.cbioportal.model.MutationSampleCountByGene;
+import org.cbioportal.service.GeneService;
+import org.cbioportal.service.MutationEnrichmentService;
+import org.cbioportal.service.MutationService;
+import org.cbioportal.service.exception.GeneticProfileNotFoundException;
+import org.cbioportal.service.util.BenjaminiHochbergFDRCalculator;
+import org.cbioportal.service.util.FisherExactTestCalculator;
+import org.cbioportal.service.util.LogRatioCalculator;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.stereotype.Service;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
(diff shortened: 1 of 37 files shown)


service/src/main/java/org/cbioportal/service/impl/MutationEnrichmentServiceImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    private BenjaminiHochbergFDRCalculator benjaminiHochbergFDRCalculator;
+    @Autowired
+    private GeneService geneService;
+
+    @Override
+    @PreAuthorize("hasPermission(#geneticProfileId, 'GeneticProfile', 'read')")
+    public List<AlterationEnrichment> getMutationEnrichments(String geneticProfileId, List<String> alteredSampleIds,
+                                                             List<String> unalteredSampleIds,
+                                                             List<Integer> entrezGeneIds)
+        throws GeneticProfileNotFoundException {
```
onursumer: indentation looks weird here

service/src/main/java/org/cbioportal/service/impl/MutationEnrichmentServiceImpl.java
```diff
@@ -0,0 +1,129 @@
...
+            alterationEnrichment.getNumberOfSamplesInAlteredGroup());
+        alterationEnrichment.setEntrezGeneId(mutationSampleCountByGene.getEntrezGeneId());
+        alterationEnrichment.setHugoGeneSymbol(gene.getHugoGeneSymbol());
+        alterationEnrichment.setCytoband(gene.getCytoband());
+        setLogRatio(alterationEnrichment, alteredSampleCount, unalteredSampleCount);
+        setPValue(alterationEnrichment, alteredSampleCount, unalteredSampleCount);
+        return alterationEnrichment;
+    }
+
+    private void setQValue(List<AlterationEnrichment> alterationEnrichments) {
```
onursumer: I try to avoid `set...` methods if they are not simple setters, `calc...` might be better in those cases.

ersinciftci:
[(13) implementing APIs needed for the enrichments tab](https://trello.com/c/08kdSGis/231-13-implementing-apis-needed-for-the-enrichments-tab)

service/src/main/java/org/cbioportal/service/impl/MutationEnrichmentServiceImpl.java
ersinciftci: Indeed, but it's correct :)

service/src/main/java/org/cbioportal/service/impl/MutationEnrichmentServiceImpl.java
ersinciftci: You are right, I'll change it in the next CNA enrichments PR.

PR #2419, opened 2017-05-09
Fix to ImportGenePanel.java

n1zea144:
# What? Why?
Fix so that ImportGenePanel.java works directly from scripts jar.

- Explicit call to SpringUtil.initDatSource() prior to call to importData().


fedde-s:
We addressed the same issue more completely in #2427.

PR #2426, opened 2017-05-10
Fix hierarchy pop-up and update documentation

sandertan:
# What? Why?
- Fix bug with hierarchy pop-up, in which root node was returned twice by API. 
- Updated version numbers in documentation.
- Removed trailing white spaces in documentation.


pieterlukasse:
commit was moved directly to `geneset` branch

PR #2427, opened 2017-05-10
Document and complete support for loading gene panel data

fedde-s:
Building on @n1zea144's work in #2159, this will document how to load gene panels, and add support to annotate data sets with gene panels used using the documented Python data set loading CLIs.


fedde-s:
This includes the fix in #2419, together with documentation on how to use the fix, and the same fix in `ImportGenePanelProfileMap.java`.

PR #2429, opened 2017-05-11
Support logo without authentication

oplantalech:
# What? Why?
Some of the header style adjustments that @alisman made in `header_bar.jsp` (https://github.com/cBioPortal/cbioportal/commit/4348eaf) caused that now the institute logo is not displayed unless the cBioPortal instance has authentication configured. I believe this is due to the name of the `<div>`, which I also propose to change, apart from reverting the code to the one @jjgao wrote (https://github.com/cBioPortal/cbioportal/commit/8587270).


jjgao:
@oplantalech thanks for the pr. Could you please fix the screenshots?

alisman:
Hi guys, this was done for a reason having to do with the way that the nav bar gets pushed down.  I should have followed up sooner with a fix but chose to wait for my next release, which makes more changes to the header.  Please don't merge this as I think this will cause another problem.  I will fix this and make a hotfix PR

oplantalech:
@alisman Thanks for the answer, I did not know if anybody had noticed this bug so that's why I tried to change it back. Good to know that you are already working on a solution.

alisman:
Hi guys, I haven't managed to fix this.  The problem is that without institude logo, the nav is positioned improperly (b/c I rely on presence or non-presenece of sibling to reposition nav).  I don't know which is worse, alignment of nav at bottom or absence of logo.  This will be fixed in the next version of portal (in beta this week) so if it can wait a little longer, we should do so.

oplantalech:
I'm closing this since it has been fixed in release 1.7.0.

PR #2454, opened 2017-05-18
Fix for mutation_event comparison / mutation data loading script

pieterlukasse:
# What? Why?
Fix for issue with UQ constraint violation in `mutation_event` table when importing mutation data. It turns out MySQL is not case sensitive when it comes to UQ constraints. However, the java code that determines whether a new `mutation_event` record should be added or not was comparing values with `.equals()` which is case sensitive. This was leading to constraint violations, e.g.:

```
ABORTED! java.lang.RuntimeException: org.mskcc.cbio.portal.dao.DaoException: 
DB Error: only 1043 of the 1044 records were inserted in `mutation_event`. 
More error/warning details: Duplicate 
entry '15-41796352-41796353-NA-62524-E813*-stop_gained' for key 'CHR' 
See tmp file for more details: /tmp/mutation_event....tempTable
```

- Changed java code to be case insensitive when comparing mutation event objects

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/model/ExtendedMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/model/ExtendedMutation.java
@@ -307,17 +307,19 @@ public final class ExtendedMutation
             this.canonicalTranscript = canonicalTranscript;
         }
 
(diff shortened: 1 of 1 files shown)


pieterlukasse:
notify: @onursumer

PR #2474, opened 2017-05-23
Oncoprint heatmap clustering improvements

pieterlukasse:
# What? Why?
Improved clustering distance function to push samples with no data to outside. This solution should push samples that have no data to the outer edge of the resulting hierarchy (and of the resulting oncoprint heatmap). See screenshots.

**BEFORE**
image

**NOW (with solution propose here)**
image

Additional improvements:
- added missing LICENSE for clustering js library
- added support for clustering (sorting) on one track only (see screenshot below)

image

### DIFF
--- a/portal/src/main/webapp/js/src/clustering/clustering-worker.js
+++ b/portal/src/main/webapp/js/src/clustering/clustering-worker.js
@@ -65,20 +65,18 @@ var isAllNaNs = function(values) {
  *   
  */
 var preRankedSpearmanDist = function(item1, item2) {
-	//take the arrays from the preProcessedValueList:
-	var ranks1 = item1.preProcessedValueList;
-	var ranks2 = item2.preProcessedValueList;
-	var item1AllNaNs = isAllNaNs(item1.orderedValueList);
-	var item2AllNaNs = isAllNaNs(item2.orderedValueList);
 	//rules for NaN values:
-	if (item1AllNaNs && item2AllNaNs) {
+	if (item1.isAllNaNs && item2.isAllNaNs) {
 		//return distance 0
 		return 0;
 	}
-	else if (item1AllNaNs || item2AllNaNs) {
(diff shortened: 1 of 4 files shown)


portal/src/main/webapp/js/src/clustering/clustering-worker.js
```diff
@@ -54,11 +68,26 @@ var preRankedSpearmanDist = function(item1, item2) {
 	//take the arrays from the preProcessedValueList:
 	var ranks1 = item1.preProcessedValueList;
 	var ranks2 = item2.preProcessedValueList;
+	var item1AllNaNs = isAllNaNs(item1.orderedValueList);
```
adamabeshouse: this should probably be precomputed

pieterlukasse: hi @adamabeshouse, good point. Will do.

pieterlukasse:
@adamabeshouse I updated the code and also made a small fix for an existing bug (before this PR) I noticed during testing.

adamabeshouse:
Code looks good but heroku is crashing, I think you need to rebase onto rc

pieterlukasse:
@adamabeshouse rebased again.

jjgao:
@pieterlukasse can we make this to `hotfix` or `release-1.6.0`?

pieterlukasse:
@jjgao: changed to `hotfix`

adamabeshouse:
@pieterlukasse I'm still seeing errors on Heroku and when I tried to run it locally using the database that's used for hotfix (cgds_public), it says theres a database version mismatch.

pieterlukasse:
@adamabeshouse the database version mismatch is not related to this PR, but is rather a problem in the DB itself (it needs to be upgraded to the most recent version). Adding @sheridancbio to this thread (I think he is somehow involved in upgrading these DBs - see also #1806, recently merged)

PR #2525, opened 2017-06-02
Keycloak improvements

pieterlukasse:
# What? Why?
Made a number of improvements to keycloak documentation while testing/reviewing it. 

- various fixes and improvements to the documentation page
- fixed `header_bar.jsp` logout section to restore global logout URL (the logout parameter had been overwritten by mistake by @brittanystoroz in a previous PR)


pieterlukasse:
@khzhu can you review as well?

pieterlukasse:
@khzhu next to the fixes, I added some TODOs to the document. The main part missing is: how to link keycloak to your local authentication service? The current documentation only has an example of keycloak working as an authentication service itself...but of course the strength of this solution is to let keycloak work as a proxy between the institute's authentication service and cBioPortal.

khzhu:
Hi Pieter, 
I will have to ask our IT department for this, I've not involved in that process. What I normally be given are the URL to the keycloak server and a realm for cbioportal to use.  Here is what I know about the sever configuration:
- run in a standalone mode
- use mySQL as its persist storage
- sync user login information from local LDAP servers

I would agree with Ben and Manda, the cBioPortal part of keycloak document should focus mainly on how to authenticate user using keycloak, but not how to set up the keycloak server, since you can find very instructive online documents at the KeyCloak's own web site.(http://www.keycloak.org/documentation.html).

fedde-s:
Neither `docs/README.md` nor `docs/User-Authorization.md` nor `docs/Authenticating-Users-via-SAML.md` seem to refer to the document about KeyCloak. Are users supposed to find out that it is supported at all via some other link? Or is it too experimental to publically document it just yet?

pieterlukasse:
@khzhu @fedde-s thanks for the feedback. I made improvements based on your suggestions. 

Only 1 TODO remains (in next comment).

docs/Authenticating-Users-via-keycloak-SAML.md
```diff
@@ -105,9 +102,37 @@ keytool -genkey -alias secure-key -keyalg RSA -keystore samlKeystore.jks
...
+ the user also gains the roles associated with that composite. This inheritance is recursive so any composite of composites also gets inherited.
+![](images/previews/assign-role-to-user.png)
+
+### Institute's IDP users: assigning roles
+
+When integrating keycloak with your company-wide authentication service, the user credentials will reside at your institute's users directory and keycloak
+will work as a proxy between your LDAP (or SAML) IDP service and cBioPortal. Nevertheless, the authorization roles
+should still be configured for the users. The steps below detail how to do this.
+
+TODO
```
pieterlukasse: @khzhu here I would like to see more information about how to configure authorization rules for users that are not local keycloak users. Will the admin have to create local users anyway? If not, where to find these users in the keycloak menu?

pieterlukasse: @fedde-s maybe you can also help fill in this part once you start testing with keycloak?

khzhu: If @fedde-s can help with the testing that would be great. I am testing standard alone keycloak server with our IT department. Once the testing is done (tried to wrap up hg38-support this week). I will finish up your TODOs and fill more details in.
Thanks for the patience!

khzhu: @pieterlukasse 
to your questions:
> how to configure authorization rules for users that are not local keycloak users. Will the admin have to create local users anyway?

you can import external users to keycloak for example import from a LDAP server. However, you will need to manually assign roles to those external users (or using APIs, I've not tried this option yet, was one of Ben's question as well), Or alternatively you can create some default roles for a client, then those default roles will automatically apply to those external users of the client. And, those external users will shows up in the "Users" menu just as those users you created using keycloak.

fedde-s:
I tested and corrected the Keycloak configuration steps, tweaked the outline of headings of the documenation to make it easier to follow, resolved the conflict with `rc` by re-implementing @pieterlukasse's templating fix in the new context, and cleaned up a workaround from the SAML connection code that is not necessary if a relevant portal property is set to a more suitable value.

@khzhu, to adjust to the last change I mentioned, you would have to set this option in your `portal.properties` file:
```properties
filter_groups_by_appname=false
```

That causes the server to skip expecting all group (and study) names to be prefixed with `cbioportal:`; Keycloak has better ways to distinguish which permissions concern which service.

fedde-s:
@khzhu and @pieterlukasse, I could get rid of the revert and merge commits by re-basing the changesets in this pull request on top of the `rc` branch. Shall I do that or do you prefer to postpone it until the changes are reviewed and ready to release?

khzhu:
Hi @fedde-s, I normally rebase to rc and squash all commits to one for the review.

fedde-s:
@khzhu, I like to be careful to discuss with reviewers like yourself before replacing commits that they might already have started reviewing though, and to ask authors (@pieterlukasse in this case) for permission before replacing/removing commits they have pushed. It can be helpful during review to keep additional changes separate from ones already reviewed. As long as amendments/fixup commits are squashed before merging ­– a commit message like ‘Fix typo noticed by reviewer’ will not be very helpful to future devs trying to understand why a line of code was written.

khzhu:
@fedde-s, I agree. That is why I normally do squash before creating pull request. 

I've been working very closely with our IT personals last two weeks and have enough information to fill in more details on 
> how to link keycloak to your local authentication service?

Perhaps, add more details on how to use an external mysql db as its persist storage, which I thought was essential and safer for a production use.

docs/Authenticating-Users-via-keycloak-SAML.md
```diff
@@ -107,16 +111,22 @@ keytool -genkey -alias secure-key -keyalg RSA -keystore samlKeystore.jks
...
+## Integrate Keycloak with your company-wide authentication service
 
-When integrating Keycloak with your company-wide authentication service, the user credentials will reside at your institute's users directory and Keycloak
-will work as a proxy between your LDAP (or SAML) service and cBioPortal.
+When integrating Keycloak with your company-wide authentication service, the user credentials will reside at your institute's users directory and Keycloak will work as a proxy between your LDAP (or SAML) service and cBioPortal.
 
-Please refer to Keycloak documentation [http://www.keycloak.org/documentation.html](http://www.keycloak.org/documentation.html) for more information on how to integrate Keycloak with your local LDAP/SAML service.
+Please refer to [the Keycloak documentation](https://keycloak.gitbooks.io/documentation/server_admin/topics/user-federation.html) for more information on how to integrate Keycloak with your local LDAP/SAML service.
 
 ## Authorization with Keycloak
```
pieterlukasse: should we make this a level 1 header, just like the introduced "Authentication" section in line 1?

khzhu:
good idea, Pieter. thanks!

khzhu:
do you need me help with this. PM internal instance now is connect to local LDAP server. Also, we might want to add an additional section for how to connect to an external database. What do you think?

docs/Authenticating-Users-via-keycloak-SAML.md
fedde-s: No, a level 1 heading is semantically the heading for the page as a whole, and displayed as such in the Readthedocs outline. The top-level sections within the page are level 2 headings.

security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/keycloak/SAMLUserDetailsServiceImpl.java
```diff
@@ -107,8 +107,7 @@ public Object loadUserBySAML(SAMLCredential credential)
             if (attrName.equals(SAML_IDP_METADATA_ROLE_ATTR_NAME)) {
                 List<XMLObject> attributeValues = cAttribute.getAttributeValues();
                 if (!attributeValues.isEmpty()) {
-                    userRoles.add(new StringBuilder(APP_NAME).append(":").append(
-                        getAttributeValue(attributeValues.get(0))).toString());
+                    userRoles.add(getAttributeValue(attributeValues.get(0)));
```
pieterlukasse: in theory Keycloak will only send the roles for the specific client application, so this change should be OK. However, we just noticed that Keycloak's SAML response contains all roles of **all** clients. This means that the use case of having multiple cbioportal applications using the same keycloak service will not work...Possible actions:
- find Keycloak documentation to configure Keycloak in such a way that  only the client roles are sent back (you can log a support request - see http://www.keycloak.org/community.html ). Adjust our documentation and/or code accordingly.
- if Keycloak solution is not available: log an issue at Keycloak to ensure only the client roles are sent back
- document alternative way / workaround, e.g. adding "cbioportal:" in front of every Keycloak role

fedde-s:
@khzhu, setting up Keycloak with an external database is part of the Keycloak installation and not specific to cBioPortal. I'd say that it is better kept in one place: the Keycloak documentation itself. So far, we already say this about it:

> Please note that configuring your local instance to use Keycloak SAML authentication requires a Keycloak Identity Provider to be set up. For details on how to set up a Keycloak server, please read online document at https://keycloak.gitbooks.io/documentation/server_installation/index.html.

docs/Authenticating-Users-via-keycloak-SAML.md
pieterlukasse: Ok, so we could change the name of the doc to Authenticating-and-Authorizing-Users.... and the level 1 heading reflecting this. Then Authentication and Authorization sections could be level 2.

security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/keycloak/SAMLUserDetailsServiceImpl.java
fedde-s: I found the issue: when configuring the client in the Keycloak admin panel, the **Scope** tab seems to be set to ‘_Full scope allowed_’ by default. If I disable this, Keycloak only tells cBioPortal about the roles specifically defined for the cBioPortal client, and shows a panel to also explicitly include specific realm roles and other roles if required. If I add an instruction to flip this setting to our guide, we do not need any workarounds.

khzhu: keycloak sends all roles the user has. If the user can access multiple cbioportal instances using the same keycloak identity provider, then all roles of user will be included in the SAML credential. And, this will not cause any problems, since the permission evaluator of cBiPortal only matches the role of interest and ignore the others.

fedde-s: That's  the setting I found: I think we should document configuring Keycloak to only send only roles that belong the the specific cBioPortal instance (or _client_, in Keycloak parlance). That way, it is up to the Keycloak administrator to decide whether groups/studies in different instances should be visible to a user even if they share the same ID. This is what the `appname:` prefix does when roles are defined in the cBio database instead of in Keycloak.

khzhu: That should work in theory. Keycloak cloud server may be better solution for your scenario, in which you can have a dedicated keycloak for each client.

At PM, the purpose of choosing keylcoak is to create a user federation to unify the users from different institutes to have one portal which shows a different view for each individual base on the roles granted. Our clinicians had hard time to remember which patients in which portals for which version of datasets.

khzhu: @pieterlukasse @fcriscuo ,we will have to use "**APP_NAME**" to support both models (authorization by keycloak or other identity providers) till we stop using **users** and **authorities** two database tables.

pieterlukasse: @khzhu as @fedde-s mentioned, he found a solution to the problem I mentioned in the first comment of this thread. Given that you are using only one portal, this change does not really affect your installation in any way. Based on what I see in the code, the `users` and `authorities` tables are also not used anymore in case your instance is configured to use Keycloak. I.e. your `SAMLUserDetailsServiceImpl.java` class parses all authorities from SAML token only, and as far as I checked this is the only part of the login workflow where authorities are filled. 

So, I think this change is correct and we should keep it. Could you please test it on your side with this new code and let us know if it works for you? Although I do not think it is the case here, there is always a chance we missed something, so a test on your side would be nice. Thanks!

fedde-s: Do note that the change requires this line in your `portal.properties`:

```properties
filter_groups_by_appname=false
```

khzhu:
@pieterlukasse @fedde-s , 
the **CancerStudyPermissionEvaluator** consumes the AuthorityList created by **SAMLUserDetailsServiceImpl**. It expects the authority is in a format of **APP_NAME:CANCER_STUDY**. And, you can not adjust this format since the authorities table is still in use, although we do not use.

fedde-s:
Ah, I see where the confusion comes from. The `authorities` table is not used in installations that are configured to use Keycloak instead, and `CancerStudyPermissionEvaluator` only expects the *APP_NAME:CANCER_STUDY* format in installations for which the config property I mentioned is set to true or unset; see [this conditional](https://github.com/cBioPortal/cbioportal/blob/v1.8.1/security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java#L348).

khzhu:
@fedde-s , sorry haven't paid an attention to the tag, then we are safe :).

khzhu:
@fedde-s , the change made to logout did not work for me. Here are what I did:
1. adjust the header_bar.jsp page
`
                    - <a href="<c:url value="/saml/logout?local=true"/>">Sign out</a>
                    + 
                    + <a href="<c:url value='<%=samlLogoutURL%>'/>">Sign out</a>
`
2. portal.properties file
`
#saml.logout.url=/login.jsp?logout_success=true
saml.logout.local=false
saml.logout.url=/
`
anything I missed? Thanks!

screen shot 2017-08-17 at 11 06 54 am

fedde-s:
@khzhu, I seem to be having the same issue. But @pieterlukasse pointed out to me that when `saml.local.logout` was set to true, the logout button didn't actually log the user out at all, which is an even bigger problem.

khzhu:
@fedde-s , I am think if **_Fine rain SAML Endpoint Configuration_** has anything to do with this error. I left all fields there blank. Do you use any of this fields?
screen shot 2017-08-18 at 10 44 56 am

pieterlukasse:
excellent changes @fedde-s, thanks! :+1:

docs/Authenticating-and-Authorizing-Users-via-keycloak.md
```diff
@@ -0,0 +1,257 @@
...
+* Realm-level roles are in global namespace shared by all clients. 
+* Client roles have basically a namespace dedicated to a client. 
+* A composite role is a role that has one or more additional roles associated with it. 
+
+
+### How does Keycloak SAML Authentication work?
+
+Keycloak supports both OpenID-Connect and SAML authentication. When you use SAML authentication, the Keycloak server exchanges XML documents with a web application. XML signatures and encryption are then used to verify requests from the application.
+
+## Configure Keycloak to authenticate your cbioportal instance
```
oplantalech: Very minor comment: `cbioportal` should be changed to `cBioPortal`.

messersc:
After setting up Keycloak for a cbioportal instance, I have to say that the documentation in this PR is improved a lot and fixes some niggles that I had when using the old docs. 

Happy to see this get merged :+1:

sheridancbio:
Earlier this morning I tried building the beta war for MSKCC based on this code branch. I applied our overrides for building the mskcc internal portal (such as MSKCCPortalUserDetailsService.java MSKCCRestfulAuthenticationSuccessHandler.java mskcc-saml-metadata.xml, ...) and tried to adjust the portal.properties to include all necessary values (including saml.logout.url=/login.jsp?logout_success=true) but I don't think I was able to get the build correct. The final error I got was:
WARNING: Exception while attempting to add an entry to the access log
        at org.apache.catalina.connector.CoyoteAdapter.log(CoyoteAdapter.java:512)
        at org.apache.coyote.ajp.AjpProcessor.process(AjpProcessor.java:191)
But I believe that this was a failure to fully configure the application correctly. I will try again later, or may need some guidance from the frontend team if the process for building beta wars has evolved.

khzhu:
@sheridancbio ,
have you tried to roll back the change @fedde-s made in portal/src/main/webapps/WEB-INF/jsp/global/header_bar.jsp page (please see my comments made on Aug17)
to have something like and see what happens?
`<% if (authenticationMethod.equals("saml")) { %>`
 `<a href="<c:url value="/saml/logout?local=true"/>">Sign out</a>`
  `<%}`

security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/keycloak/SAMLUserDetailsServiceImpl.java
messersc: Independent of #3061, should `PUBLIC` studies still work? The user has not been given any authorities via Keycloak and I set
`always_show_study_group=PUBLIC` and imported `study_es_0`, but I don't see any studies in the portal.

This is relevant part of the log, I believe:
```
hasPermission(), checking permissions on targetDomainObject
getAppName(), APP_NAME = cbioportal
filterGroupsByAppName(), FILTER_GROUPS_BY_APP_NAME = false
PUBLIC_CANCER_STUDIES_GROUP= null
hasPermission(), cancer study stable id: study_es_0
hasPermission(), user: org.cbioportal.security.spring.authentication.PortalUserDetails@99cb6bf6: Username: xx@yy.de; Password: [PROTECTED]; Enabled: true; AccountNonExpired: true; credentialsNonExpired: true; AccountNonLocked: true; Not granted any authorities
```

khzhu:
@messersc ,
when you use the keycloak, you have to create roles for all studies in the portal, including those public datasets. So, always_show_study_group=PUBLIC flag will be ignored or has no any effects at all. 
The way to make a dataset public in keycloak is to grant all users in your system those roles assigned to the public dataset. For example, you have to create a role "study_es_0" in keycloak ,and then assign the role "study_es_0" to all users in your realm by default.

pieterlukasse:
@sheridancbio thanks for reporting back. The `MSKCCPortalUserDetailsService.java` should only be used when connecting your cBioPortal directly to MSKCC's SAML service. In this PR you would like to test the connection of cBioPortal to a local Keycloak server, which means you can follow the documentation as is.

pieterlukasse:
@khzhu @fedde-s I found the solution for the logout issue you discussed around here:
https://github.com/cBioPortal/cbioportal/pull/2525#issuecomment-323374237

The solution was indeed to set the logout URL to `http://localhost:8080/cbioportal/saml/logout` here:

image

fedde-s:
@pieterlukasse, could you add that to the documentation changes around the step where it explains which fields to fill out in that form? I used **boldface** when referring to text appearing in the UI and _italics_ or `verbatim monospace` for user input in most of the guide to help readers who are following along.

pieterlukasse:
- [ ] TODO: document the logout solution in my last comment.

portal/src/main/webapp/WEB-INF/jsp/global/header_bar.jsp
```diff
@@ -49,6 +49,9 @@
             "/" + GlobalProperties.getRightLogo() : GlobalProperties.getRightLogo();
     pageContext.setAttribute("rightLogo", rightLogo);
 %>
+<c:url var="samlLogoutUrl" value="/saml/logout">
+    <c:param name="local" value="${GlobalProperties.getSamlIsLogoutLocal()}" />
```
n1zea144: I understand this change is not part of this PR, but It looks like global logout is the default in GlobalProperties.  I don't think we want this as the default behavior.  From what I understand, a global logout would terminate other SP sessions, not just the cBioPortal session.  Unless there are objections, can we update the default in this PR?

fedde-s: I don't want this comment to get lost, but I will have to change this particular line of code for other reasons (Tomcat 7 compatibility, same issue as in https://github.com/cBioPortal/cbioportal/pull/3107). @pieterlukasse, I think you know more about the considerations Ben mentions here than I do.

fedde-s:
I rebased everything on the current state of the `rc` development branch, adjusting a line to address the Tomcat 7 issue I mentioned [in a comment above](https://github.com/cBioPortal/cbioportal/pull/2525#discussion_r140523540). The integration of the pull request functionality with the new state of `rc` still has to be tested before merging, probably by @pieterlukasse on Monday.

security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/keycloak/SAMLUserDetailsServiceImpl.java
pieterlukasse: @messersc for PUBLIC to work, you have to make sure the `groups` column in `cancer_study` table also contains PUBLIC for the public studies. To automate this, you need to make sure the `groups` meta data field in used in your `meta_study.txt` file when loading a study into cBioPortal. See here for more info: https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#cancer-study and https://github.com/cBioPortal/cbioportal/blob/master/docs/User-Authorization.md#configuring-public-studies

pieterlukasse: @messersc also remember when changing a portal.property like `always_show_study_group=PUBLIC`, you normally need to rebuild and redeploy the .war. Did you do this?

messersc: @pieterlukasse I rebuild my docker image, yes. Also, `study_es_0` has `PUBLIC` in its groups, I just checked. So that should have been fine.

So I think this comment https://github.com/cBioPortal/cbioportal/pull/2525#issuecomment-326616201 is correct? Or would you expect different behaviour?

pieterlukasse:
thanks @fedde-s . I tested on tomcat 7 and 8 and both work fine. On tomcat 8 it only works after applying the extra patch of https://github.com/cBioPortal/cbioportal/pull/3169, so the #3169 will need to be merged first.

pieterlukasse:
@jjgao ready to merge, but :warning: : merger **after** #3169

PR #2535, opened 2017-06-05
Expose show hotspot/oncokb/civic/mcg properties

inodb:
This exposes show hotspot/oncokb/civic/mycancergenome props to the frontend


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -718,7 +718,7 @@ public static String getSessionServiceUrl()
     {
         return sessionServiceURL;
     }
-
```
inodb: removed some trailing whitespace ;)

PR #2544, opened 2017-06-06
Convert Datatype in Genetic Profile back to enum

ersinciftci:
There are some profiles that doesn't have proper datatype values, we need to
revert this back until the data is fixed.


ersinciftci:
[remove GeneticProfile DATATYPE enum](https://trello.com/c/BSreVY1i/301-remove-geneticprofile-datatype-enum)

PR #2546, opened 2017-06-07
Validator & Importer: white spaces and [Not Available] in HGVSp_Short

sandertan:
# What? Why?
A previous PR added a change to block all whitespaces in `HGVSp_Short` column in MAF because of an issue found with a _surrounding whitespace_. This PR makes the validator less strict by allowing whitespaces, because some mutations, such as splice sites, do not have an amino acid change. Values such as 'splice site 1234+1A>t' or 'Met Met rearrangement' (in the case of Fusion) in this field would not be loadable.  

To trim  _surrounding whitespaces_ I added a .trim() function to the importer.

I also added 2 small improvements to this PR:

- Add `Fusion` as valid value for this `Variant_Classification`. It is allowed according to mutationMapper.js
- Add `[Not Available]` to the values parsed as 'not available'. This occurs in almost all TCGA studies. When the value is recognized as 'not available', the `Amino Acid Change` column is used for in this field. If there is no Amino Acid Change, the field says _MUTATED_.


sandertan:
Also added `Fusion` as valid value for this column.

sandertan:
Add `[Not Available]` to the values parsed as 'not available'.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -520,7 +520,7 @@ private String transformOMAScore( String omaScore) {
         } else if( omaScore.equalsIgnoreCase("N") || omaScore.equalsIgnoreCase("neutral")) {
             return "N";
         } else if( omaScore.equalsIgnoreCase("[sent]")) {
-            return ExtendedMutationUtil.NOT_AVAILABLE; // TODO temp workaround for invalid sent values
+            return "NA"; // TODO temp workaround for invalid sent values
```
pieterlukasse: ?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1221,11 +1221,11 @@ def checkAminoAcidChange(self, value):
         # may require bundling the hgvs package:
         # https://pypi.python.org/pypi/hgvs/
         if value not in self.NULL_AA_CHANGE_VALUES:
-            # give error if a white space is found in value (this leads to errors in
+            # give error if a trailing white space is found in value (this leads to errors in
```
pieterlukasse: please do for leading / trailing white space...not only trailing.

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
```diff
@@ -175,7 +179,7 @@ public static boolean isValidProteinChange(String proteinChange)
 		boolean invalid = proteinChange == null ||
 		                  proteinChange.length() == 0 ||
 		                  proteinChange.equalsIgnoreCase("NULL") ||
-		                  proteinChange.equalsIgnoreCase(NOT_AVAILABLE);
+		                  NOT_AVAILABLE.contains(proteinChange);
```
pieterlukasse: now this is case sensitive? It wasn't. Please fix, e.g. change list to all lower case and do:
`NOT_AVAILABLE.contains(proteinChange.toLowerCase())`

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
sandertan: This method was using the `ExtendedMutationUtil.NOT_AVALABLE` constant without being related to the `HGVSp_Short`-empty values. 

I changed the `HGVSp_Short`-empty values from a string to a list, so this method had some trouble what to return then. The other lines in this method return a character directly, so I changed this line to do that as well. See:

```
    private String transformOMAScore( String omaScore) {
        if( omaScore == null || omaScore.length() ==0) {
            return omaScore;
        }
        if( omaScore.equalsIgnoreCase("H") || omaScore.equalsIgnoreCase("high")) {
            return "H";
        } else if( omaScore.equalsIgnoreCase("M") || omaScore.equalsIgnoreCase("medium")) {
            return "M";
        } else if( omaScore.equalsIgnoreCase("L") || omaScore.equalsIgnoreCase("low")) {
            return "L";
        } else if( omaScore.equalsIgnoreCase("N") || omaScore.equalsIgnoreCase("neutral")) {
            return "N";
        } else if( omaScore.equalsIgnoreCase("[sent]")) {
            return "NA"; // TODO temp workaround for invalid sent values
        } else {
            return omaScore;
        }
    }
```

core/src/main/scripts/importer/validateData.py
sandertan: Done

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
sandertan: Thanks, I restored the previous functionality and added the new scenario:
```
public static final List <String> NOT_AVAILABLE = Arrays.asList("NA", "[Not Available]");
.......
proteinChange.equalsIgnoreCase(NOT_AVAILABLE.get(0)) ||
proteinChange.equals(NOT_AVAILABLE.get(1));
```

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
```diff
@@ -175,8 +179,8 @@ public static boolean isValidProteinChange(String proteinChange)
 		boolean invalid = proteinChange == null ||
 		                  proteinChange.length() == 0 ||
 		                  proteinChange.equalsIgnoreCase("NULL") ||
-		                  proteinChange.equalsIgnoreCase(NOT_AVAILABLE);
-
+		                  proteinChange.equalsIgnoreCase(NOT_AVAILABLE.get(0)) ||
```
zheins: This can be changed to `proteinChange.upper() in map(str.upper, NOT_AVAILABLE)` or something like that instead of having a discrete check for each item in the list.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1221,11 +1221,11 @@ def checkAminoAcidChange(self, value):
...
         # https://pypi.python.org/pypi/hgvs/
         if value not in self.NULL_AA_CHANGE_VALUES:
-            # give error if a white space is found in value (this leads to errors in
+            # give error if a leading or trailing white space is found in value (this leads to errors in
             # other layers, e.g. COSMIC keywords will not be matched)
-            if ' ' in value:
+            if value.rstrip() != value:
                 # return with an error message
-                self.extra = 'White space found in amino acid change column'
+                self.extra = 'Trailing white space found in amino acid change column'
```
zheins: Is this really a problem worth reporting? Can our java import code not handle stripping white space? If it can't we should just fix it in the java layer. I might just be misunderstanding here.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
pieterlukasse: hm...we need to think of another solution

core/src/main/scripts/importer/validateData.py
pieterlukasse: good point :+1: . Now that it is not just any white space, but leading and trailing white spaces only we could just use the `string.trim()` method in the java layer.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
pieterlukasse: Please use `MutationDataUtils.OMA_SCORE_NOT_AVAILABLE_VALUE` constant

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
pieterlukasse: @sandertan try `NOT_AVAILABLE.stream().anyMatch(s -> s.equalsIgnoreCase(proteinChange))`

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
pieterlukasse: @zheins what are these OMA scores? They don't seem to be in DB...maybe just remove this code??

pieterlukasse: seem to be used for filling the field `mutation_event.LINK_XVAR` with values like `getma.org/?cm=var&var=hg19,2,167263066,C,T&fts=all`....which are not even valid links (was this the old mutation assessor page)? Maybe @n1zea144 can comment as well?

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
zheins: @sandertan this still hasn't been updated. The solution @pieterlukasse suggested looks good, I for some reason thought I was looking at python when I wrote my comment :-)

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
```diff
@@ -45,7 +47,9 @@
  */
 public class ExtendedMutationUtil
 {
-	public static final String NOT_AVAILABLE = "NA";
+	// originally NA (case ignored) was the only value for empty values in this column, but [Not Available]
+	// occurs in many TCGA studies.
+	public static final List <String> NOT_AVAILABLE = Arrays.asList("NA", "[Not Available]");
```
zheins: If we're going to have a bunch of these that could potentially grow, maybe we should use an enum as in the [clinical data loader](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportClinicalData.java#L66). Just a suggestion, or maybe can be a TODO for later. I don't think it needs to block merging.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
n1zea144: Hi @pieterlukasse I'm not familiar with this code (@onursumer may have written it).  It does seem to be related to the mutation assessor.  @sheridancbio is still involved with the support of the site and he tells me that the getma.org reference is no longer active.  I think we still need to provide links to the site - either updating the code to support the new URL or finding another solution entirely.

onursumer: @pieterlukasse `getma.org` is dead, ideally we should replace them with the correct "mutationassessor.org" links. Currently, we are dealing with this dead links in the frontend by string replacing... But eventually, we hope to move mutation assessor related code from portal backend to genome nexus instead.

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
sandertan: @pieterlukasse @zheins done, thanks for suggestions :)

core/src/main/scripts/importer/validateData.py
sandertan: I removed this whole part in the validator and added the .trim() as Pieter suggested

core/src/test/scripts/test_data/mutations/data_mutations_wrong_aa_change.maf
```diff
@@ -6,4 +6,3 @@ A2M	2	genome.wustl.edu	37	12	9230409	9230409	+	Missense_Mutation	SNP	T	T	C	novel
...
-PIEZO1		genome.wustl.edu	37	16	88790292	88790292	-1	Missense_Mutation	SNP	T	T	C			TCGA-A1-A0SB-01	TCGA-A1-A0SB-10	T	T					Unknown	Untested	Somatic	Phase_IV	WXS	none	1	dbGAP	Illumina GAIIx	NA	NA	NA	Missense_Mutation	p.A195V p.I167I	Q9NQ94	NM_138932	NP_620310	A1CF_HUMAN	Q9NQ94	c.(583-585)GCG>GTG	c.584C>T	6	195	195	Missense_Mutation	p.A203V	A1CF	NM_138933	NP_620311	A1CF_HUMAN	Q9NQ94	c.(607-609)GCG>GTG	c.608C>T	8	203	203	medium	2.955	A195V	getma.org/?cm=msa&ty=f&p=A1CF_HUMAN&rb=138&re=203&var=A195V	getma.org/pdb.php?prot=A1CF_HUMAN&from=138&to=203&var=A195V	getma.org/?cm=var&var=hg19,10,52595854,G,A&fts=all	10	52595854	52595854	G	A	SNP	A1CF	NM_138932	human	genbank	57_37b	-1	reviewed	missense	c.584	p.A195V	1.000	superfamily_RNA-binding domain RBD,HMMSmart_SM00360,HMMPfam_RRM_1	superfamily_RNA-binding domain RBD,HMMSmart_SM00360,HMMPfam_RRM_1,superfamily_dsRNA-binding domain-like	-	no_errors
```
sandertan: This was removed because the unit test was removed.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
pieterlukasse: @sandertan please use MutationDataUtils.OMA_SCORE_NOT_AVAILABLE_VALUE constant

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
pieterlukasse: I would vote for a TODO comment here

sandertan: I've added a TODO.

sandertan:
@pieterlukasse I've implemented your requested changes in the latest commit (5dd476b)

sandertan:
Squashed and rebased on hotfix.

PR #2547, opened 2017-06-07
Fix mutation_count_by_keyword query

ersinciftci:
It was counting all the samples, instead of distinct samples. Added the new query to
migration.sql to fix the old data.


jjgao:
@sheridancbio this is a pull request to fix the data. The database scheme does not change. Do you think it's OK?

PR #2560, opened 2017-06-09
Portal project changes to enable populating query form selection with new react component

adamabeshouse:
# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -40,7 +40,8 @@
     <%
     String siteTitle = GlobalProperties.getTitle();
  
-  
+    String selectedCancerStudyId =
+		    (String) request.getAttribute(QueryBuilder.CANCER_STUDY_ID);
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/WEB-INF/jsp/index.jsp
```diff
@@ -57,7 +59,12 @@
 
     <jsp:attribute name="head_area" >
         <script>
-        window.loadReactApp({ defaultRoute: 'home' });
+            window.loadReactApp({ defaultRoute: 'home' });
+            window.selectedCancerStudyId = '${selectedCancerStudyId}';
```
alisman: as a rule we put the settings prior to loadReactApp invocation.   The rendering of components will sometimes be synchronous in which case the above wouldn't work.  For reasons I can explain, I will probably do this on homepage, so this is important here.

adamabeshouse: that makes sense

Issue #2534, opened 2017-06-05
Update Modify Query


onursumer:
When clicked on Modify Query button, show the query selector on top of the page, instead of inside a Modal.

PR #2566, opened 2017-06-09
Updated modify query button behavior

onursumer:
Upon clicking modify query button, showing the query selector on top of the page instead of inside a modal (fix issue #2534)

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
@@ -80,11 +80,14 @@
 
 <jsp:include page="global/header.jsp" flush="true"/>
 
+<script type="text/javascript" src="js/src/modifyQuery.js?<%=GlobalProperties.getAppVersion()%>"></script>
 
 <script>
 window.appVersion = '<%=GlobalProperties.getAppVersion()%>';
     
 window.historyType = 'memory';
+
+window.maxTreeDepth = '<%=GlobalProperties.getMaxTreeDepth()%>';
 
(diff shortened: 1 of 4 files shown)


portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
```diff
@@ -80,12 +80,15 @@
...
 <jsp:include page="global/header.jsp" flush="true"/>
 
+<script type="text/javascript" src="js/src/modifyQuery.js?<%=GlobalProperties.getAppVersion()%>"></script>
 
 <script>
 window.appVersion = '<%=GlobalProperties.getAppVersion()%>';
     
 window.historyType = 'memory';
 
+window.maxTreeDepth = '<%=GlobalProperties.getMaxTreeDepth()%>';
```
inodb: Let's move this to `template.tag` if possible. Nice to keep all properties in one spot. There are a bunch of other properties for the query component such as `priorityStudies`

onursumer: @inodb Actually this variable is already in `template.tag`.

I had a discussion with @alisman already about this. It is not trivial to include `template.tag` in those files (`visualize.jsp` and `cross_cancer_results.jsp`). So this is a temporary workaround for now.

alisman: probably what we should do is put all these things in include.  That way, we don't get bit when we replace these pages.   But I'll accept this PR and then we can follow up with that.

Issue #2567, opened 2017-06-12
Discrete Copy Number endpoint 404


inodb:
https://cbioportal-release-1-7-0.herokuapp.com/case.do#/patient?caseId=TCGA-A2-A0YH&studyId=brca_tcga

Two 404s

https://cbioportal-release-1-7-0.herokuapp.com/api/genetic-profiles/brca_tcga_gistic/discrete-copy-number/fetch?discreteCopyNumberEventType=ALL&projection=DETAILED

https://cbioportal-release-1-7-0.herokuapp.com/api/genetic-profiles/brca_tcga_gistic/copy-number-counts/fetch

PR #2572, opened 2017-06-12
Fix Distrece CNA error when alteration type is ALL

ersinciftci:
Removed the validation for CNA in GeneticDataService, because if alterationType=ALL,
Discrete CNA API is using that service.


inodb:
@ersinciftci fixes one of the errors in #2567. Screenshot test failing because of a previous style change it seems

PR #2577, opened 2017-06-13
add priority studies to results view and cross-cancer

alisman:
Priority studies not showing up in study selector on results page or cross cancer. We need to write config info to those jsps, since they are legacy pages 

https://github.com/cBioPortal/cbioportal/issues/2569

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
@@ -29,13 +29,14 @@
  - You should have received a copy of the GNU Affero General Public License
  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
 --%>
-
 <%@ page import="org.mskcc.cbio.portal.servlet.QueryBuilder" %>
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp
```diff
@@ -89,6 +90,28 @@ window.historyType = 'memory';
 
 window.maxTreeDepth = '<%=GlobalProperties.getMaxTreeDepth()%>';
 
+window.priorityStudies = {};
+<%
+List<String[]> priorityStudies = GlobalProperties.getPriorityStudies();
```
adamabeshouse: it would be great to put these lines into a jsp file so it can be DRYer

alisman: I tried it but it turns out to be a hassle with the tag file.  It was more work than it was worth.  I think we should just continue to work toward applying the template to all pages, which will solve the DRY issue once and for all.

Issue #2595, opened 2017-06-15
Study view crashes when clinical attribute contains special characters


pieterlukasse:
Steps to reproduce:

- take a test study and edit the clinical data file. e.g. in `/cbioportal/core/src/test/scripts/test_data/study_es_3/data_samples.txt` change column id `DAYS_TO_COLLECTION` to `% DAYS_TO_COLLECTION` (so add a % and a white space in front of it)
- load the test study
- open study view. It will fail to load with http 400 on request http://localhost:8080/cbioportal/api-legacy/clinicaldata/samples and study view will hang (see screenshot below)

![image](https://user-images.githubusercontent.com/2900303/27180688-e93d7a4a-51d4-11e7-8b05-b5abbc813907.png)

jensreeder:
Hi,

While loading data into a local instance, I encountered a similar behavior when the column id contains a . (dot) character.
Validator and loader handle it just fine, but the javascript then interprets the . as a js properties accessor and the summary page fails to load. Would be great if the validator catches these cases.

zhx828:
@pieterlukasse does the issue still exist after switching to new API?

sandertan:
@pieterlukasse I tested this, and the study view displays correctly when there's an attribute ID with % in it. In this screenshot, I gave the Subtype column the clinical attribute id `% DAYS_TO_COLLECTION`. Still want to restrict the attributes to alphanumeric and underscores, in #2596? If so, we need to add it to the documentation as well.

![screen shot 2018-01-09 at 17 04 23](https://user-images.githubusercontent.com/9624990/34730273-31968a74-f55f-11e7-84fb-8ea6714e4a4c.png)

pieterlukasse:
thanks @sandertan , did you also test with `.` (dot)? See @jensreeder's comment above.

sandertan:
@pieterlukasse Good point. With a `.` in the attribute name, the diagram is not drawn.

![screen shot 2018-02-05 at 16 10 33](https://user-images.githubusercontent.com/9624990/35811627-39cb8858-0a8f-11e8-9902-80c985015bc1.png)

pieterlukasse:
@zhx828 any news on this one?

zhx828:
@pieterlukasse apparently I missed this ticket, sorry for the delay. Do we have any public database having this issue?

zhx828:
I don't think this ticket still applicable in the new study view. Please let me know if you still see the issues. Close for now.

PR #2596, opened 2017-06-15
Fix attribute ids validation

pieterlukasse:
# What? Why?
Fix to prevent scenario described in #2595. 

- validation script now gives ERROR when clinical attribute ID contains any special character (except for _ underscore). Special characters are still allowed in attribute name and description (which is what appears on UI).


pieterlukasse:
here is a list of studies that will fail validation after this fix:

https://github.com/bsir/cBioPortal-analysis/blob/master/attribute_formatting_check.ipynb

@sandertan @ritikakundra let's discuss how to fix this in datahub.

n1zea144:
I think this is a good idea and we should coordinate the update of data with this fix.  I think @ritikakundra should help with this and update data in datahub that is out of sync with the public portal.

pieterlukasse:
thanks @sandertan . Yes, I would like to keep this rule since it makes our data a bit more developer friendly IMO (many developers wouldn't assume an ID to have spaces or other special characters).

jjgao:
@pieterlukasse should we still merge this? If so, could you rebase?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1792,6 +1792,15 @@ def checkHeader(self, cols):
                     extra={'line_number': self.line_number,
                            'column_number': col_index + 1,
                            'cause': col_name})
+            # do not allow special characters in attribute names
+            if not re.match(
+                    # regex to allow only alpha numerics and _ underscore:
+                    r'^[A-Za-z0-9_]+$', col_name):
```
jjgao: should we support `[^A-Za-z0-9\._-]`?

jjgao:
@pieterlukasse @n1zea144 I still think it's a good idea to normalize attribute IDs. We can query our database and see how much data we need to fix.

core/src/main/scripts/importer/validateData.py
pieterlukasse: @jjgao any specific public studies you have in mind?

jjgao:
@pieterlukasse any plan for this?

jjgao:
I am closing this one since it's very old and #2595 has been addressed in frontend. Please open a new one if this is still relevant.

PR #2645, opened 2017-06-23
Fix field mix up in Study API

ersinciftci:
Microarray and miRNA count fields were mixed up


ersinciftci:
https://github.com/cBioPortal/cbioportal/pull/2646

PR #2649, opened 2017-06-26
Validator: Check for wrong gene positions in mutation files

oplantalech:
# What? Why?
We have encountered MAF files where the start and end positions of genes are floats instead of integers (e.g. `16449539.0` instead of `16449539`). Having float numbers instead of integers breaks the chromosome plot in the patient view:
screen shot 2017-06-20 at 11 32 50

- Raise an error during validation when the gene positions in mutation files are not integers.


core/src/main/scripts/importer/validateData.py
```diff
@@ -1346,6 +1348,30 @@ def checkSwissProt(self, value):
                 return True
         # if no reasons to return with a message were found, return valid
         return True
+    
+    def checkStartPosition(self, value):
+        """Check that the Start_Position value is an integer."""
+        if value.isdigit() == False or (value.isdigit() and '.' in value):
+            self.logger.error(
+                'The start position of this gene is not '
```
fedde-s: Are you sure that this is about the end position of a gene?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1346,6 +1348,30 @@ def checkSwissProt(self, value):
...
+                       'column_number': self.cols.index('Start_Position'),
+                       'cause': value})
+        # if no reasons to return with a message were found, return valid
+        return True
+    
+    def checkEndPosition(self, value):
+        """Check that the End_Position value is an integer."""
+        if value.isdigit() == False or (value.isdigit() and '.' in value):
+            self.logger.error(
+                'The end position of this gene is not '
```
fedde-s: Same as in [my comment above](https://github.com/cBioPortal/cbioportal/pull/2649/files#r124505678): are you sure that this is about the end position of a gene?

core/src/main/scripts/importer/validateData.py
fedde-s: As I believe that it is about the start position of a DNA mutation.

fedde-s: According to [the MAF specification](https://wiki.nci.nih.gov/display/TCGA/Mutation+Annotation+Format+(MAF)+Specification):

> Lowest numeric position of the reported variant on the genomic reference sequence. Mutation start coordinate (1-based coordinate system).

oplantalech: @fedde-s yeap, it's the mutation position! Completely my fault, I'm going to change that.

core/src/main/scripts/importer/validateData.py
oplantalech: @fedde-s Same error as previous comment.

sandertan:
I tested this, and it works as expected.

I also tested scientific notation `1e6` (I have seen data transformation code that output chromosomal locations as scientific notation, probably by accident). The importer does not load this as integer, but as `0`, so this PR is correct in marking this as error.

oplantalech:
@jjgao Can you merge this PR? Thank you.

PR #2689, opened 2017-07-03
Utilize priorities from database

zhx828:
## Updates
- Implement column priority based on priority column from database
- Add config to set the limitation on number of charts to be visualized
- Added layout matrix
- Improve layout sorting method
- Relayout after resizing only if user hasn't moved charts
- Replace data-number with attribute-id;

## Priorities

The priority system is represented with a final score.
The higher the final (numeric) score, the higher priority assigned.

The default score is 1.

To promote certain chart in study view, please increase priority in the database to a certain number.
The higher the score, the higher priority it will be displayed in the study view.
If you want to hide chart, please set the priority to 0.
For combination chart, as long as one of the clinical attributes has been set to 0, it will be hidden.

Currently, we preassigned priority to few charts, but as long as you assign a priority in the database except than 1, these preassigned priorities will be overwritten.

**CANCER_TYPE**:          3000,
**CANCER_TYPE_DETAILED**: 2000

**Overall survival plot**:                400 (This is combination of OS_MONTH and OS_STATUS),
**Disease Free Survival Plot**:           300 (This is combination of DFS_MONTH and DFS_STATUS),
**Mutation Count vs. CNA Scatter Plot**:  200

**Mutated Genes Table**:          90,
**CNA Genes Table**:              80,
**study_id**:                     70,
**With Mutation Data Pie Chart**: 60,
**With CNA Data Pie Chart**:      50,
**`#` of Samples Per Patient**:     40,
**Mutation Count Bar Chart**:     30,
**CNA Bar Chart**:                20

**GENDER, SEX**:  9,
**AGE**:          8

## Layout
The layout is determined mainly based on priority. Higher priority will promote chart closer to the left-top. But if the layout solely based on priority, sometimes the layout may not be displayed as expected. See the example below. 
image
In order to improve the layout, we added a layout algorithm layer. The algorithm tries to fit all charts into a 2 by 2 matrix (Mutated Genes Table occupies 2 by 2 space). When a chart can not be fitted in the first matrix, the second matrixed will be generated. And the second matrix will have lower priority than the first one. If later chart can fit into the first matrix, then its priority will be promoted.
screen shot 2017-07-18 at 12 48 11 pm

In the example above, `With Mutated Data` and `With CNA Data`  will then be promoted to the empty space (The dashed border boxes in the above image) and the following image shows the expected result.
image

## Heroku
Heroku instance: https://column-priority.herokuapp.com/

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers
Code has been reviewed under https://github.com/cBioPortal/iViz/pull/503


sandertan:
Hi @zhx828 this looks like a very useful addition to cBio. I have 2 (related) questions regarding this implementation.

1. In your example you mention a priority of [2, 3, 4, 5]. The higher the total score, the higher the priority. This array contains values from all 4 tiers, how does that work? Because you explained that all variables belong to a specific tier.

2. Let's say I would add a custom clinical attribute, for example COFFEE_PER_DAY, and give it a score of 5000. It would end up in tier 4, with [0,0,0,5000], and it will be higher than the scores in tier 1, so it would move to the front. So what is the purpose of the total score? 

Also it would be great if you could add a markdown for the documentation section explaining this functionality.

zhx828:
Hi @sandertan I think both your questions are related, and totally understandable about the confusion. Tier system is just trying to help user understander stand how study view differentiates different charts. Not all of charts are generated based on clinical attributes(Tier 3) and some tiers are combinations of clinical attribtues(tier 2). These are predefined(manually added) charts on the front end. We need to find a way to layout them with certain priorities and at the same time, give the backend the flexibility to prioritize charts over these manually added charts.

At the end of the day, the final score will determine where the chart located. Even COFFEE_PER_DAY is in the fourth tier with score 5000, but the final score is higher than second, third and most of the first tiers. Actually, it will be treated as the first chart in study view since the highest predefined score is 3000, CANCER_TYPE.

I thought this is just for testing purpose. We may need to update the markdown later, but yeah, I agree. I should add it anyway and will update it if needed. Thanks.

Please let me know if you have more questions.

docs/File-Formats.md
```diff
@@ -125,6 +125,21 @@ The first four rows of the clinical data file contain tab-delimited metadata abo
 - Row 2: **The attribute Descriptions**: Long(er) description of each clinical attribute
 - Row 3: **The attribute Datatype**: The datatype of each clinical attribute (must be one of:  STRING, NUMBER, BOOLEAN)
 - Row 4: **The attribute Priority**: A number which indicates the importance of each attribute.  In the future, higher priority attributes will appear in more prominent places than lower priority ones on relevant pages (such as the [Study View](http://www.cbioportal.org/study?id=brca_tcga)). A lower number indicates a higher priority.
+    ```
+    To promote certain chart in study view, please increase priority in the database to certain number. The higher the score, the higher priority it will be displayed in the study view.
```
pieterlukasse: Maybe remove the reference to "database" here, since this page is just about documenting the file format.

pieterlukasse:
@sandertan the description states that: 
The weight of each tier is a multiple of 10. For example, a [2, 3, 4, 5] priority array can be calculated into a score 2345 from 2 x 1000 + 3 x 100 + 4 x 10 + 5.

This means that:

- if COFFEE_PER_DAY is given 5,000 in fourth tier, its score is 5,000
- if CANCER_TYPE is given 3,000 in first tier, its score is 3,000 x 1,000 = 3,000,000, thus it still has a higher priority than COFFEE_PER_DAY. 

So, the system of tiers helps ensure that the items given a high priority by developers (first 3 tiers) are not easily overridden by mistake.

portal/src/main/webapp/js/src/dashboard/cbio-vendor.js
```diff
@@ -2709,62 +2624,46 @@ window.DataManagerForIviz = (function($, _) {
...
       array = array.sort(function(a, b) {
-        var priorityA = a.priority || -1;
-        var priorityB = b.priority || -1;
-
-        if (priorityA === -1) {
-          return 1;
-        }
-        if (priorityB === -1) {
-          return -1;
+        var score = iViz.util.compareLists(a.priority, b.priority, false);
```
pieterlukasse: is this the same as `content.compareLists` in `iviz.js`? If yes, I don't see the expected multiplications by 1000, 100 and 10....  or perhaps this function is not really responsible for the priority sorting (and name is misleading)?

portal/src/main/webapp/js/src/dashboard/iviz.js
```diff
@@ -2255,6 +2351,165 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
+        ? clinicalAttrsPriority[id][3] : 1;
+    }
+
+    /**
+     * Convert priority array to a number
+     * @param {array} priority
+     * @return {number}
+     */
+    content.calcPriorityNumber = function(priority) {
+      return _.isArray(priority) ? parseInt(priority.join('')) : 0;
```
pieterlukasse: according to documentation, [10,20,30,40] = 10x1000 + 20x100 + 30x10 + 40 = 12,340. But this function makes [10,20,30,40] = 10,203,040, which is quite a different number!?

pieterlukasse: Ok, looking at code like... 
```javascript
    function getSecondTierPriority(id) {
      var priority = 1;
      if (id) {
        switch (id) {
        case 'DFS_SURVIVAL':
          var _dfsStatus = getForthTierPriority('DFS_STATUS');
```
...I understand now that some attributes are in fact preassigned to a tier, and all custom configuration for these attributes (in the fourth tier) will be calculated with the preassigned tier weight. E.g. in example above, `DFS_SURVIVAL` is part of second tier. If we add custom configuration for priority of `DFS_SURVIVAL` (fourth tier), e.g. priority = 20, then priority will in fact be 100x20=2,000 and **not** the documented 300+20=320. *This is unexpected, given it does not match the given documentation.*

portal/src/main/webapp/js/src/dashboard/cbio-vendor.js
zhx828: This function is comparing tiers, not the final score.

portal/src/main/webapp/js/src/dashboard/iviz.js
zhx828: Actually, this function is only used to get default priority front-end reserved for clinical attributes. If you have priority from the database other than 1, it will trump the default priority. For the case above, DFS_SURVIVAL is a combination chart from DFS_STATUS and DFS_MONTH. DFS_SURVIVAL has default priority [0,4,0,0] which means it has been categorized under the second layer with priority 4. The final score for DFS_SURVIVAL will be 400. But if you set DFS_MONTH in the database as 20, DFS_STATUS as 1, then the final score should be 10.5. I realize the second tier is calculated not expected, I will update that.

zhx828: I will fix `calcPriorityNumber` function

zhx828:
@pieterlukasse @sandertan any more suggestions?

docs/File-Formats.md
```diff
@@ -126,10 +126,10 @@ The first four rows of the clinical data file contain tab-delimited metadata abo
...
     ```
-    To promote certain chart in study view, please increase priority in the database to certain number. The higher the score, the higher priority it will be displayed in the study view.
+    To promote certain chart in study view, please increase priority to certain number. The higher the score, the higher priority it will be displayed in the study view.
     If you want to hide chart, please set the priority to 0. For combination chart, as long as one of the clinical attribute has been set to 0, it will be hidden.
     
-    Currently, we preassigned priority to few charts, but as long as you assign a priority in the database except than 1, these preassigned priorities will be overwritten.
+    Currently, we preassigned priority to few charts, but as long as you assign a priority except than 1, these preassigned priorities will be overwritten.
```
pieterlukasse: why do we need such an exception? Is this because all legacy studies already have 1 as priority in all attributes? I think -1 would make more sense, or are negative numbers also allowed?

docs/File-Formats.md
```diff
@@ -126,10 +126,10 @@ The first four rows of the clinical data file contain tab-delimited metadata abo
 - Row 3: **The attribute Datatype**: The datatype of each clinical attribute (must be one of:  STRING, NUMBER, BOOLEAN)
 - Row 4: **The attribute Priority**: A number which indicates the importance of each attribute.  In the future, higher priority attributes will appear in more prominent places than lower priority ones on relevant pages (such as the [Study View](http://www.cbioportal.org/study?id=brca_tcga)). A lower number indicates a higher priority.
     ```
-    To promote certain chart in study view, please increase priority in the database to certain number. The higher the score, the higher priority it will be displayed in the study view.
+    To promote certain chart in study view, please increase priority to certain number. The higher the score, the higher priority it will be displayed in the study view.
```
pieterlukasse: I just realized that this sentence contradicts the previous sentence above: "A lower number indicates a higher priority.". Please update the previous sentence as well to make it consistent.

portal/src/main/webapp/js/src/dashboard/cbio-vendor.js
pieterlukasse: but it returns a score, right? `var score = iViz.util.compareLists(a.priority, b.priority, false);`

pieterlukasse:
@zhx828 thanks for the fixes. I added some minor comments and one small documentation fix request. I will give it a test now. 

As a general remark, I think the implementation could be more straightforward without the "tiers" concept. In fact, as I understand it now, the tiers are not really used in the code in any special way. This means that this code: 
```javascript
  "priority": {
    "CANCER_TYPE": [
      3,
      0,
      0,
      0
    ],
    "CANCER_TYPE_DETAILED": [
      2,
      0,
      0,
      0
    ],
 ...
```
can be replaced by:
```javascript
  "priority": {
    "CANCER_TYPE": 3000,
    "CANCER_TYPE_DETAILED": 2000,
   ...
```
and respective changes in other places of the code. Since the idea of "tiers" is there but is not used in any special way, it just causes unnecessary confusion. If you decide to stick to the current "tiers" implementation, I would recommend adding a clear note about this in the code, to avoid confusion and perhaps add a TODO so that it can be simplified later. On the other hand, I am already quite happy this is *not* mentioned in the user documentation (File-formats.md) - there you kept the text simple and straightforward.

docs/File-Formats.md
```diff
@@ -125,6 +125,21 @@ The first four rows of the clinical data file contain tab-delimited metadata abo
...
+    Currently, we preassigned priority to few charts, but as long as you assign a priority except than 1, these preassigned priorities will be overwritten.
+    
+    CANCER_TYPE: 3000, CANCER_TYPE_DETAILED: 2000,
+    Overall survival plot: 400 (This is combination of OS_MONTH and OS_STATUS) 
+    Disease Free Survival Plot: 300 (This is combination of DFS_MONTH and DFS_STATUS) 
+    Mutation Count vs. CNA Scatter Plot: 200,
+    Mutated Genes Table: 90, CNA Genes Table: 80, study_id: 70, 
+    With Mutation Data Pie Chart: 60, With CNA Data Pie Chart: 50, 
+    Mutation Count Bar Chart: 30, CNA Bar Chart: 20,
+    GENDER: 9, SEX: 9, AGE: 8
```
pieterlukasse: before GENDER pie chart I do see another pie chart with title "# of Samples Per Patient". Is this a mistake in the code or is it missing in the documentation here?

pieterlukasse:
Test results: for breast cancer study, I found "With Mutation Data" and "With CNA Data" to be at an unexpected place (see screenshots):

Found:
image

Expected:
image

Questionremark: this will change the current default position of some of the elements in the study view, is this really intended? I must say the current one found on cbioportal.org looks nicer (see screenshot below). Why are cancer type and cancer type detailed preferred now?

image

pieterlukasse:
One small bug I noticed when zooming in/out with the browser (or just resizing the window): whatever order  I set by hand (by dragging the widgets around) is reset when the browser window is resized. This should not be the case (items should not move around to other places just because the window is resizing - at least not now since study view UI does not have a responsive layout).

pieterlukasse:
Test result: when using the custom configuration, I got the expected results (screenshot below), except for the "With Mutation Data", "With CNA Data" and "# of Samples Per Patient" pie charts which should **not** be rendered as the first charts.

These were my custom settings:
- GENDER: 10,000
- CANCER_TYPE: 10
- CANCER_TYPE_DETAILED: 11

This was the result:
image

docs/File-Formats.md
zhx828: You are correct. Because the current setting is using 1 as default, so we decided to use 1 as default in the front-end as well. Any integer would work, only 1 and 0 are treated specially

zhx828:
@pieterlukasse thanks so much for the detailed review. For # of Samples Per Patient, it is a special chart which has predefined priority as well. I will update that in the document.

For the tiers system, I agree. Thr original intention was trying to differentiate different types of charts. I will go ahead remove the tiers system in all documents, JSON file and the code as much as possible.

I'm glad you brought up the layout issue. In the brca_tcga example, what you found is expected. What you expect was the initial layout. Let me give you more background.
In current cbioportal.org, Cancer Type(Detailed) has different priorities across different studies. For genie, mskimpact, we always show Cancer Type(Detailed) before the survival plot, but another way around for other studies. This causes a lot of confusion and inconsistency in the study view. So we decided in the column-priority pull request, we make Cancer Type(Detailed) having the highest predefined priority. After doing so, the initial load which you are expecting to see was generated, but we realize the page will become really 'messy'. Then we decide to take a step further, which leads us to layout algorithm. It's function `updateLayoutMatrix` in the `mainTemplate.js` The algorithm tries to fit all charts into 2 by 2 matrix (Mutated Genes Table occupies 2 by 2 space), that's why `With Mutated Data` and `With CNA Data` be promoted to the empty space.(The dashed border boxes).

screen shot 2017-07-18 at 12 48 11 pm

Hope this can solve some confusions.

jjgao:
@zhx828 nice figure :)

pieterlukasse:
@zhx828 thanks for the details. I will review again once you've pushed the changes. 

Here are my remarks:
- if cancer type and cancer type detailed are to get a higher priority in only a few studies, then I think this should **not** be the default. The current solution is not backwards compatible and all studies (except genie and mskimpact) will look odd. Everyone migrating to this new version will have to change the priorities and reimport all studies, even if they were happy with the old defaults. So I would like to propose you change the defaults to what they currently are for most studies, and do the custom priority and reimporting of the special case studies like genie and mskimpact.
- I like the "smart filling of empty spaces" algorithm :+1: Please make sure this is documented. You could add the figure you made to the documentation.  
- regarding resizing: as far as I can see, the current study view has a fixed width and height, so there is never true resizing. I think it is strange to have widgets moving around in a fixed size region when resizing the window. Maybe you can turn off the algorithm as soon as the user has customized the view (i.e. moved one of the items around)?

zhx828:
@pieterlukasse thanks for the reply.

1. @jjgao could you comment on this?
2. Where do you think I should put this document?
3. I will update the algorithm

jjgao:
1. It's a good point, but I also think it makes sense to promote Cancer_Type and Cancer_Type_Detailed when there are multiple ones, especially in prep for virtual studies. @pieterlukasse if we give this two charts higher priority but hide them when there is only one type (it is the case for most studies), do you think it's ok?

pieterlukasse:
@jjgao regarding point 1: I think this should be OK, although this will affect studies like http://www.cbioportal.org/study?id=brca_tcga#summary .  Maybe it would be best to have the scatter plot appear before the survival curves. So instead of this:

image

we would have this: 

image

@zhx828 regarding point 2: please add to `File-formats.md`, close to the location where you document `Row 4: The attribute Priority: ` part. Or (maybe even better) make a separate `Attribute-priorities.md` page explaining what users can expect when setting the priorities and add your explanation and the figure there. Then in `File-formats.md` you just add a link to this new `Attribute-priorities.md` page.

pieterlukasse:
@jjgao testing some more, I discovered that your idea of hiding the cancer type / cancer type detailed lists when lenght = 1 already seems to be implemented in this PR (screenshot). I think I like it...but please clearly document this exception. 

image

side note (unrelated to this PR, but to the figure above): why does my SKCM_TCGA study have more cancer type detailed values than the one on cbioportal.org?

jjgao:
@pieterlukasse any more comments?

docs/File-Formats.md
```diff
@@ -125,6 +125,33 @@ The first four rows of the clinical data file contain tab-delimited metadata abo
...
+    
+    Please note: 
+    Priority is not the sole factor determining which chart will be displayed first.
+    A layout algorithm in study view also makes a minor adjustment on the layout.
+    The algorithm tries to fit all charts into a 2 by 2 matrix (Mutated Genes Table occupies 2 by 2 space).
+    When a chart can not be fitted in the first matrix, the second matrixed will be generated. 
+    And the second matrix will have lower priority than the first one. 
+    If later chart can fit into the first matrix, then its priority will be promoted.
+    
+    Please see [here](https://github.com/cBioPortal/cbioportal/pull/2689) for an example.
```
pieterlukasse: please do make a separate documentation page and link to that instead of linking to a PR.

docs/File-Formats.md
```diff
@@ -125,6 +125,33 @@ The first four rows of the clinical data file contain tab-delimited metadata abo
 - Row 2: **The attribute Descriptions**: Long(er) description of each clinical attribute
 - Row 3: **The attribute Datatype**: The datatype of each clinical attribute (must be one of:  STRING, NUMBER, BOOLEAN)
 - Row 4: **The attribute Priority**: A number which indicates the importance of each attribute.  In the future, higher priority attributes will appear in more prominent places than lower priority ones on relevant pages (such as the [Study View](http://www.cbioportal.org/study?id=brca_tcga)). A lower number indicates a higher priority.
+    ```
```
pieterlukasse: Please fix in line 127 (above): "A lower number indicates a higher priority."  Should be: "A higher number indicates a higher priority."

jjgao:
@zhx828 let's merge it soon. Could you rebase and fix the screenshots?

jjgao:
@zhx828 why the screenshot test failed. I don't see a difference.

zhx828:
@jjgao Ino mentioned could be the tooltip issue. I made a fix, hope this can fix the problem.

PR #2713, opened 2017-07-07
Integrate new mutations tab

adamabeshouse:
# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.


adamabeshouse:
@alisman how hard would this be to add to MSKTabs? 
image

onursumer:
http://maslianok.github.io/react-responsive-tabs/ 
@adamabeshouse @alisman how about something like this?

portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
```diff
@@ -45,6 +45,14 @@ window.maxTreeDepth = '<%=GlobalProperties.getMaxTreeDepth()%>';
 window.skinExampleStudyQueries = '<%=GlobalProperties.getExampleStudyQueries().replace("\n","\\n")%>'.split("\n");
 
 window.priorityStudies = {};
+
+// global variables required for MutationMapper annotation column
+window.showCivic = <%=GlobalProperties.showCivic()%>;
```
onursumer: I think, for now, we should move the global variables related to mutation mapper into a separate jsp file. Then we can include the same jsp file in the cross cancer page as well.

adamabeshouse: that sounds good to me

inodb: I think we should have all props for frontend defined in one file. That's easier then defining global variables per component/page. We use `template.tag` for that. We don't apply the template on all the pages, when we for instance call an exposed render method directly. I guess we can define another template or jsp that just has the imports. That can be included both by `template.tag` and wherever other components are mounted. So  `portal/src/main/webapp/WEB-INF/jsp/visualize.jsp` and `portal/src/main/webapp/WEB-INF/jsp/cross_cancer_results.jsp` should also just use this one file with all the properties.

PR #2733, opened 2017-07-12
Use mutation_count_by_keyword table in Variant Counts API

ersinciftci:
Use the pre-calculated table in Variant Counts API, instead of dynamically
calculating it.


ersinciftci:
[(5) switch to mutaton_count_by_keyword in api](https://trello.com/c/SFW3r22y/297-5-switch-to-mutatoncountbykeyword-in-api)

PR #2760, opened 2017-07-18
When no mutation data is available an empty list is returned

### LINKED ISSUE #2576: Patient view fails when user does not have access to a study with mutation data
Function metaData in DaoSampleProfile.java uses a join on id2MutationProfile.keySet() which, if the user does not have access to any study with mutation data leads to the query failing and the patient view showing an Oops error message.

j-hudecek:
# What? Why?
fixes #2576 

- return empty list if there no mutations

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@ersinciftci


jjgao:
@j-hudecek thank you!

ersinciftci:
The new patient view doesn't use `DaoSampleProfile.java` anymore, so I don't think this fix is required anymore? I feel like this issue https://github.com/cBioPortal/cbioportal/issues/2576 was happening in an old version of cbioportal with the old patient view, not the latest one.

PR #2792, opened 2017-07-19
Add enrichmentType parameter to Enrichments API

ersinciftci:
Add enrichmentType parameter to Enrichments API that can either be
`SAMPLE` or `PATIENT`, to indicate whether enrichment calculation will be based
on sample count or patient count.


jjgao:
@ersinciftci would you like to release this in 1.8.0? If so, we can still merge rc to release-1.8.0.

PR #2829, opened 2017-07-24
Adding clonal_status column to db and importer

gawilson79:
Want to add the clonal status of mutations to the mutation table. 

Edited the migration.sql script to create extra column - CLONAL_STATUS. The column placement had to be specified (in this case after NORMAL_REF_COUNT) otherwise the AMINO_ACID_CHANGE column was incorrectly filled.

In order for the meta_importer script to correctly insert the new field, 6 scripts were edited. It wasn't entirely clear what the purpose of each script was so would be good to clarify this before repeating the process for other proposed new fields (e.g. cluster, CCF).


ersinciftci:
@gawilson79 I think using BigDecimal instead of Float in API models is safer in terms of accuracy. Can you change `Float ccf` in `model/src/main/java/org/cbioportal/model/Mutation.java` to `BigDecimal ccf`?

ersinciftci:
And you need to add the new columns to `cgds.sql` as well, that's why unit tests are failing.

jjgao:
Closing this one since it's out of date.

PR #2836, opened 2017-07-24
Add 1 Endpoint to Clinical Data API

ersinciftci:
POST /studies/{studyId}/clinical-data/fetch

### DIFF
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
@@ -2,7 +2,7 @@
 <!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
 
 <mapper namespace="org.cbioportal.persistence.mybatis.ClinicalDataMapper">
-    <cache/>
+    <cache size="100000"/>
 
     <sql id="selectSample">
         clinical_sample.INTERNAL_ID AS "${prefix}internalId",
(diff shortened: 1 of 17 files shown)


persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -53,12 +53,24 @@
...
-                </foreach>
+                <if test="studyIds.size() == 1">
+                    cancer_study.CANCER_STUDY_IDENTIFIER = #{studyIds[0]} AND
+                    sample.STABLE_ID IN
+                    <foreach item="item" collection="sampleIds" open="(" separator="," close=")">
+                        #{item}
+                    </foreach>
+                </if>
+                <if test="studyIds.size() > 1">
+                    <foreach index="i" collection="sampleIds" open="(" separator="OR" close=")">
```
jjgao: You can also use `IN` for pairs:

https://dba.stackexchange.com/questions/34266/selecting-where-two-columns-are-in-a-set

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -53,12 +53,24 @@
                 cancer_study.CANCER_STUDY_IDENTIFIER = #{studyIds[0]}
             </if>
             <if test="sampleIds != null">
-                <foreach index="i" collection="sampleIds" open="(" separator="OR" close=")">
-                    cancer_study.CANCER_STUDY_IDENTIFIER = #{studyIds[${i}]} AND sample.STABLE_ID = #{sampleIds[${i}]}
-                </foreach>
+                <if test="studyIds.size() == 1">
+                    cancer_study.CANCER_STUDY_IDENTIFIER = #{studyIds[0]} AND
+                    sample.STABLE_ID IN
```
jjgao: Can we avoid joining tables? Would doing two queries (query internal sample IDs and query clinical data from internal sample IDs) improve the performance? Should we cache all samples?

ersinciftci: Inner joins are almost always faster than multiple queries. We can consider caching all samples and patients, but let's do it after the release.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
ersinciftci: The reason I didn't do it is our in-memory unit test database H2 doesn't support it, because it's not part of standard SQL, so we can't write unit tests for this case if we use this syntax. And I also didn't want to use MySQL(or any other DBMS) specific features. But of course we can use it if we have to (for virtual cohort for example), but I fixed the problem for now with a new endpoint, which was already needed anyway.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
jjgao: Sounds good. Let's see how it is doing without caching and revisit it if it's too slow.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
jjgao: Sounds good. 

Let's revisit this when we do this for the study view page / virtual cohorts.

jjgao:
Please fix the frontend code.

PR #2506, opened 2017-05-30
Validator and Importer: Fusion as mutation genetic profile

sandertan:
# What? Why?
This PR adds functionality to import fusion data as a mutation genetic profile, while the `stable_id` in the fusion metafile is `fusion`. This was discussed in https://github.com/cBioPortal/cbioportal/issues/2462 and in https://github.com/cBioPortal/cbioportal/pull/1438#issuecomment-229983745. In the future we should load fusion data as fusion genetic profile, and this workaround should be removed.

After adding this functionality as a workaround, I discovered that meta files for study data were parsed twice. ~~This was annoying, because it would require to do the workaround twice~~(moved workaround to java layer).  To prevent this, I repurposed `study_meta_dictionary` so that it can contain the `meta_dictionary` from different meta files. 

While doing this, I was confused by the inconsistent naming of some variables, i.e. these are synonyms: `meta_dictionary`, `metaDictionary`, `meta_dict` and `meta`. I refactored this and a small number of others, to make them consistent. 

I also added a lot of comments for clarity and adjusted white spaces to improve readability of the code.

In the process of testing importing fusion data with `study_es_0` I noticed an issue during the gene panel import. I updated documentation and added a fix together with @pieterlukasse .

# Changes proposed in this pull request:
- Add workaround to import fusion data as mutation data.
- Repurpose of meta_study_dictionary to prevent double validation.
- Refactor inconsistent naming of `meta_dictionary`.
- Add many comments and update styling to improve readability.
- Improvements to documentation for loading sample study `study_es_0`.
- Add perl wrapper script for loading gene panels.
- Gene panel documentation improvements.
- Fix for loading gene panels #2683 .

PR #2883, opened 2017-07-28
Validator & Importer: Rename column names fusion data

sandertan:
# What? Why?
Fix https://github.com/cBioPortal/datahub/issues/62.

### 1.
Rename in header of fusion data:
`DNA Support` -> `DNA_Support`
`RNA Support` -> `RNA_Support`

@ritikakundra will update the data on datahub. Is there an internal pipeline, which outputs these files, that also needs to be updated?

### 2.
Also made file naming of fusion files consistent to:
`data_fusions.txt`
`meta_fusions.txt`

(it was confirmed in Slack data-hub channel that these are the correct names, similar to `data_mutations.txt`)
### 3.
Updated documentation to:
`**stable_id**: fusion` which was enabled in #2506


sandertan:
I tested this one with msk_impact_2017 (had to make some manual changes because the current version from datahub does not pass the validator due to unrelated (errors))

PR #2896, opened 2017-08-01
Custom highlighting of driver and passenger mutations

oplantalech:
# What? Why?
The current cBioPortal Oncoprint has functionality to highlight actionable mutations based on a fixed set of information sources, where missense, inframe and truncating mutations are subdivided in putative driver and putative passenger mutations. This PR proposes a solution to allow for this highlighting to occur based on an annotation added to the data directly, so already present at the time of importing the data into the database. Specifically, two different types of annotations are included: "driver/passenger annotation" (which only allows for `Putative_Driver`, `Putative_Passenger`, `Unknown`, `NA` and empty values) and "tiers annotation" (which allows for any value you wish to include, e.g. tiers or categories where the current mutation will be a driver).

- Add new configuration options in `portal.properties`:  in `oncoprint.custom_driver_annotation.binary.menu_label` and `oncoprint.custom_driver_annotation.tiers.menu_label` you can specify the name that the filtering is going to have in the Oncoprint's "Mutation Color" menu. `oncoprint.custom_driver_annotation.default`, `oncoprint.custom_driver_tiers_annotation.default`, `oncoprint.oncokb_hotspots.default`, and `oncoprint.hide_passenger.default ` allow to customize which options in the "Mutation Color" menu of the Oncoprint are active by default.
- Extend the mutation data format: two columns, called `cbp_driver` and `cbp_driver_tiers` have been added to the format. `cbp_driver` only allows for the values`Putative_Driver`, `Putative_Passenger`, `Unknown`, `NA` and empty, while `cbp_driver_tiers` allows for free text, limited to 50 characters. Furthermore, two more columns are added (`cbp_driver_annotation` and `cbp_driver_tiers_annotation`), which allow to add annotations on the previous two columns and are limited to 80 characters.
- Make changes in the options of "Mutation Color" menu of the Oncoprint:
screenshot menu

- Add extra information in the tooltip: an icon shows the annotation source per sample, and when hovering it, the annotation information is displayed:
screenshot tooltip


docs/File-Formats.md
```diff
@@ -543,7 +543,7 @@ The following extra annotation columns are also important for making sure mutati
...
 
-The MAF format recognized by the portal (excluding the annotation columns already mentioned above) has 32 columns + 4 columns with information on reference and variant allele counts in tumor and normal samples. A more detailed example MAF can be found on our [Downloads](Downloads.md#maf-example) page. Description of each column is provided below:
+The MAF format recognized by the portal (excluding the annotation columns already mentioned above) has 32 columns + 4 columns with information on reference and variant allele counts in tumor and normal samples + 4 columns with custom annotation of driver and passenger mutations. A more detailed example MAF can be found on our [Downloads](Downloads.md#maf-example) page. Description of each column is provided below:
```
sandertan: This text was a bit confusing, and now it's even more confusing I think. Maybe we can make it nice with this PR? Something like:

The MAF format recognized by the portal (excluding the annotation columns already mentioned above) has:
- 32 columns [link tcga maf format]
- 4 columns with information on reference and variant allele counts in tumor and normal samples
- 4 columns with custom annotation of driver and passenger mutations. 

A more detailed example MAF can be found on our Downloads page[The download is not working. We can link this to the maf of `study_es_0`]. Description of each column is provided below:

docs/File-Formats.md
```diff
@@ -581,6 +581,10 @@ The MAF format recognized by the portal (excluding the annotation columns alread
 34. **t_ref_count (Optional)**: Reference allele count (tumor).
 35. **n_alt_count (Optional)**: Variant allele count (normal).
 36. **n_ref_count (Optional)**: Reference allele count (normal).
+37. **cBioPortal_binary_filter (Optional)**: "PUTATIVE_PASSENGER" or "PUTATIVE_DRIVER".
+38. **cBioPortal_binary_filter_annotation (Optional)**: field to give more information about the cBioPortal_binary_filter value (limited to 80 characters). This field can only be present if the cBioPortal_binary_filter is also present in the MAF file.
+39. **cBioPortal_multi_filter (Optional)**: free label that marks the mutation as a putative driver (limited to 20 characters).
+40. **cBioPortal_multi_filter_annotation (Optional)**: field to give more information about the cBioPortal_multi_filter value (limited to 80 characters). This field can only be present if the cBioPortal_multi_filter is also present in the MAF file.
```
sandertan: BioPortal_binary_filter does not fully describe the contents. What about:
- cbp_driver_passenger
- cbp_driver_passenger_annotation
- cbp_driver_class
- cbp_driver_class_annotation

The short name in column is also done in GDC MAF: https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/ . The multi annotations are all drivers ‘tiers’ so calling this 'driver_class' makes sense I think. Another option would be:
- cbp_custom_driver_class
- cbp_custom_driver_class_annotation

If you agree with renaming the columns, it would also be nice if the variable names are improved

sandertan: It would be nice if we can add a new markdown with screenshots that describes this functionality. I would suggest then to use examples with values that are used in real applications, to make the concept more easy to grasp.

portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
```diff
@@ -107,6 +116,17 @@
...
+                           <% }
+                           if (!GlobalProperties.getMultiCustomAnnotationMenuLabel().trim().equals("")) { %>
+                           <div class="checkbox"><label><input type="checkbox" name="multi_filter" value="multi_filter"/> <c:out value="${GlobalProperties.getMultiCustomAnnotationMenuLabel()}" /> <img id="colorby_multi_filter_info" src="images/button-seagreen.png" alt="multi filter" style="height:15px; width:15px; cursor:pointer;"/></label></div>
+                           <div id="multi_values" style="margin-left: 30px;">
+                               <c:forEach var="multiValue" items="${DaoMutation.getMultiFilterValues(requestScope[QueryBuilder.CANCER_STUDY_ID])}" varStatus="loop">
+                                   <div class="checkbox"><label><input type="checkbox" name="multi_filter_${loop.index}" value="${fn:escapeXml(multiValue)}" /> <c:out value="${multiValue}" /></label></div>
+                               </c:forEach>
+                           </div>
+                           <% } %>
```
sandertan: Change colors of cbioportal_driver_filter and cbioportal_multi_filter, the greens icons are confusing with greens in legend below OncoPrint.

sandertan: The icons are not working for me:
screen shot 2017-08-02 at 16 11 19

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -202,3 +202,10 @@ ucsc.build=hg19
 
 # default view in oncoprint (sample, patient (default))
 oncoprint.defaultview=patient
+
+# only set the custom annotation menu labels if we want custom annotations to appear 
+custom_annotation.binary.menu_label =
+custom_annotation.multi.menu_label =
```
sandertan: Is it important to have a property to name the labels in the OncoPrint? This applies a _cbioportal specific property_ to what might be a _study specific property_. I would suggest just calling them in the Oncoprint:
- Custom driver/passenger
- Custom driver class

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -202,3 +202,10 @@ ucsc.build=hg19
...
+
+# select the custom annotations by default
+custom_annotation.default = false
```
sandertan: Clarify comment text that this is specific for the OncoPrint view and regards custom driver/passenger annotation. I would also rename the property (custom_annotation.default) to something more descriptive such as `select_custom_driver_passenger_default` or `select_custom_driver_class_default`

sandertan: Small detail: remove spaces for consistency.

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -151,6 +151,12 @@ var tooltip_utils = {
 		if (d.oncokb_oncogenic) {
 		    ret.append('<img src="images/oncokb-oncogenic-1.svg" title="'+d.oncokb_oncogenic+'" style="height:11px; width:11px;margin-left:3px"/>');
 		}
+		if (d.binary_filter) {
+		    ret.append('<img src="images/button-yellow.png" title="'+d.binary_filter_annotation+'" alt="multi filter" style="height:11px; width:11px;margin-left:3px"/>');
+		}
+		if (d.multi_filter) {
+		    ret.append('<img src="images/button-seagreen.png" title="'+d.multi_filter_annotation+'" alt="multi filter" style="height:11px; width:11px;margin-left:3px"/>');
+		}
```
sandertan: Hovering over the icon for annotation info is not working for me (might be related that the icon itself is not displayed). Also I don’t think its clear you could hover over this. Would it be possible to include annotations in the tooltip between `Mutation:` and `Coverage:` ?

(sorry if I didnt place this comment above the correct section)

src/main/resources/portal.properties.EXAMPLE
sandertan: I think a user wants to have an option that by default, only his own passenger / driver system is selected (if present in the data). I think we can accomplish this as follows:
if custom_annotation.default == true: Disable OncoKb, Hotspots and maybe even Type

core/src/test/scripts/test_data/mutations/data_mutations_annotation_more_than_80_characters.maf
```diff
@@ -0,0 +1,4 @@
+Hugo_Symbol	Entrez_Gene_Id	Center	Tumor_Sample_Barcode	Verification_Status	Validation_Status	Mutation_Status	Sequencer	Chromosome	Start_position	End_position	Variant_Classification	HGVSp_Short	MA:FImpact	MA:link.MSA	MA:link.pdb	SWISSPROT	cbioportal_driver_filter	cbioportal_driver_filter_annotation	cbioportal_multi_filter	cbioportal_multi_filter_annotation
```
sandertan: Minor thing: I would rename the unit test data files to something more descriptive about the column it is testing.

portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
oplantalech: Now the colors are seagreen and red-purple. I think they are different enough from the colors in the oncoprint legend.

oplantalech: I forgot to add the images to the previous commit. This should be fixed now.

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -2214,6 +2271,84 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
...
+			window.QuerySession.setKnownMutationSettings(new_known_mutation_settings);
+
+			updateMutationColorForm();
+			updateSortByForm();
+			
+			oncoprint.suppressRendering();
+			updateRuleSets();
+			updateSortComparators();
+			State.refreshData();
+			oncoprint.releaseRendering();
```
pieterlukasse: is there no wrapper function that executes all 7 steps above? Are they all needed?

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -932,9 +938,43 @@ public static String getDefaultOncoprintView() {
         }
         return defaultOncoprintView.trim();
     }
-
+    
+    public static boolean useCustomDriverAnnotations() {
+        String useCustomDriverAnnotations = properties.getProperty(USE_CUSTOM_DRIVER_ANNOTATIONS);
```
pieterlukasse: if not set, the default should be true. I think this will do it:
```
String useCustomDriverAnnotations = properties.getProperty(USE_CUSTOM_DRIVER_ANNOTATIONS, "true");
```

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -1042,16 +1066,20 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
-	    'cosmic_count_thresh': 10,
-	    'recognize_hotspot': true,
-	    'recognize_oncokb_oncogenic': true,
+		'ignore_unknown': false,
+		'recognize_cbioportal_count': false,
+		'cbioportal_count_thresh': 10,
+		'recognize_cosmic_count': false,
+		'cosmic_count_thresh': 10,
+		'recognize_hotspot': !customDriverAnnotationsByDefault,
+		'recognize_oncokb_oncogenic': !customDriverAnnotationsByDefault,
```
pieterlukasse: ok 👍

pieterlukasse: will this also ensure the oncokb checkbox is unchecked?

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -251,6 +251,12 @@ public void setCivicUrl(String property) {
             }
         }
     }
+    
+    public static final String USE_CUSTOM_DRIVER_ANNOTATIONS = "oncoprint.custom_driver_annotation.default";
+    public static final String BINARY_CUSTOM_DRIVER_ANNOTATION_MENU_LABEL = "oncoprint.custom_driver_annotation.binary.menu_label";
+    public static final String CLASS_CUSTOM_DRIVER_ANNOTATION_MENU_LABEL = "oncoprint.custom_driver_annotation.class.menu_label";
+    public static final String SHOW_BINARY_CUSTOM_DRIVER_ANNOTATION = "oncoprint.custom_driver_annotation.binary.show";
```
oplantalech: We discussed it again with @sandertan and @Sjoerd-van-Hagen, and we decided to define a default name for both columns. Therefore, we need the "SHOW" attributes.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
oplantalech: In `portal.properties.EXAMPLE` the custom Driver Annotations are set to `false`. So do we want only to set this parameter to `true` when the property is empty, or also "by default"? (as stated in `portal.properties.EXAMPLE`)

portal/src/main/webapp/js/api/cbioportal-datamanager.js
oplantalech: Yes, this is for starting the checkboxes. So basically, when custom driver annotations is set to `true` in the `portal.properties`, then `customDriverAnnotationsByDefault` is `true`. Therefore, the checkboxes of `binary_filter`, `multi_filter` and the first class/tier of `multi_filter` are checked, and the checkboxes for `hotspot` and `oncokb` are unchecked. If `customDriverAnnotationsByDefault` is `false`, the checked checkboxes are inverted.

portal/src/main/webapp/js/src/oncoprint/setup.js
oplantalech: Before, those 7 steps were only used in [here](https://github.com/cBioPortal/cbioportal/pull/2896/files/50a05a46536d80c89ac3fa544b1ebfb1bd83db66#diff-a14ee7841f1c852a98e1959c6ea45a91L2206). I will extract this into a single function.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -202,3 +202,12 @@ ucsc.build=hg19
 
 # default view in oncoprint (sample, patient (default))
 oncoprint.defaultview=patient
+
+# custom annotation of driver and passenger mutations in the oncoprint 
+oncoprint.custom_driver_annotation.binary.menu_label=Custom cBioPortal binary annotation
+oncoprint.custom_driver_annotation.class.menu_label=Custom cBioPortal classes
```
pieterlukasse: please change to 
```
oncoprint.custom_driver_annotation.binary.menu_label=Custom binary annotation
oncoprint.custom_driver_annotation.class.menu_label=Custom tiers
```

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
pieterlukasse: yes, if omitted from portal.properties, it means "true" by default and the logic below will set it to false if it is not applicable. You can still have it in portal.properties.EXAMPLE (which has become a kind of place where the properties are documented, but have it commented out there. Something like:
```
# set this to false to disable the automatic selection of the custom driver and passenger mutations annotation in the oncoprint (instead of OncoKB and Hotspots). By default this is true. Setting this to false will select OncoKB and Hotspots as the default annotation source instead.
#oncoprint.custom_driver_annotation.default=false
```

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
pieterlukasse: hm...not sure it this makes sense. One will probably be looking to change these labels anyway, so there is little added value in having a default value for them. I prefer the added value of 2 less portal properties.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -941,10 +939,10 @@ public static String getDefaultOncoprintView() {
     
     public static boolean useCustomDriverAnnotations() {
         String useCustomDriverAnnotations = properties.getProperty(USE_CUSTOM_DRIVER_ANNOTATIONS);
-        boolean showBinaryCustomDriverAnnotation = showBinaryCustomDriverAnnotation();
-        boolean showClassCustomDriverAnnotation = showClassCustomDriverAnnotation();
+        String getBinaryCustomDriverAnnotationMenuLabel = getBinaryCustomDriverAnnotationMenuLabel();
+        String getClassCustomDriverAnnotationMenuLabel = getClassCustomDriverAnnotationMenuLabel();
```
pieterlukasse: style comment: please remove  "get" from the variable name.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -941,10 +939,10 @@ public static String getDefaultOncoprintView() {
...
         if (useCustomDriverAnnotations == null || useCustomDriverAnnotations.isEmpty() 
-            || !(showBinaryCustomDriverAnnotation || showClassCustomDriverAnnotation))
+            || (getBinaryCustomDriverAnnotationMenuLabel==null || getClassCustomDriverAnnotationMenuLabel==null))
```
pieterlukasse: I think you will need to compare to empty string as well, since the variable could be there, but just set to nothing. E.g. 
```
oncoprint.custom_driver_annotation.class.menu_label=
```

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -932,7 +936,24 @@ public static String getDefaultOncoprintView() {
         }
         return defaultOncoprintView.trim();
     }
-
+    
+    public static boolean useCustomDriverAnnotations() {
+        String useCustomDriverAnnotations = properties.getProperty(USE_CUSTOM_DRIVER_ANNOTATIONS, "true");
+        if (getBinaryCustomDriverAnnotationMenuLabel()==null && getClassCustomDriverAnnotationMenuLabel()==null)
```
pieterlukasse: I think you will need to compare to empty string as well, since the variable could be there, but just set to nothing. E.g. 
```
oncoprint.custom_driver_annotation.class.menu_label=
```

docs/File-Formats.md
```diff
@@ -613,12 +613,12 @@ A more detailed example MAF can be found in [study_es_0](https://raw.githubuserc
...
+39. **cbp_driver_class (Optional)**: free label/category that marks the mutation as a putative driver (limited to 20 characters). This field is free text. Example values for this field are: "Tier 1", Tier 2"...
+40. **cbp_driver_class_annotation (Optional)**: field to give more information about the cbp_driver_class value (limited to 80 characters). This field can only be present if the cbp_driver_class is also present in the MAF file. This field is free text. Example values for this field are: "Highly Actionable", "Potentially Actionable", "Not currently actionable"...
 
 The "cbp_driver" columns are used in the oncoprint, providing two customized annotations of driver and passenger mutations:
-![schreenshot mutation color menu](https://user-images.githubusercontent.com/20278013/28964502-52692514-790d-11e7-9d34-72e0ebc5d506.png) 
+![schreenshot mutation color menu](https://github.com/thehyve/cbioportal/blob/7b0897e5d22f6e72861fcf90dbe5e77662b1bd1f/docs/images/screenshot-mutation-color-menu.png?raw=true)
```
pieterlukasse: change to just (`images/screenshot-mutation-color-menu.png`)

docs/portal.properties-Reference.md
```diff
@@ -66,6 +66,18 @@ app.name should be set to the name of the portal instance referenced in the "AUT
...
 oncoprint.defaultview=sample
+```
+## Custom annotation of driver and passenger mutations
+cBioPortal allows for two different ways of displaying custom annotation of driver and passenger mutations in the MAF files (see [file formats](File-Formats.md#extending-the-maf-format). To enable the view of those annotations, you must specify a name for their labels (those labels appear in the oncoprint "Mutation color" menu):
+```
+oncoprint.custom_driver_annotation.binary.menu_label=Custom binary annotation
+oncoprint.custom_driver_annotation.class.menu_label=Custom tiers
+```
+
+By default, when at least one of the custom annotations is enabled, custom annotations are automatically selected, instead of OncoKB and Hotspots. If you do not wish to automatically select custom annotations, but select OncoKB and Hotspots instead, set this parameter to false:
```
pieterlukasse: I would write: "If you wish to change this default behavior and have OncoKB and Hotspots as the default selected annotations, set this parameter to false:"

pieterlukasse:
thanks @oplantalech , just 3 minor comments left.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
pieterlukasse: so use the "show" methods as discussed

docs/File-Formats.md
```diff
@@ -608,6 +612,13 @@ The MAF format recognized by the portal (excluding the annotation columns alread
...
+37. **cbp_driver_passenger (Optional)**: "PUTATIVE_PASSENGER" or "PUTATIVE_DRIVER".
+38. **cbp_driver_passenger_annotation (Optional)**: field to give more information about the cbp_driver_passenger value (limited to 80 characters). This field can only be present if the cbp_driver_passenger is also present in the MAF file. This field is free text. Example values for this field are: "Pathogenic" or "VUS".
+39. **cbp_driver_class (Optional)**: free label/category that marks the mutation as a putative driver (limited to 20 characters). This field is free text. Example values for this field are: "Tier 1", Tier 2"...
+40. **cbp_driver_class_annotation (Optional)**: field to give more information about the cbp_driver_class value (limited to 80 characters). This field can only be present if the cbp_driver_class is also present in the MAF file. This field is free text. Example values for this field are: "Highly Actionable", "Potentially Actionable", "Not currently actionable"...
+
+The "cbp_driver" columns are used in the oncoprint, providing two customized annotations of driver and passenger mutations:
+![schreenshot mutation color menu](images/screenshot-mutation-color-menu.png)
```
sandertan: I also have a design question regarding the class column. A user might have these classes in his `cbp_driver_class` column:
- Confirmed driver
- Unknown
- Confirmed passenger

If he would have these 3 values in this column, 'Confirmed Passenger' would show up as class for a Driver in the OncoPrint. That wouldn't make sense right? In that case, would he have to replace all 'Confirmed passenger'-values with NA? If so, I think it's important to explain this in `39. **cbp_driver_class (Optional)**:` and/or the text below. 

@pieterlukasse what do you think?

docs/File-Formats.md
```diff
@@ -608,6 +612,13 @@ The MAF format recognized by the portal (excluding the annotation columns alread
...
+39. **cbp_driver_class (Optional)**: free label/category that marks the mutation as a putative driver (limited to 20 characters). This field is free text. Example values for this field are: "Tier 1", Tier 2"...
+40. **cbp_driver_class_annotation (Optional)**: field to give more information about the cbp_driver_class value (limited to 80 characters). This field can only be present if the cbp_driver_class is also present in the MAF file. This field is free text. Example values for this field are: "Highly Actionable", "Potentially Actionable", "Not currently actionable"...
+
+The "cbp_driver" columns are used in the oncoprint, providing two customized annotations of driver and passenger mutations:
+![schreenshot mutation color menu](images/screenshot-mutation-color-menu.png) 
 
 ⚠️ Please make sure that, even if you are using the MAF format, all required columns specified in the beginning of the Data File section are present!
```
sandertan: Please move this up, after the list with MAF format columns. I'd suggest an improvement would be: 
......using the MAF format, all 4 required columns are present.

business/src/main/java/org/mskcc/cbio/portal/model/DBMutationData.java
```diff
@@ -36,4 +36,8 @@
     public String variant_read_count_normal;
     public String cosmic_count;
     public String keyword;
+    public String binary_filter;
+    public String binary_filter_annotation;
+    public String multi_filter;
+    public String multi_filter_annotation;
```
sandertan: I'm not a fan of these variables names, (see other comment). Can also change in a later iteration of this PR.

docs/File-Formats.md
```diff
@@ -608,6 +612,13 @@ The MAF format recognized by the portal (excluding the annotation columns alread
 34. **t_ref_count (Optional)**: Reference allele count (tumor).
 35. **n_alt_count (Optional)**: Variant allele count (normal).
 36. **n_ref_count (Optional)**: Reference allele count (normal).
+37. **cbp_driver_passenger (Optional)**: "PUTATIVE_PASSENGER" or "PUTATIVE_DRIVER".
+38. **cbp_driver_passenger_annotation (Optional)**: field to give more information about the cbp_driver_passenger value (limited to 80 characters). This field can only be present if the cbp_driver_passenger is also present in the MAF file. This field is free text. Example values for this field are: "Pathogenic" or "VUS".
+39. **cbp_driver_class (Optional)**: free label/category that marks the mutation as a putative driver (limited to 20 characters). This field is free text. Example values for this field are: "Tier 1", Tier 2"...
+40. **cbp_driver_class_annotation (Optional)**: field to give more information about the cbp_driver_class value (limited to 80 characters). This field can only be present if the cbp_driver_class is also present in the MAF file. This field is free text. Example values for this field are: "Highly Actionable", "Potentially Actionable", "Not currently actionable"...
```
sandertan: I think it's important to explain the ordering of these classes. If you would have these 3 examples values, a user wants them in this specific order.

core/src/test/scripts/test_data/study_es_0/brca_tcga_pub.maf
```diff
@@ -1,15 +1,15 @@
...
+#version 2.4
```
sandertan: Please remove these tabs.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -202,3 +202,13 @@ ucsc.build=hg19
...
+
+# labels to be displayed in oncoprint "Mutation color" menu for custom annotation of driver and passenger mutations in the oncoprint.
+# Set these properties to enable the respective menu options in oncoprint:
+# oncoprint.custom_driver_annotation.binary.menu_label=Custom binary annotation
+# oncoprint.custom_driver_annotation.class.menu_label=Custom tiers
+
+# set this to false to disable the automatic selection of the custom driver and passenger mutations annotation in the oncoprint 
+# (instead of OncoKB and Hotspots). By default this is true. Setting this to false will select OncoKB and Hotspots as the default 
+# annotation source instead.
+# oncoprint.custom_driver_annotation.default=false
```
sandertan: GitHub complains there's no new line here. Not sure if this is important.

sandertan: I first loaded test data without the properties. I think it's not correct to place the icons in this case, because I did not configure the customization correctly:

screen shot 2017-08-14 at 11 27 27

core/src/test/scripts/test_data/study_es_0/brca_tcga_pub.maf
```diff
@@ -1,15 +1,15 @@
...
...-			TCGA-A1-A0SB-01	TCGA-A1-A0SB-10	G	G					Unknown	Untested	Somatic	Phase_IV	WXS	none	1	dbGAP	Illumina GAIIx			NA				db9d40fb-bfce-4c3b-a6c2-41c5c88982f1	a3254f8e-3bbd-42fc-abea-a5f25b7648b3	c.463delG	p.Ala155ArgfsTer59	p.A155Rfs*59	ENST00000434431	2/3	0			0			TMEM247,frameshift_variant,p.Ala155ArgfsTer59,ENST00000434431,NM_001145051.2;TMEM247,intron_variant,,ENST00000432241,;	-	ENSG00000187600	ENST00000434431	Transcript	frameshift_variant	462/659	462/659	154/219	E/X	gaG/ga	COSM1408208,~rs70940616	1		TMEM247	HGNC	42967	protein_coding	YES	CCDS56117.1	ENSP00000388684	TM247_HUMAN		UPI0000366EF8	NM_001145051.2			2/3		Coiled-coils_(Ncoils):Coil,Pfam_domain:PF15444								-:0.0202	-:0.0439		1						HIGH	1	deletion		1	1	2	46707888	46707888	G	-	DEL	TMEM247	ENST00000434431	human	ensembl	69_37n	1	known	frame_shift_del	c.462	p.A155fs	83	NULL	NULL	-	no_stop_codon:bad_bp_length_for_coding_region	TMEM247	HGNC	ENSG00000187600	20	0	0.00	7	3	30.00	-	-	-	2	46707888	46707888	G	-	DEL	TMEM247	ENST00000434431	human	ensembl	69_37n	1	known	frame_shift_del	c.462	p.A155fs	83	NULL	NULL	-	no_stop_codon:bad_bp_length_for_coding_region	TMEM247	HGNC	-	-	-	PUTATIVE_PASSENGER	Test passenger	Class2	Class annotation
```
sandertan: Did you add `Class2` instead of `Class 2` on purpose to show the ordering?

src/main/resources/portal.properties.EXAMPLE
sandertan: When loading them with the correct properties, it works as described. I do have some more design questions ( @pieterlukasse ). (Point 6 most important).

1. For new users, it is not clear that a mouse hover above the purple / green icon shows the custom annotation.
2. The time it takes for the popup (above the green/purple icon) to show up, is very long. +2 seconds. Most people would have moved away by then. Can we make this instant?
3. The popup first popup (Mutation: 489A etc...) is large enough. Can we add the custom annotations here?
4. A popup in a popup sound like a bad idea to me.
5. The Classes/Tiers: I think we should use one name. Class makes the most sense to me atm because there is no underlying tiering). I know this can be customized, but let's make the example clear as well.
6. I think the BEST solution would be that every class/tier, passenger and driver get all their own colors. These colors would also be displayed in the legend below. This way the user can instantly see, which colors correspond to which class/passenger/driver. And a mouse hover would show the color, which corresponds to the VALUE of the column, not the column itself.

screen shot 2017-08-14 at 11 35 16

core/src/test/scripts/test_data/study_es_0/brca_tcga_pub.maf
oplantalech: It is a typo.

src/main/resources/portal.properties.EXAMPLE
oplantalech: I think it makes sense to have the icons, they just give information about the mutation stored in the database. If there is information about OncoKB, the icon is also shown there (even if you have the OncoKB annotation disabled). I think the tooltip and the menu should be seen as related, but separated parts.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1375,6 +1390,36 @@ def checkEndPosition(self, value):
         # if no reasons to return with a message were found, return valid
         return True
 
+    def checkDriverPassenger(self, value):
+        """Validate the values in the binary column."""
+        if value not in ('PUTATIVE_PASSENGER', 'PUTATIVE_DRIVER'):
```
sandertan: The MAF format usually has their values in title case, such as Missense_Mutation and Unknown. For consistency, we could have the examples also in this format, such as Putative_Passenger and Unknown. But case sensitivity is kind of annoying, so it would also be nice if the all values are validated and imported case insensitive, but saved in a fixed format.

src/main/resources/portal.properties.EXAMPLE
pieterlukasse: decisions:
- [ ] improve the popup by adding the annotation next to the icon. Format will be, e.g. : "Mutation V89A ( :cyclone: annotation), ( :o: annotation 2), etc". This largely solves comments 1 to 4 and 6 from @sandertan .
- [ ] for point 5 stick to tiers, but improve naming in code (make it more consistent)
- [ ] icons mentioned in point 6 will be improved and will not be simple a circle of one color. Feedback is still needed to decide on best icons.

docs/File-Formats.md
oplantalech: I think for now it is better to leave the functionality like it is.

docs/portal.properties-Reference.md
```diff
@@ -66,6 +66,18 @@ app.name should be set to the name of the portal instance referenced in the "AUT
 The default view in OncoPrint ("patient" or "sample") can be set with the following option. The default is "patient".
 ```
 oncoprint.defaultview=sample
+```
+## Custom annotation of driver and passenger mutations
```
oplantalech: I think it is better to include them in the "Oncoprint" section (look also at the properties, they start with `oncoprint`.

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -890,6 +901,19 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
 		    d.cosmic_count = cosmic_count;
 		    ret = ret || ((typeof cosmic_count !== "undefined") && (cosmic_count >= known_mutation_settings.cosmic_count_thresh));
 		}
+		if (known_mutation_settings.recognize_binary_filter) {
+		    if (d.binary_filter == "PUTATIVE_DRIVER") {
+			ret = true; //Set the element to be a driver
+		    }
+		}
+		if (known_mutation_settings.recognize_multi_filter) {
+		    for (var value in getMultiValuesMap()) {
```
adamabeshouse: to do this safely you need to add a `hasOwnProperty` check, or you can use a lo-dash function to iterate

portal/src/main/webapp/js/api/cbioportal-datamanager.js
```diff
@@ -890,6 +901,19 @@ window.initDatamanager = function (genetic_profile_ids, oql_query, cancer_study_
...
+			if (known_mutation_settings["recognize_"+value] && getMultiValuesMap()[value] == d.multi_filter) {
```
adamabeshouse: I dont think this (`"recognize_"+value`) is the cleanest way to do it.. I think `known_mutation_settings.recognize_custom[value]` or something like that

src/main/resources/portal.properties.EXAMPLE
adamabeshouse: I agree with @sandertan - the feature should either be enabled or not, it's not clear what those symbols mean otherwise.

docs/portal.properties-Reference.md
sandertan: In that case `###` would be more clear :)

oplantalech:
@sandertan @adamabeshouse @pieterlukasse I've reviewed all your suggestions and made some changes.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1375,6 +1390,36 @@ def checkEndPosition(self, value):
         # if no reasons to return with a message were found, return valid
         return True
 
+    def checkDriver(self, value):
+        """Validate the values in the cbp_driver column."""
+        if value not in ('PUTATIVE_PASSENGER', 'PUTATIVE_DRIVER', 'NA', 'Unknown'):
```
sandertan: Please make it `Putative_Passenger` and `Putative_Driver` so that it at least looks like MAF format. As I said in my previous comment: case sensitivity is annoying, so it would also be nice if the values are validated and imported case insensitive, but saved in a fixed format. We can always implement this vali/imp enhancement at a later time, but the first part not, because that will be in the data.

docs/portal.properties-Reference.md
```diff
@@ -68,6 +68,18 @@ The default view in OncoPrint ("patient" or "sample") can be set with the follow
...
 ```
 
+## Custom annotation of driver and passenger mutations
+cBioPortal allows for two different formats of custom annotation of driver and passenger mutations in the MAF files (see [file format documentation](File-Formats.md#extending-the-maf-format)). To enable the view of those annotations (just one or all both), you must specify a name for the label(s) that you want to enable (those labels appear in the oncoprint "Mutation color" menu):
+```
+oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
+oncoprint.custom_driver_annotation.tiers.menu_label=Custom tiers
+```
+
+By default, when at least one of the custom annotations is enabled, custom annotations are automatically selected, instead of OncoKB and Hotspots. If you wish to change this default behavior and have OncoKB and Hotspots as the default selected annotations, set this parameter to false:
```
oplantalech: Yes.

oplantalech: Should I rephrase the information?

sandertan: Yes, could you rephrase it to ".... one of the custom annotations is enabled in portal.properties and the required columns are present in MAF files, custom annotations are......."

sandertan:
An improvement would be if "NA" would be parsed as empty value, and cannot be a custom tier.

jjgao:
I added @n1zea144 and @sheridancbio. Could either of you review the backend/db change?

@oplantalech is there a test instance for this PR? Our Heroku instance does not contain any custom annotation so we cannot really test it. If you can set a test instance up, let's run it by @schultzn and @ecerami before merging.

oplantalech:
@jjgao You can test in http://cbioportal-test.thehyve.net/cbioportal/ with these genes: OR11H1 TMEM247 ABLIM1 ADAMTS20 CADM2 DTNB KAT2A MSH3 MYB NPIPB15 OTOR P2RY10 PIEZO1

tmazor:
We deployed this locally and have been testing it out. @priti88 and I have some feedback but it might be easier to discuss - can we set a time to talk next week?

jjgao:
Just wanted to point out that the changes to the mutation data were made on the legacy code, ie. not in the new API. Once this pr gets merged, when refactoring the OncoPrint code / store, we would need to expose these data too. Maybe as part of the PR, the data should get into the API? @oplantalech @ersinciftci.

pieterlukasse:
@jjgao I would propose that "getting this data into new API" should be part of the next refactoring PR. I.e. we just need to make sure this is not forgotten if the refactoring starts before this PR is merged (maybe log a ticket about this with label "refactoring"?).

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -202,3 +202,13 @@ ucsc.build=hg19
...
+
+# labels to be displayed in oncoprint "Mutation color" menu for custom annotation of driver and passenger mutations in the oncoprint.
+# Set any of these properties to enable the respective menu options in oncoprint:
+# oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
+# oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
+
+# set this to false to disable the automatic selection of the custom driver and passenger mutations annotation in the oncoprint 
+# (instead of OncoKB and Hotspots). By default this is true. Setting this to false will select OncoKB and Hotspots as the default 
+# annotation source instead.
+# oncoprint.custom_driver_annotation.default=false
```
sandertan: `# oncoprint.custom_driver_annotation.default=false`
If this enabled, this option disables OncoKB and Hotspots, and ONLY enables the Putative_Driver (Binary) checkbox. The class checkbox cannot be enabled, because it is not known which class is preferred. The name and documentation should make this clear.

pieterlukasse:
I did various tests on the last code and all passed :+1:

core/src/main/java/org/mskcc/cbio/maf/MafUtil.java
```diff
@@ -560,7 +560,11 @@ public MafRecord parseRecord(String line)
 		record.setDriverFilter(TabDelimitedFileUtil.getPartStringAllowEmpty(driverIndex, parts));
 	    }
 	    record.setDriverFilterAnn(TabDelimitedFileUtil.getPartStringAllowEmpty(driverAnnIndex, parts));
-	    record.setDriverTiersFilter(TabDelimitedFileUtil.getPartStringAllowEmpty(driverTiersIndex, parts));
+	    if (driverTiersIndex != -1 && parts[driverTiersIndex].trim().equals("NA")) {
```
pieterlukasse: you could write this as:
`if (TabDelimitedFileUtil.getPartStringAllowEmpty(driverTiersIndex, parts).equals("NA"))`

or better write a method `TabDelimitedFileUtil.getPartStringAllowEmptyAndNA` and just use that.
```java
	public static String getPartStringAllowEmptyAndNA(int index, String[] parts)
	{
		String value = getPartStringAllowEmpty(index, parts);
		if (value.equals(NA_STRING)) {
				value = "";
		}
		return value;
}
```

core/src/main/scripts/importer/validateData.py
```diff
@@ -1463,15 +1463,15 @@ def checkEndPosition(self, value):
...
+        if value not in ('Putative_Passenger', 'Putative_Driver', 'NA', 'Unknown', ''):
+            self.extra = 'Only "Putative_Passenger", "Putative_Driver", "NA", "Unknown" and "" (empty) are allowed.'
             self.extra_exists = True
             return False
         return True
     
     def checkDriverTiers(self, value):
         """Report the tiers in the cbp_driver_tiers column (skipping the empty values)."""
-        if value is not "":
+        if value not in ('NA', ''):
```
pieterlukasse: I think we are keeping track of such lists in variables like `NULL_AA_CHANGE_VALUES`, `NULL_VALUES`

j-hudecek:
Hi, this is a great feature, would it be possible to display this information as extra columns on the mutations tabs in query view and patient view?

pieterlukasse:
hi @j-hudecek , thanks for the suggestion. Yes, this is something we have in our backlog. Not sure yet when this part will be implemented.

sandertan:
I've tested the latest version and the found no issues with the current implementation. The added portal properties are a nice addition to control the default behavior of checkboxes selection.

core/src/main/java/org/mskcc/cbio/maf/MafUtil.java
```diff
@@ -426,6 +437,14 @@ else if(header.equalsIgnoreCase(T_REF_COUNT) ||
...
             } else if(header.equalsIgnoreCase(NORMAL_VAF)) {
                 normalVafIndex = i;
+            } else if(header.equalsIgnoreCase(DRIVER_FILTER)) {
+            		driverIndex = i;
+            } else if(header.equalsIgnoreCase(DRIVER_FILTER_ANNOTATION)) {
+            		driverAnnIndex = i;
+            } else if(header.equalsIgnoreCase(DRIVER_TIERS_FILTER)) {
+            		driverTiersIndex = i;
+            } else if(header.equalsIgnoreCase(DRIVER_TIERS_FILTER_ANNOTATION)) {
+            		driverTiersAnnIndex = i;
```
sheridancbio: Whoa .. this monster statement really should be reframed as a hashmap lookup or something. [ in a future cleanup ]

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -1465,4 +1471,92 @@ public static void deleteAllRecords() throws DaoException {
...
+                        driverValues.add(rs.getString("DRIVER_FILTER"));
+                    }
+                } catch (SQLException e) {
+                    throw new DaoException(e);
+                } finally {
+                    JdbcUtil.closeAll(DaoMutation.class, con, pstmt, rs);
+                }
+            }
+        }
+        if (driverValues.size() > 0) {
```
sheridancbio: These lines could be simplified to "return driverValues.size() > 0;"

sheridancbio:
I think @jjgao makes a good point that this expansion of functionality is delivered through a legacy Web API, and it would be preferable to deliver new functionality through the new API. But other features (such as the mutation fingerprint) have been first deployed through the "legacy" api, which also has served as a developmental / trial area for new features before the public API is expanded, so I think that can be done in this case as well. But the update of the public API should be planned / scheduled soon.

priti88:
@tmazor and I were able to test the PR locally and everything looks good.

jjgao:
I am going to merge this. 

@oplantalech @pieterlukasse could you make it into the new API soon. We will need it when we refactor the oncoprint tab -- we will do that in the next few weeks.

pieterlukasse:
@jjgao thanks. I made a ticket regarding the new API topic: https://github.com/cBioPortal/cbioportal/issues/3086

pieterlukasse:
hi @j-hudecek, logged the enhancement request in #3095. Feel free to :+1: it there and add your ideas.

PR #2945, opened 2017-08-08
Remove special case for SAML institute logo URL

fedde-s:
The template logic for the header bar contained a special case in which
the institute logo was be prepended with a slash when using SAML
authentication, which caused issues when no logo was applicable
(signalled by an empty string). The logo seems to display just fine
without the special case, and removing it both solves the issue and
gets rid of some non-templating logic in a JSP scriptlet.

### Before:
The top-right corner showed the alternative text of the logo (the literal words ‘Institute logo’) next to the account actions icon, when no logo was applicable to a SAML-authenticated portal.
Screenshot showing the literal words 'Institute logo' next to the account action icon in the top right

### After:
The top-right corner does not show anything there in those circumstances, exactly as it does when SAML authentication is not enabled.
Screenshot showing only the account action icon in the top right without any logo

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/global/header_bar.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/global/header_bar.jsp
@@ -43,12 +43,9 @@
     else if (authenticationMethod.equals("googleplus") || authenticationMethod.equals("saml") || authenticationMethod.equals("ad")) {
         principal = "principal.username";
     }
-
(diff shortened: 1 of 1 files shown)


Sjoerd-van-Hagen:
@jjgao can this one be merged?

jjgao:
@Sjoerd-van-Hagen @fedde-s we need someone to approve it.

@alisman does this look good to you?

PR #2949, opened 2017-08-08
update frontend version for hotfix release and add new dev mode clear…

alisman:
… buttons

# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend
@cBioPortal/backend
@cBioPortal/devops

and notify them here:

### DIFF
--- a/portal/src/main/webapp/js/src/load-frontend.js
+++ b/portal/src/main/webapp/js/src/load-frontend.js
@@ -1,3 +1,10 @@
+
+function clearDevState(e){
+    localStorage.removeItem('localdev');
+    localStorage.removeItem('heroku');
+    window.location.reload();
+}
+
 window.loadReactApp = function(config) {
 
     // Set frontend route to /patient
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/js/src/load-frontend.js
```diff
@@ -18,6 +25,7 @@ window.loadReactApp = function(config) {
                 '<div class="alert alert-warning">' +
                 '<button type="button" class="close" data-dismiss="alert">&times;</button>' +
                 'cbioportal-frontend dev mode, using localhost:3000' +
+                '&nbsp;<a href="#" onclick="clearDevState()">clear</a>' +
```
adamabeshouse: this is a good idea. while you're at it, do you want to add something in here that looks for a URL parameter "localdev" or "heroku" and sets it automatically?

alisman: I thought about doing this but I think it's a security no-no.  Suppose we let someone designate a heroku instance.  What's to stop a hacker from sending someone a link with the hacker's heroku instance in the url param?  Anyone can make an app on the herokuapp.com domain.  So I think the only safe way to do this would be to have a custom herokuapp domain and limit embeds to that.  Maybe that's possible @inodb.  If we could do what @adamabeshouse suggests, it would make it easier to send around things for review.

adamabeshouse: right, now I remember the discussion we had. you're right. I think it would definitely be worth it for that reason to set up some secure way to do it

PR #2953, opened 2017-08-09
Add support of negative values to the bar rule set

rnugraha:
# What? 
This pull request is to add support of negative values when displaying clinical track in the OncoPrint tab. 

# Why?
Currently in the clinical track, all values whether they are positive or negative are represented by positive bar plot with the same color.

- Non positive values are plotted as negative bar with default colour red.
- Non negative values are plotted as positive bar with default colour green.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

# Any screenshots or GIFs?

## Before
screen shot 2017-08-09 at 16 08 32

## After
screen shot 2017-08-09 at 15 29 39

# Notify reviewers
@cBioPortal/frontend

### DIFF
--- a/portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintlegendrenderer.js
+++ b/portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintlegendrenderer.js
@@ -127,6 +127,17 @@ var OncoprintLegendView = (function() {
 	    root.appendChild(svgfactory.text(display_range[1], 50, 0, 12, 'Arial', 'normal'));
 	    var mesh = 100;
 	    var points = [];
+        var linear_gradient = svgfactory.linearGradient();
+        if (config.range_type === 'NON_POSITIVE') {
+            linear_gradient.appendChild(svgfactory.stop(100, config.negative_color));
+        } else if (config.range_type === 'NON_NEGATIVE') {
+            linear_gradient.appendChild(svgfactory.stop(100, config.positive_color));
+        } else if (config.range_type === 'ALL') {
+        	var offset = Math.abs(display_range[0]) / (Math.abs(display_range[0]) + display_range[1]) * 100;
+            linear_gradient.appendChild(svgfactory.stop(offset, config.negative_color));
+            linear_gradient.appendChild(svgfactory.stop(offset, config.positive_color));
+        }
+        root.appendChild(linear_gradient);
 	    points.push([5, 20]);
 	    for (var i=0; i<mesh; i++) {
 		var t = i/mesh;
@@ -135,7 +146,7 @@ var OncoprintLegendView = (function() {
 		points.push([5 + 40*i/mesh, 20-height]);
 	    }
 	    points.push([45, 20]);
-	    root.appendChild(svgfactory.path(points, config.color, config.color));
+        root.appendChild(svgfactory.path(points, null, null, linear_gradient));
 	} else if (config.type === 'gradient') {
 	    var num_decimal_digits = 2;
 	    var display_range = config.range.map(function(x) {
--- a/portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
+++ b/portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
@@ -497,12 +497,6 @@ var LinearInterpRuleSet = (function () {
 					range_lower = range[0],
 					range_higher = range[1];
 
-            if (val <= range_lower) {
-                return val < 0 && range_lower < 0 ? -1 : 0;
-            } else if (val >= range_higher) {
-                return val < 0 && range_lower < 0 ? 0 : 1;
-            }
-
                 if (rangeType === rangeTypes.NON_POSITIVE) {
                     // when data only contains non positive values
                     return (val - range_higher) / range_spread;
--- a/portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
+++ b/portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
@@ -472,11 +472,17 @@ var LinearInterpRuleSet = (function () {
 	this.value_key = params.value_key;
 	this.value_range = params.value_range;
 	this.log_scale = params.log_scale; // boolean
-	this.inferred_value_range;
+
+	this.rangeTypes = {
+		'ALL': 'ALL',                   // all values positive, negative and zero
+		'NON_NEGATIVE': 'NON_NEGATIVE', // value range all positive values inclusive zero (0)
+		'NON_POSITIVE': 'NON_POSITIVE'  // value range all negative values inclusive zero (0)
+	};
 
 	this.makeInterpFn = function () {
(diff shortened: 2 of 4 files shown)


pieterlukasse:
are we expecting that many changes to oncoprint-bundle.js? @adamabeshouse I know you like to ignore this file, but I can't recall seeing it with such a big diff.

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintlegendrenderer.js
```diff
@@ -127,6 +127,17 @@ var OncoprintLegendView = (function() {
 	    root.appendChild(svgfactory.text(display_range[1], 50, 0, 12, 'Arial', 'normal'));
 	    var mesh = 100;
 	    var points = [];
+        var linear_gradient = svgfactory.linearGradient();
+        if (config.range_type === 'NON_POSITIVE') {
```
pieterlukasse: maybe call this 'NEGATIVE'?

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
```diff
@@ -472,11 +472,17 @@ var LinearInterpRuleSet = (function () {
 	this.value_key = params.value_key;
 	this.value_range = params.value_range;
 	this.log_scale = params.log_scale; // boolean
-	this.inferred_value_range;
+
+	this.rangeTypes = {
+		'SIGNED': 'SIGNED',
+		'UNSIGNED': 'UNSIGNED',
+		'NON_POSITIVE': 'NON_POSITIVE'
```
pieterlukasse: Can you explain a bit more about the naming choice here? What is the difference between SIGNED and NON_POSITIVE (i.e. negative)? Don't the negative values have a `-` sign? Or does SIGNED mean "mixed" (i.e. mix of positive and negative values)?

adamabeshouse: maybe @rnugraha means that if the range includes 0, its non-positive, as opposed to negative.

in any case I think the naming is a bit confusing, mixing the "signed" language with "non_positive". Maybe just positive, negative, both, with 0's inclusion being implied?

adamabeshouse:
@pieterlukasse that files massive diffs are most likely caused by some re-ordering of files in the bundle which is functionally irrelevant, that's why I say ignore it. All that is is a bundle of the other files - looking at the other files suffices to understand what has changed. My thinking is: f the bundle seems to be working, nothing real must have changed besides the code changes in the other files, so we can trust that anything thats changed is some quirk of the bundler.

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
rnugraha: @pieterlukasse exactly what @adamabeshouse said. I used the term non-positive to implies that it includes 0.

I would also go with Adam's suggestion but I'm afraid it will mislead the inclusion of 0's.

What about:
- NON_POSITIVE = negative values inclusive 0
- NON_NEGATIVE = positive values inclusive 0
- ALL

rnugraha:
@adamabeshouse @pieterlukasse Maybe considering to integrate gulp execution in mvn install? Although the draw back is mvn install will have more dependencies like to install npm & gulp

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
pieterlukasse: @rnugraha ok, the three new names you suggest are better (plus: a small documentation line about the "inclusive 0" part - this would make it clear enough). Thanks.

pieterlukasse:
@rnugraha how long does the gulp step take? I am afraid that if it is >30s, this extra "compilation time" could already be considered "prohibitive" by some developers.

rnugraha:
@pieterlukasse Should be < 30s when one already have it installed. But this will create hard dependency on gulp when you start developing cbioportal (which one might not be interested). I'm suggesting to discuss this as separate topic as future improvement.  Any comment on this @adamabeshouse?

adamabeshouse:
@rnugraha we're actually going to move all development on the oncoprint into the separate oncoprint js repo

portal/src/main/webapp/js/src/oncoprint/webgl/src/js/oncoprintruleset.js
adamabeshouse: personally I think its a bit confusing - you see POSITIVE and you think oh ok POSITIVE, then you need to see NON. its not a big deal but, just a bit of a double take. maybe POSITIVE_OR_ZERO?

pieterlukasse:
@rnugraha I tested it locally and found a number of small issues. See list below:

- [ ] Issue 1. Test with Glioblastoma study, Days to Last follow-up attribute. All values are POSITIVE (according to legend), but still the histogram is split up in half. "NA" values are shown as full bars and values only use the upper half of the space. It should use the full height. Furthermore, the position of the bars is not aligned, i.e. they don't all start on the same line (!?). See screenshots:

image

image

- [ ] issue 2. Test with Glioblastoma, "Birth from Initial patho..." attribute (but also happens for same attribute in Breast cancer study). All negative values and plot seems initially fine, but then there is this 1 sample where the bar is drawn outside its boundaries. See screenshot:

image

- [ ] issue 3. Tested with modified Breast cancer study (see SQL query below to reproduce the data) and "Days to sample collection" attribute. Again the bars are not aligned (see screenshot). Furthermore the scale is not correct: the max positive value (1979) is shown with roughly same size bar as smallest min value (-7858), also in screenshot.

image

image

- [ ] issue 4. Same test as in issue 3. Sorted the track (A-Z). The display shows "gaps" between the lower values. Seems like glitch, perhaps related to the alignment issue reported above. 


image



### Appendix: SQL step to update "Days to sample collection" attribute to contain some negative values as well. Used for issue 3 above. 
```sql
UPDATE cbioportal.clinical_sample
SET ATTR_VALUE = CONCAT('-', ATTR_VALUE)
 where attr_id like '%DAY%%COLL%' 
AND ATTR_VALUE > '2000';
```

rnugraha:
@pieterlukasse Should be fix now. Please test it again.

pieterlukasse:
thanks @rnugraha . Issues 1 and 2 are solved. Issues 3 and 4 remain.

pieterlukasse:
@rnugraha furthermore, the scaling when having positive and negative values is not correct. In the example of issue 3, you will see that the bar for `+1979` is the same size as `-7858`, while it should be in fact ~4 times smaller.
See the legend for example, there the scaling is correct: 
image

pieterlukasse:
@rnugraha all tests are passing, thanks. Will take a look at the code now.

pieterlukasse:
@schultzn @jjgao FYI the colors are changing from "pale brown" to green and red. I hope this is OK? See before/after figures in description of this PR.

schultzn:
Looks great!

jjgao:
@adamabeshouse I am merging this. Please make sure this is captured in the refactoring.

PR #2965, opened 2017-08-11
Fix #2963 Session persistence on app redeploy

### LINKED ISSUE #2963: Add property to allow session persistence on tomcat when redeploying
Tested this on MSKCC beta portal and seems to work.

Change `META-INF/context.xml` to have a `pathname`:
```
<?xml version="1.0" encoding="UTF-8"?>
<Context antiJARLocking="true" path="/cbioportal">
    <Manager pathname="/path/to/SESSIONS.ser"/>
</Context>
```

Now we just want `pathname` to be a property, so we can specify where the sessions should eb stored (default should be `SESSIONS.ser` as per tomcat guidelines).

inodb:
Fix #2963 

This allows to set a location where the session is stored when redeploying an app. By default this is `SESSIONS.ser` in the root of the extracted war. This works fine when restarting tomcat, but is a problem when redeploying the app, since that folder gets overwritten. Simply specifying a folder outside of the app solves the issue. All we have have to do is update properties for each portal with authentication of where on the filesystem the session should be stored. For example for beta portal I specified this:

```
sessions.pathname=/srv/www/msk-tomcat/beta-SESSIONS.ser
```

See also: https://stackoverflow.com/questions/27938643/tomcat-deploy-war-without-losing-session

### DIFF
--- a/portal/src/main/webapp/META-INF/context.xml
+++ b/portal/src/main/webapp/META-INF/context.xml
@@ -1,2 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Context antiJARLocking="true" path="/cbioportal"/>
+<Context antiJARLocking="true" path="/cbioportal">
+    <Manager pathname="${sessions.pathname:SESSIONS.ser}" />
+</Context>


inodb:
Actually I'm not 100% sure if this is actually an issue. I redeployed the mskcc beta portal a few times, but it seems to keep me logged in. @sheridancbio @angelicaochoa @n1zea144 ?

alisman:
@ino, it's possible that with some authentication methods, login status wouldn't be dependent on the jsession, which is what would be refreshed.  But I believe there are other things we put in session which would be.  Like, query or some such thing.  Did you test that?

inodb:
@alisman for query we use session_service which is external: https://github.com/cBioPortal/session-service. So that shouldn't be affected.

I'm not sure what else we would store in the session

PR #2974, opened 2017-08-11
Fix #2778 frontend.url, frontend.url.runtime props

### LINKED ISSUE #2778: Use property/env variable to specify frontend version
Currently we can only change frontend version by changing `portal/pom.xml`. We should change this to an environment variable/property so on heroku we can for instance just change the env variable of the instance instead of having to make changes to the `cbioportal` repo's `pom` file. The latter will then only have to be done at the final stage when we deploy officially.

inodb:
- frontend.url allows us to specify what cbioportal-frontend url should be used. It overrides whatever frontend version is provided in the WAR through jitpack.
- frontend.url.runtime allows us to specify a file that contains the runtime version (eventually this should also support a URL)

Also allow changing `FRONTEND_URL` env variable on heroku

Fix #2778


inodb:
@alisman One advantage of this is that we can for instance point to `cbioportal-frontend.herokuapp.com` and always have latest master branch.

I've got an idea to deal with cache busting as well. We can just add the commit hash in the `dist/` folder of the frontend repo. Then we can use that as the cache buster parameter in the script include tag for the frontend. Either do AJAX call right before including or if it's slow we can use localStorage and store it there, then offer user unobtrusive option for reload.

inodb:
You can btw see this working here: https://cbioportal-pr-2974.herokuapp.com/case.do#/patient?caseId=P04&studyId=lgg_ucsf_2014 (this is using a different version than the one specified in the pom, it is using https://github.com/cBioPortal/cbioportal-frontend/pull/562).

portal/src/main/webapp/js/src/load-frontend.js
```diff
@@ -56,11 +56,21 @@ window.loadReactApp = function(config) {
         });
 
     } else {
+        if (window.frontendUrl === undefined) {
+            if (console.error) {
+                console.error('ERROR: No frontend URL defined, should at least be empty string');
+            } else {
+                console.log('ERROR: No frontend URL defined, should at least be empty string');
+            }
+            // even though this should never happen, probably better to include
```
alisman: maybe we should make it fail loudly though.  so devs will notice.  by falling back, they could think it's working

inodb: @alisman yeah that's maybe better. Should never happen anyway. It will just try to include a script tag with `undefined` plus you'll get the error message above, so that should be p obvious

inodb: updated

PR #2979, opened 2017-08-14
Alter api-legacy model for DBStudy : pmid now String

### LINKED ISSUE #1456: Mismatched PMID field type in study API
[DBStudy.java](https://github.com/cBioPortal/cbioportal/blob/rc/business/src/main/java/org/mskcc/cbio/portal/model/DBStudy.java#L21) expects the PMID retrieved from the database to be parseable as an integral number, but in [the database](https://github.com/cBioPortal/cbioportal/blob/hotfix/core/src/main/resources/db/cgds.sql#L83) it is stored as a text string. This makes the study API throw exceptions if a study has been loaded listing a non-numeric value in its metadata file.

sheridancbio:
# fix to allow non-numeric PMID output
Fix #1456 
Exceptions were observed when a study with a non-numeric PMID value in the database was part of the output of the api endpoint https://cbioportal.mskcc.org/api-legacy/studies.  The non-numeric PMID was accidentally entered, however nothing in the importer disallows the import of a non-numeric value.

- Alter the model of the legacy Api-Controller to output PMID as a string instead of Long

# Checks
- [x] Runs on Heroku.
  the `-s` option)
  hotfix.

# Notify reviewers
@cBioPortal/backend

### DIFF
--- a/business/src/main/java/org/mskcc/cbio/portal/model/DBStudy.java
+++ b/business/src/main/java/org/mskcc/cbio/portal/model/DBStudy.java
@@ -18,7 +18,7 @@ public class DBStudy implements Serializable {
     public String name;
     public String short_name;
     public String description;
-    public Long pmid;
+    public String pmid;
     public String citation;
     public String groups;
     public Integer internal_id;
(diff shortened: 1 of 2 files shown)


sheridancbio:
Code has been build and deployed and API has been tested. I also searched for dependencies on the pmid field within the frontend code and was not able to find any within the cBioPortal repository.

jjgao:
Should we do a hotfix release instead?

sheridancbio:
I don't think we need this to go into hotfix ... we can do normal testing of a beta release, etc.

jjgao:
sounds good.

Issue #601, opened 2015-12-15
Runtime class cast exception in /pancancerMutations.json


morungos:
I've found that in some cases, we get a casting error from an empty set to a list. The Java backtrace is as follows:

    WARNING [org.eclipse.jetty.servlet.ServletHandler] /pancancerMutations.json
    java.lang.ClassCastException: java.util.Collections$EmptySet cannot be cast to java.util.List
	at org.mskcc.cbio.portal.servlet.PancancerMutationsJSON.doGet(PancancerMutationsJSON.java:260)
	at org.mskcc.cbio.portal.servlet.PancancerMutationsJSON.doPost(PancancerMutationsJSON.java:265)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)

The problematic line is https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/servlet/PancancerMutationsJSON.java#L260, and it occurs when the helper method `byProteinPosStarts` returns `Collections.emptySet()`.

I suspect it's because an immutable set cannot be made into a list. 

Suggested change: amend L260 to:

    JSONArray.writeJSONString(new ArrayList<Map<String, Object>>(data), writer);

Issue #600, opened 2015-12-15
cross cancer expression issues and improvements


jjgao:
- [x] legend not match https://github.com/cBioPortal/cbioportal/issues/594
 - [x] legend: change "No mutation" to "Not mutated"
 - [x] change the color of "Not sequenced" dots to gray
 - [x] put box plot in front of dots
 - [x] query of multiple genes does not work
 - [x] when not showing mutation data, do not use different styles for sequenced and unsequenced samples.
 - [ ] load data and render data in "Expression" tab only after it is clicked or directly accessed
 - [x] separate the menu into two lines: first line with gene and download; second line with other items, checkbox should be in front of text.
 - [x] use cbioportal_client.getSampleLists instead of cbioportal_client.getPatientLists

jjgao:
Here is the original pr: https://github.com/cBioPortal/cbioportal/pull/325

PR #3025, opened 2017-08-21
check valid data in download tab for user saved studies

victoria34:
# What? Why?
After users submit query gene for user saved virtual studies, we need to check the length of altered_study_sample_map[studyIds[i]] to avoid an empty array.
Change download data format to "CANCER_STUDY_ID:SAMPLE_ID" per line

Fix [#600](https://github.com/cBioPortal/iViz/issues/600) and [#601](https://github.com/cBioPortal/iViz/issues/601) .

### DIFF
--- a/portal/src/main/webapp/js/src/data_download.js
+++ b/portal/src/main/webapp/js/src/data_download.js
@@ -137,7 +137,7 @@ var DataDownloadTab = (function() {
                     return a.localeCompare(b);
                 });
                 for (var j=0; j<sample_ids.length; j++) {
-                    strs.alt_type += studyIds[i] + "\t";
+                    strs.alt_type += studyIds[i] + ":";
                     strs.alt_type += sample_ids[j] + "\t";
                     var alt_types = sample_to_line_to_alt_type[sample_ids[j]];
                     alt_types && (strs.alt_type += alt_types.join("\t"));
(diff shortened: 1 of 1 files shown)


portal/src/main/webapp/js/src/data_download.js
```diff
@@ -176,10 +176,12 @@ var DataDownloadTab = (function() {
                     return a.localeCompare(b);
                 });
                 for (var j=0; j<sample_ids.length; j++) {
-                    strs.case_matrix += studyIds[i] + "\t";
-                    strs.case_matrix += sample_ids[j] + "\t";
-                    strs.case_matrix += (altered_study_sample_map[studyIds[i]][sample_ids[j]] ? "1" : "0") + "\t";
-                    strs.case_matrix += "\n";
+                    if (_.keys(altered_study_sample_map[studyIds[i]]) > 0) {
```
adamabeshouse: _.keys returns an array, you're looking for _.keys.length

adamabeshouse: it would be better to make a variable to hold `altered_study_sample_map[studyIds[i]]` so we dont keep accessing it, and pull it out of the loop since it doesnt change as `j` changes.

victoria34: Good idea. Thanks!

PR #3033, opened 2017-08-23
Validator: Add allowed timeline header without STOP_DATE

sandertan:
# What? Why?
Fix https://github.com/cBioPortal/datahub/issues/73 .

I also added test data. 

A future enhancement would be to remove the required order of the columns in. I didn't make a unit test, because it would be more time efficient to implement this enhancement, and then make the appropriate tests. 

https://github.com/cBioPortal/cbioportal/blob/c628bfc06e0f7446b4c2a51892f7acc8b6087325/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTimelineData.java#L49


sandertan:
We can do this if it ever becomes a priority.

PR #3037, opened 2017-08-24
Switch to cbioportal repo jitpack - iviz

zhx828:


zhx828:
@adamabeshouse I'm switching to cbioportal repo instead of yours

portal/pom.xml
```diff
@@ -93,7 +93,7 @@
             <configuration>
               <artifactItems>
                 <artifactItem>
-                  <groupId>com.github.adamabeshouse</groupId>
+                  <groupId>com.github.cbioportal</groupId>
                   <artifactId>cbioportal-frontend</artifactId>
                   <version>iviz-SNAPSHOT</version>
```
inodb: @zhx828 @adamabeshouse Just a note here. Using `iviz-SNAPSHOT` is fine for dev btw, but don't get to this to `rc`. Makes it hard to figure out when errors get introduced, cause compilation differs over time. Also note that people have to clear their local mvn cache to actually get latest iviz branch (mvn doesn't know when to update `iviz-SNAPSHOT`)

zhx828: @inodb ah good to know. Thanks! This is for testing only.

PR #3043, opened 2017-08-25
Change Genetic Profile and Genetic Data to Molecular Profile and Mole…

ersinciftci:
…cular Data

No database changes for now, we can change later if we want. No changes in Geneset related
APIs as they are used by the old frontend, The Hyve can change the frontend and
the APIs if they want.


khzhu:
@ersinciftci , do I need to manually merge your commit 67a271c into this PR? Thanks!

PR #3045, opened 2017-08-25
Changes to support multiple study mutations tab changes in frontend

adamabeshouse:
Corresponding to: https://github.com/cBioPortal/cbioportal-frontend/pull/587

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp
@@ -71,11 +71,6 @@
                 }
                 props.samplesSpecification = samplesSpecification;
                 window.renderMutationsTab(mutationsTab[0], props);
-                /*{
-                    genes: QuerySession.getQueryGenes(),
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp
```diff
@@ -42,11 +42,20 @@
 
             if (mutationsTab.hasClass('cbioportal-frontend')) {
                 // backwards compatible with the case when mutationDetailLimitReached
-                window.renderMutationsTab(mutationsTab[0], {
+                var props = {
+                    genes: QuerySession.getQueryGenes()
+                };
+                if (["-1", "all"].indexOf(QuerySession.getCaseSetId()) > -1) {
```
alisman: could you just put some comments in to explain this code.  i would ask in person but better to get this understanding in code.

adamabeshouse: sure. I'll put it in a comment but, "-1" means its a custom case set, "all" means its all cases from all the queried studies - neither are real case lists that you can query

Issue #597, opened 2015-12-15
Samples with mutation data not in sequenced sample list


yichaoS:
window.cbioportal_client.getPatientLists

PR #3071, opened 2017-09-05
Fixed expression tab loading issue

victoria34:
# What? Why?
The expression tab keeps loading after submitting a query. That's because we discard CrossCancerStudyServlet.java, which leads to some jsp and js files are not invoked. So we add related js files in js_include.jsp and move code related to expression tab from cross_cancer_results.jsp to visualize.jsp

Fix [#597](https://github.com/cBioPortal/iViz/issues/597) .


portal/src/main/webapp/WEB-INF/jsp/global/js_include.jsp
```diff
@@ -104,6 +104,11 @@
 <script type="text/javascript" src="js/lib/jstree.js?<%=GlobalProperties.getAppVersion()%>"></script>
 <script type="text/javascript" src="js/api/cbioportal-client.js?<%=GlobalProperties.getAppVersion()%>"></script>
 
+<%--copy these 3 files from cross_cancer_results.jsp for making expression tab works--%>
```
adamabeshouse: this comment will be confusing when cross_cancer_results.jsp doesn't exist anymore.. maybe just say its for the expression tab

PR #3080, opened 2017-09-07
Change virtual "Save" button to "share" button

victoria34:
# What? Why?
Hide the "save" button and add "share" button in study view. Click on share button will display the URL of virtual study.
If users select limited studies, it will show url field and "Copy" button:
screen shot 2017-09-14 at 3 03 06 pm

After click "Copy", it will return successful tip:
image
Fix [#608](https://github.com/cBioPortal/iViz/issues/608) .


portal/src/main/webapp/js/src/dashboard/iviz.js
```diff
@@ -8178,3 +8179,191 @@ window.LogRankTest = (function(jStat) {
...
+                                // Copy virtual study link to clipboard
+                                var $temp = $("<input>");
+                                $("body").append($temp);
+                                $temp.val(tooltip.find('.virtual-study-link').attr('href')).select();
+                                document.execCommand("copy");
+                                // Check if users copy url successfully
+                                if ($temp.val() === tooltip.find('.virtual-study-link').attr('href')) {
+                                    tooltip.find('.shared').css('display', 'block');
+                                    tooltip.find('.dialog').css('display', 'none');
+                                    $temp.remove();
```
adamabeshouse: I think you should remove `$temp` either way, right? (outside the `if` statement)

portal/src/main/webapp/js/src/dashboard/iviz.js
```diff
@@ -8178,3 +8179,191 @@ window.LogRankTest = (function(jStat) {
...
+                                self_.updateStats = true;
+                                
+                                self_.$nextTick(function() {
+                                    var _selectedSamplesNum = 0;
+                                    var _selectedPatientsNum = 0;
+                                    var saveCohort = false;
+                                    
+                                    if (_.isObject(self_.stats.selectedCases)) {
+                                        var different_patients = [];
+                                        var different_samples = [];
```
adamabeshouse: I dont think you need to define these (`different_patients`, `different_samples`) up here, you only use them once

portal/src/main/webapp/js/src/dashboard/iviz.js
```diff
@@ -8178,3 +8179,191 @@ window.LogRankTest = (function(jStat) {
...
+                            tooltip.find('.shared').css('display', 'none');
+                            
+                            var cohortName = $('#study_name').val();
+                            var cohortDescription = $('#study_desc').val();
+                            if (_.isObject(vcSession)) {
+                                self_.updateStats = true;
+                                
+                                self_.$nextTick(function() {
+                                    var _selectedSamplesNum = 0;
+                                    var _selectedPatientsNum = 0;
```
adamabeshouse: move these `_selected..Num` below between 8313 and 8314, where they're used

adamabeshouse:
Also, as we talked about on Slack (just want to write it here for record), please change the way you check if the selection has changed so it doesnt depend on the order of studies in selectedCases

portal/src/main/webapp/js/src/dashboard/iviz.js
victoria34: Yes, I should move it outside the `if` statement.

portal/src/main/webapp/js/src/dashboard/iviz.js
```diff
@@ -8178,3 +8179,199 @@ window.LogRankTest = (function(jStat) {
...
+                                                cohortName, cohortDescription || '')
+                                                .done(function (response) {
+                                                    self_.savedVC = response;
+                                                    tooltip.find('.cohort-link').html(
+                                                    '<a class="virtual-study-link" href="' + window.cbioURL + 
+                                                        'study?id=' + self_.savedVC.id + '" onclick="window.open(\'' + 
+                                                        window.cbioURL + 'study?id=' + self_.savedVC.id + '\')">' + 
+                                                        window.cbioURL + 'study?id=' + self_.savedVC.id + '</a>');
+                                                    _.each(self_.stats.selectedCases, function(study){
+                                                       previousSelectedCases[study.studyID] = study;
```
adamabeshouse: you've confirmed this works? if `selectedCases` changes in place, then the `study` object will not hold onto the previous state, so it will never create a new URL. The way to fix this would be to do a deep copy of `study`

Issue #3106, opened 2017-09-14
broken institute logo


jjgao:
The right logo is broken. Is it related to #2945?

public portal (should be no logo):
![image](https://user-images.githubusercontent.com/840895/30409262-fc2ca056-98d1-11e7-8ad6-0fc5f55c9dbf.png)

msk portal (msk logo):
![image](https://user-images.githubusercontent.com/840895/30409275-0f5e5c0a-98d2-11e7-99a0-040a46cc9466.png)

fedde-s:
Yes, it might be related. @jjgao, may I ask which release of Tomcat these installations use?

fedde-s:
I see, I can reproduce it by reverting to Tomcat 7.0.x.

fedde-s:
I believe that this is solved now that my patch has been merged into the hotfix development branch.

![Screenshot of the same front page region as the first one in this issue, but without the literal words ‘Institute Logo’](https://user-images.githubusercontent.com/4929431/30473940-4f50d9ec-9a02-11e7-9cd8-bdb3a2be67a5.png)

fedde-s:
As [part of the discussion](https://github.com/cBioPortal/cbioportal/pull/3107#issuecomment-329525587) about the pull request I explained the reasoning behind my choice to use JSTL tags over scriptlets.

PR #3107, opened 2017-09-14
Fix broken institute logo in older Tomcat versions

fedde-s:
The JSP template for the header bar used an EL expression which called a
static Java method. This turned out not to work in Tomcat 7, causing the
template to insert a broken  tag no matter whether a logo was
enabled or not. The native JSP way to solve this would be to define the
static method as a function in a taglib, but as this is just a single
function call I decided to keep it short and evaluate a Spring SpEL
expression instead.

This should fix issue #3106.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/global/header_bar.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/global/header_bar.jsp
@@ -32,6 +32,7 @@
 
 <%@ page import="org.mskcc.cbio.portal.util.GlobalProperties" %>
 <%@ taglib prefix='c' uri='http://java.sun.com/jsp/jstl/core' %>
+<%@ taglib prefix="s" uri="http://www.springframework.org/tags" %>
 <%@ taglib prefix="sec" uri="http://www.springframework.org/security/tags" %>
 <%
     String principal = "";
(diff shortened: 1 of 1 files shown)


fedde-s:
@jjgao, this should fix the issue you reported; I tested it in on both Tomcat 7.0.x and Tomcat 8.5.x with Java 8, with and without a logo, and it behaved as I expected in all four cases; no more grey letters spelling out the words ‘Institute Logo’.

fedde-s:
JSTL tags have superseded scriptlets in JSP back in 2003, in part because scriptlets encourage mixing arbitrary business logic into the templating layer, but also because a declarative domain-specific language could actually be a lot more easy to read when interleaved with HTML than imperative Java code.

The line I insert here certainly is less readable though, as I'm using yet another language included in Spring to avoid having to define a JSP tag library just for this one fix. I did accompany it by a comment, as a suggestion that if anyone considers writing more code like it in the future, they might want to use a taglib; I'm not sure how much of the JSP layer will remain as we move on to the new external frontend.

fedde-s:
As [part of the discussion](https://github.com/cBioPortal/cbioportal/pull/3107#issuecomment-329525587) about the pull request I explained the reasoning behind my choice to use JSTL tags over scriptlets.

PR #3155, opened 2017-09-21
Make Oncoprint heatmap tracks sortable

fedde-s:
Users have requested the option to sorting the Oncoprint columns by specific heatmap tracks similar to how one sorts clinical attribute values. After enabling this functionality, the remaining commits hopefully avoid confusion and make using this feature more smooth.

## Pre-review Screenshot ##
Screenshot of an Oncoprint in which a clinical track and a heatmap track are sorted, marked by black triangles next to their track options buttons

## Updated screenshot ##
Similar screenshot as above, except marking the sorted tracks by rows of bars of ascending and descending height


fedde-s:
@adamabeshouse, could you review this?

fedde-s:
Right, the triangles may be especially puzzling when an Oncoprint includes protein expression alterations, also visualised as triangles. Although to be consistent with sorting icons in other software I've seen, I think I still prefer (wide) vertical arrows. Does this look any better?

Similar screenshot as before but using heavy black vertical arrows

Or is this closer to what you had in mind?

Similar screenshot as before but using little heavy eastward-diagonal arrows

adamabeshouse:
personally I prefer the to-the-right arrows, for some reason it just hits me more quickly that way. @jjgao thoughts?

jjgao:
How about something like this (http://fontawesome.io/icon/sort-amount-asc/) -- transposed it and maybe w/o the arrow?

fedde-s:
Something like <http://fontawesome.io/icon/signal/> then, except with a linear increase and a mirrored version. Sounds good!

fedde-s:
I updated the code to use the icon I mentioned in [my last comment](https://github.com/cBioPortal/cbioportal/pull/3155#issuecomment-331830461), using CSS to mirror it for tracks sorted in descending order—see the updated screenshot in [the PR summary](https://github.com/cBioPortal/cbioportal/pull/3155#issue-259541895). I had to change one implementation detail in aligning it with the menu button because Firefox handled inline-block elements inconsistently with other browsers.

jjgao:
@adamabeshouse I assume it won't take too much time to move the code to oncoprint.js?

PR #3198, opened 2017-09-28
Download Oncoprint in tabular format

adamabeshouse:
This PR adds an extra option to the "Download" menu of the Oncoprint which allows to download the Oncoprint in tabular format.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
@@ -155,11 +155,12 @@
                <button type="button" class="btn btn-default dropdown-toggle" id="oncoprint_diagram_mutation_color_dropdown" data-toggle="dropdown" style="background-color:#efefef;margin:0px">
                  <span>Download</span>&nbsp;<span class="caret"></span>
                </button>
-               <div class="dropdown-menu" style="padding: 10px 5px; width: 70px;min-width: 70px;">
-                    <button class='oncoprint-diagram-download' type='pdf' style='font-size:13px; cursor:pointer;width:50px;'>PDF</button> <br/>
-                    <button class='oncoprint-diagram-download' type='png' style='font-size:13px; cursor:pointer;width:50px;'>PNG</button> <br/>
-                    <button class='oncoprint-diagram-download' type='svg' style='font-size:13px; cursor:pointer;width:50px;'>SVG</button> <br/>
-                    <button class='oncoprint-sample-download'  type='txt' style='font-size:13px; cursor:pointer;width:50px;'>Sample order</button>
+               <div class="dropdown-menu" style="padding: 10px 5px; width: 120px;min-width: 70px;">
+                    <button class='oncoprint-diagram-download' type='pdf' style='font-size:13px; cursor:pointer;width:100px;'>PDF</button> <br/>
+                    <button class='oncoprint-diagram-download' type='png' style='font-size:13px; cursor:pointer;width:100px;'>PNG</button> <br/>
+                    <button class='oncoprint-diagram-download' type='svg' style='font-size:13px; cursor:pointer;width:100px;'>SVG</button> <br/>
+                    <button class='oncoprint-sample-download'  type='txt' style='font-size:13px; cursor:pointer;width:100px;'>Sample order</button>
+                    <button class='oncoprint-tabular-download'  type='tsv' style='font-size:13px; cursor:pointer;width:100px;'>Tab-separated values</button>
                </div>
             </div>
             
--- a/portal/src/main/webapp/js/src/oncoprint/setup.js
+++ b/portal/src/main/webapp/js/src/oncoprint/setup.js
@@ -2516,9 +2516,188 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 		    cbio.download.initDownload(content, downloadOpts);
 		});
 	    });
+	    
+	    $('body').on('click', '.oncoprint-tabular-download', function () {
+		//Get caseNames (either patients or samples)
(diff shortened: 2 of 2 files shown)


portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -2499,9 +2499,153 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
 		    cbio.download.initDownload(content, downloadOpts);
 		});
 	    });
+	    
+	    $('body').on('click', '.oncoprint-tabular-download', function () {
+		//Get caseNames (either patients or samples)
+		window.QuerySession.getUIDToCaseMap().then(function (uid_to_case) {
+		    var caseNames = [];
+		    oncoprint.getIdOrder().map(function (uid) {
+			caseNames.push(uid_to_case[uid]);
```
adamabeshouse: just want to confirm `map` always goes in order

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -2499,9 +2499,153 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
...
+		    };
+		    var fusionMap = {
+			    'true': 'Fusion'
+		    };
+
+		    //Add genetic data
+		    var track_ids = utils.objectValues(State.genetic_alteration_tracks);
+		    patientOrSample = (State.using_sample_data ? "sample" : "patient");
+		    for (var track_id=0; track_id<track_ids.length; track_id++){
+			var currentTrackData = oncoprint.getTrackData(track_ids[track_id]);
```
adamabeshouse: a more principled way to do both of these is to use QuerySession.get(Sample/Patient)GenomicEventData - the oncoprint_data attribute of those elements is the track data, and gene is on there.

portal/src/main/webapp/js/src/oncoprint/setup.js
```diff
@@ -2499,9 +2499,153 @@ window.CreateCBioPortalOncoprintWithToolbar = function (ctr_selector, toolbar_se
...
+		    var content = 'track_name\ttrack_type';
+		    //Add the cases to the content
+		    for (var i=0; i<caseNames.length; i++) {
+			content += '\t'+caseNames[i];
+		    }
+		    //Add final header line
+		    content += '\n';
+		    
+		    //Iterate over oncoprintData and write it to content
+		    for (var j in oncoprintData) {
```
adamabeshouse: I dont think this type of iteration is stable with respect to order, which would be a problem when generating the table?

portal/src/main/webapp/js/src/oncoprint/setup.js
oplantalech: I've tried to do it with `QuerySession.get(Sample/Patient)GenomicEventData`, but this data do not reflect if the mutations are putative drivers or passengers. `oncoprint.getTrackData` does reflect this.

portal/src/main/webapp/js/src/oncoprint/setup.js
oplantalech: > `map` calls a provided `callback` function **once for each element** in an array, in order, and constructs a new array from the results.

From: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

portal/src/main/webapp/js/src/oncoprint/setup.js
adamabeshouse: if you call it with the argument `true` it will reflect that

`QuerySession.get(..)GenomicEventData(true)`

adamabeshouse:
I'm sorry I didn't think about this sooner but, I think your original approach would make sense, except that we should move it INTO the oncoprint as a new functionality of the library - export to tabular format. Then the internal references would be kosher. And it would make sense to be getting the data internally from the tracks and such.

oplantalech:
@adamabeshouse I've moved the code into a function.

adamabeshouse:
sorry to run you around..

so, I think if we want to move it into the oncoprint library, we need to make it a lot more general, and move all of the logic and assocation with genetic concepts into callbacks and track parameters. So I think there are two options
(1) Spend more time doing this the right way in oncoprintjs
(2) Use your previous commit where you do it outside the oncoprint.

oplantalech:
@adamabeshouse Can you take a look at the changes I made in the code? Thank you.

portal/src/main/webapp/WEB-INF/jsp/oncoprint/main.jsp
```diff
@@ -155,11 +155,12 @@
...
-                    <button class='oncoprint-diagram-download' type='svg' style='font-size:13px; cursor:pointer;width:50px;'>SVG</button> <br/>
-                    <button class='oncoprint-sample-download'  type='txt' style='font-size:13px; cursor:pointer;width:50px;'>Sample order</button>
+               <div class="dropdown-menu" style="padding: 10px 5px; width: 120px;min-width: 70px;">
+                    <button class='oncoprint-diagram-download' type='pdf' style='font-size:13px; cursor:pointer;width:100px;'>PDF</button> <br/>
+                    <button class='oncoprint-diagram-download' type='png' style='font-size:13px; cursor:pointer;width:100px;'>PNG</button> <br/>
+                    <button class='oncoprint-diagram-download' type='svg' style='font-size:13px; cursor:pointer;width:100px;'>SVG</button> <br/>
+                    <button class='oncoprint-sample-download'  type='txt' style='font-size:13px; cursor:pointer;width:100px;'>Sample order</button>
+                    <button class='oncoprint-tabular-download'  type='tsv' style='font-size:13px; cursor:pointer;width:100px;'>Tab-separated values</button>
```
adamabeshouse: @jjgao is `Tab-separated values` a good name for this?

PR #3221, opened 2017-10-02
Add genetic profiles to list and fix DaoMutation calls

adamabeshouse:
Fixes bug introduced by https://github.com/cBioPortal/cbioportal/pull/3207/files

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -1487,9 +1487,9 @@ public final class DaoMutation {
         List<String> tiers = new ArrayList<String>();
         ArrayList<GeneticProfile> geneticProfiles = new ArrayList<>();
 	for (String cancerStudyStableId: cancerStudyStableIds) {
-		DaoGeneticProfile.getAllGeneticProfiles(
+		geneticProfiles.addAll(DaoGeneticProfile.getAllGeneticProfiles(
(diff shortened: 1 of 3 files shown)


adamabeshouse:
@oplantalech hm how did you test it? because from where I'm looking, the argument to hasDriverAnnotations and numTiers will be a comma separated list of studies, and the oncoprint is not crashing. 

Try running it after typing this in the console: ```localStorage.heroku = "cbioportal-frontend-pr-666"```

oplantalech:
@adamabeshouse That's what I see:
screenshot

When I debug, I see that the `hasDriverAnnotations` is called twice, the first time with a list of the queried cancer ids, and the second one with a list containing `all`, which is causing the error.

adamabeshouse:
okay thanks @oplantalech, when you get a chance could you please try it again with the latest commit, still with the heroku? I was able to reproduce the crash and it seems to me I have fixed it

inodb:
Note that you might have to do hard refresh to get around caching of heroku (not sure how long frontend artifacts are stored)

adamabeshouse:
that shouldnt be a problem since these changes are actually in the backend

oplantalech:
@adamabeshouse I tried it once again and now the `hasDriverAnnotations` is not called with `all` as parameter, but with `""` (apart from the iteration with the correct list of cancer IDs). It might also be that we're just working with different Tomcat versions. I'm running Tomcat  8.5.16.

adamabeshouse:
thanks to @inodb for setting up test database to test this

PR #3248, opened 2017-10-11
Query Page: configurable list of sets of genes

oplantalech:
# What? Why?
We have created a property in portal.properties (`querypage.setsofgenes.location`) where the user can specify a different file that describes the list of sets of genes displayed in the query page. This file must be a JSON string.

This PR goes together with the PR cbioportal/cbioportal-frontend#689.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -37,9 +37,11 @@ import org.mskcc.cbio.portal.servlet.QueryBuilder;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.io.ClassPathResource;
 import org.springframework.security.core.Authentication;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -1054,4 +1058,21 @@ public static boolean hidePassengerMutations() {
...
+    public static String getQuerySetsOfGenes() {
+	String result = new String();
+	try {
+	    BufferedReader br = new BufferedReader(new FileReader(ResourceUtils.getFile(properties.getProperty(SETSOFGENES_LOCATION, "classpath:/setsofgenes.json"))));
+	    for (String line = br.readLine();line != null;line = br.readLine()) {
+		line = line.trim();
+		result = result + line;
+	    }
+	    br.close();
+	} catch (FileNotFoundException e) {
```
pieterlukasse: @oplantalech is good to have a comment here explaining why we catch and re-throw as RuntimeException: because of the way the getQuerySetsOfGenes method is used in .jsp it needs to be static and plain (no "throws" in signature).

PR #3253, opened 2017-10-16
Clarification in documentation stable_id continuous RNA-seq

dionnezaal:
Added clarification in documentation about stable_id of continuous RNA-seq data after question in google groups (https://groups.google.com/forum/#!topic/cbioportal/LPh5zBkdkHc).

### DIFF
--- a/docs/File-Formats.md
+++ b/docs/File-Formats.md
@@ -414,6 +414,7 @@ The expression metadata file should contain the following fields:
 
 #### Supported stable_id values for MRNA_EXPRESSION
 For historical reasons, cBioPortal expects the `stable_id` to be one of those listed in the following static set.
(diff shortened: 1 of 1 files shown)


docs/File-Formats.md
```diff
@@ -414,6 +414,7 @@ The expression metadata file should contain the following fields:
 
 #### Supported stable_id values for MRNA_EXPRESSION
 For historical reasons, cBioPortal expects the `stable_id` to be one of those listed in the following static set.
+The stable_id for continuous RNA-seq data has two options. These options were added to distinguish between two different TCGA pipelines. For custom datasets either one of these stable_id can be chosen.
```
pieterlukasse: I would add something like below, which contains a bit more background info: 

The two possible stable ids  `rna_seq_mrna` and `rna_seq_v2_mrna` relate to the two types of normalization, see: https://wiki.nci.nih.gov/display/tcga/rnaseq+version+2. However, for custom datasets either one of these stable_id can be chosen.

Issue #3144, opened 2017-09-20
Study view hangs for meso_tcga study on public portal


pieterlukasse:
![image](https://user-images.githubusercontent.com/2900303/30637621-230ea81e-9df9-11e7-9348-472754ab0120.png)

Steps to reproduce: 
- go to: http://www.cbioportal.org/study?id=meso_tcga#summary

pieterlukasse:
probably related to #3145

PR #3254, opened 2017-10-16
Handle data issue more gracefully in Study View

zhx828:
When there is a data issue, API issue, Study View needs to respond accordingly instead of hanging. In this pull request, all API calls are attached with a fail function which contains a response from the error.

JS code has been reviewed in there: https://github.com/cBioPortal/iViz/pull/626
But the jsp files need to be reviewed.
 
Related to #3144 

The following message will be given:
screen shot 2017-10-16 at 5 40 42 pm

The available messages are:
- Failed to load sample list data from study.
- Failed to load sample list from study.
- Failed to load patient clinical data.
- Failed to load sample clinical data.

Study Views needs above data to generate charts. For other optional API calls, Study View will silent pass the error and give a warning messages(In the next implementation).

# Checks
- [ ] Runs on Heroku.
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases, multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
  the `-s` option)
  hotfix.

# Notify reviewers
@jjgao @victoria34 @pieterlukasse @cBioPortal/frontend


portal/src/main/webapp/WEB-INF/jsp/dashboard/dashboard.jsp
```diff
@@ -285,22 +285,35 @@
...
-            }else {
-                $.when(window.iviz.datamanager.getConfigs()).then(function(configs){
+                        if (_.isObject(configs)) {
+                            opts = configs;
+                        }
+                        if (emailContact_) {
+                            opts.emailContact = emailContact_;
+                        }
+                        $.when(
+                            window.iviz.datamanager.initialSetup()
```
pieterlukasse: this call used to be in parallel with `window.iviz.datamanager.getConfigs()` . Is it not possible to keep running both in parallel and go to "fail" is either one of them fails?

zhx828: The fail responses are different now. And since we are reading from a JSON file for the configs, so the performance wouldn't be affected.

jjgao:
What happened to the screenshots?

jjgao:
and it's not working... https://cbioportal-pr-3254.herokuapp.com/study?id=acc_tcga#summary

image

zhx828:
@jjgao it is fixed now. It was caused due to two the second commit.

PR #3282, opened 2017-10-24
Add an API enpoint to shorten URLs using bit.ly

ersinciftci:
http://cbioportal.org/api/url-shortener?url=http://google.com

This PR adds a new property in portal.properties called `bitly.access.token`. No need to use the old `bitly.url`, `bitly.user` and `bitly.apiKey` with this new API.

### DIFF
--- a/web/src/main/resources/applicationContext-url-shortener.xml
+++ b/web/src/main/resources/applicationContext-url-shortener.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:mvc="http://www.springframework.org/schema/mvc"
+       xmlns:context="http://www.springframework.org/schema/context"
+       xsi:schemaLocation="
+     http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd
+     http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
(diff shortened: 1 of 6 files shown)


jjgao:
Thanks, @ersinciftci! Do you think it makes sense to organize these kind of servlets into a private utils location, e.g. /utils/url-shortener or even in a util api.

ersinciftci:
@jjgao I don't prefer extra paths like `/utils/` that won't have any content when visited directly, like `cbioportal.org/utils/`. And I prefer not having any Swagger documentation, even internal, for these kind of services that we never would want 3rd parties to use.

jjgao:
Good point on no swagger doc. 

I still think it’s a good idea to organize this type of endpoint under s specific path. I think it’s ok that `cbioportal.org/utils/` is not visitable.

web/src/main/resources/applicationContext-url-shortener.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
inodb: @ersinciftci So does this apply to all properties now? That they can be overridden by system props? I guess except for anything defined in `GlobalProperties`

ersinciftci: @inodb it was always like this for all properties, that's how you could use `-Ddb.host` kind of things.

inodb: @ersinciftci well `-Ddb.host` actually still doesn't work, or at least some of the `db.` parameters don't update when passed to jar/war through `-D`. Not sure why, maybe cause in some places we use `GlobalProperties` instead of the spring prop directly through annotation.

Why do we need a separate `applicationContext` for the shortener?

Seems like only this line `<context:component-scan base-package="org.cbioportal.url_shortener" />` is new compared to applicationContext.xml

PR #3287, opened 2017-10-25
Pass frontend config via window.frontendConfig

inodb:
Pass frontend config in one object, instead of separate ones. Also allow user
to give path to frontendConfig file (property called frontend.config).


inodb:
blocked by: https://github.com/cBioPortal/cbioportal-frontend/pull/715

inodb:
@onursumer fixed

PR #3311, opened 2017-11-02
Fix passing frontendConfig

inodb:
- historyType wasn't passed correctly
- also maxTreeDepth should be passed as number

Uses: https://github.com/cBioPortal/cbioportal-frontend/pull/723


alisman:
Not critical but are any of the props in myindex.ejs file (template for html file that we sometimes use in dev) involved (and going to break)?

inodb:
@alisman I updated those, but yeah i should double check nothing broke

PR #3317, opened 2017-11-03
Add logging to all new API methods in all layers using AspectJ

ersinciftci:


inodb:
@ersinciftci Thanks! Seems to have some failed core tests?

ersinciftci:
@inodb fixed, thanks.

PR #3318, opened 2017-11-06
Add check in GlobalProperties for showing twitter in bookmark buttons.

Tamba32:
Connected with cbioportal/cbioportal-frontend PR: https://github.com/cBioPortal/cbioportal-frontend#727

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -140,6 +140,9 @@ public class GlobalProperties {
     public static final String ONCOKB_PUBLIC_API_URL = "oncokb.public_api.url";
     public static final String SHOW_ONCOKB = "show.oncokb";
 
+    @Value("${bookmark.twitter_link.enabled:}")
+    public static final String SHOW_TWITTER_BOOKMARK = "bookmark.twitter_link.enabled";
+
     private static String sessionServiceURL;
     @Value("${session.service.url:}") // default is empty string
     public void setSessionServiceURL(String property) { sessionServiceURL = property; }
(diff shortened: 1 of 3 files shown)


portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
```diff
@@ -13,6 +13,7 @@ window.legacySupportFrontendConfig = {
     genomeNexusApiUrl : '<%=GlobalProperties.getGenomeNexusApiUrl()%>',
     showCivic : <%=GlobalProperties.showCivic()%>,
     showHotspot : <%=GlobalProperties.showHotspot()%>,
+    showTwitter: '<%=GlobalProperties.showTwitterBookmark()%>',
```
inodb: @Tamba32 @alisman remove quotes there to pass as boolean instead of string

Tamba32: @inodb @alisman thanks!

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -140,6 +140,8 @@
     public static final String ONCOKB_PUBLIC_API_URL = "oncokb.public_api.url";
     public static final String SHOW_ONCOKB = "show.oncokb";
 
+    public static final String SHOW_TWITTER_BOOKMARK = "bookmark.twitterLink.enabled";
```
inodb: @Tamba32 Can you either
-  add the `@Value{$bookmark.twitter_link.enabled}` decorator? That way we can change the value on runtime with `-d bookmark.twitter_link.enabled=true` or `-d bookmark.twitter_link.enabled=false`. -- Or -
- another way might be to use  bookmark.twitter_link.enabled directly in the JSP w/o using GlobalProperties class (@ersinciftci might know).

Also for consistency you prolly don't want to use camelCase on the properties, they all use underscores and dots.

jjgao:
Should this go to master branch?

PR #3344, opened 2017-11-10
Fix #3330 Make query button work for study view

### LINKED ISSUE #3330: Study is not selected after clicking query button from study view page
This is tested from iviz branch.

nov-08-2017 16-44-02

inodb:
This fixes the query button on study view for multiple studies. Also fixes
issues with passing of case lists and studies to the backend. We now no longer
create a virtual cohort on the frontend but instead just pass the relevant
studyIds and caseLists. Works both for when the session service is enabled or
disabled.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/dashboard/dashboard.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/dashboard/dashboard.jsp
@@ -336,9 +336,6 @@
                             });
                         }
                     }
-                    $("#submit_button").click(function(){
-                        iViz.submitForm(true);
-                    });
                 } else if (cohortIdsList.length >= 2) {
                     var study_name = 'Combined Studies';
                     var study_description = 'Total ' + cohortIdsList.length + ' studies.';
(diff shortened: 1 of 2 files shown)


kalletlak:
@inodb are the changes to `iviz.js` from iviz repository(https://github.com/cBioPortal/iViz) or done locally?

inodb:
@kalletlak locally I see, I guess I should update iViz first? Could you maybe give me some pointers on slack of how to do this

jjgao:
@kalletlak could you review the code first and see if there are any issues? We hope to merge soon so that @zhx828 can present the feature next Monday :)

At the same time, @inodb can probably move the code back to iviz.

@zhx828 we should set up something similar to cbioportal-frontend repo so that we don't have to copy source file. Maybe @inodb can help with that?

portal/src/main/webapp/WEB-INF/jsp/dashboard/dashboard.jsp
```diff
@@ -361,6 +358,10 @@
                             .prev().toggle();                    // Hide/show the text
                     });
                 }
+                $("#submit_button").click(function(){
+                    debugger;
```
inodb: @kalletlak we need to remove this when updating iviz

PR #3354, opened 2017-11-13
Web service for clinical attribute count for all samples from study, list of sample IDs or sample List ID

dionnezaal:
# What?
The web service counts the number of samples which have a value for the clinical attributes in a study. All samples from a study, a list of sample IDs or a sample list ID can be used for the clinical attribute count. 

The web service was added as internal API: GET /studies/{studyId}/clinical-attributes-with-stats.


dionnezaal:
@pieterlukasse ready for review

jjgao:
Thanks for the PR. Is it also possible to support multiple studies, and even virtusl studies (a subset of samples in multiple studies)?

jjgao:
... and I assume this API is to support dropdown box to select a clinical attribute?

pieterlukasse:
@jjgao correct, this will be for selecting clinical tracks in oncoprint, and maybe later also useful for the clinical drop-down in study view. 
Good point about the multiple studies. We can add that. This means the API will change to API: GET `/studies/clinical-attributes-with-stats` and a `studyIds` list becomes part of the POST request parameters.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalAttributeWithStatsMapper.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
+<mapper namespace="org.cbioportal.persistence.mybatis.ClinicalAttributeWithStatsMapper">
+    <cache/>
+    
+    <select id="getAllClinicalAttributesWithStatsNoSampleId" resultType="org.cbioportal.model.ClinicalAttributeWithStats">
+        SELECT clinical_attribute_meta.ATTR_ID AS attrId, cancer_study.CANCER_STUDY_IDENTIFIER AS cancerStudyIdentifier, clinical_attribute_meta.DISPLAY_NAME AS displayName, clinical_attribute_meta.DESCRIPTION AS description, clinical_attribute_meta.DATATYPE AS datatype, clinical_attribute_meta.PATIENT_ATTRIBUTE AS patientAttribute, clinical_attribute_meta.PRIORITY AS priority, clinical_attribute_meta.CANCER_STUDY_ID AS cancerStudyId, COUNT(DISTINCT(patient.INTERNAL_ID)) AS attrCount
```
pieterlukasse: @dionnezaal as discussed, please remove the SQL code duplication.

dionnezaal:
@jjgao @pieterlukasse I have added support for multiple studies to the API, therefore the API is now: `GET /studies/clinical-attributes-with-stats`. SampleIDs from studies can also be entered and are then selected for the clinical attribute count. 

Only one Sample List ID can be provided, please let me know if this also needs to be adapted to accept more than one Sample List ID.

jjgao:
Thanks. It'd be good to have @ersinciftci to review this PR too.

ersinciftci:
Why do we need the counts for the dropdown box? We already have a Clinical Attributes API, can’t we add a new POST endpoint to that controller to support sample IDs/sample list ID and multiple studies?

jjgao:
@ersinciftci we would like to order attributes based on data availability. It is important for studies with many attributes (oftentimes most attributes have data in a subset of samples).

ersinciftci:
@jjgao I see. @dionnezaal you can add a POST endpoint to the already existing controller, and make the ‘count’ part of DETAILED projection.

dionnezaal:
@ersinciftci @jjgao @pieterlukasse  Have integrated the web service in the current clinical attribute API. Documentation of the added POST endpoint: 

`POST /studies/clinical-attributes-with-stats`

_Required_
`studyIds`: List of study IDs

_Optional_
`sampleIds`: List of sample IDs
`sampleListId`: Sample list ID


getAllClinicalAttributesInStudyWithStats

```
{
   "attrCount": 0,
   "clinicalAttributeId": "string",
   "datatype": "string",
   "description": "string",
   "displayName": "string",
   "patientAttribute": true,
   "priority": "string",
   "studyId": "string"
}
```

dionnezaal:
- Changed name of API POST endpoint to `/clinical-attributes/fetch`
- Added attribute count variable (`attrCount`) to `ClinicalAttribute` model and removed `ClinicalAttributeWithStats` model
- Clinical attribute count is only shown when DETAILED projection is chosen by user
- Replaced input of `studyId` and `sampleId` with a request body `sampleIdentifiers` which requires a pair of these two variables. Modified code and SQL query to accept the `sampleIdentifiers` and `sampleListId` without study specified.
- Added sample with one clinical attribute to the test database to be able to perform unit test to check the clinical attribute count when a` sampleId` is present in multiple studies. 
- Added unit tests for exceptions, check for detailed projection and to check the clinical attribute count calculated. 

Documentation of current API POST endpoint:
`POST /clinical-attributes/fetch`

sampleIdentifiers: `
[
  {
    "sampleId": "string",
    "studyId": "string"
  }
]
`
sampleListId: String

Model:
```
{
    "attrCount": 0,
    "clinicalAttributeId": "string",
    "datatype": "string",
    "description": "string",
    "displayName": "string",
    "patientAttribute": true,
    "priority": "string",
    "studyId": "string"
}
```

web/src/main/java/org/cbioportal/web/ClinicalAttributeController.java
```diff
@@ -114,4 +121,39 @@
...
     }
+    
+    @RequestMapping(value = "/clinical-attributes/fetch", method = RequestMethod.POST,
+                    produces = MediaType.APPLICATION_JSON_VALUE)
+    @ApiOperation("Get all clinical attributes in specified sampleIdentifiers or sampleListID with clinical attribute count")
+    public ResponseEntity<List<ClinicalAttribute>> getAllClinicalAttributesInStudies(
+    			 @ApiParam(required = false, value = "List of sample identifiers")
+             @RequestBody List<SampleIdentifier> sampleIdentifiers,
+             @ApiParam("Sample list ID e.g. acc_tcga_all")
+             @RequestParam(required = false) String sampleListId,
```
ersinciftci: As I said earlier, you need to create a filter like https://github.com/cBioPortal/cbioportal/blob/rc/web/src/main/java/org/cbioportal/web/parameter/DiscreteCopyNumberFilter.java and put sampleIdentifiers, and sampleListId there, and make sure only one of them is filled using `@AssertTrue`. And also check `@Size(min, max)` just like in the example.

web/src/main/java/org/cbioportal/web/ClinicalAttributeController.java
```diff
@@ -114,4 +121,39 @@
...
+		    	else if (sampleIdentifiers != null && sampleIdentifiers.size() > 0) {
+    				List<String> studyIds = new ArrayList<>();
+	    			List<String> sampleIds = new ArrayList<>();
+    				for (SampleIdentifier sampleIdentifier : sampleIdentifiers) {
+    		            studyIds.add(sampleIdentifier.getStudyId());
+    		            sampleIds.add(sampleIdentifier.getSampleId());
+    		        }
+    				return new ResponseEntity<>(clinicalAttributeService.getAllClinicalAttributesInStudies(studyIds, sampleIds, projection.name(), sortBy == null ? null : sortBy.getOriginalValue(), direction.name()), HttpStatus.OK);
+
+    			} else {
```
ersinciftci: You won't need this when you create the request body filter object I mentioned above.

web/src/main/java/org/cbioportal/web/parameter/sort/ClinicalAttributeSortBy.java
```diff
@@ -8,7 +8,8 @@
     datatype("datatype"),
     patientAttribute("patientAttribute"),
     priority("priority"),
-    studyId("cancerStudyIdentifier");
+    studyId("cancerStudyIdentifier"),
+	clinicalAttributeCount("attrCount");
```
ersinciftci: Indentation, and the name `clinicalAttributeCount` doesn't match with the property name. These two are supposed to be different only if the alias in the SQL query is different than what we return to the user.

pieterlukasse:
@dionnezaal please add some performance test results. The service should not take > 1s for a normal study (e.g. brca_tcga). Should also not take much more time when you combine brca with msk_impact for example (2s max).

dionnezaal:
@pieterlukasse Performance testing results on current SQL query: 

_MSK impact (10946 samples, 17 Clinical Attributes)_
Using sampleIdentifiers: 1.545 sec
Using sampleListId: 1.397 sec

_TCGA provisional brca_tcga (1105 samples, 137 Clinical Attributes)_
Using sampleIdentifiers: 2.7 sec
Using sampleListId: 2.8 sec

ersinciftci:
@pieterlukasse Looks good but I just have a question, why do we union sample and patient attributes? Are there some attribute ids that are both sample and patient attributes?

pieterlukasse:
@ersinciftci the union is needed indeed because we want all clinical attributes in the list. Ideally attributes are only sample attributes or only patient attributes...although I think this constraint has been removed after the clinical data model changed to a model that enforces less harmonization between studies (which is a bit ironic now that we want to combine studies). 

So to answer the question if there are some attributes that will be both sample and patient attributes: yes if you combine studies, (probably) no if you don't combine them. Anyway, I don't think this is an issue that should be taken into account in this web-service. The web-service will simply display an attribute twice, which will make the underlying data problem more visible. Perhaps the web-service can have an extra attribute telling whether it is a patient or a sample attribute (if not present yet).

jjgao:
@dionnezaal could you rebase?

dionnezaal:
@jjgao rebased on rc.

PR #3370, opened 2017-11-14
workaround for missing cytoscape.js

### LINKED ISSUE #3349: network view comes up empty
An "Uncaught TypeError" is seen in the console.

onursumer:
# What? Why?
Fix #3349.

# Checks
- [ ] Runs on Heroku.
  the `-s` option)
  hotfix.

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/global/js_include.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/global/js_include.jsp
@@ -109,6 +109,9 @@
 <script type="text/javascript" src="js/src/plots-tab/util/stylesheet.js"></script>
 <script type="text/javascript" src="js/src/plots-tab/util/plotsUtil.js"></script>
 
+<%-- TODO: workaround: cytoscape.js needs to be included before virtual cohort session libraries --%>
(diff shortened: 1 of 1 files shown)


portal/src/main/webapp/WEB-INF/jsp/global/js_include.jsp
```diff
@@ -109,6 +109,9 @@
 <script type="text/javascript" src="js/src/plots-tab/util/stylesheet.js"></script>
 <script type="text/javascript" src="js/src/plots-tab/util/plotsUtil.js"></script>
 
+<%-- TODO: workaround: cytoscape.js needs to be included before virtual cohort session libraries --%>
+<script type="text/javascript" src="js/lib/cytoscape_js/cytoscape.js?<%=GlobalProperties.getAppVersion()%>"></script>
```
adamabeshouse: then do you need to remove a script tag somewhere else?

onursumer: Ideally yes, but there is also the `js_include_standard.jsp` where we don't include all the libraries, so we may still need the other script tag in its original place (`networks.jsp`).

I also noticed that virtual cohort session libraries in `js_include_standard.jsp` are included without the `$.noConflict()` call. So, actually, it is pretty confusing...

Issue #3382, opened 2017-11-16
BETA : Error while loading study view


kalletlak:
![screen shot 2017-11-16 at 1 56 32 pm](https://user-images.githubusercontent.com/12956870/32909951-01774580-cad6-11e7-8813-4b86b864edfb.png)

zhx828:
Already fixed that in another pull request, will submit soon. It does not affect anything.

Issue #3379, opened 2017-11-16
edits to custom cohort study label


tmazor:
When I first browse a custom study, the top of the page lists the full names of the studies:
![image](https://user-images.githubusercontent.com/10508276/32897499-cc4dff42-cab3-11e7-8451-e3cf7c1c6c0b.png)

Once I click the "share" button and load up the custom cohort, the labels on the top get much less descriptive:
![image](https://user-images.githubusercontent.com/10508276/32897520-e12bd15a-cab3-11e7-9dd1-5f97c44edb2f.png)

I think it would help to have the full study names. There also needs to be some punctuation to separate the two studies -- e.g. add a semicolon between "61 patients" and "Glioma (TCGA)"

Or why not keep the same format as on the combined study page - with the plus sign and the list of studies, including links back to the individual study summary pages, and just add the number of samples/patients included in the custom cohort?

tmazor:
If I start by running a query in multiple studies, there is a complete lack of information about which studies I queried at the top:
![image](https://user-images.githubusercontent.com/10508276/32902875-8a7ff404-cac1-11e7-8b10-bb5ec71a435b.png)

Can this also be replaced by the plus sign & the full list of studies?

jjgao:
Good point. The cancer study track should show up int he OncoPrint but it's still good idea to have the same information up there. 

@zhx828 given other issues, I would not worry about this for this release if it takes time. We can do this on later.

zhx828:
@jjgao @tmazor for release 1.10. We are going to show `Selected Studies` instead of `Custom Cohort` for virtual study and show exactly the same studies information like when you just combined the study. # of samples info is no longer available in 1.10 but will be enabled in next release when we enable user authentication and would be able to let them save the virtual study under their account.

inodb:
@zhx828 should we close this?

zhx828:
@inodb we could.

Issue #3360, opened 2017-11-13
progress bar when loading multiple studies on study view page


jjgao:
Loading study 1…
Loading study 2…

PR #3391, opened 2017-11-17
Add progress bar & Show studies info in the virtual study description section

zhx828:
Add progress bar when initializing the page;
Add progress bar for MutatedGene/CNA tables;
Replace loading bar with same progress bar library;
Keep the consistency, use Selected Studies instead of Custom Cohort for release 1.10;

This solves issue #3360 and #3382
This is related to #3379;

Loading main page:
loading-page

Loading mutated genes:
loading-mutated-genes

When interecting charts:
when-interecting-charts


portal/src/main/webapp/WEB-INF/jsp/dashboard/dashboard.jsp
```diff
@@ -326,40 +352,33 @@
                                 if (typeof response === 'string') {
                                     response = JSON.parse(response); 
                                 }
-                                $("#show_study_details").css('display','block');
-                                $("#study_name").html(response['data']['name']);
-                                $("#study_desc").html(response['data']['description']);
-                                //$("#submit_button").css('display','none');
+                                var _description = response['data']['description'] || '';
```
kalletlak: Over here we invoke session service to get name and description of virtual study. But in this case we can directly set name to `Selected Study` or `Selected Studies` and description directly if its a virtual study

zhx828: Ture, but thinking about that the user can still modify the virtual study name and description in the future, we may want to keep this.

kalletlak: 👍

PR #3469, opened 2017-11-28
Fix heatmap bug

adamabeshouse:


fedde-s:
Yup, I don't immediately see a problem either.

PR #3506, opened 2017-12-04
Improve documentation custom annotations and custom gene sets.

sandertan:
# What? Why?
In the original documentation:
- It was unclear what the format of custom gene sets for gene querying was.
- There was documentation regarding gene sets and custom driver annotations in `portal.properties.EXAMPLE` that should have been in `portal.properties-Reference.md`.
- The documentation that explains custom driver annotations could be improved.

- Remove documentation from `portal.properties.EXAMPLE`.
- Improve the documentation regarding custom driver annotations.
- Improve the documentation regarding custom gene sets for gene querying.
- Add table of contents to `portal.properties-Reference.md`.

### DIFF
--- a/docs/portal.properties-Reference.md
+++ b/docs/portal.properties-Reference.md
@@ -1,5 +1,19 @@
 This page describes the main properties within portal.properties.
 
+- [Database Settings](#database-settings)
+- [Segment File URL](#segment-file-url)
+- [Bitly API Username and Key](#bitly-api-username-and-key)
+- [Google Analytics](#google-analytics)
+- [Password Authentication](#password-authentication)
+- [CIViC integration](#civic-integration)
+- [OncoPrint](#oncoprint)
+- [Custom annotation of driver and passenger mutations](#custom-annotation-of-driver-and-passenger-mutations)
+	- [Enabling custom annotations in the OncoPrint](#enabling-custom-annotations-in-the-oncoprint)
+	- [Automatic selection of custom annotations](#automatic-selection-of-custom-annotations)
+	- [Automatic selection of OncoKB annotations](#automatic-selection-of-oncokb-annotations)
+	- [Automatic hiding of putative passenger mutations](#automatic-hiding-of-putative-passenger-mutations)
+- [Gene sets used for gene querying](#gene-sets-used-for-gene-querying)
+
 # Database Settings
 
 ```
@@ -61,6 +75,17 @@ authorization=true
 ```
 app.name should be set to the name of the portal instance referenced in the "AUTHORITY" column of the "AUTHORITIES" table.  See the [User Authorization](User-Authorization.md) for more information.
 
+# CIViC integration
+
+CIViC integration can be turned on or off with the following property (default: true):
+```
+show.civic=true|false
+```
+The CIViC API url is set to https://civic.genome.wustl.edu/api/ by default. It can be overridden using the following property:
+```
+civic.url=
+```
+
 # OncoPrint
 
 The default view in OncoPrint ("patient" or "sample") can be set with the following option. The default is "patient".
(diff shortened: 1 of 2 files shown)


sandertan:
According to the failed screenshot test, these properties affect the data shown in the mutation tab. I'll try to figure out what's happening here.

docs/portal.properties-Reference.md
```diff
@@ -61,52 +75,59 @@ authorization=true
...
 
-By default, when at least one of the custom annotations is enabled in portal.properties and the required columns are present in MAF files, custom annotations are automatically selected. You can change this behaviour and set the following properties to "false" to automatically unselect one or both custom driver/passenger mutations (by default, those properties are set to true):
+#### Automatic selection of custom annotations
+By default, when at least one of these custom annotations is enabled in `portal.properties` and the required columns are present in MAF files, custom annotations are automatically selected instead of OncoKB and Hotspots. To disable this default behavior and have OncoKB and Hotspots as the default selected annotations, set the following properties to `false`.
 ```
-oncoprint.custom_driver_annotation.default=false
-oncoprint.custom_driver_tiers_annotation.default=false
+oncoprint.custom_driver_annotation.default=true
+oncoprint.custom_driver_tiers_annotation.default=true
 ```
```
inodb: @sandertan this should prolly stay false right?

docs/portal.properties-Reference.md
```diff
@@ -61,52 +75,59 @@ authorization=true
...
 oncoprint.oncokb_hotspots.default=custom
-```
+``` 
 
-By default, the putative passenger mutations are not hidden. If you want to automatically hide Putative Passenger Mutations, set this property to true:
+#### Automatic hiding of putative passenger mutations
+By default, the selection box to hide putative passenger mutations is unchecked. If you want to automatically hide Putative Passenger Mutations, set this property to `true`:
 ```
-oncoprint.hide_passenger.default=true
+oncoprint.hide_passenger.default=false
```
inodb: @sandertan same here. Maybe the idea is that you're thinking it is false by default and I should set it to true if I read the text above, but that seems confusing to me.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -206,25 +206,13 @@ ucsc.build=hg19
...
 # oncoprint.oncokb_hotspots.default=custom
+# oncoprint.hide_passenger.default=false
 
-# Set this to true if you want to check the "hide Putative Passenger Mutations" checkbox by default.
-# oncoprint.hide_passenger.default=true
-
-# Change this to custom file (e.g. file:/<path>) to have a custom set of genes in query page:
-# querypage.setsofgenes.location =
\ No newline at end of file
+# Custom gene sets
```
inodb: why remove the example described in the text `file:/<path>`?

docs/portal.properties-Reference.md
sandertan: By setting `true`, @oplantalech implemented in a way that custom driver annotations are only enabled when data is present, so it should not influence instances without this data.

We could make it `false` to make it an conscious decision when you want to enable this functionality.

src/main/resources/portal.properties.EXAMPLE
sandertan: I wanted to add more documentation because knowing the example was insufficient. For example, you still needed to know the format of the JSON file. 

But the complete description was too long for the portal.properties I thought. So I moved it everything to: https://github.com/thehyve/cbioportal/blob/f55ee15f68c8ba9d2d1a1c0210bfe79c1babd28e/docs/portal.properties-Reference.md#gene-sets-used-for-gene-querying

docs/portal.properties-Reference.md
sandertan: By default Putative Passenger Mutations are **not** hidden in the portal. So to represent this behaviour with the new property, this property should be `false`. 

We can also keep all these properties commented out (as it was intially, and I have in latest version of this branch), to prevent them from having any influence on default installations.

docs/portal.properties-Reference.md
pieterlukasse: The documentation should just list the options. I like how civic option is documented above:

```
CIViC integration can be turned on or off with the following property (default: true):
show.civic=true|false
```

src/main/resources/portal.properties.EXAMPLE
pieterlukasse: I agree that the example should read:
```
# Change this to custom file (e.g. file:/<path>) to have a custom set of genes in query page:		
#querypage.setsofgenes.location=file:/<path>
```

docs/portal.properties-Reference.md
```diff
@@ -61,52 +75,59 @@ authorization=true
...
-Civic integration can be turned on or off with the following property (default: true):
-```
-show.civic=true|false
+# Gene sets used for gene querying
+To change the gene sets used for gene querying, create a JSON file and add gene sets in line with the format of [gene_lists.ts](https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/shared/components/query/gene_lists.ts). The JSON file should contain a structure equal to the part after `const gene_lists = `. Set the path to this file (e.g. file:/<path>) in the following property and restart Tomcat to apply the update.
 ```
-The Civic API url is set to https://civic.genome.wustl.edu/api/ by default. It can be overridden using the following property:
-```
-civic.url=
+querypage.setsofgenes.location=
```
pieterlukasse: this can be clearer if we write:
```
querypage.setsofgenes.location=file:/<path>
```

docs/portal.properties-Reference.md
sandertan: @pieterlukasse Done. Also made docu "Automatic selection of custom annotations" even simpler.

src/main/resources/portal.properties.EXAMPLE
sandertan: done.

docs/portal.properties-Reference.md
sandertan: Done

sandertan:
@inodb @pieterlukasse I've implemented your feedback

sandertan:
Rebased on top of hotfix.

Issue #3509, opened 2017-12-04
Loading of Query Page (index.jsp) very slow (~20s) for multiple studies


inodb:
Just getting index.do is ~20s (~9s download and 10s is spent in TTFB)

A lot of time is spent getting the data for the request:

I've used the debugger for profiling. This takes about 8s:

https://github.com/cBioPortal/cbioportal/blob/hotfix/core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java#L452-L468

Then this takes about 5s:

https://github.com/cBioPortal/cbioportal/blob/hotfix/core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java#L530-L539

(We might be able to get rid of DOWNLOAD_LINKS in the session completely, but it is still being used on the query page)

One option might be to parallelize these. We could also remove this information from the request completely. I think it is only used for initializing the QuerySession object, but would have to double check. If that's the case, it might be better to get this information through the API instead, which is what we want to do eventually anyway

inodb:
Partial fix that removes DOWNLOAD_LINKS: https://github.com/cBioPortal/cbioportal/pull/3510

PR #3510, opened 2017-12-04
For Queries exclude DOWNLOAD_LINKS in session

inodb:
The DOWNLOAD_LINKS attribute in the session is only necessary when downloading
the data. Otherwise we can remove it from the session.

Partial fix for #3509, takes about 5s less to download index.do when running on all studies


inodb:
Superseded by #3511

PR #3517, opened 2017-12-05
"No OQL" warning on plots, mutations, co expression tabs

adamabeshouse:
The screenshots are a bit confusing in this regard, but to clarify: the message is the same size on each tab (its the same jsp injected), but it is not placed the same, i.e. with the same padding. I can change that, I wasn't sure what the cleanest way to handle that would be though (help me @alisman ?)

image

image

image

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp
@@ -29,8 +29,12 @@
  - You should have received a copy of the GNU Affero General Public License
  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
 --%>
+
 <div class='section cbioportal-frontend' id='mutation_details'>
-    <img src='images/ajax-loader.gif' alt='loading'/>
+    <jsp:include page="global/no_oql_warning.jsp" flush="true" />
+    <div id="mutation_details_inner">
+        <img src='images/ajax-loader.gif' alt='loading'/>
+    </div>
 </div>
 
 <script type="text/javascript">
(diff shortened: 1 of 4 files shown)


portal/src/main/webapp/WEB-INF/jsp/mutation_details.jsp
```diff
@@ -29,8 +29,12 @@
  - You should have received a copy of the GNU Affero General Public License
  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
 --%>
+
 <div class='section cbioportal-frontend' id='mutation_details'>
-    <img src='images/ajax-loader.gif' alt='loading'/>
+    <jsp:include page="global/no_oql_warning.jsp" flush="true" />
+    <div id="mutation_details_inner">
```
adamabeshouse: @alisman I dont think I messed anything up here, but please pay extra attention

jjgao:
added @schultzn to get his approval.

jjgao:
@adamabeshouse is it possible to display the message when OQL is used?

adamabeshouse:
@jjgao you mean "this tab reflects the OQL specification from your query" ?

jjgao:
@adamabeshouse: I mean if the user just queries genes, it may not be necessary to show the message. Is it possible to display the message only when something like "EGFR: MUT" is queried. @tmazor thoughts?

tmazor:
I agree - the message should only be displayed if the user has used OQL in their query. If the query is just for gene names with no OQL, then there's no need to display the warning.

alisman:
@adamabeshouse lets discuss if we should maybe try to do this new codebase instead.  also would be nice to use bootstrap alert for consistency with other alerts.  lets talk.  sorry i'm late to the discussoin here.

adamabeshouse:
@alisman good idea

tmazor:
Hi all - any update on when this may get deployed? @adamabeshouse @jjgao @alisman

tmazor:
hey guys - any update on this? @adamabeshouse @alisman

adamabeshouse:
since we're going to merge plots in the next release, let's do it at the same time, all those tabs will be in react now

Issue #3201, opened 2017-09-29
Different sample counts in oncoprint and study view vs new variant-counts API


pieterlukasse:
The new variant-counts API shows a different "# samples" count than what is found in oncoprint and study view. 

Steps to reproduce:
1. query "Breast Invasive Carcinoma (TCGA, Provisional). This will result in oncoprint reporting there are 982 sequenced samples:
![image](https://user-images.githubusercontent.com/2900303/31012758-4e695c40-a513-11e7-978c-0f43c6cfa443.png)

2. go to the study view of this study: http://www.cbioportal.org/study?id=brca_tcga#summary . There you will see the number samples reported as 982:
![image](https://user-images.githubusercontent.com/2900303/31012139-054cf276-a511-11e7-97b2-30b7e938509d.png)

3. go to the patient view, e.g.  http://www.cbioportal.org/case.do#/patient?sampleId=TCGA-A8-A09Z-01&studyId=brca_tcga . Here you will find the numberOfSamples is 981 instead of 982:

![image](https://user-images.githubusercontent.com/2900303/31012832-9c3cca92-a513-11e7-98a5-0955203b64f0.png)

ersinciftci:
@pieterlukasse `SELECT COUNT(DISTINCT(mutation.SAMPLE_ID)) from mutation inner join genetic_profile on mutation.GENETIC_PROFILE_ID 
= genetic_profile.GENETIC_PROFILE_ID where genetic_profile.STABLE_ID = 'brca_tcga_mutations';` returns 981, so it means there is 1 sequenced sample that doesn't have any mutations.

pieterlukasse:
@ersinciftci Ok, thanks. Maybe the field `numberOfSamples` should get a different name, one that better covers this case? From the name, I expected it to be the number of profiled samples, which is not the case. This relates back to the discussion here: https://github.com/cBioPortal/cbioportal/issues/3067

ersinciftci:
@pieterlukasse yeah let's wait for that discussion to conclude, then I can either change the name or the number.

jjgao:
Agreed with @pieterlukasse: we should use the number of profiled samples as the denominator.

ersinciftci:
@jjgao `genetic_profile_samples` table doesn't have data for mutations, how do we get the number of profiled samples for mutation profiles?

jjgao:
[...]

PR #3562, opened 2017-12-12
Fix sample count in Variant Counts API

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/3201


inodb:
@ersinciftci seem to be core test errors? Could that be because of the db update?

ersinciftci:
@inodb all core tests are failing in all APIs, it seems someone changed the integration test database.

PR #3599, opened 2017-12-15
Fix for fusion integration test

pieterlukasse:
# What? Why?
Fixes failing integration test.

- new test now makes no assumptions regarding order of the data (which wasn't guaranteed)


inodb:
Thanks so much @pieterlukasse !

PR #3641, opened 2017-12-21
Hide With Mutation Data/With CNA Data for mskimpact study

zhx828:
No need to review. Code has been reviewed under iviz repo.


jjgao:
@zhx828 Could you try to find a general solution (priorities) for hiding auto-generated charts such as these two?

https://github.com/cBioPortal/cbioportal/issues/3642

PR #3284, opened 2017-10-24
Multiple genome support - change in DB schema

khzhu:
# What? 
This pull request proposed two new tables (reference_genome, reference_genome_gene) to track genome references and reference genes for multiple genome builds used by molecular profiling in tissue samples.

# Why?
Currently, cBioPortal only supports single reference genome build, since chromosome locus from the genetic coordinates are obtained by parsing the **_cytoband_** field in the gene table. All genes in the gene table having mapping cytogenetic bands from a particular genome build.

Changes proposed in this pull request:

- [ ] add a new **_reference_genome_** table to track genome builds such as GRCh38, GRCh37, GRCm38 used by molecular profiling
- [ ]  add a new **_reference_genome_gene_** table to track the reference genes for multiple genome builds
- [ ]  add a new **_select_** object in persistence myBatis gene mapper to retrieve the data from the new **_reference_genome_gene_** table 
- [ ]  modify database migration script to populate the **_reference_genome_** table and to backfill the **_reference_genome_gene_** table from existing **_gene_** table

# Checks
- [ ] Runs on Heroku.
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Google Style Guide](https://github.com/google/styleguide).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)
- [ ] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to
  hotfix.

# Any screenshots or GIFs?
If this is a new visual feature please add a before/after screenshot or gif
here with e.g. [GifGrabber](http://www.gifgrabber.com/).

# Notify reviewers
Read our [Pull request merging
policy](../CONTRIBUTING.md#pull-request-merging-policy). If you are part of the
cBioPortal organization, notify the approprate team (remove inappropriate):

@cBioPortal/frontend
@cBioPortal/backend
@cBioPortal/devops

If you are not part of the cBioPortal organization look at who worked on the
file before you. Please use `git blame <filename>` to determine that
and notify them here:

PR #3703, opened 2018-01-10
Multiple genome support mutation

khzhu:
## What? 
This pull request is a continuation of the pull request  **Multiple genome support - change in DB schema #3284**. It proposes a long-term solution for cBioPortal to support multiple reference genomes in mutational profiling

## Why?
Currently, cBioPortal only supports single genome build, since chromosome locus from the genetic coordinates are obtained by parsing the "cytoband" fields in the gene table. All genes in the gene table have mapping cytogenetic bands from particular genome build for example hg19.

- [ ] tracking reference genome used by adding a new field **REFRENCE_GENOME_ID** to both **mutation_event** and **genetic_profile** tables
- [ ] removing existing **NCBI_BUILD** field from the **mutation_event** table
- [ ] propagating **chr** from the **mutation_event** table and including this **chr** in query return-set for frontend react component to consume. Currently, chromosome names are parsed from the **cytoband** property of **gene** object.
- [ ] adjusting business model, data access objects, and persistence myBatis ORM mapping to reflect changes made above
- [ ] modifying importer script to include new **REFRENCE_GENOME_ID** in the MAF record object to add to both **mutation_event** and **genetic_profile** tables
- [ ] adjusting the gene data import script to take care of importing genes to the **reference_genome_gene** table
- [ ] adding a new import script to load new reference genome(s) to the database
- [ ] updating migration scripts to backfill **REFRENCE_GENOME_ID** to both **mutation_event** and **genetic_profile** tables
- [ ] adjusting test cases and scripts to work with the new table structures

# Checks
- [ ] Runs on Heroku.
- [ ] Make sure your commit messages end with a Signed-off-by string (this line can be automatically added by git if you run the `git-commit` command with the `-s` option)
- [ ] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to hotfix.

# Any screenshots or GIFs?

# Notify reviewers


khzhu:
@jjgao @n1zea144 @pieterlukasse 
the second part of **multiple reference genome support** pull request is ready for your review. This pull request only addresses issues associated with mutational profiling. Please also attached schema diagram for your reference. Thank you!
multiple_genome_profiling

db-scripts/src/main/resources/cgds.sql
```diff
@@ -285,14 +285,16 @@ CREATE TABLE `genetic_profile` (
...
+  `REFERENCE_GENOME_ID` int(4) NOT NULL,
   `GENETIC_ALTERATION_TYPE` varchar(255) NOT NULL,
   `DATATYPE` varchar(255) NOT NULL,
   `NAME` varchar(255) NOT NULL,
   `DESCRIPTION` mediumtext,
   `SHOW_PROFILE_IN_ANALYSIS_TAB` BOOLEAN NOT NULL,
   PRIMARY KEY (`GENETIC_PROFILE_ID`),
-  UNIQUE (`STABLE_ID`),
-  FOREIGN KEY (`CANCER_STUDY_ID`) REFERENCES `cancer_study` (`CANCER_STUDY_ID`) ON DELETE CASCADE
+  UNIQUE (`STABLE_ID`, `REFERENCE_GENOME_ID`),
```
ersinciftci: So genetic profiles won't be unique by stableId anymore? This is a big change, might cause a lot of complications. @jjgao

jjgao: Agreed with @ersinciftci. STABLE_ID should be unique.

jjgao: @khzhu do you have any use case that the same stable ID is required for different genome builds in the same study?

khzhu: @jjgao @ersinciftci , how about we set a rule: per study per genome build? The plan was to support multiple genome profiling within the same study.

ersinciftci: I don't know much about this new feature, but it feels like we should do this by using a relationship table, instead of changing the genetic_profile table. So one profile would have relations with more than one genome profile (or build? not sure what it means), but it still would be unique by stable id.

khzhu: We can do more normalization by introducing a new cross-reference genome profile table. But that would require quite changes in the code base, since the frontend would have to allow users to select the genome build of interest (having a reference genome dropdown box for instance), and then web services would also need to be changed to allow query the genetic profile based on the genome selection.

ersinciftci: As I said I don't know about the functionality much, but we have to change the whole application if genetic_profile can't be system-wise unique by the STABLE_ID anymore.

khzhu: @ersinciftci @jjgao ,please find revised schema diagram. I've added a new table named **reference_genome_profile** as a cross reference table for **genetic_profile** and **reference_genome** table. Now, **genetic_profile** can still be uniquely identified by **STABLE_ID**. In case of multiple genome profiling, then **reference_genome_profile** table will be used as a look up. 
reference_genome_profile

jjgao: @khzhu   I don't think we need reference_genome_profile table because one profile will only have one reference_genome_id. I think we should just change the current to "UNIQUE (`STABLE_ID`)".

jjgao: @khzhu what do you mean by multiple genome profiling? Do you mean, for example, mutations called by both grch38 and grch37?

khzhu: yes, @jjgao, do we leave **REFERENCE_GENOME_ID** in the genetic_profile table as it is now just remove it from the unique constraint?

jjgao: @khzhu in that case, should we have different profiles for different reference genomes, e.g. study_id_mutations_grch37 and study_id_mutations_grch38? Why do we need to have them in the same profile?

khzhu: @jjgao , the reference genome used in profiling is a very important piece of information once the system had multiple reference genome datasets, having hg19 datasets coexists with hg38 datasets for example. I would suggest to have it properly tracked in a table. 
Since we already had a rule, per profile per genome, that ruled out having a new reference_genome_profile table (Normally, one to one relation does not be tracked in a different table). Besides, both copy number alteration and segment tables (Third PR) will have to use the reference genome to pull out correct cytobands or genomic coordinate for display. Currently, the reference genome information used by the copy number segments is tracked in copy_number_seg_file table, that table is going to be deprecated soon. 
Also, the frontend QueryStore.ts file expected the format of STABLE_ID is in a format of **study_id+genetic_alteration_type**, adding a reference genome suffix might have more impact than adding a new **reference_genome_id** field since it only affect importer scripts.

ersinciftci: @khzhu do you know what line in QueryStore.ts expects a geneticProfileId to be in `study_id+genetic_alteration_type` format? Because it must be a mistake, we should never assume that.

khzhu: @ersinciftci , line 946 ~ 979.

khzhu: @ersinciftci, not geneticProfileId but sampleListID.

jjgao: @khzhu @ersinciftci: I think I am confused -- may we all are :). Let's see if we can agree on the followings:

1. The profile stable ID should be unique and therefore it should be `UNIQUE(STABLE_ID)`
2. One profile should only have one reference genome version, and therefore it is 1-to-N between genetic_profile and reference_genome table. In that case, we don't need reference_genome_profile table, and reference_genome_id will reside in genetic_profile table.
3. The frontend should not make any assumption about profile stable ID. We definitely have multiple profiles per alteration_type in a study.

Are you agree with the above?

khzhu: @jjgao , crystal clear, I could not agree more :).

ersinciftci: @jjgao `One profile should only have one reference genome version, and therefore it is 1-to-N between genetic_profile and reference_genome table.` I'm confused now :) you mean 1-to-1 (instead of 1-to-N)?

jjgao: @ersinciftci I think it's 1-N (or 1-M) because 1 profile can only have 1 genome but 1 genome may exist in multiple profiles. It is similar to Employee and Department relationship in this post (http://www.oracle.com/technetwork/issue-archive/2011/11-nov/o61sql-512018.html).

ersinciftci: @jjgao Ah I see, ok it's clear to me too now.

pieterlukasse:
Thanks @khzhu . As a general comment: I am wondering whether the `ImportGeneData.java` should perhaps be adjusted first before we start linking study data (mutation data in this PR) to the new table structure. Currently one can use `ImportGeneData.java` to update the `gene` table. With the current PR you are proposing, mutation data will be referring to `reference_genome_ gene` which cannot be easily updated, unless you adjust `ImportGeneData.java` accordingly. 
@jjgao @n1zea144 : thoughts on this one?

pieterlukasse:
@khzhu @jjgao @n1zea144 Another general comment: I would prefer to give `reference_genome_gene` its own auto generated primary key and let `mutation_event` refer to it instead of referring to the composite key. In this way, we slowly but surely will decouple molecular data from ENTREZ. 

Another question: why is this PR only updating `mutation_event` and not `mutation`?

khzhu:
@pieterlukasse , thank you so much for reviewing the PR. 

> ImportGeneData.java should perhaps be adjusted first before we start linking study data (mutation data in this PR) to the new table structure

We do not change anything in the gene table for this PR (both **Cytoband** and **length** will be dropped from the gene table in a future PR). I would vote for leaving ImportGeneData.java as it is but adding a new script to populate the reference_genome_ gene table. I do have my own python scripts to add genes to the reference_genome_ gene table which I can share. 

> Another question: why is this PR only updating mutation_event and not mutation?

The mutation table already has genetic_profile_id tracked which can be used to link back to the genetic_profile table to find reference genome used. 

>  I think GENETIC_ENTITY_ID should move from gene to reference_genome_gene and mutation_event should refer to GENETIC_ENTITY_ID from the genetic_entity table (thus should not be linked directly to reference_genome_gene). 

I will think about it deeply and get back to you...thanks again!

pieterlukasse:
@jjgao could you please share your thoughts on my last comments? We will be discussing it tomorrow.

ersinciftci:
@khzhu FYI, the db version is 2.5.0 now in `rc`, so you will need to update your db version in the PR

khzhu:
@ersinciftci , updated my DB version to 2.5.1, thanks!

db-scripts/src/main/resources/migration.sql
```diff
@@ -425,4 +426,31 @@ INSERT INTO reference_genome_gene (ENTREZ_GENE_ID, CYTOBAND, EXONIC_LENGTH, CHR,
...
--- ========================= end of reference genes related tables ========================================================================
+-- ========================= end of reference genes related tables =====================================================
+
+##version: 2.4.2
+-- ================== new reference genome foreign key in mutation_event and genetic_profile tables   ==================
+ALTER TABLE `mutation_event` ADD COLUMN `REFERENCE_GENOME_ID` INT(4) NULL AFTER `ENTREZ_GENE_ID`;
+ALTER TABLE `mutation_event`
+    ADD FOREIGN KEY (`REFERENCE_GENOME_ID`) REFERENCES `reference_genome` (`REFERENCE_GENOME_ID`) ON DELETE CASCADE;
+
+UPDATE `mutation_event` set REFERENCE_GENOME_ID = 1 WHERE NCBI_BUILD in ('hg19', 'GRCh37','37');
```
pieterlukasse: you shouldn't assume 1, 2 ,3 since the `REFERENCE_GENOME_ID` is auto incremented. You can get the ID using a select statement.

khzhu:
@n1zea144 @pieterlukasse ,
I've updated the diagram based on our meeting today. Thanks again for your time!
multi_genome_support_pr 2

db-scripts/src/main/resources/migration.sql
khzhu: @pieterlukasse , good idea, change made, thanks!

khzhu:
hi @pieterlukasse , are you agree with updated schema diagram? thanks!

khzhu:
@n1zea144 ,@pieterlukasse , 
since the main purpose to this PR is to modify the mutation table to support multiple reference genomes. A future PR after this PR is to focus on the copy number segment table. So, I am planning to have a fourth PR working on the gene table, doing things such as dropping **cytoband**, **lenght**, updating gene import scripts. cBioPortal has been build around **gene** , any changes made to the gene table has to be well investigated. What do you think? Thanks!

db-scripts/src/main/resources/migration.sql
```diff
@@ -380,6 +380,8 @@ ALTER TABLE `mutation` ADD COLUMN `DRIVER_FILTER_ANNOTATION` VARCHAR(80) NULL;
 ALTER TABLE `mutation` ADD COLUMN `DRIVER_TIERS_FILTER` VARCHAR(50) NULL;
 ALTER TABLE `mutation` ADD COLUMN `DRIVER_TIERS_FILTER_ANNOTATION` VARCHAR(80) NULL;
 
+UPDATE info SET DB_SCHEMA_VERSION="2.4.0";
```
inodb: was 2.4.0 missing?

khzhu: @inodb, added back on in this PR. was accidentally chopped off.

khzhu:
@pieterlukasse @n1zea144 

> Actually, I think GENETIC_ENTITY_ID should move from gene to reference_genome_gene and mutation_event should refer to GENETIC_ENTITY_ID from the genetic_entity table (thus should not be linked directly to reference_genome_gene). We already decoupled genetic_alteration table from gene, perhaps now is the time to do the same for mutation data (unless we have no plans to support mutations in anything else than genes).

I had deep thought on this and will not go with your suggestion, since **genetic_alteration** table will also be affected if we go with that normalization, not to mention an extra mapping we had to do for each mutation (keep in mind **mutation_event** is a huge table). 
It is hard to steer away from **ENTREZ_GENE_ID** at moment, as it is one of major keys of the cBioPortal, having 1541 usages from the backend alone.

db-scripts/src/main/resources/migration.sql
```diff
@@ -425,23 +426,51 @@ INSERT INTO reference_genome_gene (ENTREZ_GENE_ID, CYTOBAND, EXONIC_LENGTH, CHR,
...
+    c1.`CANCER_STUDY_ID` = cancer_study.`CANCER_STUDY_ID` GROUP BY cancer_study.`CANCER_STUDY_ID`, `SAMPLE_ID` HAVING SUM(`END`-`START`) > 0;
+
+UPDATE info SET DB_SCHEMA_VERSION="2.5.0";
+
+##version: 2.5.1
+
+-- ================== new reference genome foreign key in mutation_event and genetic_profile tables   ==================
+ALTER TABLE `mutation_event` ADD COLUMN `REFERENCE_GENOME_ID` INT(4) NULL AFTER `ENTREZ_GENE_ID`;
+ALTER TABLE `mutation_event`
+    ADD FOREIGN KEY (`REFERENCE_GENOME_ID`) REFERENCES `reference_genome` (`REFERENCE_GENOME_ID`) ON DELETE CASCADE;
```
pieterlukasse: ENTREZ is missing here. You should drop old FK to `gene` table and make new one referring to the composite key `REFERENCE_GENOME_ID, ENTREZ_GENE_ID`.

pieterlukasse:
hi @khzhu , thanks for the updates. Regarding your last comment, `genetic_alteration` table is already decoupled from ENTREZ, so it wouldn't be affected. Anyway, I agree that it is perhaps out of scope to try this decoupling for mutation data now...let's keep the scope focused on multiple genome support indeed. 

One remaining issue with this PR is the lack of support for importing other genes (e.g. from mouse build) or updated set of genes with `ImportGeneData.java`. A solution could be to update the documentation here (https://github.com/cBioPortal/cbioportal/blob/master/docs/Updating-gene-and-gene_alias-tables.md) with some extra steps like (example for mouse/GRCm38, inspired on what you already added on `migration.sql`): 
```sql
INSERT INTO reference_genome_gene (ENTREZ_GENE_ID, CYTOBAND, EXONIC_LENGTH, CHR, REFERENCE_GENOME_ID)
(SELECT 
	ENTREZ_GENE_ID, 
	CYTOBAND, 
	LENGTH,
    SUBSTRING_INDEX(SUBSTRING_INDEX(SUBSTRING_INDEX(gene.CYTOBAND,IF(LOCATE('p', gene.CYTOBAND), 'p', 'q'), 1),'q',1),'cen',1),
	(SELECT REFERENCE_GENOME_ID FROM reference_genome WHERE BUILD_NAME = 'GRCm38') 
FROM `gene`);
```

khzhu:
Hi @pieterlukasse, 
I am updating ImportGeneData.jave script so that it will also add/update the *reference_genome_gene** table, plus a new reference genome import script for adding/updating the reference genome. I will be able to wrap up and commit the changes this week. Thanks again for taking time review the codes!!

db-scripts/src/main/resources/migration.sql
khzhu: @pieterlukasse , updated the ** REFERENCE_GENOME_ID** foreign key, now references to the **reference_genome_gene** table.

khzhu:
Hi @pieterlukasse @n1zea144,
here comes the updates:

1. gene import script has been adjusted to the new table structure. It now populates the **reference_genome_gene** table from the genecode gtf file as well. The new "genome-build" option has been added to the gene import script to indicate the genome build the reference genome gene references to. For example,

> ./importGenes.pl --genes /Users/kelsyzhu/Downloads/Homo_sapiens.gene_info --gtf /Users/kelsyzhu/Downloads/gencode.v27.annotation.gtf --genome-build GRCh38

2. The gene importer documents has been updated. 
3. The code has been rebased to the latest **rc** and pushed back to the origin. 

If any of you can take a look that would be great appreciated. Thanks!

core/src/main/java/org/mskcc/cbio/portal/dao/DaoReferenceGenome.java
```diff
@@ -85,13 +96,13 @@ public static void addReferenceGenome(ReferenceGenome referenceGenome, boolean o
...
      */
-    public static void deleteAllRecords() throws DaoException {
+    public void deleteAllRecords() throws DaoException {
 
         Connection con = null;
         PreparedStatement pstmt = null;
         ResultSet rs = null;
         try {
-            con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            con = JdbcUtil.getDbConnection(DaoReferenceGenome.class);
```
pieterlukasse: a TRUNCATE won't normally work since there are foreign keys to this table, right?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoReferenceGenome.java
```diff
@@ -103,13 +114,74 @@ public static void deleteAllRecords() throws DaoException {
         }
     }
 
+    /**
+     * Delete Reference Genome by internal ID
+     * @param referenceGenomeId internal ID of reference genome
+     * @throws DaoException Database Error.
+     */
+    public void deleteRecord(int referenceGenomeId) throws DaoException {
```
pieterlukasse: this means all genes are also deleted and hence all studies are deleted...quite drastic measure if all you want is to UPDATE the genome build. Wouldn't it be better to implement an UPDATE statement?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -435,7 +453,7 @@ public static int calculateGeneLength(List<long[]> loci) {
             bitSet.set((int)(l[0]-min), ((int)(l[1]-min)));
         }
         
-        return bitSet.cardinality();
+        return new long[]{min, max, bitSet.cardinality()};
```
pieterlukasse: nice

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportGeneData.java
```diff
@@ -42,6 +42,7 @@
 import org.mskcc.cbio.portal.dao.DaoGeneOptimized;
 import org.mskcc.cbio.portal.model.CanonicalGene;
 import org.mskcc.cbio.portal.util.ProgressMonitor;
+import org.omg.CORBA.IMP_LIMIT;
```
pieterlukasse: why is this needed?

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportReferenceGenome.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.File;
+
+import org.mskcc.cbio.portal.dao.DaoReferenceGenome;
+import org.mskcc.cbio.portal.model.ReferenceGenome;
+import org.mskcc.cbio.portal.util.ProgressMonitor;
+import org.omg.CORBA.IMP_LIMIT;
```
pieterlukasse: why is this needed?

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportGeneData.java
khzhu: No need. Removed.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportReferenceGenome.java
khzhu: removed as well.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoReferenceGenome.java
khzhu: good point! Change made.

khzhu:
Hi @inodb @ersinciftci,
I compared failed screenshots (on the right) with ones in the repo (on the left). 
- patient_view_lgg_ucsf_2014_case_id_P04.png look same but reported not same/failed by Travis CI, which do know why
screen shot 2018-02-28 at 4 11 57 pm

- study_view_lgg_ucsf_2014.png are indeed different, since the whole top row of study view was cut off ie. no cancer type search box, no cancer type detailed, and overall survival plot (most top right corner).  So, Travis CI was right. However, that was not the error introduced by this PR. Any suggestions to get this solved? 
screen shot 2018-02-28 at 4 19 25 pm

db-scripts/src/main/resources/migration.sql
```diff
@@ -446,9 +447,37 @@ c1.`CANCER_STUDY_ID` = cancer_study.`CANCER_STUDY_ID` GROUP BY cancer_study.`CAN
...
+-- ================== new reference genome foreign key for mutation_event and genetic_profile tables   ==================
+ALTER TABLE `mutation_event` ADD COLUMN `REFERENCE_GENOME_ID` INT(4) NULL AFTER `ENTREZ_GENE_ID`;
+ALTER TABLE `mutation_event`
+    ADD FOREIGN KEY (`REFERENCE_GENOME_ID`) REFERENCES `reference_genome_gene` (`REFERENCE_GENOME_ID`) ON DELETE CASCADE;
+
+UPDATE `mutation_event` set REFERENCE_GENOME_ID = (SELECT `REFERENCE_GENOME_ID` FROM `reference_genome` WHERE BUILD_NAME = 'GRCh37') WHERE NCBI_BUILD in ('hg19', 'GRCh37','37');
+UPDATE `mutation_event` set REFERENCE_GENOME_ID = (SELECT `REFERENCE_GENOME_ID` FROM `reference_genome` WHERE BUILD_NAME = 'GRCh38') WHERE NCBI_BUILD in ('hg38', 'GRCh38');
+UPDATE `mutation_event` set REFERENCE_GENOME_ID = (SELECT `REFERENCE_GENOME_ID` FROM `reference_genome` WHERE BUILD_NAME = 'GRCm38') WHERE NCBI_BUILD in ('mm10', 'GRCm38');
+
+ALTER TABLE `mutation_event` DROP COLUMN `NCBI_BUILD`;
```
pieterlukasse: @khzhu : have you tried importing a new study using metaImport.py tool after dropping this field? So a new study that has not been imported before and that contains new mutations that are not yet in `mutation_event`.  I think this will fail, because the `NCBI_BUILD`  field is dropped here, but I don't see that it is dropped from the dao and model layers used by the importer in this PR.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -101,6 +101,7 @@ public static int addMutationEvent(ExtendedMutation.MutationEvent event) throws
         MySQLbulkLoader.getMySQLbulkLoader("mutation_event").insertRecord(
                 Long.toString(event.getMutationEventId()),
                 Long.toString(event.getGene().getEntrezGeneId()),
+                Integer.toString(DaoReferenceGenome.getInstance().getReferenceGenomeByName(event.getNcbiBuild())),
```
pieterlukasse: can we please cache this? There could be many, many thousands of such records, all executing the same query here. I expect this will significantly slow down the importing....

.travis.yml
```diff
@@ -77,9 +77,9 @@ script:
           -Dfinal.war.name=cbioportal \
           -Ddb.user=cbio_user \
           -Ddb.password=cbio_pass \
-          -Ddb.portal_db_name=public_test \
-          -Ddb.connection_string=jdbc:mysql://sm4sa6ozhn5ya6.cuooweljyhix.us-east-1.rds.amazonaws.com:3306/ \
-          -Ddb.host=sm4sa6ozhn5ya6.cuooweljyhix.us-east-1.rds.amazonaws.com \
+          -Ddb.portal_db_name=cgds_test \
+          -Ddb.connection_string=jdbc:mysql://pughlabdb.cobap3rgryeo.us-east-1.rds.amazonaws.com:3306/ \
```
pieterlukasse: this should not go to `rc`

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
```diff
@@ -83,6 +83,7 @@ public void setUp() throws DaoException {
 		newGeneticProfile.setStableId("study_tcga_pub_test");
 		newGeneticProfile.setProfileName("Barry CNA Results");
 		newGeneticProfile.setDatatype("test");
+		newGeneticProfile.setReferenceGenomeId(1);
```
pieterlukasse: same here

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
```diff
@@ -275,6 +276,7 @@ private void runImportRnaData1() throws DaoException, IOException{
         geneticProfile.setDatatype("CONTINUOUS");
         geneticProfile.setProfileName("MRNA Data");
         geneticProfile.setProfileDescription("mRNA Data");
+        geneticProfile.setReferenceGenomeId(1);
```
pieterlukasse: same here

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
```diff
@@ -328,6 +330,7 @@ public void testImportmRnaData2() throws Exception {
...
```
pieterlukasse: here

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
```diff
@@ -420,6 +423,7 @@ public void testImportRppaData() throws Exception {
         geneticProfile.setDatatype("LOG2-VALUE");
         geneticProfile.setProfileName("RPPA Data");
         geneticProfile.setProfileDescription("RPPA Data");
+        geneticProfile.setReferenceGenomeId(1);
```
pieterlukasse: and here

core/src/test/java/org/mskcc/cbio/portal/web_api/TestGetProfileData.java
```diff
@@ -74,6 +74,7 @@ public void setUp() throws DaoException {
 		newGeneticProfile.setStableId("study_tcga_pub_test");
 		newGeneticProfile.setProfileName("Barry CNA Results");
 		newGeneticProfile.setDatatype("test");
+		newGeneticProfile.setReferenceGenomeId(1);
```
pieterlukasse: here

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
```diff
@@ -14,6 +14,17 @@
...
         </if>
     </sql>
+
+    <sql id="selectRefGene">
+        gene.ENTREZ_GENE_ID AS "${prefix}entrezGeneId",
+        gene.HUGO_GENE_SYMBOL AS "${prefix}hugoGeneSymbol"
+        <if test="projection == 'SUMMARY' || projection == 'DETAILED'">
+            ,
+            gene.TYPE AS "${prefix}type",
+            reference_genome_gene.cytoband AS "${prefix}cytoband",
```
pieterlukasse: there is no join with this `reference_genome_gene` table?

db-scripts/src/main/resources/migration.sql
khzhu: @pieterlukasse , yes, I did import public lgg_unsf_2014 dataset on a migrated database and a fresh created database. The mutation_event table populated correctly.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
...
+4. nonN_bases: the total number of non N bases
+5. URL: the URL to download the reference genome 
+6. release_date: the date when reference genome released, in a format of yyyy-mm-dd hh:mm:ss
+
+here is a sample reference genome file:
+```
+#species	name	build_name	nonN_bases	URL	release_date
+human	hg19	GRCh37	2897310462	http://hgdownload.cse.ucsc.edu/goldenPath/hg19	2009-02-01 00:00:00
+human	hg38	GRCh38	3049315783	http://hgdownload.cse.ucsc.edu/goldenPath/hg38	2013-12-24 00:00:00
+mouse	mm10	GRCm38	2652783500	http://hgdownload.cse.ucsc.edu//goldenPath/mm10	2011-12-01 00:00:00
```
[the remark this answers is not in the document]
khzhu: nice catch, fixed.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
+# Add new reference genome to the database
+This manual is intended to guide you to load new reference genome(s) to the database. 
+
+### What is **Reference Genome** (also know as reference assembly)?
+A digital nucleic acid sequence database assembled by scientists as a representative example of a species' set of genes. Reference genomes can be accessed online using dedicated browsers such as Ensembl or UCSC Genome Browser. 
+
+### Why does reference genomes matter?
+Raw reads from sample tissue are often aligned to a particular reference genome. Choose the same reference genome used by the alignment help us accurately identify the chromosomal location of the gene and the position of cyto-genetic bands within a chromosome.
```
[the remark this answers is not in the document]
khzhu: take your words for it!

.travis.yml
khzhu: a temp solution to move things forward. @inodb  will at certain point help to set up a public aws to replace the pughlab one.

core/src/main/java/org/mskcc/cbio/portal/model/GeneticProfile.java
```diff
@@ -65,6 +66,7 @@ public GeneticProfile(String stableId, int cancerStudyId, GeneticAlterationType
       this.profileName = profileName;
       this.profileDescription = profileDescription;
       this.showProfileInAnalysisTab = showProfileInAnalysisTab;
+      this.referenceGenomeId = 1;
```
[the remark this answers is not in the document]
khzhu: it really does not matter, since later on importer will update this field based on the reference genome listed in the portal properties file. Can use a deprecated tag.

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -226,6 +227,13 @@ public static GeneticProfile loadGeneticProfileFromMeta(File file) throws IOExce
         String profileDescription = properties.getProperty("profile_description");
         String geneticAlterationTypeString = properties.getProperty("genetic_alteration_type");
         String datatype = properties.getProperty("datatype");
+        String genomeBuild = properties.getProperty("reference_genome_id");
+        int referenceGenomeId = 1;
```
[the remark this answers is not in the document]
khzhu: good idea, will add a validation rule.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
khzhu: it will be supplied by calling mutationMapper object.

core/src/test/java/org/mskcc/cbio/portal/dao/TestDaoGeneticProfile.java
```diff
@@ -95,6 +95,7 @@ public void testDaoCreateGeneticProfile() throws DaoException {
 		geneticProfile.setStableId("test");
 		geneticProfile.setGeneticAlterationType(GeneticAlterationType.FUSION);
 		geneticProfile.setDatatype("test");
+		geneticProfile.setReferenceGenomeId(1);
```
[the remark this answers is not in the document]
khzhu: @pieterlukasse , all test suites use a in memory DB. If you take a look at seed_mini.sql, reference_genome_ids are assigned by insertion sql statements not auto generated.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportExtendedMutationData.java
```diff
@@ -75,6 +75,7 @@ public void setUp() throws DaoException {
 		geneticProfile.setStableId("test");
 		geneticProfile.setGeneticAlterationType(GeneticAlterationType.MUTATION_EXTENDED);
 		geneticProfile.setDatatype("test");
+		geneticProfile.setReferenceGenomeId(1);
```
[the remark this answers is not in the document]
khzhu: reference_id assigned not auto generated. will not change.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
khzhu: reference_id assigned not auto generated. will not change.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
khzhu: reference_id assigned not auto generated. will not change.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTabDelimData.java
khzhu: reference_id assigned not auto generated. will not change.

core/src/test/java/org/mskcc/cbio/portal/web_api/TestGetProfileData.java
khzhu: reference_id assigned not auto generated. will not change.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
khzhu: for sure. Let's have one!

khzhu: @pieterlukasse , I've added the cache system to DaoReferenceGenome class.

khzhu:
@pieterlukasse , cached reference genome for not slow down importing :). Also, all issues raised have been addressed. The code is ready for the review. Thanks!

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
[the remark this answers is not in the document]
khzhu: done.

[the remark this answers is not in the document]
khzhu: **reference_genome_id** is already checked by the cbioportal_common.py file. 
```
    MetaFileTypes.SEG: {
        'cancer_study_identifier': True,
        'genetic_alteration_type': True,
        'datatype': True,
        'reference_genome_id': True,
        'data_filename': True,
        'description': True
    },
```
what do you think we also add **reference_genome_id** key to the meta_study file and make it an option key, since we only support one genome build per study? 
```
    MetaFileTypes.STUDY: {
        'cancer_study_identifier': True,
        'type_of_cancer': True,
        'name': True,
        'description': True,
        'short_name': True,
        'citation': False,
        'pmid': False,
        'groups': False,
        'reference_genome_id': False,
        'add_global_case_list': False
    },
```
In that way, we can also remove the following attributes
```
# species and genomic information
species=human
ncbi.build=37
ucsc.build=hg19
``` 
from the portal.properties file to support multiple genomes.

khzhu:
@pieterlukasse , at this point, all issues you raised have been addressed. Here comes my proposed change to the **species and genomic information** section in portal.properties file to support multiple genome builds. Please have your final review, thank you!
> species=human
genome.build=37,hg19,GRCh37,GRCh38
genome.name=hg

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
pieterlukasse: when? In next PR?

db-scripts/src/main/resources/migration.sql
pieterlukasse: ok, thanks

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
pieterlukasse: nice, thanks!

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -228,11 +229,7 @@ public static GeneticProfile loadGeneticProfileFromMeta(File file) throws IOExce
         String geneticAlterationTypeString = properties.getProperty("genetic_alteration_type");
         String datatype = properties.getProperty("datatype");
         String genomeBuild = properties.getProperty("reference_genome_id");
-        int referenceGenomeId = 1;
-        if (genomeBuild != null && !genomeBuild.isEmpty()) {
-            referenceGenomeId = DaoReferenceGenome.getReferenceGenomeIdByName(genomeBuild);
-        }
-
+        int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeIdByName(genomeBuild);
```
pieterlukasse: I meant something else in my last comment. The previous code was almost correct. I mean something like:
```java
int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeIdByName("GRCh37");	
if (genomeBuild != null && !genomeBuild.isEmpty()) {	
           referenceGenomeId = DaoReferenceGenome.getReferenceGenomeIdByName(genomeBuild);	
}
```

pieterlukasse: so, `int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeIdByName("GRCh37");	` instead of `int referenceGenomeId = 1 `

core/src/main/scripts/importer/validateData.py
```diff
@@ -1227,6 +1230,8 @@ def checkLine(self, data):
                         'No Amino_Acid_Change or HGVSp_Short value. This '
                             'mutation record will get a generic "MUTATED" flag',
                         extra={'line_number': self.line_number})
+        if ("NCBI_Build") not in self.cols or data[self.cols.index('NCBI_Build')] not in ["GRCh37", "GRCh38","GRCm38"]:
```
pieterlukasse: I think we already have a validation for this field. I actually meant a warning for the **meta** file property that is read in the java code in `GeneticProfileReader.java` :  `String genomeBuild = properties.getProperty("reference_genome_id");`

core/src/main/java/org/mskcc/cbio/portal/model/GeneticProfile.java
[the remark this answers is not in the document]
pieterlukasse: added some comments to the new code. It will need some extra fixes.

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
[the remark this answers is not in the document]
pieterlukasse: added some comments to the new code. It will need some extra fixes.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoReferenceGenome.java
```diff
@@ -230,18 +230,17 @@ public static ReferenceGenome getReferenceGenomeByGenomeName(String genomeName)
...
+     * @param name   Name of Reference Genome or Genome Assembly
      * @throws DaoException Database Error.
      */
 
     public static int getReferenceGenomeIdByName(String name) throws DaoException {
         try {
             return genomeInternalIds.get(name);
         } catch (java.lang.NullPointerException exp) {
-            return ReferenceGenome.REFERENCE_GENOME_ID; // NCBI_BUILD field was an optional in the past
+            return genomeInternalIds.get("GRCh37"); // NCBI_BUILD field was an optional in the past
```
pieterlukasse: aha, ok, got it. Maybe you can ignore my previous comment.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoReferenceGenome.java
```diff
@@ -230,18 +230,17 @@ public static ReferenceGenome getReferenceGenomeByGenomeName(String genomeName)
         return byGenomeName.get(genomeName);
     }
     
-    
     /**
      * Retrieve reference genome of interest by genome name or genome assembly name
-     * @param buildName   Name of Reference Genome or Genome Assembly
+     * @param name   Name of Reference Genome or Genome Assembly
```
pieterlukasse: perhaps add here that it returns the id for "GRCh37" if the name is not found.

core/src/main/scripts/importer/validateData.py
```diff
@@ -268,9 +268,9 @@ def load_genome_info(self, properties_filename):
                 sp_line = line.split('=', 1)
                 if sp_line[0] == 'species':
                     self.species = sp_line[1]
-                elif sp_line[0] == 'ncbi.build':
-                    self.ncbi_build = sp_line[1]
-                elif sp_line[0] == 'ucsc.build':
+                elif sp_line[0] == 'genome.build':
+                    self.ncbi_build = sp_line[1].split(",")
```
pieterlukasse: maybe rename this variable to `genome_builds` ?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1250,15 +1250,7 @@ def printDataInvalidStatement(self, value, col_index):
...
-        if value != '':
-            # based on MutationDataUtils.getNcbiBuild
-            if self.portal.species == "human":
-                if value not in [str(self.portal.ncbi_build), self.portal.genome_build, 'GRCh'+str(self.portal.ncbi_build)]:
-                    return False
-            elif self.portal.species == "mouse":
-                if value not in [str(self.portal.ncbi_build), self.portal.genome_build, 'GRCm'+str(self.portal.ncbi_build)]:
-                    return False
-        return True
+        return value in (self.portal.ncbi_build)
```
pieterlukasse: I think I don't agree with this simplification of the code. Previous validation did check if the build was according to species expected. I don't think we support  multiple species yet in a single portal instance, so this check is still relevant. For mouse you need another set of entrez ids and another set of (possibly conflicting) hugo symbols. I think more work will be needed to support multiple same hugo symbols pointing to different entrez ids...I think some parts of the code try to come up with a single entrez for a given hugo...I think multi species might break this assumption.

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
pieterlukasse: Ok, I see latest code has something similar :+1:

core/src/main/scripts/importer/validateData.py
pieterlukasse: I see latest code uses the correct method :+1:

pieterlukasse:
@khzhu thanks for the update. I did add some comments on the validateData.py for which I think the validation needs to be fixed (see comments around `checkNCBIbuild`).

core/src/main/java/org/mskcc/cbio/portal/dao/DaoReferenceGenome.java
khzhu: @pieterlukasse , the code already did that. Now, I am thinking about add a new param of species, since if the instance is for mouse, then by default we shall return the id for "GRCm37". what do you think?
```
            if (species.equalsIgnoreCase("mouse")) {
                return genomeInternalIds.get("GRCm37");
            } else if (species.equalsIgnoreCase("human")){
                return genomeInternalIds.get("GRCh37"); // NCBI_BUILD field was an optional in the past
            }  else {
                throw new DaoException("Species not supported yet");
            } 
```

core/src/main/scripts/importer/validateData.py
khzhu: agree! will make changes.

core/src/main/scripts/importer/validateData.py
khzhu: @pieterlukasse, no need, since 
if human, then we have
```
species=human
genome.build=37,hg19,GRCh37,GRCh38
genome.name=hg
```
if mouse, then we have
```
species=mouse
genome.build=GRCm37,GRCm38
genome.name=mm
```
do we have any use cases to support more than one species in single instance? if that is the case then we need something like below in the portal.properties file?
```
species=human,mouse
genome.build=37,hg19,GRCh37,GRCh38,GRCm37,GRCm38
genome.name=mam
```

khzhu:
@pieterlukasse , thank you for the review, I have addressed your comments. Please let me know if you have any further questions.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
...
+6. release_date: the date when reference genome released, in a format of yyyy-mm-dd hh:mm:ss
+
+here is a sample reference genome file:
+```
+#species	name	build_name	nonN_bases	URL	release_date
+human	hg19	GRCh37	2897310462	http://hgdownload.cse.ucsc.edu/goldenPath/hg19	2009-02-01 00:00:00
+human	hg38	GRCh38	3049315783	http://hgdownload.cse.ucsc.edu/goldenPath/hg38	2013-12-24 00:00:00
+mouse	mm10	GRCm38	2652783500	http://hgdownload.cse.ucsc.edu/goldenPath/mm10	2011-12-01 00:00:00
+```
+##2. Using import script to load reference genome(s) data to the database:
```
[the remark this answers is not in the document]
sandertan: Just noticed that in your previous PR, those 3 reference genomes are included in the database schema. Nice!

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -42,7 +42,14 @@ After downloading, go to your downloads directory, decompress the file and add i
...
  export PORTAL_HOME=<your_cbioportal_dir>
-./importGenes.pl --genes <ncbi_species.gene_info> --gtf <gencode.v25.annotation.gtf>
+./importGenes.pl --genes <ncbi_species.gene_info> --gtf <gencode.v25.annotation.gtf> --genome-build <GRCh37>
+```
+**IMPORTANT NOTE**: 
+1. The **reference_genome** table needs to be populated before updating the **gene** table. Further details can be found in [this document](import-reference-genome.md). 
+2. Loading genes only to the **reference_genome_gene** table without updating the **gene** table, please use the following command:
+```
+./importGenes.pl --gtf <gencode.v27.annotation.gtf> --genome-build <GRCh38>
+
```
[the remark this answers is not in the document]
khzhu: Hi @sandertan,
When using a clean database using the seed database, does this include a reference genome? What is the default reference genome?
If the seed database does not include a reference genome, why do I have to add a reference genome to update the genes?
>>> you will need to run the migration script to update you schema. The migration script will populate the reference genome table as well.

As a user, I do not understand the difference between the reference_genome_gene table and gene table. Can you document this somewhere?
>>> how about document in the **updating-gene-and-gene_alias-tables.md** file

Can genes be in the both the reference_genome_gene table and gene table? Which one is used in that case?
>>>yes. Depends on what you are looking for. For **HUGO_GENE_SYMBOL** then reference the **gene** table, for "cytoband", "exonic_length", "start/end" of the gene then reference the **reference_genome_gene** table because those attributes differ between genome builds.

As far as I understand, this PR tries to force the user when updating genes and genomic coordinates, to include a reference genome. This makes a lot of sense. Can you document this reasoning in this markdown?
>>> there is a document **import-reference-genome.md** included in the PR . Can you please check if anything missed?
Thank for help with testing! I will push back those changes based on Pieter comments this morning.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
+# Add new reference genome to the database
+This manual is intended to guide you to load new reference genome(s) to the database.
```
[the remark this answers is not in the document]
khzhu: @sandertan, because we display **cytoband** in the CNA table which is tied with genomic coordinates. Also, those plots on the patient view are using genomic coordinates.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
...
+
+## How to load reference genome(s) to the database?
+
+##1. prepare a tab delimited txt file with the following fields:
+1. species: the name of organic species e.g human
+2. name: the name of reference genome eg. hg38
+3. build_name: the name of genome assembly e.g GRCh38
+4. nonN_bases: the total number of non N bases
+5. URL: the URL to download the reference genome 
+6. release_date: the date when reference genome released, in a format of yyyy-mm-dd hh:mm:ss
```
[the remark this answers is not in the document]
khzhu: @sandertan , you will need to download mm9 fasta file from UCSC, and then count the number of bases (A, C, G, T) other than "N". Sometimes also called effective genome size.
You can always find the release date from the README file in the genome assembly.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
+# Add new reference genome to the database
+This manual is intended to guide you to load new reference genome(s) to the database. 
+
+### What is **Reference Genome** (also know as reference assembly)?
```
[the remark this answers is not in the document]
khzhu: good to know, thanks!

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
...
+A digital nucleic acid sequence database assembled by scientists as a representative example of a species' set of genes. Reference genomes can be accessed online using dedicated browsers such as Ensembl or UCSC Genome Browser. 
+
+### Why does reference genomes matter?
+Some of the views in cBioPortal rely on the reference genome version matching the reference genome version used to analyse the original data (before it was loaded into cBioPortal DB).
```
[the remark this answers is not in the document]
khzhu: @sandertan , the patient view and all tables using cytoband, mutation/seg plots using genomic coordinates.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -196,8 +196,8 @@ priority_studies=
 
 # species and genomic information
 species=human
-ncbi.build=37
-ucsc.build=hg19
+genome.build=37,hg19,GRCh37,GRCh38
+genome.name=hg
```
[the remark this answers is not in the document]
khzhu: @pieterlukasse ,@sandertan , one cBioPortal instance can support multiple reference genomes from multiple species. I am working on validateData.py script to have it work with a portal.properties like below:
```
# species and genomic information
species=human;human,mouse
genome.build=37,GRCh37;GRCh38;GRCm38
genome.name=hg19;hg38;mm10
```

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
...
+
+## How to load reference genome(s) to the database?
+
+##1. prepare a tab delimited txt file with the following fields:
```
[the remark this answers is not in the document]
khzhu: corrected.

docs/Import-reference-genome.md
```diff
@@ -0,0 +1,47 @@
...
+4. nonN_bases: the total number of non N bases
+5. URL: the URL to download the reference genome 
+6. release_date: the date when reference genome released, in a format of yyyy-mm-dd hh:mm:ss
+
+here is a sample reference genome file:
+```
+#species	name	build_name	nonN_bases	URL	release_date
+human	hg19	GRCh37	2897310462	http://hgdownload.cse.ucsc.edu/goldenPath/hg19	2009-02-01 00:00:00
+human	hg38	GRCh38	3049315783	http://hgdownload.cse.ucsc.edu/goldenPath/hg38	2013-12-24 00:00:00
+mouse	mm10	GRCm38	2652783500	http://hgdownload.cse.ucsc.edu/goldenPath/mm10	2011-12-01 00:00:00
```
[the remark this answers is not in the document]
khzhu: you will get errors from sql insert statements if no timestamp supplied.

docs/Import-reference-genome.md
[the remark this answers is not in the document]
khzhu: also added more details in Import-reference-genome.md file.

docs/Import-reference-genome.md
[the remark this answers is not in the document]
khzhu: > 1. Do I have to do this before I load any study, or can I also load it afterward? Or is it best to do this after starting a clean database?

you only need to run importReferenceGenome scripts after you create a new database. You can  add a new reference genome anytime, just make sure your reference_genome.txt file not included those existing reference genomes in the DB.

> 2. Is cBioPortal going to use the reference genome from http://hgdownload.cse.ucsc.edu/goldenPath/* ? Will it actually download the reference genome?

genomic coordinates used by the frontend react component are downloaded form those URLs and saved in a JSON file for a runtime lookup.

> 3. What will happen if I load a study for which I do not have a reference genome?

you can not load any study if you do not have a reference genome used by that study imported to the DB first, you will get an error during loading.

> 4. Do I have to synchronize this with the value of the new property in portal.properties?

Yes, you have to update species, genome.name, and genome.build to include new reference genome.

> 5. Will it overwrite or add new reference genomes? After loading new reference genomes, are my previously loaded studies affected?

you will not be added any reference genomes if already exists. After loading new reference genome, your previously loaded studies will not affected, since cBioPortal only supports one reference genome per study. For example, if you have one study profiled previously using hg19, you would like to reanalyze it using hg38, then you will have to create a new study.

khzhu:
@sandertan, @pieterlukasse , I've commented on all your questions. Thank you, it really helps me refine the work have done!! Please hold your testing till I got updated code push back and let me know if you have any further questions.

sandertan:
@khzhu Ok, I'll continue testing when you've updated the code :)

Two more things I noticed:
1. In migration.sql you add the following data in the reference genome table, but this is **missing** in cgds.sql
```
INSERT INTO `reference_genome` 
VALUES (1, 'human', 'hg19', 'GRCh37', NULL, 'http://hgdownload.cse.ucsc.edu/goldenPath/hg19/bigZips', '2009-02-01');
INSERT INTO `reference_genome` 
VALUES (2, 'human', 'hg38', 'GRCh38', NULL, 'http://hgdownload.cse.ucsc.edu/goldenPath/hg38/bigZips', '2013-12-01');
INSERT INTO `reference_genome` 
VALUES (3, 'mouse', 'mm10', 'GRCm38', NULL, 'http://hgdownload.cse.ucsc.edu//goldenPath/mm10/bigZips', '2012-01-01');
```
2. This data is missing the genome size. Perhaps this is useful to include?

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GeneMapper.xml
khzhu: @pieterlukasse , sorry missed this one. In this PR, if you take a look at MatationMapper.xml file you will find join statements.

khzhu:
@sandertan , we do not include any insert statements in the cgds.sql file, it is purely for schema definitions. You will need to use importReferenceGenome script to input reference genome.  
I will update the migration script to have genome sizes.

sandertan:
@khzhu Ah right. So when we release an updated seed database (it's currently 2.4.0) (https://github.com/cBioPortal/datahub/tree/master/seedDB) we should make sure the reference genome table is filled.

Currently there is a seed database with mouse genes (https://github.com/cBioPortal/datahub/tree/master/seedDB_mouse), which will not be migrated correctly with this method. Mouse genes will be linked to hg19. I discussed this with @pieterlukasse and he suggested to to remove the mouse seed database with a PR for now (we can do that), and update it manually when this is necessary.

Could you add a comment above
```
INSERT INTO `reference_genome` 
VALUES (1, 'human', 'hg19', 'GRCh37', NULL, 'http://hgdownload.cse.ucsc.edu/goldenPath/hg19/bigZips', '2009-02-01');
```
to describe that this migration step is only correct if the database contains hg19 genes?

core/src/test/resources/applicationContext-dao.xml
```diff
@@ -48,8 +48,8 @@
 	<bean id="dbcpDataSource" destroy-method="close" class="org.apache.commons.dbcp2.BasicDataSource">
 		<property name="driverClassName" value="com.mysql.jdbc.Driver" />
 		<property name="url" value="jdbc:mysql://localhost:3306/cgds_test" />
-		<property name="username" value="${db.user}" />
-		<property name="password" value="${db.password}" />
+		<property name="username" value="cbio_user" />
+		<property name="password" value="somepassword" />
```
sandertan: Could you revert these back to `${db.user}` and `${db.password}`? This is installation specific.

khzhu: no problem. accidentally pushed back.

khzhu: @pieterlukasse , @sandertan , are we planning to support multiple species or just multiple genomes per portal instance?

khzhu:
```
INSERT INTO `reference_genome` 
VALUES (1, 'human', 'hg19', 'GRCh37', NULL, 'http://hgdownload.cse.ucsc.edu/goldenPath/hg19/bigZips', '2009-02-01');
``` 
will add hg19 to the reference genome table, you will also need to use the importGeneData script to populate the reference_genome_gene table for your seed database. Please reference  Updating-gene-and-gene_alias-tables.md included in this PR.

khzhu:
@pieterlukasse @sandertan  I've pushed the changes back to the origin and passed the build! Please go ahead with your side of testing. Thanks!

khzhu:
@pieterlukasse , may I have you approval or anything else I need to do? thanks!

docs/Import-reference-genome.md
```diff
@@ -1,15 +1,15 @@
 # Add new reference genome to the database
 This manual is intended to guide you to load new reference genome(s) to the database. 
 
-### What is **Reference Genome** (also know as reference assembly)?
+### What is Reference Genome (also know as reference assembly)?
```
sandertan: `What is **Reference Genome** (also know as reference assembly)?` This section is not necessary for cBioPortal. Perhaps remove it?

core/src/main/java/org/mskcc/cbio/portal/model/ReferenceGenome.java
```diff
@@ -35,20 +35,13 @@
...
-     * @param buildName         Name of genome assembly      
-     */
-    public ReferenceGenome(String genomeName, String species, String buildName) {
-        super();
-        this.genomeName = genomeName;
-        this.species = species;
-        this.buildName = buildName;
-    }
-
+    public static String HOMO_SPECIES = "human";
```
sandertan: HOMO_SAPIENS

khzhu:
@sandertan , I've added a new section in the portal.properties-Reference.md file to document the reference genome as following. 
> # Reference Genome information for importing gene and study
1. use **species** to indicate type of organismas of the reference genome, for example, **human** for homo species.
2. use **genome.name** to name the reference genome, for example **hg** for human, **mm** for mouse, etc.
3. use **genome.build** to specify different genome assemblies of the reference genome, use a comma to separate each genome version.
for example,
```
# species and genomic information
species=human
genome.name=hg
genome.build=37,GRCh37,GRCh38
```
**NOTE**
currently the portal only supports multiple reference genome builds for single species.

docs/portal.properties-Reference.md
```diff
@@ -131,3 +131,16 @@ To change the gene sets used for gene querying, create a JSON file and add gene
 ```
 querypage.setsofgenes.location=file:/<path>
 ```
+# Reference Genome information for importing gene and study
```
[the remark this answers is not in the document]
sandertan: 6. From a user perspective, it would make sense to just give one type of build name to add it to his configuration, instead of using 3 overlapping properties. Ideally something like this should be enough:
```genome.build= GRCh37; GRCh38```

Can we get the link Human <- hg19 <- GRCh37 from somewhere else?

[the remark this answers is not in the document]
Comment (khzhu) [reply docs/portal.properties-Reference.md]: @sandertan ,
thanks for the suggestions. Here is modified version. Please take  a look see if anything missed. 
>
# Reference Genome information for validating the genomic data
The reference genome information will be used by the importer validation script to ensure the reference genome version matching the reference genome version used to analyse the genomic data. 
Further details on the reference genome can be found in [this document](import-reference-genome.md); details on how to load reference genes can be found in [this document](Updating-gene-and-gene_alias-tables.md)

1. use **species** to indicate type of organisms of the reference genome, for example, **human** for homo species.
2. use **genome.name** to name the reference genome. In theory you can give any names to a reference genome. By convention, the name of reference genome always has a prefix of **hg** for human, for mouse a prefix of **mm**.
3. use **genome.build** to specify different versions of sequence database assembled by different publishers for a given reference genome. For example, 
a genome assembly published by NCBI always has a prefix of **GRCh** for homo species. 

```
# species and genomic information
species=human
genome.name=hg19,hg19,hg38
genome.build=37,GRCh37,GRCh38
```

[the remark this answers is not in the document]
Comment (khzhu) [reply docs/portal.properties-Reference.md]: @sandertan , to question 6:
the order those three properties does not matter, but one genome could have multiple builds. In the script three properties will be parsed and stored in a dictionary:
```
{'human':{'hg19':['37','GRCh37']},{'hg38':['GRCh38']}}
```

**Comment by khzhu** [comment]
@pieterlukasse @sandertan , I've pushed back the changes based on Sander' latest comments. Also,
checkNCBIbuild method enhanced now checks both species and builds (Pieter's previous comment). You can take a look at the history log if interested.
```
     def checkNCBIbuild(self, value):
-        return value in (self.portal.ncbi_build)
+        return value in (self.portal.genome_build[self.portal.species].values()[0])
```

File: docs/portal.properties-Reference.md [hunk 44]
```diff
@@ -131,3 +131,25 @@ To change the gene sets used for gene querying, create a JSON file and add gene
...
+2. After existing database is migrated to a new schema version using migration.sql (details on how to update your existing cBioportal can be found in [this document](Updating-your-cBioPortal-installation.md)) 
+3. After new database is created from a seed database that contains these reference genomes (details on how to load seed databases can be found in [this document](Import-the-Seed-Database.md)), 
+
+Below is the sample reference genome information for a human
+```
+# species and genomic information
+species=human
+genome.name=hg19,hg19,hg38
+genome.build=37,GRCh37,GRCh38
+```
```
sandertan: On second thought, @pieterlukasse and I think it's better to remove the `genome.name` and `genome.build` property. The original purpose of `genome.build` was to validate that the value in `NCBI_Build` in the MAF file corresponds with the **species** of the cBioPortal instance: Human or Mouse. It was also used to **block** `GRCh38` mutation events from being loaded into the database.

With the database solution you implemented, the valid reference genomes are now in the database. We realize that creating a web-service to provide the reference genome to the validator is out of the scope of this PR, but we think that removing this property is important. This will greatly simplify the configuration.

For now, we propose the following solution:
1. Keep the species property. This is important to keep the cBioPortal instance specific for one species at a time.
2. Remove `genome.name` and `genome.build`. 
3. Hardcode in the validator that `GRCh37` and `GRCh38` are valid `NCBI_Build` values when `species` property is `human` and `GRCm38` is valid for `mouse`. Raise an error if the value is anything else.
4. Hardcode in the validator to retrieve the correct USCS names: hg19 for GRCh37, hg38 for GRCh38, mm10 for GRCm38.

In the next pull request, the `BUILD_NAME` value from the database should be retrieved by an API call. That way,  the validator can check if the value for `NCBI_Build` in MAF corresponds to a value in the `BUILD_NAME` column in the database. Also the UCSC name could possibly be retrieved from the `NAME` column.

sandertan: I think the structure of his markdown can still be improved. I've included a sample text in [this](https://gist.github.com/sandertan/c54ca55995602c9327486ee9313cfe41) gist, assuming `genome.name` and `genome.build` are removed. This also saves a lot of explaining what and why these properties exist :) 

Additionally, could you update the table of contents at the start of this document?

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -196,8 +196,8 @@ priority_studies=
 
 # species and genomic information
 species=human
-genome.build=37,hg19,GRCh37,GRCh38
-genome.name=hg
+genome.name=hg19,hg19,hg38
+genome.build=37,GRCh37,GRCh38
```
sandertan: As proposed above, please remove the `genome.` properties.

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -46,7 +46,10 @@ After downloading, go to your downloads directory, decompress the file and add i
 ```
 **IMPORTANT NOTE**: 
 1. The **reference_genome** table needs to be populated before updating the **gene** table. Further details can be found in [this document](import-reference-genome.md). 
-2. Loading genes only to the **reference_genome_gene** table without updating the **gene** table, please use the following command:
+2. use **--species** option if you import genes from other kinds of organisms other than homo species
```
sandertan: Please update:

2. Use **--species** option when importing genes for a species other than human.

khzhu: @pieterlukasse @sandertan , thanks for your review. I will find time later today to get code modified. I totally agree with your both to have genome.name and genome.build removed from the portal.properties and using hard-code genome information in the validator for now.

khzhu:
@pieterlukasse @sandertan , I've addressed those feedback from your code-review and pushed back the changes. Thank you!

db-scripts/src/main/resources/migration.sql
```diff
@@ -409,7 +411,6 @@ CREATE TABLE `reference_genome_gene` (
     `EXONIC_LENGTH` int(11) DEFAULT NULL,
     `START` bigint(20) DEFAULT NULL,
     `END` bigint(20) DEFAULT NULL,
-    `ENSEMBL_GENE_ID` varchar(64) DEFAULT NULL,
```
pieterlukasse: all changes in 2.4.1 done here should move to 2.7.0. 
I.e. the version 2.4.1 should remain as is. This means that we need to add the following to 2.7.0:
- truncate tables reference_genome_gene and reference_genome
- drop the column `ENSEMBL_GENE_ID`
- repeat the insert steps with the fixes made 

btw: the build name GRCh37 should perhaps also not be used hard coded here, but should rather come from portal.properties. This can be achieved by using a variable here, e.g. `${buildname}`. @sandertan will add some code to `migrate_db.py` (in this PR, if you are OK with that) to ensure this `buildname` variable is correctly replaced by the correct build name during migration execution in `migrate_db.py`.

db-scripts/src/main/resources/migration.sql
```diff
@@ -452,3 +453,29 @@ ALTER TABLE gistic_to_gene DROP FOREIGN KEY gistic_to_gene_ibfk_2;
...
 UPDATE info SET DB_SCHEMA_VERSION="2.6.0";
+
+##version: 2.7.0
+-- ================== new reference genome foreign key for mutation_event and genetic_profile tables   ==================
+ALTER TABLE `mutation_event` ADD COLUMN `REFERENCE_GENOME_ID` INT(4) NULL AFTER `ENTREZ_GENE_ID`;
+ALTER TABLE `mutation_event`
+    ADD FOREIGN KEY (`REFERENCE_GENOME_ID`) REFERENCES `reference_genome_gene` (`REFERENCE_GENOME_ID`) ON DELETE CASCADE;
+
+UPDATE `mutation_event` set REFERENCE_GENOME_ID = (SELECT `REFERENCE_GENOME_ID` FROM `reference_genome` WHERE BUILD_NAME = 'GRCh37') WHERE NCBI_BUILD in ('hg19', 'GRCh37','37');
+-- UPDATE `mutation_event` set REFERENCE_GENOME_ID = (SELECT `REFERENCE_GENOME_ID` FROM `reference_genome` WHERE BUILD_NAME = 'GRCm38') WHERE NCBI_BUILD in ('mm10', 'GRCm38');
```
pieterlukasse: please enable this one as well, and include one for GRCh38 as well.

pieterlukasse: @sandertan  `migrate_db.py` should run `select distinct(NCBI_BUILD) from mutation_event`. If this results in more than 1 build (e.g. 37 and 38), then the migration should give an error telling the user to start from seed DB (i.e. remove all studies first).

db-scripts/src/main/resources/migration.sql
khzhu: @pieterlukasse @sandertan , 
- No need truncate reference_genome_gene and reference_genome tables, instead create two tables only if they do not exists
- genome.name and genome.build have been dropped from the portal.properties as you suggested. You want to add them back?

khzhu: @pieterlukasse , can you please have a separate PR for your proposal for the migrate_db.py script, thanks!

sandertan: hi @khzhu, sure I'll make a PR to your branch, with an update to `migrate_db.py` and our suggestion for `migration.sql`. We still want to remove the 2 properties! Hopefully it'll be ready tomorrow, so you can review it.

khzhu:
moved to new pull request: https://github.com/cBioPortal/cbioportal/issues/5652.

PR #3710, opened 2018-01-11
Fix validation for IGR Variant Classification

sandertan:
- Fix validation for IGR Variant Classification.
- Create new info message when IGR variant classifications are filtered out.

This issue originated from a recent feature that implemented the possibility to load variant classifications that were traditionally skipped, such as Silent or 5'Flank. Also IGR could be removed from the "Skip"-list, but IGR would still be rejected in the case there is no Entrez ID and no Symbol.

To summarize: If hugo_symbol == 'Unknown' and entrez_id == '0', the mutation will be filtered out, regardless of the settings in meta_mutation property variant_classification_filter and the SKIPPED_VARIANT_TYPES list.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1361,19 +1361,24 @@ class MutationsExtendedValidator(Validator):
         entrez_id = '0'
         if 'Entrez_Gene_Id' in self.cols:
             entrez_id = data[self.cols.index('Entrez_Gene_Id')]
-        if hugo_symbol == 'Unknown' and entrez_id == '0' and variant_classification != 'IGR':
-            # the MAF specification documents the use of Unknown and 0 here
-            # for intergenic mutations, and since the Variant_Classification
-            # column is often invalid, cBioPortal interprets this combination
-            # (or just the symbol if the Entrez column is absent) as such,
-            # but with a warning:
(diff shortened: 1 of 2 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -1360,19 +1360,25 @@ def skipValidation(self, data):
...
-            self.logger.warning(
-                "Gene specification for this mutation implies "
-                "intergenic even though Variant_Classification is "
-                "not 'IGR'; this variant will be filtered out",
-                extra={'line_number': self.line_number,
-                       'cause': "Gene symbol 'Unknown', Entrez gene id 0"})
+        if hugo_symbol == 'Unknown' and entrez_id == '0':
             is_silent = True
+            if variant_classification == 'IGR':
+                self.logger.info("cBioPortal does not support variants with Variant Classification 'IGR'. "
```
pieterlukasse: Please remove "does not support", since strictly speaking this is not completely accurate. IGR would still be loaded in a specific configuration (i.e. with custom filtering configured to not filter IGR and if entrez/hugo are set to valid value).

sandertan: Thanks, I fixed it.

sandertan:
Squashed and rebased on latest master.

PR #3735, opened 2018-01-16
Tell frontend what it will need for gene set Oncoprints

fedde-s:
I have been working to draw heatmaps of gene set scores in the Oncoprint if the user requests them via the query page fields added in cBioPortal/cbioportal-frontend/pull/742, but doing this in the style of the surrounding code needs these two modifications:
- provide the list of gene sets parsed from the query in the `serverVars` global
- provide correct unique cross-study keys for samples and patients in the web API to request gene set scores

In addition, as an added convenience that would take more work to leave out than in, the gene set score API annotates its records with their patient and study IDs in new separate JSON fields, as suggested by @ersinciftci.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
@@ -92,6 +92,8 @@ public class GenesetDataServiceImpl implements GenesetDataService {
                     GenesetMolecularData genesetData = new GenesetMolecularData();
                     genesetData.setMolecularProfileId(molecularProfileId);
                     genesetData.setSampleId(sample.getStableId());
+                    genesetData.setPatientId(sample.getPatientStableId());
+                    genesetData.setStudyId(sample.getCancerStudyIdentifier());
                     genesetData.setGenesetId(genesetAlteration.getGenesetId());
                     genesetData.setValue(genesetAlteration.getSplitValues()[indexOfSampleId]);
                     genesetDataList.add(genesetData);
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
```diff
@@ -92,6 +92,8 @@
                     GenesetMolecularData genesetData = new GenesetMolecularData();
                     genesetData.setMolecularProfileId(molecularProfileId);
                     genesetData.setSampleId(sample.getStableId());
+                    genesetData.setPatientId(sample.getPatientStableId());
```
adamabeshouse: for optimal compatibility we also need uniqueSampleKey and uniquePatientKey like in other endpoints

ersinciftci: they already exist in this endpoint, it was just not working properly because patient and study id was not being set.

adamabeshouse: oh okay

fedde-s:
The screenshot test is once again broken by pre-existing issues in the `rc` branch, and it seems unlikely to me that these changes would affect the user interface at all.

ersinciftci:
@fedde-s master is finally merged back to rc, so you will need to rebase.

fedde-s:
Sure!

fedde-s:
I rebased the commits on the current state of the `rc` branch and checked that the functionality I implemented is still there; it is.

PR #3737, opened 2018-01-16
Revert study optimizations

inodb:
- MSKCC and private portal (probably all authentication portals) were not showing right amount of studies. Oddly enough it did show up correctly locally and on cbioportal.mskcc.org/beta


ersinciftci:
Closing this for now to prevent confusion, we can re-open if we find any evidence that this is the problem.

PR #3745, opened 2018-01-17
decode URI-encoded oql query

adamabeshouse:
companion to https://github.com/cBioPortal/cbioportal-frontend/pull/899


adamabeshouse:
solving this in frontend where the variable is used

PR #3753, opened 2018-01-18
decodeURIComponent(serverVars.theQuery)

adamabeshouse:


inodb:
@adamabeshouse prolly good to update frontend commit here as well, no? I assume they have to work together. Prolly better to push a branch directly to cBioPortal/cbioportal-frontend repo (instead of your own fork) so we don't have to point the `pom.xml` to your fork.

PR #3790, opened 2018-01-25
Change error to warning for incorrect UniProtKB accession

sandertan:
# What? Why?
Currently, the validator gives an error for incorrect UniProtKB accession value in SWISSPROT column. The front-end does not break because it retrieves the UniProtKB accession and name by querying MyGeneInfo and UniProt. 
(UPDATE June 2018: I believe GenomeNexus is currently queried)

- Change error to a warning. It is still informative for the user to fix the data, although this value is currently not used.

Discussion regarding using external services or data for UniProtKB accession in https://github.com/cBioPortal/cbioportal/issues/3802


sandertan:
As discussed in #3802, this PR can go through. Other validator changes, and an update to  documentation regarding using external services or data for UniProtKB accession can be added in a future PR.

Issue #3351, opened 2017-11-13
API for NA vs WT


jjgao:
Currently, we have to check both case list and gene panel data to determine whether a gene is sequenced in a sample. Can we develop an API to combine that?

input: a list of samples, a list of genes.
output: a matrix to indicate whether a gene is sequenced in a sample.

... so that we can have a simple way to address this: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/cbioportal/Hoj3PVj4hGI/8qy865P_AQAJ

jjgao:
From @ersinciftci : "isn't http://www.cbioportal.org/api/swagger-ui.html#!/Gene_Panels/getGenePanelDataUsingPOST already doing https://github.com/cBioPortal/cbioportal/issues/3351? If it is, then do you want basically the same thing as @adama wants in https://github.com/cBioPortal/cbioportal/issues/3628?"

jjgao:
@ersinciftci Currently, we have to use three end points (Mutations, Gene_Panel, and sample list) to know whether a sample is (mutated, wildtype, or not sequenced) in a gene (based on a profile). I think we need a simpler end point to achieve that. I think that will also solve @adamabeshouse's #3628. We don't have to do exact the one I proposed above though.

ersinciftci:
@jjgao so basically an endpoint that returns a two dimensional array for each sample and gene combination, and specify the gene panel if exists, or return a keyword like "Whole Exome" if whole exome, or return something like "NA" if not sequenced?

adamabeshouse:
How about: a new projection for the samples endpoint? Query parameters: samples, genes

ersinciftci:
A samples endpoint having genes parameter doesn't sound right, because samples and genes are two independent entities in the system. This data doesn't belong to samples or genes, it's a property of their combination in a genetic profile, so I can just make `/gene-panel-data/fetch` more verbose, i.e. return data for `whole exome` and `not sequenced` combinations too.

adamabeshouse:
That's a good point

jjgao:
@ersinciftci yes.

"so basically an endpoint that returns data for each sample and gene combination, and specify the gene panel id if exists, or return a keyword like "Whole Exome" if whole exome, or return something like "NA" if not sequenced?"

jjgao:
I was also thinking of returning NA vs WT from mutations end point. Currently, mutations end point only return mut cases, but we could return NA and WT too.

[...]

PR #3793, opened 2018-01-25
Add includeNonMutated parameter to fetch endpoints in Mutations API

ersinciftci:
Default by `false`. When `true`, the endpoints will also return data for non-sequenced and wild-type combinations.
Added `sequenced` and `wildType` fields to the response type.
Fixes https://github.com/cBioPortal/cbioportal/issues/3351

### DIFF
--- a/model/src/main/java/org/cbioportal/model/Mutation.java
+++ b/model/src/main/java/org/cbioportal/model/Mutation.java
@@ -34,6 +34,8 @@ public class Mutation extends Alteration implements Serializable {
     private String driverFilterAnnotation;
     private String driverTiersFilter;
     private String driverTiersFilterAnnotation;
+    private Boolean sequenced;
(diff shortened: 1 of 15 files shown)


adamabeshouse:
includeNonMutated is false by default, right? for backwards compatibility

ersinciftci:
Yes

jjgao:
@adamabeshouse : is this new option being used in the frontend?

adamabeshouse:
@jjgao anything in particular you're referring to?

jjgao:
@adamabeshouse I was thinking maybe oncoprint can use it? But if everything is already implemented in the frontend, it's not essential to change for now.

adamabeshouse:
@jjgao I think it will be useful for the download issue you've mentioned here: https://github.com/cBioPortal/cbioportal/issues/3972

PR #3844, opened 2018-02-02
Precalculate fraction genome altered and add to a table

ersinciftci:


core/src/main/java/org/mskcc/cbio/portal/dao/DaoCopyNumberSegment.java
```diff
@@ -63,6 +63,19 @@ public static int addCopyNumberSegment(CopyNumberSegment seg) throws DaoExceptio
...
     }
+
+    public static int addFractionGenomeAltered(int cancerStudyId, int sampleId) throws DaoException {
+        if (!MySQLbulkLoader.isBulkLoad()) {
+            throw new DaoException("You have to turn on MySQLbulkLoader in order to insert fraction genome altered");
+        } else {
+            MySQLbulkLoader.getMySQLbulkLoader("fraction_genome_altered").insertRecord(
+                    Integer.toString(cancerStudyId),
+                    Integer.toString(sampleId),
+                    Double.toString(getCopyNumberActeredFraction(sampleId, cancerStudyId, GlobalProperties.getPatientViewGenomicOverviewCnaCutoff()[0]))
```
jjgao: Should we just hard code this cutoff?

db-scripts/src/main/resources/migration.sql
```diff
@@ -426,3 +426,16 @@ FROM `gene`);
...
+
+CREATE TABLE `fraction_genome_altered` (
+  `CANCER_STUDY_ID` int(11) NOT NULL,
+  `SAMPLE_ID` int(11) NOT NULL,
+  `VALUE` double NOT NULL,
+  KEY (`CANCER_STUDY_ID`,`SAMPLE_ID`),
+  FOREIGN KEY (`CANCER_STUDY_ID`) REFERENCES `cancer_study` (`CANCER_STUDY_ID`) ON DELETE CASCADE,
+  FOREIGN KEY (`SAMPLE_ID`) REFERENCES `sample` (`INTERNAL_ID`) ON DELETE CASCADE
+);
+
```
jjgao: We will need migration code to fill the data for existing samples.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCopyNumberSegment.java
```diff
@@ -63,6 +63,19 @@ public static int addCopyNumberSegment(CopyNumberSegment seg) throws DaoExceptio
             return 1;
         }
     }
+
+    public static int addFractionGenomeAltered(int cancerStudyId, int sampleId) throws DaoException {
```
jjgao: Instead of doing it one sample at a time maybe we can do it for the whole study (and later for the whole copy number segment profile once we have that). similar to what we did for mutation_count: https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java#L133

db-scripts/src/main/resources/migration.sql
ersinciftci: I'll see if I can convert it to a SQL-only calculation (it's most likely), if not, we might need to calculate the values for every study and then export insert statements of values (a long list of inserts).

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCopyNumberSegment.java
ersinciftci: Once I convert it to a SQL-only solution, I think I can do it with just one insert query per study.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCopyNumberSegment.java
ersinciftci: I can, if we are going to delete (or not use in the new code) that portal property eventually. Should I also remove the cutoff parameter from the new API?

jjgao: Let's remove that portal property.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -129,8 +129,6 @@
             "TCGA working groups directly.";
 
     public static final String PATIENT_VIEW_PLACEHOLDER = "patient_view_placeholder";
-    public static final String PATIENT_VIEW_GENOMIC_OVERVIEW_CNA_CUTOFF = "patient_view_genomic_overview_cna_cutoff";
-    public static final double[] DEFAULT_GENOMIC_OVERVIEW_CNA_CUTOFF = new double[]{0.2,1.5};
```
jjgao: Maybe also remove it from the example properties file? 1.5 could also be hard coded in the frontend. -- or mabye it's another pr?

ersinciftci: OK removed, and it looks like it's hardcoded in the frontend https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/genomicOverview/tracksHelper.ts#L19

PR #3856, opened 2018-02-07
NormalizeExpressionLevels parse header fix.

angelicaochoa:
# What? Why?
If `Entrez_Gene_Id` column is missing then `NormalizeExpressionLevels` will interpret the first column of data in the copy number file (`data_CNA.txt`) as an entrez gene id since data values will match the regex `[0-9]+`.

- Determine indices for `Entrez_Gene_Id` and `Hugo_Symbol`. Use entrez gene id by default if column is present in header. Otherwise use the hugo symbol. 
- Throw fatal error if neither column is present.

# Checks
- [ ] Runs on Heroku.
- [x] Make sure your commit messages end with a Signed-off-by string (this line can be automatically added by git if you run the `git-commit` command with the `-s` option)
- [x] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to hotfix.

# Any screenshots or GIFs?

# Notify reviewers
@cBioPortal/backend 

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/NormalizeExpressionLevels.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/NormalizeExpressionLevels.java
@@ -90,6 +90,8 @@ import org.mskcc.cbio.portal.util.SpringUtil;
 public class NormalizeExpressionLevels{
 
    public static final String TCGA_NORMAL_SUFFIX = "-11";
+   private static final String ENTREZ_GENE_ID_COLUMN = "Entrez_Gene_Id";
+   private static final String HUGO_SYMBOL_COLUMN = "Hugo_Symbol";
 
    static HashMap<Long, ArrayList<String[]>> geneCopyNumberStatus = null;
    static int SAMPLES;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/NormalizeExpressionLevels.java
```diff
@@ -218,10 +225,11 @@ private static void computeZScoreXP(String file){
             values = line.split("\t");
                     
             CanonicalGene gene;
-            if (values[1].matches("[0-9]+")) {
-                gene = daoGeneOptimized.getGene(Long.parseLong(values[1]));
-            } else {
-                gene = daoGeneOptimized.getNonAmbiguousGene(values[0]);
+            if (entrezGeneIdIndex != -1) {
```
sheridancbio: here we prefer to use entrezGeneId if the column is available (over hugo gene symbol).
We could consider using hugo gene symbol if a record has a blank in entrezGeneId ... but we probably expect all entrezGeneId values to be populated if the column is present.

n1zea144: This behavior is consistent in other places of the codebase, like [ImportTabDelimData.java](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java).

PR #3869, opened 2018-02-09
Fix Study View Query button to Main Page issue

zhx828:
Include selected_study_variables.jsp in index.jsp due to new structure
changes introduced in 1.11


core/src/main/java/org/mskcc/cbio/portal/servlet/QueryBuilder.java
```diff
@@ -326,6 +326,13 @@ protected void doPost(HttpServletRequest httpServletRequest,
                     if (errorsExist) {
                         httpServletRequest.setAttribute(QueryBuilder.USER_ERROR_MESSAGE, "Please fix the errors below.");
                     }
+                    if(cancerStudyId.equals("all") && _isVirtualStudy && cohortDetails != null) {
```
kalletlak: I would just check for `cohortDetails != nll`

portal/src/main/webapp/WEB-INF/jsp/global/selected_study_variables.jsp
```diff
@@ -21,9 +21,8 @@
 %>
 <script type="text/javascript">
 
-    var isVirtualStudy = <%=isVirtualStudy%>;
-    var cancerStudyIdList = '<%=cancerStudyIdListStr%>'; // empty string if single study
-    var cancerStudyId = '<%=cancerStudyId%>'; // if multi-studies, this is always "all"
-    window.cohortIdsList = (cancerStudyIdList === 'null')? [cancerStudyId]: cancerStudyIdList.split(',');
+    window.cancerStudyIdList = '<%=cancerStudyIdListStr%>'; // empty string if single study
```
alisman: Is this file used on multiple pages?  We just need to make sure this doesn't break them.  We had a bug recently having to do this this.

PR #3879, opened 2018-02-13
Script for validation of multiple studies and updates to the validator for stricter MAF format checks

dionnezaal:
# What? Why?
A script that is able to run the validator for multiple studies. Needed for CircleCI implementation in `cBioPortal/datahub`. The script will return the validation status of each study in the console.  Also, an overall exit status of 0 or 1 is returned, where 1 indicates that at least one of the input studies has failed validation. Documentation was added to `docs/Using-the-dataset-validator.md` about how the script can be run. Furthermore, a system test was added which checks all possible exit statuses.

Later also the following validator checks were implemented:
-  Checking if SEG file is more than once in the study folder, can only be loaded once.
- Check for duplicate samples in data_gene_panel_matrix.txt [(datahub issue #166)](https://github.com/cBioPortal/datahub/issues/166)
- Check for duplicate case list category
- Checks for mutation files: [MAF file checks](https://wiki.nci.nih.gov/display/TCGA/Mutation+Annotation+Format+(MAF)+Specification) and checks based on [test output maf files from vcf2maf](https://github.com/mskcc/vcf2maf/tree/master/tests).

An extra parameter in `validateData.py` was also implemented for the last mutation file checks added: `-m`. This parameter enforces `strict_mode` which will result in errors instead of warnings for these mutation file checks.


docs/Using-the-dataset-validator.md
```diff
@@ -414,3 +414,78 @@ As an example, the command for the "Example 1" listed above incorporating the `-
...
+usage: validateStudies.py [-h] [-d ROOT_DIRECTORY] [-l LIST_OF_STUDIES]
+                          [-html HTML_FOLDER]
+                          [-u URL_SERVER | -p PORTAL_INFO_DIR | -n]
+                          [-P PORTAL_PROPERTIES]
+
+Wrapper where cBioPortal study validator is run for (multiple) input studies
+
+optional arguments:
+-h, --help            show this help message and exit
+-d ROOT_DIRECTORY, --root-directory ROOT_DIRECTORY
```
sandertan: Perhaps rename `root directory` to `parent directory`? I'm not sure. Root is usually the highest level in a hierarchy. What do you and @pieterlukasse think?

docs/Using-the-dataset-validator.md
```diff
@@ -414,3 +414,78 @@ As an example, the command for the "Example 1" listed above incorporating the `-
...
+```
+
+### Example 3
+The script can also be run on specific studies in a certain directory by combining the `--root-directory` and `--list-of-studies` parameter:
+```bash
+./validateStudies.py -d ../../../test/scripts/test_data/ -l study_es_0,study_es_1
+```
+
+### Example 4
+When HTML validation reports are desired for each of the studies validated an output folder for these HTML files can be specified. This folder does not have to exist, the script can create the folder. The HTML validation reports will get the following name: `<study_name>-validation.html`. To create HTML validation reports for each study the `--html-folder` parameter needs to be defined:
```
sandertan: Please make it a bit shorter:

"When HTML validation reports are desired, an output folder for these HTML files can be specified."

core/src/test/scripts/system_tests_validate_studies.py
```diff
@@ -0,0 +1,73 @@
...
+
+import unittest
+from importer import validateStudies
+
+
+# globals:
+PORTAL_INFO_DIR = 'test_data/api_json_system_tests'
+
+class ValidateStudiesSystemTester(unittest.TestCase):
+    '''Test cases around running the complete validateData script
```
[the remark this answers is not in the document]
sandertan: Could you replace all `'''` with double quotes `"""`?

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -0,0 +1,154 @@
...
+            try:
+                with open(os.path.join(study, "meta_study.txt"), 'r') as meta_study:
+                    for line in meta_study:
+                        if 'cancer_study_identifier' in line:
+                            study_identifier = line.split(': ')[1].strip('\n').strip('\r')
+                            html = study_identifier + "-validation.html"
+                html_file = os.path.join(output_folder, html)
+                validator_args.append('--html')
+                validator_args.append(html_file)
+            except:
```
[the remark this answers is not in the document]
dionnezaal: There can be indeed multiple errors: Supplied study folder does not exist, no meta_study file or no cancer_study_identifier in the HTML file. The warning message is indeed not clear, but the traceback of the error that occurred is printed after it. By adding the traceback I hoped that the warning will be clear to the user what went wrong. What do you think @sandertan? I will add to the warning message that the validation will not be written to a HTML file.

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -0,0 +1,154 @@
...
+            validator_logger = logging.getLogger(validateData.__name__)
+            validator_logger.handlers = []
+
+    return validation_exit_status
+
+
+def interface(args=None):
+    """Parse arguments from input"""
+
+    parser = argparse.ArgumentParser(description='Wrapper where cBioPortal study validator is run for (multiple) input studies')
```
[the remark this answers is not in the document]
sandertan: Sorry I meant to post this URL with the comment on top of this one.

core/src/main/scripts/importer/validateStudies.py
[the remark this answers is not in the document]
sandertan: " I will add to the warning message that the validation will not be written to a HTML file." with the trackback should indeed be sufficient!

dionnezaal:
Two questions from review @sandertan remain:
- Should we add a port to default from url_server? [This comment](https://github.com/cBioPortal/cbioportal/pull/3879/files/4b3fc01476e4edba44eae1462d78c9614b155a07#r167855887)
- Should we rename the parameter `root-directory` to `parent-directory`?

@pieterlukasse what do you think?

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -0,0 +1,169 @@
...
+import validateData
+
+
+def main(args):
+    """Process arguments and run validation"""
+
+    # Set up logger, stdout from validation will be written to a log file
+    logging.basicConfig(level=logging.DEBUG,
+                        filename='log-validate-studies.txt',
+                        filemode='w')
```
sandertan: Let's not write the log file, and throw away the standard out. I checked the log files and the contents are not very useful compared to what is in HTML reports.

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -0,0 +1,154 @@
...
+    parser.add_argument('-html', '--html-folder',
+                        type=str, help='Path to folder for output HTML reports')
+    # Only one of the portal arguments can be given, therefore set as mutually exclusive group
+    portal_mode_group = parser.add_mutually_exclusive_group()
+    portal_mode_group.add_argument('-u', '--url_server',
+                                   type=str,
+                                   default='http://localhost/cbioportal',
+                                   help='URL to cBioPortal server. You can '
+                                        'set this if your URL is not '
+                                        'http://localhost/cbioportal')
```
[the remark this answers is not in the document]
pieterlukasse: yes, `8080` is better in the tomcat context

docs/Using-the-dataset-validator.md
```diff
@@ -414,3 +414,78 @@ As an example, the command for the "Example 1" listed above incorporating the `-
 ```
 ./validateData.py -s ../../../test/scripts/test_data/study_es_0/ -P ../../../../../src/main/resources/portal.properties -u http://localhost:8080/cbioportal -v
 ```
+
+## Running the validator for multiple studies
```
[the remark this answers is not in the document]
pieterlukasse: toc :+1:

core/src/main/scripts/importer/validateStudies.py
pieterlukasse: please write log file to temp dir

dionnezaal: After discussion with @pieterlukasse we decided to keep the log file, but that it has to be written to the HTML-folder (if provided) otherwise to a created tmp folder.

dionnezaal:
Port was added to default for url_server input parameter in validateData and validateStudies. 
Directory name was kept `root-directory`.

docs/Using-the-dataset-validator.md
```diff
@@ -1,3 +1,11 @@
+* [Introduction](##introduction)
+* [Running the validator](##running-the-validator)
+* [Offline Validation](##offline-validation)
+* [Validation of non-human data](##validation-of-non-human-data)
+* [Running the validator for multiple studies](##running-the-validator-for-multiple-studies)
```
sandertan: Please change `##` to a single `#`, else it doesn't work.

dionnezaal: Fixed :)

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -0,0 +1,185 @@
...
+        parser.error('No action requested, add -d or -l')
+
+    return args
+
+
+if __name__ == '__main__':
+    parsed_args = interface()
+    exit_status = main(parsed_args)
+
+    print "\n\nOverall exit status: %s" %exit_status
```
pieterlukasse: you still need to use this to let the script exit with 1 if there is any error. I.e. add:
```
sys.exit(exit_status)
```

docs/Using-the-dataset-validator.md
```diff
@@ -414,3 +422,78 @@ As an example, the command for the "Example 1" listed above incorporating the `-
...
+-d ROOT_DIRECTORY, --root-directory ROOT_DIRECTORY
+                      Path to directory with all studies that should be
+                      validated
+-l LIST_OF_STUDIES, --list-of-studies LIST_OF_STUDIES
+                      List with paths of studies which should be validated
+-html HTML_FOLDER, --html-folder HTML_FOLDER
+                      Path to folder for output HTML reports
+-u URL_SERVER, --url_server URL_SERVER
+                      URL to cBioPortal server. You can set this if your URL
+                      is not http://localhost/cbioportal
```
sandertan: Found 2 more!

Could you update these as well with our http://localhost:8080/cbioportal update?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1228,6 +1232,409 @@ def checkLine(self, data):
...
+                    self.logger.warning(log_message, extra={'line_number': self.line_number,
+                                                            'cause': '(%s, %s, %s)' % (ref_allele, tumor_seq_allele1,
+                                                                                       tumor_seq_allele2)})
+                else:
+                    self.logger.error(log_message, extra={'line_number': self.line_number,
+                                                          'cause': '(%s, %s, %s)' % (ref_allele, tumor_seq_allele1, 
+                                                                                     tumor_seq_allele2)})
+
+            # Check in case of insertion if the Reference_Allele contains '-'
+            if variant_type == "INS" and '-' not in ref_allele:
```
sandertan: I'm not sure if this test is correct. Does_ the reference always have to be `-` in case of an insertion? What about A -> GT ?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1228,6 +1232,409 @@ def checkLine(self, data):
...
+                log_message = "Variant_Type indicates an insertion, but Reference_Allele does not equal -."
+                if not self.strict_mode:
+                    self.logger.warning(log_message, extra={'line_number': self.line_number, 'cause': ref_allele})
+                else:
+                    self.logger.error(log_message, extra={'line_number': self.line_number, 'cause': ref_allele})
+
+            # In case of deletion, check when Reference_Allele is the same length as both Tumor_Seq_Allele if at least
+            # one of the Tumor_Seq_Alleles is a deletion ('-') otherwise a SNP
+            if variant_type == "DEL" and len(ref_allele) == len(tumor_seq_allele1) \
```
sandertan: Similar question as above. What about GT -> A ?

core/src/main/scripts/importer/validateData.py
```diff
@@ -1228,6 +1232,409 @@ def checkLine(self, data):
...
+                    self.logger.warning(log_message, extra={'line_number': self.line_number,
+                                                            'cause': '(%s, %s, %s)' % (ref_allele, tumor_seq_allele1, 
+                                                                                       tumor_seq_allele2)})
+                else:
+                    self.logger.error(log_message, extra={'line_number': self.line_number, 
+                                                          'cause': '(%s, %s, %s)' % (ref_allele, tumor_seq_allele1, 
+                                                                                     tumor_seq_allele2)})
+
+            # In case of ONP there is checked if the Allele based columns are bigger than 3 by the MAF file checks
+            # but not if they are all the same length.
```
sandertan: This should be part of Maf file check 11f, right?

sandertan: If you'd like to keep this separate, please improve the grammar in this comment.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2367,6 +2774,27 @@ class GenePanelMatrixValidator(Validator):
...
+        sample_id = data[self.cols.index('SAMPLE_ID')]
+        if sample_id in self.gene_panel_sample_ids:
+            self.logger.error(
+                'Duplicated Sample ID.',
+                extra={'line_number': self.line_number,
+                'cause': '%s (already defined on line %d)'
+                         % (sample_id,
+                            self.gene_panel_sample_ids[sample_id])})
+        else:
+            self.gene_panel_sample_ids[sample_id] = self.line_number
```
sandertan: Nice. Short and effective check.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3025,6 +3454,15 @@ def process_metadata_files(directory, portal_instance, logger, relaxed_mode):
...
             else:
                 stable_ids.append(stable_id)
+        # Validate datatype, for segmentation file only one can be loaded
+        if 'datatype' in meta_dictionary:
+            if meta_dictionary['datatype'].lower() == "seg":
+                if study_data_type is not None:
+                    logger.error(
+                        'datatype SEG is repeated. Only one segmentation file can be loaded.')
+                else:
+                    study_data_type = meta_dictionary['datatype']
```
sandertan: `study_data_type` is not a proper name to check "presence of segment datafile". 

What about making a list, like the stable_ids list, such as study_data_types.
```
study_data_types = []
if 'datatype' in meta_dictionary:
    if meta_dictionary['datatype'].lower() in study_data_types:
        if meta_dictionary['datatype'].lower() == "seg":
            error......
    else:
         study_data_types.append(meta_dictionary['datatype'].lower())

File: core/src/main/scripts/importer/validateData.py [hunk 17]
```diff
@@ -3632,6 +4075,9 @@ def main_validate(args):
     relaxed_mode = False
     if hasattr(args, 'relaxed_clinical_definitions') and args.relaxed_clinical_definitions:
         relaxed_mode = True
+    strict_mode = False
+    if hasattr(args, 'strict_mutation_checks') and args.strict_mutation_checks:
+        strict_mode = True
```
sandertan: I'm more of a fan of using `default=False` in the parser.add_argument, and then using `strict_mode = args.strict_mutation_checks`

core/src/main/scripts/importer/validateData.py
dionnezaal: You indeed may be correct: 
A -> GT will be checked correct by the following check: 
`If Variant_Type == "INS", then (End_position - Start_position + 1 == length (Reference_Allele) or End_position - Start_position == 1) and length(Reference_Allele) <= length(Tumor_Seq_Allele1 and Tumor_Seq_Allele2)`

And the MAF file information says that the flanking reference allele cannot be in there, but says nothing about when it is a different one.
screen shot 2018-03-13 at 11 15 04 am

core/src/main/scripts/importer/validateData.py
dionnezaal: In this case, it is first checking if the reference and tumor seq alleles are the same sizes. In case they are, there should be '-' in the tumor seq alleles, otherwise it indicates an SNP.

(check if length reference allele is bigger than tumor seq alleles is done in MAF file check 11)

core/src/main/scripts/importer/validateData.py
dionnezaal: I misread MAF file check 11F, it is indeed in there, i will move it there.

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -84,14 +84,15 @@ class DataFileTestCase(LogBufferTestCase):
     particular validator class and collect the log records emitted.
     """
 
-    def validate(self, data_filename, validator_class, extra_meta_fields=None, relaxed_mode=False):
+    def validate(self, data_filename, validator_class, extra_meta_fields=None, relaxed_mode=False, strict_mode=False):
```
pieterlukasse: can we give this a specific name, like `strict_maf_checks`?

dionnezaal: Done :)

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -529,6 +529,21 @@ def test_both_name_and_entrez_but_invalid_hugo(self):
         self.assertEqual(record_list[0].cause, 'XXACAP3')
         self.assertEqual(record_list[1].cause, 'XXAGRN')
 
+    def test_both_name_and_entrez_but_hugo_starts_with_integer(self):
+        """Test when a file has both the Hugo name and Entrez ID columns,
+        but hugo is invalid because starts with integer."""
+        self.logger.setLevel(logging.ERROR)
+        record_list = self.validate('data_cna_genecol_presence_both_invalid_hugo_integer.txt',
+                                    validateData.CNAValidator,
+                                    extra_meta_fields={'meta_file_type': 'test_CNA'})
```
sandertan: Can you use a real possible value of extra_meta_fields, perhaps just `CNA`.

core/src/test/scripts/test_data/data_cna_genecol_presence_both_invalid_hugo_integer.txt
```diff
@@ -0,0 +1,8 @@
+Hugo_Symbol	Entrez_Gene_Id	TCGA-A1-A0SB-01	TCGA-A1-A0SD-01	TCGA-A1-A0SE-01	TCGA-A1-A0SH-01	TCGA-A2-A04U-01
+1-ACAP3	116983	0	-1	0	-1	2
+2-AGRN	375790	0	-1	0	-1	2
+7SK	55210	0	-1	0	-1	2
+7SK|test	83858	0	-1	0	-1	2
+ATAD3C	219293	0	-1	0	-1	2
```
sandertan: Excellent to have the alias 7SK in the test data!

core/src/main/scripts/importer/validateData.py
```diff
@@ -643,6 +643,33 @@ def checkGeneIdentification(self, gene_symbol=None, entrez_id=None):
         # set to upper, as both maps contain symbols in upper
         if gene_symbol is not None:
             gene_symbol = gene_symbol.upper()
+            # Check in case gene symbol is not null if it starts with an integer
+            if gene_symbol is not '':
+                # In case of CNA data the gene_symbol possibly needs to be splitted
+                if 'meta_file_type' in self.meta_dict.keys():
```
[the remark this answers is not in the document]
dionnezaal: With complete validation this is indeed the case. The tests that are run from unit_tests_validate_data.py do not have this and therefore they will give an error when running.

core/src/main/scripts/importer/validateData.py
```diff
@@ -643,6 +643,33 @@ def checkGeneIdentification(self, gene_symbol=None, entrez_id=None):
         # set to upper, as both maps contain symbols in upper
         if gene_symbol is not None:
             gene_symbol = gene_symbol.upper()
+            # Check in case gene symbol is not null if it starts with an integer
+            if gene_symbol is not '':
```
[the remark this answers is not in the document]
sandertan: We decided to keep it in, since validation for other file types do not set empty values to `none`, and it is out of scope for this PR to refactor that.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
pieterlukasse: We should not adjust the code because of constraints in our test set. Rather adjust the test data / unit tests in this case.

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -0,0 +1,154 @@
...
+            print var_traceback + '\x1b[30m'
+            validation_exit_status = 1  # Here also set invalid validation exit status for failing circleCI
+            # Reset logger handlers to suppress stdout output, will be written to log file
+            validator_logger = logging.getLogger(validateData.__name__)
+            validator_logger.handlers = []
+
+    return validation_exit_status
+
+
+def interface(args=None):
```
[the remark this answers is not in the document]
pieterlukasse: @dionnezaal is this done?

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
pieterlukasse: Try something like 
```python
            if type(self) is CNAValidator:
                ...
```
but also move the logic to the `CNAValidator` layer or the one closest to it that implements  `parseFeatureColumns`  (which is `GenewiseFileValidator` in this case)

dionnezaal:
@pieterlukasse Have moved CNA split logic to GenewiseValidator as you indicated. 
Also squashed the commits and are rebased on latest RC

dionnezaal:
Rebased on RC

PR #3880, opened 2018-02-13
readOnly Mybatis read-only second-level cache-redux

n1zea144:
Updates to allow read-only MyBatis second-level cache. One further optimization is to bypass copy of references returned from MyBatis layer in the service layer before returning to controller when authentication is turned off (no @POST_FILTER call) since it is not necessary.  Also, I intentionally did not address legacy MyBatis code.

Note I looked into extending/customizing spring-security@PostFilter annotation or using aop to implement the copy of references returning from the MyBatis layer and keep the service layer "clean", but I couldn't find a solution that didn't introduce more complexity than what is contained in this PR.

### DIFF
--- a/business/src/main/resources/applicationContext-business.xml
+++ b/business/src/main/resources/applicationContext-business.xml
@@ -76,11 +76,14 @@
             <property name="dataSource" ref="businessDataSource" />
         </bean>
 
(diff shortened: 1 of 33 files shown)


n1zea144:
Yeah, thats what I'm afraid of.  I'm going to do a survey of the current *ServiceImpl classes and take a further look.  One idea is to make the lists returned from the persistence layer immutable and see if any exceptions are thrown.  This won't catch any individual objects that may be modified in the service layer though.

Issue #3102, opened 2017-09-13
Deletion of Cancer Study via Foreign Key Cascade failing


sheridancbio:
`org.mskcc.cbio.portal.dao.deleteCancerStudyByCascade(int internalCancerStudyId)` relies on the "ON DELETE CASCADE" mode of foreign key constraints in database tables to remove all associated records in related tables when a records is deleted from the cancer_study table.

A foreign key constraint exists in database table "gistic" which will trigger deletion (via cascade) of any records in gistic which are linked through a foreign key constraint to table cancer_study (on field CANCER_STUDY_ID) when the cancer_study linked record is deleted.

A foreign key constraint also exists in database table "gistic_to_gene" which requires each record to have a corresponding parent record in table "gistic". However, this foreign key constraint does not specify the `ON DELETE CASCADE `clause and so this constraint would be violated if a linked record in gistic were to be deleted (linked through field GISTIC_ROI_ID).

Because the deletion of the cancer_study record will trigger the deletion of associated gistic records, but any associated gistic_to_gene records will prevent deletion of the gistic records, the call to deleteCancerStudyByCascade() will fail any time there are associated gistic_to_gene records for that cancer_study and a foreign key enabled database engine is in use (such as InnoDB). The observed exception was:
Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Cannot delete or update a parent row: a foreign key constraint fails (`cgds_public`.`gistic_to_gene`, CONSTRAINT `gistic_to_gene_ibfk_2` FOREIGN KEY (`GISTIC_ROI_ID`) REFERENCES `gistic` (`GISTIC_ROI_ID`))

Proposed fix: add the `ON DELETE CASCADE` clause to both foreign key constraints in table gistic_to_gene (the second foreign key constraint references table gene, which is not related to this issue, but a similar problem would occur if gene records were to be deleted)

This issue was reported by @angelicaochoa 
@cBioPortal/backend

sheridancbio:
This has been addressed by #3894, and will no longer be an issue after merge/deployment to master.

PR #3894, opened 2018-02-15
GISTIC_TO_GENE FKC fix

angelicaochoa:
# What? Why?
`cancer_study` delete cascade does not propagate to `gistic_to_gene` because `gistic_to_gene` is missing the `DELETE ON CASCADE` reference to FKC on `gistic` 

- Modify `gistic_to_gene` FKC on `gistic` to include `ON DELETE CASCADE`

# Checks
- [ ] Runs on Heroku.
- [x] Make sure your commit messages end with a Signed-off-by string (this line can be automatically added by git if you run the `git-commit` command with the `-s` option)
- [x] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to hotfix.

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/db-scripts/src/main/resources/migration.sql
+++ b/db-scripts/src/main/resources/migration.sql
@@ -445,3 +445,10 @@ ABS(c2.`SEGMENT_MEAN`) >= 0.2) / SUM(`END`-`START`)) AS `VALUE` FROM `copy_numbe
 c1.`CANCER_STUDY_ID` = cancer_study.`CANCER_STUDY_ID` GROUP BY cancer_study.`CANCER_STUDY_ID`, `SAMPLE_ID` HAVING SUM(`END`-`START`) > 0;
 
 UPDATE info SET DB_SCHEMA_VERSION="2.5.0";
+
+##version 2.6.0
+-- modify fkc for gistic_to_gene
+ALTER TABLE gistic_to_gene DROP FOREIGN KEY gistic_to_gene_ibfk_2;
+ALTER TABLE gistic_to_gene ADD CONSTRAINT `gistic_to_gene_ibfk_2` FOREIGN KEY (`GISTIC_ROI_ID`) REFERENCES `gistic` (`GISTIC_ROI_ID`) ON DELETE CASCADE;
+
+UPDATE info SET DB_SCHEMA_VERSION="2.6.0";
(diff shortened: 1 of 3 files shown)


db-scripts/src/main/resources/migration.sql
```diff
@@ -444,4 +444,11 @@ WHERE c2.`CANCER_STUDY_ID` = c1.`CANCER_STUDY_ID` AND c2.`SAMPLE_ID` = c1.`SAMPL
 ABS(c2.`SEGMENT_MEAN`) >= 0.2) / SUM(`END`-`START`)) AS `VALUE` FROM `copy_number_seg` AS c1, cancer_study WHERE 
 c1.`CANCER_STUDY_ID` = cancer_study.`CANCER_STUDY_ID` GROUP BY cancer_study.`CANCER_STUDY_ID`, `SAMPLE_ID` HAVING SUM(`END`-`START`) > 0;
 
-UPDATE info SET DB_SCHEMA_VERSION="2.5.0";
\ No newline at end of file
+UPDATE info SET DB_SCHEMA_VERSION="2.5.0";
+
+##version 2.6.0
+-- modify fkc for gistic_to_gene
+ALTER TABLE gistic_to_gene DROP FOREIGN KEY gistic_to_gene_ibfk_2;
```
sheridancbio: Do we know that the name for the foreign key *gistic_to_gene_ibfk_2* will be found on all database installations? I think that foreign key constraint operations are silently ignored on MyISAM engine databases ... and perhaps this foreign key naming pattern is canonical across all INNODB engine installations. But I would double check this before we commit to this schema update (since schema version changes are immutable once committed to).
If we are wrong about this, we may need to create documentation for installations on systems where the foreign key name differs from this pattern ... helping them to manually adjust the foreign key without the migration.sql script.

Maybe (in a different PR) we would want to assign chosen names to all foreign key constraints to allow future deletions / modifications. But maybe this is not possible if the FKC name is set implicitly.

sheridancbio:
Also for any db_schema_change, I would recommend that we get PR review from other institutes. (@pieterlukasse, @ersinciftci)

sheridancbio:
There is an issue that this will resolve : #3102

db-scripts/src/main/resources/migration.sql
angelicaochoa: @sheridancbio yes, the name of the constraint is `gistic_to_gene_ibfk_2` - the constraint wasn't actually missing but it was missing the `DELETE ON CASCADE`

pieterlukasse: I hope no one is still on myisam...?

PR #3907, opened 2018-02-21
Addresses use of ArrayList.removeIf which is no longer allowed with m…

n1zea144:
Addresses use of ArrayList.removeIf which is no longer allowed with mybatis read-only cache.


n1zea144:
@ersinciftci I considered allowing this list modification, but because the query does not filter based on alteration type, this cannot be done as is.

PR #2889, opened 2017-07-31
Validator: Move attributes from sample to patient level

sandertan:
# What? Why?
Fix https://github.com/cBioPortal/datahub/issues/51

This PR moves the following attributes in the Validator from sample to patient level:
- SERUM_PSA
- TUMOR_GRADE
- HISTOLOGY
- TUMOR_SITE
- KNOWN_MOLECULAR_CLASSIFIER

PR #3910, opened 2018-02-23
Modify validation of clinical attributes

sandertan:
# What? Why?
This PR removes validation to be either a patient or sample attribute for many attributes but leaves a **core** set of patient attributes.

Fix https://github.com/cBioPortal/datahub/issues/51.
Previous PR that was only moving some attributes from sample to patient level #2889.

A later PR will update the documentation of these and non-core attributes in File-Formats.

- For many attributes, remove validation of being in sample or patient data
- Move `CANCER_TYPE`, `CANCER_TYPE_DETAILED`, `HISTOLOGY` and `KNOWN_MOLECULAR_CLASSIFIER` from sample to patient level
- For a core set of attributes, set an ERROR if in incorrect data file
- Remove deprecated attributes, like `TYPE_OF_CANCER` 
- Added new attributes like `MOUSE_STRAIN`

### Core set of patient attributes
- `AGE`
- `CANCER_TYPE`
- `CANCER_TYPE_DETAILED`
- `DFS_STATUS`
- `DFS_MONTHS`
- `GENDER`
- `HISTOLOGY`
- `KNOWN_MOLECULAR_CLASSIFIER`
- `OS_STATUS`
- `OS_MONTHS`
- `PATIENT_DISPLAY_NAME`
- `SEX`

### Core set of sample attributes
- `SAMPLE_ID`
- `OTHER_SAMPLE_ID`
- `SAMPLE_DISPLAY_NAME`
- `TUMOR_TISSUE_SITE`
- `TUMOR_TYPE`

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1541,27 +1541,23 @@ class ClinicalValidator(Validator):
                       'datatype',
                       'priority')
 
-    # Attributes required to have certain properties because of hard-coded use.
-    # Note: the 'when_wrong' property (found in some attributes like METASTATIC_SITE),
-    # can be set to WARNING to indicate that only a WARNING should be given
-    # if this attribute is found in the "wrong" file (e.g. a PATIENT attribute found
-    # in a SAMPLE file or vice-versa).
+    # Only a core set of attributes must be either specific in the patient or sample clinical data.
     PREDEFINED_ATTRIBUTES = {
         'AGE': {
             'is_patient_attribute': '1',
             'datatype': 'NUMBER'
         },
         'CANCER_TYPE': {
-            'is_patient_attribute': '0',
+            'is_patient_attribute': '1',
             'datatype': 'STRING'
         },
         'CANCER_TYPE_DETAILED': {
-            'is_patient_attribute': '0',
+            'is_patient_attribute': '1',
             'datatype': 'STRING'
         },
-        'DETAILED_CANCER_TYPE': {
-            'is_patient_attribute': '0',
-            'datatype': 'STRING'
+        'DERIVED_NORMALIZED_CASE_TYPE': {
+        },
+        'DERIVED_SAMPLE_LOCATION': {
         },
         'DFS_STATUS': {
             'is_patient_attribute': '1',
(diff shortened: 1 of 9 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -1622,58 +1607,44 @@ class ClinicalValidator(Validator):
...
         'SAMPLE_CLASS': {
-            'is_patient_attribute': '0',
             'datatype': 'STRING'
         },
         'SAMPLE_DISPLAY_NAME': {
             'is_patient_attribute': '0',
             'datatype': 'STRING'
         },
         'SAMPLE_TYPE': {
-            'is_patient_attribute': '0',
```
pieterlukasse: by removing this, does it mean is can now also be a patient attribute? If that is the case, then I don't think this makes sense. All `SAMPLE_` attributes should remain sample type attributes IMO.

jjgao:
@sandertan I think `CANCER_TYPE` and `CANCER_TYPE_DETAILED` are sample attributes here. The reason is that one patient may have multiple diagnoses.

cc'ing @n1zea144 on this.

sandertan:
Thanks @jjgao . I tested `CANCER_TYPE`,  `CANCER_TYPE_DETAILED`, `HISTOLOGY` and `KNOWN_MOLECULAR_CLASSIFIER` as both SAMPLE and PATIENT attribute, and the front-end correctly displays them in Study and Patient view in both cases.

I removed these from the core set of values. I can add `CANCER_TYPE*` to the core set of SAMPLE attributes if you want to make this more strict.

core/src/main/scripts/importer/validateData.py
sandertan: `'is_patient_attribute': '0',` will make it a SAMPLE attribute. Removing `'is_patient_attribute'`will make it flexible.

I've updated the commit, now added all SAMPLE_*-related attributes to the core sample set.

jjgao:
@sandertan OK to remove `CANCER_TYPE` etc from the core set. 

It may be a good idea to add `TUMOR_SITE`, `PRIMARY_SITE`, and `METASTATIC_SITE` to the core set of SAMPLE attributes.

inodb:
@sandertan `DERIVED_NORMALIZED_CASE_TYPE` is only used on frontend. It's not stored in the db. Same with `DERIVED_SAMPLE_LOCATION`. The latter uses `METASTATIC_SITE` , `PRIMARY_SITE` or `TUMOR_SITE` to get the location of the sample afaik

sandertan:
Updated the PR to include the suggestions from JJ and Ino.

Perhaps in a future PR, we could modify the way the attributes that are now synonyms are handled.

fedde-s:
I agree that a migration script for databases that have loaded studies using the old validator-loader pipeline would make the data that users see a lot less questionable, but it has implications that may be bigger than first appears. The current pipeline would ensure that studies with, say, `GENDER` fields per sample or a `METASTATIC_SITE` field on the patient level never end up in the database. Migrating an existing database to match would imply removing the study, or at least setting the study's `STATUS` flag to disable it.

inodb:
If we decide to do a SQL migration script, as a first step I would love to see some sort of database validator/sanity checker.

It might be nice to version and name the clinical data specification. Then we could say: the study complies with this clinical data spec name and version. The spec doesn't have to be one linear thing. I'd rather say a "hierarchical spec" makes more sense. We could e.g. have some studies that apply to just a very basic set of clinical attributes and some that comply to more complex ones.

jjgao:
Agreed that we should first have a specification of clinical data vocabulary. Once we have that, it may be a good idea to have some normalization script.

But I not sure if migration.sql is the right place to do data normalization. We should allow a local instance to use a different spec of clinical data.

PR #3917, opened 2018-02-28
Move security annotations from service to web layer.

n1zea144:
It was revealed during backend code profiling that authenticate added significant cost to the time it takes to process requests (see attached image as example).  After some further analysis, one reason for this increase is because security takes place in the service layer and intra-service communication results in multiple authorization checks for the same front-end request.  This is compounded when you have multiple requests entering the backend to load a single web page like Query Results.

This PR attempts to alleviate this issue by moving the security implementation from the service layer to the web layer.  As a result of this PR, the number or authorizations per request is reduced to one.

It was first thought that the task would amount to moving spring-security annotations from the  service classes to the controller classes.  The work became more involved as the use of "filter" classes from the web/src/main/java/org/cbioportal/web/parameter package were discovered.  As a result, additional code had to be added to CancerStudyPermissionEvaluator.  I did my best to keep this class as clean as possible and made improvements when there was opportunity.  For example, I utilized a Set when permission checks came in for lists 
 containing duplicate identifiers (for GENIE a list containing 22k+ identical cancer study IDs could be checked for read permission).

This PR also contains necessary updates so that SpringFox/Swagger documentation would operate properly with security running at the web level.

I should mention that there was code added to RC in a previous PR that added significant time to loading the query results page.  After some profiling it looks like the culprit is in fetchGenePanelDataInMultipleMolecularProfiles.  Specifically a call to java.util.stream.ReferencePipeline.findFirst in a method createGenePanelData (28 seconds for this call alone as its a nested for-loop over all genes over all samples (almost 22 thousand).  This code  clouds the optimizations made by this PR and should be addressed in a separate PR before the RC branch is put into production.

### DIFF
--- a/security/security-spring/pom.xml
+++ b/security/security-spring/pom.xml
@@ -17,7 +17,7 @@
     <dependencies>
       <dependency>
         <groupId>org.mskcc.cbio</groupId>
-        <artifactId>persistence-mybatis</artifactId>
+        <artifactId>web</artifactId>
         <version>${project.version}</version>
       </dependency>
       <!-- Spring Security -->
--- a/security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
+++ b/security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
@@ -38,6 +38,8 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.ArrayList;
+import java.util.Collection;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -50,6 +52,19 @@ import org.cbioportal.persistence.SampleRepository;
 import org.cbioportal.persistence.SampleListRepository;
 import org.cbioportal.persistence.SecurityRepository;
 import org.cbioportal.persistence.StudyRepository;
+import org.cbioportal.web.parameter.ClinicalAttributeFilter;
+import org.cbioportal.web.parameter.ClinicalDataMultiStudyFilter;
+import org.cbioportal.web.parameter.ClinicalDataIdentifier;
+import org.cbioportal.web.parameter.GenePanelMultipleStudyFilter;
+import org.cbioportal.web.parameter.SampleMolecularIdentifier;
+import org.cbioportal.web.parameter.MolecularDataMultipleStudyFilter;
+import org.cbioportal.web.parameter.MolecularProfileFilter;
+import org.cbioportal.web.parameter.MutationMultipleStudyFilter;
+import org.cbioportal.web.parameter.PatientFilter;
+import org.cbioportal.web.parameter.PatientIdentifier;
+import org.cbioportal.web.parameter.SampleFilter;
+import org.cbioportal.web.parameter.SampleIdentifier;
+import org.cbioportal.web.util.UniqueKeyExtractor;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
@@ -83,6 +98,9 @@ class CancerStudyPermissionEvaluator implements PermissionEvaluator {
 
     @Autowired
     private SampleListRepository sampleListRepository;
+
+    @Autowired
+    private UniqueKeyExtractor uniqueKeyExtractor;
 
     @Value("${app.name:}")
     private String APP_NAME;
@@ -123,6 +141,9 @@ class CancerStudyPermissionEvaluator implements PermissionEvaluator {
      *   'SampleList',
      *   'List<CancerStudyId>', 
      *   'List<MolecularProfileId>', 
+     *   'MolecularDataMultipleStudyFilter',
+     *   'MolecularProfileFilter',
+     *   'MutationMultipleStudyFilter',
      *   or 'List<SampleListId>'
      * @param permission
      */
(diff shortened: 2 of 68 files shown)


security/security-spring/pom.xml
```diff
@@ -20,6 +20,11 @@
         <artifactId>persistence-mybatis</artifactId>
```
ersinciftci: We can remove this dependency as web already depends on this.

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -262,6 +315,119 @@ public boolean hasPermission(Authentication authentication, Object targetDomainO
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean hasAccessToMolecularProfiles(Authentication authentication, Collection<String> molecularProfileIds, Object permission)
+    {
+        for (String molecularProfileId : molecularProfileIds) {
+            MolecularProfile molecularProfile = molecularProfileRepository.getMolecularProfile(molecularProfileId);
```
ersinciftci: we can use fetch methods to get multiple profiles in one call, to prevent making db calls in a loop. (the optimization I mentioned a few weeks ago)

n1zea144: Yes, good idea. I can do this.

security/security-spring/pom.xml
n1zea144: Ok.

n1zea144: Done!

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -262,6 +315,120 @@ public boolean hasPermission(Authentication authentication, Object targetDomainO
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean hasAccessToSampleLists(Authentication authentication, List<String> sampleListIds, Object permission)
+    {
+        for (String sampleListId : sampleListIds) {
+            SampleList sampleList = sampleListRepository.getSampleList(sampleListId);
```
ersinciftci: Sorry forgot to say, this one also has the same problem.

n1zea144:
I've added a second commit with a proposed fix to the recently discovered performance bottleneck in GenePanelServiceImpl.  The patch dramatically improves loading a genie oncoprint (50+sec -> ~1).  I used an Multikey map to eliminate the use of stream/filter/findFirst constructs.  These are EVIL when the sample list is large:
genePanelDataList.stream().filter(g -> g.getMolecularProfileId().equals(
                    molecularProfileId) && g.getSampleId().equals(sampleId)).findFirst();

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -262,6 +315,120 @@ public boolean hasPermission(Authentication authentication, Object targetDomainO
...
+        }
+        return true;
+    }
+
+    private boolean hasAccessToMolecularProfiles(Authentication authentication, Collection<String> molecularProfileIds, Object permission)
+    {
+        List<String> profileIds = (molecularProfileIds instanceof List) ?
+            (List<String>)molecularProfileIds : new ArrayList<String>(molecularProfileIds);
+        for (MolecularProfile molecularProfile : molecularProfileRepository.getMolecularProfiles(profileIds, "SUMMARY")) {
+            if (molecularProfile == null || !hasPermission(authentication, molecularProfile, permission)) {
```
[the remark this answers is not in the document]
mandawilson: @ersinciftci is the problem that in hasPermission `cancerStudy == null` will be true if it is a molecular profile here, and then it will look up the cancer study using the studyRepository?  Would this change fix it? `molecularProfileRepository.getMolecularProfiles(profileIds, "SUMMARY")` -> `molecularProfileRepository.getMolecularProfiles(profileIds, "DETAILED")`.  I am not sure but I *think* that the sample list will have a cancer study included and so the test `cancerStudy == null` will be false, and there will not be a call to the DB there.  It has been a long time since I have looked at this and I don't know MyBatis well, so I am not completely sure about what I am saying!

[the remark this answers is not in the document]
ersinciftci: @mandawilson You are right, if we call `DETAILED` projection for molecularProfiles and sampleLists, `cancerStudy == null` would always be false, so wouldn't make a call to db in a loop anymore.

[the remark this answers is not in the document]
mandawilson: @ersinciftci @n1zea144 I think that for sampleLists, it already has the DETAILED projection, see here: https://github.com/cBioPortal/cbioportal/blob/rc/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/SampleListMyBatisRepository.java#L39  And that it is just that `molecularProfileRepository.getMolecularProfiles(profileIds, "SUMMARY")` needs to change to `molecularProfileRepository.getMolecularProfiles(profileIds, "DETAILED")`.  Though of course if the default behavior for getSampleList ever changed so that it stopped being a DETAILED projection, it would silently start accessing the DB from the loop.

PR #3924, opened 2018-03-01
write custom case list order to serverVars

adamabeshouse:


inodb:
Fixed in: https://github.com/cBioPortal/cbioportal/pull/4053

Issue #4003, opened 2018-03-20
importer logic to determine protein start position based on protein_change is incorrect


inodb:
See: https://github.com/genome-nexus/genome-nexus/issues/151

This only happens when protein_position is empty in the data file

PR #4004, opened 2018-03-20
Improve the logic to annotate proteinStart and proteinEnd

zhx828:
Reuse OncoKB logics to annotate variant;
This solves issue #4003


inodb:
@zhx828 is the oncokb logic also in Java or in JS? If it's in java would be nice to reuse the functions directly by importing some shared library (i think genome nexus might need to use the same logic).

zhx828:
@inodb it's in JAVA, we could share the utils. But need some work to clean up everything. Will update the changes later on once that is ready.

inodb:
Sounds good, thanks!

core/src/main/java/org/mskcc/cbio/portal/util/ExtendedMutationUtil.java
```diff
@@ -354,4 +360,99 @@ public static ExtendedMutation newMutation()
 
 		return mutation;
 	}
+
+    private static Map<String, Integer> annotateProteinChange(String proteinChange) {
```
onursumer: @inodb @zhx828 we have a `ProteinChangeResolver` this class in `GenomeNexus`: https://github.com/genome-nexus/genome-nexus/blob/67a3ebd56f953a8c3e093e5f909eba4b00a9ee04/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/ProteinChangeResolver.java

It might be possible to merge these two implementations within `GenomeNexus`.

zhx828: This is an issue in the importer. The problem happened here is that the study was annotated long time ago, and the mutation didn't get any protein start and protein end but only protein change. If you reannotate the study, the mutation will be shown with correct protein start. This pull request is trying to fix the issue for the studies without proper annotation.

inodb: yeah i think it makes sense to merge this logic. In the proteinchangeresolver class it also tries to determine start/end based on protein change. So we have 3 places now where this is done, genome-nexus, genome-nexus-annotation-pipeline and cbioportal importer. Let's follow up later and make a single utility function we can import

zhx828:
@inodb @onursumer do you think we can merge this?

PR #4020, opened 2018-03-22
Gene panel data API change

ersinciftci:
Reverted https://github.com/cBioPortal/cbioportal/pull/3793 because it used some properties that I changed now, will do it again later without affecting the Gene Panel Data API

### DIFF
--- a/model/src/main/java/org/cbioportal/model/Mutation.java
+++ b/model/src/main/java/org/cbioportal/model/Mutation.java
@@ -35,10 +35,6 @@ public class Mutation extends Alteration implements Serializable {
     private String driverFilterAnnotation;
     private String driverTiersFilter;
     private String driverTiersFilterAnnotation;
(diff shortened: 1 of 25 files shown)


model/src/main/java/org/cbioportal/model/Mutation.java
```diff
@@ -35,10 +35,6 @@
     private String driverFilterAnnotation;
     private String driverTiersFilter;
     private String driverTiersFilterAnnotation;
-    @NotNull
-    private Boolean sequenced;
-    @NotNull
-    private Boolean wildType;
```
adamabeshouse: why are you reverting this?

ersinciftci: because these values depends on the verbose gene panel data logic, which we change here, so I'll introduce them later.

Issue #4067, opened 2018-04-04
Oncoprint not appearing in current master; gene panel request error


fedde-s:
When I run the current master branch (8784f73b0b73207b84aa4f926a1e9a4a63d8a359) and open the results view, no Oncoprint appears. Instead, the backend throws a NullPointerException and sends a HTTP 500 Internal Server Error as the response to a post request for `api/gene-panel-data/fetch`.

fedde-s:
The bug seems to have been introduced by this commit:

```
commit 48bffceb7eb65c2fd905a67b7c30dea3836e5018
Author: Ersin Ciftci <arseniusthegreat@gmail.com>
Date:   Mon Apr 2 11:11:31 2018 -0400

    Gene panel improvement

    Fixes #3965
```

fedde-s:
@ersinciftci, are you aware of this?

ersinciftci:
@jjgao it seems @fedde-s 's local DB is missing the `_sequenced` list (I thought it was enforced somehow), should I assume all samples are sequenced if the study doesn't have `_sequenced` list?

PR #4068, opened 2018-04-04
Use all sample list if study has no sequenced list

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4067


fedde-s:
I can confirm that it fixes #4067. I built and ran this commit (3510f366ee3b1b7944b2a269e421cb8bfd569bed) in the same containerised setup as before, migrating a database starting from the same snapshot, and I see an Oncoprint now 👍

PR #4114, opened 2018-04-19
Fallback on gene symbol if gene not resolved by entrez

### LINKED ISSUE #1615: Hugo_symbol not looked at in data_CNA.txt importing
If both `Hugo_symbol` and `Entrez_Gene_Id` columns are present in CNA file, the hugo symbol is not used if the entrez gene Id cannot be found, causing the record to be skipped.

angelicaochoa:
# What? Why?
Fix #1615 

- Fallback on hugo gene symbol if could not resolve `CanonicalGene` by entrez id

# Notify reviewers
@jjgao @pieterlukasse 

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
@@ -334,39 +334,39 @@ public class ImportTabDelimData {
                     ProgressMonitor.logWarning("Ignoring gene ID:  " + geneSymbol);
                     return false;
                 } else {
-                	List<CanonicalGene> genes = null;
-                	//If rppa, parse genes from "Composite.Element.REF" column:
-                	if (rppaProfile) {
+                    List<CanonicalGene> genes = null;
+                    //If rppa, parse genes from "Composite.Element.REF" column:
+                    if (rppaProfile) {
                         genes = parseRPPAGenes(geneSymbol);
                         if (genes == null) {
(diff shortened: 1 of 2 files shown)


angelicaochoa:
@jjgao yes, all have been good 👍

pieterlukasse:
@angelicaochoa @jjgao is falling back on symbol really correct? Maybe we need to look at why entrez gene ids are deprecated in the first place, before deciding for this strategy. Did you do that? If yes, could you please share why the fallback is the correct approach? My concern is with the scenario where entrez/symbol combinations are deprecated because what was first thought to be a gene is either not a gene, or multiple genes, etc, leading to repurposing of the symbol. In this case, falling back to the symbol would be wrong as the old alterations measured in one genomic location would now assigned to a different one.

sandertan:
Falling back on symbol can be dangerous and this should be reviewed by the user. Perhaps deprecation of a gene was necessary because the gene has significantly changed, and any assumptions made based on the old gene should be reviewed. Symbol repurposing and incorrect mapping can be another scenario where falling back on Symbol is unwanted. 

As long as Entrez ID is the primary gene ID in cBioPortal, I think "not resolved by Entrez" should lead to an error or strong warning, and the user that imports should further investigate these cases.

From a high level perspective, I wonder if cBioPortal should keep matching Entrez/Symbol with an internal gene database (build from an NCBI gene table). Perhaps just load and show every Symbol the user tries to import. The UI requires Symbols and not Entrez IDs, so perhaps Symbols should be the primary gene ID.

Issue #4104, opened 2018-04-16
Bug in the oncoprint tab when calculating the percentage of altered cases/patients


alisman:
From Google Groups:

https://groups.google.com/forum/#!topic/cbioportal/luYgF0D1l3M

It seems like there is a small bug in the oncoprint tab when calculating the percentage of altered cases/patients. This can be reproduced by selecting the TCGA LUAD provisional dataset, selecting the tumour samples with mRNA data (517) and looking at whichever genes of interest.

The percentage of altered cases is calculate with respect to the number of complete tumours (230) instead of the number of selected tumours. The problem is that the number of altered cases refers to all selected cases. It is even possible to get a >100% result in the message (see screenshot).

http://www.cbioportal.org/index.do?cancer_study_id=luad_tcga&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=luad_tcga_rna_seq_v2_mrna&gene_list=ADAM10%2520ADAM17%2520APH1A%2520APH1B%2520ARRDC1%2520CIR1%2520CTBP1%2520CTBP2%2520CUL1%2520DLL1%2520DLL3%2520DLL4%2520DTX1%2520DTX2%2520DTX3%2520DTX3L%2520DTX4%2520EP300%2520FBXW7%2520HDAC1%2520HDAC2%2520HES1%2520HES5%2520HEYL%2520ITCH%2520JAG1%2520JAG2%2520KDM5A%2520LFNG%2520MAML1%2520MAML2%2520MAML3%2520MFNG%2520NCOR2%2520NCSTN%2520NOTCH1%2520NOTCH2%2520NOTCH3%2520NOTCH4%2520NRARP%2520NUMB%2520NUMBL%2520PSEN1%2520PSEN2%2520PSENEN%2520RBPJ%2520RBPJL%2520RFNG%2520SNW1%2520SPEN%2520HES2%2520HES4%2520HES7%2520HEY1%2520HEY2&geneset_list=+&tab_index=tab_visualize&Action=Submit&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=luad_tcga_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=luad_tcga_gistic&genetic_profile_ids_PROFILE_MRNA_EXPRESSION=luad_tcga_mrna_median_Zscores

adamabeshouse:
[...]

PR #4118, opened 2018-04-20
Add 'profiled' field instead of 'wholeExomeSequenced' to Gene Panel D…

ersinciftci:
…ata API


fedde-s:
This looks like it's written in a way that will break existing code and code currently waiting for review :(   But I guess that's intentional.

fedde-s:
Although in that case, note that it's directed to the master branch.

pieterlukasse:
Is this a fix to an existing bug? If not, maybe move it to `rc`?

ersinciftci:
@jjgao asked for this change for https://github.com/cBioPortal/cbioportal/issues/4104

PR #4124, opened 2018-04-24
Add CircleCI E2E tests

inodb:
Run e2e tests from frontend repo against cbioportal backend

### DIFF
--- a/portal/pom.xml
+++ b/portal/pom.xml
@@ -105,9 +105,9 @@
             <configuration>
               <artifactItems>
                 <artifactItem>
-                  <groupId>com.github.cbioportal</groupId>
+                  <groupId>com.github.inodb</groupId>
                   <artifactId>cbioportal-frontend</artifactId>
-                  <version>v1.12.1</version>
+                  <version>95702b4e681de7a0f69d08d257c4537c41bd676e</version>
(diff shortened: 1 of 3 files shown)


inodb:
Still getting ~15 errors: https://circleci.com/gh/inodb/cbioportal/32

portal/pom.xml
```diff
@@ -105,9 +105,9 @@
             <configuration>
               <artifactItems>
                 <artifactItem>
-                  <groupId>com.github.cbioportal</groupId>
+                  <groupId>com.github.inodb</groupId>
```
adamabeshouse: make sure to revert this

adamabeshouse:
cool, so this will run based on master branch?

inodb:
@adamabeshouse it will run the backend in the PR with the frontend defined in the pom against the e2e tests in the same frontend

jjgao:
@inodb what's the plan on this?

inodb:
@jjgao 	I think we should close this for now and revisit where these tests should live. We also want to run them before data updates to the portal. We should come up with some interface that makes it a little easier to run the tests without pulling the entire frontend repo imo. E.g. some docker compose file that runs the whole suite of tests

PR #4129, opened 2018-04-25
Added alternate freq allele MAF columns.

angelicaochoa:
# What? Why?
Addresses issue in TCGA-BRCA where non-standardard allele frequency columns are provided and therefore not imported.

Mappings for alt column names for calculating allele frequency:
- tumor_ref_reads = t_ref_count
- tumors_var_reads = t_alt_count
- normal_ref_reads = n_ref_count
- normal_var_reads = n_alt_count

- `MafUtil` now interprets alternate allele freq columns when parsing MAF records

# Notify reviewers
@cBioPortal/backend 


n1zea144:
@angelicaochoa Is this the only TCGA file that contains alternative names for allele frequencies?  If so, why not change the column headers in the MAF instead of this patch?

PR #4177, opened 2018-05-04
Change N/S to N/P

inodb:
All references to not sequenced should be changed to not profiled


adamabeshouse:
i dont think this is necessary, it was still showing up that way because of mismatch tolerance

PR #4187, opened 2018-05-08
Use maintained copy of OQL parser to query virtual studies

fedde-s:
As @adamabeshouse noticed in review of the feature presented in cBioPortal/cbioportal-frontend#1143, multi-study queries used a copy of the OQL parser that's no longer actively maintained. As it turned out, it was to decide between showing the Oncoprint tab or the Cancer Types Summary tab after querying. This caused issues integrating new OQL syntaxes, which are readily fixed by using the new parser instead. That solution has the additional advantage that `DATATYPES` statements are no longer counted as genes in that decision.


fedde-s:
Done, @adamabeshouse! Just rebased on the latest `rc`.

PR #4212, opened 2018-05-11
updated News for 1.13.0 release

jjgao:


inodb:
@jjgao Thanks for updating! The link for me keeps going to: http://www.cbioportal.org/index.do?session_id=59357ca8498e5df2e2919824&found_session=true#mutation_details. Automatically going to mutations tab instead of oncoprint, do you have that as well?

PR #4216, opened 2018-05-11
Create "isolateNetwork" mode in which network tab will render alone

alisman:
This code will have no effect unless the isolateNetwork parameter is passed in URL.   When that is so, styles will hide portal's chrome so only network tab shows.  This will allow network tag to be shown inside of iframe

# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
@@ -38,7 +38,14 @@
 <jsp:include page="global/header.jsp" flush="true" />
 
 <script>
+
+if (/isolateNetwork=1/.test(window.location.search)) {
+    window.isolateNetwork = true;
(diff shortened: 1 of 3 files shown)


portal/src/main/webapp/WEB-INF/jsp/visualize.jsp
```diff
@@ -38,7 +38,14 @@
 <jsp:include page="global/header.jsp" flush="true" />
 
 <script>
+
+if (/isolateNetwork=1/.test(window.location.search)) {
```
adamabeshouse: i guess this is okay although it does feel hacky..

alisman: oh this is a whole thing is a glorious hack.  i think the above is ok.  it's in a throwaway codebase.

PR #4227, opened 2018-05-15
Issue 4165 fix.

n1zea144:
Filters out fusion record in the case that gene is not part of gene panel.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
@@ -39,6 +39,7 @@ import org.mskcc.cbio.portal.dao.*;
 import org.mskcc.cbio.portal.model.*;
 import org.mskcc.cbio.portal.model.ExtendedMutation.MutationEvent;
 import org.mskcc.cbio.portal.util.*;
+import org.mskcc.cbio.portal.repository.GenePanelRepositoryLegacy;
 
 /**
  * Imports a fusion file.
@@ -51,13 +52,13 @@ public class ImportFusionData {
     public static final String FUSION = "Fusion";
     private File fusionFile;
     private int geneticProfileId;
-    private String genePanel;
+    private String genePanelID;
     private Set<String> sampleSet  = new HashSet<>();
 
-    public ImportFusionData(File fusionFile, int geneticProfileId, String genePanel) {
+    public ImportFusionData(File fusionFile, int geneticProfileId, String genePanelID) {
         this.fusionFile = fusionFile;
         this.geneticProfileId = geneticProfileId;
-        this.genePanel = genePanel;
+        this.genePanelID = genePanelID;
     }
 
     public void importData() throws IOException, DaoException {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
```diff
@@ -120,6 +121,12 @@ public void importData() throws IOException, DaoException {
                     ProgressMonitor.logWarning("Gene not found:  " + geneSymbol + " ["
                             + entrezGeneId + "]. Ignoring it "
                             + "and all fusion data associated with it!");
+                    continue;
+                }
+                if (genePanel != null && !GeneticProfileUtil.geneInPanel(gene, genePanel)) {
+                    ProgressMonitor.logWarning("Gene is not profiled in gene panel:  " + geneSymbol + " ["
```
pieterlukasse: @n1zea144 any such new warnings in the loader also need to be implemented in the validator. Can you please add it there as well?

core/src/test/scripts/test_data/study_es_0/data_fusions.txt
```diff
@@ -1,4 +1,5 @@
 Hugo_Symbol	Entrez_Gene_Id	Center	Tumor_Sample_Barcode	Fusion	DNA_support	RNA_support	Method	Frame
 FGFR3	2261	saturn	TEST-A2B8-01	TMPRSS2-ERG fusion	yes	yes	unknown	unknown
 ERBB2	2064	jupiter	TEST-A2FF-01	Fusion2	yes	yes	unknown	unknown
-ERBB2	2064	uranus	TCGA-GI-A2C8-01	ERBB2-TEST Fusion	yes	yes	unknown	frameshift
\ No newline at end of file
+ERBB2	2064	uranus	TCGA-GI-A2C8-01	ERBB2-TEST Fusion	yes	yes	unknown	frameshift
+TP53	7157	uranus	TCGA-GI-A2C8-01	TP53-TEST Fusion	yes	yes	unknown	frameshift
```
pieterlukasse: @n1zea144 can you please remove this one? In study_es_0 the idea is to have data that does *not* result in any warnings. You could add it to study_es_3, which is the study where we test the warnings.

n1zea144:
@pieterlukasse I've created a new PR to address your issues.  Lets move the conversation there:

[PR-4305](https://github.com/cBioPortal/cbioportal/pull/4305)

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileUtil.java
```diff
@@ -93,4 +91,13 @@ public static int getGenePanelId(String panelId) {
         GenePanel genePanel = genePanelRepositoryLegacy.getGenePanelByStableId(panelId).get(0);
         return genePanel.getInternalId();
     }
+
+    public static boolean geneInPanel(CanonicalGene gene, GenePanel genePanel) {
```
sheridancbio: I know this is after the fact, but one thought I had while reading this is that we seem to be adding new dependencies into the core module here.  Maybe this functionality should be in a util class in another module?
Also, I'm surprised that we need to loop through the genes in a panel in order to find if a gene is a member. We could put this functionality into class GenePanel ... and store (if necessary) a hash set of entrezGeneId for members.

Issue #3979, opened 2018-03-15
ln?q=EGFR:MUT rediction


jjgao:
http://cbioportal.org/ln?q=EGFR:MUT is still redirecting to cross_cancer.do? Please change it to index.do.

alisman:
@inodb can you take this?

jjgao:
I guess it has not been pushed out?

inodb:
@jjgao this is live now - note that it does take quite a while before something appears on the page but hopefully that'll be solved once we switch to single page app

PR #4249, opened 2018-05-21
Point all ln? links to use new index.do

inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/3979 (don't use crosscancer.do)

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/servlet/LinkOut.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/servlet/LinkOut.java
@@ -110,26 +110,28 @@ public class LinkOut extends HttpServlet {
     }
 
     private String createCrossCancerForwardingUrl(String hostURL, String geneList) {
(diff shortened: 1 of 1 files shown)


kalletlak:
@inodb is there a possibility for sending a list of study ids instead of `all` to `crosscancer.do`?

inodb:
@kalletlak could you elaborate? The goal is to not use `crosscancer.do` anymore (we want to get rid of it). I'm sending a list of study ids in `cancer_study_list` and `cancer_study_id=all` to `index.do`

kalletlak:
If someone already created `crosscancer.do` url with a subset of studies instead of `all`, should we support them

inodb:
@kalletlak good point - we might want to do that. I think that is a different issue tho than redirecting ln? to index.do. That would be redirecting cross_cancer.do to index.do. Right?

jjgao:
@inodb @kalletlak don't worry about crosscancer.do. I don't see much traffic pointing there now.

PR #4264, opened 2018-05-23
Display only the study-specific clinical variables in Plots tab

### LINKED ISSUE #4256: Display name of clinical variables in plots tab
There is a bug in how the Plots tab displays the name of the clinical variables. If I have two studies in an instance with different display names for clinical variables, e.g. `OS_STATUS` and `DFS_STATUS` are `OS Status` and `DFS Status` in study 1, and `Test` and `Test` in study 2; and then I query ONLY study 1, I see two `Test` instead of `OS Status` and `DFS Status`.

oplantalech:
# What? Why?
Fix #4256. 

The new function `generateJson` filters all the clinical variables retrieved from the database by the cancer study ID of the queried study. Therefore, the temporary solution in `clinSpec.js` is useless and I have removed it.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/web_api/GetClinicalData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/web_api/GetClinicalData.java
@@ -129,15 +129,17 @@ public class GetClinicalData {
         ClinicalData c = data.get(0);
         return "" + c.getStableId() + "\t" + c.getAttrId() + "\t" + c.getAttrVal();
     }
-
+    
     /**
      * Creates a json object with data and attributes fields that correspond to the data
-     * in the clinicals and the set of attributes that exist in the clinicals
+     * in the clinicals and the set of attributes that exist in the clinicals, filtered
+     * by the defined cancer study ID.
      * @param clinicals
+     * @param cancer_study_id
      * @return
      * @throws DaoException
      */
-    public static JSONObject generateJson(List<ClinicalData> clinicals) throws DaoException {
+    public static JSONObject generateJson(List<ClinicalData> clinicals, Integer cancer_study_id) throws DaoException {
         Set<JSONObject> attrs = new HashSet<JSONObject>();
         JSONObject toReturn = new JSONObject();
         JSONArray data = new JSONArray();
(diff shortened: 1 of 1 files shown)


adamabeshouse:
I don't think I'm qualified to comment on this, I don't touch the backend. Maybe @ersinciftci ?

portal/src/main/webapp/js/src/plots-tab/view/components/clinSpec.js
```diff
@@ -3,9 +3,7 @@ var clinSpec = (function() {
     function appendAttrList(axis) {
         $("#" + ids.sidebar[axis].spec_div).append("<br><h5>Clinical Attribute</h5>");
         $("#" + ids.sidebar[axis].spec_div).append("<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<select class='chosen' style='max-width: 250px;' id='" + ids.sidebar[axis].clin_attr + "'>");
-        //TODO: this is a temp solution (duplications in API results)
-        var _uniqClinMeta = _.uniq(metaData.getClinAttrsMeta(), "id");
-        $.each(_uniqClinMeta, function(index, obj) {
+        $.each(metaData.getClinAttrsMeta(), function(index, obj) {
```
pieterlukasse: @oplantalech is this change really needed?

oplantalech: Well, this is filtering all clinical variables by their id. Theoretically, with the new fix, they shouldn't be duplicates, otherwise this would mean that we have two clinical variables in the same study with the same attribute name (id; not to confuse with display name).

core/src/main/java/org/mskcc/cbio/portal/web_api/GetClinicalData.java
```diff
@@ -149,9 +151,11 @@ public static JSONObject generateJson(List<ClinicalData> clinicals) throws DaoEx
             attrIds.add(c.getAttrId());
 //            }
         }
-        
+
         for (ClinicalAttribute attr : DaoClinicalAttributeMeta.getDatum(attrIds)) {
```
jjgao: Maybe we should change DaoClinicalAttributeMeta.getDatum to accept study id?

oplantalech: Done.

oplantalech:
Following the suggestion by @jjgao, I discovered that there is already a `DaoClinicalAttributeMeta.getDatum` function that accepts the cancer study id as a parameter and returns only the clinical variables from the mentioned cancer study. Therefore, now this function is used and I've removed the redundant code that filters clinical variables by the cancer study id.

jjgao:
@oplantalech could you please rebase before we can merge?

oplantalech:
@jjgao I've rebased it. Now it's ready to merge!

PR #4267, opened 2018-05-23
Pass baseurl for routing on frontend

inodb:
P much the same as window.frontendConfig.apiRoot but good to be explicit


portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
```diff
@@ -106,4 +106,5 @@ window.frontendConfig.frontendUrl = window.frontendConfig.frontendUrl? window.fr
 window.frontendConfig.apiRoot = window.frontendConfig.apiRoot? window.frontendConfig.apiRoot : '<%=baseURL%>';
 // default, override on per page bases, set to hash if full react page
 window.frontendConfig.historyType = window.frontendConfig.historyType? window.frontendConfig.historyType : 'memory'; 
+window.baseUrl = '<%=baseUrl%>';
```
alisman: could we just namespace this inside of frontendConfig?

Issue #4259, opened 2018-05-22
Gene panel API error 


pieterlukasse:
The API call to 
http://www.cbioportal.org/api/swagger-ui.html#/Gene_Panels/getGenePanelDataUsingPOST below

```
curl -X POST "http://www.cbioportal.org/api/molecular-profiles/nsclc_unito_2016_mutations/gene-panel-data/fetch" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"sampleListId\": \"string\"}"
```
Results in 

```
<!doctype html><html lang="en"><head><title>HTTP Status 500 – Internal Server Error</title><style type="text/css">h1 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:22px;} h2 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:16px;} h3 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:14px;} body {font-family:Tahoma,Arial,sans-serif;color:black;background-color:white;} b {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;} p {font-family:Tahoma,Arial,sans-serif;background:white;color:black;font-size:12px;} a {color:black;} a.name {color:black;} .line {height:1px;background-color:#525D76;border:none;}</style></head><body><h1>HTTP Status 500 – Internal Server Error</h1><hr class="line" /><p><b>Type</b> Exception Report</p><p><b>Message</b> Request processing failed; nested exception is org.springframework.jdbc.BadSqlGrammarException: </p><p><b>Description</b> The server encountered an unexpected condition that prevented it from fulfilling the request.</p><p><b>Exception</b></p><pre>org.springframework.web.util.NestedServletException: Request processing failed; nested exception is org.springframework.jdbc.BadSqlGrammarException: 
### Error querying database.  Cause: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near &#39;&#39; at line 17
### The error may exist in org&#47;cbioportal&#47;persistence&#47;mybatis&#47;GenePanelMapper.xml
### The error may involve defaultParameterMap
### The error occurred while setting parameters
[...]

PR #4270, opened 2018-05-24
Fix Gene Panel API invalid sampleListId

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4259

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
@@ -21,6 +21,7 @@ import org.apache.commons.collections.map.MultiKeyMap;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
```diff
@@ -104,6 +104,9 @@ public GenePanel getGenePanel(String genePanelId) throws GenePanelNotFoundExcept
 
         molecularProfileService.getMolecularProfile(molecularProfileId);
         List<String> sampleIds = sampleListRepository.getAllSampleIdsInSampleList(sampleListId);
+        if (sampleIds.isEmpty()) {
+            return new ArrayList<>();
```
onursumer: Why not `Collections.emptyList()`?

ersinciftci: it's not type safe, but I'll change to Collections.<GenePanelData>emptyList()

PR #4305, opened 2018-05-30
Remove data in study_es_0 which generate warnings (its warning free).

n1zea144:
Removed data that was put into study_es_0 from [PR-4227](https://github.com/cBioPortal/cbioportal/pull/4227).  This dataset should be warning free.

To validate the filtering of fusion partners requires gene panel data.  It looks like there are only place holders in the validator for gene panel / matrix validation.  If someone can add that, I can piggyback off of it, otherwise I suggest leaving it out.  Technically speaking, I don't think its incorrect to include a fusion event which includes a partner gene that is not profiled if it is in fact a partner in the event, but I did update File-Formats.md to state that the event will be filtered out during import.


core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -147,7 +147,7 @@ public void testLoadStudyEs0() throws Throwable {
             assertEquals(31, mutations.size());
 
             //===== Check FUSION data ========
-            // Are there 4 fusion entries in mutation profile, but one is off panel, so only 3 should be imported
```
pieterlukasse: this is a good test, if done for study_es_3

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -130,7 +130,7 @@ public void testLoadStudyEs0() throws Throwable {
                 }
             }
             //check that there are only warnings for empty positions in fake data:
-            assertEquals(13, countWarnings);
+            assertEquals(12, countWarnings);
```
pieterlukasse: Let's revert this to an assert that checks that there are no warnings. To get this working, you will also need to remove the cause of the warnings from study_es_0. I see now that the confusion was introduced by @dionnezaal in commit https://github.com/cBioPortal/cbioportal/commit/bbbdae3bef9c8dd95e9e2b34e51abe98a90cd834 . The solution for these extra 12 warnings will be to fix the study_es_0 MAF data so that it does not give warnings. 

Original code was: 
```java
          //check that there no warnings:
          assertEquals(0, countWarnings);
```

pieterlukasse:
Thanks @n1zea144 . I added my comments.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
sandertan: Apologies for the late reply. I fixed this in a PR to @n1zea144 's branch: https://github.com/n1zea144/cbioportal/pull/1

I also reverted many floats back to integers, for example for Start_Position, which was accidentally done in cBioPortal/cbioportal@bbbdae3#diff-3cd4ce7cd10421c2d456537975d4d810 , when adding the Germline mutations.

Issue #2478, opened 2017-05-23
Data base schema import failure


cbio1:
hi,

I am truly amazed by your work of cbioportal!

Just tried to install the local copy. followed every steps and tried to import the database schema. and it reported
ERROR 1071 (42000) at line 321: Specified key was too long; max key length is 767 bytes
hacked the .sql file 
![image](https://cloud.githubusercontent.com/assets/28907745/26375439/91baf1d4-3fce-11e7-91d4-ab1d8d416057.png)
by reducing the varchar(400) to varhar(255) and it worked.

hopefully this hack wont damage the whole system after they are installed?

thanks!

jjgao:
Looks like Innodb has a limitation on that.

https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes

cbio1:
Thanks jjgao.

While I continued the configuration, i found two versions of the schema sql import file:

one is the current repo's version:
https://github.com/cBioPortal/cbioportal/blob/master/db-scripts/src/main/resources/cgds.sql

another one is referenced in the installation doc page:
https://cbioportal.readthedocs.io/en/latest/Docker-Prerequisites.html

which does provide a varchar(255) version of the mutation_event table:
https://raw.githubusercontent.com/cBioPortal/cbioportal/v1.3.1/core/src/main/resources/db/cgds.sql

Hope this helps!

sheridancbio:
Yes, if you are using a mysql configuration with character encoding UTF8 then varchar(255) is the longest key element allowed. In our production databases we currently are using varchar(255) for TUMOR_SEQ_ALLELES because we switched to UTF8 encoding recently. We had an issue with long discussion of this topic ( #1801 ). I see that that issue was closed and I was supposed to create a new issue to track the update of the database installation documentation to carefully explain all the options to installers. I'll check the documentation and create that issue (or alter this issue so that what is needed is spelled out here)

fedde-s:
I'm running into this issue as well, when I run the latest mysql:8.0 [from Docker Hub](https://hub.docker.com/_/mysql/). I could work around it by reverting to mysql:5.7 for now, but we should solve it eventually.

fedde-s:
For completeness, I'm using the latest seed DB from Datahub, not the outdated ones you linked.

fedde-s:
[...]

PR #4333, opened 2018-05-31
Recommend MySQL <=5.7.x

fedde-s:
Tell people who want to install a cBioPortal server not to use MySQL 8.0 for now unless they know what they're doing, as the new default character encoding is not compatible with the cBioPortal
database schema for now. See https://github.com/cBioPortal/cbioportal/issues/2478.


fedde-s:
I will be unavailable for amendments until Wednesday, but the reviewers should be able to push additional commits to this branch if necessary before merging.

fedde-s:
Rebased now that v1.13.1 has been released.

Issue #4371, opened 2018-06-14
Mutational Signature Support


n1zea144:
Provide full-stack support for Mutational Signature (MS) data.

**Tasks for importing MS data into database:**

- [x] Extended [database schema](https://github.com/cBioPortal/cbioportal/blob/master/db-scripts/src/main/resources/cgds.sql) to capture MS metadata.
- [x] Create bean to represent an MS record in the context of importing ([model](https://github.com/cBioPortal/cbioportal/tree/master/core/src/main/java/org/mskcc/cbio/portal/model) package within core maven project).
- [x] Create Data Access Object for reading/writing MS metadata records. ([dao](https://github.com/cBioPortal/cbioportal/tree/master/core/src/main/java/org/mskcc/cbio/portal/dao) package within core maven project).
- [x] Create script for importing MS metadata into the database.
- [ ] Modify [ImportTabDelimData](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java) to support importing of MS data.

**Tasks for reading MS data out of database**
To be completed

**Task for proper integration of code**
To be completed

jjgao:
@n1zea144 @dippindots is this issue still relevant?

n1zea144:
No, I will close.

PR #4392, opened 2018-06-22
Importing mutational signature (MS) data into database

jxu8:
# What? Why?
For #4371 

- Extend database schema to capture mutational signature (MS) metadata
- create bean to represent MS record
- create DAO for MS
- create import script for MS data


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureData.java
```diff
@@ -0,0 +1,164 @@
...
+
+public class ImportMutationalSignatureData extends ConsoleRunnable {
+
+    @Override
+    public void run() {
+        try {
+            String progName = "ImportMutationalSignatureData";
+            String description = "Import mutational signature data files.";
+
+            OptionParser parser = new OptionParser();
```
n1zea144: I think you only need three arguments here -  help, mutational-signature-file, and replace (or override-existing).

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureData.java
```diff
@@ -0,0 +1,164 @@
...
+     * @param mutationalSignatureFile File with mutational signature information
+     * @throws IOException
+     * @throws DaoException
+     */
+    
+    public static void importData (File mutationalSignatureFile, boolean updateInfo) throws IOException, DaoException{
+        try (FileReader reader = new FileReader(mutationalSignatureFile)) {
+            BufferedReader buf = new BufferedReader(reader);
+            String line;
+            while ((line = buf.readLine()) != null){
```
n1zea144: Does this compile?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureData.java
```diff
@@ -0,0 +1,164 @@
...
+                ProgressMonitor.incrementCurValue();
+                ConsoleUtil.showProgress();
+                
+                String[] parts = line.split("\t");
+
+                //assume mutational signature id and description are first two columns in file
+                MutationalSignature mutationalSignature = new MutationalSignature();
```
n1zea144: I would add some checks to parts to make sure fields are not empty before creating a new MutationalSignature() and setting its fields.

business/src/main/java/org/mskcc/cbio/portal/model/MutationalProfile.java
```diff
@@ -3,12 +3,12 @@
 import java.io.Serializable;
 import java.util.Map;
 
-public class MutationalSignature implements Serializable {
+public class MutationalProfile implements Serializable {
```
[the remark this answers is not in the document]
jxu8: Sure, but should we rename this class to MutationalProfile to avoid confusion with the MutationalSignature bean?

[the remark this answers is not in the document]
n1zea144: Oh I see - this was a previous class with the same package/name.  No, you can't change without the risk of  breaking existing code - you would have to also update the old code to use the new name.  I would rename the other class - MutationalSignatureMetadata

core/src/main/java/org/mskcc/cbio/portal/model/MutationalSignature.java
```diff
@@ -0,0 +1,46 @@
+package org.mskcc.cbio.portal.model;
```
n1zea144: We should rename this class MutationalSignatureMeta to avoid clash with existing MutationalSignature class.

model/src/main/java/org/cbioportal/model/MutationalSignature.java
```diff
@@ -0,0 +1,56 @@
+package org.cbioportal.model;
```
n1zea144: This class doesn't need to be in this PR - you have the other class in the core module.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureData.java
```diff
@@ -0,0 +1,196 @@
...
+            boolean replace = commandLine.hasOption("replace");
+
+            ProgressMonitor.setConsoleMode(true);
+
+            File mutationalSignatureFile;
+            int numLines;
+            if(commandLine.hasOption("mut-sig-file")) {
+                mutationalSignatureFile = new File((String) commandLine.getOptionValue("mut-sig-file"));
+
+                System.out.println("Reading gene data from:  " + mutationalSignatureFile.getAbsolutePath());
```
n1zea144: This should be reading mutational signature metadata....

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureData.java
```diff
@@ -0,0 +1,196 @@
...
+                    //assume mutational signature id and description are first two columns in file
+                    MutationalSignature mutationalSignature = new MutationalSignature();
+                    mutationalSignature.setMutationalSignatureId(parts[0]);
+                    mutationalSignature.setDescription(parts[1]);
+
+                    //check if mutational signature already exists by the mutational signature id
+                    MutationalSignature existingMutationalSignature = DaoMutationalSignature.getMutationalSignatureById(mutationalSignature.getMutationalSignatureId());
+
+                    // only replaces description
+                    if (replace) {
```
n1zea144: This looks backwards - you should check if the mutational signature exists and if it does, update only if replace is true.

if (existingMutationalSignature)
{
if (replace) replace;
}
// signature does not exist, add it
else {
add to database
}

core/src/main/java/org/mskcc/cbio/portal/oncoPrintSpecLanguage/MutationalSignature.java
```diff
@@ -0,0 +1,56 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+import javax.validation.constraints.NotNull;
+
+public class MutationalSignature implements Serializable {
```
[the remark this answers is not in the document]
n1zea144: I'm still not sure why this class is here

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,184 @@
...
+                    
+                    if (existingMutationalSignature == null){ //add a new mutational signature if database does not contain it
+                        ProgressMonitor.setCurrentMessage("Adding: " + mutationalSignatureMeta.getMutationalSignatureId());
+                        DaoMutationalSignature.addMutationalSignature(mutationalSignatureMeta);
+                    }
+                    else{//if it already exists, either replace description or do nothing
+                        if (replace){
+                            ProgressMonitor.setCurrentMessage("Updating mutational signature: " + parts[0]);
+                            DaoMutationalSignature.updateMutationalSignature(mutationalSignatureMeta);
+                            }
```
n1zea144: I would add a ProgressMonitor statement that says signature exists, but replace flag is false, so skipping.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,184 @@
...
+import joptsimple.OptionSet;
+import joptsimple.OptionSpec;
+
+import org.apache.commons.cli.*;
+
+import java.io.*;
+import java.util.*;
+
+
+public class ImportMutationalSignatureMetaData extends ConsoleRunnable {
```
n1zea144: Can we rename this ImportMutationalSignatureMetadata

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,184 @@
...
+            CommandLineParser parser = new DefaultParser();
+            CommandLine commandLine = null;
+            commandLine = parser.parse(options, args);
+
+            if(commandLine.hasOption( "h" ) ){
+                System.out.println("Command line arguments are: 'h' for help, 'mut-sig-meta-file' for the mutational signature metadata file, 'replace' to replace data already in the database");
+            }
+            
+            // Check options
+            boolean replace = commandLine.hasOption("replace");
```
n1zea144: Can we change "replace" to "r", like "-r"

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,184 @@
...
+
+            ProgressMonitor.setConsoleMode(true);
+
+            File mutationalSignatureMetaFile;
+            int numLines;
+            if(commandLine.hasOption("mut-sig-meta-file")) {
```
n1zea144: change "mut-sig-meta-file" to "metafile"

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,184 @@
...
+    @Override
+    public void run() {
+        try {
+            String progName = "ImportMutationalSignatureMetaData";
+            String description = "Import mutational signature metadata file.";
+
+            Options options = ImportMutationalSignatureMetaData.getOptions(args);
+            CommandLineParser parser = new DefaultParser();
+            CommandLine commandLine = null;
+            commandLine = parser.parse(options, args);
```
n1zea144: Check if no arguments have been passed to the routine (like if argument count == 0) and if so, display the help message and exit.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,184 @@
...
+
+            if(commandLine.hasOption( "h" ) ){
```
n1zea144: Move this message into a "help" method, which can be called if the 'h' arg is passed or if no arguments have been passed to the routine.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMutationalSignatureMetaData.java
```diff
@@ -0,0 +1,196 @@
...
+import org.mskcc.cbio.portal.model.MutationalSignatureMeta;
+import org.mskcc.cbio.portal.util.*;
+
+import org.apache.commons.cli.*;
+
+import java.io.*;
+import java.util.*;
+
+
+public class ImportMutationalSignatureMetadata extends ConsoleRunnable {
```
zhx828: @n1zea144 do we actually need this script? The meta data already included in the migration file. The purpose of this script is for future update?

n1zea144: If I recall, this is to  import metadata into a fresh database (or update data).

PR #2407, opened 2017-05-03
Fix #2406

j-hudecek:
ternary operator has higher precedence than concatenation

# What? Why?
Fix #2406

Changes proposed in this pull request:
- fixing syntax error in generated sql

# Checks
- [x] Runs on Heroku.
- [x] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [x] Follows the [Google Style Guide](https://github.com/google/styleguide).
- [x] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)
- [x] If this is a feature, the PR is to rc. If this is a bug fix, the PR is to
  hotfix.

# Any screenshots or GIFs?
If this is a new visual feature please add a before/after screenshot or gif
here with e.g. [GifGrabber](http://www.gifgrabber.com/).

# Notify reviewers
Read our [Pull request merging
policy](../CONTRIBUTING.md#pull-request-merging-policy). If you are part of the
cBioPortal organization, notify the approprate team (remove inappropriate):

If you are not part of the cBioPortal organization look at who worked on the
file before you. Please use `git blame <filename>` to determine that
and notify them here:
@n1zea144

PR #4414, opened 2018-06-28
fixing #2406

### LINKED ISSUE #2406: Loading of CNA overview on Study View never finishes
When opening the Study overview of a study with CNA data the cna overview never finishes loading. There is a request to cna.json? which receives 500 with this exception:

    HTTP Status 500 - org.mskcc.cbio.portal.dao.DaoException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') AND ALTERATION IN (-2,2) AND SAMPLE_ID in ('2','3','4','5','6','7','8','9','10' at line 1

with root cause

    org.mskcc.cbio.portal.dao.DaoException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') AND ALTERATION IN (-2,2) AND SAMPLE_ID in ('2','3','4','5','6','7','8','9','10' at line 1
	org.mskcc.cbio.portal.dao.DaoCnaEvent.getCnaEvents(DaoCnaEvent.java:176)
	org.mskcc.cbio.portal.servlet.CnaJSON.processGetCnaRequest(CnaJSON.java:135)

Looking at the code that constructs the query (around line 153):

                    + " AND sample_cna_event.CNA_EVENT_ID=cna_event.CNA_EVENT_ID"
                    + (entrezGeneIds==null?"":" AND ENTREZ_GENE_ID IN(" + StringUtils.join(entrezGeneIds,",") + ")")
                    + " AND ALTERATION IN (" + StringUtils.join(cnaLevels,",") + ")"

if entrezGeneIds == null then we get ""+ StringUtils.join(entrezGeneIds,",") + ")") 
That is just the ")" that causes the syntax error

jjgao:
moving code in #2407 to fix #2406


jjgao:
@pieterlukasse could you help review this code?

pieterlukasse:
hi @jjgao, sorry for the late reply. I think this is basically dead code now. Maybe we can remove the method instead?

jjgao:
@n1zea144 @sheridancbio is this code dead?

n1zea144:
@jjgao Just checked - seems it was only used by CnaJSON.java, so it is dead.

PR #4057, opened 2018-04-03
Update to structural variant data format

sandertan:
# What? Why?
This new format will have support for detailed fusion data, with the goal of drawing fusion breakpoint events in a new fusion tab. 

This PR contains 3 main commits, which were reviewed in separate PRs.
- Backend changes to DB schema, importer, and validator by @sandertan, reviewed in this PR.
- Backend changes to the API by @dionnezaal, reviewed and approved here: https://github.com/thehyve/cbioportal/pull/25
- Display fusion tab based on genetic profiles of a study by @rnugraha, reviewed and approved here https://github.com/thehyve/cbioportal/pull/31

**RFC:** [RFC 31_b](https://docs.google.com/document/d/1tzSW2iB59bcns4xNCSJjMTxaX2uavCILDD4lo3HwcE4/edit#)

## Backend changes to DB schema, importer, and validator
- Update database schema
- Create importer for structural variants
- Create validator for structural variants
- Add specific validation for fusion events
- Add representable test data to `study_es_0`
- Remove old structural variant functionality (which was not used)

**Column names:** New, old and updated column names in all separate layers (data, database, java) can be found in [this Google Sheets doc](https://docs.google.com/spreadsheets/d/1Xc1QrA4cBynxrG-s7rXJv49QBI0Mjavz5gRfApYsGpI/edit?usp=sharing).

This is a fresh PR, after closing the slightly cluttered PR https://github.com/cBioPortal/cbioportal/pull/3398.

PR #4441, opened 2018-07-09
Support fusion GML events import

angelicaochoa:
# What? Why?
Hack for supporting germline fusion/structural variant events. The idea is to be able to annotate these events with red "Germline" sup-script next to the protein change.

Example 
screen shot 2018-07-09 at 4 30 25 pm

- Override mutation status for fusion events to value in `Fusion_Status` column if available

# Checks

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/maf/FusionFileUtil.java
+++ b/core/src/main/java/org/mskcc/cbio/maf/FusionFileUtil.java
@@ -44,74 +44,76 @@ import java.util.HashMap;
  */
 public class FusionFileUtil
 {
-	public static final String FUSION = "Fusion";
-	public static final String DNA_SUPPORT = "DNA_support";
-	public static final String RNA_SUPPORT = "RNA_support";
-	public static final String METHOD = "Method";
-	public static final String FRAME = "Frame";
+    public static final String FUSION = "Fusion";
+    public static final String DNA_SUPPORT = "DNA_support";
+    public static final String RNA_SUPPORT = "RNA_support";
+    public static final String METHOD = "Method";
+    public static final String FRAME = "Frame";
+    public static final String FUSION_STATUS = "Fusion_Status";
 
-	// number of headers in the header line
-	private int headerCount;
+    // number of headers in the header line
+    private int headerCount;
 
-	// mapping for all column names (both standard and custom columns)
-	private HashMap<String, Integer> columnIndexMap;
+    // mapping for all column names (both standard and custom columns)
+    private HashMap<String, Integer> columnIndexMap;
 
-	/**
-	 * Constructor.
-	 *
-	 * @param headerLine    Header Line.
-	 */
-	public FusionFileUtil(String headerLine)
-	{
-		// init column index map
-		this.columnIndexMap = new HashMap<String, Integer>();
+    /**
+     * Constructor.
+     *
+     * @param headerLine    Header Line.
+     */
+    public FusionFileUtil(String headerLine)
+    {
+        // init column index map
+        this.columnIndexMap = new HashMap<String, Integer>();
 
-		// split header names
-		String parts[] = headerLine.split("\t");
+        // split header names
+        String parts[] = headerLine.split("\t");
 
-		// update header count
-		this.headerCount = parts.length;
+        // update header count
+        this.headerCount = parts.length;
 
-		// find required header indices
-		for (int i=0; i<parts.length; i++)
-		{
-			String header = parts[i];
+        // find required header indices
+        for (int i=0; i<parts.length; i++)
+        {
+                String header = parts[i];
 
-			// put the index to the map
-			this.columnIndexMap.put(header.toLowerCase(), i);
-		}
-	}
+                // put the index to the map
+                this.columnIndexMap.put(header.toLowerCase(), i);
+        }
+    }
 
-	public FusionRecord parseRecord(String line)
-	{
-		String parts[] = line.split("\t", -1);
+    public FusionRecord parseRecord(String line)
+    {
+        String parts[] = line.split("\t", -1);
 
-		FusionRecord record = new FusionRecord();
+        FusionRecord record = new FusionRecord();
 
-		record.setHugoGeneSymbol(TabDelimitedFileUtil.getPartStringAllowEmpty(this.getColumnIndex(MafUtil.HUGO_SYMBOL), parts));
-		record.setEntrezGeneId(TabDelimitedFileUtil.getPartLong(this.getColumnIndex(MafUtil.ENTREZ_GENE_ID), parts));
-		record.setCenter(TabDelimitedFileUtil.getPartString(this.getColumnIndex(MafUtil.CENTER), parts));
-		record.setTumorSampleID(TabDelimitedFileUtil.getPartString(this.getColumnIndex(MafUtil.TUMOR_SAMPLE_BARCODE), parts));
-		record.setFusion(TabDelimitedFileUtil.getPartString(this.getColumnIndex(FUSION), parts));
-		record.setDnaSupport(TabDelimitedFileUtil.getPartString(this.getColumnIndex(DNA_SUPPORT), parts));
-		record.setRnaSupport(TabDelimitedFileUtil.getPartString(this.getColumnIndex(RNA_SUPPORT), parts));
-		record.setMethod(TabDelimitedFileUtil.getPartString(this.getColumnIndex(METHOD), parts));
-		record.setFrame(TabDelimitedFileUtil.getPartString(this.getColumnIndex(FRAME), parts));
+        record.setHugoGeneSymbol(TabDelimitedFileUtil.getPartStringAllowEmpty(this.getColumnIndex(MafUtil.HUGO_SYMBOL), parts));
+        record.setEntrezGeneId(TabDelimitedFileUtil.getPartLong(this.getColumnIndex(MafUtil.ENTREZ_GENE_ID), parts));
+        record.setCenter(TabDelimitedFileUtil.getPartString(this.getColumnIndex(MafUtil.CENTER), parts));
+        record.setTumorSampleID(TabDelimitedFileUtil.getPartString(this.getColumnIndex(MafUtil.TUMOR_SAMPLE_BARCODE), parts));
+        record.setFusion(TabDelimitedFileUtil.getPartString(this.getColumnIndex(FUSION), parts));
+        record.setDnaSupport(TabDelimitedFileUtil.getPartString(this.getColumnIndex(DNA_SUPPORT), parts));
+        record.setRnaSupport(TabDelimitedFileUtil.getPartString(this.getColumnIndex(RNA_SUPPORT), parts));
+        record.setMethod(TabDelimitedFileUtil.getPartString(this.getColumnIndex(METHOD), parts));
+        record.setFrame(TabDelimitedFileUtil.getPartString(this.getColumnIndex(FRAME), parts));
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
```diff
@@ -138,6 +139,10 @@ public void importData() throws IOException, DaoException {
                     // TODO we may need get mutation type from the file
                     // instead of defining a constant
                     mutation.setMutationType(FUSION);
+                    // override default "Mutation_Status" if germline data
+                    if (isGermlineData) {
+                        mutation.setMutationStatus("GERMLINE");
```
mandawilson: "GERMLINE" string should probably be a static variable like FUSION.

core/src/main/java/org/mskcc/cbio/portal/model/GeneticAlterationType.java
```diff
@@ -37,7 +37,9 @@
 public enum GeneticAlterationType {
     MUTATION_EXTENDED,
     FUSION,
+    FUSION_GML,
     STRUCTURAL_VARIANT,
+    STRUCTURAL_VARIANT_GML,
```
mandawilson: Is STRUCTURAL_VARIANT_GML used anywhere?

angelicaochoa:
@mandawilson not yet but there are plans to support structural variant data soon(ish) so I added it preemptively

core/src/main/java/org/mskcc/cbio/portal/model/GeneticAlterationType.java
angelicaochoa: @mandawilson not yet but there are plans to support structural variant data soon(ish) so I added it preemptively

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
```diff
@@ -50,15 +50,18 @@
...
     public static final String FUSION = "Fusion";
+    private final String GERMLINE = "GERMLINE";
     private File fusionFile;
     private int geneticProfileId;
     private String genePanelID;
+    private GeneticAlterationType geneticAlterationType;
     private Set<String> sampleSet  = new HashSet<>();
 
-    public ImportFusionData(File fusionFile, int geneticProfileId, String genePanelID) {
+    public ImportFusionData(File fusionFile, int geneticProfileId, String genePanelID, GeneticAlterationType geneticAlterationType) {
```
sheridancbio: I hope this does not break any other clients ... in particular, our pipeline importer code

angelicaochoa: @sheridancbio shouldn't break our importer.. the import class is called with arguments 
```--data <datafile> --meta <metafile>``` so the way it's constructed here

**Comment by sheridancbio** [comment]
I notice that core/src/main/java/org/mskcc/cbio/portal/util/ProfileMerger.java has a list of most other profiles, but does not mentions fusions at all. I'm not sure if this code is dead ... maybe it just doesn't support fusions? not sure.

**Comment by sheridancbio** [comment]
There is also a comment in model/src/main/java/org/cbioportal/model/MolecularProfile.java which says that this class is copied from GeneticAlterationType and that these classes should be kept in sync. Do we know what this is and why it was copied?

**Comment by angelicaochoa** [comment]
@mandawilson @sheridancbio this is the first I've learned about `ProfileMerger`. Not sure what that's for

**Comment by sheridancbio** [comment]
A lot of the new API seems to use MolecularProfile ... maybe we won't have profiles of FUSION_GML, so maybe we don't need the new API to service this GeneticAlterationType? is that right @angelicaochoa ?

**Comment by sheridancbio** [comment]
@angelicaochoa :

I did search through the code for switch-like code across GeneticAlterationType and did not find anything that included fusions. I did see that model/src/main/java/org/cbioportal/model/MolecularProfile.java has an enum copied from core/src/main/java/org/mskcc/cbio/portal/model/GeneticAlterationType.java (I think cloning code in this way was a bad idea .. but it has been done) ... and it seems like the intention was that changes to GeneticAlterationType should be mirrored in the member Enum inside MolecularProfile.java ... so I would recommend that we add FUSION_GML into that enum as well.

And in service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java, there is a check for {MolecularAlterationType.MUTATION_EXTENDED, MolecularAlterationType.FUSION} which throws a MolecularProfileNotFoundException if the profile is one of these types. Perhaps a clause should be added for FUSION_GML there too. That is is the only reference to MolecularAlterationType.FUSION that I saw.

I know that we will not be creating profiles of type FUSION_GML ... at least not yet. So there probably would not be any problems encountered by the cbioportal backend services. But hypothetically sometime in the future someone may request that we see profiles of germline fusions, so if this is a true GeneticAlterationType we may as well support it in the same way we support regular (tumor sample) fusions. (which also are a potential future profile feature - currently we import fusions under mutation profiles I think)

**Comment by sheridancbio** [comment]
Maybe we should get comments from @ersinciftci or @inodb about the appropriateness of adding GeneticAterationType.FUSION_GML to org.cbioportal.model.MolecularProfile.MolecularAlterationType ... to keep these two classes in sync. (see previous comment)

**Comment by angelicaochoa** [comment]
@sheridancbio I updated `model/src/main/java/org/cbioportal/model/MolecularProfile.java` and `service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java` with the new genetic alteration type. Tests are passing.. I've requested a review from @ersinciftci as well to get additional input. Thanks!

**Comment by angelicaochoa** [comment]
@jjgao yeah I thought about that too but the ImportFusionData class needs to know what datatype it is since Fusion data gets imported under the mutations profile

**Comment by jjgao** [comment]
@angelicaochoa thanks. I realized that too and was edit my comments above. I think we should add a column to the fusion data format (Fusion_Status) to indicate whether a fusion event is somatic or germline. It would be a much smaller change than adding a new data type.

**Comment by angelicaochoa** [comment]
@jjgao I thought the same thing actually - I guess I should have opened up that question for discussion with others but would definitely but less code change

**Comment by angelicaochoa** [comment]
@jjgao sounds good to me! I’ll get to it tomorrow. Thanks!

**Comment by jjgao** [comment]
@angelicaochoa thank you! I think that will also address @sheridancbio's comment on sync with MolecularDataServiceImpl.

**Comment by angelicaochoa** [comment]
@sheridancbio @mandawilson @jjgao updated PR to parse a new column `Fusion_Status` and set it to the mutation record `Mutation_Status` for import into DB table. 

In a moment I will update the documentation as well regarding new/option `Fusion_Status` col for `data_fusions.txt`

File: core/src/main/java/org/mskcc/cbio/maf/FusionFileUtil.java [hunk 4]
```diff
@@ -44,74 +44,77 @@
...
+        this.headerCount = parts.length;
+
+        // find required header indices
+        for (int i=0; i<parts.length; i++)
+        {
+                String header = parts[i];
+
+                // put the index to the map
+                this.columnIndexMap.put(header.toLowerCase(), i);
+        }
```
sheridancbio: maybe after putting all the found headers into the index map, we should add a check that all expected headers were found, and throw an exception if not.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
```diff
@@ -138,6 +138,7 @@ public void importData() throws IOException, DaoException {
                     // TODO we may need get mutation type from the file
                     // instead of defining a constant
                     mutation.setMutationType(FUSION);
+                    mutation.setMutationStatus(record.getFusionStatus().toUpperCase());
```
sheridancbio: a little confusing .. (set mutation status to fusion status) .. but the real issue is presenting fusions under datatype mutation. We are lucky that we can reuse the "germline" flag of mutations for this datatype.

docs/File-Formats.md
```diff
@@ -783,6 +783,7 @@ A fusion data file is a two dimensional matrix with one gene per row.  For each
 7. **RNA_support**: Fusion detected from RNA sequence data, "yes" or "no".
 8. **Method**: Fusion detected algorithm/tool.
 9. **Frame**: "in-frame" or "frameshift".
+10. **Fusion_Status (OPTIONAL)**: An assessment of the mutation type (i.e., "SOMATIC", "GERMLINE", "UNCALLED", etc.)
```
sheridancbio: good. (keeping docs current)

sheridancbio:
@jjgao I think we are ready to merge this PR ... do you want to clear your requested changes now that it has been reorganized? (or I can dismiss)

sandertan:
This looks like an effective solution to support this type of data.

I'm wondering how this will look in the OncoPrint. We recently added Germline visualization in the OncoPrint, but obviously we haven't tested this with a Fusion event. Could you check if it works as expected? 

Example of germline mutation in oncoprint:
http://www.cbioportal.org/index.do?session_id=5af5a304498eb8b3d56615b6

sandertan:
@angelicaochoa It would be nice to have some test data for this. Could you add this column to `data_fusions.txt` in `study_es_0` to make sure this data passes the validator & importer without any warnings/errors?

https://github.com/cBioPortal/cbioportal/blob/master/core/src/test/scripts/test_data/study_es_0/data_fusions.txt

docs/File-Formats.md
sandertan: Small comment: I'm not sure what you mean with 'etc'. Currently it's not clear if this should contain specific values or is free text. Perhaps it's best that this column has the same functionality as "Mutation_Status" in the MAF format and refer to that.


Mutation_Status in File-Formats:

26. **Mutation_Status (Optional):** "Somatic" or "Germline" are supported by the UI in Mutations tab. "None", "LOH" and "Wildtype" will not be loaded. Other values will be displayed as text.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportFusionData.java
sandertan: This looks like an effective solution to support this type of data.

I'm wondering how this will look in the OncoPrint. We recently added Germline visualization in the OncoPrint, but obviously, we haven't tested this with a Fusion event. Could you check if it works as expected?

Example of germline mutation in oncoprint:
http://www.cbioportal.org/index.do?session_id=5af5a304498eb8b3d56615b6

docs/File-Formats.md
sandertan: We should change  "in Mutations tab" to in Mutations tab and OncoPrint"

angelicaochoa:
@sandertan About the oncoprint, the events still show up as fusion events w/o any germline indication. We'll (@schultzn @n1zea144) be discussing with frontend team how we can annotate these as germline in the oncoprint soon.

angelicaochoa:
@sandertan @jjgao updated my PR, ready for review again whenever you have a chance. Thanks!

jjgao:
@angelicaochoa once we have the germline data in the database, please ask @adamabeshouse to take a look and show them as germline in oncoprint.

sandertan:
We will have to add a column for this data to https://github.com/cBioPortal/cbioportal/pull/4057 as well, but this PR doesn't conflict.

Will there be a public study in the portal with these kind of data or is this for an internal study?

jjgao:
@sandertan AFAIK, it's for internal for now.

Issue #4391, opened 2018-06-22
validateStudies.py crashes for skcm_tcga_pan_can_atlas_2018


sandertan:
This issue was reported in the Datahub channel and refers to an issue found on CircleCI when validating the SKCM TCGA PanCancer study (`skcm_tcga_pan_can_atlas_2018`) 

DataHub PR with the problematic study:
https://github.com/cBioPortal/datahub/pull/239

CircleCI error:
https://circleci.com/gh/cBioPortal/datahub/279?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I reproduced this issue locally when using the same wrapper to validate multiple studies, [`validateStudies.py`](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/scripts/importer/validateStudies.py). The study data for`skcm_tcga_pan_can_atlas_2018` can be found here: https://github.com/cBioPortal/datahub/tree/eb96197c4750ce04c09cfcb12d3806a343663588/public/skcm_tcga_pan_can_atlas_2018

Strangely enough, when validating using validateData.py on this study, there is no crash. Most likely this has something to do with how the logger is controlled in `validateStudies.py`.

fedde-s:
I can't find the validateStudies.py script in cBioPortal v1.13.2. Is it to be introduced in v1.14.0?

fedde-s:
The issue isn't just present in the master branch, it has indeed been released in v1.14.0.

fedde-s:
Pull request #4453 should fix it.

sandertan:
Thanks for the fix!

PR #4453, opened 2018-07-13
Fix the validateStudies flush-to-console crash

fedde-s:
This should fix issue #4391.

The logging framework used by the validation script does not expect to
run in the same Python interpreter process as another run, and doing
this anyway caused problems particularly when running the multi-study
validator on studies with more than a million (possible) errors. Running
the validation script in a subprocess fixes.

One small complication of not having access to the same logger was that
the study header had to be explicitly flushed to the log file before
running the validation script, so that it doesn't stay in a buffer until
a later line.


fedde-s:
Rebased on the current state of the `master` branch.

core/src/test/scripts/system_tests_validate_studies.py
```diff
@@ -80,6 +113,86 @@ def test_exit_status_multiple_studies(self):
...
+                kwargs = {k: v for k, v in kwargs.items() if k != 'capacity'}
+                # pass 3 as the capacity argument
+                super(EagerFlusher, self).__init__(3, *args, **kwargs)
+        class EagerFlushingCollapser(
+                cbioportal_common.CollapsingLogMessageHandler,
+                EagerFlusher):
+            """CollapsingLogMessageHandler with EagerFlusher overrides."""
+            pass
+        self.original_collapser = cbioportal_common.CollapsingLogMessageHandler
+        cbioportal_common.CollapsingLogMessageHandler = EagerFlusher
```
fedde-s: I realised, this mock no longer works now that `validateData` runs in its own process. The test isn't technically testing what it says it's testing any more. However, I don't see a better way to fix this than by making the buffer size configurable by passing a flag to `validateData`.

fedde-s:
Updated on top of `master` again as @sandertan requested, and made `system_tests_validate_studies.py` executable as a script for consistency.

PR #4475, opened 2018-07-19
Handle unknown status in validateStudies and decrease log cache size

sandertan:
# What? Why?
- Fix a crash when an unknown exit status occurs. 
- Decrease the size of validation log cache from 1,000,000 to 500,000. This will solve an out-of-memory issue when running the validator on CircleCI.

This is the case in https://circleci.com/gh/cBioPortal/datahub/326?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link for PR https://github.com/cBioPortal/datahub/pull/239

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -4036,7 +4036,7 @@ def main_validate(args):
     text_handler.setFormatter(
         cbioportal_common.LogfileStyleFormatter(study_dir))
     collapsing_text_handler = cbioportal_common.CollapsingLogMessageHandler(
-        capacity=1e6,
+        capacity=5e5,
         flushLevel=logging.CRITICAL,
         target=text_handler)
     collapsing_text_handler.setLevel(output_loglevel)
@@ -4078,7 +4078,7 @@ def main_validate(args):
         # TODO extend CollapsingLogMessageHandler to flush to multiple targets,
         # and get rid of the duplicated buffering of messages here
         collapsing_html_handler = cbioportal_common.CollapsingLogMessageHandler(
-            capacity=1e6,
+            capacity=5e5,
             flushLevel=logging.CRITICAL,
             target=html_handler)
         collapsing_html_handler.setLevel(output_loglevel)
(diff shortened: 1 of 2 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -4036,7 +4036,7 @@ def main_validate(args):
     text_handler.setFormatter(
         cbioportal_common.LogfileStyleFormatter(study_dir))
     collapsing_text_handler = cbioportal_common.CollapsingLogMessageHandler(
-        capacity=1e6,
+        capacity=5e5,
```
fedde-s: Now that we actually run into this limitation, this is a good time to re-evaluate [the idea to get rid of buffering before collapse altogether](https://github.com/cBioPortal/cbioportal/blob/v1.14.0/core/src/main/scripts/importer/cbioportal_common.py#L425), monoidally. Hitting the buffer capacity currently means that the message will appear in the output multiple times, being collapsed in batches. But for the short term, this seems like a good fix for the Datahub CI issues as the number was always arbitrary anyway.

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -131,11 +131,12 @@ def main(args):
             exit_status_study = 2
 
         # Check exit status and print result
-        if exit_status_study == 1 or exit_status_study == 2:
-            print '\x1b[0m' + "Result: " + '\x1b[31m' + possible_exit_status[exit_status_study] + '\x1b[0m'
-            validation_exit_status = 1  # When invalid check the exit status to one, for failing circleCI
+        exit_status_message = possible_exit_status.get(exit_status_study, 'Unknown status: {}'.format(exit_status_study))
+        if exit_status_study == 0 or exit_status_study == 3:
```
fedde-s: This line doesn't clearly express that it's checking whether the study potentially passed validation. It might be nicer to extract the condition into a variable with a name that does clarify that. But it will do.

core/src/main/scripts/importer/validateStudies.py
```diff
@@ -131,11 +131,12 @@ def main(args):
...
+            print '\x1b[0m' + "Result: %s" % exit_status_message
```
fedde-s: Moving these ANSI colour codes into a little namespace, or reusing the existing one in [metaImport.py](https://github.com/cBioPortal/cbioportal/blob/v1.14.0/core/src/main/scripts/importer/metaImport.py#L22), would make this code a *lot* cleaner, if you're touching this code anyway. But again, if you feel it's disproportionate to the scope of this PR, this will do!

sandertan:
Thanks @fedde-s for the feedback. It would indeed be nice to have these improvements. Let's move these to a backlog.

Issue #4510, opened 2018-07-23
Verify that legacy patient links redirect properly to new HTML5 links


alisman:
@inodb i'm give this to you just to get a different set of eyes on it.  

- [x] Legacy patient links work
  - [x] nav case ids
- [x] Legacy sample links works
- [x] Legacy cohort links work
- [x] Check links to patient page from other areas of app
  - [x] study view (top button, clinical data tab)
  - [x] oncoprint (mouseover tooltip)
  - [x] mutations tab (first column)
  - [x] plots tab (mouseover tooltip)

inodb:
@alisman all seems to work - i just updated some urls in prs to backend and frontend, but they work anyway cause of the rewrite.

PR #4512, opened 2018-07-24
Change case.do/#/patient urls to /patient

inodb:
Related to #4510

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -699,13 +699,13 @@ public class GlobalProperties {
 
     public static String getLinkToPatientView(String caseId, String cancerStudyId)
     {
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -694,13 +694,13 @@ public static String getGenomicBuild(){
 
     public static String getLinkToPatientView(String caseId, String cancerStudyId)
     {
-        return "case.do#/patient?caseId=" + caseId
+        return "/patient?caseId=" + caseId
```
alisman: i believe these need to relative b/c otherwise on instance nested in a subfolder, user will get taken to root

inodb: @alisman seems like this is already being pasted after full url of instance (see references to `getLinkToPatientView`

PR #4526, opened 2018-07-30
Update gene set importer and documentation

sandertan:
- Remove unnecessary info message
- Fix documentation

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenesetHierarchy.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenesetHierarchy.java
@@ -68,25 +68,19 @@ public class ImportGenesetHierarchy extends ConsoleRunnable {
 						"'data' argument required");
 			}
 
-
 			File genesetFile = new File(options.valueOf(data));
-
 			ProgressMonitor.setCurrentMessage("Input file:\n" + genesetFile.getPath() + "\n");
 
-			// Check if geneset_hierarchy_node already filled
-			boolean emptyDatabase = !DaoGenesetHierarchyNode.checkGenesetHierarchy();
-			if (emptyDatabase) {
-				ProgressMonitor.setCurrentMessage("Table `geneset_hierarchy_node` is empty.\n");
-			} else {
-				ProgressMonitor.setCurrentMessage("Table `geneset_hierarchy_node` is not empty.\n");
-			}
-
-			// First we want to validate that the gene sets we're adding, are in database.
+			// First validate if gene sets in the hierarchy file, are in database. Therefore start an import process
+			// with validation set to 'true'. This will mock the import process and checks the gene sets.
 			boolean validate = true;
 			importData(genesetFile, validate);
 
-			// Make the database empty
-			if (!emptyDatabase) {
+			// Check if gene set hierarchy tables are already filled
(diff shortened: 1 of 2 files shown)


docs/File-Formats.md
```diff
@@ -1175,12 +1175,12 @@ geneset_def_version: msigdb_6.1
 ### GSVA score data file
 The data file will be a simple tab separated format, similar to the expression data file: each sample is a column, each gene set a row, each cell contains the GSVA score for that sample x gene set combination.
 
-The first column is the GENESET_ID. This contains the EXTERNAL_ID or "stable id" (MsigDB calls this "standard name") of the gene set. The other colums are the sample columns: An additional column for each sample in the dataset using the sample id as the column header.
+The first column is the `geneset_id`. This contains the name or "stable id" (MSigDB calls this "standard name") of the gene set. The other columns are sample columns: An additional column for each sample in the dataset using the sample id as the column header.
```
oplantalech: > this contains the name or "stable id"

I guess you mean: this contains the name "stable id"

docs/File-Formats.md
```diff
@@ -1219,12 +1219,12 @@ geneset_def_version: msigdb_6.1
 ### GSVA p-value data file
 The data file will be a simple tab separated format, similar to the score file: each sample is a column, each gene set a row, each cell contains the p-value for the score found for sample x gene set combination.
 
-The first column is the GENESET_ID. This contains the EXTERNAL_ID or "stable id" (MsigDB calls this "standard name") of the gene set. The other colums are the sample columns: An additional column for each sample in the dataset using the sample id as the column header.
+The first column is the `geneset_id`. This contains the name or "stable id" (MSigDB calls this "standard name") of the gene set. The other columns are sample columns: An additional column for each sample in the dataset using the sample id as the column header.
```
oplantalech: Same error as previous comment.

docs/File-Formats.md
sandertan: @oplantalech I simplified it.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenesetHierarchy.java
```diff
@@ -75,11 +75,6 @@ public void run() {
 
 			// Check if geneset_hierarchy_node already filled
 			boolean emptyDatabase = !DaoGenesetHierarchyNode.checkGenesetHierarchy();
-			if (emptyDatabase) {
```
fedde-s: Would moving the `emptyDatabase` boolean down to the line on which it is needed break things? And if so, could you add a comment on how `importData()` (which apparently validates something) does interact with `checkGenesetHierarchy()`?

sandertan:
Thanks for the feedback @fedde-s . I improved the readability, can you review again?

sandertan:
Hi @n1zea144 thanks for looking at this PR. I don't think this is a common use case (yet). 

For the related datatype, the gene sets, I added an option to force remove all existing gene sets and add the new ones using a very long parameter(`boolean confirmEmptyGenesets = options.has("confirm-delete-all-genesets-hierarchy-genesetprofiles");`

If this is every required for gene set hierarchy, we can easily add it. Because this hierarchy data is portal wide, we thought the prompt is useful for the most common use case.

n1zea144:
@sandertan Sounds good.

sandertan:
Rebased on master fixed ci.

PR #4535, opened 2018-07-30
Enable basic authentication for session service

inodb:
When properties session.service.user and session.service.password are provided
enable basic authentication for the session service proxy.

Requires this to be merged: https://github.com/cBioPortal/session-service/pull/16

TODO:
- [x] Util and Controller class seem like they should just be merged? Little hard to understand what the diff is between the two

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/SessionServiceUtil.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/SessionServiceUtil.java
@@ -32,13 +32,22 @@
 
 package org.mskcc.cbio.portal.util;
 
+import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.mskcc.cbio.portal.model.virtualstudy.VirtualStudy;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.httpclient.HttpException;
 
+
 import com.fasterxml.jackson.core.type.TypeReference;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
 import org.springframework.web.client.HttpStatusCodeException;
 import org.springframework.web.client.RestTemplate;
 
@@ -57,10 +66,13 @@ import java.net.MalformedURLException;
 
 import java.util.Map;
 
+import org.mskcc.cbio.portal.web.ProxySessionServiceController;;
+
 /**
  *
  * @author Manda Wilson
  */
+
 public class SessionServiceUtil {
 
     private static Log LOG = LogFactory.getLog(SessionServiceUtil.class);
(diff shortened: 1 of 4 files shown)


inodb:
@kalletlak do you know how to util and controller class are related. Seems to me like we should just use the one controller class, no?

kalletlak:
@inodb `SessionServiceUtil` is used in servlet controllers. we can get rid of this once we have SPA.
`ProxySessionServiceController` is exposed as an api and is directly used in the front-end

core/src/main/java/org/mskcc/cbio/portal/util/SessionServiceUtil.java
```diff
@@ -38,7 +38,15 @@
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.httpclient.HttpException;
 
+
```
inodb: we can get rid of this file once we have SPA. Code here is kinda hacky with some duplication, but we'll remove the file later so not a big deal imo

zhx828:
@kalletlak could you do a review again?

web/src/main/java/org/mskcc/cbio/portal/web/ProxySessionServiceController.java
```diff
@@ -75,16 +93,33 @@
         }
         return null;
     }
+
+    public HttpHeaders createHeaders(){
```
kalletlak: may be we have condition inside the object initialization, something like
```
private HttpHeaders createHeaders(){
	return new HttpHeaders() {{
		if (sessionServiceUser != null && !sessionServiceUser.equals("") && sessionServicePassword != null && !sessionServicePassword.equals("")) {
			String auth = sessionServiceUser + ":" + sessionServicePassword;
			byte[] encodedAuth = Base64.encodeBase64(auth.getBytes(Charset.forName("US-ASCII")));
			String authHeader = "Basic " + new String(encodedAuth);
			set( "Authorization", authHeader);
		}
		set( "Content-Type", "application/json");
	}};
}
```
and no need to check for null at the methods invoking it

inodb: @kalletlak 	good point - thanks for reviewing! Updated it

web/src/main/java/org/mskcc/cbio/portal/web/ProxySessionServiceController.java
```diff
@@ -198,7 +237,11 @@ private void updateVirutalStudyUser(String              id,
                 }
             }
             virtualStudy.getData().setUsers(users);
-            new RestTemplate().put(sessionServiceURL + SessionType.virtual_study + "/" + id, virtualStudy.getData());
+            RestTemplate restTemplate = new RestTemplate();
+            HttpEntity<Object> httpEntity = null;
+            httpEntity = new HttpEntity<Object>(virtualStudy.getData(), getHttpHeaders());
```
kalletlak: directly initialize it 
```HttpEntity<Object> httpEntity = new HttpEntity<Object>(virtualStudy.getData(), getHttpHeaders());```

inodb: good catch - fixed

inodb:
doing e2e test here: https://github.com/cBioPortal/cbioportal-frontend/pull/1678

PR #4536, opened 2018-07-31
Merge master to release 1.15.0

sheridancbio:
# Resolves conflict during merge of News.md


docs/File-Formats.md
```diff
@@ -350,6 +358,14 @@ data_filename: data_log2CNA.txt
 
 The log2 copy number data file follows the same format as expression data files.  See [Expression Data](#expression-data) for a description of the expression data file format.
 
+#### GISTIC2 Format
```
inodb: @sheridancbio yeah let's maybe remove one of these GISTIC2 ones - fine in subsequent PR

sheridancbio: I see that there is one GISTIC2 section in the Discrete CNA section, and another in the Continuous CNA section. Maybe this note is being put in both places because GISTIC2 might be used to produce both discrete CNA calls and also continuous CNA calls? (so the reader might want the info when reading either of the larger sections?)

inodb: @jjgao can gistic2 results be used both for continuous cna data as well as discrete?

inodb: Filed an issue for this so we can merge it: https://github.com/cBioPortal/cbioportal/issues/4564

jjgao: Yes, gistic2 results contains both discrete and continous cna data.

PR #4558, opened 2018-08-01
Move TUMOR_SITE to patient attributes

sandertan:
# What? Why?
Change the following attribute to be a PATIENT attribute instead of a SAMPLE attribute
- `TUMOR_SITE`

This was requested by @yichaoS to let several studies pass validation on CircleCI, such as `blca_tcga_pub_2017`, see:
https://384-63335718-gh.circle-artifacts.com/0/test-reports/ERRORS/blca_tcga_pub_2017-validation.html
I have **not** tested if the front end works as expected when these data types are in the clinical patient file.

Also, File-Formats is updated to reflect these changes. @jjgao @n1zea144 please comment if you agree with this data format.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1948,7 +1948,7 @@ class ClinicalValidator(Validator):
         'TUMOR_GRADE': {
         },
         'TUMOR_SITE': {
-            'is_patient_attribute': '0',
+            'is_patient_attribute': '1',
             'datatype': 'STRING',
         },
         'TUMOR_STAGE_2009': {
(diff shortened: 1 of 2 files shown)


jjgao:
@yichaoS @ritikakundra Maybe we should not reuse the same attribute for both patient and sample b/c they may have different meanings and will create issues for cross-study analysis. If we need to use them in patients, should we just have different attributes, e.g. primary_site_patient?

yichaoS:
@sandertan @jjgao @ritikakundra  Do we have examples of these attributes ever being used for patients?

ritikakundra:
@sandertan only tumor_site needs to be changed to patient based. Rest all are sample level.

sandertan:
@ritikakundra done.

fedde-s:
New strict rules for some of these attributes from #3910 seem to have been included in the `release-1.15.0` branch. Should these amendments be included in that release as well, so that the release doesn't break compatibility with pre-existing studies?

fedde-s:
Never mind! I see that it's already directed to `master`.

Issue #4531, opened 2018-07-30
optional log in: is it possible to stay logged in betweens sessions?


jjgao:
just similar to Gmail...

jjgao:
@zhx828 do we need to change the tomcat start script?

zhx828:
@jjgao yes, but only when we deploy the optional login. @inodb I will remind you again when we decide to release optional login. Then you could update our deployment script by adding `-Dtomcat.session.timeout=43200` This is a 30 days timeout.

PR #4560, opened 2018-08-01
Update session timeout from command line

zhx828:
You can set it to -1 to keep session always alive.
This solves issue https://github.com/cBioPortal/cbioportal/issues/4531

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?
@cBioPortal/backend


pom.xml
```diff
@@ -204,6 +204,7 @@
     <maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
     <jacoco-maven-plugin.version>0.7.6.201602180812</jacoco-maven-plugin.version>
     <mockito.version>1.10.19</mockito.version>
+    <session.timeout>720</session.timeout>
```
n1zea144: Can you prefix this with "tomcat" and move it near the other tomcat property:

[https://github.com/zhx828/cbioportal/blob/999f892fae96698f154752875aad5fd1eaf7d572/pom.xml#L218](https://github.com/zhx828/cbioportal/blob/999f892fae96698f154752875aad5fd1eaf7d572/pom.xml#L218)

n1zea144: Can you prefix this with "tomcat" and move it near the other tomcat property?
[https://github.com/zhx828/cbioportal/blob/999f892fae96698f154752875aad5fd1eaf7d572/pom.xml#L218](https://github.com/zhx828/cbioportal/blob/999f892fae96698f154752875aad5fd1eaf7d572/pom.xml#L218)

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -216,3 +216,4 @@ oncoprint.defaultview=patient
 # Disable cache (for non authorized portals this can sometimes avoid running
 # out of memory with relatively little performance loss)
 # mybatis.cache.enabled=false
+
```
n1zea144: I missed this - do you want to remove this file from the changeset?

zhx828: It probably caused due to my previous implementation. Let me update.

Issue #4557, opened 2018-07-31
Optimize clinical track data counts API and reenable it in oncoprint


schultzn:
I am testing the new clinical track feature in OncoPrints on a slow internet, and the track availability feature seems to be very slow. OncoPrint loads very quickly, but the clinical tracks don't show up for large studies (internal MSK-IMPACT), even when selecting just a subset of the data set.

![image](https://user-images.githubusercontent.com/11892704/43488652-69f8fee6-94e8-11e8-9200-eb586b198cbb.png)

jjgao:
@adamabeshouse let's discuss this and see there is a way to improve the API performance.

jjgao:
If not we should exclude this from this release.

adamabeshouse:
yes the API is very slow, that's the same reason I excluded clinical track availability from the plots tab, I couldnt find a way to decouple it from the general tab loading. we need @ersinciftci

adamabeshouse:
i just merged a PR disabling this functionality. I changed the title of this issue so that its reenabling it later

ersinciftci:
The problem with this endpoint is that we can't get a count directly from the database because it's a comma separated list in the database, so we need to get all the data, parse, exclude 'NA's and then count. so it's not faster than getting the data itself, even though we return only the count. I'll profile and see if there is any fixable bottleneck in the Molecular Data endpoints in general.

ersinciftci:
Actually seems like I misunderstood, I thought this was related to https://github.com/cBioPortal/cbioportal/issues/4431, but apparently it's related to Clinical Data, not Molecular Data. I'll check.

PR #4587, opened 2018-08-06
Change LEFT JOINs to INNER JOINs in clinical attribute count endpoint

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4557


jjgao:
@ersinciftci I think you need to change some unit tests.

PR #4611, opened 2018-08-09
Caches needed data for security so mybatis cache can be disabled.

n1zea144:
This code boosts performance in password-protected instances of portal by pre-caching data required by the CancerStudyPermissionEvaluator during permission checking, including molecularProfiles, CancerStudies, and sampleLists.  We avoided use of MyBatis cache layer is it is still suspect in causing crashes to tomcat servers, especially when queries contain large numbers of samples.

We also gave preferential treatment to performance by pre-caching cancer studies over the risk of missing a cancer study group update (which can happen at MSK).  This risk seems minimal as the more common use case is the update to user authorities (which are read from the database at login) and not the update to cancer study groups.

Note, there are some repositories that are autowired into the class that are not used, but by autowiring them seem to overcoming the following exception

`
org.apache.ibatis.builder.IncompleteElementException: Could not find SQL statement to include with refid org.cbioportal.persistence.mybatis.PatientMapper.select
`
apparently an inter-mapper reference between a mapper required directly by the class and one indirectly.  After attempts at various spring/mybatis configurations, auto-wiring the repositories not directly used seem to be the only thing to address the exception.

# Checks
- [ ] Runs on Heroku.

### DIFF
--- a/security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
+++ b/security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
@@ -33,43 +33,20 @@
 package org.cbioportal.security.spring;
 
 // imports
+import java.util.*;
 import java.io.Serializable;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.ArrayList;
-import java.util.Collection;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
+import org.apache.commons.logging.*;
 
-import org.cbioportal.model.CancerStudy;
-import org.cbioportal.model.MolecularProfile;
-import org.cbioportal.model.SampleList;
-import org.cbioportal.persistence.MolecularProfileRepository;
-import org.cbioportal.persistence.SampleRepository;
-import org.cbioportal.persistence.SampleListRepository;
-import org.cbioportal.persistence.SecurityRepository;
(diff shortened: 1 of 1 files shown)


security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -410,7 +430,8 @@ private boolean hasAccessToMolecularProfiles(Authentication authentication, Coll
     {
         List<String> profileIds = (molecularProfileIds instanceof List) ?
             (List<String>)molecularProfileIds : new ArrayList<String>(molecularProfileIds);
-        for (MolecularProfile molecularProfile : molecularProfileRepository.getMolecularProfiles(profileIds, "DETAILED")) {
+        for (String molecularProfileId : profileIds) {
```
mandawilson: We switch from "DETAILED" to "SUMMARY" view here, is that OK?

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -420,7 +441,8 @@ private boolean hasAccessToMolecularProfiles(Authentication authentication, Coll
 
     private boolean hasAccessToSampleLists(Authentication authentication, List<String> sampleListIds, Object permission)
     {
-        for (SampleList sampleList : sampleListRepository.getSampleLists(sampleListIds, "DETAILED")) {
```
mandawilson: Here we also switch from "DETAILED" view to "SUMMARY".

PR #4630, opened 2018-08-13
Add documentation, tests and test data for protein data

sandertan:
# What? Why?
Importing custom protein level data can be quite difficult due to the `Composite.Element.REF` column.  This column can contain multiple gene symbols, as well as an antibody. It is currently not possible to import protein level data based on Entrez ID or gene symbol. 

# This PR adds:
- Documentation on protein level data and `Composite.Element.REF` column
- Tests and test data, which was missing for these data types. 
- To avoid the generation of 39 warnings due to phosphogenes that are added to the database on the fly, I created a info logger where the "phosphogene gene added to database" messages are stored.


sandertan:
This PR does not include (yet):
- Checks whether the data is correctly loaded in `TestIntegrationTest.java`
- Functionality for `source_stable_id` as was implemented for the z-scored expression and gene set profiles.
- Updated validator to import protein data based on solely Entrez ID or Symbol. I'm not sure if we want to allow this, but the importer supports this.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -222,7 +222,10 @@ public void testLoadStudyEs0() throws Throwable {
                 }
             }
             assertEquals(50, countGte2Lt3);
-            
+
+            //===== Check RPPA and Protein Quantification data ========
+            // TODO
```
pieterlukasse: I agree

pieterlukasse: I mean...please add it here ;)

pieterlukasse:
This PR looks good to me 👍 I can approve once the integration test is added.

jjgao:
@Sjoerd-van-Hagen is there any plan on this one? The PR is not passing travis. I tried to resolve the conflicts - hopefully I didn't mess up with anything.

Sjoerd-van-Hagen:
@jjgao not really. Sander opened this, I am not sure what gave rise to this PR to be honest. 

@sandertan care to take a look?

sandertan:
@jjgao I think the only thing failing is the system test. This can easily be checked and solved.
`test_data/study_es_0/result_report.html` is probably old. It would be best to run the system test locally, then check the output and see if all looks fine. This output can be found in `test_data/study_es_0/result_report.html~`. I usually rename this file to something like `result_report2.html` so that my browser can open it.

If all looks fine (all green, no warnings, no errors), replace `test_data/study_es_0/result_report.html` with `test_data/study_es_0/result_report.html`

PR #4639, opened 2018-08-14
Validation for hg18 segment files and gene names that start with number

sandertan:
- Store hg18, hg19 as valid reference genomes in segment data
Validation can be restored to normal when hg18 data on public portal and data hub has been
liftovered to hg19. It was decided in the data hub call of August 14 2018 to remove validation until then.

@yichaoS Can you check out this branch and test it locally with a hg19 and hg18 segment file study?

- Allow gene names that start with a number if they are in the gene table
- Also update json files with genes and gene aliases to allow testing of these gene names.


sandertan:
Tested with hg18 and hg19 segment data, and validation works as expected.

PR #4667, opened 2018-08-20
attempt to fix DaoPatient stall during reCache()

sheridancbio:
- all primitive patient records extracted from record set before cancer studies are retrieved
- primitive patient records are processed after recordset/connection is closed (cancer study objects are retrieved and Patient model objects are created)
- also removed apparently unnecessary static initializer block in DaoGeneticProfile
- removed "throws DaoException" specifier from DaoCancerStudy.getCancerStudyByInternalId() (exception is no longer possible)
- unnecessary try/catch blocks removed in related classes
- cleaned up cachePatient() function to use only the passed Patient object, which contains the needed studyId in the CancerStudy member

These fixes, and this new branch is to attempt to resolve an intermittent stalling of the import process during the recaching step when using the jdbc datasource. This branch (importer-daopatient-stall-fix) will be kept current with master (rebasing as necessary) and will be used to build the import pipeline. When it appears that the intermittent stalling has been resolved, the code will be tested for use in the production web service portals and a PR for a merge into a release branch will be opened.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoPatient.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoPatient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2015 - 2018 Memorial Sloan-Kettering Cancer Center.
  *
  * This library is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoPatient.java
```diff
@@ -60,20 +83,20 @@ private static void clearCache()
...
         Connection con = null;
         PreparedStatement pstmt = null;
         ResultSet rs = null;
         try {
             con = JdbcUtil.getDbConnection(DaoPatient.class);
             pstmt = con.prepareStatement("SELECT * FROM patient");
             rs = pstmt.executeQuery();
-            ArrayList<Patient> list = new ArrayList<Patient>();
             while (rs.next()) {
-                Patient p = extractPatient(rs);
```
averyniceday: call to getCancerStudy separated out from result set - connection closed first

sheridancbio:
I have tested import using this codebase. I still saw 2 reCache reports:

core/src/main/java/org/mskcc/cbio/portal/dao/DaoPatient.java
```diff
@@ -60,20 +83,20 @@ private static void clearCache()
     public static synchronized void reCache()
     {
         clearCache();
-
+        Vector<PrimitivePatientRecord> patientRecordsRetrieved = new Vector<>();
```
angelicaochoa: just curious.. why a vector?

sheridancbio: The Vector class is thread safe, and we use synchronized methods in DaoCancerStudy .. presumably to prevent two parallel thread from both calling reCacheAll.
So if multithreading is a worry, it seemed safer to stay with the threadsafe class (rather than ArrayList).

angelicaochoa: @sheridancbio maybe you can add a comment saying so? that's useful to know. either in the pr or here at this line?

sheridancbio: I will add comment to the code

sheridancbio: Actually, on deeper reflection, this variable is already inside a synchronized function call. And as a local variable, it would only have visibility to a single thread anyway. To say that Vector is "thread safe" really means that the read/write methods are synchronized on the object ... but we don't need that inside a synchronized method really. So I'm going to switch it to ArrayList

sheridancbio:
I have changed the Vector into an ArrayList inside DaoPatient.reCache()
I have also added a second commit where a 4 second delay is added after closing the connection on an addCancerStudy() call before we reCacheAll() ... under the theory that perhaps a large transaction needs some time to be committed/cleared before we can try re-reading/re-caching studies.

sheridancbio:
4 second delay increased to 5 seconds (after checking with @averyniceday)

PR #4675, opened 2018-08-20
Add in missing line to update db version schema

sheridancbio:
# What? Why?
The migration.sql script is missing the line which updates the schema version in table info after the changes that are made in version 2.6.0.

The appropriate line is added

It seems to me that even if someone has migrated their database to 2.6.0, it is harmless for the script to re-apply the changes between 2.6.0 and 2.6.1, so we probably do not need to warn users who may have applied the migration.sql script on a database without getting the bump to 2.6.1 in the info table.

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


angelicaochoa:
@sheridancbio i dont have write permissions so someone else will have to merge. good catch though

PR #4680, opened 2018-08-21
Add disabled tabs to config_service and add new getter to global prop…

alisman:
…s to return string from props

# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -1005,6 +1005,18 @@ public static String getDarwinRegex() {
         return (tabs.length > 0 && disabledTabs.length() > 0) ? Arrays.asList(tabs) : new ArrayList<String>();
     }
     
+    public static String getDisabledTabList() {
```
inodb: might be better to name this e.g. `getDisabledTabAsString` or something, feels weird to return String for a function named `..List()`

PR #4681, opened 2018-08-21
Double cors headers

alisman:
Double cors header is rejected by browser

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/portal/src/main/webapp/WEB-INF/web.xml
+++ b/portal/src/main/webapp/WEB-INF/web.xml
@@ -69,22 +69,7 @@
       <filter-name>CorsFilter</filter-name>
       <url-pattern>/*</url-pattern>
     </filter-mapping>
-
-    <filter>
-      <filter-name>APICorsFilter</filter-name>
-      <filter-class>org.apache.catalina.filters.CorsFilter</filter-class>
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -69,22 +69,7 @@
...
-        <param-value>${session.service.origin}</param-value>
-      </init-param>
-    </filter>
-
-    <filter-mapping>
-      <filter-name>APICorsFilter</filter-name>
-      <url-pattern>/api-legacy/proxy/session/*</url-pattern>
-    </filter-mapping>
-
+
```
alisman: session service was matching both CorsFilter url patterns and thus the filter was getting applied twice, resulting in an invalid HTTP header which was rejected by browser.  @mandawilson @kalletlak annotate shows both of you guys touched this code in last couple years.  just want to make sure i'm not missing something.   the api should be covered by the blanket wildcard of above filter

kalletlak: 👍

PR #4689, opened 2018-08-22
Add `rc` build status in the README

fedde-s:
The one icon by itself caused confusion as to which state of the code it was
referring to; it could be read as ‘the branch you're currently looking at’.
Additionally, this will make more people notice when the code in the mainline
feature development branch fails the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,10 @@ If you would like to know how to setup a private instance of the portal and/or g
 If you are interested in coordinating the development of new features, please contact cbioportal@cbio.mskcc.org.
 
 ## Test status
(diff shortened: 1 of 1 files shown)


inodb:
@fedde-s thanks! i do like the table style slightly better: https://github.com/cBioPortal/cbioportal-frontend#test-status--code-quality

You prefer the style in this PR?

fedde-s:
How about this compromise, @inodb? It still targets people unfamiliar with a project, but in table form to make it a bit easier to skim :)

inodb:
@fedde-s perfect! Thanks!

PR #4691, opened 2018-08-22
Fix validator crash when no symbol in CNA data

sandertan:
# What? Why?
Fix https://github.com/cBioPortal/datahub/issues/318

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -990,9 +990,10 @@ class GenewiseFileValidator(FeaturewiseFileValidator):
             # treat empty string as a missing value
             if hugo_symbol == '':
                 hugo_symbol = None
-            # In case of CNA data the Hugo Symbol should be split when gene symbol contains pipe
-            if (type(self) is CNAValidator or type(self) is CNAContinuousValuesValidator) and '|' in hugo_symbol:
-                hugo_symbol = hugo_symbol.split('|')[0]
+            if hugo_symbol is not None:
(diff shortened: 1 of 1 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -990,9 +990,10 @@ def parseFeatureColumns(self, nonsample_col_vals):
             # treat empty string as a missing value
             if hugo_symbol == '':
                 hugo_symbol = None
-            # In case of CNA data the Hugo Symbol should be split when gene symbol contains pipe
-            if (type(self) is CNAValidator or type(self) is CNAContinuousValuesValidator) and '|' in hugo_symbol:
-                hugo_symbol = hugo_symbol.split('|')[0]
+            if hugo_symbol is not None:
+                # In case of CNA data the Hugo Symbol should be split when gene symbol contains pipe
+                if (type(self) is CNAValidator or type(self) is CNAContinuousValuesValidator) and '|' in hugo_symbol:
```
fedde-s: You shouldn't have to modify a function that affects parsing of all possible gene-wise matrices in order to change something specific to CNA files. Would it beat all possible to move this logic to in a subclass?

fedde-s: Having the superclass know about its subclasses in addition to vice-versa makes the code very hard to read.

sandertan:
I'm incorporating this commit, including a solution to @fedde-s feedback, in https://github.com/cBioPortal/cbioportal/pull/4666

PR #4693, opened 2018-08-22
New ER diagram for version 2.7.0 of DB schema

pieterlukasse:
# What? Why?
An updated version of the ER diagram, according to DB schema version 2.7.0

This diagram is useful for anyone thinking about proposing DB changes or doing maintenance in the backend. It also shows possible issues or weak points in current model.


pieterlukasse:
It already identified 2 issues with current cgds.sql:
- `clinical_attribute_meta` is not referenced by `clinical_patient` nor by `clinical_sample` tables
- `sample_list` is not referenced by `sample_list_list` table

PR #4705, opened 2018-08-23
spring-security 5 upgrade

n1zea144:
This PR upgrades the following security related packages:

- spring security
- spring saml
- spring social

Most of the changes are a result of moving to/through spring security 4.  Most notably, CSRF prevention is turned on by default in spring security 4.  As a result all get requests no longer work.  Therefore this PR turns off CSRF support to limit the scope of this PR to mainly configuration file changes.  Also in spring security 4, the default logout url is /logout.  To minimize disruptions to the codebase, this default has been set back to j_spring_security_logout.  The other changes are the result of package/class movement between spring 3 and spring 4.  Finally, there are dependency changes (the removal of the guava package from core/pom.xml is because there is a dependency to guave defined in the parent pom), spring-social-google requires guice (note the repos change).

This PR has been tested against googleplus and saml (using NetIQ IDP) profiles. 

For MSK folks, there is also an accompanied change to the internal Hg server with similar changes to private versions (private-beta, mskcc) of applicationContext-security.xml (see spring-security-5-upgrade branch in the portal-configuration repos).   This PR was also tested against restful login backdoor used by external systems at MSK.

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/security/security-spring/pom.xml
+++ b/security/security-spring/pom.xml
@@ -10,8 +10,8 @@
     <artifactId>security-spring</artifactId>
 
     <properties>
-      <spring.social.version>1.1.0.RELEASE</spring.social.version>
-      <org.springframework.social.google-version>1.0.0.RELEASE</org.springframework.social.google-version>
+      <spring.social.version>1.1.6.RELEASE</spring.social.version>
+      <org.springframework.social.google-version>1.1.3</org.springframework.social.google-version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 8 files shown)


security/security-spring/src/main/resources/applicationContext-security.xml
```diff
@@ -57,6 +57,7 @@
     <!-- This must come before the default entry point which will capture everything not matched by pattern -->
     <http use-expressions="true" pattern="/api/**" entry-point-ref="restAuthenticationEntryPoint" create-session="never">
         <intercept-url pattern="/**" access="isAuthenticated()"/>
+        <csrf disabled="true"/>
```
sheridancbio: It was necessary to disable csrf requirements to be backwards compatible. We might want to think this through for the future ... and consider the potential risks of session hijacking.

n1zea144: Yes, I agree.  If we enable csrf, we will have to change all front-end endpoints to post and start passing a csrf token.  Since our default behavior to this point has been disabled csrf, I figured we could leave this for another effort.

security/security-spring/pom.xml
```diff
@@ -84,10 +84,16 @@
       </dependency>
       <!-- Spring Social Google-->
       <dependency>
-        <groupId>org.springframework.social</groupId>
+        <groupId>com.github.spring-social</groupId>
         <artifactId>spring-social-google</artifactId>
         <version>${org.springframework.social.google-version}</version>
       </dependency>
+      <!-- Google guice -->
```
pieterlukasse: @n1zea144 why is this one needed?

PR #4708, opened 2018-08-24
Update rc after v1.16.0

fedde-s:
Now that v1.16.0 has been released, I believe that the next release that the `rc` branch will be building towards is v1.17.0.

Resolve merge conflicts by changing the versions of the Java projects to
v1.17.0-SNAPSHOT, but leave the frontend version up for discussion in te pull
request as it didn't change between cBioPortal v1.15.1 and v1.16.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,11 @@ install:
 - |
   if [[ "${TEST}" == python-validator ]]
   then
-      sudo -H pip install -r requirements.txt
+      sudo apt-get install python3.4-venv
+      python3.4 -m venv ~/python-env-validator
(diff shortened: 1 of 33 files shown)


fedde-s:
@inodb, do you see problems with this way to resolve the conflict and change the release number in `rc`? And do we want the frontend project to match it, presumably by skipping from 1.16.0 to 1.17.0?

inodb:
@fedde-s yup this is great! We can update the frontend commit to latest rc in a subsequent pr. There are some failing tests atm on frontend's rc so we'll need to resolve those first

PR #4740, opened 2018-08-31
Spa

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


portal/src/main/webapp/WEB-INF/jsp/dashboard/dashboard.jsp
```diff
@@ -569,16 +570,9 @@
...
-<script src="js/src/load-frontend.js"></script>
 
-<script>
-        window.loadReactApp({ defaultRoute: 'blank' });
-</script>
 
-<div id="reactRoot" style="display:none"></div>
-    
     
+
```
alisman: @kalletlak @zhx828 getting rid of import of new app into old studyview jsp.  was causing some complications

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -647,7 +644,20 @@
         <url-pattern>/ln</url-pattern>
     </servlet-mapping>
 
-
+    <servlet>
+        <servlet-name>SinglePageApplication</servlet-name>
+        <jsp-file>/index.jsp</jsp-file>
+    </servlet>
+    <!--mapping-->
+    <servlet-mapping>
```
alisman: @inodb note that every time we add a new route to SPA frontend, we have to make sure that tompcat with resolve that route to the SPA page (index.jsp).   This is kind of cumbersome and can be dealt with more generally but would require some learning/tinkering.  The strategy I've seen is to get server to resolve 404s to the SAP.  Then frontend can either match the route or show a 404 if there is no match.  But we DON'T want to resolve api 404s to SPA.   Food for thought.

portal/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,76 @@
+<%
```
alisman: @inodb This is a simple jsp which serves as a replacement for the .ejs file.   It's possible that we can just get rid of this eventually and simply serve the output of the ejs file from the WAR file.    But there may be reasons we still want to allow server to write some stuff to the page.   For example, if we want to load configuration information here instead of forcing the app to fetch it.   We can easily allow the application to do both/either.

portal/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,76 @@
...
+            sessionServiceUrl:'http://www.cbioportal.org/rc/api-legacy/proxy/session/',
+            disabledTabs:[],
+            skinShowDataSetsTab:true,
+            skinShowWebAPITab:true,
+            skinShowRmatLABTab:true,
+            skinShowTutorialsTab:true,
+            skinShowAboutTab:true,
+            skinIsMarkdownDocumentation:true,
+            skinDocumentationBaseUrl:"https://raw.githubusercontent.com/cbioportal/cbioportal/release-1.15.0/docs",
+            skinFaqSourceURL:"https://docs.google.com/document/d/e/2PACX-1vSWTtIJZF2tuBimihr8ke-d00DpKh7fydFIQb5xYpE_bMYM9hZyY9OP1Vz1Ts0ow7ob-3h2S19cuB5O/pub?embedded=true",
```
alisman: this stuff is crap.  should't be hardcoded here.

portal/src/main/webapp/js/src/dashboard/hijackSubmission.js
```diff
@@ -0,0 +1,47 @@
+
+function handleLegacySubmission(url, query){
```
alisman: @zhx828 these are dirty tricks to hijack submission of queries to results page and patient page from legacy study view.   We we localStorage to avoid having to post giant queries to the SPA(which would require backend support).    Obviously all this goes away when we introduce new study view, although we may actually need the localStorage trick if we want app to open in new window.

portal/src/main/webapp/WEB-INF/web.xml
inodb: resolved ☝️

portal/src/main/webapp/config_service.jsp
```diff
@@ -55,16 +31,161 @@ studies += "}";
...
+
+    <%
+    
+        // Set API root variable for cbioportal-frontend repo
+        String url = request.getRequestURL().toString();
+        String baseURL = url.substring(0, url.length() - request.getRequestURI().length()) + request.getContextPath();
+        baseURL = baseURL.replace("https://", "").replace("http://", "");
+    
+        String[] propNameArray = {
+            "app.version",
```
inodb: we'll test these more thoroughly before release: https://github.com/cBioPortal/cbioportal/issues/4972

Issue #3955, opened 2018-03-09
Can we show only the top mutated genes in study view


jjgao:
The Mutated Genes table currently shows cBioPortal cancer genes with 1 or more mutations, as well as any gene with 2 or more mutations. This has caused some confusion since not all mutated genes can be found in the table.

The question is whether it's useful to show so many genes. Can we show only the top, say 100, mutated genes? If not, what's the use cases to show less frequently mutated genes? 

Similarly to CNA Genes table -- which is worse b/c we only show cancer genes.

![image](https://user-images.githubusercontent.com/840895/37219208-c5203f74-2390-11e8-8f30-7cf97dc73c62.png)

schultzn:
I really like the ability to search for specific genes in the table, even if they are low frequency... can we keep that somehow?

jjgao:
@schultzn maybe we can have the complete list in the Mutated Genes tab? Or do you think it is important to support interactive analysis between low frequently mutated genes with other charts?

schultzn:
I don’t really want to lose the functionality… I never go to the Mutated Genes tab.

Why are you trying to remove the low frequency genes? To save loading time or to make changes to the visualization?

Let’s discuss in person maybe?

jjgao:
@schultzn the big problem is we are not showing all mutated genes, i.e. we are not showing a gene if it's only mutated once and is not a cancer gene. In the case of CNA genes, we only show cancer genes. There have a few users asking about the "discrepancy" between the Mutated Genes table and the results page.

Maybe we should just try to show all mutated genes and CNA genes. @zhx828 @ersinciftci maybe we can try to have some backend support for pagination and filtering?

zhx828:
@jjgao the issue will be real-time filtering. It will have delay everytime user filters the table.

ersinciftci:
The new study view APIs returns all mutated and CNA genes, so I'm closing this issue for now. We can re-open if we don't like the performance.

schultzn:
It would still be nice to search for genes and retrieve the frequency of alterations.. maybe we can show only the top 10 at first, but then quickly compute the other frequencies when a user searches for genes?

jjgao:
I like @schultzn's idea.

@ersinciftci @zhx828 How difficult is it to do pagination and query? For query, you may want to match genes that start with the user input once they have typed two characters.

ersinciftci:
[...]

PR #4761, opened 2018-09-10
Fix performance issue for cna-genes and mutated-genes endpoints

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/3955


ersinciftci:
Seems like this will make big studies slower than it is. I'll try to find a different way to solve this.

ersinciftci:
The new indexes added in https://github.com/cBioPortal/cbioportal/pull/4773 made the big studies slower as well, so maybe this change will cancel out with that one and fix the big studies performance problem.

ersinciftci:
Didn't work.

PR #4762, opened 2018-09-10
Change classes to be serializable for Redis compatibility

averyniceday:
# What? Why?
Recent changes to the backend to use Redis to store sessions created bugs in the portal. Information stored in the session were not serializable and resulted in an exception being thrown when downloading data.

i.e Message java.io.IOException: java.lang.RuntimeException: Class org.mskcc.cbio.portal.model.DownloadLink does not implement Serializable or externalizable (also with GeneticProfile)

- Change classes to implement Serializable

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/model/GeneticProfile.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/model/GeneticProfile.java
@@ -32,6 +32,7 @@
 
 package org.mskcc.cbio.portal.model;
 
+import java.io.Serializable;
 import java.util.Properties;
 
 import org.apache.commons.lang.builder.ToStringBuilder;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/portal/model/GeneticProfile.java
```diff
@@ -32,14 +32,15 @@
...
+import java.io.Serializable;
 import java.util.Properties;
 
 import org.apache.commons.lang.builder.ToStringBuilder;
 
 /**
  * Class for genetic profile
  */
-public class GeneticProfile {
+public class GeneticProfile implements Serializable {
```
sheridancbio: A member of this class is of type GeneticAlterationType ... but that is an enum and inherently serializable (so no declaration needed)

Issue #3180, opened 2017-09-25
add more index to tables


jjgao:
e.g. 

- `STABLE_ID` in `patient`
- `STABLE_ID` in `sample`
- `ATTR_VALUE` in `clinical_patient`
- `ATTR_VALUE` in `clinical_sample`
- `GENE_A` and `GENE_B` in `interaction`

jjgao:
there may be more, but don't worry about tables listed here: https://github.com/cBioPortal/cbioportal/issues/2054

ersinciftci:
Added indexes for 
`STABLE_ID` in `patient`
`STABLE_ID` in `sample`
will add for the others after evaluating the usage patterns.

PR #4773, opened 2018-09-11
Add indexes for patient and sample stable_id columns

ersinciftci:
Partially fixes https://github.com/cBioPortal/cbioportal/issues/3180


inodb:
Actually - seems like this has more to do with db config: https://dba.stackexchange.com/questions/10732/adding-an-index-very-slow-is-there-a-mysql-cmd-to-get-an-eta-or-show-progress. So not much we can do really

ersinciftci:
Didn't work

PR #4779, opened 2018-09-11
Replacing known problematic unicode characters with non unicode equivalents

onursumer:
# What? Why?
This is a workaround to solve the issues related to unicode characters when converting from SVG to PDF.

# Checks
- [ ] Runs on Heroku.


onursumer:
Seems like this changes the default font of the converted PDF.

onursumer:
Closing this for now. Replacing the characters on the front-end instead.

PR #4781, opened 2018-09-12
Warning instead of error for some incorrectly placed clinical attributes

sandertan:
Recently we changed the validation of whether clinical attributes should be in the sample or patient file (https://github.com/cBioPortal/cbioportal/pull/3910). This PR makes this validation less strict. 

This is required to load old study data (all TCGA provisional studies) that can't be migrated to the new format.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1883,7 +1883,8 @@ class ClinicalValidator(Validator):
                       'datatype',
                       'priority')
 
-    # Only a core set of attributes must be either specific in the patient or sample clinical data.
+    # A core set of attributes must be either specific in the patient or sample clinical data.
+    # See GET /api/ at http://oncotree.mskcc.org/cdd/swagger-ui.html#/
     PREDEFINED_ATTRIBUTES = {
         'AGE': {
             'is_patient_attribute': '1',
@@ -1984,6 +1985,12 @@ class ClinicalValidator(Validator):
             'datatype': 'STRING'
         },
     }
+
+    # For some attributes, only a warning should be reported when they are in the wrong data file.
+    # This was added to successfully validate old study data.
+    WARNING_ATTRIBUTES = ['PRIMARY_SITE', 'METASTATIC_SITE', 'TUMOR_TISSUE_SITE']
+
+    # Invalid characters in attributes
     INVALID_ID_CHARACTERS = r'[^A-Za-z0-9._-]'
 
 
@@ -2135,12 +2142,22 @@ class ClinicalValidator(Validator):
                     if attr_property == 'is_patient_attribute':
                         expected_level = self.PREDEFINED_ATTRIBUTES[col_name][attr_property]
                         if self.PROP_IS_PATIENT_ATTRIBUTE != expected_level:
-                            self.logger.error(
-                                'Attribute must to be a %s-level attribute',
-                                {'0': 'sample', '1': 'patient'}[expected_level],
-                                extra={'line_number': self.line_number,
-                                       'column_number': col_index + 1,
-                                       'cause': col_name})
+                            if col_name in self.WARNING_ATTRIBUTES:
+                                self.logger.warning(
+                                    'Clinical Data Dictionary defines this as a %s-level attribute. '
+                                    'See GET /api/ at http://oncotree.mskcc.org/cdd/swagger-ui.html#/',
+                                    {'0': 'sample', '1': 'patient'}[expected_level],
+                                    extra={'line_number': self.line_number,
+                                           'column_number': col_index + 1,
+                                           'cause': col_name})
+                            else:
+                                self.logger.error(
+                                    'According to Clinical Data Dictionary, this attribute must to be a %s-level '
+                                    'attribute. See GET /api/ at http://oncotree.mskcc.org/cdd/swagger-ui.html#/',
+                                    {'0': 'sample', '1': 'patient'}[expected_level],
+                                    extra={'line_number': self.line_number,
+                                           'column_number': col_index + 1,
+                                           'cause': col_name})
                     # check pre-header datatype property:
                     if attr_property == 'datatype':
                         # check pre-header metadata if applicable -- if these were
--- a/core/src/test/scripts/unit_tests_validate_data.py
+++ b/core/src/test/scripts/unit_tests_validate_data.py
@@ -254,7 +254,7 @@ class ClinicalColumnDefsTestCase(PostClinicalDataFileTestCase):
         self.assertEqual(record.column_number, 2)
         self.assertIn(record.cause, 'STRING')
 
-        # Expect warning for sample attribute in patient clinical data
+        # Expect error for sample attribute in patient clinical data
         record = next(record_iterator)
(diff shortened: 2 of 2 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -2135,12 +2142,22 @@ def checkHeader(self, cols):
...
-                            self.logger.error(
-                                'Attribute must to be a %s-level attribute',
-                                {'0': 'sample', '1': 'patient'}[expected_level],
-                                extra={'line_number': self.line_number,
-                                       'column_number': col_index + 1,
-                                       'cause': col_name})
+                            if col_name in self.WARNING_ATTRIBUTES:
+                                self.logger.warning(
+                                    'Clinical Data Dictionary defines this as a %s-level attribute. '
+                                    'See GET /api/ at http://oncotree.mskcc.org/cdd/swagger-ui.html#/',
```
fedde-s: I'm not sure what that ‘Clinical Data Dictionary’ is for, so how should I respond when I see this warning? What breaks if I completely ignore it?

sandertan: We could add an explanation of this in File-Formats. If I understand correctly, the CDD is the format used by MSKCC to normalize clinical data, and should be followed for the study to be compared to other studies. Not following this format is okay for non-core attributes, but then the users should know that the clinical data cannot be compared.

fedde-s: Is it possible to make the warning message itself let the user know that, so that they can quickly decide whether they want to accept this potential problem or refer to external docs to fix it?

sandertan: @fedde-s Good point! I changed the warning and added the CDD to file formats.

fedde-s: That resolves the potential issue I saw 👍

sandertan:
I also added a commit to update/remove some broken URLs in the documentations and fix some typos.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1984,6 +1985,12 @@ class ClinicalValidator(Validator):
             'datatype': 'STRING'
         },
     }
+
+    # For some attributes, only a warning should be reported when they are in the wrong data file.
+    # This was added to successfully validate old study data.
+    WARNING_ATTRIBUTES = ['PRIMARY_SITE', 'METASTATIC_SITE', 'TUMOR_TISSUE_SITE']
```
pieterlukasse: please add a few words on how it was tested/verified that having these attributes (only) in the wrong file does not affect functionality. And does this mean these were previously set to `logger.error` without a good reason? Or is the original problem solved (e.g. in the frontend)?

docs/Docker-Prerequisites.md
```diff
@@ -12,9 +12,8 @@ First, make sure that you have the latest version of Docker installed on your ma
...
 You can download these files by using the links below:
 
-- **Schema 1.3.0**: [SQL file with create table statements for portal release 1.3.1](https://raw.githubusercontent.com/cBioPortal/cbioportal/v1.3.1/core/src/main/resources/db/cgds.sql) 
-- **Seed data, part1**: [cbioportal-seed SQL (.gz) file - part1 (no pdb_ tables)](https://github.com/cbioportal/datahub/raw/b69c86803c40d543080bf31a645721d06c82d08d/seedDB/seed-cbioportal_no-pdb_hg19.sql.gz)
-- **Seed data, part2 (optional)** [cbioportal-seed SQL (.gz) file - part2 (only pdb_ tables)](https://github.com/cbioportal/datahub/raw/b69c86803c40d543080bf31a645721d06c82d08d/seedDB/seed-cbioportal_only-pdb.sql.gz)
+- **Schema 2.6.0**: [SQL file with create table statements for cBioPortal 1.12.0 or higher](https://raw.githubusercontent.com/cBioPortal/cbioportal/v1.13.1/db-scripts/src/main/resources/cgds.sql)
+- **Seed data**: [cbioportal-seed SQL (.gz) file containing genes, gene aliases and cancer types](https://github.com/cBioPortal/datahub/raw/054429f83529a4e5229c1f7670d1d4c891cd8a90/seedDB/seed-cbioportal_hg19_v2.6.0.sql.gz)
```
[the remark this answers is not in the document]
sandertan: It saves the effort of making a separate PR. I put it in a separate commit.

docs/File-Formats.md
```diff
@@ -210,7 +210,7 @@ The file containing the sample attributes has two **required** columns:
 
 By adding `PATIENT_ID` here, cBioPortal will map the given sample to this patient. This enables one to associate multiple samples to one patient. For example, a single patient may have had multiple biopsies, each of which has been genomically profiled. See [this example for a patient with multiple samples](http://www.cbioportal.org/patient?studyId=lgg_ucsf_2014&caseId=P04).
 
-The following columns are required if you want the [pan-cancer summary statistics tab in a pan-cancer study](http://www.cbioportal.org/index.do?cancer_study_list=cellline_ccle_broad&cancer_study_id=cellline_ccle_broad&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=cellline_ccle_broad_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=cellline_ccle_broad_CNA&Z_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=cellline_ccle_broad_cnaseq&case_ids=&patient_case_select=sample&gene_set_choice=prostate-cancer%3A-ar-signaling-%2810-genes%29&gene_list=SOX9+RAN+TNK2+EP300+PXN+NCOA2+AR+NRIP1+NCOR1+NCOR2&clinical_param_selection=null&tab_index=tab_visualize&Action=Submit#pancancer_study_summary):
+The following columns are required for the pan-cancer summary statistics tab.
```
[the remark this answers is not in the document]
sandertan: Well it says tab. So it's a tab. But ok, I've added back a link that isn't broken.

core/src/main/scripts/importer/validateData.py
sandertan: See https://github.com/cBioPortal/cbioportal/blob/d82c5fc7b40dd1f4a5fcffa67976e150dad4f382/docs/File-Formats.md#custom-columns-in-clinical-data

These properties don't lead issues in the UI, but can make the data incomparable between studies. I tested this in the patient & sample view.

docs/Docker-Prerequisites.md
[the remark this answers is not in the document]
pieterlukasse: please update based on https://github.com/cBioPortal/datahub/pull/354. Please also update the title of this PR to reflect the fact that it is about 2 things.

core/src/main/scripts/importer/validateData.py
pieterlukasse: What happens when these attributes are in both patient and sample file? What if they have conflicting values? Is sample given priority? Is there a warning during validation? Will the UI be able to handle this?

What happens when the patient has multiple samples, but only patient has the attribute METASTATIC_SITE? Will all samples be labelled metastatic?

sandertan:
Replaced by #4836 and issue https://github.com/cBioPortal/datahub/issues/361.

PR #4787, opened 2018-09-12
Allow changing config at runtime via $PORTAL_HOME/portal.properties

fedde-s:
Allow changing the configuration from portal.properties, including the database credentials, by pointing `$PORTAL_HOME` at a folder containing a `portal.properties` config file when running from any entrypoint (including command-line tools/scripts). This should greatly simplify and clarify deployment.

This approach should retain compatibility with existing deployments, by using the existing `$PORTAL_HOME` variable. If no `$PORTAL_HOME/portal.properties` file exists, it still falls back to the compiled-in defaults from `<cbioportal_source_folder>/src/main/resources/` and `GlobalProperties.java`.

Caveat: the web server still requires database credentials to be duplicated in the Tomcat resource.

# Checks
- [x] Runs on Heroku.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -563,8 +563,13 @@ public class GlobalProperties {
     public static String getEmailContact()
     {
         String emailAddress = portalProperties.getProperty(SKIN_EMAIL_CONTACT);
-        return (emailAddress == null) ? DEFAULT_EMAIL_CONTACT :
-            ("<span class=\"mailme\" title=\"Contact us\">" + emailAddress + "</span>");
+        if (emailAddress == null)
+            emailAddress = DEFAULT_EMAIL_CONTACT;
+        return (
+                "<span class=\"mailme\" title=\"Contact us\">" +
+                emailAddress +
+                "</span>"
+        );
     }
 
     public static boolean includeNetworks()
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -32,12 +32,9 @@
 
 package org.mskcc.cbio.portal.util;
 
-import org.mskcc.cbio.portal.servlet.QueryBuilder;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.core.io.ClassPathResource;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Component;
@@ -308,8 +305,48 @@ public class GlobalProperties {
     public void setFrontendUrlRuntime(String property) { frontendUrlRuntime = property; }
 
     private static Log LOG = LogFactory.getLog(GlobalProperties.class);
-    private static Properties portalProperties = initializeProperties(PORTAL_PROPERTIES_FILE_NAME);
+    private static ConfigPropertyResolver portalProperties = new ConfigPropertyResolver();
     private static Properties mavenProperties = initializeProperties(MAVEN_PROPERTIES_FILE_NAME);
+
+    /**
+     * Minimal portal property resolver that takes system property overrides.
+     *
+     * Provides properties from runtime or the baked-in
+     * portal.properties config file, but takes overrides from <code>-D</code>
+     * system properties.
+     */
+    private static class ConfigPropertyResolver {
+        private Properties configFileProperties;
+        /**
+         * Finds the config file for properties not overridden by system props.
+         *
+         * Either the runtime or buildtime portal.properties file.
+         */
+        public ConfigPropertyResolver() {
+            configFileProperties = initializeProperties(PORTAL_PROPERTIES_FILE_NAME);
+        }
+        /**
+         * Finds the property with the specified key, or returns the default.
+         */
+        public String getProperty(String key, String defaultValue) {
+            String propertyValue = configFileProperties.getProperty(key, defaultValue);
+            return System.getProperty(key, propertyValue);
+        }
+        /**
+         * Finds the property with the specified key, or returns null.
+         */
+        public String getProperty(String key) {
+            return getProperty(key, null);
+        }
+        /**
+        * Tests if a property has been specified for this key.
+        *
+        * @return true iff the property was specified, even if blank.
+        */
+        public boolean containsKey(String key) {
+            return getProperty(key) != null;
+        }
+    }
 
     private static Properties initializeProperties(String propertiesFileName)
     {
--- a/docs/Deploying.md
+++ b/docs/Deploying.md
@@ -12,11 +12,28 @@ To make it available to your bash shell, add the following to your `.bash_profil
 
 Note:  If you are following the [recommended Ubuntu instructions](https://www.digitalocean.com/community/tutorials/how-to-install-apache-tomcat-8-on-ubuntu-14-04):  you should set ```export CATALINA_HOME=/opt/tomcat```
 
+## Prepare the global configuration file
+
+The portal is configured using a global configuration file, `portal.properties`.
+An example file is available in the `src/main/resources` folder.
+Use it as a template to create your own:
+
+    cd src/main/resources
+    cp portal.properties.EXAMPLE $HOME/cbioportal/portal.properties
+
+For more information about the `portal.properties` file,
+see the [reference](portal.properties-Reference.md) page.
+
 ## Add PORTAL_HOME to Tomcat
 
-The `PORTAL_HOME` environment variable needs to be available to the `cbioportal.war` file which runs within the Tomcat server. To make it available to Tomcat, edit your Tomcat startup file (typically `$CATALINA_HOME/bin/catalina.sh`) and add the following line anywhere within this file (we typically add it near the `JAVA_OPTS` statements):
+The `PORTAL_HOME` environment variable needs to be available to
+the `cbioportal.war` file which runs within the Tomcat server,
+so that it can find the configuration file. To make it available to Tomcat,
+edit your Tomcat startup file (typically `$CATALINA_HOME/bin/setenv.sh`)
+and add a line like the following, pointing it to the folder containing
+`portal.properties`:
 
-    export PORTAL_HOME= $CATALINA_HOME + "/webapps/cbioportal/WEB-INF/classes/"
+    export PORTAL_HOME=/Users/johndoe/cbioportal
 
 ## Add the MySQL JDBC Driver to Apache Tomcat
 
@@ -70,7 +87,7 @@ Lastly, open a browser and go to:
 ## Important
 
 - Each time you modify any java code, you must recompile and redeploy the WAR file.
-- Each time you modify any properties (see customization options), you must recompile and redeploy the WAR file.
+- Each time you modify any properties (see customization options), you must restart tomcat.
 - Each time you add new data, you must restart tomcat.
 
 ## Developer Tip
--- a/docs/Deploying.md
+++ b/docs/Deploying.md
@@ -54,8 +54,9 @@ or, if you are following the [recommended Ubuntu instructions](https://www.digit
 
 After the tomcat server has been started, to deploy the WAR file, run the following command:
 
-    sudo cp portal/target/cbioportal-*-SNAPSHOT.war $CATALINA_HOME/webapps/cbioportal.war
-    
+```
+sudo cp portal/target/cbioportal-*-SNAPSHOT.war $CATALINA_HOME/webapps/cbioportal.war
+```
 
 After doing this, you can look in the tomcat log file (`$CATALINA_HOME/logs/catalina.out`) to see if the portal has been proper deployed.  You should see something like:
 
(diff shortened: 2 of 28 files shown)


fedde-s:
I just heard that @alisman is also planning to look into how configuration is provided to the portal. Maybe he should read through the steps of the deployment manual as I document them here—again, I believe it to be fully compatible with the traditional way to configure the portal.

alisman:
@fedde-s looking through this, I think our efforts are not really related.  I am creating a service which will expose some configuration properties to the frontend.   There are also a couple ways to override this configuration at runtime for development purposes.  At some point we might want to split the frontend related configuration from the backend/api configuration.  Not sure yet.

fedde-s:
I'm glad we didn't accidentally do duplicate, or worse, conflicting work then, @alisman. I was worried for a moment. 😅

portal/pom.xml
```diff
@@ -131,7 +131,8 @@
             <resource>
               <directory>${project.parent.basedir}/business/src/main/resources</directory>
               <targetPath>WEB-INF/classes</targetPath>
-              <filtering>true</filtering>
+              <!-- do not hard-set portal.properties config at build-time -->
+              <filtering>false</filtering>
```
inodb: this might be a thing for pipelines to check if this is fine

test/end-to-end/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ cbioportal:
     - $PWD/../../:/cbioportal
   command: java -Ddbconnector=dbcp -jar portal/target/dependency/webapp-runner.jar --expand-war portal/target/cbioportal.war
   environment:
-    PORTAL_HOME: /cbioportal
+    PORTAL_HOME: /cbioportal/test/end-to-end/runtime-config
```
inodb: nice

fedde-s:
`GlobalProperties.java` supported overriding the properties file by placing one in `$PORTAL_HOME/` all along. The main pain point was the several distinct sources of database config, which is what I sorted out here. The rest is changes to the script and docs so that they no longer assume other scripts and Maven build products to be replicated in every configuration folder.

fedde-s:
And of course I demonstrated  the new flexibility in the ‘deployment steps’ part of the manual!

fedde-s:
> We can maybe keep them as command line parameters there [in heroku deployment], then we can make sure that both command line arguments and using $PORTAL_HOME works.

I don't think we've ever had a supported use-case for using command-line parameters to bake different defaults into the `.war` during compilation without touching the `src/` folders. That was a quirk of how a few particular properties were implemented, and it was already broken for all others.

It might be a good idea to use `src/main/resources/portal.properties` to make it test whether baking custom defaults into the `.war` that way still works, as that's what has been documented so far.

fedde-s:
I'm not familiar enough with Heroku to get it to work there, and it doesn't seem to allow me to see the logs. I've pushed a commit that should be a step in the right direction, removing the use of now-obsolete implementation details in the build process, but the app still doesn't run. @inodb, would it be easier for you to see what's going wrong?

inodb:
@fedde-s Thanks for having a look! When debugging it might be easier to deploy to your own account. Go here: https://github.com/thehyve/cbioportal/tree/jdbcutil-datasource n click on the button "deploy to heroku" in the README (might have to make a heroku account). Then you can see build logs and app logs. Looks like something still goes wrong with the database connection. My guess is that the `heroku/` folder isn't being included in the war. Heroku's java buildpack only deploys the war, not the entire source folder afaik.

If possible I think supporting spring boot style configuration (https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html) of the db parameters would be nicer. Then we could switch heroku to use `SPRING_APPLICATION_JSON` in `app.json`. Or do you think this would take too much effort?

fedde-s:
I feel like adding another configuration API to cBioPortal would make things more confusing rather than less, if it's only documented in pull request discussions. And documenting it sound like a lot of work if it works for some properties but not others, uses different file formats and environment variables in addition to the pre-existing ones, and requires deployers to learn what Spring is. That would be a feature to review in a separate pull request.

fedde-s:
It might be feasible though, to have `GlobalProperties.java` take it into account when initialising its internal `portalProperties` field, so that it works for all properties. If you know how to make that pull request, I could use it in this one after that's merged.

fedde-s:
Do you think you could implement that fairly easily, @inodb? You probably know Spring better than I do, as well as the use-case you had in mind. It sound like it would be way easier than integrating the defaults via https://github.com/inodb/cbioportal-buildpack or something.

inodb:
@fedde-s Agree that it would be good to only document that type of configuration once it's actually working for everything 

For heroku the only properties that we use there for which spring boot config is not working are the db parameters. I'll have a look if it's easy to support that.

To apply for all configuration, I think this PR solves most of it https://github.com/bihealth/cbioportal/compare/1352749b72a03db5a9091f79ceead4e268503b4c...bihealth:feature/spring-style-properties. But for now Id be happy if we can just get the db settings through spring.

- [ ] There's one more thing that I thought of. There might actually be portal maintainers that are already setting PORTAL_HOME n those will suddenly see that when changing that file they might be updating their production config. Maybe it's good to add a mvn profile for the runtime configuration using `$PORTAL_HOME/src/main/resources`. So one can only change the runtime config through opt-in. What do you think?

fedde-s:
> There might actually be portal maintainers that are already setting PORTAL_HOME n those will suddenly see that when changing that file they might be updating their production config.

This was already the case, but it led to undefined behaviour when there was a `portal.properties` file in the root `$PORTAL_HOME/` folder as it didn't work for all properties or entrypoints. I don't think this PR will introduce new issues there.

fedde-s:
@inodb, do you know how to get Spring to inject these properties into the `portalProperties` internal of `GlobalProperties.java`? That's where they're read from: https://github.com/cBioPortal/cbioportal/blob/v1.16.0/core/src/main/java/org/mskcc/cbio/portal/util/DatabaseProperties.java#L55

inodb:
@fedde-s I think this should work: https://github.com/bihealth/cbioportal/compare/1352749b72a03db5a9091f79ceead4e268503b4c...bihealth:feature/spring-style-properties#diff-d5234aefb7508dcc8a8f17cb9684d495R262

inodb:
@fedde-s hmm link doesn't work, but look at `GlobalProperties.java`, `resolveProperty` and `EmbeddedValueResolver`

inodb:
@fedde-s this fixes it https://github.com/inodb/cbioportal/tree/jdbcutil-datasource. Got it running here: https://cbioportal-db-runtime.herokuapp.com/. It gives a few memory quota exceeded errors but that's expected, just refresh a few times n it'll show. You can merge in that one commit to this branch maybe? I haven't included all the properties as @holtgrewe did in his branch (thanks for all ur work! Sorry it's been taking so long to get this in). I just copied what we wanted to use. We could add it or do that at a later stage. Seems to work for db configuration on heroku at least, so that's good.

fedde-s:
> Agree that it would be good to only document that type of configuration once it's actually working for everything 

More importantly, I think we shouldn't use them on Heroku until the use case is documented for others in the community to understand. Doing that is what led this pull request to be delayed in the first place.

inodb:
@fedde-s I don't agree. This is a transition phase. Eventually all parameters should be configurable in a similar manner. The Heroku test did a good job catching that this style of configuration no longer worked for the parameters for which it used to work. A lot of our features start out as internal use only followed by generalizing it for other cBioPortal deployers. That's why many things aren't documented that are only in the "internal phase". We could add a documentation section for developmental use. That might be a good idea for some functionality, but writing this is a matter of prioritization and if it's a choice between documentation and tests I'd much rather have a test then vice versa. 

I think the main reason why this PR was delayed is because we weren't on the same page of what the eventual goal was. We should have prolly hashed that out on a call/issue/slack or RFC before the PR was submitted. I also don't think a delay of 6 days so far is really that bad, given that we've had several proposals of changing this configuration for well over a year that never made it through.

I'm very happy and appreciative though that you are pushing this forward. Thanks a ton!

Given you're comment I assumed you were not too keen on including my/holtgrewe's changes that i pointed to ☝️, but hopefully I've been able to shed some light on the reasons for including this? Thanks!

fedde-s:
> hopefully I've been able to shed some light on the reasons for including this?

You have, thanks! But if we are adding this new complexity into the shared code, it's no longer in the ‘internal phase’ and I'd prefer others who want to build on the code to understand these reasons as well. Even if it's only in comments inside the code.

fedde-s:
> we weren't on the same page of what the eventual goal was

That's exactly why I made the pull request, and wrote a comment at the top of it :)

fedde-s:
> we weren't on the same page of what the eventual goal was

That's exactly why I made a pull request, and wrote a comment at the top of it :)

inodb:
@fedde-s so how about this, we merge in the changes from my branch and follow up the documentation issues n other runtime configuration properties in a new PR addressing https://github.com/cBioPortal/cbioportal/issues/4832.

fedde-s:
Unfortunately, it looks like that change breaks the credentials in both the runtime and buildtime `portal.properties` altogether, leaving *only* the new `-D` options (and the Tomcat resource, in the case of the webserver) working. I'm not sure what to do about that. Do you see a way around it, @inodb?

fedde-s:
Presumably, the resolver should use configuration similar to this block in `applicationContext-business.xml`:

```xml
        <bean id="propertyPlaceholderConfigurer" class="org.springframework.beans.factory.config.PropertyPlaceholderConfigurer">
            <property name="systemPropertiesModeName" value="SYSTEM_PROPERTIES_MODE_OVERRIDE" />
            <property name="searchSystemEnvironment" value="true" />
            <property name="ignoreResourceNotFound" value="true" />
            <property name="locations">
                <list>
                    <value>classpath:/portal.properties</value>
                    <value>file:///#{systemEnvironment['PORTAL_HOME']}/portal.properties</value>
                </list>
            </property>
        </bean>
```

fedde-s:
What's going wrong, is that the code tries to call `EmbeddedValueResolverProvider.getResolver().resolveStringValue()`, but that first method returns `null` in scripts that don't boot up the Spring framework. And that's most of them.

fedde-s:
If the Spring framework should be initialised in order to access the database credentials, do you  know how the `GlobalProperties.resolveProperty` method itself can do that?

fedde-s:
Applied a quick fix that should get Heroku to run again, as discussed with @inodb elsewhere. It doesn't introduce Spring to entrypoints that didn't previously use them just yet, but does respect Java system properties passed with `-D` as overrides for system properties.

fedde-s:
Any objections from the reviewers to merging this now?

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -308,9 +305,49 @@ public static String parseUrl(String url)
...
+         */
+        public ConfigPropertyResolver() {
+            configFileProperties = initializeProperties(PORTAL_PROPERTIES_FILE_NAME);
+        }
+        /**
+         * Finds the property with the specified key, or returns the default.
+         */
+        public String getProperty(String key, String defaultValue) {
+            String propertyValue = configFileProperties.getProperty(key, defaultValue);
+            return System.getProperty(key, propertyValue);
```
inodb: i like it

inodb:
Thanks so much @fedde-s v clean solution!

fedde-s:
Thanks!

docs/Deploying.md
```diff
@@ -12,11 +12,28 @@ To make it available to your bash shell, add the following to your `.bash_profil
...
-The `PORTAL_HOME` environment variable needs to be available to the `cbioportal.war` file which runs within the Tomcat server. To make it available to Tomcat, edit your Tomcat startup file (typically `$CATALINA_HOME/bin/catalina.sh`) and add the following line anywhere within this file (we typically add it near the `JAVA_OPTS` statements):
+The `PORTAL_HOME` environment variable needs to be available to
+the `cbioportal.war` file which runs within the Tomcat server,
+so that it can find the configuration file. To make it available to Tomcat,
+edit your Tomcat startup file (typically `$CATALINA_HOME/bin/setenv.sh`)
+and add a line like the following, pointing it to the folder containing
+`portal.properties`:
 
-    export PORTAL_HOME= $CATALINA_HOME + "/webapps/cbioportal/WEB-INF/classes/"
+    export PORTAL_HOME=/Users/johndoe/cbioportal
```
sheridancbio: right ... the configuration can be read from anywhere now. it does not need to come from the baked in copy.

docs/Deploying.md
```diff
@@ -64,13 +82,12 @@ After doing this, you can look in the tomcat log file (`$CATALINA_HOME/logs/cata
...
 Lastly, open a browser and go to:  
-
-[http://localhost:8080/cbioportal/](http://localhost:8080/cbioportal/)
+<http://localhost:8080/cbioportal/>
 
 ## Important
 
 - Each time you modify any java code, you must recompile and redeploy the WAR file.
-- Each time you modify any properties (see customization options), you must recompile and redeploy the WAR file.
+- Each time you modify any properties (see customization options), you must restart tomcat.
```
sheridancbio: 👍

portal/pom.xml
sheridancbio: I guess we will check this during the next importer build. We definitely use filtering, but I would guess that the filtering we need is done inside the (spring-based) importer codebase and that the importer pom.xml is the controlling file.

pieterlukasse:
@inodb thanks for the review and merge. Discussing the release here: https://github.com/thehyve/cbioportal-docker/pull/40#discussion_r220882196

sheridancbio:
One thing which I did not realize previously is that we have a very overloaded and multi-modal GlobalProperties.java class now. This class probably needs to be cleaned up and broken up now ... and we will have to do so when removing the core module.

fedde-s:
This pull request has been merged and in active use on many production servers for months; if you see an issue with the current state of the cBioPortal software, I'd suggest logging it.

PR #4789, opened 2018-09-12
Authenticates any request with an authorization bearer header 

mandawilson:
Currently authenticates user to be "fakeuser@mskcc.org", which obviously needs changing.  Only tested with "googleplus" security profile.

# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/security/security-spring/src/main/resources/applicationContext-security.xml
+++ b/security/security-spring/src/main/resources/applicationContext-security.xml
@@ -53,14 +53,38 @@
     <http pattern="/reactapp/**" security="none"/>
     <http pattern="/api/swagger-resources/**" security="none"/>
     <http pattern="/api/swagger-ui.html" security="none"/>
-        
+
     <!-- This must come before the default entry point which will capture everything not matched by pattern -->
     <http use-expressions="true" pattern="/api/**" entry-point-ref="restAuthenticationEntryPoint" create-session="never">
         <intercept-url pattern="/**" access="isAuthenticated()"/>
         <csrf disabled="true"/>
+        <custom-filter ref="tokenAuthenticationFilter" before="FORM_LOGIN_FILTER"/>
     </http>
 
     <b:bean id="restAuthenticationEntryPoint" class="org.cbioportal.security.spring.sessionservice.RestAuthenticationEntryPoint"/>
+
+    <b:bean id="tokenAuthenticationFilter" class="org.cbioportal.security.spring.authentication.token.TokenAuthenticationFilter">
+        <b:property name="authenticationManager" ref="tokenAuthenticationManager" />
+        <b:property name="authenticationSuccessHandler" ref="tokenSuccessRedirectHandler"/>
+    </b:bean>
+
+    <!-- WARNING: if you define another authentication-manager AFTER this one, this one
+        will be overridden, even if you use <http ... pattern="/api/**" ... authentication-manager-ref="tokenAuthenticationManager">
+        AND even though the authenticationManager is defined as a property of tokenAuthenticationFilter
+        so you must add <authentication-provider ref="tokenUserDetailsAuthenticationProvider"/> to your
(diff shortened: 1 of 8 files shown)


security/security-spring/src/main/resources/applicationContext-security.xml
```diff
@@ -533,6 +567,7 @@
     <!-- authenticate is off but we have session-service beans -->
     <b:beans profile="noauthsessionservice">
         <global-method-security pre-post-annotations="disabled"/>
+        <b:bean id="restAuthenticationEntryPoint" class="org.cbioportal.security.spring.sessionservice.RestAuthenticationEntryPoint"/>
```
mandawilson: For some reason if "noauthsessionservice" was selected, it would throw an error that it couldn't find the "restAuthenticationEntryPoint" bean, even though it is globally defined.  I don't think any portals are using the "noauthsessionservice" option though, I think session service is working with authentication and authorization = false, so perhaps we don't need this profile anymore.  Anyway maybe this PR is not the place to add this, because it is not related to the token issue.

security/security-spring/src/main/resources/applicationContext-security.xml
```diff
@@ -53,14 +53,38 @@
     <http pattern="/reactapp/**" security="none"/>
     <http pattern="/api/swagger-resources/**" security="none"/>
     <http pattern="/api/swagger-ui.html" security="none"/>
-        
+
     <!-- This must come before the default entry point which will capture everything not matched by pattern -->
-    <http use-expressions="true" pattern="/api/**" entry-point-ref="restAuthenticationEntryPoint" create-session="never">
+    <http use-expressions="true" pattern="/api/**" entry-point-ref="restAuthenticationEntryPoint" create-session="never" authentication-manager-ref="tokenAuthenticationManager">
```
n1zea144: @mandawilson Doesn't this mean that when this code gets merged, web page requests will need tokens?  Ah wait no, in the filter, you only requiresAuthentication if an Authorization header is present otherwise control moves to the next filter in the framework......nice.

n1zea144: Why do you need authentication-manager-ref defined here and within the TokenAuthenticationFilter?

n1zea144: It doesn't look like you based these changes on the spring security 5 code - was this intentional?

security/security-spring/src/main/resources/applicationContext-security.xml
n1zea144: No, its defined only when authentication is turned on:

 <b:beans profile="googleplus,openid,saml,ad,ldap">
...
</b:beans>

security/security-spring/src/main/resources/applicationContext-security.xml
mandawilson: @n1zea144 Yes it isn't based on Spring Security 5 code because those changes hadn't been merged in when I started this.  I will get them now.  I am pretty confident it won't change anything because most the time I was reading security documentation for more recent Spring Security versions and then checking to make sure they hadn't changed since version 3.

I can remove authentication-manager-ref, I left it in after testing it and realizing it wasn't helpful (it didn't stop the second manager from overriding this referenced one).

security/security-spring/src/main/resources/applicationContext-security.xml
sheridancbio: I was about to merge this .. but I will wait for the update to spring-security v5

security/security-spring/src/main/resources/applicationContext-security.xml
sheridancbio: I was about to merge this .. but I will wait for the update to spring-security v5

sheridancbio:
code changes have been pushed directly to the feature branch in the cBioPortal repo ... Do Not Merge

PR #4801, opened 2018-09-14
Add names to all database indexes to prevent index override

ersinciftci:


ersinciftci:
Seems like the index name didn't actually have an effect.

Issue #4812, opened 2018-09-17
MUTATION_COUNT issues with %_uncalled profile


inodb:
- [x] Importer should not add mutation count for uncalled profile
- [x] Change datatype from MUTATION_EXTENDED to MUTATION_UNCALLED in portal importer configuration file for `mutation-uncalled`

PR #4841, opened 2018-09-19
For MUTATION_COUNT  include only MUTATION_EXTENDED

inodb:
At MSKCC there are some MUTATION_EXTENDED_UNCALLED and FUSION profiles that
shouldn't be included when determining the number of mutations for a sample

Partially solves https://github.com/cBioPortal/cbioportal/issues/4812. Also need to update the pipeline to actually give uncalled profiles the correct genetic_alteration_type

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -145,6 +145,7 @@ public final class DaoMutation {
                     "AND mutation.`MUTATION_EVENT_ID` = mutation_event.`MUTATION_EVENT_ID` " +
                     "AND mutation.`MUTATION_STATUS` <> 'GERMLINE' " +
                     "AND genetic_profile.`GENETIC_PROFILE_ID`=? " +
+                    "AND genetic_profile.`GENETIC_ALTERATION_TYPE` = 'MUTATION_EXTENDED' " +
                     "GROUP BY genetic_profile.`GENETIC_PROFILE_ID` , `SAMPLE_ID`;");
             pstmt.setInt(1, geneticProfile.getGeneticProfileId());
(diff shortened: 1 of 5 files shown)


jjgao:
@inodb we should also not include `fusion` events that are in MUTATION_EXTENDED profiles.

inodb:
@jjgao fusions have the genetic alteration type `FUSION` so that should be fine

jjgao:
@inodb AFAIK, we are still importing fusions into the mutation profile. Please double check.

jjgao:
Actually, you can learn it just from querying mutation profile for fusions, e.g. in this query (http://www.cbioportal.org/index.do?session_id=5ba30993498eb8b3d567ef55), only the mutation profile was selected but ERG fusions are queried out.

It was a hack that lived too long, unfortunately.

inodb:
@jj good catch! Missed that - ill check if there's some field in the mutation table that we can use to ignore them. Will update the PR

inodb:
@jjgao 	updated the PR. For now I left the mutation_count_by_keyword table alone. Like you said it shouldn't really matter since it's computed on a per profile basis and only used on the patient view. I filed an issue to make sure the frontend handles it properly: https://github.com/cBioPortal/cbioportal/issues/4848

For fusions in MUTATION_EXTTENDED this the mysql query to get them:
```
select distinct genetic_profile.stable_id from mutation, mutation_event, genetic_profile where genetic_profile.genetic_alteration_type != 'FUSION' and mutation.genetic_profile_id=genetic_profile.genetic_profile_id and mutation.mutation_event_id=mutation_event.mutation_event_id and  mutation_type <> 'Fusion' and protein_change;
```
We could clean that up at some point but necessary for 1.17 id say

inodb:
I tested the migration, should take ~5-10m on production. I did not test importing, but the core tests importing the test study did a good job at least catching the errors i made

PR #4869, opened 2018-09-24
MUTATION_UNCALLED valid /mutations/fetch profile

inodb:
This fixed the problem that the migration caused for msk internal %_uncalled
profiles to now be of genetic alteration type MUTATION_UNCALLED. It was no longer possible to retrieve those mutations on the patient view page.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
@@ -216,6 +216,7 @@ public class MolecularDataServiceImpl implements MolecularDataService {
         MolecularProfile molecularProfile = molecularProfileService.getMolecularProfile(molecularProfileId);
 
         if (molecularProfile.getMolecularAlterationType().equals(MolecularAlterationType.MUTATION_EXTENDED) || 
+            molecularProfile.getMolecularAlterationType().equals(MolecularAlterationType.MUTATION_UNCALLED) ||
             molecularProfile.getMolecularAlterationType().equals(MolecularAlterationType.FUSION)) {
 
             throw new MolecularProfileNotFoundException(molecularProfileId);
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/cbioportal/service/impl/VariantCountServiceImpl.java
```diff
@@ -58,8 +58,9 @@ private MolecularProfile validateMolecularProfile(String molecularProfileId) thr
 
         MolecularProfile molecularProfile = molecularProfileService.getMolecularProfile(molecularProfileId);
 
-        if (!molecularProfile.getMolecularAlterationType()
-            .equals(MolecularProfile.MolecularAlterationType.MUTATION_EXTENDED)) {
+        if (!(molecularProfile.getMolecularAlterationType()
+            .equals(MolecularProfile.MolecularAlterationType.MUTATION_EXTENDED) || !molecularProfile.getMolecularAlterationType()
```
inodb: @ersinciftci im not really sure where the variant count api is used but figured i added this anyway

service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
```diff
@@ -216,6 +216,7 @@ private void validateMolecularProfile(String molecularProfileId) throws Molecula
         MolecularProfile molecularProfile = molecularProfileService.getMolecularProfile(molecularProfileId);
 
         if (molecularProfile.getMolecularAlterationType().equals(MolecularAlterationType.MUTATION_EXTENDED) || 
+            molecularProfile.getMolecularAlterationType().equals(MolecularAlterationType.MUTATION_UNCALLED) ||
```
inodb: MUTATION_UNCALLED shouldn't be used for `molecular-data`, only used for `mutations`

inodb:
tested on cbioportal.mskcc.org/beta

service/src/main/java/org/cbioportal/service/impl/VariantCountServiceImpl.java
inodb: actually there's a double negation here - will update in follow up PR

Issue #4895, opened 2018-09-28
Handle missing MUTATION_COUNT data


inodb:
On OncoPrint the MUTATION_COUNT now shows up as having 92% data available, even though it is 100%, we are just treating missing data as 0. Possible solutions

- Fill in missing data on frontend as 0
- Add 0 values in the clinical database
- Fill in 0 values in backend at API level

I guess advantage of adding it in db is that we won't have to deal with it in multiple places. Downside is that it increase table size n response size

As reported by @schultzn 

CC @adamabeshouse 	@ersinciftci

adamabeshouse:
@inodb filling in missing data would also have to take into account if the sample is sequenced at all, right? Personally I think it would be great not to have to code this as a special case, and fill it in at the clinical database level. But I dont know whats involved with that.

inodb:
@adamabeshouse Correct - Im not sure how this was originally handled

jjgao:
@inodb @adamabeshouse it was missing from the database previously too and was handled in the front end. https://github.com/cBioPortal/cbioportal/issues/4799

@inodb @ersinciftci: we should add the sequenced sample without mutaitons and assign 0 as their mutation_count. Could you prioritize that?

yichaoS:
@schultzn ^

sheridancbio:
PR #4907 has been merged and deployed. This corrected almost all cases, but after deployment it was noticed that the coverage for the mskimpact study went from 92% to 99%  (not 100%). The issue was looked into, and we think we understand what happened. If a sample is in the profile and yet has no mutation records in the MAF, we use a LEFT JOIN operation to construct a single record which does not populate any fields from the mutation / mutation_event tables. This is handled properly, and a single MUTATION_COUNT record with value 1 is inserted into clinical_sample.

However, if there *are* mutation records from the MAF, and all mutation records are filtered out of the count by the filtering of GERMLINE and Fusion records, then no records at all are present in the result, and no MUTATION_COUNT 0 record is set for that sample.

In the mskimpact study, there seem to be 451 samples that fall into this category .. where mutations are comprised only of GERMLINE or Fusion records. Here are some example sample ids:

inodb:
There's also an internal issue where MUTATION_COUNT fails with a key constraint when you import mutation data split over two files, I've created a separate issue here: https://github.com/cBioPortal/cbioportal/issues/4950

PR #4907, opened 2018-10-01
Add MUTATION_COUNT = 0 for samples that is profiled but has no mutation

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4895


core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -140,16 +140,17 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
...
-                    "SELECT `SAMPLE_ID`, COUNT(DISTINCT mutation_event.`CHR`, mutation_event.`START_POSITION`, " +
+                    "SELECT sample_profile.`SAMPLE_ID`, COUNT(DISTINCT mutation_event.`CHR`, mutation_event.`START_POSITION`, " +
                     "mutation_event.`END_POSITION`, mutation_event.`REFERENCE_ALLELE`, mutation_event.`TUMOR_SEQ_ALLELE`) AS MUTATION_COUNT " +
-                    "FROM `mutation` , `genetic_profile`, `mutation_event` " +
-                    "WHERE mutation.`GENETIC_PROFILE_ID` = genetic_profile.`GENETIC_PROFILE_ID` " +
-                    "AND mutation.`MUTATION_EVENT_ID` = mutation_event.`MUTATION_EVENT_ID` " +
-                    "AND mutation.`MUTATION_STATUS` <> 'GERMLINE' " +
+                    "FROM `sample_profile` " +
+                    "LEFT JOIN mutation ON mutation.`SAMPLE_ID` = sample_profile.`SAMPLE_ID`" +
+                    "LEFT JOIN mutation_event ON mutation.`MUTATION_EVENT_ID` = mutation_event.`MUTATION_EVENT_ID`" +
```
inodb: @ersinciftci maybe good to add a comment to this query - just to clarify that we're counting 0 when the sample is profiled. No count in MUTATION_COUNT when it hasn't been

ersinciftci: @inodb done

inodb:
@ersinciftci this uses `REPLACE` for the mutation count insert: https://github.com/inodb/cbioportal/commit/6358abe510701e2e481193cb080ca4d918465cbd so we can handle the case of having mutation data split over multiple files. Seems like main problem was the bulk insert through DaoClinicalData. That always does an INSERT, even if you change `SAMPLE_INSERT` as we noticed before, because it loads data through `LOAD DATA LOCAL INFILE`. 

Another solution is to extend DaoClinicalData to add a `replaceDatum` method or something like that, which never uses the bulk loader, but this seemed clean enough to me.

Issue #4916, opened 2018-10-02
Spring contexts and GlobalProperties use separate logic to read portal.properties


fedde-s:
The logic on which configuration files override which for the  is duplicated between at least three files: `GlobalProperties.java` (used for most customisation, and since PR #4787 also database credentials), `applicationContext-business.xml` (used for a few of the integrations with other web services and some details of caching and database integration) and `applicationContext-security.xml` (used for auth config). The logic isn't identical between the three, and this has already bitten me in #4905.

Apart from the bug referenced above, I believe that the Spring config in the `applicationContext-*.xml` files refer to the compiled-in file copied from `portal.properties.EXAMPLE` whenever a property is not specified in the runtime `$PORTAL_HOME/portal.properties`, whereas non-Spring configuration in `GlobalProperties.java` only refers to the baked-in file if no runtime file was supplied. This is difficult to document and maintain.

jjgao:
@fedde-s @alisman : Maybe we should list and review the different configuration files and how they are being used?

fedde-s:
@alisman: the issue I reported was about a need to refactor to avoid growing some technical debt across modules of the backend, written in Java and in Spring XML. Could you explain how that's now about the frontend? Or was that a misclick?

fedde-s:
@jjgao: I think that ensuring that the various configuration files are clearly documented is better handled as a separate issue; more closely related to #4774. I reported this issue here just to keep track of some technical debt in how the logic to locate `portal.properties` is implemented.

alisman:
I don't think this ticket is frontend related.   We do of course use portal.properties for frontend configuration properties but it doesn't sound like that's what's at issue.

Issue #4832, opened 2018-09-18
Document `-D` parameter configuration


inodb:
Following up on https://github.com/cBioPortal/cbioportal/pull/4787:

- [x] copy over runtime configuration for all parameters in GlobalProperties using resolveStringValue from heltgrewe's PR https://github.com/bihealth/cbioportal/compare/1352749b72a03db5a9091f79ceead4e268503b4c...bihealth:feature/spring-style-properties
- [x] Make sure the properties work
- [x] Document the following priority order of property loading:
    ```
    1. -D parameters
    2. PORTAL_HOME/portal.properties
    3. baked in portal.properties
    ```
- [x] Document that one can use `dbconnector=dbcp` so one doesn't need to set the db parameters in a tomcat recourse

jjgao:
@inodb could you update this one?

pieterlukasse:
@inodb the order should be 
1. PORTAL_HOME/portal.properties
2. -D parameters
right? Can we remove the baked in portal.properties altogether? See also discussion in https://github.com/cBioPortal/cbioportal/issues/5982

pieterlukasse:
Should `dbcp` be the default? Would be nice if we don't need to set this explicitly.

fedde-s:
I'd expect the way to specify options that is supplied last, closest to the runtime behaviour itself, to override any properties set earlier, in less flexible ways. So command-line parameters to the server process take precedence over environment variables and runtime configuration files, which in turn override compile-time options (if we continue to support those), which override defaults in the source code itself.

fedde-s:
Also, isn't the first of the four check marks resolved already? Most, if not all, properties that are retrieved by methods of `core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java` should be overridable by `-D`-style Java system properties since the changes we implemented in [commit cbcbd05 of pull request 4787](https://github.com/cBioPortal/cbioportal/pull/4787/commits/cbcbd050ff224cbad75aedd5a062db859eedf870#diff-d5234aefb7508dcc8a8f17cb9684d495).

inodb:
@fedde-s @pieterlukasse made a PR here: https://github.com/cBioPortal/cbioportal/pull/6046. We can discuss further

pieterlukasse:
[...]

PR #4917, opened 2018-10-02
Make security-spring respect runtime configuration

### LINKED ISSUE #4905: security-spring does not honour runtime configuration
PR #4787 moved the recommended way to configure installations to `$PORTAL_HOME/portal.properties`. However, the `security/security-spring/` component (or at least the Keycloak integration code within it) bypasses the central logic on where configuration comes from and instead uses `applicationContext-security.xml`.

fedde-s:
In PR #4787, I adjusted the central logic on reading configuration files, as
well as the deployment manual, moving the main way to supply
configuration to $PORTAL_HOME/portal.properties at runtime. Later, I
found out that the security-spring applicationContext uses its own logic to read the same
files, and this logic was broken.

As a quick fix to make configuration of auth settings work again, adjust
security-spring's duplicate logic to behave like GlobalProperties at
least when these settings are actually listed in the runtime
portal.properties.

Turn `authenticate=` into a required JVM system property (`-D` arg) when using runtime configuration, as it's used before `portal.properties` is read.

Fix #4905. The main issue, at least; the remaining maintainability and deployability issue I logged as #4916.

I pointed this pull request to `rc` rather than `master` even though it's a bugfix, because it only really matters after #4787.

# Checks
- [x] Runs on Heroku.

### DIFF
--- a/docs/portal.properties-Reference.md
+++ b/docs/portal.properties-Reference.md
@@ -73,11 +73,17 @@ googleplus.consumer.secret=2jCfg4SPWdGfXF44WC588dK
 ```
 (note: these are just examples, you need to get your own) You will also need to go to "Google+ API" and click Enable button. In case of problems make sure to enable DEBUG logging for org.springframework.social and org.springframework.security.web.authentication.
 
-
-To active password authentication, then the following properties are required:
+To activate password authentication,
+pass the `-Dauthenticate=googleplus` argument to the web server JVM.
+This can be done by adding the following line to the Tomcat config file
+`$CATALINA_HOME/bin/setenv.sh`:
+```sh
+CATALINA_OPTS='-Dauthenticate=googleplus'
 ```
-app.name=
-authenticate=googleplus
+
+In addition, set these properties in `portal.properties`:
+```
+app.name=cbioportal
 authorization=true
 ```
 app.name should be set to the name of the portal instance referenced in the "AUTHORITY" column of the "AUTHORITIES" table.  See the [User Authorization](User-Authorization.md) for more information.
--- a/app.json
+++ b/app.json
@@ -21,7 +21,7 @@
     },
     "SPRING_OPTS": {
         "description":"set spring properties with e.g. -Dshow.civic=true (TODO: not all props work atm)",
-        "value":"-Dtomcat.catalina.scope=runtime -Ddb.user=cbio_user -Ddb.password=cbio_pass -Ddb.portal_db_name=public_test_small -Ddb.connection_string=jdbc:mysql://devdb.cbioportal.org:3306/ -Ddb.host=devdb.cbioportal.org -Dshow.civic=true -Ddbconnector=dbcp -Dsuppress_schema_version_mismatch_errors=true"
+        "value":"-Dauthenticate=false -Dtomcat.catalina.scope=runtime -Ddb.user=cbio_user -Ddb.password=cbio_pass -Ddb.portal_db_name=public_test_small -Ddb.connection_string=jdbc:mysql://devdb.cbioportal.org:3306/ -Ddb.host=devdb.cbioportal.org -Dshow.civic=true -Ddbconnector=dbcp -Dsuppress_schema_version_mismatch_errors=true"
     }
   },
   "buildpacks": [
--- a/docs/Authenticating-Users-via-LDAP.md
+++ b/docs/Authenticating-Users-via-LDAP.md
@@ -23,13 +23,15 @@ The cBioPortal code has no means of storing user name and passwords and no means
 # Configuring LDAP within cBioPortal
 
 
-## Modifying portal.properties
+## Modifying configuration
 
-Within portal.properties, modify the section labeled `authentication`.  For example:
+Make Tomcat pass the authentication method as a JVM argument
+by adding this line to `$CATALINA_HOME/bin/setenv.sh`:
 
-    # authentication
-    authenticate=ldap
-    
+    CATALINA_OPTS='-Dauthenticate=ldap'
+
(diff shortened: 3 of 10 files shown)


fedde-s:
I'm still testing whether the fix actually works.

fedde-s:
It doesn't seem to, but I'm not familiar enough with Spring to figure this out on my own. Is the syntax I'm trying to use correct? Previously, I believe, the XML file tried to use the `propertyPlaceholderConfigurer` bean to determine the value of `$PORTAL_HOME` as part of a parameter to initialise the `propertyPlaceholderConfigurer` bean. I tried a different approach in commit https://github.com/cBioPortal/cbioportal/pull/4787/commits/39c6c073902d5b5f02b6999f35436ecc9947430c and copied it here, but now I still suspect that [`web.xml`](https://github.com/cBioPortal/cbioportal/blob/39c6c073902d5b5f02b6999f35436ecc9947430c/portal/src/main/webapp/WEB-INF/web.xml#L26) is not setting the right profile based on `$PORTAL_HOME/portal.properties` at runtime.

fedde-s:
I found something really puzzling in `web.xml`. It looks like Maven transforms these lines:
```xml
    <context-param>
        <param-value>${dbconnector:jndi},${authenticate}</param-value>
    </context-param>
```
into these:
```xml
    <context-param>
        <param-value>${dbconnector:jndi},false</param-value>
    </context-param>
```

Making the authenticate setting non-configurable.

fedde-s:
I think that requires changes to some `pom.xml` file?

fedde-s:
No, that probably wouldn't work. I think `web.xml` cannot access the configuration in `portal.properties` at all, so if we want to make `authenticate` configurable and use it in `web.xml`, it would have to be a Java system property, supplied to the JVM via a `-D` parameter in `CATALINA_OPTS`.

fedde-s:
I can't currently think of a way to set the profile that satisfies all three of these criteria:
- [ ] behaves as before for installations that have authentication enabled in the buildtime properties and do not supply runtime properties,
- [ ] respects the runtime properties if they are supplied and the buildtime properties were left untouched (leaving Maven to copy in `portal.properties.EXAMPLE`), and
- [ ] defaults to `false` if the runtime properties do not mention `authenticate` and the buildtime properties were left untouched (leaving Maven to copy in `portal.properties.EXAMPLE`)

fedde-s:
A solution that basically satisfies two of these, would be removing the `authenticate` property from `portal.properties.EXAMPLE`, and documenting that it should *always* be supplied as `-D` property in `$CATALINA_OPTS` (see https://github.com/cBioPortal/cbioportal/issues/4832). It would probably mean that the server crashes if neither is supplied though.

mandawilson:
@fedde-s We run multiple web apps on the same Tomcat with different authentication methods.  We would have to segregate our web apps into different Tomcats where the authentication method needed was the same for all apps, is that correct?  Or we could just choose to continue to put the authenticate property into our application-specific classpath:portal.properties files, correct?  We would continue to things as we used to, building different wars for each application.

fedde-s:
The solution I proposed would still be a step forward though: everyone else, deploying a separate process for each cBioPortal server (maybe using Docker), can start using runtime configuration that way.

fedde-s:
Unless you guys have an idea on how to access the `authenticate=` property from `portal.properties` to set the profile before `web.xml` starts `applicationContext-security.xml`, the code I pushed now gets runtime configuration to work without breaking pre-existing buildtime Maven filtering.

It does come with the somewhat awkward requirement that the JVM parameter must *always* be supplied if not using the old buildtime configuration setup, as I'm not aware of a syntax to set the default that can interchangeably be parsed by both Maven and Spring.

fedde-s:
@pieterlukasse suggested that not having a default for the `authenticate=` property may actually be a good thing: it can help avoid accidentally publishing sensitive or unlicensed data.

fedde-s:
I tested it with a different Dockerfile, and found out that something still seems to get hard-coded from `portal.properties.EXAMPLE` somewhere. If I don't replace `portal.properties.EXAMPLE` with an empty file, I get this error at runtime:

```
Failed to instantiate [org.springframework.security.saml.key.JKSKeyManager]: Constructor threw exception; nested exception is java.lang.RuntimeException: Error initializing keystore
```

But this time I don't think it's in `web.xml`, so it might be easier to fix.

fedde-s:
Found it—`applicationContext-security.xml` was configured to be edited by Maven at buildtime. It does contain a `propertyPlaceHolderConfigurer` to read the properties from the right `portal.properties` file (baked-in or `$PORTAL_HOME/`) itself though, so I that can safely be removed. I tested it with SAML using both buildtime and runtime configuration.

Sjoerd-van-Hagen:
@n1zea144 hi Ben, can we try to move this one along?

docs/portal.properties-Reference.md
```diff
@@ -73,11 +73,17 @@ googleplus.consumer.secret=2jCfg4SPWdGfXF44WC588dK
 ```
 (note: these are just examples, you need to get your own) You will also need to go to "Google+ API" and click Enable button. In case of problems make sure to enable DEBUG logging for org.springframework.social and org.springframework.security.web.authentication.
 
+To activate password authentication,
```
sheridancbio: It seems like there should be a parallel document "Authenticating-and-Authorizing-Users-via-Googleplus.md" rather than showing that method here. Instead I would have put references here to the three/four/five? available guides on how to configure for different authentication methods.
But that change probably belongs in a different PR

inodb:
i added one commit such that heroku is public by default

sheridancbio:
This code has been built into the latest importers at mskcc and has not caused problems. Merging PR now.

PR #4918, opened 2018-10-03
Feature data access tokens

sheridancbio:
- introduced DataAccessToken model for both persistence and API response [JSON mapping]
    - experimenting with full upgrade to spring 5.0.7 throughout code (pom.xml)
    - jwtUtils.createToken() now returns DataAccessToken class instance (full info)
    - web unit tests moved to security module (dependant on security utils)
    - api and service interface modified to return DataAccessToken model rather than string when appropriate
    - added service for retrieving full token info (getDataAccessTokenInfo) from plain token text
    - added service/endpoint for revoking a single token
    - JWT services now throw UnsupportedOperation for all methods involving persistence (revoke, retrieve)
    - some additional ExceptionHandlers still needed .. such as DataAccessTokenNotFoundException for revoke / getInfo methods


web/src/main/java/org/cbioportal/weblegacy/DataAccessTokenController.java
```diff
@@ -69,54 +80,64 @@ private void initializeUsersWhoCannotUseTokenSet() {
...
-                                    @RequestParam(required = false) Boolean allowRevocationOfOtherTokens) {
+    @RequestMapping(method = RequestMethod.POST, value = "/dataAccessToken", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<DataAccessToken> createDataAccessToken(Authentication authentication,
+                                    @RequestParam(required = false) Boolean allowRevocationOfOtherTokens) throws HttpClientErrorException {
         HttpHeaders responseHeaders = new HttpHeaders();
-        String createdToken = dataAccessTokenService.createDataAccessToken(getAuthenticatedUser(authentication), allowRevocationOfOtherTokens);
+        responseHeaders.setContentType(MediaType.APPLICATION_JSON);
+        DataAccessToken createdToken = dataAccessTokenService.createDataAccessToken(getAuthenticatedUser(authentication), allowRevocationOfOtherTokens);
         if (createdToken == null) {
-            return new ResponseEntity<String>("Unable to create a new token.", responseHeaders, HttpStatus.NOT_FOUND);
```
mandawilson: We should probably create a new exception + handler for this.

PR #4935, opened 2018-10-03
Fix config

alisman:
Change configuration service so that it bypasses (mostly) custom methods in GlobalProperties, using only that classes getProperty method to retrieve from portal.properties

default values will now be determined on frontend.  i documented them in this spreadsheet

https://docs.google.com/spreadsheets/d/1eC0-mU-kdmiInmi1LZRj_-L-oZofmTNOzBYBMr12eUE/edit#gid=192138605

my process has been as follows:
1. JJ marked ~50 properties which are relevant to frontend
2. I combed through GlobalProperties to determine what the defaults are and put those defaults in a configuration object on frontend.  

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


portal/src/main/webapp/config_service.jsp
```diff
@@ -1,40 +1,12 @@
...
-}
-studies += "}";
-
-%>
-
-<%= request.getParameter("callback") %>({
+<%--
+legacyConfig = 
+{
     enableDarwin : <%=CheckDarwinAccessServlet.CheckDarwinAccess.existsDarwinProperties()%>,
```
alisman: i just left this legacy config here for reference.  will delete before merge

portal/src/main/webapp/config_service.jsp
```diff
@@ -88,8 +63,106 @@ studies += "}";
...
+            "skin.show_web_api_tab",
+            "skin.title",
+            "skin.authorization_message",
+            "mdacc.heatmap.study.meta.url",
+            "mdacc.heatmap.study.url",
+            "oncoprint.custom_driver_annotation.tiers.menu_label"
+        }; 
+    
+        // for each above, add json prop and lookup value in portal.properties
+        for (int i = 0; i < propNameArray.length; i++){
```
alisman: i'm sure there is some better way to serialize all this stuff but this works pretty well.  actually, there i still one problem which is that boolean values are delivered as strings.  I correct this on frontend but we should figure that out here.

alisman: ditto numeric values

PR #4951, opened 2018-10-05
Fix #4950 split mutations data with same stable_id

### LINKED ISSUE #4950: Internal Importer Issue fail when importing two data_mutation_extended.txt with same `stable_id`
We do this for instance for fusion data to make it show up in oncoprint

Related to: https://github.com/cBioPortal/cbioportal/issues/4895 and https://github.com/cBioPortal/cbioportal/pull/4907

inodb:
This fixes the problem when importing mutation data split over two files with
the same stable_id. This would throw a key constraint error before because that
combination of genetic_profile, sample_id and MUTATION_COUNT would already
exist.


core/src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
```diff
@@ -469,23 +469,6 @@ private static String generateIdsSql(Collection<Integer> ids) {
         return "'" + StringUtils.join(ids, "','") + "'";
     }
 
-    public static void deleteRecords(String attrId) throws DaoException {
```
inodb: this was added before to get the tests to pass

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -106,6 +129,54 @@ public void testImportMutationsFile() throws Exception {
         ImportProfileData secondRunner = new ImportProfileData(secondArgs);
         secondRunner.run();
```
inodb: this would fail before because of the key constraint

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -106,6 +129,54 @@ public void testImportMutationsFile() throws Exception {
...
+        // also confirm mutation count for first sample is still correct
+        clinicalData = DaoClinicalData.getSampleData(study.getInternalId(), new ArrayList<String>(Arrays.asList("TCGA-AA-3664-01", "TCGA-AA-3665-01")), clinicalAttribute);
+        assert(clinicalData.size() == 2);
+        assertEquals("8", clinicalData.get(0).getAttrVal());
+        assertEquals("1", clinicalData.get(1).getAttrVal());
+    }
+
+    @Test
+    public void testImportSplitMutationsFile() throws Exception {
+        /* Mutations file split over two files with same stable id */
```
inodb: the test above tests a lot of things, this makes it a little cleaner by just testing the counting in the split mutation files. We can add some fusion data related tests in the PR that addresses the issue with counting fusion and germline only samples

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportExtendedMutationData.java
```diff
@@ -292,8 +293,16 @@ private void checkGermlineMutations() throws DaoException {
 
     private void checkMutationCounts() throws DaoException {
 
-        ClinicalData clinicalData = DaoClinicalData.getDatum("study_tcga_pub", "TCGA-AA-3664", "MUTATION_COUNT");
```
inodb: This seemed to have been checking `MUTATION_COUNT` using a patient id before. Not sure how this was working before, gets a null exception now. I think `MUTATION_COUNT` should only be at the sample level, so seems ok to change

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -137,10 +137,22 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
...
+            }
+
             // do not include germline and fusions (msk internal) when counting mutations
             // we do not add the MUTATION_COUNT clinical data for the sample if it's not profiled
+            // use replace to overwrite existing counts if they exist (e.g. when
+            // the mutation data is split over multiple files with same profile
+            // id)
             pstmt = con.prepareStatement(
-                    "SELECT sample_profile.`SAMPLE_ID`, COUNT(DISTINCT mutation_event.`CHR`, mutation_event.`START_POSITION`, " +
+                    "REPLACE `clinical_sample` " +
```
sheridancbio: REPLACE is an extension to standard SQL syntax ... we are already tied to the mysql DBMS, so maybe this is not an issue, but this would be one more thing that a deployer using a different (DB2?) DBMS would need to port.
Also, I needed to look up this syntax and read docs in order to know what was being done, so there is additional overhead on developers.
On the other hand, this command does seem to do exactly what we need, and concisely.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -87,16 +95,33 @@ public void testImportMutationsFile() throws Exception {
         runner.run();
         //This test is to check if the ImportProfileData class indeed adds the study stable Id in front of the
         //dataset study id (e.g. studyStableId + "_breast_mutations"):
-        DaoClinicalData.deleteRecords("MUTATION_COUNT");
-        MySQLbulkLoader.flushAll();
         String studyStableId = "study_tcga_pub";
-        studyId = DaoCancerStudy.getCancerStudyByStableId(studyStableId).getInternalId();
+        CancerStudy study = DaoCancerStudy.getCancerStudyByStableId(studyStableId);
```
sheridancbio: This is a complex test .. maybe a comment at the start of the test function explaining what conditions are being tested and how would be helpful. The comment about whether the studyStableId is added in front or not is obsolete / misleading.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -106,6 +131,62 @@ public void testImportMutationsFile() throws Exception {
...
+        // assume a MUTATION_COUNT record has been added for the sample and the
+        // count is 1, the other one is a germline mutation
+        // also confirm mutation count for first sample is still correct
+        clinicalData = DaoClinicalData.getSampleData(study.getInternalId(), new ArrayList<String>(Arrays.asList("TCGA-AA-3664-01", "TCGA-AA-3665-01")), clinicalAttribute);
+        assert(clinicalData.size() == 2);
+        assertEquals("8", clinicalData.get(0).getAttrVal());
+        assertEquals("1", clinicalData.get(1).getAttrVal());
+        // remove profile at the end
+        DaoGeneticProfile.deleteGeneticProfile(geneticProfile);
+        assertNull(DaoGeneticProfile.getGeneticProfileByStableId(studyStableId + "_breast_mutations"));
```
sheridancbio: this unit test is not checking whether the deleteGeneticProfile() works properly, so we may not need this assert

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -106,6 +131,62 @@ public void testImportMutationsFile() throws Exception {
...
+        ImportProfileData secondRunner = new ImportProfileData(secondArgs);
+        secondRunner.run();
+        // assume a MUTATION_COUNT record has been updated for both samples (both +1)
+        clinicalData = DaoClinicalData.getSampleData(study.getInternalId(), new ArrayList<String>(Arrays.asList("TCGA-AA-3664-01", "TCGA-AA-3665-01")), clinicalAttribute);
+        assert(clinicalData.size() == 2);
+        assertEquals("4", clinicalData.get(0).getAttrVal());
+        assertEquals("1", clinicalData.get(1).getAttrVal());
+        // remove profile at the end
+        DaoGeneticProfile.deleteGeneticProfile(geneticProfile);
+        assertNull(DaoGeneticProfile.getGeneticProfileByStableId(studyStableId + "_breast_mutations"));
```
sheridancbio: this unit test is not checking whether the deleteGeneticProfile() works properly, so we may not need this assert

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -106,6 +131,62 @@ public void testImportMutationsFile() throws Exception {
...
+                "--data","src/test/resources/splitMutationsData/data_mutations_extended_continued.txt",
+                "--meta","src/test/resources/splitMutationsData/meta_mutations_extended.txt",
+                "--loadMode", "bulkLoad"
+        };
+        ImportProfileData secondRunner = new ImportProfileData(secondArgs);
+        secondRunner.run();
+        // assume a MUTATION_COUNT record has been updated for both samples (both +1)
+        clinicalData = DaoClinicalData.getSampleData(study.getInternalId(), new ArrayList<String>(Arrays.asList("TCGA-AA-3664-01", "TCGA-AA-3665-01")), clinicalAttribute);
+        assert(clinicalData.size() == 2);
+        assertEquals("4", clinicalData.get(0).getAttrVal());
```
sheridancbio: nice clear test

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
inodb: The differing dbms argument will become less of an issue as we move to container based deployments. So not too worried about this in the long term.

I can add a comment in the code that replace is basically insert+update depending on whether data is available

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
inodb: yeah that test happens a few lines down, ill put a comment at the top that describes the entire test

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
inodb: this could be done a bit cleaner by doing this in `@After`, since each test assumes the mutation profile doesn't exist. I do think it's good to make sure there that the profile is completely deleted even though that's not what's being tested. Otherwise it might be p unclear why these tests are failing. It's kinda like using it as an `assume that x` rather than the test itself

inodb:
@sheridancbio tried to address your comments, updated the PR

sheridancbio:
This code is running in the latest built importers. If the mskimpact study imports successfully tonight, this PR (and #4895) can be merged.

PR #4962, opened 2018-10-08
UUID implementation for data access tokens

averyniceday:
This is the first iteration of the UUID implementation (just to verify everything is connected and does what's expected). I am working on polishing up the code. 

# What? Why?
UUID implementation as an alternative to JWT. Allows for token management (revoking outdated or compromised tokens) and longer TTL. 

- additional data members in DataAccessToken class
- removed constructor to allow automatic mapping
- repository layer which interacts with sql database through mybatis
    - get all data access tokens (given a username)
    - get a data access token (given a token identifier)
    - insert a new record
    - delete a record
- UUID service implementation
    - creating a data access token will delete oldest token if user already has max number of tokens allowed

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


model/src/main/java/org/cbioportal/model/DataAccessToken.java
```diff
@@ -1,22 +1,36 @@
...
 
 import java.io.Serializable;
+import java.util.Date;
 import javax.validation.constraints.NotNull;
 
 public class DataAccessToken implements Serializable {
 
     private String token;
-
-    public DataAccessToken(String token) {
```
averyniceday: having the constructor defined in the class was breaking the auto-mapping behavior when using mybatis. Planning to look into this more.

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+@Component
+public class UuidDataAccessTokenServiceImpl implements DataAccessTokenService {
+
+    @Autowired
+    private DataAccessTokenRepository dataAccessTokenRepository;
+
+    @Value("${jwt.ttl_seconds}")
+    private int datTtlSeconds;
+
+    @Value("${max.number.dat}")
```
sheridancbio: for grouping purposes, maybe all dat properties should start with dat ... dat.max_number_per_user?

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    private int maxNumberOfAccessTokens;
+
+    private static final Log log = LogFactory.getLog(UuidDataAccessTokenServiceImpl.class);
+
+    // create a data access token (randomly generated UUID) and insert corresponding record into table with parts:
+    // username
+    // uuid
+    // expiration date (current time + 1 month)
```
sheridancbio: recommend adding creation date

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.SECOND, datTtlSeconds);
+        Date expirationDate = calendar.getTime();
+       
+        DataAccessToken dataAccessToken = new DataAccessToken(); 
+        dataAccessToken.setToken(uuid);
+        dataAccessToken.setUsername(username);
+        dataAccessToken.setExpiration(expirationDate);
+        if (getNumberOfTokensForUsername(username) == maxNumberOfAccessTokens) {
+            revokeOldestDataAccessTokenForUsername(username);
```
sheridancbio: Here is where the argument to the function should be used. If allowRevocationOfOtherTokens is false, we should not revoke this token, and should fail with a thrown exception (maybe we need a new exception class like "DataAccessTokenUnableToCreateException")
Also, maybe this revoke logic should happen before we make the effort to generate a new UUID

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    // uuid
+    // expiration date (current time + 1 month)
+    @Override
+    public DataAccessToken createDataAccessToken(String username, boolean allowRevocationOfOtherTokens) {
+        if (username == null || username.trim().length() == 0) {
+            throw new IllegalArgumentException("subject cannot be empty");
+        }
+        String uuid = UUID.randomUUID().toString();
+        Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.SECOND, datTtlSeconds);
```
sheridancbio: Before this, we can pull out the creation time from the calendar

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    public DataAccessToken getDataAccessTokenInfo(String token) {
+        DataAccessToken dataAccessToken = dataAccessTokenRepository.getDataAccessToken(token);
+        return dataAccessToken;
+    }
+
+    @Override
+    public void revokeAllDataAccessTokens(String username) {
+        List<DataAccessToken> allDataAccessTokens = dataAccessTokenRepository.getAllDataAccessTokensForUsername(username);
+        for (DataAccessToken dataAccessToken : allDataAccessTokens) {
+            dataAccessTokenRepository.removeDataAccessToken(dataAccessToken.getToken());
```
sheridancbio: we can avoid this loop by adding the "removeAllDataAccessTokensForUser" mybatis mapper, which can do it with one step

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        return dataAccessToken.getExpiration();
+    }
+   
+    @Override
+    public Boolean isValid(String dataAccessToken) {
+        DataAccessToken storedDataAccessToken = null;
+        try {
+            storedDataAccessToken = dataAccessTokenRepository.getDataAccessToken(dataAccessToken);
+        } catch (Exception e) {
+            log.error("Error retrieving data access token, " + dataAccessToken + " from token store");
```
sheridancbio: we should probably distinguish between not finding the DAT in the table (DataAccessTokenDoesNotExistException) from the "something went wrong with mysql server just now" case (DataAccessTokenRetrievalException)

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        Calendar calendar = Calendar.getInstance();
+        Date currentDate = calendar.getTime();
+        if (storedDataAccessToken.getExpiration().before(currentDate)) {
+            return Boolean.FALSE;
+        } 
+        return Boolean.TRUE;
+    }
+
+    private int getNumberOfTokensForUsername(String username) {
+        List<DataAccessToken> allDataAccessTokens = dataAccessTokenRepository.getAllDataAccessTokensForUsername(username);
```
sheridancbio: This can be made into a custom query which just returns the results of select count(*) from data_access_token where username = `$?` .. that way less data is passed around and ultimately ignored

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        return allDataAccessTokens.size();
+    }
+
+    // revokes oldest token in token management system for a user
+    private void revokeOldestDataAccessTokenForUsername(String username) {
+        List<DataAccessToken> allDataAccessTokens = dataAccessTokenRepository.getAllDataAccessTokensForUsername(username);
+        DataAccessToken oldestDataAccessToken = getOldestDataAccessToken(allDataAccessTokens);
```
sheridancbio: finding the oldest token can be done within mysql using an ORDER BY clause and an index in the table for the expiration date. This will be faster than linearly scanning the tokens below.

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        dataAccessTokenRepository.removeDataAccessToken(oldestDataAccessToken.getToken());
+    }
+    
+    // helper function for getting the newest data access token out of a list of data access tokens
+    private DataAccessToken getNewestDataAccessToken(List<DataAccessToken> dataAccessTokens) {
+        DataAccessToken newestDataAccessToken = null;
+        for (DataAccessToken dataAccessToken : dataAccessTokens) {
```
sheridancbio: similar to getOldest, this can be more efficient by using an ORDER BY clause in the mysql with an index rather than a linear scan.

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,203 @@
...
+            newestDataAccessToken = (newestDataAccessToken != null) ? (getNewerDataAccessToken(newestDataAccessToken, dataAccessToken)) : (dataAccessToken);
+        }
+        return newestDataAccessToken;
+    }
+
+    // helper function for comparing two data access tokens and returning the newer one
+    private DataAccessToken getNewerDataAccessToken(DataAccessToken newestDataAccessToken, DataAccessToken dataAccessToken) {
```
sheridancbio: rather than a method here to compare two tokens, this could be added to the model class so that other places the code would be able to compare to tokens.
public boolean expiresSoonerThan(DataAccessToken otherToken);

service/src/main/java/org/cbioportal/service/util/JwtUtils.java
```diff
@@ -92,9 +92,11 @@ public DataAccessToken createToken(String subject) {
             .setIssuedAt(creationDate)
             .setExpiration(expirationDate)
             .signWith(SignatureAlgorithm.HS256, decodedSecretKey).compact();
-        return new DataAccessToken(jws);
+        DataAccessToken dataAccessToken = new DataAccessToken();
```
sheridancbio: can be reverted after default constructor is added

web/src/main/java/org/cbioportal/weblegacy/DataAccessTokenController.java
```diff
@@ -87,7 +87,7 @@ private void initializeUsersWhoCannotUseTokenSet() {
         responseHeaders.setContentType(MediaType.APPLICATION_JSON);
         DataAccessToken createdToken = dataAccessTokenService.createDataAccessToken(getAuthenticatedUser(authentication), allowRevocationOfOtherTokens);
         if (createdToken == null) {
-            return new ResponseEntity<>(new DataAccessToken(null), responseHeaders, HttpStatus.NOT_FOUND);
```
sheridancbio: can be reverted after default DataAccessToken constructor is added

db-scripts/src/main/resources/migration.sql
```diff
@@ -489,4 +499,54 @@ HAVING SUM(`END`-`START`) > 0;
...
+FROM copy_number_seg AS c2 WHERE c2.`CANCER_STUDY_ID` = c1.`CANCER_STUDY_ID` AND c2.`SAMPLE_ID` = c1.`SAMPLE_ID`
+AND ABS(c2.`SEGMENT_MEAN`) >= 0.2) IS NULL, 0, (SELECT SUM(`END`-`START`) FROM copy_number_seg AS c2
+WHERE c2.`CANCER_STUDY_ID` = c1.`CANCER_STUDY_ID` AND c2.`SAMPLE_ID` = c1.`SAMPLE_ID`
+AND ABS(c2.`SEGMENT_MEAN`) >= 0.2) / SUM(`END`-`START`)) AS `VALUE` FROM `copy_number_seg` AS c1 , `cancer_study`
+WHERE c1.`CANCER_STUDY_ID` = cancer_study.`CANCER_STUDY_ID` GROUP BY cancer_study.`CANCER_STUDY_ID` , `SAMPLE_ID`
+HAVING SUM(`END`-`START`) > 0;
+
+UPDATE `info` SET `DB_SCHEMA_VERSION`="2.7.2";
+
+##version: 2.7.3
```
averyniceday: new code starting here -- additional changes above this were taken from the MUTATION_COUNTS pull request

sheridancbio:
Unit test failures due to missing functions in Guava?? Not sure what happened there. Maybe we need to rebase this branch in front of rc ... or check the dependency tree and make sure we are linking to the right guava version for our code.

service/src/main/java/org/cbioportal/service/DataAccessTokenService.java
```diff
@@ -34,6 +34,7 @@
 
 import java.util.*;
 import org.cbioportal.model.DataAccessToken;
+import org.cbioportal.service.exception.MaxNumberTokensExceededException;
```
sheridancbio: not needed

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,167 @@
...
+    // create a data access token (randomly generated UUID) and insert corresponding record into table with parts:
+    // username
+    // uuid
+    // expiration date (current time + 1 month)
+    @Override
+    public DataAccessToken createDataAccessToken(String username, boolean allowRevocationOfOtherTokens) {
+        if (username == null || username.trim().length() == 0) {
+            throw new IllegalArgumentException("subject cannot be empty");
+        } 
+        if (getNumberOfTokensForUsername(username) == maxNumberOfAccessTokens) {
```
sheridancbio: use >= for test

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public void revokeAllDataAccessTokens(String username) {
+        dataAccessTokenRepository.removeAllDataAccessTokensForUsername(username);
+    }
+
+    @Override
+    public void revokeDataAccessToken(String token) {
+        DataAccessToken dataAccessToken = dataAccessTokenRepository.getDataAccessToken(token);
+        dataAccessTokenRepository.removeDataAccessToken(dataAccessToken.getToken());
```
sheridancbio: test: if it doesn't exist what happens?

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,167 @@
...
+        DataAccessToken storedDataAccessToken = null;
+        try {
+            storedDataAccessToken = dataAccessTokenRepository.getDataAccessToken(dataAccessToken);
+        } catch (Exception e) {
+            log.error("Error retrieving data access token, " + dataAccessToken + " from token store");
+            return Boolean.FALSE;
+        }
+        Calendar calendar = Calendar.getInstance();
+        Date currentDate = calendar.getTime();
+        if (storedDataAccessToken.getExpiration().before(currentDate)) {
```
sheridancbio: maybe this is switched? ... no.. probably right.

web/src/main/java/org/cbioportal/web/config/CustomObjectMapper.java
```diff
@@ -45,6 +45,7 @@
 import org.cbioportal.model.ClinicalEvent;
 import org.cbioportal.model.ClinicalEventData;
 import org.cbioportal.model.CopyNumberSeg;
+import org.cbioportal.model.DataAccessToken;
```
sheridancbio: maybe this code is duplicated .. see legacy controller

service/src/main/java/org/cbioportal/service/impl/UuidDataAccessTokenServiceImpl.java
averyniceday: if it doesn't exist it just returns a "success" response -- so I think we should keep the getDataAccessToken check first and throw and exception if the response is null

PR #4963, opened 2018-10-08
Fix #4895 MUTATION_COUNT germline/fusion only

### LINKED ISSUE #4895: Handle missing MUTATION_COUNT data

inodb:
If a sample has only germline or fusion events it would not get 0 count. By
including the criteria in the left join this is solved

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -144,11 +144,11 @@ public final class DaoMutation {
                     "mutation_event.`END_POSITION`, mutation_event.`REFERENCE_ALLELE`, mutation_event.`TUMOR_SEQ_ALLELE`) AS MUTATION_COUNT " +
                     "FROM `sample_profile` " +
                     "LEFT JOIN mutation ON mutation.`SAMPLE_ID` = sample_profile.`SAMPLE_ID` " +
+                    "AND ( mutation.`MUTATION_STATUS` <> 'GERMLINE' OR mutation.`MUTATION_STATUS` IS NULL ) " +
                     "LEFT JOIN mutation_event ON mutation.`MUTATION_EVENT_ID` = mutation_event.`MUTATION_EVENT_ID` " +
+                    "AND ( mutation_event.`MUTATION_TYPE` <> 'Fusion' OR mutation_event.`MUTATION_TYPE` IS NULL ) " +
                     "INNER JOIN genetic_profile ON genetic_profile.`GENETIC_PROFILE_ID` = sample_profile.`GENETIC_PROFILE_ID` " +
                     "WHERE genetic_profile.`GENETIC_ALTERATION_TYPE` = 'MUTATION_EXTENDED' " +
-                    "AND ( mutation.`MUTATION_STATUS` <> 'GERMLINE' OR mutation.`MUTATION_STATUS` IS NULL ) " +
-                    "AND ( mutation_event.`MUTATION_TYPE` <> 'Fusion' OR mutation_event.`MUTATION_TYPE` IS NULL ) " +
                     "AND genetic_profile.`GENETIC_PROFILE_ID`=? " +
                     "GROUP BY sample_profile.`GENETIC_PROFILE_ID` , sample_profile.`SAMPLE_ID`;");
             pstmt.setInt(1, geneticProfile.getGeneticProfileId());
--- a/core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
+++ b/core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
@@ -24,6 +24,7 @@
 package org.mskcc.cbio.portal.scripts;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -35,6 +36,7 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.mskcc.cbio.portal.dao.DaoCancerStudy;
+import org.mskcc.cbio.portal.dao.DaoClinicalAttributeMeta;
 import org.mskcc.cbio.portal.dao.DaoClinicalData;
 import org.mskcc.cbio.portal.dao.DaoCnaEvent;
 import org.mskcc.cbio.portal.dao.DaoException;
@@ -45,8 +47,11 @@ import org.mskcc.cbio.portal.dao.DaoSample;
 import org.mskcc.cbio.portal.dao.MySQLbulkLoader;
 import org.mskcc.cbio.portal.model.CancerStudy;
 import org.mskcc.cbio.portal.model.CanonicalGene;
+import org.mskcc.cbio.portal.model.ClinicalAttribute;
+import org.mskcc.cbio.portal.model.ClinicalData;
 import org.mskcc.cbio.portal.model.CnaEvent;
 import org.mskcc.cbio.portal.model.ExtendedMutation;
+import org.mskcc.cbio.portal.model.GeneticProfile;
 import org.mskcc.cbio.portal.util.ConsoleUtil;
 import org.mskcc.cbio.portal.util.ImportDataUtil;
 import org.mskcc.cbio.portal.util.ProgressMonitor;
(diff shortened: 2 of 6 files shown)


db-scripts/src/main/resources/migration.sql
```diff
@@ -515,3 +515,27 @@ AND ( mutation_event.`MUTATION_TYPE` <> 'Fusion' OR mutation_event.`MUTATION_TYP
...
+FROM `sample_profile` 
+LEFT JOIN mutation ON mutation.`SAMPLE_ID` = sample_profile.`SAMPLE_ID`
+AND ( mutation.`MUTATION_STATUS` <> 'GERMLINE' OR mutation.`MUTATION_STATUS` IS NULL )
+LEFT JOIN mutation_event ON mutation.`MUTATION_EVENT_ID` = mutation_event.`MUTATION_EVENT_ID`
+AND ( mutation_event.`MUTATION_TYPE` <> 'Fusion' OR mutation_event.`MUTATION_TYPE` IS NULL )
+INNER JOIN genetic_profile ON genetic_profile.`GENETIC_PROFILE_ID` = sample_profile.`GENETIC_PROFILE_ID`
+WHERE genetic_profile.`GENETIC_ALTERATION_TYPE` = 'MUTATION_EXTENDED'
+GROUP BY sample_profile.`GENETIC_PROFILE_ID` , sample_profile.`SAMPLE_ID`;
+
+DELETE FROM `clinical_sample` WHERE clinical_sample.`ATTR_ID` = 'FRACTION_GENOME_ALTERED';
```
inodb: recompute fraction genome altered, because there was a bug with this before so some studies that were imported might have the wrong value for this

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -108,6 +113,42 @@ public void testImportMutationsFile() throws Exception {
...
+        assertEquals("0", clinicalData.get(0).getAttrVal());
+
+        // check if the three germline mutations have been inserted
+        validateMutationAminoAcid (geneticProfileId, sampleId, 64581, "T209A");
+        validateMutationAminoAcid (geneticProfileId, sampleId, 50839, "G78S");
+        validateMutationAminoAcid (geneticProfileId, sampleId, 2842, "L113P");
+
+        // remove profile at the end
+        DaoGeneticProfile.deleteGeneticProfile(geneticProfile);
+        assertNull(DaoGeneticProfile.getGeneticProfileByStableId(studyStableId + "_breast_mutations"));
```
sheridancbio: since this is not a test of the DaoGeneticProfile.deleteGeneticProfile() function, we probably do not need an assert statement here.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
```diff
@@ -108,6 +113,42 @@ public void testImportMutationsFile() throws Exception {
...
+        String studyStableId = "study_tcga_pub";
+        CancerStudy study = DaoCancerStudy.getCancerStudyByStableId(studyStableId);
+        studyId = study.getInternalId();
+        int sampleId = DaoSample.getSampleByCancerStudyAndSampleId(studyId, "TCGA-AA-3664-01").getInternalId();
+        GeneticProfile geneticProfile = DaoGeneticProfile.getGeneticProfileByStableId(studyStableId + "_breast_mutations");
+        geneticProfileId = geneticProfile.getGeneticProfileId();
+
+        // assume clinical data for MUTATION_COUNT was created
+        ClinicalAttribute clinicalAttribute = DaoClinicalAttributeMeta.getDatum("MUTATION_COUNT", studyId);
+        assertNotNull(clinicalAttribute);
```
sheridancbio: this is the key test ... that the MUTATION_COUNT attribute was computed for the import of this all-GERMLINE (for this sample) profile

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -205,6 +206,20 @@ public void testLoadStudyEs0() throws Throwable {
             assertEquals(3, clinicalAttributes.size());
             clinicalAttributes = apiService.getPatientClinicalAttributes();
             assertEquals(4, clinicalAttributes.size());
+            List<DBClinicalSampleData> clinicalComputedSampleData = apiService.getSampleClinicalData("study_es_0", Arrays.asList("MUTATION_COUNT","FRACTION_GENOME_ALTERED"), Arrays.asList("TCGA-A2-A04P-01"));
+            Boolean mutationCountExists = false;
+            Boolean fractionGenomeAlteredExists = false;
+            for (DBClinicalSampleData dbClinicalSampleData: clinicalComputedSampleData) {
+                if (dbClinicalSampleData.attr_id.equals("MUTATION_COUNT")) {
+                    mutationCountExists = true;
+                    assertEquals("TCGA-A2-A04P-01 has one mutation in MUTATION_COUNT", "1", dbClinicalSampleData.attr_val);
```
sheridancbio: maybe these messages should also be phrased with "should have one mutation ..." as below.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -205,6 +206,20 @@ public void testLoadStudyEs0() throws Throwable {
...
+                } else if (dbClinicalSampleData.attr_id.equals("FRACTION_GENOME_ALTERED")) {
+                    fractionGenomeAlteredExists = true;
+                    assertEquals("TCGA-A2-A04P-01 has 0 FRACTION_GENOME_ALTERED", 0.0, Float.parseFloat(dbClinicalSampleData.attr_val), 0.01);
```
sheridancbio: This is new to me .. the "double is equal within a delta range" construction. But it looks like this is the preferred way, and it does make sense for issues related to precision.

db-scripts/src/main/resources/migration.sql
```diff
@@ -515,3 +515,27 @@ AND ( mutation_event.`MUTATION_TYPE` <> 'Fusion' OR mutation_event.`MUTATION_TYP
 GROUP BY sample_profile.`GENETIC_PROFILE_ID` , sample_profile.`SAMPLE_ID`;
 
 UPDATE `info` SET `DB_SCHEMA_VERSION`="2.7.1";
+
+##version: 2.7.2
+DELETE FROM `clinical_sample` WHERE clinical_sample.`ATTR_ID` = 'MUTATION_COUNT';
+INSERT INTO `clinical_sample` SELECT sample_profile.`SAMPLE_ID`, 'MUTATION_COUNT', COUNT(DISTINCT mutation_event.`CHR`, mutation_event.`START_POSITION`,
```
sheridancbio: unnecessary trailing whitespace on this line, and the next two lines

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportProfileData.java
inodb: i've addressed this in https://github.com/cBioPortal/cbioportal/pull/4951. These lines can be removed after merging that

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
inodb: it actually gave me an error when doing a float comparison without using delta range, so that was nice

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -144,11 +144,11 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
                     "mutation_event.`END_POSITION`, mutation_event.`REFERENCE_ALLELE`, mutation_event.`TUMOR_SEQ_ALLELE`) AS MUTATION_COUNT " +
                     "FROM `sample_profile` " +
                     "LEFT JOIN mutation ON mutation.`SAMPLE_ID` = sample_profile.`SAMPLE_ID` " +
+                    "AND ( mutation.`MUTATION_STATUS` <> 'GERMLINE' OR mutation.`MUTATION_STATUS` IS NULL ) " +
```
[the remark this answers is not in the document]
inodb: got it from here: https://stackoverflow.com/questions/26115649/return-0-if-no-records-found-on-count-query-in-mysql

sandertan:
@inodb Please also update `cgds.sql` to 2.7.2.

inodb:
@sandertan 	https://github.com/cBioPortal/cbioportal/pull/5057

Issue #5015, opened 2018-10-16
performance - new study view


jjgao:
- [x] sql potential improvement (single study vs multiple) #4802
- [ ] filtering improvement (ideally only do filtering once for all ajax calls)
- [x] reduce ajax call by combining all unfiltered pie charts in one call and the same with filtered pie chars. And the same with bar charts.
- [x] scatter plot imporvement (hide it for genie, and support density plot)
- [ ] max simultaneous calls
- [ ] cache (mybatis or custom), e.g. cache by filtering criteria for getting the list of samples (session-based, user-based, global?)

zhx828:
@jjgao @ersinciftci @n1zea144 is this done?

jjgao:
I have removed this from 1.19.0, but let's keep it open - there is still room to improve.

PR #5017, opened 2018-10-16
Support multiple attributes in clinical data count endpoint

ersinciftci:
Partially fixes https://github.com/cBioPortal/cbioportal/issues/5015


web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -81,8 +81,16 @@
...
-        return new ResponseEntity<>(clinicalDataService.fetchClinicalDataCounts(studyIds, sampleIds, 
-            Arrays.asList(attributeId), clinicalDataType.name()).get(attributeId), HttpStatus.OK);
+        Map<String, List<ClinicalDataCount>> resultForSampleAttributes = clinicalDataService.fetchClinicalDataCounts(
+            studyIds, sampleIds, attributes.stream().filter(a -> a.getClinicalDataType().equals(ClinicalDataType.SAMPLE))
+            .map(a -> a.getAttributeId()).collect(Collectors.toList()), ClinicalDataType.SAMPLE.name());
+        Map<String, List<ClinicalDataCount>> resultForPatientAttributes = clinicalDataService.fetchClinicalDataCounts(studyIds, sampleIds, 
+        attributes.stream().filter(a -> a.getClinicalDataType().equals(ClinicalDataType.PATIENT))
+            .map(a -> a.getAttributeId()).collect(Collectors.toList()), ClinicalDataType.PATIENT.name());
+        Map<String, List<ClinicalDataCount>> combinedResult = new HashMap<>();
+        combinedResult.putAll(resultForSampleAttributes);
```
kalletlak: there might be a possibility for same key in `resultForSampleAttributes` and `resultForPatientAttributes`

ersinciftci: actually attribute id is unique for a cancer study in clinical_attribute_meta, so they can't be the same. I just require the clinicalDataType from the frontend so that we wouldn't need to check the clinical_attribute_meta table first and then use the appropriate table when getting clinical data, because frontend already has that information

ersinciftci: Multi-study queries actually can have a sample attribute in one study and the a patient attribute with the same id in another study, so I'll change the response a little bit.

PR #5030, opened 2018-10-18
Change return type of clinical data counts endpoint

ersinciftci:


model/src/main/java/org/cbioportal/model/ClinicalDataCountItem.java
```diff
@@ -0,0 +1,35 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+import java.util.List;
+
+public class ClinicalDataCountItem implements Serializable {
+
+    private String attributeId;
+    private String clinicalDataType;
```
zhx828: The type should be ClinicalDataType

PR #5052, opened 2018-10-19
Update portal.properties.EXAMPLE

tweep:
Added the MySQL port to the connection details. Important if you don't run on the Mysql default port ( 3306 )


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -4,10 +4,10 @@ app.name=cbioportal
 # database
 db.user=cbio_user
 db.password=somepassword
-db.host=localhost
+db.host=localhost:MYSQLPORT
 db.portal_db_name=cbioportal
 db.driver=com.mysql.jdbc.Driver
-db.connection_string=jdbc:mysql://localhost/
+db.connection_string=jdbc:mysql://localhost:MYSQLPORT/
```
inodb: @tweep 	could we make this `:3306` instead? If no properties file is provided it uses this EXAMPLE default, so better to make it connect to 3306

jjgao:
@inodb change to 3306, please review.

PR #5073, opened 2018-10-23
update for "groups" configuration for keycloak

pieterlukasse:
# What? Why?
Updated "groups" configuration section of Keycloak documentation to reflect what we have
been doing in practice.


fedde-s:
Sounds like a good way to make better use of the features of Keycloak, and manage permissions in one place, with the new insights I didn't have when I wrote the initial recommendation 👍

Issue #5132, opened 2018-10-29
Links from study view to static pages not working


inodb:
Still point to e.g. http://www.cbioportal.org/s/datasets

PR #5134, opened 2018-10-29
Remove links to /s/ (old prefix spa static pages)

inodb:


inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/5132

PR #5148, opened 2018-10-30
Fix Molecular Data API sort issue

ersinciftci:
It seems like in Java, `_` comes before letters when sorted, but in MySQL it comes after. The code was depending on them being the same order, so had to do this trick.


ersinciftci:
@adamabeshouse it's all because of these comma separated values in the database, hopefully will get rid of them in the future :)

PR #5198, opened 2018-11-06
Fix redirection of case.do and index.do legacy urls to SPA urls

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/portal/src/main/webapp/WEB-INF/web.xml
+++ b/portal/src/main/webapp/WEB-INF/web.xml
@@ -454,10 +454,6 @@
         <servlet-name>patient_view</servlet-name>
         <servlet-class>org.mskcc.cbio.portal.servlet.PatientView</servlet-class>
     </servlet>
-    <servlet-mapping>
-        <servlet-name>patient_view</servlet-name>
-        <url-pattern>/case.do</url-pattern>
-    </servlet-mapping>
 
     <servlet>
         <servlet-name>CrossCancerJSON</servlet-name>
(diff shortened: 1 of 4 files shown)


portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -672,8 +668,9 @@
         <url-pattern>/results/*</url-pattern>
         <url-pattern>/patient/*</url-pattern>
         <url-pattern>/newstudy/*</url-pattern>
-        <url-pattern>/mutation_mapper/*</url-pattern>
+        <url-pattern>/mutation_mapper/*</url-pattern>n
```
adamabeshouse: this seems like a typo

web/src/main/java/org/mskcc/cbio/portal/web/ProxyController.java
```diff
@@ -65,7 +65,7 @@
   public void setHotspotsURL(String property) { this.hotspotsURL = property; }
   
   private String bitlyURL;
-  @Value("${bitly.url}")
+  @Value("${bitly.url:''}")
```
adamabeshouse: why is this still here if you're getting rid of the property?

alisman: I would rather get leave things be and then retire it all en-masse

alisman: i'm not quite taking it out.    i want to remove it from EXAMPLE property file because it actually is defunct and does nothing.  but the property needs to have a default value or the above code will cause compilation error.  so i think this is the lightest touch.

adamabeshouse: ah okay

PR #5225, opened 2018-11-10
Data Access Tokens properties update and endpoint migration to internal API

angelicaochoa:
* Added new `portal.properties` property (`dat.revoke_other_tokens`)
* Fall back on default value for `dat.revoke_other_tokens` if user does not provide parameter to `dataAccesstoken` endpoint
* Migrated DAT endpoints from `/api-legacy` to main API

TO-DO:
- [ ] document feature and properties
- [ ] update e2e screenshot tests
- [x] support DAT for other authentication methods
- [ ] SAML testing

**NOTE: Tests will keep failing until we support other implementations of data access tokens.** 


angelicaochoa:
@sheridancbio right, this situation arose because how would the frontend "Download Token" button know whether to revoke other tokens or not since the user isn't interact with the endpoint directly.. It's just a function that the download link does by itself when it is clicked. The default value should be provided in some way and of course I don't want to hard code it in the frontend for this download token link is implemented. Instead it's best to make this some property that someone deploying a portal can set based on their needs.

angelicaochoa:
@sheridancbio I updated this PR to migrate the DAT endpoints to the main API.

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -88,7 +79,13 @@ private void initializeUsersWhoCannotUseTokenSet() {
                                     @RequestParam(required = false) Boolean allowRevocationOfOtherTokens) throws HttpClientErrorException {
```
ersinciftci: Can we use `-` separated paths in all endpoints to make it consistent with all other APIs? And also can we use plural for all resource names in the path, like `data-access-tokens`.

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -88,7 +79,13 @@ private void initializeUsersWhoCannotUseTokenSet() {
...
         HttpHeaders responseHeaders = new HttpHeaders();
         responseHeaders.setContentType(MediaType.APPLICATION_JSON);
```
ersinciftci: I think this happens automatically when your controller is annotated with `@RestController`, so you can delete it from all endpoints.

ersinciftci:
You also need to move `DataAccessTokenControllerTest.java` under web module too, and you probably don't need `DataAccessTokenControllerConfig.java`. You can take a look at the other controller tests under web module to create similar tests.

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
angelicaochoa: @ersinciftci done

angelicaochoa:
@ersinciftci thanks for the review! I updated the PR with your suggestions:

- updated DAT endpoint paths `-` separated to match rest of APIs
- moved exception handlers to `GlobalExceptionHandler`
- removed redundant response headers

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -0,0 +1,123 @@
...
+        else {
+            createdToken = dataAccessTokenService.createDataAccessToken(getAuthenticatedUser(authentication));
+        }
+        if (createdToken == null) {
+            return new ResponseEntity<>(new DataAccessToken(null), HttpStatus.NOT_FOUND);
+        }
+        return new ResponseEntity<>(createdToken, HttpStatus.CREATED);
+    }
+
+    @RequestMapping(method = RequestMethod.GET, value = "/data-access-tokens/fetch")
```
ersinciftci: You don't actually need `/fetch` for `GET` requests, it's for POST requests that sends a query object to get a result.

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @RequestMapping(method = RequestMethod.GET, value = "/data-access-tokens/{token}")
+    public ResponseEntity<DataAccessToken> getDataAccessToken(
+            @ApiParam(required = true, value = "token") @PathVariable String token) {
+        DataAccessToken dataAccessToken = dataAccessTokenService.getDataAccessTokenInfo(token);
+        return new ResponseEntity<>(dataAccessToken, HttpStatus.OK);
+    }
+
+    @RequestMapping(method = RequestMethod.DELETE, value = "/data-access-tokens/revoke")
```
ersinciftci: You don't need `/revoke` either, those verbs at the end of the paths is for POST requests that does querying (when it's a complex or long query that can't be done with GET parameters), to distinguish them from the POST requests that creates/updates resources.

security/security-spring/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -30,16 +30,17 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-package org.cbioportal.weblegacy;
+package org.cbioportal.web;
```
ersinciftci: I meant that it can be moved to the `web` module instead of `security`, since `DataAccessTokenController.java` is in the `web` module. And you can annotate this class with `@Configuration` and move the beans from `DataAccessTokenControllerConfig.java` to here, and delete the config file. You can look at the controller tests under `web` module for some examples.

angelicaochoa:
@ersinciftci resolved all comments

ersinciftci:
@angelicaochoa seems like some tests are failing

angelicaochoa:
@ersinciftci yeah there are some tests that are going to fail until we support other authentication profiles. Right now it'll only work for google plus which is why we're working in a dev branch

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -30,69 +30,73 @@
...
+import org.springframework.security.core.userdetails.User;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 import org.springframework.test.context.web.WebAppConfiguration;
 
 @RunWith(SpringJUnit4ClassRunner.class)
-//@WebAppConfiguration
-@ContextConfiguration(classes = {DataAccessTokenControllerConfig.class})
+@WebAppConfiguration
+@ContextConfiguration("/applicationContext-web.xml")
```
sheridancbio: Ok, I think I see. We are moving the test into the web module, so we don't need to do package-scan to find the controller beans. I worry a little about bringing in the whole runtime applicationContext-web file ... potentially changes to this file in the future will break our unit tests. But maybe that is good (that problems show up in the unit tests before we commit changes to applicationContext-web)

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -30,69 +30,73 @@
...
+@Configuration
 public class DataAccessTokenControllerTest {
+     @Autowired
+     DataAccessTokenService tokenService;
```
sheridancbio: why drop the "private" access modifier? Do we access this member externally?

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -30,69 +30,73 @@
...
 
-    @Autowired
-    private DataAccessTokenService tokenService;
+     @Autowired
+     DataAccessTokenController dataAccessTokenController;
```
sheridancbio: same comment

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -30,69 +30,73 @@
...
+
     @Before
     public void setup() {
         Mockito.reset(tokenService);
     }
-   
+
     protected Authentication makePrincipal(String username, boolean isAuthentic) {
-        PortalUserDetails portalUserDetails = makePortalUserDetails(username);
+        User user = new User(username, "unused", new ArrayList<GrantedAuthority>());
```
sheridancbio: this avoids bringing in dependencies from the security module (because we don't need PortalUserDetails)

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -120,10 +124,6 @@ private void checkIfResponseMatches(ResponseEntity<DataAccessToken> response, Ht
         if (responseStatus != expectedStatus) {
             Assert.fail("Response from controller handler (" + responseStatus + ") did not match expected status : " + expectedStatus);
         }
-        MediaType responseType = response.getHeaders().getContentType();
```
sheridancbio: Why was this part of the test dropped? The response type should be JSON (unless the request fails due to authentication issues)

ersinciftci:
> @ersinciftci yeah there are some tests that are going to fail until we support other authentication profiles. Right now it'll only work for google plus which is why we're working in a dev branch
> 
> I think @sheridancbio just started working on this actually

@angelicaochoa failing end-to-end tests are sometimes okay but some unit tests are also failing, and since they are supposed to be isolated from other until, they should ideally always pass (I think currently there is a Spring bean dependency related problem, lots of unit tests are failing).

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
angelicaochoa: correct

angelicaochoa:
@ersinciftci right, there was a build issue due to some beans related to the DAT authentication provider / user detail service when the authentication profile is not `googleplus`. We have a fix in place that makes these beans available to any kind of authentication profile now. We plan on refining further as we move forward with development. I was able to successfully run the tests locally with @sheridancbio and @averyniceday after adding a couple missing drop table statements in `cgds.sql` though. I'll be sure to keep you in the loop in the next few days as we move forward with adding more unit tests and additional refactoring.

PR #5245, opened 2018-11-14
Resolve issue #5242

### LINKED ISSUE #5242: cgds and migration.sql mismatch
see https://github.com/cBioPortal/cbioportal/pull/3012/files#r233434163

angelicaochoa:
# What? Why?
Fix #5242 

- Resolve mismatch in cgds.sql and migration.sql varchar size for `mutation_count_by_keyword.KEYWORD`

# Checks
- [ ] Runs on Heroku.

### DIFF
--- a/db-scripts/src/main/resources/cgds.sql
+++ b/db-scripts/src/main/resources/cgds.sql
@@ -475,7 +475,7 @@ CREATE TABLE `mutation` (
 -- --------------------------------------------------------
 CREATE TABLE `mutation_count_by_keyword` (
     `GENETIC_PROFILE_ID` int(11) NOT NULL,
-    `KEYWORD` varchar(50) DEFAULT NULL,
+    `KEYWORD` varchar(255) DEFAULT NULL,
     `ENTREZ_GENE_ID` int(11) NOT NULL,
     `KEYWORD_COUNT` int NOT NULL,
     `GENE_COUNT` int NOT NULL,


angelicaochoa:
🔔  @pieterlukasse

sheridancbio:
I have confirmed .. an alter table statement which keeps the same data definition for a column does not cause an error. Such as : alter table `info` modify column `DB_SCHEMA_VERSION` varchar(24) default null;

inodb:
@sheridancbio @angelicaochoa  we should prolly still create a new migration for this right? Maybe we can add it as part of the migration currently in `release-2.1.0` branch?

angelicaochoa:
@inodb the mutation piece is already prsent

inodb:
@angelicaochoa isn't the issue @sheridancbio mentioned unresolved? If somebody recently installed a seed cgds.sql from e.g. `v2.6.1`, they would never have this varchar increase?

If that's indeed the case it might be good to duplicate the same migration as part of `v2.7.4` (which is currently in `release-2.1.0` branch). Basically add the statement ```ALTER TABLE `mutation_count_by_keyword` MODIFY COLUMN `KEYWORD` VARCHAR(255);``` to the `v2.7.4` migration.

angelicaochoa:
@inodb ah, I see. Good point. Think I should just move the original migration statement to the appropriate place then or duplicate it?

inodb:
@angelicaochoa duplicating it should be fine, maybe with a comment in the migration.sql referring to this PR

PR #5262, opened 2018-11-16
Change db parameters dbcp for production

inodb:
We want to use dbcp on k8s/AWS. This copies over the db parameters we are currently
using in production in our context.xml when using jndi instead of dbcp


fedde-s:
I haven't tested this, but from a brief glance it looks like it will align the way cBioPortal runs across different institutes, which would generally be good if there's no specific reason to differ 👍

fedde-s:
Although I feel that the comment sounds like it should be in a commit message, since it refers to the history of the code instead of the reasoning behind it.

PR #5264, opened 2018-11-16
Make DAT implementation configurable via portal.properties

angelicaochoa:
# What? Why?
Make DAT implementation configurable via portal.properties

* DAT implementation method used is now configurable via `/portal.properties:dat.method`
* Supported values for `dat.method` are `uuid`,`jwt`, or `none` (default)
* Throw `AccessDeniedException` is user tries to interact with DAT endpoints on an unauthenticated portal.
* DATController Tests use TestPropertySource instead of importing ContextConfiguration from /applicationContext-web.xml
* Fixed failing tests in `TokenAuthenticationFilterTest` and `JwtUtilsTest` by setting default value for properties in respective beans


service/src/main/java/org/cbioportal/service/DataAccessTokenServiceFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package org.cbioportal.service;
+
+/**
+ *
+ * @author ochoaa
+ */
+public interface DataAccessTokenServiceFactory {
```
angelicaochoa: this interface is used to get the `DataAccessTokenService` by bean/component name (DAT method implementation value from `dat.method` in `portal.properties`)

It is part of the `ServiceLocatorFactoryBean` in `applicationContext-security.xml`
```
    <b:bean  id="tokenServiceFactory" class="org.springframework.beans.factory.config.ServiceLocatorFactoryBean">
        <b:property name="serviceLocatorInterface" value="org.cbioportal.service.DataAccessTokenServiceFactory"/>
    </b:bean>
```

sheridancbio:
There is one failing unit test currently in the security module:
org.springframework.security.authentication.BadCredentialsException: Invalid token
	at org.cbioportal.security.spring.authentication.token.TokenAuthenticationFilter.attemptAuthentication(TokenAuthenticationFilter.java:102)
	at org.cbioportal.security.spring.authentication.token.TokenAuthenticationFilterTest.testAttemptAuthentication_success(TokenAuthenticationFilterTest.java:90)

Perhaps these older unit tests in the filter were expecting JWT token signature sections in the provided tokens? Ideally the filter only calls the (correct) service layer's .validateToken() method, which is where knowledge about how to validate tokens is encapsulated.

sheridancbio:
another failing "sanity check" is that the db_schema_version numbers need to migrate together. migration.sql has moved to 2.7.4, while cgds.sql and the portal pom are at 2.73:

Making sure all db versions are the same in cgds.sql, pom.xml and migration.sql
pom.xml db version is 2.7.3
db-scripts/src/main/resources/cgds.sql db version is 2.7.3
db-scripts/src/main/resources/migration.sql db version is 2.7.4

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -53,32 +56,42 @@
...
-    private String[] USERS_WHO_CANNOT_USE_TOKENS = {"anonymousUser", "servcbioportal"};
-    private Set<String> usersWhoCannotUseTokenSet = null;
+    @Value("${dat.unauth_users:anonymousUser}")
+    private String[] USERS_WHO_CANNOT_USE_TOKENS;
+    private List<String> usersWhoCannotUseTokenSet;
 
     @Autowired
     private void initializeUsersWhoCannotUseTokenSet() {
-        usersWhoCannotUseTokenSet = new HashSet<>(Arrays.asList(USERS_WHO_CANNOT_USE_TOKENS));
+        usersWhoCannotUseTokenSet = Arrays.asList(USERS_WHO_CANNOT_USE_TOKENS);
```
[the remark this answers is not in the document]
angelicaochoa: @sheridancbio resolved

service/src/test/java/org/cbioportal/service/util/JwtUtilsTestConfiguration.java
```diff
@@ -49,33 +49,8 @@
 package org.cbioportal.service.util;
 //TODO package org.cbioportal.security.spring.authentication.token;
 
-import io.jsonwebtoken.Claims;
```
sheridancbio: thanks for the copious cleanup

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -119,6 +119,8 @@ ldap.attributes.given_name=givenName
 ldap.attributes.email=mail
 
 # data access token settings
+dat.unauth_users=
```
sheridancbio: I think the property name is not clear. A self-documenting name such as "dat.ineligible_user_list" would be better, or we could give a comment above the property explaining what goes into the field : (comma/whitespace?) separated list  of users who will be refused access to the Data Access Token methods

angelicaochoa: @sheridancbio I added some detailed descriptions of all of these properties in the documentation PR (#5306)

PR #5306, opened 2018-11-20
Data Access Token documentation

angelicaochoa:
# What? Why?
Added documentation for Data Access Token:
- obtaining and usage
- implementation 

Co-authored-by: Benjamin Gross <benjamin.gross@gmail.com>
Co-authored-by: Robert Sheridan <sheridan@cbio.mskcc.org>


docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
+# Downloading Controlled-Access Data
+
+## [Introduction](introduction)
+A user authentication token (data access token) is required for downloading controlled-access data programmatically from the cBioPortal. See [Obtaining an Authentication Token](#obtaining-an-authentication-token) for instructions on how to download a token and [Using Authentication Tokens](#using-authentication-tokens) for how to use it.
+
```
sheridancbio: Other pages in the documentation are in the style of a guild/tutorial to the cBioPortal deployer, rather than the user. They step the reader through the steps of how to configure and deploy the portal, so I think this page should try to take the same approach. I think the first section would guide the user on the choices available in portal.properties and give advice on how to select a token format, and then tell them what other settings are available. This could be organized property by property, or having a subsection for UUID configuration, a subsection for JWT configuration, and a subsection for the "none" mode - disallowing DAT in this portal instance.

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+## [Obtaining an Authentication Token](obtaining-an-authentication-token)
+To obtain an authentication token, simply select the `Download Token` option from the dropdown menu at the top right corner of the cBioPortal home page (where your username is displayed). Your token will be downloaded as file with the following format:
```
sheridancbio: An image would be valuable here. It might be helpful to also mention that when token authentication is disabled (or for users who are prohibited from access) the menu option will not be present (a contrasting image).

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+token: your-authentication-token
```
sheridancbio: Rather than showing a template of a token, an actual generated token would be better:
An example of a UUID formatted token:
[file contents]
An example of a JWT formatted token:
[file contents]

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+creation_date: YYYY-MM-DD HH:MM:SS
+expiration_date: YYYY-MM-DD HH:MM:SS
+```
+
+_Note that the expiration date may vary between different controlled-access instances of the cBioPortal._
```
sheridancbio: I don't understand what this means. Maybe you are saying that the duration of a token can be configured on different portals. If that is the meaning, it can be covered in the section on how to configure the access tokens in portal.properties.

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+
+## [Using Authentication Tokens](using-authentication-tokens)
+Once downloaded, the token can be sent in the `Authorization` header when making requests to the [cBioPortal web API](cBioPortal-Web-API.md).
+```
+Authorization: Bearer <token>
+```
```
sheridancbio: Maybe attach a curl command showing a request to GET /api/studies where the header is attached (using a sample token from above)

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+
+## [Implementing Controlled-Access Tokens](implementing-controlled-access-tokens)
+To implement controlled-access tokens for your cBioPortal instance:
```
sheridancbio: my previous comments suggested moving this section up sooner in this documentation page.
Instead of the term "implementing" I would refer to "configuring".
Instead of "Controlled-Access Tokens" I would stick with "Authentication Tokens" or "Data Access Tokens". We should perhaps discuss what our terminology will be for this feature and then apply it consistently throughout the code and the documentation.

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+
+1. Decide on an implementation ([JWT](#jwt-implementation) or [UUID](#uuid-implementation)).
```
sheridancbio: I would give more guidance to the deployer for the decision between styles of tokens (and also point out that they can disable token use if that is their preference). I agree that this should be the first decision they make (whether or not to provide token authentication and if so, what kind of token format to use)

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## [Implementing Controlled-Access Tokens](implementing-controlled-access-tokens)
+To implement controlled-access tokens for your cBioPortal instance:
+
+1. Decide on an implementation ([JWT](#jwt-implementation) or [UUID](#uuid-implementation)). 
+2. Add or modify your `portal.properties` with the following:
+
+
+|           Property          |                                      Description                                          |    Default    |      Type Allowed     |
```
sheridancbio: this (table) format matches what is on Customizing-your-instance-of-cBioPortal.html

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+| --------------------------- | ----------------------------------------------------------------------------------------- | ------------- | --------------------- |
+| dat.unauth_users            | Users that should never be allowed to download an authentication token                    | anonymousUser | comma-delmited string |
+| dat.method                  | Authentication token implementation method                                                | none          | jwt, uuid, none       |
```
sheridancbio: instead of implementation method (a programming term), we should simply talk about token format (Authentication token format)

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+| dat.ttl_seconds             | Time in seconds until token expiration                                                    | -1            | integer               |
+| jwt.secret_key ($)          | Signature used to verify the token is valid and not modified.                             | none          | string                |
```
sheridancbio: I think it would be better to rename all properties related to data access tokens to have the prefix "dat". So we could  have dat.jwt_secret_key and dat.jwt_issuer (if the properties are currently jwt.secret_key and jwt.issuer then this would need to be a separate code update PR as well)

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+| jwt.issuer ($)              | Issuer of the token.                                                                      | none          | string                |
+| dat.revoke_other_tokens (*) | Determines whether existing tokens should be revoked if a new token is created for a user | true          | true or false         |
```
sheridancbio: Benjamin recommended dropping this configuration choice. I personally think it is worthwhile, but this description is imprecise. Even if this property is true, existing tokens will not be revoked so long as the user is below their token quota (dat.max_number_per_user). Only when they are at their quota will a (one) token that exists be expired to "make room" for the issuance of a new token. So if quota is 5 and a user has 5 existing tokens, then when they request a new token with the setting dat.revoke_other_tokens=true, one of the 5 existing tokens (the soonest to expire) will be revoked, the other 4 existing tokens will remain, and the new token will be issued.
I'm not sure how to phrase this succinctly, but the current description implies that all 5 existing tokens would be revoked.

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+| dat.max_number_per_user (*) | Maximum number of tokens allowed per user                                                 | -1            | integer               |
+
+- \* UUID-specific implementation
```
sheridancbio: I would say "Options with this footnote are only applicable when using the UUID token format"

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+- ($) JWT-specific implementation
```
sheridancbio: I would say "Options with this footnote are only applicable when using the JWT token format"

docs/Downloading-Controlled-Access-Data.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+### [JWT implementation](jwt-implementation)
+Detailed information on JSON Web Tokens (JWT) can be found [here](https://jwt.io/introduction/).
+
+### [UUID implementation](uuid-implementation)
+_Note: UUID implementation requires a backend database / token store but allows for greater control over tokens issued._
```
sheridancbio: A table in the cBioPortal database is used to store tokens when using the UUID token format. Revocation of a token is achieved by deleting the token from this table.
[I would also show the table schema here and give field descriptions]

docs/Downloading-Controlled-Access-Data.md
averyniceday: @sheridancbio I actually think it might better to keep this as implementation (or another word meaning the same thing). I think a user selecting between the two cares more about the management rather than the format the token comes back as.

docs/README.md
```diff
@@ -64,4 +65,4 @@ We also maintain an active [list of RFCs (Requests for Comments)](RFC-List.md) w
 
 ### 5.2 Datasets
 * [Downloads](Downloads.md)
-
+* [Downloading Controlled-Access Data](Downloading-Controlled-Access-Data.md)
```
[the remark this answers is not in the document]
n1zea144: Agreed.

docs/Authenticating-Users-via-Tokens.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Property**: dat.unauth_users (optional)
+- **Description**: A list of users that should not be allowed to download a data access token.
+- **Permissible Values**: A comma-delimited list of valid user names as found in the users table.
+- **Default Value**: anonymousUser
+
+**Property**: dat.ttl_seconds (required)
+- **Description**: The time in seconds between token creation and token expiration.
+- **Permissible Values**: An integer value greater than zero.
+- **Default Value**: -1
```
n1zea144: What happens when then the default value is -1?  I think we should set the default value to a 30 day expiration.

docs/Authenticating-Users-via-Tokens.md
```diff
@@ -0,0 +1,89 @@
+# Introduction
+
+The cBioPortal provides access to data within its database through the [cBioPortal web service](http://www.cbioportal.org/api/swagger-ui.html).  While most users interact with this web service indirectly while visiting cBioPortal web pages, it is possible to access this web service directly.  Direct access to the web service is appealing for those users with an interest in mining or analyzing cBioPortal data beyond the confines of a web browser, such as from ‘R’ or Python clients.
+
+If your instance of the cBioPortal will not require [user authorization](https://docs.cbioportal.org/2.2-authorization-and-authentication/user-authorization) (login), direct access to the web service is available without any additional server configuration or client side prerequisites.  However, if your instance of the cBioPortal will require login, additional **data access token** server configuration is necessary before direct access to the web service is available to your users. Additionally, prior to web service access, clients will need to obtain a data access token and present this token with each web service request.
```
n1zea144: I think we can move the bold around "data access token"

docs/Authenticating-Users-via-Tokens.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Property**: dat.jwt.secret_key (required only when dat.method = jwt)
+- **Description**: The signature used to sign JWT based data access tokens during creation.  This signature is used in subsequent web service requests to validate the request.  This property is used only when dat.method = jwt.
+- **Permissible Values**: A string value.
+- **Default Value**: none
+
+**Property**: dat.uuid.max_number_per_user (required only when dat.method = uuid)
+- **Description**: This property determines the number of data access tokens that can be allocated to a single user at any given time.  If a user requests a data access token that puts them over this limit, then based on the value of dat.uuid.revoke_other_tokens (see below), the cBioPortal will either refuse to present a new token to the user or it may invalidate the oldest one before presenting the new one to the user.  This property is used only when dat.method = uuid.
+- **Permissible Values**: An integer value greater than zero.
+- **Default Value**: -1
```
n1zea144: What happens with a default value of -1...unlimited number of tokens?  I think we should have the default value to 1 and set revoke_other_tokens default to true.

PR #5319, opened 2018-11-21
fix api parameter description in 2 swagger endpoints

sheridancbio:
# What? Why?
Swagger API Documentation for gene-panel-data/fetch endpoint and mutations/fetch contain incorrect descriptions

fix api parameter description in gene-panel-data and mutations endpoint
    - gene-panel-data/fetch description refers to request supplying Gene IDs, but no such ability remains in the service interface
    - mutations/fetch description does not mention mode where sampleIds are not supplied (only molecularProfileId)

# Checks
- [ ] Runs on Heroku.


ersinciftci:
Thanks @sheridancbio !

sheridancbio:
Possibly an update to the behavior of the mutation/fetch endpoint spec was accidentally applied to the gene-panel-data/fetch endpoint

Issue #5214, opened 2018-11-08
bad links in documentation


tmazor:
The "Import Gene Panels" page has links to other pages within the documentation, but those links are wrong.

Go here: https://cbioportal.readthedocs.io/en/latest/Import-Gene-Panels.html
Click on "File Format" under #2.
It brings you to: https://cbioportal.readthedocs.io/File-Formats.html#gene-panel-data
But the correct link is: https://cbioportal.readthedocs.io/en/latest/File-Formats.html#gene-panel-data

"Loading a Sample Study" under #3 has the same issue.

inodb:
Seems to be a known issue with the docs parser: https://github.com/rtfd/recommonmark/issues/8 . Looks like someone wrote a hacky fix, might be worth trying.

inodb:
moved to gitbook instead

PR #5337, opened 2018-11-27
Add files for config of gitbook.com

inodb:
This addresses this issue by moving to a different service, gitbook.com instead of readthedocs.org (suggested by @sandertan ): 

https://github.com/cBioPortal/cbioportal/issues/5214

Overall seems the MarkDown integration is better, see for an example:

https://cbioportal.gitbook.io/cbioportal/v/gitbook/

Also it adds an integration in the github ci, so you can easily few the documentation live (make sure to select the `gitbook` version):

screen shot 2018-11-27 at 2 59 58 pm


tmazor:
gitbook looks nice!

sandertan:
@inodb Nice! So when this is merged, the version will be 'master'? 

And is the plan to switch from readthedocs to gitbook completely? There are a few references to readthedocs in our current code/documentation.

Also, I was wondering what this version 1.0.0 is:
screenshot 2018-11-28 at 13 31 24

inodb:
@sandertan 	that version will disappear once merged to master (1.0.0 refers to `master` branch for whatever reason). Use `gitbook` version for now: cbioportal.gitbook.io/cbioportal/v/gitbook

I think we can run them side by side after merging this. If we don't find issues with gitbook we can make that the official one.

I found one minor thing:

- [ ] When a link is referring to a file outside SUMMARY.md it will point to the file on GitHub instead of showing it inside gitbook. e.g. https://cbioportal.gitbook.io/cbioportal/v/gitbook/5.1-data-loading/data-loading/file-formats#gene-panel-data click on the link `Import-Gene-Panels` will take you to GitHub. Might be good to reorganize those files such that they are part of SUMMARY.md

inodb:
@sandertan made some changes to the docs:

- Add missing markdown to SUMMARY/README
- Update API docs
- Remove PFAM update guide (this comes from Genome Nexus now)

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -55,11 +55,9 @@ SELECT count(*) FROM cbioportal.gene;
 SELECT count(*) FROM cbioportal.gene_alias;
 ```
 
-8- Additionally, there are several other tables you may want to update now (only in human).
+8- Additionally, there are other tables you may want to update now (only in human).
 
-1. Updating the pfam graphics if you want to see 3D structures, explained in [this document](Updating-pfam_graphics-table.md)
```
inodb: this is no longer required, since all this data lives in Genome Nexus now as far as I know

fedde-s:
I see no particular problems introduced by this, except that changing the contents of the docs (such as whether to update the PFAM graphics table) does not belong in a commit that's about configuring Gitbook integration. That only makes review and later maintenance/referencing harder.

I hope GitBook will support linking to documentation for previous releases in addition to `master`. If not, we'll still have GitHub to fall back on for that use-case, so that should be no problem.

A few pages still use GitHub-specific emoji sequences, such as `:information_source:` in [the Keycloak instructions](https://docs.cbioportal.org/v/gitbook/2.2-authorization-and-authentication/authenticating-and-authorizing-users-via-keycloak#configure-keycloak-to-authenticate-your-cbioportal-instance), but that issue is not introduced by this PR, it was already an issue on Readthedocs.

inodb:
@fedde-s good point - I'll update the title of that commit. The missing Markdown files in the ToC is not necessarily gitbook specific

There's a plug-in to generate docs for different tags. Unf u can only either show specific branches or tags, not both afaik. We should decide which ones to show

PR #5344, opened 2018-11-28
Break security module dependency on web module

sheridancbio:
# What? Why?
Web module code unit tests were located in security module because of the need to avoid a cyclic dependency between modules. Breaking this dependency allows unit tests to be located in the module containing the tested code and encapsulates the user-supplied argument types for the web layer within the web layer.

## module dependency change
Current :
`    security <- web <- service <- persistence`
After this PR : 
`    web <- service <- persistence`
`                security <- persistence`

The essential change in this PR is to remove knowledge of incoming JSON parameters to POST Web API requests from the security module. The security module no longer parses user supplied arguments. Instead those arguments are parsed in an intercepter class defined in the web module which stores a list of involved cancer studies for the request. Then the security module (through CancerStudyPermissionEvaluator) determines if the user making the request has authorization to view the involved cancer studies. So the security module now works only using the core object model defined in the model module.

A utility has been added to the persistence/persistence-mybatis module to aid in mapping ids to modeled objects such as {CancerStudy, MolecularProfile, SampleList}. This utility is used both by the new interceptor class (InvolvedCancerStudyExtractorInterceptor) and CancerStudyPermissionEvaluator for efficient spring-method-security PreAuthorize() and PostFilter() processing.

Other bugfixes and code cleanup are included (see changeset log message for full details)

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


web/src/main/java/org/cbioportal/web/util/RequestWrapper.java
```diff
@@ -0,0 +1,47 @@
+package org.cbioportal.web.util;
```
angelicaochoa: @sheridancbio this isn't used right?

sheridancbio: Maybe not .. maybe this is left in from earlier attempts at this?

sheridancbio: file removed and tests still pass

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -72,9 +72,6 @@
     private static String authenticate;
     @Value("${authenticate:false}") // default is false
     public void setAuthenticate(String property) { authenticate = property; }
-    private static String authorization;
-    @Value("${authorization:false}") // default is false
-    public void setAuthorization(String property) { authorization = property; }
```
pieterlukasse: nicely removed property in documentation :+1:

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -129,13 +131,17 @@
...
 
-    @PreAuthorize("hasPermission(#sampleFilter, 'SampleFilter', 'read')")
+    @PreAuthorize("hasPermission(#involvedCancerStudies, 'Collection<CancerStudyId>', 'read')")
     @RequestMapping(value = "/samples/fetch", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE,
         produces = MediaType.APPLICATION_JSON_VALUE)
     @ApiOperation("Fetch samples by ID")
     public ResponseEntity<List<Sample>> fetchSamples(
+        @ApiIgnore // prevent reference to this attribute in the swagger-ui interface
+        @RequestAttribute(required = false, value = "involvedCancerStudies") Collection<String> involvedCancerStudies,
+        @ApiIgnore // prevent reference to this attribute in the swagger-ui interface. this attribute is needed for the @PreAuthorize tag above.
```
pieterlukasse: `interceptedSampleFilter` does not seem to be used in the `@PreAuthorize` tag above.

sheridancbio: @pieterlukasse That is correct. interceptedSampleFilter is the unmodified / uninterpreted / verbatim argument as passed by the user. This JSON object is parsed / used in the handler function business logic that follows. The '@PreAuthorize' annotation receives an extracted set of cancer studies that contain any of these samples. This extraction is done by InvolvedCancerStudyExtractorInterceptor ... returning model module objects (CancerStudy) which can be checked by the security module without passing the SampleFilter object (defined in the web module). That is the essence of this PR - the web module concepts are hidden from the security module, and instead we are passing model module objects (the model module is a shared dependency to all other modules)

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -94,187 +77,22 @@
...
         if (log.isDebugEnabled()) {
             log.debug("setPublicCancerStudiesGroup(), always_show_study_group = " + ((property == null) ? "null" : property));
         }
         if (property != null && property.trim().isEmpty()) {
             PUBLIC_CANCER_STUDIES_GROUP = null;
-        } 
-    }
-
-    // maps used to cache required relationships - in all maps stable ids are key
-    private Map<String, MolecularProfile> molecularProfiles = new HashMap();
```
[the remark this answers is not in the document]
pieterlukasse: @angelicaochoa yes, I am referring to memory issues that I detected during load tests on an authenticated portal. So perhaps these could be a cause...

PR #5354, opened 2018-12-03
Add endpoint to calculate coexpression in gene sets

oplantalech:
This PR adds a new endpoint which allows to calculate the correlations between one gene (or gene set) from a specific profile (e.g. mRNA expression) against the genes / gene sets of another profile (mRNA expression, RPPA, GSVA scores). This endpoint is going to be used in the new co-expression tab, you can check (and find out where to test) the functionality here: https://github.com/cBioPortal/cbioportal-frontend/pull/1786.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
@@ -4,21 +4,29 @@ import org.apache.commons.lang.math.NumberUtils;
 import org.apache.commons.math3.linear.Array2DRowRealMatrix;
 import org.apache.commons.math3.linear.RealMatrix;
 import org.apache.commons.math3.stat.correlation.SpearmansCorrelation;
-import org.cbioportal.model.CoExpression;
 import org.cbioportal.model.Gene;
 import org.cbioportal.model.GeneMolecularData;
-import org.cbioportal.service.CoExpressionService;
+import org.cbioportal.model.Geneset;
+import org.cbioportal.model.GenesetMolecularData;
+import org.cbioportal.model.MolecularData;
+import org.cbioportal.model.MolecularProfile;
+import org.cbioportal.model.CoExpression.GeneticEntityType;
+import org.cbioportal.model.CoExpression;
 import org.cbioportal.service.GeneService;
+import org.cbioportal.service.GenesetDataService;
+import org.cbioportal.service.GenesetService;
 import org.cbioportal.service.MolecularDataService;
-import org.cbioportal.service.exception.MolecularProfileNotFoundException;
+import org.cbioportal.service.MolecularProfileService;
+import org.cbioportal.service.CoExpressionService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
-import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 @Service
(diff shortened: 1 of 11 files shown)


model/src/main/java/org/cbioportal/model/GenesetCoExpression.java
```diff
@@ -0,0 +1,59 @@
+package org.cbioportal.model;
```
ersinciftci: Seems like this file is not used anywhere?

oplantalech: @ersinciftci Looks like this package needs to be declared.

service/src/main/java/org/cbioportal/service/impl/MolecularProfileCorrelationServiceImpl.java
```diff
@@ -0,0 +1,589 @@
+package org.cbioportal.service.impl;
```
ersinciftci: @oplantalech @pieterlukasse When I look at this class, I see a lot of very similar code blocks repeating over and over again. There must be a better way to do this without so many repetition.

oplantalech: @ersinciftci I've been working to remove the repetitive code. I think that looks much better now 😃

ersinciftci:
It seems like `MolecularProfileCorrelationServiceImpl.java` is mostly copy paste of `CoExpressionServiceImpl.java`, and `MolecularProfileCorrelation.java` is mostly copy paste of `CoExpression.java`, `fetchGenesetMolecularData` in `MolecularDataServiceImpl.java` is mostly copy paste of `fetchMolecularData`, and `MolecularProfileCorrelationController` is mostly copy paste of `CoExpressionService.java` (as a matter of fact, the function's name is still `fetchCoExpressions`). So this PR basically replicates the functionality of gene co-expressions for genesets by duplicating the code. Considering both genes and genesets have the concept of `GENETIC_ENTITY_ID`, I think it should be done utilizing that field instead of code duplication.

oplantalech:
@ersinciftci Yes, the new "correlation" endpoint does everything what the "co-expression" endpoint did for genes, with also the possibility of changing the profile (e.g. calculate correlation of SOX9 in mRNA expression against all genesets / all RPPA values in the study), and it also provides the same functionality for gene sets. The "old" endpoint should be removed, I did not remove it here because I was afraid that this could break the current functionality in the frontend. Now I've removed the files associated with the old endpoint, so we will have to merge together this PR and the frontend one, which only uses the new endpoint from this PR (https://github.com/cBioPortal/cbioportal-frontend/pull/1786).

ersinciftci:
@oplantalech I see, I wish I had known it before reviewing :) The tab's name will still be Co-expression in the frontend, right? Is there any reason to change the endpoint's name to Correlation in the backend? Can we just keep the old names and add geneset and profile change support? I prefer using the same terms in the frontend and backend so it's easier to find what component in the frontend uses what API endpoint.

web/src/main/java/org/cbioportal/web/MolecularProfileCorrelationController.java
```diff
@@ -0,0 +1,67 @@
...
+@RestController
+@Validated
+@Api(tags = "Correlations", description = " ")
+public class MolecularProfileCorrelationController {
+    
+    @Autowired
+    private MolecularProfileCorrelationService molecularProfileCorrelationService;
+
+    @PreAuthorize("hasPermission(#molecularProfileId, 'MolecularProfile', 'read')")
+    @RequestMapping(value = "/molecular-profiles/correlations/{geneticEntityId}/{molecularProfileIdA}/{molecularProfileIdB}/fetch",
```
ersinciftci: We use the path variables only for `is-a` relationships, so all these 3 variables should be query parameters. And the name `geneticEntityId` is confusing because in the database it's the internal auto-increment id but in here we are using it as Entrez Gene Id or Geneset Id. So I suggest you create 2 separate query parameters, entrezGeneId and genesetId, and throw an exception if both of them are non-null.

service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
```diff
@@ -222,4 +224,64 @@ private void validateMolecularProfile(String molecularProfileId) throws Molecula
...
+        validateMolecularProfile(molecularProfileId);
+        List<String> sampleIds = sampleListRepository.getAllSampleIdsInSampleList(sampleListId);
+        if (sampleIds.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return fetchGenesetMolecularData(molecularProfileId, sampleIds, genesetIds, projection);
+    }
+    
+    @Override
+    public List<GenesetMolecularData> fetchGenesetMolecularData(String molecularProfileId, List<String> sampleIds,
```
ersinciftci: This is still pretty much copy paste of `fetchMolecularData` except one line. Instead of doing that, we should change `molecularDataRepository.getGeneMolecularAlterations` to accept both genes and genesets and call the appropriate mapper method based on the genetic entity type.

web/src/main/java/org/cbioportal/web/CoExpressionController.java
```diff
@@ -34,27 +33,40 @@
...
         method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE,
         produces = MediaType.APPLICATION_JSON_VALUE)
-    @ApiOperation("Fetch co-expressions in a molecular profile")
+    @ApiOperation("Calculates correlations between a genetic entity from a specific profile and another profile from the same study")
     public ResponseEntity<List<CoExpression>> fetchCoExpressions(
-        @ApiParam(required = true, value = "Molecular Profile ID e.g. acc_tcga_rna_seq_v2_mrna")
-        @PathVariable String molecularProfileId,
-        @ApiParam(required = true, value = "List of Sample IDs/Sample List ID")
+        @ApiParam(required = true, value = "Molecular Profile ID from the Genetic Entity referenced in the co-expression filter e.g. acc_tcga_rna_seq_v2_mrna")
+        @PathVariable String molecularProfileIdA,
```
ersinciftci: As I said before, these 2 should be @RequestParam because multiple resources back to back in the path would indicate that they have a parent-child relationship, but it's not the case here. So the path should be `/molecular-profiles/co-expressions/fetch`

web/src/main/java/org/cbioportal/web/CoExpressionController.java
```diff
@@ -34,27 +33,40 @@
...
         @Valid @RequestBody CoExpressionFilter coExpressionFilter,
-        @ApiParam(required = true, value = "Entrez Gene ID")
-        @RequestParam Integer entrezGeneId,
         @ApiParam("Threshold")
-        @RequestParam(defaultValue = "0.3") Double threshold) throws MolecularProfileNotFoundException {
+        @RequestParam(defaultValue = "0.3") Double threshold) throws Exception {
 
         List<CoExpression> coExpressionList;
+        String geneticEntityId = null;
+        String geneticEntityType = null;
```
ersinciftci: Please use enum for these kind of variables, with values `GENE` and `GENESET`.

service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
oplantalech: @ersinciftci My fault, there was already an available endpoint to fetch molecular data for gene sets.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -28,71 +33,125 @@
...
-        return createCoExpressions(molecularDataList, entrezGeneId, threshold);
+        List<CoExpression> computedCoExpressions = null;
+        List<? extends MolecularData>  molecularDataListA = null;
+        List<? extends MolecularData>  molecularDataListB = null;
+        if (geneticEntityType.toLowerCase().equals("gene")) {
+            molecularDataListA = molecularDataService.getMolecularData(molecularProfileIdA, sampleListId, null, "SUMMARY");
+        } else if (geneticEntityType.toLowerCase().equals("geneset")) {
+            molecularDataListA = genesetDataService.fetchGenesetData(molecularProfileIdA, sampleListId, null);
+        } else {
+                throw new Exception("Genetic Entity Type must be 'gene' or 'geneset'.");
```
ersinciftci: This PR creates a GeneticEntityType enum in the web module, you can move it inside `CoExpression.java` and use it everywhere, so you wouldn't need this string comparison and exception handling.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -28,71 +33,125 @@
...
+        }
+        if (molecularProfileIdB.matches("(.*)_gsva_scores")) {
```
ersinciftci: You should get the geneticAlterationType of this profile and check if it's equal to `MolecularAlterationType.GENESET_SCORE`, instead of checking the name of the profile.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -28,71 +33,125 @@
...
+
+        List<CoExpression> computedCoExpressions = null;
+        List<? extends MolecularData>  molecularDataListA = null;
+        List<? extends MolecularData>  molecularDataListB = null;
+        if (geneticEntityType.toLowerCase().equals("gene")) {
+            molecularDataListA = molecularDataService.fetchMolecularData(molecularProfileIdA, sampleIds, null, "SUMMARY");
+        } else if (geneticEntityType.toLowerCase().equals("geneset")) {
+            molecularDataListA = genesetDataService.fetchGenesetData(molecularProfileIdB, sampleIds, null);
+        } else {
+                throw new Exception("Genetic Entity Type must be 'gene' or 'geneset'.");
```
ersinciftci: Same problem with above.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -28,71 +33,125 @@
...
+        }
+        if (molecularProfileIdB.matches("(.*)_gsva_scores")) {
```
ersinciftci: Same problem with above.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -28,71 +33,125 @@
...
             }
             
             CoExpression coExpression = new CoExpression();
-            coExpression.setEntrezGeneId(entrezGeneId);
-            Gene gene = genes.get(entrezGeneId).get(0);
-            coExpression.setCytoband(gene.getCytoband());
-            coExpression.setHugoGeneSymbol(gene.getHugoGeneSymbol());
+            coExpression.setGeneticEntityId(entityId);
+            //check if the query gene is a gene or a gene set:
+            if (molecularDataMapB.get(entityId).get(0) instanceof GeneMolecularData) {
```
ersinciftci: We already know if molecularProfileIdB is a geneset profile or gene profile from the line 87, I don't think you need to check again.

oplantalech:
@jjgao Can we merge this PR? You can test it in https://cbioportal-test.thehyve.net:8081/cbioportal/api/swagger-ui.html#/Co-Expressions/fetchCoExpressionsUsingPOST

jjgao:
@oplantalech yes, we can since @ersinciftci has approved this pr. Could you rebase?

jjgao:
@oplantalech would this break the current frontend?

oplantalech:
@jjgao Yes, we need to synchronize this with cBioPortal/cbioportal-frontend#1786.

jjgao:
@oplantalech OK. Let's merge both once the frontend pr is ready.

PR #5359, opened 2018-12-03
DataAccessTokenController Unit Tests Switched to Use MockMVC

averyniceday:
Change controller unit tests to use mockMvc (instead of directly hitting controller methods)
Test applicationContext-security.xml file added to configure security portion

Additional small fix in the controller to use the bean-instantiation if dat.method is not defined (i.e get  `none` from dataAccessTokenServiceFactory) instead of instantiating a new UnauthDataAccessTokenServiceImpl(). Without this change, the controller does not use the mocked DataAccessTokenService bean created in the unit test configuration. 

NOTE: The current tests relies on some classes from the security module. Due to the circular dependency (security is dependent on web), they were copied over. They can be replaced when the dependency is broken. 

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -67,9 +69,8 @@
     @PostConstruct
     public void postConstruct() {
         if (datMethod == null || !SUPPORTED_DAT_METHODS.contains(datMethod)) {
```
angelicaochoa: @averyniceday since `none` is technically supported, then we can just check whether `datMethod==null` and throw some exception and if not null then always return 

```this.tokenService = this.dataAccessTokenServiceFactory.getDataAccessTokenService(this.datMethod);``` (instead of checking if it's a supported method)

File: web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java [hunk 2]
```diff
@@ -32,128 +32,231 @@
...
-        inheritLocations = false)
-public class DataAccessTokenControllerTest {    
+@Configuration
+public class DataAccessTokenControllerTest {
+
+    public static final String MOCK_USER = "MOCK_USER";
+    public static final String MOCK_PASSWORD = "MOCK_PASSWORD";
+    public static final String VALID_TOKEN_STRING = "VALID_TOKEN";
+    public static final String NONEXISTENT_TOKEN_STRING = "NONEXISTENT_TOKEN";
+    public static final String NOT_FOUND_ERROR_MESSAGE = "Specified token can not be found";
```
angelicaochoa: typo (cannot is one word)

web/src/test/java/org/cbioportal/web/DataAccessTokenControllerTest.java
```diff
@@ -32,128 +32,231 @@
...
+import org.cbioportal.service.exception.TokenNotFoundException;
 
 @RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations={"/applicationContext-web.xml", "/applicationContext-security.xml"})
 @WebAppConfiguration
-//TO-DO: Read from /applicationContext-web.xml without overriding test properties
-@ContextConfiguration(classes = DataAccessTokenControllerConfig.class)
-@TestPropertySource(
-        properties = {
-            "dat.method=uuid",
```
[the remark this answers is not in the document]
averyniceday: yup! the test properties won't affect anything since we're just making sure requests will get forwarded to the correct controller/service methods.

angelicaochoa:
Commenting here to remember to rebase on the merge of PR #5344

averyniceday:
already in release 3.0.0 branch

PR #5369, opened 2018-12-04
Study tags

oplantalech:
Added the cancer study identifier as part of the `CancerStudyTags` model, so that the endpoint retrieving multiple studies returns the cancer study identifier instead of the cancer study ID (e.g. `{"brca_tcga":{"analyst":"Frank", "load":"45"}}` instead of `{3:{"analyst":"Frank", "load":"45"}}`.


oplantalech:
We're not using this endpoint anymore.

PR #5403, opened 2018-12-06
Get project version from git

inodb:
Normally for each new release we have to create a PR just to update the version in the pom. Then afterwards we get conflicts when we try n merge master branch back into feature branches. Imagine updating master branch, now you will get conflicts in release-x branch and rc branch.

Maven should not keep track of the versioning at all, this is already managed by git. This maven-external-version plugin determines the version using a bash script, `version.sh`. This tries to get the version from git. If git is not available at compile time, one can also supply the version by setting an env variable `PROJECT_VERSION` instead.

I slightly modified maven-external-version plugin, so we can download that dependency from jitpack.

Unfortunately ${project.version} doesn't work in plugins, so we can't show the
version name in the actual war file name anymore (this is a known issue:
https://github.com/bdemers/maven-external-version/issues/7). Seems like an ok
trade-off though

Another issue I ran into was that `integration-test` didn't work anymore without running `install` first. Not sure if this is a reason for concern, but just wanted to notify @ersinciftci . There might be a way to update the plugin to avoid having to do this, but maybe it's not that big of a deal.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,9 +44,9 @@
     <build>
     <plugins>
           <plugin>
-            <groupId>com.github.inodb.maven-external-version</groupId>
+            <groupId>com.github.cbioportal.maven-external-version</groupId>
             <artifactId>maven-external-version-plugin</artifactId>
-            <version>b7bc6a6ff7d51295089d5dc7ff64084063440cc6</version>
+            <version>67e9fc4ebb17333b119ab10e1b521fe8391b7ff2</version>
             <extensions>true</extensions>
             <configuration>
               <strategy hint="script">
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,9 @@
   <artifactId>master</artifactId>
   <packaging>pom</packaging>
   <name>Portal Master</name>
-  <version>1.18.2-SNAPSHOT</version>
+  <!-- project version is generated through git or can be passed as
+	   PROJECT_VERSION env variable (see version.sh) -->
+  <version>0-unknown-version-SNAPSHOT</version>
   <description>master maven module</description>
 
   <scm>
@@ -23,6 +25,12 @@
       </snapshots>
     </repository>
   </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <id>jitpack.io</id>
+      <url>https://jitpack.io</url>
+    </pluginRepository>
+  </pluginRepositories>
 
   <profiles>
     <profile>
@@ -35,6 +43,17 @@
       </properties>
     <build>
     <plugins>
+          <plugin>
+            <groupId>com.github.inodb.maven-external-version</groupId>
+            <artifactId>maven-external-version-plugin</artifactId>
+            <version>b7bc6a6ff7d51295089d5dc7ff64084063440cc6</version>
+            <extensions>true</extensions>
+            <configuration>
+              <strategy hint="script">
+                <script>./version.sh</script>
+              </strategy>
+            </configuration>
+          </plugin>
           <plugin>
             <groupId>org.jacoco</groupId>
             <artifactId>jacoco-maven-plugin</artifactId>
@@ -250,7 +269,7 @@
     <jacoco-maven-plugin.version>0.7.6.201602180812</jacoco-maven-plugin.version>
     <mockito.version>1.10.19</mockito.version>
 
-    <final.war.name>cbioportal-${project.version}</final.war.name>
+    <final.war.name>cbioportal</final.war.name>
 
   <db.test.driver>com.mysql.jdbc.Driver</db.test.driver>
     <!-- For MySQL < 5.5 change 'default_storage_engine' to 'storage_engine' -->
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
           <plugin>
             <groupId>com.github.cbioportal.maven-external-version</groupId>
             <artifactId>maven-external-version-plugin</artifactId>
-            <version>67e9fc4ebb17333b119ab10e1b521fe8391b7ff2</version>
+            <version>f09c2b9608744881111f24e55d55a91e54e6cb5f</version>
             <extensions>true</extensions>
             <configuration>
               <strategy hint="script">
--- a/version.sh
+++ b/version.sh
@@ -2,8 +2,8 @@
 if [ -n "$PROJECT_VERSION" ]; then
     # use env variable PROJECT_VERSION if defined
     echo $PROJECT_VERSION
-elif ! [ -x "$(command -v git)" ]; then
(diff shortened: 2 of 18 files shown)


version.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+if [ -n "$PROJECT_VERSION" ]; then
+    # use env variable PROJECT_VERSION if defined
+    echo $PROJECT_VERSION
```
fedde-s: This looks should do for the use-case I was concerned about: building docker images based on a development source folder without including all other branches, reflogs, stashed patches and configuration such as remote repos into the image.

Does this run during the Maven build? If so, we can basically run the below git command _outside_ of the build container, and pass it as a `--build-arg`.

pom.xml
```diff
@@ -35,6 +43,17 @@
...
     <build>
     <plugins>
+          <plugin>
+            <groupId>com.github.inodb.maven-external-version</groupId>
+            <artifactId>maven-external-version-plugin</artifactId>
+            <version>b7bc6a6ff7d51295089d5dc7ff64084063440cc6</version>
+            <extensions>true</extensions>
+            <configuration>
+              <strategy hint="script">
+                <script>./version.sh</script>
```
fedde-s: Do you know at which phase of the Maven build lifecycle the script will run? I've been trying to use `mvn dependency:go-offline` to download at least some of the dependencies and plugins separately before building the actual source code, and if it makes sense, I believe it would be convenient for my automation if `version.sh` wouldn't be invoked until a later step.

fedde-s: Annoyingly, it looks like this script *does* run earlier in the build lifecycle, which would leave one of the workflows I maintain in `thehyve/cbioportal-docker` (making a PR to `cBioPortal/cbioportal` is backlogged for a less-busy time of year) unable to use the new feature.

pom.xml
```diff
@@ -35,6 +43,17 @@
       </properties>
     <build>
     <plugins>
+          <plugin>
+            <groupId>com.github.inodb.maven-external-version</groupId>
```
fedde-s: Should this (eventually) move from this privately-maintained repo to a cBioPortal community effort?

pom.xml
inodb: @fedde-s hmm that is tricky - im not sure if it's possible to change it to a different lifecycle. One thing we perhaps could do is to allow disabling the plugin completely. Then instead of using this script, you could do a find & replace on `0-unknown-version-SNAPSHOT`. Since we won't change the version in the pom anymore, you could say this is a new feature as well ;)?

pom.xml
inodb: @fedde-s i can fork it to the cBioPortal repo to make it look more official before merging

pom.xml
inodb: I mostly went for a maven approach in this PR, because I figured this way developers/deployers don't have to change their existing workflow. It is much easier imo to do a find and replace before running maven, but figured it be nicer not to request people to type a different cmd when compiling

fedde-s: For now, I can just make that one Dockerfile use-case hard-code the `PROJECT_VERSION` to a fixed string. It would be very useful to be able to see the commit hash here, but only if it can be read *after* dependencies are cached. I think that feature should go on a backlog and not block your PR though, as only a few people actually use that Dockerfile as far as I'm aware.

inodb:
Thanks for the quick review @fedde-s !

pom.xml
fedde-s: Ah, I see that it's a fork of an existing project. If the changes you made get accepted upstream, it might get even more people to look at it and maintain it. Do what you feel is best! 🙂

fedde-s:
Confirmed: if I tag the commit something starting with a `v`, the frontend displays the name of the tag as the version number and the `info` API call displays it without the `v` and with the suffix `-SNAPSHOT`.

inodb:
@fedde-s I get something like this locally: `1.18.1-4-g4b178bdde-dirty-SNAPSHOT` (It's in a dirty state right now and i rebased). With `git describe` it should give you something recent. Maybe you don't have all the tags pulled? `git fetch --tags` would solve that. I can include that command in the `version.sh` script

pom.xml
inodb: @fedde-s yeah it's unfortunately not actively maintained 😢. Im not sure why, seems like a lot of people still use it and I haven't found a good alternative. I've forked it to our org and made a single commit that describes the changes I made better: https://github.com/cBioPortal/maven-external-version/commit/67e9fc4ebb17333b119ab10e1b521fe8391b7ff2. It's just for allowing one to pull the dependency from jitpack (the plugin is not published on maven central)

fedde-s:
@inodb: you're right! I don't think you need to add the command: I'd specifically left them out in minimal my test environment, but will almost always base actual production deployments on a tag.

version.sh
```diff
@@ -0,0 +1,13 @@
...
+elif ! [ -x "$(command -v git)" ] || ! [ -d .git ]; then
+	# git is not installed, or not in a git repo
+    echo '0-unknown-version'
+else
+	# fetch all tags silently
+	git fetch --tags --quiet > /dev/null 2> /dev/null
+	# get human readable commit id from git
+	git describe --tags --always --dirty | sed 's/^v//'
```
sheridancbio: If git describe always goes back to a tag, does it mean that portal dependencies can only be brought in from a tagged commit that is pushed to the cbioportal github repo? If a single PR makes changes in both the service and web layer (like changing the service interface), will the "new" web layer be able to bring in the "new" service layer even if a tag has not been created (in a pr to rc, for instance)? Maybe the extra info after the tag is enough to provide the additional path since the tag (the "dirty" part I guess?) 

% git describe --tags --always --dirty

v1.18.0-120-g05c7ad6-dirty

inodb: @sheridancbio Portal dependencies can be brought in locally or remotely. It doesn't need to be tagged/committed in the github repo. Maven continues to function in the same way, so if u compile something locally the artifacts will end up in your `~/.m2` folder and you can include them in another project as a dependency.

sheridancbio: sounds good. I was mainly thinking about dependencies between the modules within cbioportal, but packaging the cbioportal code into other codebases is also needed.

inodb:
I found another issue when including cbioportal through jitpack as a dependency as we do with e.g. genome-nexus-annotation-pipeline. The scripts pom would not get the updated parent project's version. Not generating a new pom by the shaded maven plugin solved that issue

fedde-s:
What do you mean? Are you no longer generating the scripts JAR? It's required for all data loading.

fedde-s:
So if that's the case, we have to revert the merge right away.

PR #5405, opened 2018-12-06
Enforce copy number segment file format during import

averyniceday:
# What? Why?
Records were previously constructed using variables assigned through indices. Incorrect data would still be imported successfully since there were no checks. Code assumed that certain fields would be in certain positions (i.e seg.mean always at position 4). 

- Create a required field: index map to check required fields are present
- Use field:index map for variable assignment and record construction (will take into account if columns are in a different order)

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCopyNumberSegmentData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCopyNumberSegmentData.java
@@ -48,19 +48,20 @@ import java.util.*;
  */
 public class ImportCopyNumberSegmentData extends ConsoleRunnable {
 
+    private static final String[] requiredHeaderFields = {"ID", "chrom", "loc.start", "loc.end", "num.mark", "seg.mean"};
     private int entriesSkipped;
-    
+
     private void importData(File file, int cancerStudyId) throws IOException, DaoException {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCopyNumberSegmentData.java
```diff
@@ -81,20 +82,20 @@ private void importData(File file, int cancerStudyId) throws IOException, DaoExc
...
                 if (start >= end) {
                     //workaround to skip with warning, according to https://github.com/cBioPortal/cbioportal/issues/839#issuecomment-203452415
                     ProgressMonitor.logWarning("Start position of segment is not lower than end position. Skipping this entry.");
                     entriesSkipped++;
                     continue;
                 }            
-                int numProbes = new BigDecimal((strs[4])).intValue();
-                double segMean = Double.parseDouble(strs[5]);
+                int numProbes = new BigDecimal((strs[headerIndexMap.get("num.mark")])).intValue();
+                double segMean = Double.parseDouble(strs[headerIndexMap.get("seg.mean")]);
```
sheridancbio: and these indices *will* be found because getIndexMap(line) will throw and exception if all required fields are not present.

PR #5424, opened 2018-12-11
Use REPLACE for the SAMPLE_COUNT attribute

ersinciftci:


inodb:
@ersinciftci Thanks! Looks good, but would be good to add some test(s) with split patient/sample files to `TestImportProfileData.java`

PR #5459, opened 2018-12-17
Documentation to help new (backend) developers with testing part

pieterlukasse:
# What? Why?

Documentation to help new (backend) developers with testing part.

@inodb @ersinciftci maybe you have an idea of where this document should be linked so developers can easily find it?


inodb:
@pieterlukasse This is great! Maybe we can copy the whole thing to the `README.md`? Or have a link from there to this doc. Then the doc itself could be part of section `4. Development`

pieterlukasse:
Thanks @inodb . I will update it.

PR #5460, opened 2018-12-17
Implementation of treatment profile data type

pvannierop:
# What? Why?
This PR implements a new feature involving validation, import and presentation of
treatment and treatment response profile data by the cBioPortal backend. See [RFC 49](https://docs.google.com/document/d/1jbpnWjlVxEpigFkNhr9LzhZ0rmUy9WleYa_neUSV4Js) for more details.

## Changes proposed in this pull request
- Update database schema:
    - New table for the `treatment` entity.
    - New `pivot_threshold` and `sort_order` fields in the genetic_profile table.
- Provide support for IC50, EC50, GI50 and AUC (area under the curve) data types for treatment
response profile data (described in File-Formats.md).
- Added treatment profile meta and data files to the study_es_0, study_es_1 and study_es_3 test data sets.
- Python validation scripts check for consistency between different treatment response data
files in a study (includes unit and system tests)
- Implemented new Java import classes for parsing and persistence of treatments and treatment
responses profiles (includes integration tests).
- REST service updated to present treatment and treatment response profile data (includes unit tests).
- Various code improvements in python and java layers to avoid code duplication
- In order to allow a n number of treatment data files for treatments the _allowed_data_types_ file allows for a wild card (_*_) as `STABLE_ID`. The user is responsible for providing unique stable ids in the profile meta-files (checked by validation scripts).

## Remarks
- Treatments are not preloaded nor part of the seed database, but imported/updated on the fly from
`feature_id`, `name`, `description` and `url` columns in the treatment profile data files.
- h2 version used for web-service tests was updated to accommodate ENUM SQL type used in
the genetic_profile table.
- Username and password used for integration tests were changed to test credentials.

# Checks
- [ ] Runs on Heroku.


core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -941,25 +965,184 @@ def test_missing_row_gsva(self):
...
+    def test_zero_issues_error(self):
+        passConcValue("0").error.assert_called()
+        passConcValue("0.000").error.assert_called()
+        passConcValue("00000").error.assert_called()
+
+    def test_leadingzeros_are_allowed(self):
+        passConcValue("0000.123").error.assert_not_called()
+
+    def test_negativenumber_issues_error(self):
+        passConcValue("-234234.234").error.assert_called()
```
pieterlukasse: I wouldn't block negative numbers now that we know that log values could be one of the use cases.

db-scripts/src/main/resources/cgds.sql
```diff
@@ -290,16 +303,18 @@ CREATE TABLE `reference_genome` (
...
   `DATATYPE` varchar(255) NOT NULL,
   `NAME` varchar(255) NOT NULL,
   `DESCRIPTION` mediumtext,
-  `SHOW_PROFILE_IN_ANALYSIS_TAB` BOOLEAN NOT NULL,
+  `SHOW_PROFILE_IN_ANALYSIS_TAB` tinyint(1) NOT NULL,
+  `PIVOT_THRESHOLD` FLOAT DEFAULT NULL,
+  `SORT_ORDER` ENUM('ASC','DESC') DEFAULT NULL,
   PRIMARY KEY (`GENETIC_PROFILE_ID`),
-  UNIQUE (`STABLE_ID`),
+  UNIQUE KEY `STABLE_ID` (`STABLE_ID`),
```
pieterlukasse: why? Is this change also in migrate.sql?

persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/TreatmentMyBatisRepositoryTest.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright (c) 2016 Memorial Sloan-Kettering Cancer Center.
```
pieterlukasse: The Hyve

service/src/main/java/org/cbioportal/service/TreatmentDataService.java
```diff
@@ -0,0 +1,12 @@
+package org.cbioportal.service;
```
pieterlukasse: Add copyright

service/src/main/java/org/cbioportal/service/TreatmentService.java
```diff
@@ -0,0 +1,14 @@
+package org.cbioportal.service;
```
pieterlukasse: Add copyright

service/src/main/java/org/cbioportal/service/impl/TreatmentDataServiceImpl.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright (c) 2016 The Hyve B.V.
```
pieterlukasse: yes, but 2019

pieterlukasse:
@pvannierop Please also rebase and fix conflicts

core/src/main/scripts/importer/validateData.py
```diff
@@ -3421,7 +3421,7 @@ class TreatmentWiseFileValidator(MultipleDataFileValidator, metaclass=ABCMeta):
     prior_validated_header = None
     REQUIRED_HEADERS = ['treatment_id']
     OPTIONAL_HEADERS = ['name', 'description', 'url']
-    UNIQUE_COLUMNS = ['treatment_id', 'name', 'description', 'url']
+    UNIQUE_COLUMNS = ['treatment_id']
```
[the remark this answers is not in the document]
pvannierop: 1. The check for the `name` field was added to be unique in treatment data files for a single study
`UNIQUE_COLUMNS = ['treatment_id']`
became
`UNIQUE_COLUMNS = ['treatment_id', 'name']`

2. I implemented the check for the currently stored name for a given treatment_id. A warning is issued when the new treatment data set overwrites names currently stored values. This update includes new unit test for this behavior.

.travis.yml
```diff
@@ -73,8 +73,8 @@ script:
           -e \
           -Ppublic \
           -Dfinal.war.name=cbioportal \
-          -Ddb.user=cbio_user \
-          -Ddb.password=somepassword \
+          -Ddb.test.username=cbio_user \
+          -Ddb.test.password=somepassword \
```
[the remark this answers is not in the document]
pvannierop: Wellcome!

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneset.java
```diff
@@ -247,58 +247,6 @@ private static Geneset extractGeneset(ResultSet rs) throws SQLException, DaoExce
         return geneset;
     }
     
-    public static Set<Long> getGenesetGeneticEntityIds() throws DaoException {
```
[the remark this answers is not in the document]
pvannierop: Sorry. File was reverted to default state.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticProfile.java
```diff
@@ -217,6 +230,10 @@ private static GeneticProfile extractGeneticProfile(ResultSet rs) throws SQLExce
         profileType.setGeneticAlterationType(GeneticAlterationType.valueOf(rs.getString("GENETIC_ALTERATION_TYPE")));
         profileType.setDatatype(rs.getString("DATATYPE"));
         profileType.setGeneticProfileId(rs.getInt("GENETIC_PROFILE_ID"));
+        if ((Float) rs.getFloat("PIVOT_THRESHOLD") != null) {
```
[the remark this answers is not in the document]
pvannierop: Ah this is useless indeed. It was to check whether a pivot threshold was passed or not. The ResultSet.getFloat() docs mention:
>Returns:
the column value; if the value is SQL NULL, the value returned is 0

I have updated this line to:
`if (rs.getFloat("PIVOT_THRESHOLD") != 0) {
...}`

db-scripts/src/main/resources/cgds.sql
pvannierop: No idea. Line was reverted to original state.

persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/TreatmentMyBatisRepositoryTest.java
pvannierop: Changed to:
` * Copyright (c) 2019 The Hyve.`

service/src/main/java/org/cbioportal/service/TreatmentDataService.java
pvannierop: Copyright statement added.

service/src/main/java/org/cbioportal/service/TreatmentService.java
pvannierop: Copyright statement added.

service/src/main/java/org/cbioportal/service/impl/TreatmentDataServiceImpl.java
pvannierop: Copyright statement added.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTreatments.java
```diff
@@ -0,0 +1,237 @@
...
+    */
+    public static void startImport(OptionSet options, OptionSpec<String> data, boolean updateInfo){
+        try {
+            if (options.hasArgument(data)) {
+                File treatmentFile = new File(options.valueOf(data));
+                importData(treatmentFile, updateInfo);
+            }
+        }
+        catch (Exception ex) {
+            ex.printStackTrace();
```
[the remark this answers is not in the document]
pvannierop: Updated. This code is the same ImportGeneSets.java. Should I add a TODO above this code?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoTreatment.java
```diff
@@ -0,0 +1,272 @@
+/*
+ * Copyright (c) 2016 Memorial Sloan-Kettering Cancer Center.
```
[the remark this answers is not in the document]
pvannierop: Changed.

core/src/main/java/org/mskcc/cbio/portal/model/GeneticProfile.java
```diff
@@ -51,150 +51,176 @@
...
+        this.setGeneticProfileId(template.geneticProfileId);
+        this.setTargetLine(template.getTargetLine());
+        this.setOtherMetadataFields(template.getAllOtherMetadataFields());
+    }
+    
+    public int getGeneticProfileId() {
         return geneticProfileId;
     }
-
+
```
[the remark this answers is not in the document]
pvannierop: Fixed.

core/src/main/java/org/mskcc/cbio/portal/model/Treatment.java
```diff
@@ -0,0 +1,170 @@
+/*
+ * Copyright (c) 2016 Memorial Sloan-Kettering Cancer Center.
```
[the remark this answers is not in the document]
pvannierop: Done in all larger new files.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -508,6 +553,54 @@ private boolean parseGenesetLine(String line, int nrColumns, int sampleStartInde
...
+     * @param treatmentIdIndex  index of the column that uniquely identifies a sample
+     * @param filteredSampleIndices
+     * @param daoGeneticAlteration
+     * @return
+     * @throws DaoException 
+     */
+    // TODO: TREATMENT BACKLOG this code is a duplicate of the parseGenesetLine.
+    // consider refactoring do that both functions are covered by a single
+    // base class
+    private boolean parseTreatmentLine(String line, int nrColumns, int sampleStartIndex, int treatmentIdIndex,
```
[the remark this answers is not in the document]
pvannierop: ImportTreatment.java imports treatments whereas ImportTabDelimitedData imports treatment profile data. Both need to be there. You can see the logic in ImportProfileData.java (lines 90-96):
>            } else if (geneticProfile.getGeneticAlterationType() == GeneticAlterationType.TREATMENT_RESPONSE) {
>
>                // add all missing Treatments to the database
>                ImportTreatments.importData(dataFile, true);
>
>                ImportTabDelimData treatmentProfileImporter = new ImportTabDelimData(dataFile, >geneticProfile.getTargetLine(), geneticProfile.getGeneticProfileId(), genePanel);
>                treatmentProfileImporter.importData(numLines);

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -150,47 +156,58 @@ private static GeneticProfileLink createGeneticProfileLink(GeneticProfile geneti
         // In the future with other types of genetic profile links, this should be configurable in the meta file. 
         String referenceType;
         if (Arrays.asList("P-VALUE", "Z-SCORE").contains(geneticProfile.getDatatype())) {
-        	referenceType = "STATISTIC";
+            referenceType = "STATISTIC";
```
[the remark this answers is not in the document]
pvannierop: Did not find what was the difference, but I changed the white space to the state in rc.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3341,7 +3477,127 @@ def checkValue(self, value, col_index):
...
+
+        # value is not defined
+        if value == "":
+            self.logger.error("Cell is empty. An effective concentration value is expected. Use 'NA' to indicate missing values.",
+                extra={'line_number': self.line_number,
+                'column_number': col_index + 1,
+                'cause': value})
+            return
+        
+        stripped_value = value.strip()
```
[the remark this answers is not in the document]
pvannierop: AFAICS the gene set routine also doe not perform white space trimming. I added this func to the ImportTabDelimitedData.java --> parseTreatmentLine:

> // trim whitespace from values
            values = Stream.of(values).map(String :: trim).toArray(String[]::new);

Should also be added to importGenesetLine?

core/src/main/scripts/importer/validateData.py
```diff
@@ -3341,7 +3477,127 @@ def checkValue(self, value, col_index):
...
+
+        # 'NA' is an allowed value. No further validations apply.
+        if stripped_value == 'NA':
+            return
+
+        # if the value is prefixed with '>' remove this prefix
+        # prior to evaluation of the numeric value
+        hasTruncSymbol = re.match("^>", stripped_value)
+        stripped_value = re.sub("^>","", stripped_value)
```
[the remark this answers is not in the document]
pvannierop: Wow, I really forgot to add this. How stupid. I have updated the validateData.py code and added a unit test that checks "<" acceptance.

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -941,25 +965,184 @@ def test_missing_row_gsva(self):
         self.assertEqual(len(record_list2), 1)
         for record in record_list2:
             self.assertEqual(record.levelno, logging.ERROR)
-        self.assertEqual('Gene sets column in score and p-value file are not equal', record.getMessage())
+        self.assertEqual('Feature columns in data files are not equal.', record.getMessage())
```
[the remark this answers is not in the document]
pvannierop: Ok, I delegated definition of the message to the child classes. Gene set data file checking now returns:

>Gene sets column in score and p-value file are not equal.

Treatment data file checking now returns:
>Treatment feature columns (`treatment_id`, ...) in treatment profile data files are not identical.

core/src/test/scripts/unit_tests_validate_data.py
pvannierop: The idea was that concentrations can never be negative. I plan to implement two treatment data types. One is TREATMENT_CONCENTRATION (must be > 0), the other TREATMENT_CONTINUOUS (any numerical value). When a neg.log of concentration is passed the TREATMENT_CONTINUOUS data type should be used. Alternatively, we only create a TREATMENT_CONTINUOUS that is applied to every data file. You prefer the latter?

core/src/main/scripts/importer/cbioportal_common.py
```diff
@@ -59,6 +59,11 @@ class MetaFileTypes(object):
     GENE_PANEL_MATRIX = 'meta_gene_panel_matrix'
     GSVA_SCORES = 'meta_gsva_scores'
     GSVA_PVALUES = 'meta_gsva_pvalues'
+    TREATMENT_IC50 = 'meta_treatment_ic50'
+    TREATMENT_EC50 = 'meta_treatment_ec50'
+    TREATMENT_GI50 = 'meta_treatment_gi50'
+    TREATMENT_AUC = 'meta_treatment_auc'
```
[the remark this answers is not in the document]
pvannierop: Handling of generic treatment response files was added.

core/src/main/scripts/importer/cbioportal_common.py
```diff
@@ -246,6 +251,54 @@ class MetaFileTypes(object):
         'data_filename': True,
         'show_profile_in_analysis_tab': True,
         'geneset_def_version': True
+    },
+    MetaFileTypes.TREATMENT_AUC: {
```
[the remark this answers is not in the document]
pvannierop: Handling of generic treatment response files was added.

core/src/main/scripts/importer/allowed_data_types.txt
```diff
@@ -36,3 +36,7 @@ PROTEIN_LEVEL	CONTINUOUS	protein_quantification
 PROTEIN_LEVEL	Z-SCORE	protein_quantification_zscores
 GENESET_SCORE	GSVA-SCORE	gsva_scores
 GENESET_SCORE	P-VALUE	gsva_pvalues
+TREATMENT_RESPONSE	IC50	treatment_ic50
+TREATMENT_RESPONSE	EC50	treatment_ec50
+TREATMENT_RESPONSE	GI50	treatment_gi50
+TREATMENT_RESPONSE	AUC	treatment_auc
```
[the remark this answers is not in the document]
pvannierop: Handling of generic treatment response files was added.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTreatments.java
```diff
@@ -109,15 +109,10 @@ public void run() {
...
     */
-    public static void startImport(OptionSet options, OptionSpec<String> data, boolean updateInfo){
-        try {
-            if (options.hasArgument(data)) {
-                File treatmentFile = new File(options.valueOf(data));
-                importData(treatmentFile, updateInfo);
-            }
-        }
-        catch (Exception ex) {
-            ex.printStackTrace();
```
pieterlukasse: yes, please add the TODO to remove this one in ImportGenesetData

docs/File-Formats.md
```diff
@@ -1251,10 +1252,58 @@ The cells contain the p-value for the GSVA score: A real number, between 0.0 and
...
+
+### Treatment response meta file
+The meta file will be similar to meta files of other genetic profiles, such as mRNA expression. For treatment response data `RESPONSE-VALUE` is used as `datatype`. Any number of treatment response parameters can be loaded by means of different data files. For treatment meta files values for `stable_id` are not pre-registered in the cBioPortal implementation, but can be freely chosen by the user. Requirement is that each treatment meta file is assigned an unique `stable_id`.
+
+Required fields: 
+```
+cancer_study_identifier: Same value as specified in study meta file
+genetic_alteration_type: TREATMENT_RESPONSE
+datatype: RESPONSE-VALUE
+stable_id: Any unique identifier within the study
```
pieterlukasse: thinking about error scenarios...it would be nice if we say "any unique identifier using a combination of alphanumeric characters and _". This is the current format for other stable ids. I just don't know what could break if one would deviate from it now. So let's validate it as such.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -584,6 +586,9 @@ private boolean parseTreatmentLine(String line, int nrColumns, int sampleStartIn
             }
             
             String values[] = (String[]) ArrayUtils.subarray(parts, sampleStartIndex, parts.length>nrColumns?nrColumns:parts.length);
+
+            // trim whitespace from values
+            values = Stream.of(values).map(String::trim).toArray(String[]::new);
```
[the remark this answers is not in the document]
pieterlukasse: yes, maybe also add to geneset part...or add a TODO there

core/src/main/scripts/importer/validateData.py
```diff
@@ -3336,16 +3514,90 @@ class GsvaPvalueValidator(GsvaWiseFileValidator):
...
+        stripped_value = value.strip()
+
+        # 'NA' is an allowed value. No further validations apply.
+        if stripped_value == 'NA':
+            return
+
+        # if the value is prefixed with '>' or '<' remove this prefix
+        # prior to evaluation of the numeric value
+        hasTruncSymbol = re.match("^[><]", stripped_value)
+        stripped_value = re.sub("^[><]","", stripped_value)
```
pieterlukasse: what happens if values is like `> 10` instead of `>10`?

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTreatmentData.java
```diff
@@ -0,0 +1,120 @@
...
+
+        // Test whether fields were populated correctly
+        assertEquals("Name of Irinotecan", treatment1.getName());
+        assertEquals("Desc of Irinotecan", treatment1.getDescription());
+        assertEquals("Url of Irinotecan", treatment1.getRefLink());
+
+        // test fields are left unchanged when updateInfo flag is not setcd 
+        // and the next data file contains different feature meta data
+        File fileNewDesc = new File("src/test/resources/treatments/data_treatment_ic50_newdesc.txt");
+        ImportTreatments.importData(fileNewDesc, updateInfo);
```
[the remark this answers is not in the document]
pieterlukasse: is not a use case I think. You might as well just always update. Less is more :)

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
+    def _get_prior_validated_sample_ids(cls):
+        return TreatmentWiseFileValidator.prior_validated_sample_ids
+
+    @classmethod
+    def _set_prior_validated_sample_ids(cls, sample_ids):
+        TreatmentWiseFileValidator.prior_validated_sample_ids = sample_ids
+    
+    @classmethod
+    def _get_message_features_do_not_match(cls):
+        return "Treatment feature columns (`treatment_id`, ...) in treatment profile data files are not identical."
```
pieterlukasse: hm...I had to read the code to understand this error..... Maybe one idea would be to add more details to this error message, perhaps in the "cause" section. Something like: "The same set of treatments should be used across the different treatment data files for this study. Check that all treatment ids of one file are also present in all other treatment files".

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticProfile.java
```diff
@@ -217,6 +230,10 @@ private static GeneticProfile extractGeneticProfile(ResultSet rs) throws SQLExce
         profileType.setGeneticAlterationType(GeneticAlterationType.valueOf(rs.getString("GENETIC_ALTERATION_TYPE")));
         profileType.setDatatype(rs.getString("DATATYPE"));
         profileType.setGeneticProfileId(rs.getInt("GENETIC_PROFILE_ID"));
+        if (rs.getFloat("PIVOT_THRESHOLD") != 0) {
+            profileType.setPivotThreshold(rs.getFloat("PIVOT_THRESHOLD"));
+            profileType.setSortOrder(rs.getString("SORT_ORDER"));
```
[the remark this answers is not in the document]
pvannierop: Ok, changed to:

>        if (rs.getFloat("PIVOT_THRESHOLD") != 0) {
>            profileType.setPivotThreshold(rs.getFloat("PIVOT_THRESHOLD"));
>        }
>        if (rs.getString("SORT_ORDER") != null && ! rs.getString("SORT_ORDER").equals("") ) {
>            profileType.setSortOrder(rs.getString("SORT_ORDER"));
>        }
>        return profileType;

core/src/main/scripts/importer/validateData.py
```diff
@@ -48,7 +48,7 @@
 from abc import ABCMeta, abstractmethod
 
 # configure relative imports if running as a script; see PEP 366
-if __name__ == "__main__" and __package__ is None:
+if __name__ == "__main__" and (__package__ is None or __package__ is '') :
```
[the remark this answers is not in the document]
pvannierop: This is the only way I can get this to work in VScode. Here __package__ returns '' not None. May this be caused by a different python interpreter (I use 3.6.7)?

core/src/main/scripts/importer/validateData.py
```diff
@@ -3421,7 +3423,32 @@ class TreatmentWiseFileValidator(MultipleDataFileValidator, metaclass=ABCMeta):
...
+    
+    def checkDifferentNameInDb(self, nonsample_col_vals):
+        # Check for different combinations of treatment_id and treatment_name
+        # in the database. If true, raise warnings for each discrepancy.
+        nonsample_cols =  self.nonsample_cols
+        if (nonsample_cols.index("name") != None):
+
+            treatmentId = nonsample_col_vals[ nonsample_cols.index("treatment_id") ]
+            fileTreatmentName = nonsample_col_vals[ nonsample_cols.index("name") ]
+            dbTreatmentName = self.portal.treatment_map[treatmentId]["name"]
```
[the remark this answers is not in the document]
pvannierop: You are right. The code did not account for the fact that a particular treatment_id was not added to the database before. Please not that if a treatment_id is entered there is _always_ a name field added; this can not be a reason for the `dbTreatmentName` to be `None`. The new code is:

>            # check whether a name for the treatment has been
>            # registered in the database
>            dbTreatmentName = None
>            if self.portal.treatment_map[treatmentId] != None:
>                dbTreatmentName = self.portal.treatment_map[treatmentId]["name"]
>
>            # when a name has been registered for this treatment and 
>            # is different from the new name, issue a warning.
>            if dbTreatmentName != None and dbTreatmentName != fileTreatmentName:
>                    self.logger.warning(

core/src/main/scripts/importer/validateData.py
```diff
@@ -3308,6 +3306,11 @@ def _get_prior_validated_sample_ids(cls):
     @abstractmethod
     def _set_prior_validated_sample_ids(cls, sample_ids):
         pass
+    
+    @classmethod
+    @abstractmethod
+    def _get_message_features_do_not_match():
+        pass
```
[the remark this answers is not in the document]
pvannierop: This is an abstract method (ABC package). The implementaton (`pass`) does not matter but I do not know whether the `pass` statements can be removed. Does there need to be a body?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
```diff
@@ -14,6 +14,10 @@
     private DaoGeneticEntity() {
     }
     
+    // TODO consider renaming this class to something similar to
```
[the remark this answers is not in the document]
pvannierop: TODO comment was updated.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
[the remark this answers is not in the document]
pvannierop: Also added to geneset code.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTreatments.java
pvannierop: Done.

docs/File-Formats.md
pvannierop: Docs were updates.

core/src/main/scripts/importer/validateData.py
pvannierop: They are not recognized. It is not mentioned that a space is allowed in File-Formats.md. But then again there is also no statement that it is not. As on the effect of my code. After removal of `>` or `<` the remaining string is passed through `float(stripped_value)`. I tried float(" 5") (notice the space here) and this evaluates to 5.0. So, if there would be a space no error would be raised and the value is accepted by the validator. I guess we may add a remark to the docs that a whitespace character is allowed (and adjust the study_es_0 to include one of these values)?

core/src/main/scripts/importer/validateData.py
pvannierop: Ok, errors now contain a cause in addition to the original message. For treatments and gene sets these are respectively:

>"message": "Treatment feature columns (`treatment_id`, ...) in treatment profile data files are not identical."
>"cause": "The same set of treatments should be used across the different treatment data files for this study. Please ensure that all treatment id's of one file are present in all other treatment files."

>"message": "Gene sets column in score and p-value file are not equal."
>"cause": "The same set of gene sets should be used in the score and p-value files for this study. Please ensure that all gene set id's of one file are present in the other gene set data file."

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
pieterlukasse: @fedde-s any ideas?

core/src/main/scripts/importer/validateData.py
pieterlukasse: maybe try `re.sub("^[><]\s*"`

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
fedde-s: This change does not belong in the code, it's an undocumented workaround for a misconfigured IDE.

As the PEP referenced on the line above says, `__package__` will be `None` if the file is run as a script, or it will be `importer` if the file is run from the module search path as `importer.validateData` (or as `from importer import validateData`).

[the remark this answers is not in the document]
fedde-s: As a more proper solution, it be worth refactoring the package structure to remove the need for this boilerplate, by either:
- replacing the relative imports by absolute ones (as the folder containing the script [will automatically end up](https://docs.python.org/3/tutorial/modules.html#the-module-search-path) in the search path), or
- renaming `cbioportalImporter.py` and validateData.py` and creating new wrapper scripts in their place, so that only the wrapper scripts need to do an absolute import.

Both options will require a change to the way the automated tests import these scripts, and should probably be reviewed as a separate pull request.

core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -14,7 +14,7 @@
 from pathlib import Path
 
 # configure relative imports if running as a script; see PEP 366
-if __name__ == "__main__" and __package__ is None:
+if __name__ == "__main__" and (__package__ is None or __package__ is ''):
```
fedde-s: As I responded above, this undocumented workaround does not belong in the code.

core/src/main/scripts/importer/metaImport.py
```diff
@@ -14,7 +14,7 @@
 from pathlib import Path
 
 # configure relative imports if running as a script; see PEP 366
-if __name__ == "__main__" and __package__ is None:
+if __name__ == "__main__" and (__package__ is None or __package__ is ''):
```
fedde-s: As I responded above, this undocumented workaround does not belong in the code.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
+                if not self._get_prior_validated_feature_ids() == self.feature_ids:
+                    self.logger.error( self._get_message_features_do_not_match() )
 
-        checkConsistencyScoresPvalue(self)
+        checkConsistencyFeatures(self)
 
-        super(GsvaWiseFileValidator, self).onComplete()
+        super(MultipleDataFileValidator, self).onComplete()
+
+# TODO: move duplicate code in GsvaWiseFileValidator and TreatmentWiseFileValidator classes to meta class
```
fedde-s: Do you mean moving them to the superclass? I see no reason for metaprogramming there.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3336,16 +3514,90 @@ class GsvaPvalueValidator(GsvaWiseFileValidator):
...
+                'cause': value})
+            return
+
+        if numeric_value % 1 == 0 and not hasTruncSymbol:
+            self.logger.warning("Value has no decimals and may represent an invalid response value.",
+                extra={'line_number': self.line_number,
+                'column_number': col_index + 1,
+                'cause': value})
+                
+        return
```
fedde-s: This `return` doesn't add anything

core/src/main/scripts/importer/validateData.py
```diff
@@ -3835,6 +4087,14 @@ def index_geneset_id_list(json_data,
             result_list.append(geneset_id)
     return result_list
 
+def index_treatment_data(json_data,
+                         id_field = "treatmentId"):
```
fedde-s: Styling: no spaces around keyword/default arguments unless combined with type annotations. ([PEP8](https://www.python.org/dev/peps/pep-0008/#other-recommendations))

core/src/main/scripts/importer/validateData.py
```diff
@@ -4196,6 +4465,12 @@ def main_validate(args):
 # ------------------------------------------------------------------------------
 # vamanos
 
+def _get_column_index(parts, name):
```
fedde-s: The comment does not apply to this function. You should either define the function _above_ the comments like all other functions, or remove the comment.

core/src/main/scripts/importer/validateData.py
fedde-s: Also, I think strings in this file tend to use single quotes unless they're docstrings or likely to contain apostrophes.

core/src/main/scripts/importer/validateData.py
fedde-s: Just to be sure, it might than also be worth [prefixing the string with an `r`]( https://docs.python.org/3/tutorial/introduction.html#strings ), to mark that backslashes are not to be read as special characters in the string literal, but passed down verbatim. That's [generally recommended in regexes](https://docs.python.org/3.5/library/re.html#module-re).

core/src/test/java/org/mskcc/cbio/portal/scripts/TestImportTreatmentData.java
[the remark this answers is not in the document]
pvannierop: Ok, `updateInfo` mechanism was removed.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
pvannierop: I do not see how my IDE is configured for this. I have removed this change.

core/src/main/scripts/importer/validateData.py
pvannierop: @fedde-s 
But the special characters of `"^[><]\s*" should be interpreted as special characters, not?

The regex was updated to `"^[><]\s*"`.

core/src/main/scripts/importer/cbioportalImporter.py
pvannierop: Change was reverted.

core/src/main/scripts/importer/metaImport.py
pvannierop: Change was reverted.

core/src/main/scripts/importer/validateData.py
pvannierop: Comment was no longer needed. Deleted.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
+        GsvaWiseFileValidator.prior_validated_sample_ids = sample_ids
+
+    @classmethod
+    def _get_message_features_do_not_match(cls):
+        return "Gene sets column in score and p-value file are not equal."
+
+
+
+class TreatmentWiseFileValidator(MultipleDataFileValidator, metaclass=ABCMeta):
+    """Groups multiple treatment response files from a study to ebsure consistency.
```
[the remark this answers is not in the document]
pvannierop: Corrected.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3336,16 +3514,90 @@ class GsvaPvalueValidator(GsvaWiseFileValidator):
...
+    # (3) NA cell value is allowed; means treatment was not tested on a sample
+    # (4) Is an empty cell value allowed? (meaning treatment was not tested on a sample)
+    #
+    # Warnings for concentrations:
+    # (1) Cell contains a value without decimals and is not prependend by ">"; value appears to be truncated but lacks ">" truncation indicator
+    def checkValue(self, value, col_index):
+        """Check a value in a sample column."""
+
+        # value is not defined
+        if value == "":
```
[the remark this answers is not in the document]
pvannierop: Updated to defend against white space-only cells:

>        # value is not defined
>        stripped_value = value.strip()
>        if stripped_value == "":
>            ...
>            return
>
>        # 'NA' is an allowed value. No further validations apply.
>        if stripped_value == 'NA':
>            return

core/src/main/scripts/importer/validateData.py
pvannierop: Removed.

core/src/main/scripts/importer/validateData.py
pvannierop: Changed to:
> def index_treatment_data(json_data,
>                         id_field='treatmentId'):

Also updated:
>def index_geneset_id_list(json_data,
>                         id_field='genesetId'):

core/src/main/scripts/importer/validateData.py
pvannierop: Comment removed.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
-    REQUIRED_HEADERS = ['geneset_id']
     def __init__(self, *args, **kwargs):
-        super(GsvaWiseFileValidator, self).__init__(*args, **kwargs)
-        self.geneset_ids = []
+        super(MultipleDataFileValidator, self).__init__(*args, **kwargs)
+        self.feature_ids = []
+
+    @classmethod
+    @abstractmethod
+    def _get_prior_validated_header(cls):
```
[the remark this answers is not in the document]
pvannierop: Changed for all occurences.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
+                    self.logger.warning(
+                    "Name " + fileTreatmentName + " for treatment `" + treatmentId
+                    + "` is different from name `"+ dbTreatmentName + "` present in the cBioPortal database. "
+                    + "Treatment names in cBioPortal always reflect treatment names in the last imported study. ",
+                    extra={'line_number': self.line_number,
+                            'cause': fileTreatmentName})
+
+    @classmethod
+    def _get_prior_validated_header(cls):
+        return TreatmentWiseFileValidator.prior_validated_header
```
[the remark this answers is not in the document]
pvannierop: Changed for all occurences.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
+    def _get_prior_validated_sample_ids(cls):
+        return GsvaWiseFileValidator.prior_validated_sample_ids
+
+    @classmethod
+    def _set_prior_validated_sample_ids(cls, sample_ids):
+        GsvaWiseFileValidator.prior_validated_sample_ids = sample_ids
+
+    @classmethod
+    def _get_message_features_do_not_match(cls):
+        return "Gene sets column in score and p-value file are not equal."
```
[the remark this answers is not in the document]
pvannierop: This is because error thrown by the abstract superclass is molymorphic depending on the deriving class.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3263,228 @@ def parse_q_value(self, value, column_number):
...
+    prior_validated_feature_ids = None
+    prior_validated_header = None
+    REQUIRED_HEADERS = ['geneset_id']
+
+    def __init__(self, *args, **kwargs):
+        super(GsvaWiseFileValidator, self).__init__(*args, **kwargs)
+
+    @classmethod
+    def _get_prior_validated_header(cls):
+        return GsvaWiseFileValidator.prior_validated_header
```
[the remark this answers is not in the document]
pvannierop: Changed all occurences.

core/src/main/scripts/importer/validateData.py
fedde-s: Looks good. Only the quotes should be special characters as far as the string literal is concerned, before the string literal is passed down to be parsed by the `re.sub()` function. Without the `r"`, the reader would have to think about whether `\s` means anything special in Python string literal syntax, in addition to what the resulting string means as a regex.

fedde-s: Don't forget to push the commit you just mentioned, though.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
fedde-s: I don't think that should make a difference. Doesn't retrieving a string by name as `self.spam` use the same resolution order as retrieving a function by name as `self.spam()`?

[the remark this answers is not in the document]
fedde-s: To illustrate what I mean:
```python
>>> class Breakfast:
...     spam = 'generic spam'
...     def make_toast(self):
...         return 'generic toast'
...     def cook(self):
...         return (self.spam, self.make_toast())
...
>>> class HealthyBreakfast(Breakfast):
...     spam = 'spam lite'
...     def make_toast(self):
...         return 'whole-wheat toast'
...
>>> todays_breakfast = HealthyBreakfast()
>>> todays_breakfast.cook()
('spam lite', 'whole-wheat toast')
```

core/src/main/scripts/importer/validateData.py
fedde-s: Now that I think about it again, this may actually have been a sensible idea 😳  But if you don't feel like it, I'd say it's fine not to do it right away.

fedde-s: Sorry, I must have been really out of practice here. Trying to solve this with metaclasses would be a terrible idea—those are about inheritance. If anything, the feature that could be used for that cleanup would be `@`decorators.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
fedde-s: After discussing this in person, it turns out that assigning to class-local variables in class methods did not work the way I expected: they were name-mangled for the right class, but assigned in the subclass. I'd suggest just going with `@staticmethod` for now, to remove the unused `cls` argument.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
fedde-s: As I explained in [another comment](https://github.com/cBioPortal/cbioportal/pull/5460#discussion_r257662107), I suggest just going with `@staticmethod` for now, to remove the unused `cls` argument.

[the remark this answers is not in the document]
fedde-s: Also, please just push the commit if you think it's ready to notify reviewers about.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3318,12 +3502,15 @@ class GsvaScoreValidator(GsvaWiseFileValidator):
     in R can calculate a GSVA score or GSVA-like score (such as ssGSEA) per sample per gene set.
     """
 
+    def __init__(self, *args, **kwargs):
+        super(GsvaScoreValidator, self).__init__(*args, **kwargs)
```
fedde-s: I think this initialiser method can be safely removed; the superclass method is next in the resolution order anyway.

core/src/main/scripts/importer/validateData.py
```diff
@@ -48,7 +48,7 @@
 from abc import ABCMeta, abstractmethod
 
 # configure relative imports if running as a script; see PEP 366
-if __name__ == "__main__" and (__package__ is None or __package__ is '') :
+if __name__ == "__main__" and (__package__ is None or __package__ is ''):
```
pieterlukasse: please remove `or __package__ is ''`

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
```diff
@@ -14,9 +14,8 @@
     private DaoGeneticEntity() {
     }
     
-    // TODO consider renaming this class to something similar to
-    // DaoEntity because the name does not correctly capture the 
-    // new Treatment `genetic_entity` type.
+    // TODO ...something similar to DaoEntity to reflect the fact that there are also entities that
```
pieterlukasse: replace ... with ` consider renaming this class to `

docs/File-Formats.md
pieterlukasse: please also add the validation to enforce it.

core/src/main/scripts/importer/validateData.py
pieterlukasse: nice, thanks.

docs/File-Formats.md
```diff
@@ -1271,7 +1271,7 @@ Required fields:
 cancer_study_identifier: Same value as specified in study meta file
 genetic_alteration_type: TREATMENT_RESPONSE
 datatype: RESPONSE-VALUE
-stable_id: Any unique identifier within the study
+stable_id: Any unique identifier using a combination of alphanumeric characters and _
```
pieterlukasse: please also add the validation to enforce it.

core/src/main/scripts/importer/validateData.py
```diff
@@ -450,6 +468,9 @@ def _validate_file(self):
                     self.logger.warning('Ignoring invalid column header. '
                         'Continuing with validation...')
 
+            # keep track of columns that should contain unique data
+            # init a dectionary for bookkeeping
```
fedde-s: I don't understand what this comment is about. Is it something you didn't intend to include in the PR?

core/src/main/scripts/importer/validateData.py
pvannierop: Removed.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
pvannierop: Replaced.

core/src/main/scripts/importer/validateData.py
pvannierop: Well spotted. These redundant lines were removed.

fedde-s: I still see them.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
pvannierop: I cannot get this to work in any other way. I have reverted back to class name references.

docs/File-Formats.md
pvannierop: Added.

pieterlukasse: can't find it

core/src/main/scripts/importer/validateData.py
fedde-s: Half the comment is still there.

core/src/main/scripts/importer/validateData.py
fedde-s: It wasn't removed, so please don't comment about it in past tense. A commit message would be a better place for such a comment.

fedde-s: But this line isn't a big deal, not blocking by itself.

core/src/main/scripts/importer/validateData.py
```diff
@@ -3229,87 +3260,238 @@ def parse_q_value(self, value, column_number):
...
+                ids = self.feature_ids
+                self.set_prior_validated_feature_ids(ids)
             else:
-                # Check if gene set ids are the same
-                if not prior_validated_geneset_ids == self.geneset_ids:
-                    self.logger.error(
-                        'Gene sets column in score and p-value file are not equal')
+                # Check if feature ids are the same
+                if not self.get_prior_validated_feature_ids() == self.feature_ids:
+                    log_info = self.get_message_features_do_not_match()
```
fedde-s: The abstract method you call here really needs a docstring, to tell subclassers what they should make it return. It seems to be dictionary containing only the one specific key 'cause', that maps to [the violating string found in the data file](https://github.com/cBioPortal/cbioportal/blob/v2.1.0/core/src/main/scripts/importer/cbioportal_common.py#L296)?

fedde-s:
I added a test for a scenario I believe to be broken.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
fedde-s: Well, there is one argument in favour of using a method for this: subclassers now benefit from the warnings issued by the `@abstractmethod` annotation you added.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
fedde-s: I agree; I was wrong there. You wrote these classes specifically to store state on their own class level, so they can't be more flexible. Static methods that reference that specific class level, instead of class methods, are the way to go.

core/src/main/scripts/importer/validateData.py
pvannierop: I thought this served as a separator. I have removed the divider line.

docs/File-Formats.md
pvannierop: Sorry for being unclear. Uniqueness of the stable id is ensured by existing code. No change from my side was needed.

core/src/main/scripts/importer/validateData.py
pvannierop: Docstring has been added.

core/src/main/scripts/importer/validateData.py
pvannierop: All empty constructor methods were removed. Tests run ok.

docs/File-Formats.md
pvannierop: Ok, I have added a check for the meta file stable_id in the data validator. A unit test for this was added to unit_tests_validate_data.py.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
pvannierop: I have merged proposed code changes by Fedde, thereby resolving this issue.

core/src/main/scripts/importer/validateData.py
fedde-s: The field that you documented as ‘text that details cause of the error’ is being logged as if it were a string found in the input file. But if you feel that that won't be an issue for the user experience, then it shouldn't block this pull request.

Sjoerd-van-Hagen:
@n1zea144 who can review this one?

ersinciftci:
Correct me if I'm wrong but this PR seems like mostly a copy of what was done for Genesets feature (which already partially duplicated the existing molecular data functionality), and I feel like duplicating the whole stack (web, service, persistence, model, db-scripts, importer, documentation, tests) for each new molecular alteration type is probably not the right way to do it. We should try to generalize the current code to support new molecular alteration types if they are simply just comma separated values in the `genetic_alteration` table like the existing alteration types. (By the way, next week is my last week at DFCI, so I think @n1zea144 should still take a look at this PR)

Sjoerd-van-Hagen:
@ersinciftci Hi Ersin,

I did not know you were leaving DFCI, sorry to hear that. 

We have discussed this in detail with JJ and Ben when we put out the RFC. The problem with generalizing is that you would have to put things of different types in the same table. When you do that, and you want to refer to one those things you will be limited in how you set up your foreign keys. For instance, if you mix genes, gene sets and treatments in one table how will you enforce that an RNA expression value cannot be attached to a gene set or treatment?
Generalization is good to reduce the amount of code but it can also make things complicated and harder to understand.
It would be really helpful to Ben if you could do a review of the quality of the code, as his team has a lot on their plate.

ersinciftci:
You can always enforce these kind of things in the importer code, doesn't have to be in the database, since we never go add data to the database manually. And I also don't think adding treatment specific columns to the `genetic_profile` is a good idea, it's a table for common properties of profiles. But anyways, as I said, I unfortunately won't be around for too long to defend my arguments, so it's up to @n1zea144 to give the final go-ahead. The quality of the code looks good, since, as I said, it's mostly a copy of the current code.

pieterlukasse:
@ersinciftci, @n1zea144  the part for the "comma separated values" that go into `genetic_alteration` is maybe the one that has less duplication (e.g. `ImportTabDelimData.java`). Regarding validation: we still need separate validation for genesets and treatments. I think most of the duplication can be found in the other layers of the api because we wanted to have separate endpoints. Part of this duplication could perhaps be removed in the future by merging the service classes of all `genetic_alteration` entities...but I'm not sure if this is worth the effort and whether it will make the code more maintainable. To make this possible, we would probably need to come up with a generic controller class as well and a very generic model class that only contains minimal attributes (i.e. the value and a unique identifier for the entity and the sample/study combination). This would also require refactoring of the frontend parts that currently use molecular data endpoints.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticProfile.java
```diff
@@ -100,17 +100,30 @@ public static int addGeneticProfile(GeneticProfile profile) throws DaoException
...
             pstmt.setString(1, profile.getStableId());
             pstmt.setInt(2, profile.getCancerStudyId());
             pstmt.setString(3, profile.getGeneticAlterationType().name());
             pstmt.setString(4, profile.getDatatype());
             pstmt.setString(5, profile.getProfileName());
             pstmt.setString(6, profile.getProfileDescription());
             pstmt.setBoolean(7, profile.showProfileInAnalysisTab());
+
+            // `pivot_threshold_value` and `value_sort_order` fields are treatment data 
+            // specific. These fields are set to null when not present in profile object.
```
n1zea144: There is no other place to put these?

pieterlukasse:
> treatment "metadata" data and response data is rolled into the same file, is that right?
@n1zea144: yes, that's right. We wanted to simplify this here and avoid a separate loader process just for the `treatment` lookup table. So treatment "metadata" and the measurements themselves are found in the same data file (not to be confused with the meta_ file which contains the profile metadata).

core/src/main/java/org/mskcc/cbio/portal/dao/DaoTreatment.java
```diff
@@ -0,0 +1,272 @@
...
+
+            // for insert statements the number of affected records is returned
+            // this requires the RETURN_GENERATED_KEYS to be set for insert statements.
+            int switchGetGeneratedKeys = PreparedStatement.NO_GENERATED_KEYS;
+            if (action == SqlAction.INSERT)
+                switchGetGeneratedKeys = PreparedStatement.RETURN_GENERATED_KEYS;
+
+            pstmt = con.prepareStatement(statement, switchGetGeneratedKeys);
+            for (int i = 0; i < keys.length; i++)
+                pstmt.setString(i+1, keys[i]);
```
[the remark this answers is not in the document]
pvannierop: Ah yes, well spotted. Apparently mysql (or jbdc) did not complain about this type mismatch. I have updated the code to pass the geneticEntityId as an int.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportProfileData.java
```diff
@@ -87,6 +87,13 @@ public void run() {
             } else if (geneticProfile.getGeneticAlterationType() == GeneticAlterationType.FUSION) {
                 ImportFusionData importer = new ImportFusionData(dataFile, geneticProfile.getGeneticProfileId(), genePanel);
                 importer.importData();
+            } else if (geneticProfile.getGeneticAlterationType() == GeneticAlterationType.TREATMENT_RESPONSE) {
+
+                // add all missing Treatments to the database
+                ImportTreatments.importData(dataFile);
```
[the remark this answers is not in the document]
pvannierop: The close coupling of the treatment meta data and treatment response data is a design feature. Treatments are very different from real `genetic entities` in the sense that there is no central point of truth of what are valid treatments. We envisioned that treatments may be different between studies. We did not want to have to repopulate the database when a new study is added that contains (some) different treatments. Instead, the treatment identifiers + meta data is loaded from the data file at moment of import. So yes, in the current implementation the importing of the meta data is necessarily tied to the import of the response data because treatments are too fluid to be part of any seed database.

n1zea144:
Sorry for the delay in responding @pvannierop, @pieterlukasse .  I was out of the lab last week and still catching up on events.  I appreciate your effort to  avoid a separate loader process for treatment lookup data.  This same topic came up here at MSK as we have an effort to support "mutational signature" data which is similar in representation (non-gene/matrix/scalar).

We have an idea which involves extending the genetic_entity table to capture metadata for entities like treatments/mutational signatures so that any new none-gene/matrix/scalar type can be introduced with little or no code changes required (I think we will always need to add to GeneticAlterationType.java).  More on this below.

Since this PR introduces the concept of combining metadata data with scalar measurements, I think its essential that we agree on a data file-format before pipelines are built that lock us into a particular format.

In a discussion with @jjgao and @sheridancbio.  Two ideas were floated - the first is that all metadata fields placed into a file that contains measurements be prefixed with a special label, like "META:".  This would allow a data file to contain an arbitrary number of metadata fields that can be processed by a single ImportTabDelimData.java class.  The second id is stuff all metadata into a single column in the data file in JSON format.

Combined with changes to the data file-format, as I eluded to earlier, we are thinking about extending the genetic_entity table to include a column containing a metadata in a simple json format (mainly name/value pairs).  We can expose this data through a new endpoint which serves genetic_entity metadata based on a genetic_entity_type or ID.

Would be interested in your thoughts.  Happy to discuss this idea further here or a call.

@pvannierop Since metadata and measurements are combined in the file, do we need to include ImportTreatments.java or files like data_treatment_ic50.txt?  Perhaps these are artifacts of this PRs development process?  Can we clean up some of these no longer required files (there are 94 files changed in this PR).

pieterlukasse:
@n1zea144 @jjgao I like the idea of extending the `genetic_entity` table. I believe we did discuss this model in the past (perhaps during geneset feature design), but would be good to plan a call to talk over it again. Then we can also discuss the data format, and whether we should fix it in this PR.

n1zea144:
@pieterlukasse @jjgao Will we all be on steering call tomorrow?  How about we discuss after that?

n1zea144:
@pieterlukasse @jjgao How about on Monday at 11am?

docs/File-Formats.md
```diff
@@ -1253,10 +1254,61 @@ The cells contain the p-value for the GSVA score: A real number, between 0.0 and
...
+
+The first column must be named `treatment_id` and contains unique identifiers for the treatments. The `treatment_id` column can be followed by optional unique `name`, `description` and `url` columns. The other columns are sample columns; an additional column for each sample in the dataset using the sample id as the column header.
+
+The cells contain treatment response values that can be postive and negative real numbers including 0. When no response value is measured for a sample/treatment-pair the value should be "NA". Empty cell are not allowed. Numerical values van have a "<" or ">" prefix to indicate that the real value is smaller or larger than the stated response value, respectivley. Example with 3 treatments and 3 samples:
+
+<table>
+<thead><tr><th>treatment_id</th><th>name</th><th>description</th><th>url</th><th>1321N1_CENTRAL_NERVOUS_SYSTEM</th><th>22RV1_PROSTATE</th><th>42MGBA_CENTRAL_NERVOUS_SYSTEM</th></tr></thead>
```
pieterlukasse: change `description` to `META:description`
change `treatment_id` to `stable_id`

docs/File-Formats.md
```diff
@@ -1253,10 +1254,61 @@ The cells contain the p-value for the GSVA score: A real number, between 0.0 and
...
+
+### Treatment response meta file
+The meta file will be similar to meta files of other genetic profiles, such as mRNA expression. For treatment response data `RESPONSE-VALUE` is used as `datatype`. Any number of treatment response parameters can be loaded by means of different data files. For treatment meta files values for `stable_id` are not pre-registered in the cBioPortal implementation, but can be freely chosen by the user. Requirement is that each treatment meta file is assigned an unique `stable_id`.
+
+Required fields: 
+```
+cancer_study_identifier: Same value as specified in study meta file
+genetic_alteration_type: TREATMENT_RESPONSE
+datatype: RESPONSE-VALUE
```
pieterlukasse: change to `LIMIT-VALUE`

docs/File-Formats.md
```diff
@@ -1253,10 +1254,61 @@ The cells contain the p-value for the GSVA score: A real number, between 0.0 and
...
+
+### Treatment response meta file
+The meta file will be similar to meta files of other genetic profiles, such as mRNA expression. For treatment response data `RESPONSE-VALUE` is used as `datatype`. Any number of treatment response parameters can be loaded by means of different data files. For treatment meta files values for `stable_id` are not pre-registered in the cBioPortal implementation, but can be freely chosen by the user. Requirement is that each treatment meta file is assigned an unique `stable_id`.
+
+Required fields: 
+```
+cancer_study_identifier: Same value as specified in study meta file
+genetic_alteration_type: TREATMENT_RESPONSE
```
pieterlukasse: change to `GENERIC_ASSAY`

docs/File-Formats.md
pieterlukasse: change `name` to `META:name`?
change `url` to `META:url`

pvannierop: 'META:' -tags were added to all meta fields. Validator and importer code were modified accordingly.

docs/File-Formats.md
pvannierop: LIMIT-VALUE data type implemented.

docs/File-Formats.md
pvannierop: GENERIC_ASSAY alteration type implemented. Validator and importer code were modified accordingly.

pvannierop:
@n1zea144 @pieterlukasse I have updated the PR with implementation of the GENERIC_ASSAY alteration type for treatment response data. Can you please check and comment?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -123,8 +124,8 @@ public void importData(int numLines) throws IOException, DaoException {
                                 && geneticProfile.getGeneticAlterationType() == GeneticAlterationType.GENESET_SCORE
                                 && parts[0].equalsIgnoreCase("geneset_id");
         boolean treatmentProfile = geneticProfile!=null
```
pieterlukasse: should we also rename `treatmentProfile` to `genericAssayProfile`?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -148,7 +149,7 @@ public void importData(int numLines) throws IOException, DaoException {
                 }
             } else if (treatmentProfile) {
                 if (treatmentIdIndex == -1) {
-                    throw new RuntimeException("Error: the following column should be present for treatment response data: treatment_id");
+                    throw new RuntimeException("Error: the following column should be present for treatment response data: stable_id");
```
pieterlukasse: maybe replace `present for treatment response data` with `present for this type of data`?

docs/File-Formats.md
```diff
@@ -1263,20 +1263,20 @@ The cells contain the p-value for the GSVA score: A real number, between 0.0 and
...
 
 ### Treatment response meta file
-The meta file will be similar to meta files of other genetic profiles, such as mRNA expression. For treatment response data `RESPONSE-VALUE` is used as `datatype`. Any number of treatment response parameters can be loaded by means of different data files. For treatment meta files values for `stable_id` are not pre-registered in the cBioPortal implementation, but can be freely chosen by the user. Requirement is that each treatment meta file is assigned an unique `stable_id`.
+The meta file will be similar to meta files of other genetic profiles, such as mRNA expression. For treatment response data `LIMIT-VALUE` is used as `datatype`. The `LIMIT-VALUE` is validated to contain any continuous number optionally prefixed with a '>' or '<' threshold symbol. One or more treatment response data files can be loaded by means of different data files. For treatment meta files values for `stable_id` are not pre-registered in the cBioPortal implementation, but can be freely chosen by the user. Requirement is that each treatment meta file is assigned an unique `stable_id` by the user. This requirement is validated by the data validation scripts.
```
pieterlukasse: just realized that it is quite confusing to have `stable_id` in meta file and also as part of the "meta fields". These are two different stable_ids....maybe they should have different names? Maybe call the new one `entity_stable_id`?

docs/File-Formats.md
```diff
@@ -1296,18 +1296,21 @@ pivot_threshold_value: 0.1
...
+
+### Treatment response data file
+The data file will be a simple tab separated format, similar to the expression data file: each sample is a column, each tested treatment (a compound or combination thereof) is a row, each cell contains treatment response values for that treatment x sample combination.
 
-The first column must be named `treatment_id` and contains unique identifiers for the treatments. The `treatment_id` column can be followed by optional unique `name`, `description` and `url` columns. The other columns are sample columns; an additional column for each sample in the dataset using the sample id as the column header.
+The first column must be named `stable_id` and contains unique identifiers for the treatments. The `stable_id` column can be followed by optional `META:name`, `META:description` and `META:url` columns. The other columns are sample columns; an additional column for each sample in the dataset using the sample id as the column header.
```
pieterlukasse: see my previous comment about `stable_id`

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
pvannierop: We agreed not to start renaming all the references to treatments in the current implementation. I have updated the reference to `treatmentProfile` in this file.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
pvannierop: Message was updated.

docs/File-Formats.md
pvannierop: All occurences of `stable_id` were renamed to `entity_stable_id`.

docs/File-Formats.md
pvannierop: All occurences of `stable_id` were renamed to `entity_stable_id`.

Sjoerd-van-Hagen:
@n1zea144 Hi Ben,

I think this one is ready to be merged. Will you do the honours?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportProfileData.java
```diff
@@ -87,6 +87,13 @@ public void run() {
             } else if (geneticProfile.getGeneticAlterationType() == GeneticAlterationType.FUSION) {
                 ImportFusionData importer = new ImportFusionData(dataFile, geneticProfile.getGeneticProfileId(), genePanel);
                 importer.importData();
+            } else if (geneticProfile.getGeneticAlterationType() == GeneticAlterationType.GENERIC_ASSAY) {
+
+                // add all missing `genetic_entities` for this assay to the database
+                ImportGenericAssayEntity.importData(dataFile);
+                
+                ImportTabDelimData genericAssayProfileImporter = new ImportTabDelimData(dataFile, geneticProfile.getTargetLine(), geneticProfile.getGeneticProfileId(), genePanel);
```
n1zea144: Not suggesting you do it now, but did you consider importing generic assay data in same process as importing genetic assay entity?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
```diff
@@ -14,10 +14,15 @@
...
     
+    // TODO: consider renaming this class to something something similar to DaoEntity 
+    // to reflect the fact that there are also entities that are "not genetic", like
+    // for example the new entity of type "TREATMENT".
+    // see: https://github.com/cBioPortal/cbioportal/pull/5460#discussion_r250148672
     public static enum EntityTypes
     {
         GENE,
         GENESET,
+        TREATMENT,
```
n1zea144: Do we need TREATMENT type here?  Can it be replaced with GENERIC_ENTITY (I question the need for this enum, but thats a different discussion).

n1zea144:
Hi @Sjoerd-van-Hagen @pvannierop A few more comments/questions...

db-scripts/src/main/resources/migration.sql
```diff
@@ -599,3 +599,25 @@ CREATE TABLE `data_access_tokens` (
...
+##version: 2.9.1
+CREATE TABLE `treatment` (
+  `ID` INT(11) NOT NULL auto_increment,
+  `STABLE_ID` VARCHAR(45) NOT NULL UNIQUE,
+  `NAME` VARCHAR(45) NOT NULL,
+  `DESCRIPTION` VARCHAR(200) NOT NULL,
+  `LINKOUT_URL` VARCHAR(400) NOT NULL,
+  `GENETIC_ENTITY_ID` INT NOT NULL,
+  PRIMARY KEY (`ID`),
+  UNIQUE INDEX `TREATMENT_GENETIC_ENTITY_ID_UNIQUE` (`GENETIC_ENTITY_ID` ASC),
```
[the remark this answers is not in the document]
pvannierop: Index added.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportProfileData.java
pvannierop: Yes, but we decided to divide responsibility of for different tables/data types to different importer scripts. Its is a matter of taste I guess. I like that looking from the outside it is clear that the Entities are loaded, followed by import of the profile data. Creating a dedicated Importer class that does both for generic assay data is possible in the future and can be discussed.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticEntity.java
pvannierop: No, this can also be renamed to GENERIC_ASSAY (I did). Please be aware that we did not removed all references to the concept of treatments from the existing code. This we envisioned to be future work associated with introduction of a generic _genetic_entity_ database structure.

pvannierop:
@n1zea144 I updated the code based on your feedback. Can you please review?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportProfileData.java
n1zea144: @dippindots is working on implementing proper support for generic_assay_support.  As part of this effort, he will refactor this code (after PR is merged) into a general solution.  We've had a few conversations at MSK about the approach to take, but haven't discussed in detail how the import process would change.  If you would like to be included in this discussion, let us know (just assumed you had other things to do).

n1zea144:
@Sjoerd-van-Hagen @pvannierop I'm ready to merge.  There are a few unresolved conversations with @fedde-s and @pieterlukasse.  @pieterlukasse comments appear to be resolved, not sure about @fedde-s .  Can you resolve them?

core/src/main/scripts/importer/validateData.py
pvannierop: I have removed the (optional) 'cause' element. This part of the code compares the lists of identifiers between two files and returns an error when not identical. Since this can be an issue of order or contents. Since it will be rather complicated to find the error I decided to remove the cause component.

fedde-s: Sure, should be fine; I said that it should block this pull request, so I'll resolve this thread.

inodb:
@n1zea144 the travis error is the flaky token test already existing in release-3.0.0 branch

inodb:
@n1zea144 	@pvannierop I see that there was some discussion about the `pivot_threshold` and `sort_order` fields in the `genetic_profile` table, but I can't really find what was decided in the end. Since they were added as columns to `genetic_profile` I assume they generalize to all `genetic_profile`s in some way? Could someone elaborate on that? Thanks!

Sjoerd-van-Hagen:
@inodb Both of them are useful in a more general sense. The pivot_threshold can help to discriminate between high and low values, while the sort order specified a default sort order. In some cases, low values are interesting, in others the high values. This could also apply to other measurements.

inodb:
Thanks @Sjoerd-van-Hagen !

PR #5511, opened 2018-12-20
enlarge cancer_study pmid field width

sheridancbio:
- update unit tests with a comma separated string
- adjust validator to allow longer pmid field width

# What? Why?
Fix # .

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


core/src/test/resources/seed_mini.sql
```diff
@@ -97,7 +97,7 @@ INSERT INTO "type_of_cancer" ("TYPE_OF_CANCER_ID","NAME","CLINICAL_TRIAL_KEYWORD
 
 -- cancer_study
 INSERT INTO "cancer_study" ("CANCER_STUDY_ID", "CANCER_STUDY_IDENTIFIER", "TYPE_OF_CANCER_ID", "NAME", "SHORT_NAME", "DESCRIPTION", "PUBLIC", "PMID", "CITATION", "GROUPS") 
-VALUES (1,'study_tcga_pub','brca','Breast Invasive Carcinoma (TCGA, Nature 2012)','BRCA (TCGA)','<a href=\"http://cancergenome.nih.gov/\">The Cancer Genome Atlas (TCGA)</a> Breast Invasive Carcinoma project. 825 cases.<br><i>Nature 2012.</i> <a href=\"http://tcga-data.nci.nih.gov/tcga/\">Raw data via the TCGA Data Portal</a>.',1,'23000897','TCGA, Nature 2012','SU2C-PI3K;PUBLIC;GDAC');
+VALUES (1,'study_tcga_pub','brca','Breast Invasive Carcinoma (TCGA, Nature 2012)','BRCA (TCGA)','<a href=\"http://cancergenome.nih.gov/\">The Cancer Genome Atlas (TCGA)</a> Breast Invasive Carcinoma project. 825 cases.<br><i>Nature 2012.</i> <a href=\"http://tcga-data.nci.nih.gov/tcga/\">Raw data via the TCGA Data Portal</a>.',1,'23000897,23092874','TCGA, Nature 2012','SU2C-PI3K;PUBLIC;GDAC');
```
jjgao: 23092874 doesn't seem to be associated to the TCGA study but if it's just for testing purposes, it's ok.

sheridancbio: It is only for testing, but I will switch it to 26451490, which is the 2015 tcga brca cell paper.

sheridancbio:
After discussion with @jjgao, we chose not to represent the list of pubmed ids as a list, but to keep the pmid data member as a simple string of comma separated values. This was because our use case allows for the insertion of the comma separated values directly into the http link we generate for the user on the datasets page. Following that link results in pubmed presenting a set of referenced papers in the query results output. So the frontend would not need to parse this string.

Also we discussed renaming the field to "PMIDS", but did not want to pay the cost of propagating that change to the meta_study.txt file format, which would require migrating all existing cancer studies in our repositories and causing disruptions to other users of cbioportal. So the database field and meta_study.txt field remain "PMID"/"pmid".

jjgao:
@sandertan @pieterlukasse is the validator part ready for this?

sandertan:
@jjgao Apologies for not responding sooner. I'll modify the validator tomorrow to check for comma separated values in this field.

sandertan:
Validator PR in https://github.com/sheridancbio/cbioportal/pull/6.

sheridancbio:
@jjgao just back from vacation today. I see the changes sent by @sandertan in https://github.com/sheridancbio/cbioportal/pull/6 ... I'll review that and work on merging this into the cbioportal RC branch soon.

sheridancbio:
@jjgao @sandertan : when importing the PMID line from meta_study.txt, should whitespace in the line be silently purged/dropped? Or should whitespace go in to the database as provided in the file? Or should whitespace be considered an error in the input file?  My guess is that we can silently drop whitespace from the value of this field ... and that we probably want to test the length of the field after whitespace has been removed.

jjgao:
@sheridancbio maybe we should not allow whitespace, so it has to be comma-delimited? @sandertan could you enforce that?

sheridancbio:
thanks @jjgao ... I think I will merge the code from @sandertan now into my PR, then make the modifications to prohibit whitespace in the value. Then @sandertan (and anyone else) can review my changes before the merge to RC

sheridancbio:
@sandertan I have rebased this PR in front of RC and added a validator test prohibiting whitespace in the PMID field value. I have not run the unit tests yet however. If you are able, please take a look.

docs/File-Formats.md
```diff
@@ -37,7 +37,7 @@ This file contains metadata about the cancer study. The file contains the follow
 3. **name**: The name of the cancer study, e.g., "Breast Cancer (Jones Lab 2013)".
 4. **description**: A description of the cancer study, e.g., "Comprehensive profiling of 103 breast cancer samples. Generated by the Jones Lab 2013". This description may contain one or more URLs to relevant information.
 5. **citation (Optional)**: A relevant citation, e.g., "TCGA, Nature 2012".
-6. **pmid (Optional)**: A relevant pubmed id. If used, the field citation has to be filled, too.
+6. **pmid (Optional)**: One or more relevant pubmed ids (comma separated). If used, the field citation has to be filled, too.
```
sandertan: I think it would be nice if we add here that white spaces are not allowed, like`(comma separated, without embedded whitespaces)`.

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -2274,6 +2274,27 @@ def test_exceed_maximum_length_meta_attribute_value(self):
...
 
+    def test_invalid_pmid_values(self):
+        """Test to check whether the validator throws an error for invalid PMID values in meta_study.txt."""
+        self.logger.setLevel(logging.ERROR)
+        validateData.process_metadata_files(
+            'test_data/meta_study/invalid_pmid_values',
+            PORTAL_INSTANCE,
+            self.logger, False, False)
+
+        # expecting two errors about invalid PMID
```
sandertan: ` # expecting two errors ` ->  `# expecting three errors `

docs/File-Formats.md
sheridancbio: done

core/src/test/scripts/unit_tests_validate_data.py
sheridancbio: done

sheridancbio:
code rebased and squashed to single commit

PR #5521, opened 2018-12-24
update OQL documentation

tmazor:


tmazor:
@adamabeshouse I updated the table based on your very good point in slack that it was confusing - let me know what you think of this version.

Also, I noticed that the OQL header line isn't showing up on production. Could you take a look and see if you can figure out why that is?

PR #5295, opened 2018-11-20
Add endpoint to retrieve study tags from multiple studies in one request

oplantalech:
Related to #4563, add an endpoint which retrieves all study tags from a list of study ids.

PR #5604, opened 2019-01-15
Remove endpoint that retrieves all tags at once

oplantalech:
Remove endpoint added in #5295  because it is not used.


Sjoerd-van-Hagen:
This is probably useful, even if not (yet) used in the front end. For instance, one could pull the metadata tags from all studies into a Python script to do some meta analysis or querying.

Sjoerd-van-Hagen:
So let's close this PR.

oplantalech:
@Sjoerd-van-Hagen I see. Ok, I'll close this PR.

Issue #5603, opened 2019-01-15
API 500 in study view


zhx828:
After fixing issue(https://github.com/cBioPortal/cbioportal/issues/5600), I notice the `api/cna-genes/fetch` returns 500 after selecting `Male`

Reproduce:
- Go here: http://www.cbioportal.org/study?id=paad_icgc,paad_qcmg_uq_2016,paad_tcga_pan_can_atlas_2018
- Select Sex = Male
- Take a look the Network

jjgao:
@zhx828 is it creating issues for frontend? Do we need to fix it soon?

zhx828:
@jjgao we need to fix it soon. The CNA table does not get any information.

jjgao:
@ersinciftci do you think you'll have time to handle this?

ersinciftci:
@jjgao I'll try to look at it tomorrow

PR #5628, opened 2019-01-18
Fix cna-genes and mutated-genes when there is no related profile in t…

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/5603

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/DiscreteCopyNumberServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/DiscreteCopyNumberServiceImpl.java
@@ -16,6 +16,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/cbioportal/service/impl/DiscreteCopyNumberServiceImpl.java
```diff
@@ -146,12 +146,17 @@ public BaseMeta fetchMetaDiscreteCopyNumbersInMolecularProfile(String molecularP
...
-        
-        List<CopyNumberCountByGene> result =  discreteCopyNumberRepository
-            .getSampleCountInMultipleMolecularProfiles(molecularProfileIds, sampleIds, entrezGeneIds, alterations);
-        
-        if (includeFrequency) {
-            geneFrequencyCalculator.calculate(molecularProfileIds, sampleIds, result);
+
+        List<CopyNumberCountByGene> result;
+        if (molecularProfileIds.isEmpty()) {
+            result = new ArrayList<>();
```
onursumer: wouldn't it be better to use `Collections.emptyList()` here?

Issue #4345, opened 2018-06-06
Database schema is incompatible with utf8, the default in MySQL 8


fedde-s:
As found out in #2478, the [`mutation_event`](https://github.com/cBioPortal/cbioportal/blob/v1.13.2/db-scripts/src/main/resources/cgds.sql#L395) table contains a unique key too large to be accepted by MySQL when using the UTF-8 character set.

Since this is the [default encoding](https://dev.mysql.com/doc/refman/8.0/en/charset-applications.html) in the current release of MySQL, I have documented a recommendation to stick to MySQL 5.7 in #4333, along with a hint at why in case users do want to tweak the configuration.

However, I feel that we should adjust the schema to support UTF-8 and MySQL 8.x out of the box.

jjgao:
Thanks, @fedde-s.

@khzhu you mentioned that you are using MySQL8. Would you mind sending your changes as a pull request?

cc'ing @n1zea144 and @sheridancbio

jjgao:
Thank you, @khzhu

sheridancbio:
Just catching up with this issue again. It sounds like we are moving towards reducing the tumor_seq_allele field width to 255 characters (which is something we did in our production databases, and diverges from the schema specified in cgds.sql). I think this is probably fine, but I think the reason the length of this field was originally raised above 255 was to capture some very long insertion / deletion events that we were getting through sequencing. I think @angelicaochoa made these adjustments in response to some missing or truncated mutation events we were processing.

 I guess in conjunction with this reduction to 255 we should ask the question : what do we do when a mutation event is imported that has an insertion of more than 255 nucleotides? We could make this part of the study validator ... that no mutation event contains an indel with more than this limit of nucleotides reported. Maybe an insert above a certain length gets converted into a structural variant? Or we can store very long insertions in a separate table perhaps?  We probably should not be silently truncating or discarding such events.  This discussion should probably be in a separate issue. cc: @n1zea144

fedde-s:
I briefly considered moving the field to a separate two-column table with a numeric primary key. That way the composite key could include the (much smaller) foreign key, and uniqueness of the string itself could be ensured by a single-column constraint, if that makes a difference. Your ideas about structural variants and involving the validator sound promising too though!

fedde-s:
[...]

PR #5658, opened 2019-01-28
Drop UNIQUE KEY from mutation_event. Set ALLELE fields to TEXT

angelicaochoa:

# What? Why?
In an effort to resolve data truncation exceptions thrown by MySQL during the import of `mutation_event` records where either the reference or tumor alleles are greater than the current text limits (255 char), all allele fields in `mutation_event` and `mutation` are now `TEXT`.

Partially resolves issue #4345

- Alter allele columns in `mutation` and `mutation_event` types to `TEXT`.
- Added script that sanity checks that table `mutation_event` still satisfies unique constraint that is being dropped in the database schema.

# Checks
- [ ] Runs on Heroku.


db-scripts/src/main/resources/migration.sql
```diff
@@ -542,12 +542,12 @@ UPDATE `info` SET `DB_SCHEMA_VERSION`="2.7.2";
...
 ##version: 2.7.3
 DELETE FROM `clinical_attribute_meta` WHERE clinical_attribute_meta.`ATTR_ID` = 'SAMPLE_COUNT';
-INSERT INTO `clinical_attribute_meta` SELECT 'SAMPLE_COUNT', 'Number of Samples Per Patient', 
+INSERT INTO `clinical_attribute_meta` SELECT 'SAMPLE_COUNT', 'Number of Samples Per Patient',
 'Number of Samples Per Patient', 'STRING', 1, '1', patient.`CANCER_STUDY_ID` FROM patient
 GROUP BY patient.`CANCER_STUDY_ID`;
 
 DELETE FROM `clinical_patient` WHERE clinical_patient.`ATTR_ID` = 'SAMPLE_COUNT';
-INSERT INTO `clinical_patient` SELECT patient.`INTERNAL_ID`, 'SAMPLE_COUNT', COUNT(*) FROM sample 
+INSERT INTO `clinical_patient` SELECT patient.`INTERNAL_ID`, 'SAMPLE_COUNT', COUNT(*) FROM sample
```
sheridancbio: I've been tempted to do the trailing whitespace cleanup myself .. I've looked over the whitespace cleanup you have done here and don't think there would be any functional effect on the earlier migration steps, so I'll approve with these included.

db-scripts/src/main/resources/migration.sql
```diff
@@ -567,3 +567,17 @@ UPDATE `info` SET `DB_SCHEMA_VERSION`="2.7.4";
 ##version: 2.8.0
 ALTER TABLE `cancer_study` MODIFY COLUMN `PMID` varchar(1024) DEFAULT NULL;
 UPDATE `info` SET `DB_SCHEMA_VERSION`="2.8.0";
+
+##version: 2.8.1
+ALTER TABLE `mutation_event` DROP INDEX `CHR`;
```
sheridancbio: We did not explicitly name this key in cgds.sql, but mysql seems to have given it the same name as the first field (CHR). I could not find this "default key name" behavior defined in the mysql reference documentation. This is similar to the problem with our (unnamed) foreign key constraints -- it is not certain that we will be able to modify them properly later. But I think your migration code is the best that we can do without getting really "hacky" ... such as dumping the table schema to a text string and then do string parsing to locate the key and discover the index name.

fedde-s:
I don't currently feel involved enough to comment. I think @pieterlukasse has feedback though.

inodb:
@angelicaochoa Nice! One minor thing: if I read the comments of the issue this stemmed from correctly we will now rely on the importer to enforce the unique constraint. It might be nice to have some sanity check for this.

Maybe we could have some new script (could be just bash) that takes the mysql login parameters and does a simple sanity check i.e. do we indeed only have unique mutations in the mutation event table.

angelicaochoa:
@inodb hmm, I think we can add this sanity check to our daily pipeline.. but that's a script we would check into our internal repo usually. This sounds like something that should be checked into `db-scripts/src/main/scripts/`. I think we can work with that. 

cc @sheridancbio @n1zea144

angelicaochoa:
@inodb updated PR with sanity checking script

fedde-s:
Do we really need Python 2 for this?

fedde-s:
Also, how would people know to run the script? Should it be documented somewhere? Or would it be possible to make `metaImport.py` or `cBioPortalImporter.py` refuse to report a study as successfully loaded if it results in a corrupted database?

angelicaochoa:
@pieterlukasse @fedde-s ok so for now I've removed the python script. 

1) because it's in python 2 and I forgot we had moved to python 3 and 
2) because I'm thinking that this might be a larger discussion to figure out the right solution to making sure removing the unique key constraint doesn't introduce any unintended consequences. 

I've tested imports with this new change and so far no issues for myself.

After manually inserting a duplicate `mutation_event` and duplicate sample mutations in `mutation` linked to this duplicate `mutation_event` I confirmed that the same event will simple show up twice for a given sample. See image below:
 screen shot 2019-01-30 at 11 05 52 am

So the API itself doesn't break at least.

angelicaochoa:
@inodb @pieterlukasse @fedde-s 

After further consideration and some investigation I'm not sure this sanity checking is entirely necessary because the "uniqueness" of records in `mutation_event` is implicit in the code. 

`mutation_event`s are represented as instances of `ExtendedMutation.java` objects, whose hash code and `equals(...)` methods we've overridden [here](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/model/ExtendedMutation.java#L315). We are just checking / comparing the `mutation_event` details such as gene, chr, start/end position, etc. and not checking the internal id (`mutation_event.MUTATION_EVENT_ID`) at all.

When we import mutations data, we first [load all existing mutation events](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java#L109). 

It's implicit here that even if there *are* "duplicate" events stored in the `mutation_event` table, only one will exist in this map of existing mutation events. Any samples with some variant X that are imported will be associated to the same `mutation_event` A (as opposed to some samples being linked to `mutation_event` A and another to the same but "duplicate"  `mutation_event` B).

I've reduced the comparison to a simple one liner MySQL query:
```
'select (select count(*) from  mutation_event) = (select count(distinct entrez_gene_id, chr, start_position, end_position, tumor_seq_allele, protein_change, mutation_type) from mutation_event)'
```

It's not really clear at what point someone would run this or if this check would be something that executes prior to import but that's all it would be ^^^

inodb:
@angelicaochoa 	Thanks! That one line looks good!

I think having some sanity check routine which executes a bunch of those sql statements would be nice. Just to answer the question, does my database still make sense? But yeah can def be addressed in another PR

angelicaochoa:
@fedde-s @inodb Hm, so perhaps then I’ll add this simple sanity check after the import of a MAF in `ImportExtendedMutationData.java`

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -1564,4 +1564,37 @@ public static boolean hasDriverAnnotations(String _cancerStudyStableIds) throws
...
+     * If the counts do not match then SQL statement returns 0 (false), otherwise returns 1 (true).
+     * @return
+     * @throws DaoException
+     */
+    public static boolean hasDuplicateMutationEvents() throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+
+        Integer returnValue = -1; // default = -1 to indicate mismatched counts from db table
```
sheridancbio: about the comment - if we reach the end of the function and the value is still -1, it would mean that there was an exception during the running of the query (the query did not complete) and we will not know whether duplicate records exist or not. If the query does complete, this value will be either 0 (duplicate records exist) or 1 (all records are distinct).

Also, I think the name 'returnValue' is commonly used to hold a value which will be returned when the function exits. This variable holds a value returned from the SQL resultSet, so maybe a clearer name would be something like "DistinctRecordQueryResult". Alternatively, we *could* end this function with `return returnValue` .. so this function would return one of three (Integer) values : -1 (don't know), 0 (duplicated records exist), 1 (all records are distinct)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -1564,4 +1564,37 @@ public static boolean hasDriverAnnotations(String _cancerStudyStableIds) throws
...
+        try {
+            con = JdbcUtil.getDbConnection(DaoMutation.class);
+            pstmt = con.prepareStatement(
+                    "SELECT "
+                    + "(SELECT COUNT(*) FROM mutation_event) = "
+                    + "(SELECT COUNT(DISTINCT ENTREZ_GENE_ID, CHR, START_POSITION, END_POSITION, TUMOR_SEQ_ALLELE, PROTEIN_CHANGE, MUTATION_TYPE) FROM mutation_event)");
+            rs = pstmt.executeQuery();
+            while (rs.next()) {
```
sheridancbio: I don't think a loop is needed here. The output from the query must contain only a single record I think. It would probably be safe to do:
rs.next();
returnValue = rs.getInt(1);
or if we wanted to verify that only one record was present in the resultSet we could do
if (rs.next()) {
returnValue = rs.getInt(1);
}
if (rs.next()) {
returnValue = -1; // some unexpected condition led to multiple records in the output
}

But a loop is probably not harmful, since we will always be performing a single iteration.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -454,6 +454,11 @@ public void importData() throws IOException, DaoException {
         if( MySQLbulkLoader.isBulkLoad()) {
             MySQLbulkLoader.flushAll();
         }
+        // run sanity check on `mutation_event` to determine whether duplicate
+        // events were introduced during current import
+        if (DaoMutation.hasDuplicateMutationEvents()) {
+            throw new DaoException("Duplicate mutation events were introduced during this import. Aborting...");
```
sheridancbio: If duplicate mutation events were present before the import and no further duplications were introduced, we would still get this exception. Perhaps it would be more accurate to say:
"After attempting to import mutation data for this study, duplicate mutation event records were detected. Aborting..."

sheridancbio:
updates look good

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -454,6 +454,11 @@ public void importData() throws IOException, DaoException {
         if( MySQLbulkLoader.isBulkLoad()) {
             MySQLbulkLoader.flushAll();
         }
+        // run sanity check on `mutation_event` to determine whether duplicate
+        // events were introduced during current import
+        if (DaoMutation.hasDuplicateMutationEvents()) {
```
pieterlukasse: Please add a unit test for this method. E.g. add duplicate events to DB and then check if this method returns true.

db-scripts/src/main/resources/migration.sql
pieterlukasse: Will the import be much slower now? There will be many queries to `mutation_event` table, and now there is no index anymore. Can you please show a comparison of loading time for a regular study (say BRCA_TCGA) with *vs* without this index on a system that already has a reasonably large set of studies (perhaps on a copy of the public portal's DB)? Perhaps we need to add some indexes to compensate the performance loss (if any)?

ersinciftci:
Seems like my cgds.sql changes in that PR was related to persistent unit tests, but I don't particularly remember the reason. I don't feel like it would break any functionality though (hopefully).

db-scripts/src/main/resources/migration.sql
n1zea144: @pieterlukasse I would think the opposite - without having to write to an index, the import would be faster, no?

jjgao:
@pieterlukasse good point on current UQ constraint + `REFERENCE_ALLELE` and `NCBI_BUILD`, but maybe we should log it and do it in another PR, because 1) the problem already exists in the current code, 2) we will need to normalize the field. I did a quick query and found five values in our public database: GRCh37, hg19, 37, 36, NA. An issue has been created about the data issue: https://github.com/cBioPortal/datahub/issues/571.

>are you planning to shrink the mutation_event table to a much smaller table
For this, maybe we should purge the obsolete records in mutation_event after we remove one study? (This won't work if we import a study first and then delete the old one.) We could also set up process to reannotate the database with Genome Nexus from time to time. I think @inodb had some script to do that.

angelicaochoa:
@pieterlukasse I've added a unit test and integration test. Please take a look when you have a chance. Thanks!

sheridancbio:
@jjgao mentioned the topic of cleaning up obsolete mutation_event records. There is already code in production which checks (after the deletion of a study) whether there are any "orphan" mutation_event records. So long as the study is deleted using DaoCancerStudy.deleteCancerStudyByCascade() (which is now the default for us and can be used by anyone running InnoDB engine), the function purgeUnreferencedRecordsAfterDeletionOfStudy() is called which deletes all records from mutation_event and cna_event which no longer have any references to their event ids.

core/src/test/resources/data_mutations_extended_duplicate_events.txt
```diff
@@ -0,0 +1,4 @@
+#sequenced_samples: TCGA-AA-3664-01 TCGA-AA-3665-01
+Hugo_Symbol	Entrez_Gene_Id	Center	Tumor_Sample_Barcode	Verification_Status	Validation_Status	Mutation_Status	Sequencer	Chromosome	Start_position	End_position	Variant_Classification	HGVSp_Short	MA:FImpact	MA:link.MSA	MA:link.PDB
+NLRP3	114548	broad.mit.edu	TCGA-AA-3664-01	Unknown	valid	Unknown	Illumina GAIIx	chr1	245674681	245674681	Missense_Mutation	G982G	--	--	--
+NLRP3	114548	broad.mit.edu	TCGA-AA-3665-01	Unknown	valid	Unknown	Illumina GAIIx	chr1	245674681	245674681	Missense_Mutation	G982G	--	--	--
```
sheridancbio: If there are fields which are not used to distinguish mutation_event records when searching for a matching mutation_event record during import, it would add clarity to show variation in those fields here. The only field with any variation is Tumor_Sample_Barcode. If Center, Verification_Status, Validation_Status, Sequencer, ... columns showed some variation, then it would be clear looking at this test which columns determined the sharing of a mutation_event record.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -331,7 +351,37 @@ else if (gistic.getChromosome() == 10) {
...
+        geneticProfileStableIds.add("test_dup_mut_events");
+        List<Mutation> mutations = mutationMapperLegacy.getMutationsDetailed(geneticProfileStableIds,null,null,null);
+        // there are 2 identical mutation records from 2 different samples in test MAF
+        // verify that only one mutation event is associated with these samples
+        assertEquals(2, mutations.size());
+        Set<Integer> events = new HashSet<>();
+        for (Mutation mut : mutations) {
+            events.add(mut.getMutationEvent().getMutationEventId());
+        }
+        assertEquals(1, events.size());
```
sheridancbio: This test shows that mutation_event records *can* be reused. If we wanted to be more clear about when a mutation_event record should be reused and when a new distinct record should be created we would need to have several boundary cases showing when small variations allows a record to be reused or not. But at least this shows that when mutation profiles are imported, nearly identical records will share a mutation_event record (see notes below from data_mutations_extended_duplicate_events.txt for related comment)

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -331,7 +351,37 @@ else if (gistic.getChromosome() == 10) {
...
 
-    
+    /**
+     * Test to confirm that 2 identical mutation events from 2 samples still point to the same
+     * mutation event despite unique constraint on `mutation_event`.
+     * @throws IOException
+     * @throws DaoException
+     */
+    @Test
+    public void testImportExtendedMutationDataExtendedWithSharedMutationEvent() throws IOException, DaoException {
```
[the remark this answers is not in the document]
pieterlukasse: good point, my bad. @angelicaochoa please move this to `TestImportExtendedMutationData.java`

angelicaochoa:
@pieterlukasse I tested the import time of `ImportExtendedMutationsData` and there is a performance penalty without the constrain in place so we're thinking of perhaps using a partial index on this table instead. Will update when we have a plan. Thanks for bringing up that suggestion.

angelicaochoa:
@pieterlukasse @n1zea144 

MAF import times:

|         condition         |      time        |
| ------------------- | ------------ |
| w/constraint            | 17m36.005s |
|  w/o constraint        | 50m43.194s |
| w/partial constraint | 18m48.635s |

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -635,17 +636,21 @@ public static boolean hasAlleleFrequencyData (int geneticProfileId, int sampleId
             pstmt.setString(6, event.getProteinChange());
             pstmt.setString(7, event.getMutationType());
             rs = pstmt.executeQuery();
-            if (rs.next()) {
-                return extractMutationEvent(rs);
-            }
-            else {
-                return null;
+            while(rs.next()) {
+                mutationEventList.add(extractMutationEvent(rs));
```
sheridancbio: this is ok .. but there is some overhead in calling extractMutationEvent() on each returned record.
If the resultSet is pointing at a record where the TumorSeqAllele value differs from the string in the argument (event), we will not be returning it below ... so why do the construction of a model object and the lookup of a gene from the hashmap in DaoGeneOptimized? Also, here we are making a list which must be searched below. That can be avoided by putting the logic here:
String queryTumorSeqAllele = event.getTumorSeqAllele();
if (queryTumorSeqAllele == null) {
    queryTumorSeqAllele = "";
}
while(rs.next()) {
    String retrievedTumorSeqAllele = rs.getString("TUMOR_SEQ_ALLELE");
    if (retrievedTumorSeqAllele == null) {
        retrievedTumorSeqAllele = "";
    }
    if (queryTumorSeqAllele.equals(retrievedTumorSeqAllele)) {
        return extractMutationEvent(rs);
    }
}
return null;

.gitignore
```diff
@@ -98,3 +98,4 @@ portal/reactapp/*
 package.json
 .vscode/
 *.tramp_history
+db-scripts/src/main/resources/cgds-test.sql
```
sheridancbio: I'm not sure why this was added to .gitignore

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
angelicaochoa: @sheridancbio thanks, I updated to return early if a match is found now instead of searching through a list at the end.

db-scripts/src/main/resources/gen-cgds-test-schema.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+CGDS_SCHEMA_FILE=$(find . -type f -name "*cgds.sql" | grep -v target)
```
sheridancbio: I *think* it is ok to rely on maven setting the working directory to be that of the pom file, so this should work even if the actual working directory was elsewhere.
Here the cgds-test.sql file is created in the src tree ... rather than target.
After building using this script, does it result in an untracked file db-scripts/src/main/resources/cgds-test.sql? Maybe it would  be safer to instead use the target tree for constructing the cgds-test.sql ... that way it will be ignored and not accidentally checked in to the repo.

persistence/persistence-mybatis/pom.xml
```diff
@@ -38,7 +38,7 @@
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
-            <version>1.4.192</version>
+            <version>1.4.197</version>
```
sheridancbio: I guess it should not hurt to update h2

pom.xml
```diff
@@ -152,6 +152,25 @@
...
+          <!-- Generate test DB schema without key `UQ_MUTATION_EVENT_DETAILS` to accommodate H2 MySQL driver not supporting partial indexes -->
+          <plugin>
+            <artifactId>exec-maven-plugin</artifactId>
+            <groupId>org.codehaus.mojo</groupId>
+            <version>1.6.0</version>
+            <inherited>false</inherited>
+            <executions>
+              <execution>
+                <id>Generate Test DB schema</id>
+                <phase>integration-test</phase>
```
sheridancbio: Will this work for both integration-test testing and regular unit testing? .. I'm not familiar with the integration-test phase and thought that our test database is also in use during plain unit tests.

.gitignore
sheridancbio: ah .. after reading through to the end, I guess it is because it *is* left in the source tree. Maybe by creating it in the target tree, this would not be necessary here.

db-scripts/src/main/resources/cgds.sql
```diff
@@ -430,7 +430,7 @@ CREATE TABLE `mutation_event` (
   `KEYWORD` varchar(255) DEFAULT NULL COMMENT 'e.g. truncating, V200 Missense, E338del, ',
   KEY (`KEYWORD`),
   PRIMARY KEY (`MUTATION_EVENT_ID`),
-  UNIQUE (`CHR`, `START_POSITION`, `END_POSITION`, `TUMOR_SEQ_ALLELE`, `ENTREZ_GENE_ID`, `PROTEIN_CHANGE`, `MUTATION_TYPE`),
+  KEY `UQ_MUTATION_EVENT_DETAILS` (`CHR`, `START_POSITION`, `END_POSITION`, `TUMOR_SEQ_ALLELE`(255), `ENTREZ_GENE_ID`, `PROTEIN_CHANGE`, `MUTATION_TYPE`),
```
pieterlukasse: @angelicaochoa @sheridancbio  is `KEY` a synonym to `INDEX`? If yes, please use `INDEX`, as `KEY` sounds a bit like this would be a unique or primary key (which I am assuming it is not...so please also remove the `UQ_` prefix of the index name).

sheridancbio: I think we have tried to match the format of CREATE TABLE statements that are generated by mysqldump. You can see other *plain* indexes annotated with KEY in the cgds_gdac table or by doing mysqldump --no-data. (For example, mutation_event has KEY `KEYWORD` (`KEYWORD`) and KEY `ENTREZ_GENE_ID` (`ENTREZ_GENE_ID`). To keep things consistent in the cgds.sql file, I think we should either use KEY or switch every plain index to INDEX. In mysql, the two words are synonymous I think.

angelicaochoa: @pieterlukasse  I renamed the key but I agree with @sheridancbio - either we update everything or remain consistent with how the `cgds.sql` is currently set up

core/src/test/java/org/mskcc/cbio/portal/scripts/TestNormalizeExpressionLevels.java
```diff
@@ -89,8 +89,7 @@ public void testNormalizeExpressionLevels() {
 			daoGene.addGene(new CanonicalGene(6205, "RPS11"));
 			daoGene.addGene(new CanonicalGene(7157, "TP53"));
 			daoGene.addGene(new CanonicalGene(367, "AR"));
-
-			NormalizeExpressionLevels.driver(args);
+			NormalizeExpressionLevels.main(args);
```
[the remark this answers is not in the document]
angelicaochoa: @sheridancbio yes,  `NormalizeExpressionLevels` should be called with `main(...)`, not `driver(...)`

db-scripts/src/main/resources/gen-cgds-test-schema.sh
angelicaochoa: @sheridancbio I added `cgds-test.sql` to the list of files to be ignored in `.gitignore`

persistence/persistence-mybatis/pom.xml
angelicaochoa: @sheridancbio yeah it's set to update in another PR already so it can either update now or later with the other PR. Doesn't really matter but to me sooner seems better

db-scripts/src/main/resources/cgds.sql
fedde-s: They are generally synonymous, except that if it occurs in column definitions, `KEY` means `PRIMARY KEY` instead for compatibility reasons.

Source: https://dev.mysql.com/doc/refman/5.7/en/create-table.html#create-table-indexes-keys

Sjoerd-van-Hagen:
@angelicaochoa  has this change been thoroughly tested for performance degradation? Text columns are basically external to the MySQL table which is why they can take more time to retrieve. How much time does it take when you run a select query returning a large number of mutation events in the new situation, compared to the old?

Sjoerd-van-Hagen:
> @pieterlukasse
> 
> @n1zea144 brought up the idea of doing a partial unique constraint on the `TUMOR_SEQ_ALLELE` field while maintaining datatype `TEXT`.
> 
> Here are the MAF import times:
> condition 	time
> w/constraint 	17m36.005s
> w/o constraint 	50m43.194s
> w/partial constraint 	18m48.635s

Did you mix up the with (w/) and without (w/o) results? I would think that without should be fastest.

angelicaochoa:
@Sjoerd-van-Hagen I did not mix up the results from the timed MAF import. Without *any* constraint the import was _extremely_ slow. With a partial constraint there was a very small difference compared to the original constraint in place.

jjgao:
@angelicaochoa is it b/c we look up for a mutation in the mutation_event table first when importing?

angelicaochoa:
@jjgao yes it's because of that

db-scripts/src/main/resources/cgds.sql
sheridancbio: I see the difference you referenced @fedde-s ... I was not familiar with that. I've gone through cgds.sql looking for any usage of the keyword KEY in a column definition, and I didn't see any. For documentation, here are all the cases where a (non-column) KEY definition is present without being PRIMARY or UNIQUE or FOREIGN:

gene has KEY `HUGO_GENE_SYMBOL` (`HUGO_GENE_SYMBOL`)
uniprot_id_mapping (obsolete table) has KEY (`UNIPROT_ID`)
uniprot_id_mapping (obsolete table) has KEY (`UNIPROT_ACC`)
sample_profile has KEY (`SAMPLE_ID`)
mutation_event has KEY (`KEYWORD`)
mutation has KEY (`GENETIC_PROFILE_ID`,`ENTREZ_GENE_ID`)
mutation has KEY (`GENETIC_PROFILE_ID`,`SAMPLE_ID`)
mutation has KEY (`GENETIC_PROFILE_ID`)
mutation has KEY (`ENTREZ_GENE_ID`)
mutation has KEY (`SAMPLE_ID`)
mutation has KEY (`MUTATION_EVENT_ID`)
mutation_count_by_keyword has KEY (`GENETIC_PROFILE_ID`,`KEYWORD`)
drug has KEY `DRUG_NAME` (`DRUG_NAME`)
sample_cna_event has KEY (`GENETIC_PROFILE_ID`,`SAMPLE_ID`)
copy_number_seg has KEY (`CANCER_STUDY_ID`,`SAMPLE_ID`)
cosmic_mutation has KEY (`KEYWORD`)
clinical_event has KEY (`PATIENT_ID`, `EVENT_TYPE`)

If we want to refactor cgds.sql to use INDEX in place of KEY for these usages, I am not strongly opposed. I think it has been KEY because of the generated CREATE TABLE statements that result from mysqldump. It seems like a useful thing to be able to dump a database schema with mysqldump --no-data and compare the schema with cgds.sql to see how you differ from the "standard" cbioportal database schema. But perhaps skipping over these small differences by eye would not be a big deal since there are other transformations necessary as well (table ordering, removal of comments, etc)

sheridancbio:
We are inclined to merge this PR tomorrow morning (0900 EST / GMT-5) unless there is any objection. There can be a follow up PR to use the INDEX keyword in place of the KEY keyword throughout cgds.sql if desired. This would not require any changes to migration.sql or any database migration by end users.

Sjoerd-van-Hagen:
@sheridancbio I would like to see how the text columns perform compared to the varchar ones. Typically they are slower, but this effect was more pronounced on magnetic disks. Has anyone run any large select queries that would retrieve the information in those columns, and compared to the old situation?

angelicaochoa:
@Sjoerd-van-Hagen see [here](https://github.com/cBioPortal/cbioportal/pull/5658#issuecomment-461129812)

Sjoerd-van-Hagen:
@angelicaochoa those are import times with and without constraints. I assume all of them are on the new situation with the TEXT columns?

angelicaochoa:
@Sjoerd-van-Hagen if you look at the table it's the timing for `ImportExtendedMutationData` for one MAF, during which (1) a map of all existing mutation events is generated and (2) the `mutation_event` table is queried for each mutation event encountered in the MAF being imported. The size of this MAF was `273,590` records, in which there are `187,105` unique variants so `187,105`.


the table reads by the conditions tested... 

|         condition         |      time        |
| ------------------- | ------------ |
| w/constraint            | 17m36.005s |
|  w/o constraint        | 50m43.194s |
| w/partial constraint | 18m48.635s |

- **w/constraint**, meaning no changes to the database and still using `varchar(255)` as the datatype for `TUMOR_SEQ_ALLELE`, etc. 
- **w/o constraint**, meaning no constraint or key _at all_  on `mutation_event`. `TUMOR_SEQ_ALLELE` is a `TEXT` datatype
- **w/partial constraint**, meaning the change `TUMOR_SEQ_ALLELE` has a partial index of 255 chars, the partial constraint is still defined by the same fields as before (tumor_seq_allele(255), entrez_gene_id, chr, start/end_position, mutation_type, protein_change)

Sjoerd-van-Hagen:
@angelicaochoa so the performance hit is 72s or 7% in this case. That is not too bad indeed.

I do wonder what happens when you need the actual contents of `TUMOR_SEQ_ALLELE` of multiple records at once, in terms of performance.

Then again, it is the performance of the application as a whole that counts, so I guess it would be fine.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -635,17 +635,19 @@ public static boolean hasAlleleFrequencyData (int geneticProfileId, int sampleId
             pstmt.setString(6, event.getProteinChange());
             pstmt.setString(7, event.getMutationType());
             rs = pstmt.executeQuery();
-            if (rs.next()) {
-                return extractMutationEvent(rs);
-            }
-            else {
-                return null;
+            while(rs.next()) {
```
pieterlukasse: why this while? Is mysql query not returning an exact match?

db-scripts/src/main/resources/gen-cgds-test-schema.sh
pieterlukasse: please add comments to this file explaining why it is needed and when it could be removed again (e.g. when h2db adds support for partial indexes)

db-scripts/src/main/resources/cgds.sql
pieterlukasse: thanks @fedde-s , @sheridancbio , let's keep it as `KEY` for now :+1:

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
angelicaochoa: @pieterlukasse this is _just in case_ there is more than one result returned. The query looking for a match against `TUMOR_SEQ_ALLELE`, and I didn't run into any such cases even when purposefully introducing variants that overlapped the first 255 nuc's, but this is just a safety put in place I guess.

db-scripts/src/main/resources/gen-cgds-test-schema.sh
angelicaochoa: @pieterlukasse how does this sound?


```
# The purpose of this script is to create a test version of the database schema to load in:
#     `./persistence/persistence-mybatis/src/test/resources/testContextDatabase.xml`
#
# The only difference between the production version of `./db-scripts/src/main/resources/cgds.sql` and the
# version loaded for testing is the dropping of the partial index on `mutation_event`.`TUMOR_SEQ_ALLELE`.
#
# This is done to accommodate the H2 MySQL driver since the latest version (1.4.197) does not yet
# support partial indexing.
#
# When the H2 MySQL driver finally does support partial indexing, this script can be removed and the production
# version can be used for testing again.
#
# author: ochoaa
# last update: 2019/02/13
```

angelicaochoa: cc @sheridancbio ☝️

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
pieterlukasse: but the previous code returns the first match, while this one does as well. The only difference is that the new code is slightly more confusing. Can you please revert this one?

db-scripts/src/main/resources/gen-cgds-test-schema.sh
pieterlukasse: @angelicaochoa excellent. Thanks.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
angelicaochoa: @pieterlukasse sure, np 👍

sheridancbio: Perhaps this loop is to handle a subtlety that we encountered once or twice in the past. We saw cases where the TUMOR_SEQ_ALLELE string contained a non-ascii encoded character (I cannot remember whether it was unicode or utf8). There was a difference in function output between the java string comparison library and the mysql `=` operator for the where clause. If I remember correctly, the mysql `=` operator is more tolerant of difference across encodings. I cannot remember what the exact character was but it may have been that a hyphen (ASCII code 0x2D) evaluates to "equal" to alternative unicode horizontal stroke characters. One example I found now is comparisons between accented letters such ä (LATIN SMALL LETTER A WITH DIAERESIS) and unaccented ascii coded letters. In mysql the query `select 'a' = _utf8 x'c3a4';` returns 1 (true).  (_utf8 x'c3a4' is the utf representation of ä)

Because we are dropping the unique index, it is now hypothetically possible that two mutation_event records could exist with the only difference being that one has TUMOR_SEQ_ALLELE value 'tta' and the other has TUMOR_SEQ_ALLELE value 'ttä' and the select query might return both values. I know this is a contrived example but we did see one or two cases like this.
So the loop here would make sure that the correct string was linked to if somehow this situation occurred.

sheridancbio:
all change requests addressed

PR #5673, opened 2019-02-01
Create separate study view page documentation; Add simple filter description

zhx828:
To test
https://github.com/zhx828/cbioportal/blob/add-study-view-docs/docs/FAQ.md
https://github.com/zhx828/cbioportal/blob/add-study-view-docs/docs/SUMMARY.md

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -38,6 +38,8 @@
 * [Customizing your cBioPortal Instance via portal.properties](Customizing-your-instance-of-cBioPortal.md)
 * [More portal.properties Settings](portal.properties-Reference.md)
 * [Configuring Caching Behavior](Caching.md)
+* [How does the study view organize the charts](study-view/layout.md)
+* [Utilize priority setting from database to visualize charts](study-view/priority.md)
 
 ### 2.4 Integration with Other Webservices
 * [OncoKB Data Access](OncoKB-Data-Access.md)
--- a/docs/study-view/layout.md
+++ b/docs/study-view/layout.md
@@ -1,0 +1,8 @@
+## How does the study view organize the charts
+Study view page is fully responsive, it will try to fit as many charts as possible based on the browser's width and height.
+
+The layout of a chart is determined mainly based on priority. Higher priority will promote chart closer to the left-top. 
+
+In order to improve the layout, we added a layout algorithm layer. The study view page is using grid layout. All charts will be put into 2-dimensional systems. For example, pie chart, by default, takes 1 block and bar chart uses two blocks. All charts will be placed from left to right, top to bottom. In order to prevent misalignment, we promote small charts to fit into the space. 
+
+For logged-in(authenticated) users, charts layout is saved to users profile i.e, whenever user tries to re-visits the same url, previously saved layout will be loaded.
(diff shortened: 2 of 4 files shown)


docs/study-view/simple-filter.md
```diff
@@ -0,0 +1,11 @@
+## Simple Filter
+
+The study view support pre-filtered values of one attribute in URL.
+
+### How to
+Two parameters should be specified in the URL:
+
+* filterAttributeId: the clinical attribute id (we only accept one attribute id). Please refer to clinical_attribute_meta table, ATTR_ID column
```
jjgao: We need to find a way to expose this. Maybe we can have it in the mouse over of chart header?

jjgao:
I think we should address the issue about exposing attribute id on study view page - maybe in a separate PR.

I think we need a better place to put the documentation about how to construct an URL (because this information is for end users and our docs are more for technical people). For example, we have a section about linking to cbioportal here: https://www.cbioportal.org/webAPI, but I think maybe we should now have a separate page for this. Alternatively, we could put it in FAQ.

jjgao:
- [] Let’s move the simple-filter to FAQ. 
- [] Please also expose the attribute ID on the study view interface, e.g. a column in Add Chart (hidden by default).

zhx828:
@jjgao @inodb could you take another look? The second item from JJ's list will be in separate PR.

jjgao:
@zhx828 I made some change to the FAQ section.

docs/study-view/priority.md
```diff
@@ -1,5 +1,5 @@
 # Study View Customization with Priority Data
-Example of study view in public portal: https://www.cbioportal.org/study?id=acc_tcga,lgg_tcga#summary
+Example of study view in public portal: http://www.cbioportal.org/study?id=acc_tcga,lgg_tcga#summary
```
jjgao: why switch back to http here?

zhx828: @jjgao probably an issue when I did the rebase. Changed back to https

docs/SUMMARY.md
```diff
@@ -36,6 +36,8 @@
 ### 2.3 Customization 
 * [Customizing your cBioPortal Instance via portal.properties](Customizing-your-instance-of-cBioPortal.md)
 * [More portal.properties Settings](portal.properties-Reference.md)
+* [How does the page layout the charts](study-view/layout.md)
```
inodb: @zhx828 maybe better:	`How does the study view organize the charts`

inodb: layout is a noun, so either `lay out` as a verb but that meaning is slightly different

docs/study-view/layout.md
```diff
@@ -0,0 +1,6 @@
+## How does the page layout the charts
```
inodb: same here

docs/study-view/layout.md
```diff
@@ -0,0 +1,6 @@
...
+Study view page is fully responsive, it will try to fit more charts in the same height as many as possible base on the browser width.
```
inodb: *based

inodb: `it will try to fit as many charts as possible based on the browser's width and height`

jjgao:
Not why this wasn't merged? @zhx828

zhx828:
@jjgao i think at then, I hesitated to include it because the filters hasn't been finalized. But I think we could now.

zhx828:
@kalletlak study view has been updated significantly since I wrote this. Do you think you can provide more insights on how this document can be updated?

PR #5682, opened 2019-02-05
Change realm name in the Keycloak documentation

oplantalech:
_demo_ is not a very appropriate name for the Keycloak realm name for cBioPortal. At the Hyve we usually use _cbio_ which is much self-explanatory, so I propose to change the name in the documentation.


khzhu:
agree! **_cbio_** is more specific than **_demo_**. 
Or shall we leave the name of realm to users themselves to decide. Instead, say something like suggested realm name...

oplantalech:
@khzhu Let's leave _cbio_, otherwise we have to include an explanation in the properties we change, and it might just overcomplicate the steps, I prefer to keep it simple.

khzhu:
Okay. no problems.

oplantalech:
@jjgao Can you merge?

jjgao:
if possible, can we use cbioportal instead of cbio?

oplantalech:
@jjgao done. Let's merge this! :)

jjgao:
@oplantalech thanks, could you rebase?

oplantalech:
@jjgao Done!

PR #5687, opened 2019-02-05
Fix problem with AD authentication after upgrade of Spring security

arnaudceol:
After the upgrade to spring-security 4.*, authentication with AD was not working anymore. In the log message we can read that the "username" is emtpy.

To fix it, I add the username-parameter to the security configuration as suggested in https://stackoverflow.com/a/31607334

### DIFF
--- a/security/security-spring/src/main/resources/applicationContext-security.xml
+++ b/security/security-spring/src/main/resources/applicationContext-security.xml
@@ -458,7 +458,9 @@
         <http use-expressions="true" auto-config="true">
             <form-login login-page="/login.jsp"
                 login-processing-url="/j_spring_security_check" 
-                default-target-url="/index.do"
+                username-parameter="j_username" 
+                password-parameter="j_password"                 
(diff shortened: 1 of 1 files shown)


n1zea144:
Hi @arnaudceol Thanks for these changes, they look fine.  Can you rebase this PR and squash your commits into a single commit?

n1zea144:
Hi @arnaudceol Thanks for these changes, but it looks like a merge into master which changes the names of these parameters will make this PR incorrect, [login.jsp](https://github.com/cBioPortal/cbioportal/blob/master/portal/src/main/webapp/login.jsp#L176).

If you rebase your branch and update this change to reflect the new form parameter names, we could merge this PR.

n1zea144:
Closing pull request as a change in [login.jsp](https://github.com/cBioPortal/cbioportal/blob/master/portal/src/main/webapp/login.jsp#L176) makes this PR moot.

PR #5712, opened 2019-02-08
Network tab needs X-Frame-Options: SAMEORIGIN

inodb:
Better to handle this at app level instead of at server config level. Makes deployment easier

### DIFF
--- a/portal/src/main/webapp/WEB-INF/web.xml
+++ b/portal/src/main/webapp/WEB-INF/web.xml
@@ -99,7 +99,7 @@
       <filter-name>CorsFilter</filter-name>
       <url-pattern>/*</url-pattern>
     </filter-mapping>
-    
+
     <filter>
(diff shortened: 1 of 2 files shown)


alisman:
I'm confused now.  I thought we wanted to load the iframe html page from netlify (or from wherever we are hosting frontend).

inodb:
@alisman this keeps behavior as it is currently, just moves the x frame headers from having to be set at nginx/Apache level to the app itself. We can discuss more tomorrow

inodb:
Noticed it doesn't work when using social_auth, need to find a solution for that

alisman:
maybe give the jsp an exception in web.xml?   ben recently made a PR for similar

inodb:
Found it https://stackoverflow.com/questions/28647136/how-to-disable-x-frame-options-response-header-in-spring-security

Issue #4837, opened 2018-09-19
Break up with core and portal modules


jjgao:
Once we have SPA running, the next step is to completely break up with the Core and Portal modules. The goal should be that the frontend only communicates with backend by web API.

- DOWNLOAD DATA form is still using QueryBuilder
- Any other dependency to QueryBuilder
- List the servlets that we are still using in frontend
  - network.do
  - svg2pdf
  - ...
- Any other dependencies to Core?
- Any dependencies to Portal module?

PR #5722, opened 2019-02-11
whole slide viewer token generator

### LINKED ISSUE #5624: support token based authentication to MSK Slide Viewer

dippindots:
back-end implementation of whole slide viewer token generator
Partial Fix #5624 .

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -259,6 +259,8 @@ public class GlobalProperties {
     
     public static final String SETSOFGENES_LOCATION = "querypage.setsofgenes.location";
 
+    public static final String MSK_WHOLE_SLIDE_VIEWER_SECRET_KEY = "msk.whole.slide.viewer.secret.key";
+
     private static boolean showCivic;
     @Value("${show.civic:false}") // default is false
     public void setShowCivic(String property) { showCivic = Boolean.parseBoolean(property); }
@@ -1183,5 +1185,17 @@ public class GlobalProperties {
     public static String getQuerySetsOfGenes() {
         String fileName = portalProperties.getProperty(SETSOFGENES_LOCATION, null);
         return readFile(fileName);
+    }
+
+    public static String getMskWholeSlideViewerToken()
+    {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        String secretKey = portalProperties.getProperty(MSK_WHOLE_SLIDE_VIEWER_SECRET_KEY);
+        
+        if(authentication != null && authentication.isAuthenticated() && secretKey != null && !secretKey.isEmpty()) {
+            return MskWholeSlideViewerTokenGenerator.generateTokenByHmacSHA256(authentication.getName(), secretKey);
+        } else {
+            return null;
+        }
     }
 }
--- a/core/src/main/java/org/mskcc/cbio/portal/util/MskWholeSlideViewerTokenGenerator.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/MskWholeSlideViewerTokenGenerator.java
@@ -1,0 +1,33 @@
+
+package org.mskcc.cbio.portal.util;
+
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.binary.Base64;
+
+public class MskWholeSlideViewerTokenGenerator {
+	// ~ Static fields
+	// ========================================================================================================
+
+	public static final String PORTAL_NAME = "cbioportal";
+
+	// ~ Methods
+	// ========================================================================================================
+
+	public static String generateTokenByHmacSHA256(String username, String secretKey) {
+        try {
+            String timeStamp = String.valueOf(System.currentTimeMillis());
+            String message = PORTAL_NAME + "?" + "user=" + username + "&t=" + timeStamp;
+
+            Mac sha256_HMAC = Mac.getInstance("HmacSHA256");
+            SecretKeySpec secret_key = new SecretKeySpec(secretKey.getBytes(), "HmacSHA256");
+            sha256_HMAC.init(secret_key);
+            String hash = Base64.encodeBase64String(sha256_HMAC.doFinal(message.getBytes()));
+
+            return hash;
+        }
+        catch (Exception e) {
+            return null;
+        }
+	}
+}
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -259,6 +259,8 @@
     
     public static final String SETSOFGENES_LOCATION = "querypage.setsofgenes.location";
 
+    public static final String WHOLE_SLIDE_VIEWER_SECRET_KEY = "whole_slide_viewer_secret_key";
```
alisman: i believe this should use dot notation.  we convert to underscores when writing these to javascript (jsp)

core/src/main/java/org/mskcc/cbio/portal/util/WholeSlideViewerTokenGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+public class WholeSlideViewerTokenGenerator {
+	// ~ Static fields
+	// ========================================================================================================
+
+	public static final String PORTAL_NAME = "cbioportal";
+
+	// ~ Methods
+	// ========================================================================================================
+
+	public static String generateToken(String username, String secretKey) {
```
alisman: i would give this function a more descriptive name

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -1184,4 +1186,15 @@ public static String getQuerySetsOfGenes() {
         String fileName = portalProperties.getProperty(SETSOFGENES_LOCATION, null);
         return readFile(fileName);
     }
+
+    public static String getWholeSlideViewerToken()
+    {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        
+        if(authentication != null && authentication.isAuthenticated()) {
```
alisman: in most cases there will be no secret key, so you should put a check to only do this when a secret key exists.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
dippindots: Changed to "viewer.key"

core/src/main/java/org/mskcc/cbio/portal/util/WholeSlideViewerTokenGenerator.java
dippindots: is generateTokenByHmacSHA256 better?

jjgao:
@dippindots could you ask Peter to review the code too?

dippindots:
@jjgao Yes, I will invite Peter to review this pr.

jetic83:
I reviewed it, and it looks good to me (authentication string seems correct, encryption as well)
Thanks,
 Peter

dippindots:
@jetic83 Thank you, Peter!

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -259,6 +259,8 @@
     
     public static final String SETSOFGENES_LOCATION = "querypage.setsofgenes.location";
 
+    public static final String WHOLE_SLIDE_VIEWER_SECRET_KEY = "viewer.key";
```
jjgao: `viewer.key` seems too generic. Maybe change it to "msk.whole.slide.viewer.secret.key". also change change the variable to `MSK_ WHOLE_SLIDE_VIEWER_SECRET_KEY`.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -1184,4 +1186,16 @@ public static String getQuerySetsOfGenes() {
         String fileName = portalProperties.getProperty(SETSOFGENES_LOCATION, null);
         return readFile(fileName);
     }
+
+    public static String getWholeSlideViewerToken()
+    {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        String secretKey = portalProperties.getProperty(WHOLE_SLIDE_VIEWER_SECRET_KEY);
+        
+        if(authentication != null && authentication.isAuthenticated() && secretKey != null) {
```
jjgao: Is it possible that secretKey is empty?

core/src/main/java/org/mskcc/cbio/portal/util/WholeSlideViewerTokenGenerator.java
```diff
@@ -0,0 +1,33 @@
+
+package org.mskcc.cbio.portal.util;
+
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.binary.Base64;
+
+public class WholeSlideViewerTokenGenerator {
```
jjgao: change to `MskWholeSlideViewerTokenGenerator `

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -1184,4 +1186,16 @@ public static String getQuerySetsOfGenes() {
         String fileName = portalProperties.getProperty(SETSOFGENES_LOCATION, null);
         return readFile(fileName);
     }
+
+    public static String getWholeSlideViewerToken()
```
jjgao: change to getMskWholeSlideViewerToken

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
dippindots: @jjgao I think it will be empty sometime because of this only works for MSK portal, but for the public portal, it may don't exist.

jjgao: then maybe also check `!secretKey.isEmpty()`?

sheridancbio:
This PR added token checking functionality into the core, which is soon to be dropped. Also, there is business logic here which was located in the overloaded GlobalProperties class. Maybe this new functionality is short term and is meant to be dropped when we drop the core? Otherwise we are going to need to find and relocate this code when the core is removed from the code base. In general, I think we should have a policy of encapsulating any new additions to the core in separate classes whenever possible ... to make it easier to find and transfer the functionality later.

alisman:
GlobalProperties is heavily used by the config service, which is responsible for piping configuration to frontend.   We will need to preserve this in some form going forward, core or not.  Ditto the index.jsp.  And probably also the login jsp.

jjgao:
@sheridancbio could you create an issue for this? If you are investigating dependencies to core and portal modules, it would be a good idea to add them to https://github.com/cBioPortal/cbioportal/issues/4837. Thanks.

PR #5734, opened 2019-02-12
Add ssl prop for db connection

averyniceday:
changes to allow imports to AWS DB with SSL
specific for imports which use the businessDataSource

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
@@ -2,6 +2,7 @@ package org.mskcc.cbio.portal.dao;
 
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.mskcc.cbio.portal.util.DatabaseProperties;
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Data source that self-initializes based on cBioPortal configuration.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -442,6 +442,11 @@ public static String getProperty(String property)
 		return (portalProperties.containsKey(property)) ? portalProperties.getProperty(property) : null;
 	}
 
+    public static String getProperty(String property, String defaultProperty)
+    {
+        return (portalProperties.containsKey(property)) ? portalProperties.getProperty(property, defaultProperty) : null;
```
angelicaochoa: @averyniceday this method already exists... [see here](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java#L344)

averyniceday: @angelicaochoa that's a method inside the private class here: https://github.com/cBioPortal/cbioportal/pull/5734/files#diff-d5234aefb7508dcc8a8f17cb9684d495R331

so we need a passthrough to it similar to what's defined here: https://github.com/cBioPortal/cbioportal/pull/5734/files#diff-d5234aefb7508dcc8a8f17cb9684d495R440

angelicaochoa: @averyniceday okay so since this is a required property now in the properties then maybe we should just treat it the same as the others so we can avoid updating `GlobalProperties`

angelicaochoa: just use this instead:
```
dbProperties.setDbUseSSL(GlobalProperties.getProperty("db.use_ssl")); 
```

angelicaochoa: or if anything, if we really don't want potentially null value then check for it in `DBProperties` instead like in the getter or in `JdbcDataSource` when constructing the url

angelicaochoa: @averyniceday basically I want to avoid making a change to `GlobalProperties` that would only be used once, especially since we plan to eventually move to a full spring application and would be dropping `GlobalProperties` and injecting properties the spring way

PR #5763, opened 2019-02-19
release-2.1.0 news

jjgao:


schultzn:
Here is a suggestion on how to slightly re-order and re-word the news.


New feature: Copy-Number Segments tab on the Study View page using igv.js v2 <https://github.com/igvteam/igv.js> - see example <http://www.cbioportal.org/study?id=ucec_tcga_pub&tab=cnSegments>
Improved Copy-Number Segments tab on the Results page 
New feature: OncoKB and Cancer Hotspots tracks in the Mutations tab on the Results page

PR #5784, opened 2019-02-25
add timestamp to the msk slide viewer token information

dippindots:
The back-end should return the token generation time together with the token. The front-end url should have the same time when we generate the token.
Partial fix https://github.com/cBioPortal/cbioportal/issues/5624.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/MskWholeSlideViewerTokenGenerator.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/MskWholeSlideViewerTokenGenerator.java
@@ -1,32 +1,35 @@
 
 package org.mskcc.cbio.portal.util;
-
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 import org.apache.commons.codec.binary.Base64;
 
 public class MskWholeSlideViewerTokenGenerator {
 	// ~ Static fields
-	// ========================================================================================================
-
-	public static final String PORTAL_NAME = "cbioportal";
+    // ========================================================================================================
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/MskWholeSlideViewerTokenGenerator.java
```diff
@@ -9,25 +9,23 @@
...
 
             Mac sha256_HMAC = Mac.getInstance("HmacSHA256");
             SecretKeySpec secret_key = new SecretKeySpec(secretKey.getBytes(), "HmacSHA256");
             sha256_HMAC.init(secret_key);
-            String hash = Base64.encodeBase64String(sha256_HMAC.doFinal(message.getBytes()));
-
-            return hash;
+            hash = Base64.encodeBase64String(sha256_HMAC.doFinal(message.getBytes()));
         }
         catch (Exception e) {
```
inodb: @dippindots maybe good to log this error?

core/src/main/java/org/mskcc/cbio/portal/util/MskWholeSlideViewerTokenGenerator.java
```diff
@@ -9,25 +9,23 @@
...
 
+    public String hash = null;
+    public String timeStamp = String.valueOf(System.currentTimeMillis());
 	public static final String PORTAL_NAME = "cbioportal";
 
 	// ~ Methods
 	// ========================================================================================================
 
-	public static String generateTokenByHmacSHA256(String username, String secretKey) {
+	public void generateTokenByHmacSHA256(String username, String secretKey) {
```
inodb: @dionnezaal why does the signature of this method have to change? Any specific reason?

PR #5787, opened 2019-02-26
Add netlify as another place for loading frontend

inodb:
Similar to Heroku, but use Netlify. Netlify builds frontend artifacts
automatically, so it will save some time having to wait for heroku deployments
to build. The instances don't sleep either, so this makes the testing
experience a little smoother.


inodb:
Related: https://github.com/cBioPortal/cbioportal-frontend/pull/2070

PR #5795, opened 2019-02-27
Fix the properties to set defaults for Oncoprint mutation annotations

oplantalech:
The current properties `oncoprint.custom_driver_annotation.default`, `oncoprint.custom_driver_tiers_annotation.default` and `oncoprint.oncokb_hotspots.default` do not work as expected in the [documentation](https://github.com/cBioPortal/cbioportal/blob/master/docs/portal.properties-Reference.md). 

This PR proposes to fix this and change some of the name properties so that it is more clear what they do. All properties are boolean, if they are set to `true`, then the checkbox in the "Mutations" menu of the Oncoprint will be checked by default:

- `oncoprint.custom_driver_annotation.binary.default`: referring to the "binary" custom annotations.
- `oncoprint.custom_driver_annotation.tiers.default`: referring to the tiers custom annotations.
- `oncoprint.oncokb.default`: referring to the OncoKB annotations.
- `oncoprint.hotspots.default`: referring to Hotspots annotations.

This PR goes together with cBioPortal/cbioportal-frontend#2074.

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -201,12 +201,13 @@ ucsc.build=hg19
 oncoprint.defaultview=patient
 
 # OncoPrint driver mutation annotations
-# oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
-# oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
-# oncoprint.custom_driver_annotation.default=true
-# oncoprint.custom_driver_tiers_annotation.default=true
-# oncoprint.oncokb_hotspots.default=custom
-# oncoprint.hide_vus.default=false
+#   oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
+#   oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
+#   oncoprint.custom_driver_annotation.binary.default=true
+#   oncoprint.custom_driver_annotation.tiers.default=true
+  oncoprint.oncokb.default=true
+  oncoprint.hotspots.default=true
+#  oncoprint.hide_vus.default=true
 
 # Custom gene sets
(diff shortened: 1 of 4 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -204,12 +204,13 @@ ucsc.build=hg19
...
 # OncoPrint driver mutation annotations
-# oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
-# oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
-# oncoprint.custom_driver_annotation.default=true
-# oncoprint.custom_driver_tiers_annotation.default=true
-# oncoprint.oncokb_hotspots.default=custom
-# oncoprint.hide_vus.default=false
+#   oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
+#   oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
+#   oncoprint.custom_driver_annotation.binary.default=true
```
alisman: @jjgao @oplantalech I think we decided that we should not have defaults represented in configuration here but instead in our codebase.   it doesn't make a lot of sense to have defaults values in configuration, right?

adamabeshouse: but this is the example file @alisman

adamabeshouse: oh, now I understand what you are saying. @alisman , .default actually refers to a configuration - it is whether the binary, oncokb, etc selection are selected *by default*

Sjoerd-van-Hagen:
@alisman any comments? If not, can you approve this one?

src/main/resources/portal.properties.EXAMPLE
alisman: ok, that makes sense.  maybe we could have a better name.  i'll leave that up to you guys.

Issue #5773, opened 2019-02-21
Making session service mandatory


jjgao:
Currently, cBioPortal can be installed w/o session service. As session service is becoming an integral part of the cBioPortal (see #5771), we should consider making it mandatory.

- [x] 1. Documentation on how to set up session service
- [ ] 2. Exception handling
  - What should we do when session service is (temporary) unavailable? 
  - One simple option would be disabling all related features #5771
- [ ] ~~3. Should we consider integrating session service as part of the cBioPortal API?~~
- [ ] ~~4. Should we provide a public session service for others to use?~~

cc'ing @cBioPortal/product @pieterlukasse @Sjoerd-van-Hagen @kalletlak @priti88 @inodb @alisman @n1zea144 @sheridancbio

pieterlukasse:
Maybe it makes sense to have it as part of cbioportal core. It is probably simple and small enough to integrate it there. This would mean we will not need to complicate the cBioPortal installation. What do you think @jjgao? Or is there a good reason for keeping this as a separate service?

jjgao:
@pieterlukasse I think that may be a good idea. We can still keep the data in the mongo database (to avoid a full refactoring), and integrate the Java code or even just include the session service as a library and expose it in the web API.

@n1zea144 @mandawilson @inodb @kalletlak thoughts?

jjgao:
I have add this to the comparison feature epic. Let's do # 1&2 for this round.

@inodb I am wondering if you can help with # 1. Or maybe there is a better person?

@kalletlak and @adamabeshouse could you take care of # 2.

We can consider # 3&4 later.

inodb:
@jjgao 	I can help with that - would be nice if whoever integrates the session-service can provide at least some basic documentation on how to use it. I can help test and improve it. Maybe help with providing a docker solution. 

It might also be nice if there's a way to run session service with in memory mongo db for development.

jjgao:
> whoever integrates the session-service can provide at least some basic documentation on how to use it

@kalletlak maybe you can work with @inodb on this?

I also talked to @n1zea144. We agreed that we should not work on its integration with cBioPortal code base for this round.

kalletlak:
@jjgao sure.

@inodb session-service is well documented at https://github.com/cBioPortal/session-service/blob/master/README.md, we can start with that.

adamabeshouse:
this is done for cohort comparison, we just show an error message if the request to the session service fails

fedde-s:
[...]

PR #5796, opened 2019-02-27
Fix evaluation of sessionServiceEnabled flag

alisman:
We weren't evaluating sessionServiceEnabled properly.  An empty string (corresponding to example properties file) was interpreted as sessionService = true.

### DIFF
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -140,7 +140,8 @@
              
         obj.put("oncoprintOncoKbHotspotsDefault",enableOncoKBandHotspots);    
         
-        obj.put("sessionServiceEnabled",GlobalProperties.getSessionServiceUrl()!= "");        
+        // we don't want to expose session service url to frontend so we need this flag
+        obj.put("sessionServiceEnabled",!GlobalProperties.getSessionServiceUrl().equals(""));      
                 
         out.println(obj.toJSONString());       
                       


portal/src/main/webapp/config_service.jsp
```diff
@@ -140,7 +140,8 @@
              
         obj.put("oncoprintOncoKbHotspotsDefault",enableOncoKBandHotspots);    
         
-        obj.put("sessionServiceEnabled",GlobalProperties.getSessionServiceUrl()!= "");        
+        // we don't want to expose session service url to frontend so we need this flag
```
alisman: the session service is always access by frontend via proxy.  that's why we don't need url.

alisman:
@jjgao this is the exact use case which makes session service required now.  

1. no session service
2. very long URL 

It can't be solved anymore by POST trick because in SPA, we don't do postback to server when a query is submitted, we just change frontend routes.     In @fedde-s case, he is refreshing a submitted url, and thus running into server's url length limit. 

It seems we already reached this conclusion but i think the best way forward is just to require session service.

alisman:
@fedde-s I think we should still merge this PR as it fixes one bug.    The fact that it still fails is a separate issue.

fedde-s:
This PR does break a use case that wasn't broken before. I think we should ask someone for review to see whether that use-case is important in practice before merging this.

fedde-s:
Even if support for large custom queries should indeed be dropped, displaying a blank page without any error message seems like a step backwards right now.

fedde-s:
It may not be worth fixing this if we're going to integrate the session service even in minimal deployments (being discussed in https://github.com/cBioPortal/cbioportal/issues/5773), but as it is, large custom queries are possible when v2.0.1 behaves as if there _is_ as session service and sends a POST request, and broken after applying this fix.

alisman:
@fedde-s i'm not sure that's quite right.  The SPA doesn't send ANY request when submitting a query.  It changes the frontend route and ajax takes care of data loading.  This works because there is no frontend limit on the length of the url (or at least a very large limit). But when the new url is refreshed or shared, that's when we get into a problem.   
We could undo the SPA and bring back posting, but i think that's drastic.  I think our application clearly calls for  session service to maintain these huge inputs.   It's unfortunate that this was overlooked, though it's hard to know how commonplace this is.    

One workaround while we deal with universal session service would be to increase the server's url limit.  I'm pretty sure this is configurable.  

As for a better error message, I will look into whether we can get a better error page for this exception.

fedde-s:
Large custom queries generally come from the study view, and open in a new tab. Might that be why it worked for me as long as I didn't try to refresh?

I do feel like not being able to refresh is almost certainly less of an issue than not being able to view the results of the query at all.

alisman:
@fedde-s i see.  in that case i believe it may be working because we engineered a special solution to use local storage to pass the params from study view to new window.   so it puts the params in local storage and then restores it to url after the window is opened.  
refreshing that page will result in the server exception you see.   however, that's only on refresh.  so, if my analysis is right, this PR shouldn't affect that.   this PR WILL introduce a blank page on refresh, but it should not do so on submission from study view.  can you confirm that's true?

fedde-s:
> Large custom queries generally come from the study view, and open in a new tab. Might that be why it worked for me as long as I didn't try to refresh?

It isn't: querying from the query page works as well, and it starts with a POST request to `<root_url>/api-legacy/proxy/session/main_session`.

alisman:
the post request is an ajax request.    i'm still confused though: are you confirming that everything is working EXCEPT refresh?

portal/src/main/webapp/config_service.jsp
```diff
@@ -140,7 +140,8 @@
...
+        obj.put("sessionServiceEnabled",!GlobalProperties.getSessionServiceUrl().equals(""));
```
adamabeshouse: ahha... classic java gotcha

fedde-s:
Without this pull request, yes. Everything was working except refresh. With this pull request, everything is working for small custom queries, but larger ones crash the backend before the Oncoprint ever displays.

fedde-s:
Sorry for the confusion—my screencast was taken on v2.0.1, so without this pull request.

inodb:
@alisman @fedde-s so should we merge this?

alisman:
Now that we require all portals to have a session service, we should probably strip this out entirely?

inodb:
Makes sense

fedde-s:
Yeah, I don't think we're affected by it at The Hyve any longer either, since we started treating the session service as mandatory to deploy 2.x.x and up. I haven't personally worked on many cBioPortal deployments in the past months, so if you do want more review on whether to close this, @oplantalech may have input (or say the same thing I just did).

oplantalech:
As @fedde-s said, since versions 2.x.x we just enforce our installations to require session service.

alisman:
I'll make PR stripping this from configuration example if that hasn't already been done.  Will also make sure that don't bother passing this property through anymore to frontend

PR #5800, opened 2019-02-28
Move the legend to the bottom of the HTML validation report

oplantalech:
- Move the legend to the bottom of the HTML validation report


Sjoerd-van-Hagen:
@n1zea144 Hi Ben, can we get this merged. We want to make sure that the more important information will be at the top of the report.

PR #5806, opened 2019-03-01
Add HRRS filter to log web requests

averyniceday:
TODO:
figure out pipeline for decoding and replaying requests

# What? Why?
Adding hrrs library to log incoming web requests - potentially useful for debugging memory leak
Logging can be turned on and off by sending a PUT request to the hrrs servlet (this might not work as easily on a password protected instance of the portal)

Add hrrs servlet filter 
new property `hrrs.logging.directory` in portal.properties - decide logging output directory

TODO:
Integrate in the `replayer` hrrs tool
Workaround for turning hrrs logging on/off on password protected portal

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -36,7 +36,6 @@
         <artifactId>service</artifactId>
         <version>${project.version}</version>
     </dependency>
-
     <!-- antisamy -->
     <dependency>
       <groupId>org.owasp</groupId>
--- a/core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2019 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
+ * FOR A PARTICULAR PURPOSE. The software and documentation provided hereunder
+ * is on an "as is" basis, and Memorial Sloan-Kettering Cancer Center has no
+ * obligations to provide maintenance, support, updates, enhancements or
+ * modifications. In no event shall Memorial Sloan-Kettering Cancer Center be
+ * liable to any party for direct, indirect, special, incidental or
+ * consequential damages, including lost profits, arising out of the use of this
+ * software and its documentation, even if Memorial Sloan-Kettering Cancer
+ * Center has been advised of the possibility of such damage.
+ */
+
+/*
+ * This file is part of cBioPortal.
+ *
+ * cBioPortal is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mskcc.cbio.portal.util;
+
+import java.io.*;
+import java.util.*;
+import org.apache.commons.logging.*;
+import org.cbioportal.model.*;
+import org.springframework.beans.factory.annotation.*;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.context.annotation.Bean;
+import org.springframework.stereotype.Component;
+import com.vlkan.hrrs.servlet.base64.Base64HrrsFilter;
+import com.vlkan.rfos.RotationConfig;
+import com.vlkan.rfos.policy.DailyRotationPolicy;
+import javax.servlet.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DebugHrrsFilter extends Base64HrrsFilter {
+
+    private static RotationConfig createRotationConfig() {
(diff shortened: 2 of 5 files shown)


core/pom.xml
```diff
@@ -38,6 +38,21 @@
...
 
+      <dependency>
+        <groupId>com.vlkan.hrrs</groupId>
+        <artifactId>hrrs-servlet-filter-base64</artifactId>
+        <version>0.5</version>
+      </dependency> 
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>23.6-jre</version>
```
sheridancbio: we are currently importing guava 19.0 in the root pom.xml ... perhaps we should upgrade that rather than load two separate versions of guava

core/pom.xml
```diff
@@ -38,6 +38,21 @@
...
+      </dependency>
+      <dependency>
+        <groupId>org.apache.httpcomponents</groupId>
+        <artifactId>httpcore</artifactId>
+        <version>4.4.8</version>
```
sheridancbio: we are currently importing httpcore 4.3.2 in the root pom.xml .. perhaps we should upgrade that rather than load two separate versions of httpcore.

portal/pom.xml
```diff
@@ -15,6 +15,21 @@
   <url>http://maven.apache.org</url>
 
   <dependencies>
+      <dependency>
```
sheridancbio: if guava and httpcore are updated in the root pom, they do not need to be here

core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
```diff
@@ -0,0 +1,71 @@
+/*
```
sheridancbio: Can we relocate this class into the portal module? We hope to drop the core module, so adding classes there means that we will need to relocate them later.

averyniceday: @sheridancbio do you have a suggestion for where inside the portal module? Looking through the portal module briefly, I don't see a clear subdirectory it belongs in. 

Also I'm using the GlobalPropertiesUtil class from the core module to read in the logging directory portal property.

sheridancbio: we explored code together .. there are other filters in core that are referenced in the portal web.xml, so this is just a new sibling to those previous dependencies. All will need to be relocated.

core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
```diff
@@ -0,0 +1,71 @@
...
+                .file(file)
+                .filePattern(filePattern)
+                .policy(DailyRotationPolicy.getInstance())
+                .build();
+        return rotationConfig;
+    }
+ 
+    public DebugHrrsFilter() {
+        super(createRotationConfig());
+        this.setEnabled(true);
```
averyniceday: @inodb I think this is where I set it to turn on logging. Shouldn't be to hard to have it read from a properties file. It's been awhile so I don't completely remember but I think some examples I saw had a controller that could let you hit an endpoint that turned the logging on/off.

inodb: 👌 if u have some time at some point, would be cool to add that. Def not urgent, since we figured out the performance problems from before in another way. Maybe some explanation in portal.properties.EXAMPLE of these new properties

jjgao:
@averyniceday any update on this?

averyniceday:
@jjgao looking at it now, we're going to try to merge it in this sprint.

core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
mandawilson: @averyniceday @inodb the API endpoint is enabled in the current version (you guys probably know that).  I deployed an instance of this to test with authentication in place.  I put in the property file `hrrs.enable.logging=false`.  I then did the following:`GET http://dev-host:8080/manda-tmp/hrrs` (first I did have to login) and got: `{"enabled": false}`.  I then edited my request in Firefox to be a `PUT http://dev-host:8080/manda-tmp/hrrs?enabled=true`, and hit the GET endpoint again and could see `{"enabled": true}` (HRRS was turned on).  I verified by hitting the website and I could see data being added to `hrrs-spring-records.csv` on the disk.  If anyone can turn this on we should be careful.  HRRS saves the payload and the request, although I think there is a maximum size it will store.  You can configure your own maximum size to store and also implement a `isRequestRecordable(HttpServletRequest request)` method in order to filter requests that are recorded.  On a public portal you would not have to be authenticated to turn it on and off.  If I wanted to cause trouble I would turn this on through the API, then hit the site over and over again on a page that returned a large payload.  I think logs are set to rotate daily.  Probably all we have to do is make sure we compress them, delete them, and limit the size of them.  We might also want to document this somewhere so they know to do the same.  Can we turn the API endpoint off? @n1zea144 @sheridancbio

core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
```diff
@@ -0,0 +1,71 @@
...
+import com.vlkan.rfos.policy.DailyRotationPolicy;
+import javax.servlet.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DebugHrrsFilter extends Base64HrrsFilter {
+
+    private static RotationConfig createRotationConfig() {
+        String hrrsLoggingDirectory = GlobalProperties.getProperty("hrrs.logging.directory");
+        String file = new File(hrrsLoggingDirectory, "hrrs-spring-records.csv").getAbsolutePath();
```
n1zea144: Why set directory via property and not filename?

core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
mandawilson: `hrrs` API endpoint has been removed.

PR #5808, opened 2019-03-01
Disable Pool Prepared Statement JDBC

inodb:
We are using -Ddbconnector=jdbc on AWS in production. Before the dbcp was only
being used for the importer. It seems that enabling this when serving the
webapp results in a huge number of queries being stored in memory. We might
eventually want to make this some property that one can only enable when using
the importer.


core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
```diff
@@ -24,7 +24,9 @@ public JdbcDataSource () {
         this.setPassword(password);
         this.setUrl(url);
         //  By pooling/reusing PreparedStatements, we get a major performance gain
```
fedde-s: Not fully sure whether this comment still adds clarity

fedde-s:
I don't know enough about this to see a problem with it. The last time I touched this logic it was to be able to reuse database credentials between parts of cBioPortal that build on Spring and ones that don't (particularly entry points of the data loading pipeline for which it may be overkill), and I don't expect that to be broken by this change, at least if the core tests pass.

pieterlukasse:
hi @ino, did you guys do some tests to check whether the import scripts performance was affected? I can imagine it gets slower...but how much slower? Maybe @angelica can do some tests in the same fashion she did here: https://github.com/cBioPortal/cbioportal/pull/5658#issuecomment-462912255?

inodb:
@pieterlukasse we haven't done any testing yet re importing - i think it would be good to test this before doing a minor release

alisman:
I thought we were going to decouple configuration?

inodb:
@alisman Do you mean having a separate configuration to enable that caching for the importer? I think that would make sense if we confirm there's a big performance win when running the importer

alisman:
agreed

paulcwvandijk:
I've tested data loading with brca_tcga data (importing and re-importing) with and without the disabling Pool Prepared Statement JDBCchange. Performance was even better (25% down) with setPoolPreparedStatements set to false.

PR #5814, opened 2019-03-04
Add 11 tests for validation of gene panel and (not-)profiled data

sandertan:
# What? Why?
Gene panel validation has been added in #5602 . This PR mostly adds test data, but also contains code for unit and system tests.

- Add 10 unit tests for the new validation checks for both gene panel and (not-) profiled data
- Add 1 system test


jjgao:
@sandertan @Sjoerd-van-Hagen any plan on this?

Sjoerd-van-Hagen:
Hi @jjgao,
This one is interesting. @sandertan created these tests but he was not happy with all the duplication. It could be done in a cleaner way but we did not have the time to do so. The idea behind creating this PR was basically that someone making changes to the gene panel code would be able to run these, but they would not be part of the master branch before cleaning them up. Obviously, the cleaning up did not happen, so now we should decide whether we want to merge them, or clean them up. Perhaps @n1zea144 has an opinion on this.

PR #5858, opened 2019-03-12
Use of hashset instead of list speeds up SampleServiceImpl.processSam…

n1zea144:
Profiled with GENIE Cohort v5.0-public (~60,000 samples) on internal MSK machine - dashi dev (numbers will probably not be so dramatic going forward, but list sizes will continue to grow, so its a good change going forward).

Using list -

6BF2ACF1-E680-4AF2-A2D4-0EC57139795F

Using set -

AF9211B3-C10C-4787-B038-1F82CE556FB6


ersinciftci:
@n1zea144 Awesome! Nice catch. @jjgao This will possibly reduce Genie study view load time from around 10 seconds to around 5 seconds.

Issue #5850, opened 2019-03-10
simplified search: allow customize the study list


jjgao:
e.g. using instance-specific configuration.

Some feedback from steering meeting last Friday:

Both physical studies and virtual studies should be supported as the default studies.

For the public portal, it would be a good idea to create a TCGA PanCancer Atlas virtual study and use that for query, so that the study name and description can be properly displayed in the Results page. (currently, it's Combined Study)

inodb:
@jjgao  there are a few ways to solve this:

1. We change the session service model to include a `name` in the data part for sessions of type `main_session` (https://github.com/cBioPortal/session-service#valid-type). Similar as we have for `virtual_study`

2. We could set the title in a property of quick search config. Say e.g. it will query study x+y+z. Then on query view we can display a diff name if all those studies are being queried (instead of the usual "combined study")

1. is prolly cleaner solution

inodb:
Discussed with Hongxin - virtual studies are actually not private if u have the id. So as long as all studies in the session are public, one can access it. So instead will create a virtual study of TCGA pancan under cbioportal account. Then at some point we can show all virtual studies owned by cbioportal on the homepage.

PR #5863, opened 2019-03-14
Add quick search beta configuration and docs

inodb:
Related to: https://github.com/cBioPortal/cbioportal/issues/5850

also fixes this: https://github.com/cBioPortal/cbioportal/issues/5869

Moves `/ln?` links to be handled by the frontend


docs/portal.properties-Reference.md
```diff
@@ -50,7 +50,46 @@ skin.show_news_tab=
...
+# for studies queried when typing in a single gene use this when session
+# service is disabled: 
+quick_search.gene_query.cancer_study_list=
+quick_search.gene_query.cancer_study_list_name=
+# or this if you do have session service (title will be pulled from the
+# virtual_study):
+quick_search.gene_query.session_id=
+```
+
+One can show quick search by setting `quick_search.show=true` (default is
```
alisman: i think we should change this "enabled" but not too important

docs/portal.properties-Reference.md
```diff
@@ -50,7 +50,46 @@ skin.show_news_tab=
...
+
+```
+quick_search.gene_query.session_id=5c8a7d55e4b046111fee2296
+```
+
+When using a session id, the title will be pulled from the virtual study. Make
+sure to create a `virtual_study` with studies that everybody has access to and
+don't use a `main_session` id.
+
+**Note:** When `quick_search.gene_query.cancer_study_list` or `quick_search.gene_query.session_id` is defined it will also be used as the default study for the `/ln?=TP53:MUT` links.
```
alisman: do you think it's weird that we are embedding a default that can be used outside of quick search inside quicksearch config?  maybe we should have a default_cross_cancer_study_list that is used by ln and quick search?

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -643,19 +643,6 @@
         <url-pattern>/oranalysis.do</url-pattern>
     </servlet-mapping>
```
alisman: we should make sure that we update frontend to compatible version (meaning it can handle ln route)

docs/portal.properties-Reference.md
inodb: @alisman yeah was thinking that as well. We'll prolly need a `default_cross_cancer_session_id` config as well then

portal/src/main/webapp/WEB-INF/web.xml
inodb: updated it

PR #5885, opened 2019-03-22
Fix /api/studies/ keyword+pageSize resp size

inodb:
The primary site keyword search would not disregard duplicates when determining
how many studies to add. This fixes that. As an example: on current public
portal querying with keyword head and selecting a pageSize of 15, it would not
return a response with all 13 studies, it would return fewer.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/StudyServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/StudyServiceImpl.java
@@ -15,6 +15,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import java.util.ArrayList;
 
 @Service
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/cbioportal/service/impl/StudyServiceImpl.java
```diff
@@ -33,21 +34,24 @@
...
-                    pageSize - allStudies.size() : primarySiteMatchingStudies.size();
-                primarySiteMatchingStudies = primarySiteMatchingStudies.subList(0, toIndex);
-            }
             for (CancerStudy cancerStudy : primarySiteMatchingStudies) {
-                if (!allStudies.stream().anyMatch(c -> c.getCancerStudyIdentifier().equals(cancerStudy.getCancerStudyIdentifier()))) {
-                    allStudies.add(cancerStudy);
+                if (!allStudiesByCancerStudyIdentifier.containsKey(cancerStudy.getCancerStudyIdentifier())) {
+                    allStudiesByCancerStudyIdentifier.put(cancerStudy.getCancerStudyIdentifier(), cancerStudy);
+                }
+                if (pageSize != null && allStudiesByCancerStudyIdentifier.size() == pageSize) {
```
dippindots: Looks good to me! Not sure if the `pageSize != null` is necessary.

inodb: i believe so, in case you don't specify pageSize as parameter to the request it will be null here. We want to return all the found matches in that case, so never break 🏎

PR #5905, opened 2019-03-27
Co-expression memory optimization

ersinciftci:
It's a memory optimization for Co-expression API, instead of creating a list of millions of `GeneMolecularData` objects to calculate the co-expression, we do the calculation on the raw `GeneMolecularAlteration` objects.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
@@ -6,19 +6,30 @@ import org.apache.commons.math3.linear.RealMatrix;
 import org.apache.commons.math3.stat.correlation.SpearmansCorrelation;
 import org.cbioportal.model.CoExpression;
 import org.cbioportal.model.Gene;
-import org.cbioportal.model.GeneMolecularData;
+import org.cbioportal.model.GeneMolecularAlteration;
+import org.cbioportal.model.MolecularProfile;
+import org.cbioportal.model.Sample;
+import org.cbioportal.persistence.MolecularDataRepository;
+import org.cbioportal.persistence.SampleListRepository;
 import org.cbioportal.service.CoExpressionService;
 import org.cbioportal.service.GeneService;
 import org.cbioportal.service.MolecularDataService;
+import org.cbioportal.service.MolecularProfileService;
+import org.cbioportal.service.SampleService;
 import org.cbioportal.service.exception.MolecularProfileNotFoundException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -28,54 +39,80 @@
...
             List<String> queryValuesCopy = new ArrayList<>(queryValues);
 
             List<Integer> valuesToRemove = new ArrayList<>();
             for (int i = 0; i < queryValuesCopy.size(); i++) {
-                if (!NumberUtils.isNumber(queryValuesCopy.get(i)) || !NumberUtils.isNumber(values.get(i))) {
+                if (!NumberUtils.isNumber(queryValuesCopy.get(i)) || !NumberUtils.isNumber(values.get(i)) 
+                    || excludedIndexes.contains(i)) {
                     valuesToRemove.add(i);
                 }
             }
```
adamabeshouse: I can't add a comment below but, we can do one further optimization. Below, instead of what you have for arrays, let it be like this:

```
double[][] arrays = new double[2][valuesNumber.length];
arrays[0] = queryValuesNumber;
arrays[1] = valuesNumber;
...
SpearmansCorrelation spearmansCorrelation = new SpearmansCorrelation((new Array2DRowRealMatrix(arrays, false)).transpose())
```

It seems wasteful to duplicate that data, right?

ersinciftci: @adamabeshouse I don't know much about matrix transformation but it sounds good to me, so I just updated the PR.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -99,13 +139,10 @@
...
-            for (int i = 0; i < valuesNumber.length; i++) {
-                arrays[i][0] = queryValuesNumber[i];
-                arrays[i][1] = valuesNumber[i];
-            }
-
-            SpearmansCorrelation spearmansCorrelation = new SpearmansCorrelation(new Array2DRowRealMatrix(arrays, false));
+            double[][] arrays = new double[2][valuesNumber.length];
+            arrays[0] = queryValuesNumber;
+            arrays[1] = valuesNumber;
+            SpearmansCorrelation spearmansCorrelation = new SpearmansCorrelation((new Array2DRowRealMatrix(arrays, false)).transpose());
```
alisman: @jjgao @n1zea144 @lemccarthy  unless I'm mistaken, it would seem that at minimum we need to bring into memory all alterations for a single gene in order to populate the array we pass to SpearmansCorrelation here.   If that's true, does a cursor help us?  I think a cursor is for handling db result record by record.  But we cannot do that, right?

PR #5906, opened 2019-03-27
Add quick search news section

inodb:


docs/News.md
```diff
@@ -1,3 +1,8 @@
+## March 19, 2019
```
jjgao: just noticed the date isn't right.

inodb: lol - fixed

Issue #2901, opened 2017-08-02
DumpPortalInfo fails to export gene set metadata


fedde-s:
As documented [here](https://github.com/cBioPortal/cbioportal/blob/v1.7.3/docs/Using-the-dataset-validator.md), the data set validation script needs certain side data (external to studies) in order to test whether a study can be loaded into a specific installation without breakage or unexpected assumptions. For setups in which the validation script cannot retrieve these metadata from the web API directly (it usually runs on a different machine than the web server, and may not be able to log in to an authentication layer), cBioPortal contains [a script](https://github.com/cBioPortal/cbioportal/blob/v1.7.3/docs/Using-the-dataset-validator.md#example-4-generating-the-portal-info-folder) that generates a folder of JSON files needed by the validation script. 

Currently, this script fails to export a JSON file for gene set data, even though the validation does require it. This disables validations related to gene set data files in these setups.

fedde-s:
@pieterlukasse, do you know what might have gone wrong here? For reference, [this](https://github.com/cBioPortal/cbioportal/blob/v1.7.3/core/src/main/java/org/mskcc/cbio/portal/scripts/DumpPortalInfo.java#L112) is the Java code of the utility that should generate the JSON files from the database, just like the web API would do.

sandertan:
Also gene panels are not exported by DumpPortalInfo.

PR #5935, opened 2019-04-05
Fix to get genesets and gene panels exported via DumpPortalInfo

pieterlukasse:
# What? Why?
Fix for #2901. 

- add service layer as a dependency in core. 
- use service layer in DumpPortalInfo script to to export genesets and gene panels JSON


pieterlukasse:
@oplantalech can you please test this in the metaImport context? This should allow for extra geneset and gene panel validation in some of our projects where the authentication is turned ON.

pieterlukasse:
Adding @ersinciftci because of the new dependency to service layer being added here. Ideally we can soon move out the scripts from core at some point. So the change proposed here (in pom.xml) is a temporary one for core.

PR #5956, opened 2019-04-10
Changed how back end sends mutation and CNA data

lemccarthy:
Instead of sending a calculated frequency, the backend will now send the total number of samples profiled, and the front end will calculate it.


lemccarthy:
I've added to this commit the changes I need to do the gene panel modal.

PR #5985, opened 2019-04-18
Apply the co-exp optimization to the new code

### LINKED ISSUE #5980: Merge master to rc branch, resolving conflict in coexpressionserviceimp
hi @oplantalech, @ersinciftci made an optimization to coexpression endpoint to reduce memory consumption.  You can see the changes (now in master) here: https://github.com/cBioPortal/cbioportal/pull/5905

There is a significant conflict between these changes and your [5354](https://github.com/cBioPortal/cbioportal/pull/5354) PR to rc. 

We were wondering if you could help resolve these conflicts.  If you could merge master to rc branch you will see the conflicts to a couple files.  You could make a PR to rc that resolve these conflicts.  

You will find just one other conflict in pom.xml.  You can resolve with the rc file.

ersinciftci:
Fixes #5980

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
@@ -5,7 +5,7 @@ import org.apache.commons.math3.linear.Array2DRowRealMatrix;
 import org.apache.commons.math3.linear.RealMatrix;
 import org.apache.commons.math3.stat.correlation.SpearmansCorrelation;
 import org.cbioportal.model.Gene;
-import org.cbioportal.model.GeneMolecularAlteration;
+import org.cbioportal.model.MolecularAlteration;
 import org.cbioportal.model.GeneMolecularData;
 import org.cbioportal.model.Geneset;
 import org.cbioportal.model.GenesetMolecularData;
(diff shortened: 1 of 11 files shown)


service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -42,65 +55,171 @@
...
+            return Collections.emptyList();
+        }
+
+        return fetchCoExpressions(molecularProfileId, sampleIds, geneticEntityId, geneticEntityType, threshold);
+    }
+
     @Override
-    public List<CoExpression> getCoExpressions(String geneticEntityId, CoExpression.GeneticEntityType geneticEntityType, String sampleListId, 
-                                              String molecularProfileIdA, String molecularProfileIdB, Double threshold) throws Exception {
+    public List<CoExpression> getCoExpressions(String geneticEntityId, CoExpression.GeneticEntityType geneticEntityType,
```
alisman: this is called overloading yes? same method, two different signatures?  so this one handles different profileIds, yes?

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -42,65 +55,171 @@
...
+        Boolean isMolecularProfileBOfGenesetType = molecularProfileB.getMolecularAlterationType()
+                .equals(MolecularProfile.MolecularAlterationType.GENESET_SCORE);
         if (isMolecularProfileBOfGenesetType) {
-            molecularDataListB = genesetDataService.fetchGenesetData(molecularProfileIdB, sampleIds, null).stream().collect(Collectors.toList());;
+            molecularDataListB = genesetDataService.fetchGenesetData(molecularProfileIdB, sampleIds, null).stream()
+                    .collect(Collectors.toList());
+            ;
         } else {
-            molecularDataListB = molecularDataService.fetchMolecularData(molecularProfileIdB, sampleIds, null, "SUMMARY").stream().collect(Collectors.toList());
+            molecularDataListB = molecularDataService
```
alisman: so in this case (multiple profiles and NOT genesets, we lose our optimization, right, because we have get the molecularData (which were the original target of optimization).

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
ersinciftci: yes

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
ersinciftci: yes

PR #6006, opened 2019-04-23
microrna gene record loading after seed database

### LINKED ISSUE #5668: mirna: how to better handle it.
The fake negative entrez gene IDs give us a lot of headache. Let me document how we generate them here and see if we can solve the issue.

The mirna id file reside in the id-mapping-mirbase.txt file under reference-data in our main mercurial repo. Also uploaded [here](https://github.com/cBioPortal/cbioportal/files/2814331/id-mapping-mirbase.txt).

This is the [Java code](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportMicroRNAIDs.java#L80) to import this file.

image

Following [the code](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneOptimized.java#L167), it uses DaoGene.addGeneWithoutEntrezGeneId.
image

following [the code](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGene.java#L71) again, you’ll see getNextFakeEntrezId() is called to generate a negative ID: 

image

Note that we also have negative entrez gene IDs for phosphoproteins too, so depending on who import first, the entrez IDs are different. So basically these IDs are generated randomly. :(

One way to solve it to separate the mirna ids into a separate table and connect them with entity table. (the same with phosphoprotein)

sheridancbio:
# What? Why?
Fix #5668 ; The microRNA "genes" have been problematic. They are not genes in the fullest sense (no translation / codons) and have a separate process of biological function (double strand RNA formation, Dicer cleavage, RISC formation). Nonetheless, display of microRNA copy number alterations (reported for microRNA precursors) and microRNA expression (reported for mature forms) were supported in cBioPortal early on through the addition of constructed precursor/mature combined entries added to the gene table with negative entrez_gene_id values (arbitrarily assigned to these microRNA combination "genes" and other genetic profile subjects such as phosphoproteins).

With scheduled updating of the gene table via ncbi downloads, we had conflicts between downloaded ncbi "genes" (with valid assigned positive entrez_gene_id) which were microRNA precursor loci and the negatively numbered precursor/mature microRNA pairs loaded into the production database.

This PR moves in the direction of suppressing the positive numbered ncbi precursor loci and standardizing the use of the negatively numbered microRNA precursor/mature combinations based on the current microRNA annotations from mirbase.org. This is intended to be a short term solution. In the long term we hope to properly model microrna profiling without use of the gene table.

- updated code and instructions to allow loading of microrna "gene" records during cBioPortal deployment
- added a resource file which maps precursors to mature to combined forms
- updated code for loading microrna genes and aliases
    - dropped the call to removePreviousMicroRNARecord() because we do not want to purge any existing gene records in the database (and linked genomic profile data)
- cbio_cancer_genes have been dropped (one "gene" in this set was a microRNA, causing loading problems)
    - source file deleted
    - old JSON servlets have cancer_gene arguments rejected now, and annotate all output with cancer_gene:false instead of actual value
- adjusted test schema generation script to not find entries in .git hidden subdirectory

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


sheridancbio:
@jjgao @n1zea144 One consequence of dropping the cbio_cancer_gene file was a reduction in functionality in legacy cnaJSON and mutationsJSON servlets. Now an exception is thrown if cnaJSON is called with the argument cbio_genes_filter=true, and all output which provide a true/false indication for cbio_cancer_gene will always report false.
I don't think these servlets are in use .. unless by the R and MATLAB clients. But because there was a change in API, I thought you should be aware and approve / merge this code.
Also I was not sure which outside institutions (if any) should be involved in reviewing this. Maybe you can give guidance.

n1zea144:
Hi @sheridancbio I noticed a link to the R source code from the [R-MATLAB](http://www.cbioportal.org/rmatlab) tab.  From there you can get to the source code for these packages.  Can you do a peek at the code and see calls to these calls will be affected?

sheridancbio:
@n1zea144 I've examined the R code and Matlab code linked to from here:
https://github.com/cBioPortal/cbioportal/blob/master/docs/RMatlab.md

Every call to the cbioportal data services go through webservice.do
We looked together earlier and you determined that webservice.do has its own standalone linkage to the persistence layer and is not dependent on cnaJSON or mutationsJSON servlets.

mandawilson:
@jjgao @n1zea144 The changes in this PR are contained in the merged PRs #6288 and #6345.  We should close this PR, but @sheridancbio should probably take a look and see if there is anything he wanted that got missed.  I think the other PRs cover everything.  In any case, we definitely do not want to merge this now.

sheridancbio:
I've looked through the merged code. It basically looks fine. I see a couple of changes:
in CnaJSON.java I had added a servlet exception if a user queried with the cbio_genes_filter option ... that was removed, which is fine I guess. The end user may not notice that no filtering is being done any more. But we are dropping this entire servlet soon anyway, so it is not very important.
Otherwise, everything looks good with the merge.

PR #6054, opened 2019-05-03
Update TimelineValidator in Validator to Check 'START_DATE' Value

averyniceday:
# What? Why?
The PDX pipeline is pulling timeline records without a `START_DATE` value (invalid). This change will separate out the data validation from the software pulling data and put it with the rest of the data validator. 

Move data validation for timeline files into the validator so validation is not spread across different repositories. 

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


averyniceday:
@pieterlukasse @fedde-s if you guys have the time, could you please take a look at this PR? I'd like to add this small step to validating timeline data (this data issue recently came up in our pipelines). Thanks!

averyniceday:
@pieterlukasse updated!

fedde-s:
I'm a bit busy right now and I trust Pieter's review. No need to wait for mine!

sheridancbio:
@pieterlukasse does the added unit test satisfy your request?

core/src/main/scripts/importer/validateData.py
```diff
@@ -2872,10 +2872,24 @@ class TimelineValidator(Validator):
...
         super(TimelineValidator, self).checkLine(data)
         # TODO check the values
+        for col_index, col_name in enumerate(self.cols):
+            # treat cells beyond the end of the line as blanks,
+            # super().checkLine() has already logged an error
+            value = ''
+            if col_index < len(data):
+                value = data[col_index].strip()
+            if col_name == 'START_DATE':
+                if value.strip() in self.NULL_VALUES:
```
pieterlukasse: following up on @sheridancbio 's comment: START_DATE in DB (see https://github.com/cBioPortal/cbioportal/blob/master/db-scripts/src/main/resources/cgds.sql#L775) is `int NOT NULL`. So in case of `na`, the code would crash during importing. Perhaps better that testing `if value.strip() in self.NULL_VALUES` is to test whether it is a valid integer, and give the error if not.

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -443,6 +442,25 @@ def test_date_in_nondate_column(self):
...
+        # set level according to this test case:
+        self.logger.setLevel(logging.ERROR)
+        record_list = self.validate('data_timeline_invalid_start_date.txt',
+                                     validateData.TimelineValidator)
+        self.assertEqual(len(record_list), 2)
+        for error in record_list:
+            self.assertEqual("ERROR", error.levelname)
+            self.assertIn("Missing START_DATE", error.getMessage())
+
+
```
pieterlukasse: 👍 good test, thanks.

core/src/main/scripts/importer/validateData.py
averyniceday: @pieterlukasse sorry for the late reply, I've updated it to check that the value consists of digits

PR #6061, opened 2019-05-06
Added configuration entry for analytics report url

### LINKED ISSUE #6058: Create new config property for google analytics reporting feature
Study the way we handle google analytics key "google_analytics_profile_id"

Namely, it needs to be added to list in this jsp, which is what communicates the config data to frontend.

/Users/aaronlisman/IntelJProjects/portal/src/main/webapp/config_service.jsp

lemccarthy:
# What? Why?
Fix #6058

### DIFF
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -37,6 +37,7 @@
             "oncoprint.oncokb_hotspots.default",
             "genomenexus.url",
             "google_analytics_profile_id",
+            "analytics_report_url",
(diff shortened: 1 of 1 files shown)


portal/src/main/webapp/config_service.jsp
```diff
@@ -30,13 +30,17 @@
         String[] propNameArray = {
             "app.version",
             "app.name",
+            "dat.uuid.revoke_other_tokens",
```
alisman: looks like there's some extra stuff in this commit

PR #6063, opened 2019-05-06
Update docker docs

inodb:
Move over thehyve/cbioportal-docker to replace the existing docker
documentation

Blocked by (dbcp): https://github.com/cBioPortal/cbioportal/pull/6046


docs/docker/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+Update the seeded database schema to match the cBioPortal version
+in the image, by running the following command. Note that this will
+most likely make your database irreversibly incompatible with older
+versions of the portal code.
+
+```
+docker run --rm -it --net cbio-net \
+    -v /<path_to_config_file>/portal.properties:/cbioportal/portal.properties:ro \
+    cbioportal:update-docker-docs \
```
inodb: will change this tag once approved - after updating the docker image

docker/web-and-data/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    && pip3 install -r /cbioportal/requirements.txt
+
+# add importer scripts to PATH for easy running in containers
+RUN find /cbioportal/core/src/main/scripts/ -type f -executable \! -name '*.pl'  -print0 | xargs -0 -- ln -st /usr/local/bin
+
+# put config files in this folder if you want to override config
+ENV PORTAL_HOME=/cbioportal
+
+# TODO: fix the workdir-dependent references to '../scripts/env.pl' and do this:
+# RUN find $PWD/core/src/main/scripts/ -type f -executable \! \( -name env.pl -o -name envSimple.pl \)  -print0 | xargs -0 -- ln -st /usr/local/bin
```
khzhu: @inodb , is it possible to remove <none> image after docker build, since it was 3.28GB from my test?

khzhu: can we remove line 38,39 if is no use?

docker/web/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# this GC with JAVA_OPTS=-XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC
+FROM maven:3.5.4 as build
+COPY $PWD /cbioportal
+WORKDIR /cbioportal
+RUN mvn -DskipTests clean install
+
+FROM shipilev/openjdk-shenandoah:8
+COPY --from=build /cbioportal/portal/target/cbioportal*.war /app.war
+COPY --from=build /cbioportal/portal/target/dependency/webapp-runner.jar /webapp-runner.jar
+CMD /usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEBAPP_OPTS} /app.war
```
khzhu: can we  have a Dockerfile for just data loading? for instance have Dockerfiles: 

- web+data
- web only 
- data loading only

docs/docker/README.md
```diff
@@ -0,0 +1,184 @@
...
+docker run -d --restart=always \
+  --name=cbioDB \
+  --net=cbio-net \
+  -e MYSQL_ROOT_PASSWORD='P@ssword1' \
+  -e MYSQL_USER=cbio \
+  -e MYSQL_PASSWORD='P@ssword1' \
+  -e MYSQL_DATABASE=cbioportal \
+  -v /<path_to_save_mysql_db>/db_files/:/var/lib/mysql/ \
+  mysql:5.7
+```
```
khzhu: can we connect to existing mysql server not in the cbio-net network? if so, how?

docs/docker/README.md
```diff
@@ -0,0 +1,184 @@
...
+  --name=load-seeddb \
+  --net=cbio-net \
+  -e MYSQL_USER=cbio \
+  -e MYSQL_PASSWORD='P@ssword1' \
+  -v /<path_to_seed_database>/cgds.sql:/mnt/cgds.sql:ro \
+  -v /<path_to_seed_database>/seed-cbioportal_<genome_build>_<seed_version>.sql.gz:/mnt/seed.sql.gz:ro \
+  mysql:5.7 \
+  sh -c 'cat /mnt/cgds.sql | mysql -hcbioDB -u"$MYSQL_USER" -p"$MYSQL_PASSWORD" cbioportal \
+      && zcat /mnt/seed.sql.gz |  mysql -hcbioDB -u"$MYSQL_USER" -p"$MYSQL_PASSWORD" cbioportal'
+```
```
khzhu: do we need to run the migration script to update the database schema, since seed database might not be up-to-date?

docs/docker/README.md
khzhu: is it possible to have a docker compose doing the following three tasks?

- set up mysql server
- restore mysql from seed database
- migrate to the latest schema

fedde-s: I don't even think that that automation requires docker-compose. It's part of the entrypoint script that the MySQL container itself runs before opening the MySQL service port.

fedde-s: However, in [thehyve/cbioportal-docker](https://github.com/thehyve/cbioportal-docker#step-2---run-mysql-with-seed-database), we specifically decided that it was more user-friendly to document loading the seed as an explicit separate step.

Back when we did use the entrypoint script to load the seed database in the same command that starts the service, new users were repeatedly confused about when it would or would not load the seed database; it only does so if you mount a `/var/lib/mysql` volume that is completely empty, not if you mount, say, an unsuccessfully loaded database folder or an old database folder with no studies loaded.

docker/web-and-data/Dockerfile
```diff
@@ -0,0 +1,39 @@
+# This image includes both the web app and the data import related scripts in
+# one image. Because of historic reasons some of these scripts run java
+# artifacts under the hood. It therefore currently makes more sense to have a
+# single image supporting both instead of moving the scripts out to a separate
+# container.
+FROM maven:3.5.4 as build
+COPY $PWD /cbioportal
```
fedde-s: Do the docs for multi-stage build explicitly state that the potentially very large amount of private branch data and configuration in the `.git/` folder will _not_ be included in the final Docker image?

docker/web-and-data/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /cbioportal
+RUN mvn -DskipTests clean install
+
+FROM shipilev/openjdk-shenandoah:8
```
fedde-s: The Docker Hub documentation for this image specifically states that it concerns a nightly builds that is untested, experimental and to be used with care. I would not feel comfortable deploying it to production servers that can affect users and sensitive data. Do the official OpenJDK images, endorsed by a trademarked organisation name, not suffice?

docker/web-and-data/Dockerfile
fedde-s: And why _is_ it commented out? In [thehyve/cbioportal-docker](https://github.com/thehyve/cbioportal-docker/blob/v2.1.1+backport6012/docs/example_commands.md#importing-data), it's actively used to run the study import scripts.

fedde-s: @khzhu, the way I understand it, Docker will keep the intermediate build container only as a cache. It would be used if you download an updated version of the runtime base image (containing the Debian base system and OpenJDK runtime), and you want to build an image that combines the cBioPortal code you already compiled with the new environment. It should not be needed at buildtime, so if you don't mind having to recompile cBioPortal in such a case, you could safely remove it after the build step.

docs/docker/README.md
fedde-s: The migration script is an even more complicated problem, as [the migration process](https://github.com/cBioPortal/cbioportal/blob/v2.2.0/docs/Updating-your-cBioPortal-installation.md#running-the-migration-script) skips blocks of statements from `migration.sql` depending on which version of the schema was already present in the seed database downloaded from Datahub.

docker/web-and-data/Dockerfile
fedde-s: @khzhu, does that work for you? Do you feel that removing images should be explicitly mentioned in the documentation?

docs/docker/README.md
fedde-s: I believe you can point the hostname in portal.properties to a MySQL instance running on another machine as usual, and there are also ways to set up a hybrid approach with a container that is not isolated from the host networking stack.

However, these docs (authored and maintained by The Hyve on [thehyve/cbioportal-docker](https://github.com/thehyve/cbioportal-docker/blob/v2.1.1%2Bbackport6012/README.md#step-2---run-mysql-with-seed-database) over the past several years) assume that most users who start deploying with Docker will not want to install a MySQL server without isolation, making it a more niche use-case.

I feel that it may be more constructive to review adding documentation for it as a separate feature. Do you think that it should block integrating these docs, @khzhu?

docs/docker/README.md
khzhu: I am not using migration python script but migration sql statements when running mysql docker container. I extract the portion of sql statements needed to get the schema updated to the latest from the migration.sql file to a sql file. 
Eventually, I will use an open source database migration tool called **_Flyway_** to automate this process.

docs/docker/README.md
khzhu: @fedde-s, you might want to put one sentence or two to let users know how to connect to an existing database. We use cbioportal docker web image with AKS, but the sql servers (master/slave) are outside the AKS.

docs/docker/README.md
fedde-s: I'd suggest making a pull request (or issue report) to review and agree such things between institutes.

docker/web-and-data/Dockerfile
inodb: @fedde-s for us it doesn't b/c they don't free the memory adequately unfortunately. If you don't care about freeing up memory as much b/c you e.g. have a single machine per app then the official image suffices. Im happy to make this use an official openjdk image though. We'd be using the one in `web/` for production

docker/web-and-data/Dockerfile
khzhu: what is line 33 for?

docker/web-and-data/Dockerfile
khzhu: I vote for an official openJDK image.

docker/web-and-data/Dockerfile
khzhu: I would suggest to document those "none" images, so users know where those images come from. You might also want to include a docker command to remove dangled images.

khzhu: - why do we need python-dev and python3-dev (line 26, 27)? 
- any reasons to use mvn 3.5.4?

inodb: @fedde-s that's copy pasted from the Dockerfile in the master branch: https://github.com/thehyve/cbioportal-docker/blob/master/Dockerfile#L57-L58

inodb: also there's a `TODO` comment in there so figured it might be important to keep?

inodb: @khzhu line 33 adds scripts like `metaImport.py` to the `PATH` so you can run the documentation @fedde-s linked without having to specify a full path

docker/web/Dockerfile
inodb: @khzhu it would be nice to have a data loading only image, but note that it's not that useful b/c of how connected the data import and core module still is. You still need Java+Python. The only difference with the `web+data` one would be that there is no war file and `webapp-runner` included. These are however not that big. So it seems to me like this would only really be useful once we move the core module out and completely separate the data import code from the web code

docs/docker/README.md
inodb: @khzhu I think a docker compose file might be useful at some point, especially for developers and testing/CI to have a one command way to run the portal. But yes let's address that later in a separate PR

docker/web-and-data/Dockerfile
inodb: @fedde-s @khzhu I'll update this Dockerfile to the official openJDK image. But I do want to provide one image with Shenandoah cause that's what we are using in production. The performance difference is huge in terms of memory consumption. I could add another `web` only image if that's desired. I get the sensitive data argument but also note that the motto of shiplev is:

`"builds.shipilev.net — still more secure than npm install"`

And we are using the npm registry extensively. Ofc that's not an argument to add another vulnerability. But people/institutions can decide for themselves how much of an issue this is

docker/web-and-data/Dockerfile
inodb: good point - that's prolly why maven could correctly determine the git tags 😄

docker/web-and-data/Dockerfile
khzhu: @inodb ,does line 38,39 try to do the same thing as line 33 except for renaming env.pl?

docs/docker/README.md
```diff
@@ -0,0 +1,184 @@
...
+Make sure to at least provide the database parameters from step 1, which are
+required for the next step:
+
+```
+db.user=cbio
+db.password=P@ssword1
+db.host=cbioDB
+db.portal_db_name=cbioportal
+db.connection_string=jdbc:mysql://cbioDB/
+```
```
khzhu: please document the use of db.use_ssl option, it will need to be set to **_true_** if your mysql server use SSL. 

> db.use_ssl = true 

or use JAVA_OPTS when run docker 

> -Ddb.use_ssl=true

docker/web-and-data/Dockerfile
inodb: @fedde-s could you elaborate on this one actually -  We're copying over the `war` and `webapp-runner` from the first image specifically, so the final image will only contain that.

Im not sure what the best procedure is here. We pull a bunch of git info using mvn for https://www.cbioportal.org/api/info, so it's useful to have `.git` available during the compile step. We could potentially pass these as env variables as well

docker/web-and-data/Dockerfile
inodb: @khzhu @fedde-s The idea of the documentation updates for this PR is to get people to run the portal using docker images. We would provide these images on docker hub.

I think building your own images and changing them is another use case for developers. Ideally people deploying the portal wouldn't have to do this.

@fedde-s has written a substantial amount of documentation on how to use Docker as a developer. But I think to include it, it would need some restructuring so we can clearly separate these use cases. The `none` images point can be addressed in that documentation

fedde-s: Sounds good to me! If @khzhu agrees, can one of you phrase an issue report to track the follow-up task of refining the docs for that use-case for devs new to Docker?

fedde-s: > @fedde-s that's copy pasted from the Dockerfile in the master branch: https://github.com/thehyve/cbioportal-docker/blob/master/Dockerfile#L57-L58

Ahh, yeah, sorry, I think _I_ actually wrote that a long time ago. It's an existing inflexibility in the cBioPortal scripts: the endpoints for loading side-data (say, updating the gene table) only work if you set the working directory to `/cbioportal/core/src/main/scripts`, so I couldn't improve usability by adding them to `$PATH` with my limited knowledge of Perl.

I should report this as an issue in https://github.com/cBioPortal/cbioportal/ instead of tracking it as a TODO comment here. I believe that that would resolve @khzhu's objection.

docker/web/Dockerfile
khzhu: @inodb @fedde-s , I got data loading only image worked out. The size of image is not that much reduced (basically only excluding two jar files), but the new image is more user friendly, since all a user needs to do is to mount a volume to a docker container. Something like below:
```
docker run -it -v /Users/kzhu/Downloads/breast_msk_2018:/study kzhupm/cbio-loader
```
I agree with you both and will not complicate things here. I will open a new PR for new data loader Docker file at some point in the near future. I will also include two optimized docker files for existing web and web-and-data (Elena from Azure team helped with the optimization, the web image now is only 282MB).

inodb: @khzhu that's great! Thanks!

docs/docker/README.md
inodb: added a few sentences

docs/docker/README.md
inodb: included that

khzhu: thank you, @inodb!

docker/web/Dockerfile
fedde-s: Sounds like a great idea, even if it's just a different default command; I never thought of that!

docker/web-and-data/Dockerfile
inodb: created this one to track https://github.com/cBioPortal/cbioportal/issues/6131

inodb: i see this has been reported now: https://github.com/cBioPortal/cbioportal/issues/6123

.dockerignore
```diff
@@ -0,0 +1,8 @@
+.project
+.pydevproject
+.classpath
+.DS_Store
+target
+docker
+docs
+*/target/*
```
khzhu: can we also add .github?

khzhu:
I've done my review. The document have been greatly improved with accurate details. Well done!

.dockerignore
inodb: done

inodb:
Thanks @khzhu !! Also - most of this work is from @fedde-s, so credit goes to him 🥇

docker/web-and-data/Dockerfile
fedde-s: My worry is that `docker history --no-trunc <image name>`, `docker inspect <image name>` or `docker inspect <container name>` or similar metadata digs into the image might leak information about the contents of the `.git/` folder other than the version description that the build process needs to include in the .war file.

If we can find some docs or practical evidence implying that the information really isn't preserved in the final image, I agree that giving the ‘mvn stage’ of the Dockerfile access is much better than passing `--build-args`, so that a conventional `docker build` command just works. Otherwise, a few lines of shell script or a documented instruction to supply the build arg could imaginably be a big security improvement for some use-cases.

inodb: @fedde-s Thanks for clarifying! That is a good point

I couldn't find an answer using google so opened a StackOverflow question: https://stackoverflow.com/questions/56278325/is-it-a-security-risk-to-include-a-git-folder-in-the-build-step-of-a-docker-mul. Please help me rephrase it if you think it's unclear

The automated image builds would be done on CircleCI or Docker Hub, so I don't think there's a risk there, since it will only pull public git info. But maybe if you build it locally and push it publicly it could result in some .git folder leaking. Hopefully the StackOverflow question can give an answer to that.

I could add some warning as comments to the image referring to that question for now. Is that an ok solution?

docker/web-and-data/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# This image includes both the web app and the data import related scripts in
```
fedde-s: I cannot find what command you're using to build an image using this Dockerfile, giving it access to .dockerignore and the right folder to `COPY` in, while ignoring the other Dockerfile that's still in the root folder of the repo. I presume you're just building it from that root folder, but with an added `-f` argument to target this Dockerfile?

docs/README.md
```diff
@@ -19,10 +19,8 @@ We also maintain an active [list of RFCs (Requests for Comments)](RFC-List.md) w
 * [Hardware Requirements](Hardware-Requirements.md)
 
 #### 2.1.1 Docker
-* [Docker Prerequisites](Docker-Prerequisites.md)
-* [Deploy using Docker](Deploy-Using-Docker.md)
-* [Uninstall Docker cBioPortal](Uninstall-Docker-cBioPortal.md)
-* [Import a Study Using Docker](Import-Study-Using-Docker.md)
+* [Docker Prerequisites](docker/prerequisites.md)
+* [Deploy with Docker](docker/deploy.md)
```
fedde-s: I don't see files by these names. Do you mean `docker/README.md` and `docker/example_commands.md`?

docker/web-and-data/Dockerfile
fedde-s: Yes! That would definitely do, and resolve my objection to merging this :+1:

docs/README.md
inodb: good catch - i messed up with rebasing

docker/web-and-data/Dockerfile
inodb: yes that's how i was using it - im not super familiar with docker hub pipelines or circle ci pushing of images so have to figure out what directory structure will work for that. Might have to change it slightly, but i can add the command im currently using to this PR, maybe as a comment on this image

docker/web-and-data/Dockerfile
inodb: updated

docs/README.md
inodb: fixed

docker/web-and-data/Dockerfile
inodb: added some usage instructions

fedde-s:
Have the other four reviewers (@n1zea144, @sheridancbio, @kalletlak and @ao508) seen any potential issues that may have to be addressed before merging, or would it be okay to retract the requests for their review?

fedde-s:
It may be worth merging the new commits from https://github.com/thehyve/cbioportal-docker since you started, if you haven't already. Althrough there should only be a few, and some of the improvements are about documentation that's no longer necessary now that the Dockerfile is going to be maintained in the same repository as the cBioPortal build POM itself.

PR #6067, opened 2019-05-07
Update validator: remove portal_info_dir argument (path to JSON files)

rmadupuri:
# What? Why?
Switching to use cbioportal API url instead of using static JSON files for validation.

- Remove the portal_info_dir argument

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -4016,11 +4016,6 @@ def interface(args=None):
                                    help='URL to cBioPortal server. You can '
                                         'set this if your URL is not '
                                         'http://localhost:8080/cbioportal')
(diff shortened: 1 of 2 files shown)


inodb:
@rmadupuri Thanks! Why should this option be removed completely? One can decide not to use the parameter right?

rmadupuri:
Hi @inodb, this PR is because we decided to completely remove the static dependencies of the validator code.  But I am closing the PR for two reasons:
1. We reverted back to using the json files since studies fail when using portal API due to gene symbol mapping to multiple entrez ids. (We will be cleaning this soon)
2. Validator unit tests are based on the json files. so removing the attribute fails the PR.

Issue #6099, opened 2019-05-14
-Ddat.method has no default value


inodb:
Running without adding `-Ddat.method` gives runtime error:

```
SEVERE: Exception sending context initialized event to listener instance]of class [org.springframework.web.context.ContextLoaderListener]
java.lang.IllegalArgumentException: Could not resolve placeholder 'dat.method' in value "${dat.method}"
```

inodb:
Lol github doesn't understand `"needs to be fixed"` vs `"fixed"`

ao508:
@n1zea144 @inodb when you ran this, did your properties contain `dat.method` with no value set for it or did `dat.method` not exist at all in your properties

PR #6108, opened 2019-05-15
Data Access Token access for api legacy and webservice dot do

mandawilson:
Includes fix for #6099


sheridancbio:
Canceled outstanding review requests (per request in slack from reviewees)

PR #6166, opened 2019-05-24
change ObjectMapper in InvolvedCancerStudyExtractorInterceptor 

inodb:
Don't use Autowiring/Bean

Just instantiate an ObjectMapper to use resolves "bean not found of type" error message in tomcat catalina.out

This is an essential fix for v2.2.1 (mskcc tomcat won't work otherwise)


sheridancbio:
@alisman We are familiar with this code change The problem was that the type of CustomObjectMapper is castable to ObjectMapper, so there was a bean wiring ambiguity that cropped up somehow late in the game.
Feel free to merge this if you approve, or if you would rather not review, that would be ok too .. just let us know.

PR #6188, opened 2019-05-29
update FAQ with group comparison

tmazor:
The group comparison questions are mostly just links to the tutorial slides, but at least it is now part of the FAQ.

If this looks good, it should probably be merged to the 3.0.0 release branch so it gets released with group comparison.


jjgao:
We should rebase this to release-3.0.0 but there is currently a conflict.

PR #6194, opened 2019-05-30
Merge master to release 3.0.0

inodb:


web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -137,6 +137,64 @@
         return true;
     }
 
+    private boolean extractAttributesFromClinicalDataBinCountFilter(HttpServletRequest request) {
```
sheridancbio: Isn't this function already defined on line 443? This seems like a second copy

web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -137,6 +137,64 @@
...
+                request.setAttribute("involvedCancerStudies", cancerStudyIdCollection);
+            }
+        } catch (Exception e) {
+            LOG.error("exception thrown during extraction of clinicalDataBinCountFilter: " + e);
+            return false;
+        }
+        return true;
+    }
+
+    private boolean extractAttributesFromClinicalDataCountFilter(HttpServletRequest request) {
```
sheridancbio: Isn't this function already defined on 463? This is a second copy

web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -137,6 +137,64 @@
...
+                request.setAttribute("involvedCancerStudies", cancerStudyIdCollection);
+            }
+        } catch (Exception e) {
+            LOG.error("exception thrown during extraction of clinicalDataBinCountFilter: " + e);
+            return false;
+        }
+        return true;
+    }
+
+    private boolean extractAttributesFromStudyViewFilter(HttpServletRequest request) {
```
sheridancbio: Another copy is on line 504 .. I do notice a bug in both of these ... the Log.error() reports clinicalDataBinCountFilter, but this (and the previous function) are parsing different types. But a bugfix should be in a separate PR.

inodb:
@sheridancbio 	this is fixed now

web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
inodb: didn't fix this - we can address it in a separate PR as suggested

sheridancbio:
Looks good @inodb 👍

PR #6234, opened 2019-06-06
Validator: do not execute tests that require API data if no_portal_checks is enabled

oplantalech:
The validator has the option `-n` (`--no_portal_checks`) which skips tests requiring information from the cBioPortal API. In #5602 the Gene Panel option is fixed, but the gene panel tests requiring API information are not escaped when using the `-n` option, therefore this option cannot be used now. This small PR fixes this bug.

You can test the fix by going to `cbioportal/core/src/main/importer` and execute the command:
```
./validateData.py -s ../../../test/scripts/test_data/study_es_0/ -n
```


jjgao:
@rmadupuri maybe you can help review this? cc'inng @n1zea144

oplantalech:
@jjgao can you merge?

PR #6264, opened 2019-06-12
Add DB property to enable/disable prepared statement pooling

ao508:
Make pooling prepared statements setting for db a portal property. 

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
@@ -15,6 +15,7 @@ public class JdbcDataSource extends BasicDataSource {
         String password = dbProperties.getDbPassword();
         String database = dbProperties.getDbName();
         String useSSL = (!StringUtils.isBlank(dbProperties.getDbUseSSL())) ? dbProperties.getDbUseSSL() : "false";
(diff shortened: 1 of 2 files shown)


ao508:
@sheridancbio resolved

pieterlukasse:
just out of curiosity: is anyone deploying this set to `true`? This was the cause of the memory leak, right?

inodb:
Yeah I think we should be very explicit here that it can only be enabled when running the importer. It might therefore be better to make it a command line parameter for the importer if possible instead

pieterlukasse:
Given @inodb's comment, maybe we should close this PR to avoid confusion?

sheridancbio:
Withdrawing request for review from @inodb with permission

sheridancbio:
@inodb I believe this could already operate as a command-line parameter with the mechanism used by @ao508 ... so perhaps you are recommending that we drop the line from the portal.properties.EXAMPLE file and instead make it an undocumented property?
Or would you prefer that we add a comment above the portal.properties line saying "#caution: db.enable_pooling is intended to be used only for data-import - memory leaks are observed setting this property to true in cbioportal web applications" ?

sheridancbio:
@inodb We have dropped the property from portal.properties. We decided this would be an undocumented feature used only for building an (network optimized) importer.

PR #6271, opened 2019-06-13
resolve gene ambiguity reporting exception during profile import

sheridancbio:
# What? Why?
### Background
During import of profile datafiles, the entrez_gene_id for each record is resolved. Records with an explicit entrez_gene_id use the explicit entrez_gene_id but if not, we accept a hugo_gene_symbol field which is converted to entrez_gene_id. If an exact match to a hugo_gene_symbol in the database gene table is not found, gene_aliases are considered. When we encounter hugo_gene_symbols which map to multiple entrez_gene_ids, the importer will reject / skip these rows in the profile datafile usually. Two exceptions are allowed : one for RPPA profiles, and one for cases which involve entrez_gene_ids which are of type "miRNA". For the miRNA exception, we attempt to import the current datafile record for every matched entrez_gene_id of type "miRNA". The reason for doing this is presence of the specially constructed microRNA combination form (precursor/mature) gene table records created with negative value entrez_gene_ids by the deployment script ImportMicroRNAIDs.java  --- because CNA profiles for precursor gene symbol aliases (such as "hsa-mir-100") will match combination form gene record "MIR-100/100" and also combination form gene record "MIR-100/100*" which combines the precursor with either the 3p or the 5p arm mature form) and the CNA profile is applicable to both of these entrez_gene_id.

Another factor is that once a profile record has been imported for any entrez_gene_id, further attempts to import profile records for the same entrez_gene_id (within the same data file) will be rejected / skipped as duplicated records.

### The bug and the fix
The bug occurred in the logging code which reported skipped (due to duplication or ambiguity) records. The logging code attempted to distinguish duplicated microRNA records (which are expected to normally match multiple entrez_gene_ids) from ambiguous records of other types (protein coding genes mainly), but the logic would not handle cases where a gene alias matches a mixture of microRNA entrez_gene_ids plus a single/unique protein coding entrez_gene_id. The fix here is combine the logging of ambiguity problems involving non-microRNA genes with the logging of ambiguity problems involving a mixture of microRNA and non-microRNA genes into a single clause.

- cases where a hugo gene symbol was ambiguous (matching both microRNA and non-microRNA) and duplicated in a data profile file threw exceptions.
- now we consolidated the logging of these ambiguities and do not throw exceptions
- additional logging for cases where microRNA import was attempted but failed (versus no microRNAs being involved)

Co-authored-by: Angelica Ochoa <aochoa4230@gmail.com>

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
@@ -249,7 +249,110 @@ public class ImportTabDelimData {
             buf.close();
         }                
     }
-    
+
+    /**
+    * Attempt to create a genetic_alteration record based on the current line read from a profile data file.
+    * <ol>
+    *   <li>Commented out lines and blank lines are always skipped (returns false)
+    *   <li>The line is split into columns by the tab delimiter
+    *   <li>The involved genes (list of entrez_gene_ids) are determined:
+    *     <ol>
+    *       <li>Hugo_Symbol and Entrez_Gene_Id column indices are read and validated
+    *       <li>if neither are available, the line is skipped
+    *       <li>if Hugo_Symbol contains '///' or '---', the line is skipped
+    *       <li>rppaProfile parsing has special rules for determining the involved genes
+    *       <li>if Entrez_Gene_Id is available, use that to determine the involved genes
+    *       <li>if Hugo_Symbol is available, use that to determine the involved genes (truncate symbols with '|' in them)
+    *       <li>if the involved genes list is still empty, the line is skipped (returns false)
+    *     </ol>
+    *   <li>We now choose between three branches:
+    *     <ol>
+    *       <li>if the set of involved genes is empty, we skip the line (returns false)
+    *       <li>if there is exactly 1 involved gene:
+    *         <ol>
+    *           <li>if this is a 'discretizedCnaProfile', normalize the CNA values and create a list of cnaEvents to be added
+    *           <li>attempt to store the record in genetic_alteration
+    *           <li>if the record is successfully stored (not duplicated), create (or update) records in sample_cna_event for the created list of cnaEvents (if any)
+    *         </ol>
+    *       <li>if there are several involved genes, loop ; for each one:
+    *         <ol>
+    *           <li>if the gene type is 'miRNA' (or this is an rppaProfile), attempt to store the record (using this gene) in genetic_alteration
+    *           <li>otherwise, increment a count of how many genes were not of type 'miRNA' (used for logging)
+    *         </ol>
+    *         <ul>
+    *           <li>after looping through all involved genes, check whether any records were successfully stored in genetic_alteration - if not log the failure
+    *         </ul>
+    *     </ol>
+    *   <li>If a record was (or more than one were) successfully stored in genetic_alteration, return true ; else false
+    * </ol>
+    * <p>
+    * During the import of any single profile data file, at most one record per Entrez_Gene_Id will be successfuly imported to genetic_alteration.
+    * Each attempt to import is done through a call to the function storeGeneticAlterations().
+    * That function will check an instance variable importSetOfGenes, and if the gene has been previously imported, no new attempt is made (failure).
+    * Each time a gene is successfully imported, it is added to importSetOfGenes.
+    * <p>
+    * MicroRNA are treated specially because of the possible presence of constructed combination forms (such as 'MIR-100/100*' and 'MIR-100/100').
+    * In these cases a Hugo_Symbol such as 'hsa-mir-100' may be expected to match the Entrez_Gene_Id for both of these combination forms.
+    * In that case, we want to import several copies of the genetic alteration profile line .. one for each matched gene of type 'miRNA'.
+    * This allows the visualization of both CNA event profiles for the microRNA precursor with expression profiles for the microRNA mature form.
+    * <p>
+    * The current implementation of this code does not attempt to "merge" / "unify" lines in the profile data file which have duplicated Entrez_Gene_Id.
+    * Instead, the first encountered line which maps to the Entrez_Gene_Id will be stored as a record in genetic_alteration (returns true).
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -436,17 +440,19 @@ else if (!entrez.matches("[0-9]+")) {
...
+                                    // this means that microRNA or RPPA could not be stored
+                                    ProgressMonitor.logWarning("Could not store microRNA or RPPA data"); //TODO detect the type of of data and give specific warning
                                 }
                                 else {
-                                    //should not occur. It would mean something is wrong in preceding logic (see else if (genes.size()==1) ) or a configuration problem, e.g. where a symbol maps to both a miRNA and a normal gene:
-                                    throw new RuntimeException("Unexpected error: unable to process row with gene " + geneSymbol);
+                                    // this case :
+                                    //      - the hugo gene symbol was ambiguous (matched multiple entrez-gene-ids)
+                                    //      - at least one of the entrez-gene-ids was not a microRNA (and this is not an rppaProfile)
+                                    //      - all of the matched microRNA ids (if any) failed to be imported (presumably already imported on a prior line)
```
pieterlukasse: I like these comments 👍 . Should we perhaps add the respective code / refactor the existing code to reflect these rules explicitly?

n1zea144: I would say its worth the effort if we plan to keep this file, but our current plan is to start cleaning up the core later this year, which will be followed up by resurrecting the spring-batch based importer.

sheridancbio:
rescinding request for review to @averyniceday by request

sheridancbio:
@pieterlukasse I think what we will do to address your points is to add additional comments in the code explaining the logic of this method with examples. Writing unit tests or refactoring the code is outside of the scope of this effort. The comments around the throwing of the runtime exception were added when we did not expect cases where a gene alias would refer both to protein coding genes in the gene table and also miRNA genes in the gene table. The logging of duplications and ambiguities which require skipping lines in the profile data file were basically categorized into 2 cases : case#1 : if there were more than one non-microRNA gene involved, we reported that we skipped the record due to being unable to resolve the gene symbol into a single entrez_gene_id. case#2 : if all the genes involved were microRNA, we reported that we were unable to import the current (miRNA) profile line from the datafile. What we did not expect was a gene alias matching exactly one protein-coding gene, and also several microRNA genes. The reporting code threw the exception in this case. If you consider that matching 2 protein-coding genes and also several microRNA genes did *not* generate the exception ... and that no protein-coding genes are even attempted to be imported if any microRNA genes are involved, maybe it makes a good argument for not throwing an exception in the case that exactly one protein coding gene is involved plus one or more microRNA (which had previously been imported to).

sheridancbio:
@pieterlukasse I have updated the documentation above this function to summarize the logical flow and explain the microRNA considerations. I've also added sample input / output cases which could function as unit test definitions. For simpler examination, I've generated the javadoc for this class (including private members) and attached it.
[ImportTabDelimData.pdf](https://github.com/cBioPortal/cbioportal/files/3299134/ImportTabDelimData.pdf)

sheridancbio:
I've fixed the line in the documentation which printed out as "4. after looping for several involved genes, ..."; now it should show up with a bullet point under the two prior lines ...
1.
2.
• after looping for several involved genes, ...

sheridancbio:
I've fixed the incorrect reference to table "genetic_alterations" and replaced it with "genetic_alteration"

sheridancbio:
[ImportTabDelimData.pdf](https://github.com/cBioPortal/cbioportal/files/3299181/ImportTabDelimData.pdf)

New version of javadoc

pieterlukasse:
thanks @sheridancbio. I like the extended documentation and test cases. These are a good basis for unit tests indeed. 
One part of the code that I still don't understand is: what happens when `discretizedCnaProfile` is `true` but `genes.size() > 1`? Will there be no CNA events added in `cnaEventsToAdd`? The code seems to contradict your documentation at
> This allows the visualization of both CNA event profiles for the microRNA precursor with expression profiles for the microRNA mature form.`

sheridancbio:
@pieterlukasse When genes.size() > 1, that takes precedence in the code; discretizedCnaProfile will not even be considered in that case.  If genes.size() > 1, and none of the matched genes are of type "miRNA", then no imports will be attempted for he current line. (One exception : rppaProfiles, which have special rules). If genes.size() > 1 and there *are* some genes of type miRNA, then there will be an attempt to import the profile on the current line for each gene of type miRNA. Failures will be logged, but if any of the miRNA genes are successfully used to import the current line profile line, the overall call to parseLine() is considered a success.

sheridancbio:
I rebased and made a small language adjustment (based on feedback from @mandawilson).

If the documentation seems to adequately explain the use cases and (complicated) code logic, we would like to move forward with merging this PR and importing a couple of studies which need this bugfix to allow import.

sheridancbio:
@pieterlukasse I see your point about discretizedCnaProfile .. you are right; for microRNA genes I think there will be no CNA Events to add to the sample_cna_event table. Maybe this was an error when we refactored this code adding the separate sample_cna_event table to hold events rather than storing the comma separated profile of events in genetic_alteration. I'll discuss this with the pipelines team. Fixing that error can be done in a separate PR ... but thank you for pointing it out.

sheridancbio:
Dismissing change request from @pieterlukasse to allow merge (under the belief that the documentation additions have sufficiently clarified the logic and use cases). I will create an issue for the sample_cna_event table non-use for microRNAs and we will fix that error in a follow up PR.

pieterlukasse:
@sheridancbio thanks for following up with an extra fix. One request: next time, please don't dismiss a change request before giving me enough time to reply. I prefer to do this myself. Please consider the time difference (I am not in Boston yet :stuck_out_tongue_winking_eye: ).
For the record: I think this PR does not introduce a new issue and is a step forward since the code better reflects a situation that is now known to be expected. I still think there should be a set of unit tests that can work as regression tests. This will make the future refactoring easier, since one can then easily check the new implementation against the existing tests.

PR #6276, opened 2019-06-17
Allow NA values in start and end position of MAF files

oplantalech:
# What? Why?
PR #2649 introduced a check to avoid `Start_Position` and `End_Position` in MAF files to contain float numbers. However, this disallowed adding `NA` values. Since both `Start_Position` and `End_Position` are optional in MAF files, it should be allowed to contain `NA`. 

This PR proposes to raise a warning when `Start_Position` or `End_Position` contains an `NA` value, to notify the user loading that this causes the chromosome plot of the Patient View to not be displayed properly for this mutation (it shows up at the beginning of the chromosome, since it does not know the position). 

PD: the lollipop in the Mutation tab is not altered by having `NA` in `Start_Position` or `End_Position` because it uses the `HGVSp_Short` to position the mutation in the lollipop.


jjgao:
Maybe we should make start and end positions as required fields. I don't see mutations without them to be very useful in the future.

oplantalech:
@rmadupuri @jjgao At that moment, if Start_Position and End_Position are optional (also according to File Formats), I think it makes sense to allow for NA. However, I do wonder: since we enforce to have the amino acid change (HGVSp_Short), isn't it possible to calculate the start and end positions from this change if they are not provided in the MAF file?

jjgao:
@oplantalech it's not one-to-one from protein to genomic change. The genomic change fields are really critical and I don't think they should be optional.

Currently they are optional b/c we also hack other events such as fusions and V3 deletion in MAF. But real mutations, we should require the genomic change fields.

oplantalech:
@jjgao ok, understood. Then, what should we do here? Close this PR and raise an issue to enforce start and end positions, merge this PR as a temporary solution...

jjgao:
@oplantalech let's merge this one and open another issue?

oplantalech:
@jjgao Sure, let's do that

jjgao:
Issue created to capture the discussion of making genomic change columns required: https://github.com/cBioPortal/cbioportal/issues/6511

PR #6283, opened 2019-06-18
Modify copy_number_seg.SEG_ID col to BIGINT(20)

ao508:

# What? Why?
Fixes out of range value for column `copy_number_seg.SEG_ID` exception during import. 

- alter `copy_number_seg.SEG_ID` type to `BIGINT(20)` in database

# Checks
- [ ] Runs on Heroku.

### DIFF
--- a/db-scripts/src/main/resources/cgds.sql
+++ b/db-scripts/src/main/resources/cgds.sql
@@ -694,7 +694,7 @@ CREATE TABLE `sample_cna_event` (
 
 -- --------------------------------------------------------
 CREATE TABLE `copy_number_seg` (
-  `SEG_ID` int(255) NOT NULL auto_increment,
+  `SEG_ID` bigint(20) NOT NULL auto_increment,
   `CANCER_STUDY_ID` int(11) NOT NULL,
   `SAMPLE_ID` int(11) NOT NULL,
   `CHR` varchar(5) NOT NULL,
@@ -819,4 +819,4 @@ CREATE TABLE `info` (
   `GENESET_VERSION` varchar(24)
 );
 -- THIS MUST BE KEPT IN SYNC WITH db.version PROPERTY IN pom.xml
-INSERT INTO info VALUES ('2.9.0', NULL);
+INSERT INTO info VALUES ('2.9.1', NULL);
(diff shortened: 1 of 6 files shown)


db-scripts/src/main/resources/cgds.sql
```diff
@@ -694,7 +694,7 @@ CREATE TABLE `sample_cna_event` (
 
 -- --------------------------------------------------------
 CREATE TABLE `copy_number_seg` (
-  `SEG_ID` int(255) NOT NULL auto_increment,
+  `SEG_ID` BIGINT(20) NOT NULL auto_increment,
```
sheridancbio: other tables use "bigint(20)" rather than BIGINT(20) ... so probably better to stay consistent with casing of types

db-scripts/src/main/resources/migration.sql
```diff
@@ -598,4 +598,8 @@ CREATE TABLE `data_access_tokens` (
     PRIMARY KEY (`TOKEN`),
     FOREIGN KEY (`USERNAME`) REFERENCES `users` (`EMAIL`) ON DELETE CASCADE
 );
-UPDATE `info` SET `DB_SCHEMA_VERSION`="2.9.0";
+
+##version: 2.9.1
+ALTER TABLE `copy_number_seg` MODIFY COLUMN `SEG_ID` BIGINT(20);
+
+UPDATE `info` SET `DB_SCHEMA_VERSION`="2.9.1";
```
sheridancbio: The python script which applies the migration script reads this file line by line - I recommend having the newline at the end of the final line. It very likely does not matter, but good to have just for safety.

n1zea144: @ao508 We need to make some minor model changes before this PR is merged:

segID should be changed to Long here:
[CopyNumberSeg.java](https://github.com/cBioPortal/cbioportal/blob/master/model/src/main/java/org/cbioportal/model/CopyNumberSeg.java#L8)

and here:

[CopyNumberSegmentControllerTest.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/test/java/org/cbioportal/web/CopyNumberSegmentControllerTest.java#L41)
[CopyNumberSegmentControllerTest.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/test/java/org/cbioportal/web/CopyNumberSegmentControllerTest.java#L51)

ao508: resolved!

inodb:
It's good to luck at the migration in the release/rc branches when doing migrations. There is already a 2.9.1 schema defined: https://github.com/cBioPortal/cbioportal/blob/release-3.1.0/db-scripts/src/main/resources/migration.sql

PR #6302, opened 2019-06-21
Refactor CoExpressionService to work with iterator over MolecularAlteration objects.

n1zea144:
In an effort to reduce memory consumption when the Co-expression tab is selected, this PR introduces the use of a MyBatis cursor to iterate over Molecular Alterations rather than fetch them all in a single database call.

The PR needs to be profiled in beta to determine the gain in memory optimization against the loss in runtime performance (it may be determined that a cursor should only be used when the cohort exceeds a certain sample size).  Also, its unclear what impact a large volume of concurrent Co-expression requests can have on the database connection pool when cursors are in use.

The PR limits the exposure of the MyBatis cursor to the persistence layer and returns an Iterable to the service layer.  The PR refactors code as needed, makes minor attempts to cleanup the existing code to avoid breaking functionality, but does add comments where time was spent understanding the codebase.

Some JProfiler screenshots comparing runtime performance /memory usage of this PR against the current master branch will be provided, but as previously mentioned, profiling in the beta environment should still be performed before this PR is merged.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
@@ -6,6 +6,8 @@ import org.apache.commons.math3.linear.RealMatrix;
 import org.apache.commons.math3.stat.correlation.SpearmansCorrelation;
 import org.cbioportal.model.Gene;
 import org.cbioportal.model.MolecularAlteration;
+import org.cbioportal.model.GeneMolecularAlteration;
+import org.cbioportal.model.GenesetMolecularAlteration;
 import org.cbioportal.model.Geneset;
 import org.cbioportal.model.MolecularData;
 import org.cbioportal.model.MolecularProfile;
@@ -26,6 +28,7 @@ import org.cbioportal.service.exception.MolecularProfileNotFoundException;
 import org.cbioportal.service.exception.SampleListNotFoundException;
 import org.cbioportal.service.CoExpressionService;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
 import org.springframework.stereotype.Service;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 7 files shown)


service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -117,31 +109,49 @@
...
+            List<Integer> queryGeneticEntityIds = Arrays.asList(Integer.valueOf(queryGeneticEntityId));
+            maItr = molecularDataService.getMolecularAlterations(molecularProfileId, queryGeneticEntityIds, "SUMMARY");
         } else if (geneticEntityType.equals(GeneticEntityType.GENESET)) {
-            molecularAlterations = genesetDataService.getGenesetAlterations(
-                molecularProfileId, null);
+            List<String> queryGeneticEntityIds = Arrays.asList(queryGeneticEntityId);
+            maItr = genesetDataService.getGenesetAlterations(molecularProfileId, queryGeneticEntityIds);
+        }
+        for (MolecularAlteration ma : maItr) {
+            queryMolecularDataList = ma;
```
ao508: should this be adding to this list or since it's just querying one genetic entity id that `queryMoleuclarDataList` is just being reassigned to whatever the first element of `maItr` is (assuming it's non-empty)

n1zea144: Basically yes - its only passing in one entity id so it only expects on record back.

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -160,29 +173,52 @@
...
+            maItr = molecularDataService.getMolecularAlterations(molecularProfileId, null, "SUMMARY");
+        } else if (geneticEntityType.equals(GeneticEntityType.GENESET)) {
+            maItr = genesetDataService.getGenesetAlterations(molecularProfileId, null);
         }
-        coExpressionList = computeCoExpressions(values, includedQueryValues, isMolecularProfileBOfGenesetType, threshold);  
-        return coExpressionList;
+
+        // For each MolecularAlteration in the profile, compute a CoExpression to return.
+        // If the MolecularAlteration is for the query gene/geneset, skip it.  Otherwise,
+        // filter out genetic_alteration values from genetic_alteration values from
```
ao508: accidentally repeated some words in this comment

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
ao508: thanks for clarifying, just wanted to confirm

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
n1zea144: Thanks for catching, just updated.

n1zea144:
Here are some screen capture comparing runtime performance between precursor/cursor implementations.  The good news is the cursor implementation is in the same ballpark as precursor implementation.  Profiling is being performed against the brca_metabric cohort hosted on public_test (devdb.cbioportal.org/public_test).  I'll follow this up with some memory usage profiling....

Current "precursor" codebase, ~37 seconds to compute coexpression of brca1 gene:

image

cursor codebase, ~39 seconds to compute coexpression of brca1 gene:

image

jjgao:
@n1zea144 just logged for using cursor for Enrichments endpoints if this works out well.

n1zea144:
Here is one of many screen captures comparing memory usage by the coexpression endpoint between the master branch and this PR (mybatis cursor implementation).  The cursor implementation uses ~half as much memory.   There is some variability in this savings and much of the savings is in the downstream computeCoExpressions method (which has me scratching my head), but I think the gain due to the use of cursors will only improve and become clearer as the datasets become larger.  The highlighted row illustrates that the current implementation is using +1,284MB more memory to satisfy the fetchCoExpression request as compared to the cursor implementation (two rows down):

image

Issue #2237, opened 2017-03-23
Code clean-up: remove *oncotator*.java


pieterlukasse:
All oncotator related classes can probably be removed (or `cgds.sql` needs to be updated) since it looks like they depend on the table `oncotator_cache` which does not exist in `cgds.sql` 

Example in `DaoOncotatorCache.java`:
```
"INSERT INTO oncotator_cache (`CACHE_KEY`,`GENE_SYMBOL`, `GENOME_CHANGE`, `PROTEIN_CHANGE`," +
                                " `VARIANT_CLASSIFICATION`," +
...
```

pieterlukasse:
@sheridancbio agree?

sheridancbio:
I'm not familiar with the oncotator codebase. There is probably a great deal we can purge from the core Dao package (and others), and this is an obvious example.

I think @onursumer may be familiar with the replacement of oncotator. I'll ask around, but I think you are right; oncotator may not be used at all any more now that we are using genome nexus.

sheridancbio:
Onur confirms that we can purge oncotator code.

pieterlukasse:
solved by pending PR. Closing here...

PR #6304, opened 2019-06-21
cleaning up dead code from portal package

pieterlukasse:
# What? Why?
Fix for https://github.com/cBioPortal/cbioportal/issues/2237

This fix goes a bit further and deletes the whole `org.*` package from the portal module.


pieterlukasse:
notify @sheridancbio

pieterlukasse:
Can we also remove other references to oncotator, like in the DB model (see `mutation_event` table)?

n1zea144:
@pieterlukasse @jjgao We have references to some of theses classes in an MSK pipeline repos.  We can't remove these yet, but will get to get when we work on cleaning up the core later this year.  For now, will close.

PR #6318, opened 2019-06-24
Refactored ImportTabDelimData to handle microRNA aliases properly

sheridancbio:
# What? Why?
When NCBI gene records for microRNA coexist in the database with constructed negatively numbered combination form microRNA entries created by ImportMicroRNAIDs, there is confusion during import when a gene record's hugo_gene_symbol matches gene_alias entries in the gene_alias table. One such example is MIR3180-2 (which is currently the hugo gene symbol for entrez_gene_id 100422956). However MIR3180-2 is also an alias (in ncbi) for entrez_gene_id 107984155 (LOC107984155), and two other negatively numbered combination form genes are created for this alias during microRNA import (MIR-3180-2/3P and MIR-3180-2/5P). The main problem was that when an exact match to the symbol in gene table exists, the alias table was being ignored. Code in importTabDelimData has been further refactored and now considers both aliases and gene symbols when importing records involving microRNA genes.

- separate function added to DaoGeneOptimized to extract linked genes for an alias (Gene_Symbol)
- detect presence or absense of microRNA in matched set of gene records
- separate logic for microRNA import (independent of the number of matched genes)
- separate logic for non-microRNA import (when more than one gene was matched, don't attempt import unless it was rppaProfile)
- drop code supporting cbioCancerGenes
- whitespace/style standardization

Co-authored-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


sheridancbio:
A complete list of gene symbols which have both positive entrez_gene_id ncbi gene records *and* negative combination (precursor/mature) forms constructed with an identical gene alias string:
MIR1178 MIR1182 MIR1184-2 MIR1184-3 MIR1233-2 MIR1236 MIR1282 MIR1292
MIR1302-11 MIR1304 MIR1306 MIR1587 MIR1908 MIR1909 MIR196B MIR1972-2
MIR21 MIR3118-4 MIR3135B MIR3179-3 MIR3180-2 MIR3184 MIR3190 MIR3591
MIR3610 MIR3615 MIR3620 MIR3652 MIR3654 MIR3670-2 MIR3680-2 MIR3916
MIR3934 MIR4271 MIR4283-1 MIR4283-2 MIR4321 MIR4435-1 MIR4435-2 MIR4444-2
MIR4454 MIR4467 MIR4468 MIR4509-1 MIR4509-3 MIR4519 MIR452 MIR4523
MIR4640 MIR4646 MIR4647 MIR4650-2 MIR4656 MIR4680 MIR4707 MIR4723
MIR4738 MIR4745 MIR4746 MIR4771-1 MIR4784 MIR4788 MIR4800 MIR490
MIR492 MIR5004 MIR5047 MIR509-2 MIR512-2 MIR515-2 MIR622 MIR636
MIR637 MIR664B MIR6834 MIR877 MIR941-3 MIR941-4 MIR943

With this code update, if a record matches any gene of type "miRNA" (through alias or directly), it will cause an attempt to import the profile data to only records of type "miRNA". If any such attempts to import succeed, the import of the line is considered successful and otherwise, a log message reports either failure to import microRNA data (if all matched genes are of type 'miRNA') or failure to import data for a symbol which refers to set of genes (a mixture of miRNA and non-miRNA). This failure will normally only be reported if a gene symbol (or alias) is duplicated in the profile data file.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -460,46 +466,64 @@ else if (!entrez.matches("[0-9]+")) {
...
+                        if ("miRNA".equals(g.getType())) {
+                            microRNAGenes.add(g);
+                        } else {
+                            nonMicroRNAGenes.add(g);
+                        }
+                    }
+                    if (!microRNAGenes.isEmpty()) {
+                        // for micro rna, duplicate the data
+                        for (CanonicalGene gene : microRNAGenes) {
+                            boolean result = storeGeneticAlterations(values, daoGeneticAlteration, gene, geneSymbol);
```
mandawilson: Instead of `boolean result = ` we could do `recordStored = ` and get rid of the next few lines.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -522,42 +546,27 @@ else if (!entrez.matches("[0-9]+")) {
...
                                     }
                                 }
-                                else {
-                                    otherCase++;
-                                }
                             }
                             if (recordStored) {
                                 //skip one, to avoid double counting:
                                 nrExtraRecords--;
                             }
```
mandawilson: Instead of `if (recordStored) {} if (!recordStored) {}` we can just do an `if (recordStored) {} else {}`.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
sheridancbio: I think this may not be equivalent. If we dropped the "if (result == true)" line, then recordStored would match the results of the final attempted record. As it is, recordStored winds up true if *any* of the microRNAGenes are imported successfully.

mandawilson: Ah yes of course you are right!

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportTabDelimData.java
```diff
@@ -435,8 +440,14 @@ else if (!entrez.matches("[0-9]+")) {
                             //can return here and avoid duplicated messages:
                             return false;
                         }    
-                    }
-                    else {
+                        if (genes.isEmpty()) {
+                            String gene = (geneSymbol != null) ? geneSymbol : entrez;
+                            ProgressMonitor.logWarning("Gene not found for:  [" + gene
```
ao508: `gene` here should be `geneSymbol` right?

PR #6324, opened 2019-06-26
Fix migration script: cytoband string split - mm10

SRodenburg:
# What? Why?
We found a bug in migration.sql when migrating from seed DB 2.4.0 into a mouse seed DB.
(https://github.com/cBioPortal/datahub/blob/master/seedDB_mouse/README.md)

The chromosome characters are derived by parsing the cytoband string. However, in the mouse seed DB these cytoband strings are (unlike human genome cytobands, e.g. '17p39') split by a space; e.g. '6 F3'.

In the file db-scripts/src/main/resources/migration.sql, we adapted the statements migrating from seed DB 2.4.0 to 2.4.1, when filling the table `reference_genome_gene` there is a series of `SUBSTRING_INDEX` statements meant to split the chromosome character from the cytoband. I adapted the code such that it is also able to split on whitespace.

### DIFF
--- a/db-scripts/src/main/resources/migration.sql
+++ b/db-scripts/src/main/resources/migration.sql
@@ -422,7 +422,13 @@ INSERT INTO reference_genome_gene (ENTREZ_GENE_ID, CYTOBAND, EXONIC_LENGTH, CHR,
 	ENTREZ_GENE_ID,
 	CYTOBAND,
 	LENGTH,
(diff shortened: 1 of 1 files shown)


inodb:
Thanks @SRodenburg 	- I think there are quite a few other issues as well for mouse support in the latest version of cBioPortal:

https://github.com/cBioPortal/cbioportal/issues/6326

I think in general the migrations shouldn't be allowed to change once in master branch, it would prolly have to be a new migration. But since this looks pretty innocuous and we're nearing version 2.9.x, maybe it's ok to change this retroactively

@khzhu 	could you maybe take a look at this one?

khzhu:
@inodb , I've tested the query on human genes. It worked.

PR #6328, opened 2019-06-26
Fix endpoint of molecularProfiles with structural variants

oplantalech:
# What? Why?
Following the merge of #4057 we found a bug: the structural variant profile does not add the ORDERED_SAMPLE_LIST in the genetic_profile_samples table, making the endpoint `/molecular-profiles/{molecularProfileId}/molecular-data/fetch` to fail when asking for a structural variant profile. This small fix solves the issue.

This is a critical bug: without the structural variants frontend PR merged, the current Plots tab in rc is broken.


oplantalech:
@jjgao can you merge?

PR #6332, opened 2019-06-26
add tutorial PDFs

### LINKED ISSUE #5989: Tutorials are (sometimes) not accessible behind corporate firewalls
Tutorials are now in google drive, and this is (sometimes) not accessible behind corporate firewalls. Would be great if a download link can be provided next to the iframes that now host the google docs.

cc: @pieterlukasse

tmazor:
fix #5989


alisman:
@inodb we have to deploy backend before frontend update

PR #6334, opened 2019-06-27
mutation fetch spark persistence layer

doori:
Spark implementation of mutated-genes/fetch api.

- Changed service level.

Reviewers:
@n1zea144 @kalletlak


core/src/main/java/org/mskcc/cbio/portal/scripts/ParquetWriter.java
```diff
@@ -0,0 +1,126 @@
+/*
```
n1zea144: As previously noted,  I would move this to a util package in the persistence-spark

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
```diff
@@ -0,0 +1,80 @@
...
+    private SparkSession spark;
+
+    @Value("${data.parquet.folder}")
+    private String PARQUET_DIR;
+
+    private static Log logger = LogFactory.getLog(SignificantlyMutatedGeneSparkRepository.class);
+
+    @Override
+    public List<MutSig> getSignificantlyMutatedGenes(String studyId, String projection, Integer pageSize, Integer pageNumber, String sortBy, String direction) {
+        System.out.println(PARQUET_DIR);
```
n1zea144: Lets remove this line.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
```diff
@@ -0,0 +1,80 @@
...
+        Dataset<Row> mutationDf = spark.read()
+            .parquet(PARQUET_DIR + "/" + studyId + "/data_mutations_extended.txt.parquet");
+
+        // for MSK
+        if (studyId.contains("msk")) {
```
n1zea144: This is just for development, we obviously don't want to hardcode study ids into the code.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
```diff
@@ -0,0 +1,80 @@
...
+            Dataset<Row> mutationsMskcc = spark.read()
+                .parquet(PARQUET_DIR + "/" + studyId + "/data_mutations_mskcc.txt.parquet");
+            mutationsMskcc = mutationsMskcc.drop("cDNA_change");
```
n1zea144: Why is this here?

core/pom.xml
```diff
@@ -24,7 +24,11 @@
       <artifactId>security-spring</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+      <dependency>
```
n1zea144: Making a dependency on persistence-spark so we can place the parquet writer here doesn't sit right with me, but it is consistent with where the other scripts go.  Should we consider keeping it in persistence-spark until a better place is identified?

core/src/main/resources/applicationContext-core.xml
```diff
@@ -7,8 +7,23 @@
...
+        <b:property name="locations">
+            <b:list>
+                <b:value>file:///${PORTAL_HOME}/portal.properties</b:value>
+                <b:value>classpath:portal.properties</b:value>
+            </b:list>
+        </b:property>
+    </b:bean>
+    
     <context:component-scan base-package="org.mskcc.cbio.portal"/>
+    <context:annotation-config />
```
n1zea144: Remind me what we need to autowire here?

core/src/test/resources/applicationContext-dao.xml
```diff
@@ -17,6 +17,8 @@
 		     <list>
 			     <value>file:///#{systemEnvironment['PORTAL_HOME']}/src/main/resources/portal.properties</value>
                  <value>file:///#{systemEnvironment['PORTAL_HOME']}/src/main/resources/maven.properties</value>
+                 <value>file:///#{systemEnvironment['PORTAL_HOME']}/portal.properties</value>
```
n1zea144: Why are these modifications necessary?

persistence/persistence-spark/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+    <artifactId>persistence-spark</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.mskcc.cbio</groupId>
+            <artifactId>persistence-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
```
n1zea144: Can we move this to top-level pom with other spring deps?
[https://github.com/cBioPortal/cbioportal/blob/master/pom.xml#L407](https://github.com/cBioPortal/cbioportal/blob/master/pom.xml#L407)

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -23,11 +23,11 @@
     <!-- jndi profile is defined here to prevent exceptions during war deployment -->
     <context-param>
         <param-name>spring.profiles.active</param-name>
-        <param-value>${dbconnector:jndi},${authenticate}</param-value>
+        <param-value>${dbconnector:dbcp},${authenticate}</param-value>
```
n1zea144: How did changes to this file get into your PR?

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
doori: The study Ids are not hardcoded, but the parquet filenames are. I added this because the other studies, genie and brca_tcga, didn't have data_mutations_mskcc.txt file, but msk_impact_2017 did. I thought msk uses the data_mutations_mskcc.txt file. If this is not true I can remove this..

doori: Maybe an enum or some static file for filenames would be cleaner than having "data_mutations_extended.txt.parquet" and so on?

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
doori: drop() is called before unionByName() because union requires both Datasets to have the same column names.

portal/src/main/webapp/WEB-INF/web.xml
doori: this was required for running webapp-runner.

core/src/test/resources/applicationContext-dao.xml
doori: These were added to resolve CI build failure - Could not resolve placeholder 'spark.app.name' in value "${spark.app.name}". An example of a failed build: https://api.travis-ci.org/v3/job/552881291/log.txt

core/src/test/resources/applicationContext-dao.xml
```diff
@@ -17,6 +17,8 @@
 		     <list>
 			     <value>file:///#{systemEnvironment['PORTAL_HOME']}/src/main/resources/portal.properties</value>
                  <value>file:///#{systemEnvironment['PORTAL_HOME']}/src/main/resources/maven.properties</value>
+                 <value>file:///${PORTAL_HOME}/portal.properties</value>
+                 <value>classpath:portal.properties</value>
```
fedde-s: Are you sure that this is correct ordering, and that that syntax to evaluate `$PORTAL_HOME` actually works in this context? This has gone wrong several times across the code base over time, leading to configurations varying from one XML to another.

fedde-s: I feel like `$PORTAL_HOME/portal.properties` should _only_ be evaluated at runtime (never compile time), and `<portal_source>/src/main/resources/*.portal.properties` (which is the same thing as `classpath:portal.properties`) should be included at build time if at all, and this code seems to do the exact opposite.

fedde-s: In addition, I feel like the runtime config file should override the buildtime defaults if it is in effect, and not the other way around.

fedde-s: Spring documentation I bellieve to be relevant: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/io/support/PropertiesLoaderSupport.html#setLocations-org.springframework.core.io.Resource...-

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
kalletlak: why not set [mergeSchema](https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#schema-merging) option to true and read all the mutation files together. Something like
`spark.read.option("mergeSchema", "true").parquet(PARQUET_DIR + "/" + studyId + "/data_mutations*.txt")`

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
```diff
@@ -0,0 +1,80 @@
...
+        if (studyId.contains("msk")) {
+            Dataset<Row> mutationsMskcc = spark.read()
+                .parquet(PARQUET_DIR + "/" + studyId + "/" + ParquetConstants.DATA_MUTATIONS_MSKCC);
+            mutationsMskcc = mutationsMskcc.drop("cDNA_change");
+            mutationDf = mutationDf.unionByName(mutationsMskcc);
+        }
+        mutationDf.createOrReplaceTempView("mutation");
+
+        StringBuilder sb = new StringBuilder("select first(Hugo_Symbol),");
+        sb.append("Entrez_Gene_Id, ");
```
kalletlak: all appends can go into `StringBuilder` constructor argument

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
```diff
@@ -0,0 +1,80 @@
...
+        sb.append("COUNT(*), ");
+        sb.append("COUNT(DISTINCT(Tumor_Sample_Barcode)) ");
+        sb.append("from mutation ");
+        sb.append("group by Entrez_Gene_Id ");
+        if (sortBy != null && direction != null) {
```
kalletlak: I think both these conditions should be handled separately

kalletlak: and could you try this sort on the dataset and compare the performance with both the methods. something like`result.orderBy(asc(sortBy))` or `result.orderBy(desc(sortBy))`

portal/src/main/webapp/WEB-INF/web.xml
kalletlak: you could pass it as runtime argument `-Ddbconnector=dbcp`

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
doori: yes it can. I broke them off for readability.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetConstants.java
```diff
@@ -0,0 +1,8 @@
+package org.cbioportal.persistence.spark.util;
```
n1zea144: As previously  mentioned, this will not scale.

portal/src/main/webapp/WEB-INF/web.xml
n1zea144: I mentioned above - the web-app-runner support has been merged into master and rc.  I think it should be removed from this PR and the PR should be rebased.

core/pom.xml
```diff
@@ -418,6 +417,28 @@
...
+            <version>2.3</version>
+            <executions>
+                <execution>
+                    <phase>package</phase>
+                    <goals><goal>copy</goal></goals>
+                    <configuration>
+                        <artifactItems>
+                            <artifactItem>
+                                <groupId>com.github.jsimone</groupId>
+                                <artifactId>webapp-runner</artifactId>
```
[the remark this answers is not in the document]
doori: Would you update the spark-parquet-persistence branch? If I rebase my branch, wouldn't it still show up on this PR unless we update the branch I'm merging into?

core/src/test/resources/applicationContext-dao.xml
doori: Even after moving Parquet Writer, the dao tests were failing without this change. Please see https://travis-ci.org/cBioPortal/cbioportal/jobs/553352211 which was a build after moving Parquet Writer.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
doori: thanks for the suggestion. Updated code to load data_mutations*

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SignificantlyMutatedGeneSparkRepository.java
doori: sortBy != null and projection != 'ID' is actually in most of the MyBatis Mappers. Do you think this is incorrect logic?

core/src/test/resources/applicationContext-dao.xml
n1zea144: @doori Yes, I see.  TestDaoCancerStudy is looking for ${spark.app.name} which has to come from portal.properties .  I think you should follow the order specified in the business context file which will allow a portal.properties file that resides inthe PORTAL_HOME environment variable to override properties found in a "baked-in" portal.properties file:

[https://github.com/cBioPortal/cbioportal/blob/499beab3ca2b15e5ec525dfc2f3f7fe974d0ea31/business/src/main/resources/applicationContext-business.xml#L19](https://github.com/cBioPortal/cbioportal/blob/499beab3ca2b15e5ec525dfc2f3f7fe974d0ea31/business/src/main/resources/applicationContext-business.xml#L19)

doori: Thanks for the explanation. I moved the classpath: line before the systemEnvironment one.

Issue #6337, opened 2019-06-28
mutated-genes/cna-genes endpoints return too much data when study has panels


zhx828:
This is a new feature introduced in https://github.com/cBioPortal/cbioportal/pull/5994/files.

But when gene panel has long list of genes, the data can easily goes up to hundred megabytes. For instance in this cell line study, the panel has around 5000 genes. https://www.cbioportal.org/study/summary?id=cellline_ccle_broad

Proposed solution:
The matchingGenePanels does not return the full list of the genes and description but only the genePanelId, and also study view will fetch the gene panel info separately for further frequency calculation.

PR #6341, opened 2019-06-28
Only include gene panel id in the AlterationCountByGene

zhx828:
In the big panel, the return data would be to big to return the detailed info

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6337

### DIFF
--- a/model/src/main/java/org/cbioportal/model/AlterationCountByGene.java
+++ b/model/src/main/java/org/cbioportal/model/AlterationCountByGene.java
@@ -2,7 +2,7 @@ package org.cbioportal.model;
 
 import java.io.Serializable;
 import java.math.BigDecimal;
-import java.util.List;
+import java.util.Set;
 
 public class AlterationCountByGene implements Serializable {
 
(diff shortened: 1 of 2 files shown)


model/src/main/java/org/cbioportal/model/AlterationCountByGene.java
```diff
@@ -12,7 +12,7 @@
 	private Integer totalCount;
 	private Integer numberOfSamplesProfiled;
     private BigDecimal qValue;
-    private List<GenePanel> matchingGenePanels;
+    private List<String> matchingGenePanelIds;
```
kalletlak: @zhx828 Is there a possibility for duplicate ids. If yes then better to use `Set`.

zhx828: The way the matchingGenePanelIds generated would avoid the duplication. But I agree the Set probably better here.

zhx828: @kalletlak updated.

jjgao:
@zhx828 maybe this should go to master?

zhx828:
@jjgao this needs the backend+frontend changes deploy at the same time. I don't think master should be the place.

jjgao:
I see. Let's make it into the next release then.

We may need to find a way to do quick bug fixes that requires both frontend and backend just like this one. @inodb

inodb:
@jjgao the process is the same:

1. disable auto deploy to netlify on frontend
2. tag frontend version
3. update frontend version in backend pom
4. tag backend
5. deploy new tagged docker image to k8s without frontend url set

PR #5679, opened 2019-02-04
Matchminer Proxy controller

victoria34:
# What? Why?
Create MatchMiner Controller as a proxy to get matched trials in patient view(Trial Match tab).

PR #6348, opened 2019-07-02
Create proxy MatchMinerController.java

victoria34:
# What? Why?
Create MatchMiner Controller as a proxy to get matched trials in patient view(Trial Match tab).


zhx828:
This is the same code from pr https://github.com/cBioPortal/cbioportal/pull/5679. Just to avoid resolving too many conflicts when rebasing to master branch.

Issue #6359, opened 2019-07-08
Webapp-runner deployment does not allow docker volume file mounts 


pvannierop:
While using the new docker solution for portal deployment we came across the limitation of  that the new solution, that runs the .war file with webapp-runner is not compatible with file mounts in docker. Placing a custom logo in the /image directory of the expanded webapp-runner source dir prior to running webapp-runner interferes with server startup.
```
  cbioportal-container:
    container_name: cbioportal-container
    image: cbioportal/cbioportal:master
    ports:
    - "8081:8080"
    volumes:
    # by default webapp-runner decompresses the war into /expanded prior to serving the content
    # line below does not work it interfered with the web-runner unpacking
    - $PWD/custom_logo.jpg:/target/tomcat.8080/webapps/expanded/images/custom_logo.jpg:ro
    - $PWD/portal.properties:/cbioportal/portal.properties:ro
    - $PWD/frontendConfig.json:/tmp/frontendConfig.json:ro
    environment:
      CATALINA_OPTS: "-Xms2G -Xmx24G"
    command: /bin/sh -c 'java -Dauthenticate=false -jar webapp-runner.jar /app.war'
    restart: always
    depends_on:
      - cbioDB
```
Is there an alternative way of deployment that allows unpacking and deployment via Tomcat in the same image?

pvannierop:
@inodb @n1zea144 @fedde-s 
I have solved this issue in [PR #6369](https://github.com/cBioPortal/cbioportal/pull/6369). Please review and merge :)

inodb:
this has been fixed

PR #6369, opened 2019-07-10
Webapp-runner deployment on decompressed webapp directory (not .war file)

pvannierop:
# What? Why?
For customization of the portal and for implementation of authentication solutions (e.g. KeyCloak) in Docker container environment, custom files are mounted into the webapp file structure using volumen mounts. With the new docker solution the portal is deployed as a .war file served by webapp-runner. As described in [Issue #6359](https://github.com/cBioPortal/cbioportal/issues/6359) running from .war compressed webapp does not allow mounting of custom files with a volume mount in a Docker container. 

This PR will update the webapp-runner deployment to run from the `/cbioportal-webapp` directory in the docker image, instead of the compressed .war file. Changes involve:
- decompression of the .war in the Dockerfile's into `/cbioportal-webapp`.
- update of the `CMD` entry in `/docker/web/Dockerfile` to correctly handle `$JAVA_OPTS` and `$WEB_OPTS` env. vars
- update of `README.md` to reflect change in webapp-runner deployment mechanism.

# Note
- Although the .war-based mechanism still works I decided to no longer feature it in the documentation since this may lead to problems with volume mounts.

### DIFF
--- a/docker/web-and-data/Dockerfile
+++ b/docker/web-and-data/Dockerfile
@@ -17,6 +17,9 @@ WORKDIR /cbioportal
 RUN mvn -DskipTests clean install
 
 FROM openjdk:8-jre
+
+ENV PORTAL_WEB_HOME=/cbioportal-webapp
+
 RUN mkdir -p /cbioportal
 COPY --from=build /cbioportal/portal/target/cbioportal*.war /app.war
 COPY --from=build /cbioportal/portal/target/dependency/webapp-runner.jar /webapp-runner.jar
@@ -45,3 +48,6 @@ RUN find /cbioportal/core/src/main/scripts/ -type f -executable \! -name '*.pl' 
 
 # put config files in this folder if you want to override config
 ENV PORTAL_HOME=/cbioportal
+
+RUN mkdir -p $PORTAL_WEB_HOME
+RUN unzip /app.war -d  $PORTAL_WEB_HOME
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -23,6 +23,14 @@ WORKDIR /cbioportal
 RUN mvn -DskipTests clean install
 
 FROM shipilev/openjdk-shenandoah:8
+
+ENV PORTAL_WEB_HOME=/cbioportal-webapp
+
 COPY --from=build /cbioportal/portal/target/cbioportal*.war /app.war
 COPY --from=build /cbioportal/portal/target/dependency/webapp-runner.jar /webapp-runner.jar
-CMD /usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEBAPP_OPTS} /app.war
+
+RUN apt-get update && apt-get install unzip
+RUN mkdir -p $PORTAL_WEB_HOME
+RUN unzip /app.war -d $PORTAL_WEB_HOME
+
+CMD ["/bin/sh", "-c", "/usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEB_OPTS} ${PORTAL_WEB_HOME}"]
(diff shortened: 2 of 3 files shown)


docker/web/Dockerfile
```diff
@@ -23,6 +23,14 @@ WORKDIR /cbioportal
...
+
+ENV PORTAL_WEB_HOME=/cbioportal-webapp
+
 COPY --from=build /cbioportal/portal/target/cbioportal*.war /app.war
 COPY --from=build /cbioportal/portal/target/dependency/webapp-runner.jar /webapp-runner.jar
-CMD /usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEBAPP_OPTS} /app.war
+
+RUN apt-get update && apt-get install unzip
+RUN mkdir -p $PORTAL_WEB_HOME
+RUN unzip /app.war -d $PORTAL_WEB_HOME
```
inodb: Good to not introduce an extra dependency, maybe try https://stackoverflow.com/questions/7882745/how-to-extract-war-files-in-java-zip-vs-jar. Apt-get update is a little unpredictable

fedde-s:
I don't have much time to phrase a review for this right now, but I think @inodb is already suggesting some improvements that would resolve the main issue I saw.

docker/web/Dockerfile
pvannierop: Ok, dependency removed.

docker/web-and-data/Dockerfile
```diff
@@ -45,3 +48,6 @@ RUN find /cbioportal/core/src/main/scripts/ -type f -executable \! -name '*.pl'
 
 # put config files in this folder if you want to override config
 ENV PORTAL_HOME=/cbioportal
+
+RUN mkdir -p $PORTAL_WEB_HOME
+RUN unzip /app.war -d  $PORTAL_WEB_HOME
```
fedde-s: Would it be an improvement to be consistent with the alternative Dockerfile below, and use the `jar` command?

pvannierop: This command is not present in that image. I checked.

inodb: @pvannierop @fedde-s Looks indeed like jre has no jar binary. It's only included in jdk images. It does have unzip included in the image so that's a good alternative

docker/web/Dockerfile
```diff
@@ -23,6 +23,13 @@ WORKDIR /cbioportal
...
+ENV PORTAL_WEB_HOME=/cbioportal-webapp
+
 COPY --from=build /cbioportal/portal/target/cbioportal*.war /app.war
 COPY --from=build /cbioportal/portal/target/dependency/webapp-runner.jar /webapp-runner.jar
-CMD /usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEBAPP_OPTS} /app.war
+
+RUN mkdir -p $PORTAL_WEB_HOME
+RUN cd $PORTAL_WEB_HOME && /root/jdk/bin/jar -xvf /app.war
+
+CMD ["/bin/sh", "-c", "/usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEB_OPTS} ${PORTAL_WEB_HOME}"]
```
inodb: @pvannierop there's a typo here in `WEB_OPTS`, should be `${WEBAPP_OPTS}`

Also double check the the environment parameters are still correctly passed this way. CMD with square brackets vs w/o brackets still confuses me 🙂

fedde-s: For reference, [these](https://docs.docker.com/engine/reference/builder/#cmd) are the docs on the semantics of the different forms.

fedde-s: As I understand it, the form without square brackets is effectively syntactic sugar for `CMD ["sh", "-c", x]`, and doesn't play well with `ENTRYPOINT`. In both cases a shell is started to parse the dollar marks, but in the square bracket case it is explicit.

pvannierop: WEB_OPTS was renamed to WEBAPP_OPTS.

pvannierop: I renamed WEB_OPTS to WEBAPP_OPTS.

inodb:
Thanks @pvannierop !

Issue #6377, opened 2019-07-11
HRRS filter PR breaks cbioportal deployment


pvannierop:
Fix for bug caused by https://github.com/cBioPortal/cbioportal/pull/5806/files.

The recent implementation of HRRS filter introduced new portal.properties params that are mandatory to set. 
```
# enable/disable hrrs logging
# filter which logs incoming web requests to specified logging directory
# logs must be decrypted for more in-depth information
hrrs.logging.filepath=
hrrs.enable.logging=
```
If not set, the portal throws an error and terminates. 

1. It would be better to define default values for both params so the portal always runs.
2. When setting `hrrs.enable.logging=false` the property `hrrs.logging.filepath` is still required. The implementation better not require `hrrs.logging.filepath` to be set, when not logging.

P.S. I wrote this is a hurry. Please excuse my terseness.

inodb:
Seems like duplicate of: https://github.com/cBioPortal/cbioportal/issues/6368

inodb:
GitHub gives recommended similar issues btw:

<img width="668" alt="Screen Shot 2019-07-11 at 11 02 46 AM" src="https://user-images.githubusercontent.com/1334004/61062052-7ea5e280-a3cb-11e9-945e-ed128c180a1e.png">

pvannierop:
@Ino ah ha ha thnx

PR #6378, opened 2019-07-11
HRRS Properties Fix 

### LINKED ISSUE #6368: hrrs configuration crashes with properties.EXAMPLE
- [ ] Would be good to have the default for hrrs be:

    `hrrs.enable.logging=false`

    it should work when all these properties are missing from the properties file:

    ```
    hrrs.logging.filepath=
    hrrs.enable.logging=
    ```
    
    This is useful for people deploying a new version of the portal. They won't have to update their properties file

- [ ] In the EXAMPLE properties file good to mention what hrrs is. Mention that it is for debugging and probably good to comment out by default? The EXAMPLE file is included automatically as the default properties file when no properties file is given so better to not assign any values there

averyniceday:
# What? Why?
Fixes #6368 and #6377 - deployment was failing when hrrs properties were left blank.

- `hrrs.enable.logging` defaults to `false`
- `hrrs.logging.filepath` defaults to `java.io.tmpdir/hrrs-logging.csv`

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
@@ -32,26 +32,67 @@
 
 package org.mskcc.cbio.portal.util;
 
-import java.io.*;
-import java.util.*;
-import org.apache.commons.logging.*;
-import org.cbioportal.model.*;
-import org.springframework.beans.factory.annotation.*;
-import org.springframework.security.core.Authentication;
-import org.springframework.security.core.authority.AuthorityUtils;
-import org.springframework.context.annotation.Bean;
-import org.springframework.stereotype.Component;
-import com.vlkan.hrrs.servlet.base64.Base64HrrsFilter;
+
+import com.vlkan.hrrs.api.HttpRequestRecordWriter;
+import com.vlkan.hrrs.api.HttpRequestRecordWriterTarget;
+import com.vlkan.hrrs.serializer.base64.Base64HttpRequestRecord;
+import com.vlkan.hrrs.serializer.base64.Base64HttpRequestRecordWriter;
+import com.vlkan.hrrs.serializer.base64.guava.GuavaBase64Encoder;
+import com.vlkan.hrrs.serializer.file.HttpRequestRecordWriterRotatingFileTarget;
+import com.vlkan.hrrs.servlet.HrrsFilter;
 import com.vlkan.rfos.RotationConfig;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/DebugHrrsFilter.java
```diff
@@ -50,8 +51,10 @@
 
 public class DebugHrrsFilter extends Base64HrrsFilter {
 
+    private static final String DEFAULT_HRRS_LOGGING_NAME = "/hrrs-logging.csv";
+
     private static RotationConfig createRotationConfig() {
```
inodb: @averyniceday 	is it possible to not run `super(createRotationConfig())` if `hrrs.enable.logging` is false? I assume `java.io.tmpdir` is always set but seems redundant to set up a logging file path if the whole feature is disabled?

averyniceday: @inodb I'll look into it some more. But the superclass Base64HrrsFilter requires a RotationConfig argument passed in.

inodb: @averyniceday 	I see - not worth spending too much time on it, but if it's an easy fix that would be great 👍 - gonna approve for now

averyniceday: @inodb I changed it so this class will extend the HrrsFilter which doesn't require a RotationConfig passed in to the constructor. Then the writer/writerTarget is only set if the logging is enabled. I left the code for defaulting to the tempdir just in case someone enables logging with no path.

inodb: @averyniceday perfect!

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -256,7 +256,9 @@ matchminer.url=
 matchminer.token=
 
 # enable/disable hrrs logging
-# filter which logs incoming web requests to specified logging directory
+# hrrs can be used for debugging - when enabled incoming web requests
+# will be logged into the specified directory
 # logs must be decrypted for more in-depth information
-hrrs.logging.filepath=
-hrrs.enable.logging=
+# see: https://github.com/vy/hrrs/blob/master/README.md
```
inodb: nice

Issue #6346, opened 2019-07-02
clinical-data-bin-counts has the precision of floating issue


zhx828:
The endpoint returns `end: 0.8999999999999999 ` when it should be 0.9 Issue found in https://www.cbioportal.org/study/summary?id=acc_tcga Fraction Genome Altered

Solution: https://stackoverflow.com/questions/179427/how-to-resolve-a-java-rounding-double-issue

PR #6381, opened 2019-07-11
Use BigDecimal over Double for a precise result

zhx828:
Signed-off-by: Hongxin Zhang <hongxin@cbio.mskcc.org>

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6346


web/src/main/java/org/cbioportal/web/parameter/ClinicalDataBinFilter.java
```diff
@@ -51,12 +52,12 @@ public void setEnd(Double end) {
     private boolean rangeIsCoveringCustomBins() {
         if (this.customBins != null && (start != null || end != null)) {
             boolean valid = true;
-            for (Double bin : this.customBins) {
-                if(start != null && start > bin) {
+            for (BigDecimal bin : this.customBins) {
+                if(start != null && start.compareTo(bin) == 1) {
```
onursumer: maybe we should use `> 0` instead of `== 1`: https://stackoverflow.com/a/34677687

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -19,19 +20,19 @@ public DataBinHelper(StudyViewFilterUtil studyViewFilterUtil) {
...
+        BigDecimal min;
         String value;
 
         if (gtMin == null && gteMin == null) {
             // no special outlier
             min = null;
             value = ">";
         }
-        else if (gtMin == null || (gteMin != null && gteMin < gtMin)) {
+        else if (gtMin == null || (gteMin != null && gteMin.compareTo(gtMin) == -1)) {
```
onursumer: same as previous, `< 0` instead of `== -1`?

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -89,64 +90,68 @@ else if (lteMax == null || (ltMax != null && lteMax < ltMax)) {
...
+        else if (q3.compareTo(new BigDecimal("0.001")) != -1 && q3.compareTo(new BigDecimal("1")) == -1) {
             //maxValue = Number((q3 + iqr * 1.5).toFixed(3));
             //minValue = Number((q1 - iqr * 1.5).toFixed(3));
-            maxValue = (new BigDecimal(q3 + iqr * 1.5)).setScale(3, BigDecimal.ROUND_HALF_UP).doubleValue();
-            minValue = (new BigDecimal(q1 - iqr * 1.5)).setScale(3, BigDecimal.ROUND_HALF_UP).doubleValue();
+            maxValue = q3upperBoundry.setScale(3, BigDecimal.ROUND_HALF_UP);
+            minValue = q1LowerBoundry.setScale(3, BigDecimal.ROUND_HALF_UP);
         } 
-        else if (q3 < 0.001) {
+        else if (q3.doubleValue() < 0.001) {
```
onursumer: why do we need to use `doubleValue()` here instead of `compareTo`?

web/src/main/java/org/cbioportal/web/parameter/ClinicalDataBinFilter.java
zhx828: @onursumer I think it's fine. The compareTo only produces -1, 0, 1.

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
zhx828: Ah, it should be converted.

PR #6386, opened 2019-07-15
Clinical data spark parquet persistence

doori:
Spark/parquet backend for clinical data apis in study view.

api/clinical-data-counts/fetch
api/clinical-data-bin-counts/fetch?dataBinMethod=STATIC
api/clinical-data-density-plot/fetch?xAxisAttributeId=FRACTION_GENOME_ALTERED&xAxisBinCount=50&xAxisStart=0&xAxisEnd=1&yAxisAttributeId=MUTATION_COUNT&yAxisBinCount=52&yAxisStart=0&clinicalDataType=SAMPLE

Reviewers:
@n1zea144 @kalletlak

### DIFF
--- a/model/src/main/java/org/cbioportal/model/ClinicalDataModel.java
+++ b/model/src/main/java/org/cbioportal/model/ClinicalDataModel.java
@@ -1,0 +1,50 @@
+package org.cbioportal.model;
+
+public class ClinicalDataModel {
+
+    private String sampleId;
+    private String patientId;
+    private String studyId;
+    private String attrId;
+    private String attrValue;
+
+    public String getSampleId() {
+        return sampleId;
+    }
+
+    public void setSampleId(String sampleId) {
+        this.sampleId = sampleId;
+    }
+
+    public String getPatientId() {
+        return patientId;
+    }
+
+    public void setPatientId(String patientId) {
+        this.patientId = patientId;
+    }
+    
+    public String getStudyId() {
+        return studyId;
+    }
+
+    public void setStudyId(String studyId) {
+        this.studyId = studyId;
+    }
+    
+    public String getAttrId() {
+        return attrId;
+    }
+
+    public void setAttrId(String attrId) {
+        this.attrId = attrId;
+    }
+
+    public String getAttrValue() {
+        return attrValue;
+    }
+
+    public void setAttrValue(String attrValue) {
+        this.attrValue = attrValue;
+    }
+}
--- a/persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
+++ b/persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
@@ -1,0 +1,289 @@
+package org.cbioportal.persistence.spark;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.apache.spark.sql.*;
+import org.apache.spark.sql.expressions.UserDefinedFunction;
+import org.apache.spark.sql.types.DataTypes;
+import org.cbioportal.model.*;
+import org.cbioportal.model.meta.BaseMeta;
+import org.cbioportal.persistence.ClinicalDataRepository;
+import org.cbioportal.persistence.PersistenceConstants;
+import org.cbioportal.persistence.spark.util.ParquetConstants;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+import scala.collection.JavaConverters;
+import scala.collection.Seq;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static org.apache.spark.sql.functions.*;
+
+
+@Component
+@Qualifier("clinicalDataSparkRepository")
+public class ClinicalDataSparkRepository implements ClinicalDataRepository {
+
+    @Autowired
+    private SparkSession spark;
+
+    @Value("${data.parquet.folder}")
+    private String PARQUET_DIR;
+
+    private static Log logger = LogFactory.getLog(ClinicalDataSparkRepository.class);
+    private static final String MUTATION_COUNT = "MUTATION_COUNT";
+    private static final String FRACTION_GENOME_ALTERED = "FRACTION_GENOME_ALTERED";
+
+    @Override
+    public List<ClinicalData> getAllClinicalDataOfSampleInStudy(String studyId, String sampleId, String attributeId, String projection, Integer pageSize, Integer pageNumber, String sortBy, String direction) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public BaseMeta getMetaSampleClinicalData(String studyId, String sampleId, String attributeId) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ClinicalData> getAllClinicalDataOfPatientInStudy(String studyId, String patientId, String attributeId, String projection, Integer pageSize, Integer pageNumber, String sortBy, String direction) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public BaseMeta getMetaPatientClinicalData(String studyId, String patientId, String attributeId) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ClinicalData> getAllClinicalDataInStudy(String studyId, String attributeId, String clinicalDataType, String projection, Integer pageSize, Integer pageNumber, String sortBy, String direction) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public BaseMeta getMetaAllClinicalData(String studyId, String attributeId, String clinicalDataType) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ClinicalData> fetchAllClinicalDataInStudy(String studyId, List<String> ids, List<String> attributeIds, String clinicalDataType, String projection) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public BaseMeta fetchMetaClinicalDataInStudy(String studyId, List<String> ids, List<String> attributeIds, String clinicalDataType) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ClinicalData> fetchClinicalData(List<String> studyIds, List<String> ids, List<String> attributeIds,
+                                                String clinicalDataType, String projection) {
+        List<Dataset<Row>> res = new ArrayList<>();
+        
+        Set<String> studyIdSet = new HashSet<>(studyIds);
+        String dataFile = ParquetConstants.CLINICAL_PATIENT;
+        if (PersistenceConstants.SAMPLE_CLINICAL_DATA_TYPE.equalsIgnoreCase(clinicalDataType)) {
+            dataFile = ParquetConstants.CLINICAL_SAMPLE;
+        }
+        Dataset<Row> clinicalData = loadStudyFiles(studyIdSet, dataFile);
+        
+        // join with clinical sample data to get sample id information.
+        if (PersistenceConstants.PATIENT_CLINICAL_DATA_TYPE.equalsIgnoreCase(clinicalDataType)) {
+            Dataset<Row> clinicalSampleData = loadStudyFiles(studyIdSet, ParquetConstants.CLINICAL_SAMPLE);
+            clinicalData = clinicalData.join(clinicalSampleData.drop("studyId"), "PATIENT_ID");
+        }
+        
+        // filter by sample id
+        if (!CollectionUtils.isEmpty(ids)) {
+            String[] caseArr = ids.stream().toArray(String[]::new);
+            if (PersistenceConstants.SAMPLE_CLINICAL_DATA_TYPE.equalsIgnoreCase(clinicalDataType)) {
+                clinicalData = clinicalData.filter(col("SAMPLE_ID").isin(caseArr));
+            } else {
+                clinicalData = clinicalData.filter(col("PATIENT_ID").isin(caseArr));
+            }
+        }
+
+        // if attributes is null return data from all columns
+        List<String> attributes = attributeIds;
+        if (CollectionUtils.isEmpty(attributeIds)) {
+            attributes = Arrays.asList(clinicalData.columns());
+        }
+        clinicalData.createOrReplaceTempView("clinicalData");
+
+        for (String attributeId : attributes) {
+            StringBuilder sb = new StringBuilder();
+            sb.append("SELECT PATIENT_ID as patientId, SAMPLE_ID as sampleId, studyId");
+            if (!MUTATION_COUNT.equalsIgnoreCase(attributeId) && !FRACTION_GENOME_ALTERED.equalsIgnoreCase(attributeId)) {
+                sb.append(", " + attributeId + " AS attrValue");
+                sb.append(", '" + attributeId + "' AS attrId");
+            }
+            sb.append(" FROM clinicalData ");
+
+            String query = sb.toString();
+            Dataset<Row> sub = spark.sql(query);
+            
+            if (MUTATION_COUNT.equalsIgnoreCase(attributeId)) {
+                sub = sub.join(getMutationCount(studyIdSet), "sampleId");
+            }
+            if (FRACTION_GENOME_ALTERED.equalsIgnoreCase(attributeId)) {
+                sub = sub.join(getFractionGenomeAltered(studyIdSet), "sampleId");
+            }
+            res.add(sub);
+        }
+        
+        Dataset<Row> ds = res.get(0);
+        // union all datasets before calling collectAsList which move all  data into the application's driver process.
+        for (Dataset<Row> r : res.subList(1, res.size())) {
+            ds = ds.unionByName(r);
+        }
+
+        Dataset<ClinicalDataModel> clinicalDataModel = ds.as(Encoders.bean(ClinicalDataModel.class));
+        List<ClinicalDataModel> clinicalDataModels = clinicalDataModel.collectAsList();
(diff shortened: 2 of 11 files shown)


persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,277 @@
...
+    @Value("${data.parquet.folder}")
+    private String PARQUET_DIR;
+
+    private static Log logger = LogFactory.getLog(ClinicalDataSparkRepository.class);
+    private static final String MUTATION_COUNT = "MUTATION_COUNT";
+    private static final String FRACTION_GENOME_ALTERED = "FRACTION_GENOME_ALTERED";
+
+    @Override
+    public List<ClinicalData> getAllClinicalDataOfSampleInStudy(String studyId, String sampleId, String attributeId, String projection, Integer pageSize, Integer pageNumber, String sortBy, String direction) {
+        throw new UnsupportedOperationException();
```
n1zea144: Are these unsupported exceptions because we will not be supporting these calls in spark/parquet?  If that is true, in this PR we should have modifications to [ClinicalDataServiceImpl.java](https://github.com/cBioPortal/cbioportal/blob/master/service/src/main/java/org/cbioportal/service/impl/ClinicalDataServiceImpl.java) which can maintain references to multiple repositories (one mybatis and one spark).  I'm not quite sure the best way to handle this.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetConstants.java
```diff
@@ -3,5 +3,7 @@
 public class ParquetConstants {
     // Parquet file names
     public static final String DATA_MUTATIONS = "data_mutations*.txt.parquet";
-    
+    public static final String CLINICAL_SAMPLE = "data_clinical_sample.txt.parquet";
+    public static final String CLINICAL_PATIENT = "data_clinical_patient.txt.parquet";
+    public static final String CNA_SEG = "data_cna_hg19.seg.parquet";
```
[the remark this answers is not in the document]
doori: I started working on documentation, and thought about the discussion we had last meeting about parquet file names. We can write multiple data files into one parquet file with df.write.mode("append"). I think we could document which data files would be written into the parquet file names we decide on, just like we have sql table names. For example, instead of wildcards, we could write all data_cna_*.seg files to cna_seg, data_mutations*.txt to mutations.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetConstants.java
```diff
@@ -3,5 +3,7 @@
 public class ParquetConstants {
     // Parquet file names
     public static final String DATA_MUTATIONS = "data_mutations*.txt.parquet";
-    
+    public static final String CLINICAL_SAMPLE = "data_clinical_sample.txt.parquet";
```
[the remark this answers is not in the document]
doori: Please see the comment below. All data_clinical_sample*.txt files can be written to clinical_sample parquet file, and we can keep conventional filenames in ParquetConstants.

model/src/main/java/org/cbioportal/model/ClinicalDataModel.java
```diff
@@ -0,0 +1,50 @@
+package org.cbioportal.model;
```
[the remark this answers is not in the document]
doori: This was a question I brought up in one of the first meeting. Spark Encoders can only map Rows to POJOs and ClinicalData is not a POJO, since it extends UniqueKeyBase and is also nested, containing ClinicalAttribute object as an attribute. I created a POJO model for this method, because we conditionally join mutation and fraction data and with join the order is not preserved, so row.get(index) doesn't work.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
doori: These methods are not called by StudyViewController and so not in scope for this project - as per my understanding. This PR does contain ClniicalDataServiceImpl.java with both mybatis and spark repos, where only the implemented method uses spark repo. Change in ClniicalDataServiceImplTest.java is also included in this PR as well.

model/src/main/java/org/cbioportal/model/ClinicalDataModel.java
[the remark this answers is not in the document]
n1zea144: Does the current implementation return a unique sample/patient key?

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetConstants.java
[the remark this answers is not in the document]
n1zea144: SEG files we will have to leave separate based on the reference genome.  The meta file for segment files should have a reference_genome_id.

model/src/main/java/org/cbioportal/model/ClinicalDataModel.java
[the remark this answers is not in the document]
doori: i think the unique keys get populated, but in this case, the Study view APIs use ClinicalData to create lists of ClinicalDataCountItem or DataBin.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetConstants.java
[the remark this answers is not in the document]
doori: sure we can keep them separate. Using data_cna*.seg for SEG files and also added /studies/ directory in ParquetConstants as decided in today's meeting.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+    @Override
+    public BaseMeta fetchMetaClinicalDataInStudy(String studyId, List<String> ids, List<String> attributeIds, String clinicalDataType) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ClinicalData> fetchClinicalData(List<String> studyIds, List<String> sampleIds, List<String> attributeIds,
+                                                String clinicalDataType, String projection) {
+        List<Dataset<Row>> res = new ArrayList<>();
+        for (String studyId : new HashSet<>(studyIds)) {
```
kalletlak: Instead of iterating studies to read parquet file why not read all files at a time. 
something like `spark.read().parquet("file1", "file2",...)`. or may be a path pattern matching. not sure if it's supported in java spark library

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+                if (MUTATION_COUNT.equalsIgnoreCase(c)) {
+                    sub = sub.join(getMutationCount(studyId), "sampleId");
+                }
+                if (FRACTION_GENOME_ALTERED.equalsIgnoreCase(c)) {
+                    sub = sub.join(getFractionGenomeAltered(studyId), "sampleId");
+                }
+                res.add(sub);
+            }
+        }
+        Dataset<Row> ds = res.get(0);
```
kalletlak: could you add comments explaining why this is done. especially line 138

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+        for (Dataset<Row> r : res.subList(1, res.size())) {
+            ds = ds.unionByName(r);
+        }
+
+        Dataset<ClinicalDataModel> clinicalDataModel = ds.as(Encoders.bean(ClinicalDataModel.class));
+        List<ClinicalDataModel> dsls = clinicalDataModel.collectAsList();
```
kalletlak: better name for dsls

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+        List<ClinicalData> clinicalData = dsls.stream()
```
kalletlak: directly return instead of creating new variable

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+            spark.sqlContext().setConf("spark.sql.caseSensitive", "false");
+            String dataFile = ParquetConstants.CLINICAL_PATIENT;
+            if (PersistenceConstants.SAMPLE_CLINICAL_DATA_TYPE.equalsIgnoreCase(clinicalDataType)) {
+                dataFile = ParquetConstants.CLINICAL_SAMPLE;
+            } 
+            Dataset<Row> clinicalData = spark.read()
+                .option("mergeSchema", true)
+                .parquet(PARQUET_DIR + ParquetConstants.STUDIES_DIR + studyId + "/" + dataFile);
+
+            if ("PATIENT".equalsIgnoreCase(clinicalDataType)) {
```
kalletlak: add a comment explaining why sample clinical data is joined when queried for patient datatype

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+
+        // if attributes is null return data from all columns
+        List<String> attributes = attributeIds;
+        if (attributeIds == null || attributeIds.isEmpty()) {
+            attributes = Arrays.asList(clinicalData.columns());
+        }
+
+        clinicalData.createOrReplaceTempView("clinicalData");
+        for (String c : attributes) {
+            StringBuilder sb = new StringBuilder();
```
kalletlak: not sure if there is a better way to do instead of iterating for each attribute. response time increases as the number of attributes increases

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
doori: It is possible to load all files at a time, like I do in fetchClinicalDataCounts() method. I iterated through to set ClinicalData.studyId.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
doori: Sure I can add the comment. i'm doing this to union all results in datasets before calling collectAsList - which is an expensive call.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
doori: yeah, I'm not sure as i'm using attributeId for the value and group by.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
kalletlak: that should work

doori: which one would work? Do you prefer I change it to load all files at once or iterate through

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,280 @@
...
+                Dataset<Row> clinicalSampleData = spark.read()
+                    .option("mergeSchema", true)
+                    .parquet(PARQUET_DIR + ParquetConstants.STUDIES_DIR + studyId + "/" + ParquetConstants.CLINICAL_SAMPLE);
+                clinicalData = clinicalData.join(clinicalSampleData, "PATIENT_ID");
+            }
+
+            // filter by sample id
+            if (sampleIds != null && !sampleIds.isEmpty()) {
+                String[] sampleArr = sampleIds.stream().toArray(String[]::new);
+                clinicalData = clinicalData.filter(col("PATIENT_ID").isin(sampleArr).or(col("SAMPLE_ID").isin(sampleArr)));
```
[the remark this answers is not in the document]
doori: thanks for the catch. i added conditional based on the clinicaldatatype. From the front end it passes patientIds or sampleIds.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
doori: I tried adding union as part of the sql, it does not show a big impact on performance. I'm not sure if there is a way to do multiple group bys in the same select...clinical-data-bin-counts

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
kalletlak: load all files at once

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
kalletlak: Leave it as is for now until we find a better solution.

doori: ok

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
doori: As I refactor, I see that setting the studyId is necessary for StudyViewController.java line 387. clinicalDataMap = clinicalDataList.stream().collect(Collectors.groupingBy(ClinicalData::getSampleId, 
                Collectors.groupingBy(ClinicalData::getStudyId)));

kalletlak: you can try adding new column `STUDY_ID` to dataset

doori: sorry, I don't follow you. if i'm merging all studies, then how would i know which studyId to set in ClinicalData?

kalletlak: you could get study id from the input file name. something like this https://stackoverflow.com/a/40849244.

`.withColumn("input_file_name", getStudyId(input_file_name()));`

you will have to implement `getStudyId` to extract study if from path

doori: sure. updated the code.

kalletlak: nice. just curious, was there any performance improvement with this approach?

doori: I added https://user-images.githubusercontent.com/2739448/61876051-f12dbc80-aeb9-11e9-90e9-5432ae913a69.png in the other comment. Doesn't seem to have much impact on performance for 2 studies.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,285 @@
...
+            }, DataTypes.StringType);
+        
+        Dataset<Row> df = spark.read()
+            .option("mergeSchema", true)
+            .parquet(fileSeq)
+            .withColumn("studyId", getStudyId.apply(input_file_name()));
+        return df;
+    }
+
+    private Dataset<Row> getMutationCount(Set<String> studyIds) {
```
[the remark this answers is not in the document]
doori: i think it should be ok, as we join (inner by default) with the sampleId/patientId - filtered clinicaldata.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -0,0 +1,289 @@
...
+        
+        Set<String> studyIdSet = new HashSet<>(studyIds);
+        String dataFile = ParquetConstants.CLINICAL_PATIENT;
+        if (PersistenceConstants.SAMPLE_CLINICAL_DATA_TYPE.equalsIgnoreCase(clinicalDataType)) {
+            dataFile = ParquetConstants.CLINICAL_SAMPLE;
+        }
+        Dataset<Row> clinicalData = loadStudyFiles(studyIdSet, dataFile);
+        
+        // join with clinical sample data to get sample id information.
+        if ("PATIENT".equalsIgnoreCase(clinicalDataType)) {
```
kalletlak: create `PATIENT` constant similar to `SAMPLE_CLINICAL_DATA_TYPE`

doori: done

PR #6390, opened 2019-07-16
oncokb analysis backend support

zhx828:


model/src/main/java/org/cbioportal/model/OncoKBDataCountItem.java
```diff
@@ -0,0 +1,40 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+import java.util.List;
+
+public class OncoKBDataCountItem implements Serializable {
+
+    public enum OncoKBDataType {
```
zhx828: We don't need data type here. oncokb will always on sample since the annotation is on mutation level

web/src/main/java/org/cbioportal/web/MutationController.java
```diff
@@ -78,7 +78,9 @@
         @ApiParam("Name of the property that the result list is sorted by")
         @RequestParam(required = false) MutationSortBy sortBy,
         @ApiParam("Direction of the sort")
-        @RequestParam(defaultValue = "ASC") Direction direction) throws MolecularProfileNotFoundException {
+        @RequestParam(defaultValue = "ASC") Direction direction,
+        @ApiParam("Annotation")
+        @RequestParam(required = false) String annotation) throws MolecularProfileNotFoundException {
```
zhx828: We can now remove the annotation in this endpoint since it was a testing project.

service/src/test/java/org/cbioportal/service/impl/MutationServiceImplTest.java
```diff
@@ -176,17 +176,17 @@ public void fetchMetaMutationsInMolecularProfile() throws Exception {
...
-        MutationMeta result = mutationService.fetchMetaMutationsInMolecularProfile(MOLECULAR_PROFILE_ID, 
+        MutationMeta result = mutationService.fetchMetaMutationsInMolecularProfile(MOLECULAR_PROFILE_ID,
             Arrays.asList(SAMPLE_ID1), Arrays.asList(ENTREZ_GENE_ID_1));
 
         Assert.assertEquals(expectedMutationMeta, result);
     }
 
     @Test(expected = MolecularProfileNotFoundException.class)
     public void fetchMetaMutationsInMolecularProfileNotFound() throws Exception {
-
```
zhx828: You need to take a look the IDE you are using, it should understand the settings in the .editorconfig file.  We have `trim_trailing_whitespace = false` to force the IDE won't trim the trailing whitespace which is what happening here and many other places.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -1,256 +0,0 @@
-# app name
```
zhx828: The .EXAMPLE file should not be deleted.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,47 @@
...
         combinedResult.addAll(resultForSampleAttributes);
         combinedResult.addAll(resultForPatientAttributes);
         return new ResponseEntity<>(combinedResult, HttpStatus.OK);
     }
 
+    @PreAuthorize("hasPermission(#involvedCancerStudies, 'Collection<CancerStudyId>', 'read')")
+    @RequestMapping(value = "/oncokb/fetch", method = RequestMethod.POST,
+        consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+    @ApiOperation("Fetch annotation data by study view filter")
+    public ResponseEntity<List<ClinicalDataCountItem>> fetchAnnotationData(
```
zhx828: Here will be `OncoKBCountItem`

zhx828: Also let's call the method fetchOncoKBData

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,47 @@
...
+        @ApiParam(required = true, value = "Annotation data filter")
+        @Valid @RequestBody(required = false)  OncoKBDataCountFilter oncoKBDataCountFilter,
```
zhx828: And you are right. I don't think the RequestBody is needed here. Even `clinicalDataCountFilter` seems not used. Let's just leave here. I will figure it out later.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,47 @@
...
+        List<SampleIdentifier> filteredSampleIdentifiers = studyViewFilterApplier.apply(studyViewFilter);
+
+        List<OncoKBDataCountItem> combinedResult = new ArrayList<>();
+        if (filteredSampleIdentifiers.isEmpty()) {
+            return new ResponseEntity<>(combinedResult, HttpStatus.OK);
+        }
+        List<String> studyIds = new ArrayList<>();
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(filteredSampleIdentifiers, studyIds, sampleIds);
+        List<ClinicalDataCountItem> resultForSampleAttributes = clinicalDataService.fetchClinicalDataCounts(
```
zhx828: Here is basically need to get mutation data first then return OncoKBDataCountItem based on mutation data.

model/src/main/java/org/cbioportal/model/OncoKBDataCountItem.java
```diff
@@ -0,0 +1,27 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+import java.util.List;
+
+public class OncoKBDataCountItem implements Serializable {
+
+    private String attributeId;
+    private OncoKBDataType oncoKBDataType;
```
zhx828: Just a placeholder: we don't need this

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MutationMapper.xml
```diff
@@ -201,6 +202,31 @@
             LIMIT #{limit} OFFSET #{offset}
         </if>
     </select>
+    
+    <select id="getMutationsInMultipleMolecularProfilesByAnnotation" resultType="org.cbioportal.model.Mutation">
+        SELECT
+        <include refid="select"/>
+        <include refid="from"/>
+        <if test="annotation != null">
+          SELECT * FROM mutation WHERE mutation.annotation = #{annotation}
```
zhx828: This is incorrect. You already have `SELECT` in this statement. `mutation.annotation = #{annotation}` should be included in the WHERE clause which is in `whereInMultipleMolecularProfiles `

web/src/main/java/org/cbioportal/web/StudyViewController.java
zhx828: You should start to use `getMutationsInMultipleMolecularProfilesByAnnotation` in mutationService

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +96,46 @@
...
+        List<OncoKBDataCountItem> combinedResult = new ArrayList<>();
+        if (filteredSampleIdentifiers.isEmpty()) {
+            return new ResponseEntity<>(combinedResult, HttpStatus.OK);
+        }
+        List<String> studyIds = new ArrayList<>();
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(filteredSampleIdentifiers, studyIds, sampleIds);
+        List<Mutation> resultForSampleAttributes = mutationService.getMutationsInMultipleMolecularProfilesByAnnotation(molecularProfileService
+            .getFirstMutationProfileIds(studyIds, sampleIds), sampleIds, null, SUMMARY, null, null);
+        combinedResult.addAll(resultForSampleAttributes);
```
zhx828: Convert Mutation to OncoKBDataCountItem

web/src/main/java/org/cbioportal/web/util/OncoKBDataFilterApplier.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public List<SampleIdentifier> apply(List<SampleIdentifier> sampleIdentifiers,
+                                        List<OncoKBDataFilter> attributes,
+                                        Boolean negateFilters) {
+        List<Mutation> annotationDataList = new ArrayList<>();
+        if (!attributes.isEmpty() && !sampleIdentifiers.isEmpty()) {
+            List<String> studyIds = new ArrayList<>();
+            List<String> sampleIds = new ArrayList<>();
+            studyViewFilterUtil.extractStudyAndSampleIds(sampleIdentifiers, studyIds, sampleIds);
+            String annotationMap = "{";
```
zhx828: Considering use JSONObject

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,55 @@
...
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(filteredSampleIdentifiers, studyIds, sampleIds);
+        List<Mutation> resultForSampleAttributes = mutationService.getMutationsInMultipleMolecularProfiles(molecularProfileService
+            .getFirstMutationProfileIds(studyIds, sampleIds), sampleIds, null, SUMMARY, 10000000, 0, null, null);
+        Set<OncoKBDataCount> set = new HashSet();
+        for(Mutation mutationProfile : resultForSampleAttributes) {
+          JSONObject jsonMutation = new JSONObject(mutationProfile.getAnnotation());
+          OncoKBDataCount oncoKBCancer = new OncoKBDataCount();
+          oncoKBCancer.setAttributeId("oncogenic");
+          oncoKBCancer.setValue(jsonMutation.getJSONObject("oncogenic").getString());
```
zhx828: You need to check whether oncogenic exists in jsonMutation

model/src/main/java/org/cbioportal/model/OncoKBDataCount.java
```diff
@@ -0,0 +1,34 @@
+package org.cbioportal.model;
+
+import java.io.Serializable;
+
+public class OncoKBDataCount implements Serializable {
+
+    private String attributeId;
+    private String value;
+    private Integer count;
```
zhx828: Set it to 0 by default then you can calculate directly

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,55 @@
...
+          oncoKBCancer.setCount(oncoKBCancer.getCount()++);
```
zhx828: See comment in OncoKBDataCount. You cannot do the calculation without initialization

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,85 @@
...
+        }
+        List<String> studyIds = new ArrayList<>();
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(filteredSampleIdentifiers, studyIds, sampleIds);
+        List<Mutation> resultForSampleAttributes = mutationService.getMutationsInMultipleMolecularProfiles(molecularProfileService
+            .getFirstMutationProfileIds(studyIds, sampleIds), sampleIds, null, SUMMARY, 10000000, 0, null, null);
+        Set<OncoKBDataCount> set = new HashSet();
+        for(Mutation mutationProfile : resultForSampleAttributes) {
+          JSONObject jsonMutation = new JSONObject(mutationProfile.getAnnotation());
+          if(jsonMutation.getJSONObject("oncokb").has("oncogenic")) {
```
zhx828: The following if conditions can actually be combined since there is just one key difference

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -96,14 +97,85 @@
...
+        if (filteredSampleIdentifiers.isEmpty()) {
+            return new ResponseEntity<>(combinedResult, HttpStatus.OK);
+        }
+        List<String> studyIds = new ArrayList<>();
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(filteredSampleIdentifiers, studyIds, sampleIds);
+        List<Mutation> resultForSampleAttributes = mutationService.getMutationsInMultipleMolecularProfiles(molecularProfileService
+            .getFirstMutationProfileIds(studyIds, sampleIds), sampleIds, null, SUMMARY, 10000000, 0, null, null);
+        Set<OncoKBDataCount> set = new HashSet();
+        for(Mutation mutationProfile : resultForSampleAttributes) {
```
zhx828: You also need to do the check whether the attribute-value combination already exist in the set, other wise the count will always be 1.

web/src/main/java/org/cbioportal/web/util/OncoKBDataFilterApplier.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public List<SampleIdentifier> apply(List<SampleIdentifier> sampleIdentifiers,
+                                        List<OncoKBDataFilter> attributes,
+                                        Boolean negateFilters) {
+        List<Mutation> annotationDataList = new ArrayList<>();
+        if (!attributes.isEmpty() && !sampleIdentifiers.isEmpty()) {
+            List<String> studyIds = new ArrayList<>();
+            List<String> sampleIds = new ArrayList<>();
+            studyViewFilterUtil.extractStudyAndSampleIds(sampleIdentifiers, studyIds, sampleIds);
+            String annotationMap = "{ oncokb: {";
```
zhx828: The annotationMap should be constructed by using JSONObject

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -204,8 +209,18 @@
...
-                .getFirstMutationProfileIds(studyIds, sampleIds), sampleIds, null, true);
+            
+            // getSampleCountInMultipleMolecularProfiles is counting the number of mutations within filtered samples
+            // But that does not guarantee they are driver mutations.
+            // I created method getSampleCountInMultipleMolecularProfilesByAnnotation in the MutationMapper.
+            // Please consider using that when updating the following code
+            List<AnnotationFilter> annotationCancer;
+            AnnotationFilter annotationCase = new AnnotationFilter();
+            annotationCase.setPath("oncokb.oncogenicity");
+            annotationCase.setValues(new List<String>("YES", "NO", "YES,NO"));
```
zhx828: Here is the same logic of the work you did in filterByDriverMutation. AnnotationFilter should be generated based on  withOncoKBDriverMutationData. If Yes, the value will be converted  to proper oncogenicity value.

web/src/main/java/org/cbioportal/web/util/StudyViewFilterApplier.java
```diff
@@ -142,25 +148,70 @@ public SampleIdentifier apply(Sample sample) {
...
+        studyViewFilterUtil.extractStudyAndSampleIds(sampleIdentifiers, studyIds, sampleIds);
+        
+        List<String> oncogenicityIdentifier;
+        if(withOncoKBDriverMutationData == true) {
+          oncogenicityIdentifier = oncoKBConverter.getOncogenicityByHasDriver("YES");
+        }
+        if(withOncoKBDriverMutationData == false) {
+          oncogenicityIdentifier = oncoKBConverter.getOncogenicityByHasDriver("NO");
+        }
+        List<AnnotationFilter> annotationProfiles;
```
zhx828: AnnotationFilter should be populated by `oncogenicityIdentifier`.

web/src/main/java/org/cbioportal/web/util/StudyViewFilterApplier.java
```diff
@@ -142,25 +148,70 @@ public SampleIdentifier apply(Sample sample) {
...
+        AnnotationFilter annotationOncoKBData = new AnnotationFilter();
+        annotationOncoKBData.setPath("oncokb.oncogenicity");
+        annotationOncoKBData.setValues(new List<String>("YES", "NO", "YES,NO"));
```
zhx828: You are doing filtering, imaging you need to get a list of mutations which qualifies the oncogenicity values we are trying to filter. If `withOncoKBDriverMutationData ` is true, `oncoKBConverter.getOncogenicityByHasDriver` will return you ['oncogenic', 'likely oncogenic', 'predicted oncogenic']. The actual mutation oncogenicity should be one of these values. Take a look at the MutationMapper.xml, how the annotationProfiles is actually used.

web/src/main/java/org/cbioportal/web/util/StudyViewFilterApplier.java
```diff
@@ -142,25 +148,70 @@ public SampleIdentifier apply(Sample sample) {
...
+        // After getting oncogenicities, you need to construct the List<AnnotationFilter> for the function below
+        // The method need to used to get all qualified mutations: mutationService.getMutationsInMultipleMolecularProfilesByAnnotation()
+        // After getting the mutations, just get the list of samples from the mutations and return
+        List<String> studyIds = new ArrayList<>();
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(sampleIdentifiers, studyIds, sampleIds);
+        
+        List<String> oncogenicityIdentifier;
+        if(withOncoKBDriverMutationData == true) {
+          oncogenicityIdentifier = oncoKBConverter.getOncogenicityByHasDriver("YES");
```
zhx828:
```suggestion
          oncogenicityIdentifier = oncoKBConverter.getOncogenicityByHasDriver(withOncoKBDriverMutationData == true ? "YES":"NO");
```

jjgao:
... and this?

PR #6417, opened 2019-07-24
Fix dockerfile shenandoah

inodb:
jar binary moved to a different path, but is in `PATH` now, so no need to supply full path


fedde-s:
Sorry! I wasn't at work. But I trust your assessment there, it didn't look to complicated :+1:

Issue #6391, opened 2019-07-17
Failing co-expression endpoint (related to refactoring of CoExpressionServiceImpl)


pvannierop:
In `release-3.1.0` branch localdb e2e-test show that co-expression functionality is broken when using gene/gene set cross comparisons. We suspect it is related to refactoring to introduce use of `Iterable` for _MolecularDataService_ (done in [commit a645fbec15](https://github.com/cBioPortal/cbioportal/blob/a645fbec15dd52036a5f55e0ee2f569c85d5cc34/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java)).

The error is thrown from the `getCoExpressions` method. It appears that this is related to the use of Transactions in CoExpressionServiceImpl:
```
// transaction needs to be setup here in order to return Iterable from molecularDataService in fetchCoExpressions
    @Transactional(readOnly=true)
    public List<CoExpression> getCoExpressions(String geneticEntityId, CoExpression.GeneticEntityType geneticEntityType,
            String sampleListId, String molecularProfileIdA, String molecularProfileIdB, Double threshold)
            throws MolecularProfileNotFoundException, SampleListNotFoundException, GenesetNotFoundException,
            GeneNotFoundException {
```

The error is thrown from line 93 when the molecularDataService is being called:
```
} else {
            molecularDataListB = molecularDataService.getMolecularData(molecularProfileIdB, sampleListId, null,
                    "SUMMARY");
        }
```

The errors thrown are:
```
org.springframework.transaction.CannotCreateTransactionException: Could not create JDBC savepoint; nested exception is java.sql.SQLException: Connection is read-only. Queries leading to data modification are not allowed.
	org.springframework.jdbc.datasource.JdbcTransactionObjectSupport.createSavepoint(JdbcTransactionObjectSupport.java:112)
	org.springframework.transaction.support.AbstractTransactionStatus.createAndHoldSavepoint(AbstractTransactionStatus.java:144)
	org.springframework.transaction.support.AbstractPlatformTransactionManager.handleExistingTransaction(AbstractPlatformTransactionManager.java:458)
	org.springframework.transaction.support.AbstractPlatformTransactionManager.getTransaction(AbstractPlatformTransactionManager.java:349)
	org.springframework.transaction.interceptor.TransactionAspectSupport.createTransactionIfNecessary(TransactionAspectSupport.java:461)
[...]

PR #6420, opened 2019-07-24
Fix for 6391 - removed unnecessary readonly transactional annotation on repos method.

n1zea144:
Not sure why I put it in the original PR, but further testing reveals this annotation is not required.

Fixes https://github.com/cbioportal/cbioportal/issues/6391


persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/MolecularDataMyBatisRepository.java
```diff
@@ -34,7 +34,6 @@ public String getCommaSeparatedSampleIdsOfMolecularProfile(String molecularProfi
 
     @Override
     // cursor processing requires a transaction
```
inodb:
```suggestion
```

n1zea144: Doh!  Gimme a sec....

PR #6440, opened 2019-07-29
spark backend for filtered-sample api

doori:
Spark backend changes for api/filtered-samples/fetch

Reviewers
@n1zea144 @kalletlak


persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public BaseMeta fetchMetaSamples(List<String> sampleListIds) {
+        throw new UnsupportedOperationException();    }
+
+    @Override
+    public List<Sample> getSamplesByInternalIds(List<Integer> internalIds) {
+        throw new UnsupportedOperationException();
+    }
+
+    private Dataset<Row> loadStudyFiles(Set<String> studyIds, String file) {
```
kalletlak: looks similar to the one in `ClinicalDataSparkRepository.java`. may be better to move it to utils class.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
```diff
@@ -0,0 +1,197 @@
...
+        Dataset<Row> samples = loadStudyFiles(new HashSet<>(studyIds), ParquetConstants.CLINICAL_SAMPLE);
+        
+        if (!CollectionUtils.isEmpty(sampleIds)) {
+            samples = samples.where(samples.col("SAMPLE_ID").isin(sampleIds.toArray()));
+        }
+        
+        samples.createOrReplaceTempView("samples");
+        StringBuilder sb = new StringBuilder("SELECT PATIENT_ID, SAMPLE_ID, studyId");
+        if (PersistenceConstants.SUMMARY_PROJECTION.equalsIgnoreCase(projection)
+            || PersistenceConstants.DETAILED_PROJECTION.equalsIgnoreCase(projection)) {
```
kalletlak: This seems odd, in both cases `SAMPLE_TYPE` is appended. Could you cross-verify if this condition is correct.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
```diff
@@ -0,0 +1,197 @@
...
+    private Sample mapToSample(Row row, String projection) {
+        Sample sample = new Sample();
+        sample.setPatientStableId(row.getString(0));
+        sample.setStableId(row.getString(1));
+        String typeOrStudy = row.getString(2);
+        if (PersistenceConstants.SUMMARY_PROJECTION.equalsIgnoreCase(projection)
+            || PersistenceConstants.DETAILED_PROJECTION.equalsIgnoreCase(projection)) {
+            Sample.SampleType sampleType = getType(sample.getStableId(), typeOrStudy);
+            sample.setSampleType(sampleType);
+            sample.setCancerStudyIdentifier(row.getString(3));
```
kalletlak: looks like studyId is always third column

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
```diff
@@ -0,0 +1,197 @@
...
+            .option("mergeSchema", true)
+            .parquet(fileSeq)
+            .withColumn("studyId", getStudyId.apply(input_file_name()));
+    }
+    
+    private Sample mapToSample(Row row, String projection) {
+        Sample sample = new Sample();
+        sample.setPatientStableId(row.getString(0));
+        sample.setStableId(row.getString(1));
+        String typeOrStudy = row.getString(2);
```
kalletlak: `typeOrStudy` can be renamed to `sampleType` and moved inside `if` condition

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
```diff
@@ -0,0 +1,197 @@
...
+            sample.setSampleType(sampleType);
+            sample.setCancerStudyIdentifier(row.getString(3));
+        } else {
+            sample.setCancerStudyIdentifier(typeOrStudy);
+
+        }
+        return sample;
+    }
+
+    private Sample.SampleType getType(String stableId, String sampleType) {
```
kalletlak: This should be in a utils class and needs to be tested

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
```diff
@@ -71,13 +71,10 @@
...
             .mapToInt(Integer::parseInt).boxed().collect(Collectors.toList());
 
-        List<Sample> samples;
-        if (sampleIds == null) {
-            samples = sampleService.getSamplesByInternalIds(internalSampleIds);
-        } else {
-            List<String> studyIds = new ArrayList<>();
-            sampleIds.forEach(s -> studyIds.add(molecularProfile.getCancerStudyIdentifier()));
-            samples = sampleService.fetchSamples(studyIds, sampleIds, "ID");
+        List<Sample> samples = sampleService.getSamplesByInternalIds(internalSampleIds);
```
kalletlak: getSamplesByInternalIds returns `UnsupportedOperationException`
https://github.com/cBioPortal/cbioportal/pull/6440/files#diff-95d528fcc25c8f87f6828986b804f6eaR124

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
doori: The logic was based on from org.mskcc.cbio.portal.model.Sample.getType

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
doori: correct, the internalIds are based on MySQL, and not accessible from parquet data.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
doori: from SampleMapper.xml Line 12 <if test="projection == 'SUMMARY' || projection == 'DETAILED'">. If projection is 'ID' then sampleType won't be included.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
doori: thanks for catching this.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
doori: Added a util class and more JUnit test cases.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
doori: Agreed. I added a util class.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
kalletlak: ok

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
kalletlak: If this is based on MySQL then why are these changes required?

doori: Thanks for asking that question. Later in this method, `int indexOfSampleId = internalSampleIds.indexOf(sample.getInternalId());` it expects to find internalId of sample, which is MySQL specific. The implemented fetchSamples cannot retrieve internalIds from txt data and it would be duplicate effort to go back to MySQL just to fetch the internalId. The main difference I saw in the `if (sampleIds==null)/else` was that it was filtering by smapleIds if provided, so the fetchSample call was removed and filtering by sampleId was added.

doori: In [SampleServiceImpl](https://github.com/cBioPortal/cbioportal/blob/4fb750f4b3c6adeb6bf89b263568c4f304c0310e/service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java#L189), some code also depend on internalId, however it doesn't affect this endpoint (the line is only called with DETAILED projection). For sample-counts endpoint (PR to be created) I had to make changes so that the code doesn't depend on `internalId` and instead on `sampleId`. If this is a hard dependency (i.e. different samples can have same `sampleId` but different `internalId`), then it would not be appropriate to implement fetchSamples Parquet.

n1zea144: Unfortunately, only internal (MySQL) sample ids are unique across studies.  The stable id, like TCGA-A1-A0SB-01 is not unique across studies as the sample could be combined with samples from other studies into a new "merged" study.  We could consider the combined study-id/sample-id to be unique.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+    
+    private Sample mapToSample(Row row, String projection) {
+        Sample sample = new Sample();
+        sample.setPatientStableId(row.getString(0));
+        sample.setStableId(row.getString(1));
+        sample.setCancerStudyIdentifier(row.getString(2));
+        if (PersistenceConstants.SUMMARY_PROJECTION.equalsIgnoreCase(projection)
+            || PersistenceConstants.DETAILED_PROJECTION.equalsIgnoreCase(projection)) {
+            Sample.SampleType sampleType = SampleTypeUtil.getType(sample.getStableId(), row.getString(3));
```
n1zea144: The SampleTypeUtil borrows from [StableIdUtil](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/util/StableIdUtil.java) which is TCGA specific.  Did you consider using the [SampleType](https://github.com/cBioPortal/cbioportal/blob/master/model/src/main/java/org/cbioportal/model/Sample.java#L7) enum defined in the Sample Class?

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/ClinicalDataSparkRepository.java
```diff
@@ -32,10 +25,9 @@
     @Autowired
     private SparkSession spark;
 
-    @Value("${data.parquet.folder}")
-    private String PARQUET_DIR;
+    @Autowired
+    private ParquetLoader parquetLoader;
```
n1zea144: I like the use of this.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
doori: Do you mean adding the util method in the Sample class?

service/src/main/java/org/cbioportal/service/impl/GenesetDataServiceImpl.java
doori: Thanks for the confirmation. I added code to check both studyId/sampleId

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/SampleSparkRepository.java
n1zea144: I was thinking of moving the SampleTypeTypeHandler out of the mybatis package and generalizing it, but I see that it uses specific Mybatis types.  I do see you are using Sample.SampleType in your util class, which is the most important. Let just leave the way it is.

Issue #6424, opened 2019-07-25
Erron on co-expression service


alisman:
Following request returns 500 error.  Error occurs if you go to  https://bit.ly/2JSyzqf
and change the right side molecular profile dropdown.

[...]

PR #6450, opened 2019-07-30
Fix for 6424

n1zea144:
This is a fix for [6424](https://github.com/cBioPortal/cbioportal/issues/6424).  This is related to the introduction of the Cursor type in the MolecularDataMapper.  The original PR tried to reuse code in the MolecularDataMapper.xml file.  When the PR was first submitted, all tests past because not all execution paths were covered.  These errors have since come to light.  This PR addresses these errors by separating the use of MolecularDataMapper.xml:getGeneMolecularAlterations into geteGeneMolecularAlterations and getGeneMolecularAlterationsIter, the latter of which is used by the Cursor Implementation.

### DIFF
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
@@ -38,7 +38,38 @@
         ORDER BY REPLACE(genetic_profile.STABLE_ID, '_', ' '), genetic_profile.STABLE_ID ASC
     </select>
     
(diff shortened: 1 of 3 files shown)


persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -62,6 +63,36 @@
         </where>
     </select>
 
+    <!-- This routine is a copy of getGeneMolecularAlterations above.  This copy is necessary because
```
alisman: i'm curious, is there a difference between them?  i couldn't spot it!   or is a copy necessary for some other reason?

PR #6456, opened 2019-07-30
Code to precache GeneServiceImpl.getGene() data.

n1zea144:
Currently GeneMapper.getAllGenes() is the only repository query behind GeneServiceImpl that gets precached on service startup.

GeneMapper.getGeneByEntrezGeneId() is a repository query that services CoExpressionServiceImpl.computeCoExpressions().  When a request for co-expressions is made, fetching gene information can add a great amount of overhead to the call.  This PR attempts to speed up co-expression requests (as well as other requests that get gene info by entrez id) by precaching the GeneMapper.getGeneByEntrezGeneId() on GeneServiceImpl startup.

Precaching gene info by entrez id is very time consuming (as of this PR there is ~150,000 gene records - we can take ~2.5 hrs to process).  In light of this, precaching of gene info by entrez id is off-loaded to a worker thread from ForkJoinPool.commonPool which allows for rapid war file deployment.  This aspect of the implementation is up for discussion.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
@@ -7,21 +7,27 @@ import org.cbioportal.persistence.GeneRepository;
 import org.cbioportal.service.GeneService;
 import org.cbioportal.service.exception.GeneNotFoundException;
 import org.cbioportal.service.util.ChromosomeCalculator;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
-import javax.annotation.PostConstruct;
+import org.slf4j.*;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import java.util.concurrent.CompletableFuture;
+
+import javax.annotation.PostConstruct;
 
 @Service
 public class GeneServiceImpl implements GeneService {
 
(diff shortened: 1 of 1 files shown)


inodb:
@n1zea144 Nice! What's the speed-up approx?

I'm wondering a bit with future performance testing how to keep track of this. If this will take 2.5h to initialize it might be good to know when testing that you shouldn't compare the performance until this caching has been completed. Maybe we should have some section in the docs on caching?

n1zea144:
It can take ~10 minutes to complete a request for coexpressions when querying against a WES/WGS study and the gene data is not cached.  After cache, the request can be satisfied ~1 minute.

n1zea144:
Maybe the performance testing should be limited to targeted panel studies for ease of CI testing with the foreknowledge that the gene data will be cached in a production environment.

service/src/main/java/org/cbioportal/service/impl/GeneServiceImpl.java
```diff
@@ -33,10 +39,36 @@
...
         // query all genes so they would be cached
-        getAllGenes(null, null, "SUMMARY", null, null, null, null);
+        List<Gene> geneList = getAllGenes(null, null, "SUMMARY", null, null, null, null);
 
         geneAliasMap = geneRepository.getAllAliases().stream().collect(Collectors.groupingBy(
             GeneAlias::getEntrezGeneId, Collectors.mapping(GeneAlias::getGeneAlias, Collectors.toList())));
+
+        // the geneGene method is behind coexpressions, lets cache the data behind that method too
+        // run asynchronously as to speedup war deployment
+        if (!precachingHasStarted) {
```
jjgao: I am wondering if we can pre-cache when we start the war?

PR #6457, opened 2019-07-30
Json annotation of mutations

zhx828:
There is one commit in this pr actually is in master. Once the rc has the latest master, I will rebase.

The db migration code is not included yet. No so sure what version that should go. I will just leave it to backend.

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/MutationRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/MutationRepository.java
@@ -1,5 +1,6 @@
 package org.cbioportal.persistence;
 
+import org.cbioportal.model.AnnotationFilter;
 import org.cbioportal.model.Mutation;
 import org.cbioportal.model.MutationCountByPosition;
 import org.cbioportal.model.MutationCountByGene;
@@ -23,6 +24,11 @@ public interface MutationRepository {
                                                            Integer pageSize, Integer pageNumber,
                                                            String sortBy, String direction);
 
+    List<Mutation> getMutationsInMultipleMolecularProfilesByAnnotation(List<String> molecularProfileIds, List<String> sampleIds,
+                                                                       List<Integer> entrezGeneIds, String projection,
+                                                                       Integer pageSize, Integer pageNumber,
+                                                                       String sortBy, String direction, List<AnnotationFilter> filters);
+    
     MutationMeta getMetaMutationsInMultipleMolecularProfiles(List<String> molecularProfileIds, List<String> sampleIds,
                                                              List<Integer> entrezGeneIds);
 
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/MolecularDataMyBatisRepository.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/MolecularDataMyBatisRepository.java
@@ -36,13 +36,7 @@ public class MolecularDataMyBatisRepository implements MolecularDataRepository {
     public List<GeneMolecularAlteration> getGeneMolecularAlterations(String molecularProfileId, 
                                                                      List<Integer> entrezGeneIds, String projection) {
 
-        List<GeneMolecularAlteration> toReturn = new ArrayList();
-        Iterable<GeneMolecularAlteration> gmasItr =
-            molecularDataMapper.getGeneMolecularAlterations(molecularProfileId, entrezGeneIds, projection);
-        for (GeneMolecularAlteration gma : gmasItr) {
-            toReturn.add(gma);
-        }
-        return toReturn;
+        return molecularDataMapper.getGeneMolecularAlterations(molecularProfileId, entrezGeneIds, projection);
     }
 
     @Override
@@ -52,7 +46,7 @@ public class MolecularDataMyBatisRepository implements MolecularDataRepository {
     public Iterable<GeneMolecularAlteration> getGeneMolecularAlterationsIterable(String molecularProfileId, 
                                                                                  List<Integer> entrezGeneIds, String projection) {
 
-        return molecularDataMapper.getGeneMolecularAlterations(molecularProfileId, entrezGeneIds, projection);
+        return molecularDataMapper.getGeneMolecularAlterationsIter(molecularProfileId, entrezGeneIds, projection);
     }
 
     @Override
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
@@ -38,7 +38,38 @@
         ORDER BY REPLACE(genetic_profile.STABLE_ID, '_', ' '), genetic_profile.STABLE_ID ASC
     </select>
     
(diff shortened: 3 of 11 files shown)


persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -62,6 +63,36 @@
         </where>
     </select>
 
+    <!-- This routine is a copy of getGeneMolecularAlterations above.  This copy is necessary because
```
ao508: is there a way that you can move all of this into a `<sql> .... </sql>` element that is referenced by these queries using `include` as it is done [here](https://github.com/cBioPortal/cbioportal/pull/6457/files#diff-4020171d42dc9dc6eb6f0c4e437a48a0R110)

persistence/persistence-api/src/main/java/org/cbioportal/persistence/MutationRepository.java
```diff
@@ -23,6 +24,11 @@ MutationMeta getMetaMutationsInMolecularProfileBySampleListId(String molecularPr
                                                            Integer pageSize, Integer pageNumber,
                                                            String sortBy, String direction);
 
+    List<Mutation> getMutationsInMultipleMolecularProfilesByAnnotation(List<String> molecularProfileIds, List<String> sampleIds,
+                                                                       List<Integer> entrezGeneIds, String projection,
+                                                                       Integer pageSize, Integer pageNumber,
+                                                                       String sortBy, String direction, List<AnnotationFilter> filters);
+
```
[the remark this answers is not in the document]
ao508: Another question - would annotations from GN also be grouped into the JSON field or is this more for expanding the annotations available in MAFs/the portal

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
zhx828: @ao508 this is not from this pr.  It's introduced in https://github.com/cBioPortal/cbioportal/commit/685912300513b9b92512c54c9dfbb84080792411 I will rebase once the latest master is in rc branch.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/MolecularDataMyBatisRepository.java
```diff
@@ -36,13 +36,7 @@ public String getCommaSeparatedSampleIdsOfMolecularProfile(String molecularProfi
...
                                                                      List<Integer> entrezGeneIds, String projection) {
 
-        List<GeneMolecularAlteration> toReturn = new ArrayList();
-        Iterable<GeneMolecularAlteration> gmasItr =
-            molecularDataMapper.getGeneMolecularAlterations(molecularProfileId, entrezGeneIds, projection);
-        for (GeneMolecularAlteration gma : gmasItr) {
-            toReturn.add(gma);
-        }
-        return toReturn;
+        return molecularDataMapper.getGeneMolecularAlterations(molecularProfileId, entrezGeneIds, projection);
```
sheridancbio: could this refactoring be done in a separate PR? It seems like it would apply independently of the changes happening in the mutation model.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -62,6 +63,36 @@
...
+         it is backing a corresponding method in MolecularDataMapper.java which returns a Cursor as
+         opposed to a List. This method should be kept in sync with getGeneMolecularAlterations above.
+         (Attempts where made to share getGeneMolecularAlterations between methods in MolecularDataMapper.java
```
sheridancbio: where = were

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -62,6 +63,36 @@
...
+            </include>
+        </if>
+        FROM genetic_alteration
+        INNER JOIN genetic_profile ON genetic_alteration.GENETIC_PROFILE_ID = genetic_profile.GENETIC_PROFILE_ID
+        INNER JOIN gene ON genetic_alteration.GENETIC_ENTITY_ID = gene.GENETIC_ENTITY_ID
+        <where>
+            genetic_profile.STABLE_ID = #{molecularProfileId}
+            <if test="entrezGeneIds != null and !entrezGeneIds.isEmpty()">
+                AND gene.ENTREZ_GENE_ID IN
+                <foreach item="item" collection="entrezGeneIds" open="(" separator="," close=")">
```
sheridancbio: These "IN (.., .., ..,)" clauses can be slow. Maybe this case of selecting the genes requested by a user is not too problematic, but we also might consider doing this filtering at the service layer after querying across all genes. That way the same query is cached once and used for all searches on a particular genetic profile.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MutationMapper.xml
```diff
@@ -151,6 +152,61 @@
         </where>
     </sql>
 
+    <sql id="whereInMultipleMolecularProfilesByAnnotation">
+        <where>
+            <if test="sampleIds != null and !sampleIds.isEmpty()">
+                mutation.SAMPLE_ID IN (
```
sheridancbio: same comment as before. Maybe we want to filter out the unwanted samples at the service layer so that we can cache and reuse a single query per profile.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MutationMapper.xml
```diff
@@ -151,6 +152,61 @@
...
+            </if>
+            <if test="sampleIds == null">
+                genetic_profile.STABLE_ID IN
+                <foreach item="item" collection="molecularProfileIds" open="(" separator="," close=")">
+                    #{item}
+                </foreach>
+            </if>
+            <if test="filters != null and !filters.isEmpty()">
+                <foreach item="filter" collection="filters">
+                    AND JSON_EXTRACT(mutation.annotation, #{filter.path}) in
```
sheridancbio: Here is the logic where the AnnotationFilter is applied ... the JSON is extracted and we are reading out (it appears) a single field ... and looking for matches to a set of supplied values

jjgao:
@zhx828 what's the plan with this?

zhx828:
@jjgao never got a chance to come back work on this ticket.

jjgao:
@zhx828 @averyniceday has this already been used in the FACETS/ASCN implementation?

zhx828:
@jjgao I believe FACETS is using a separate column in the table. @averyniceday could comment on the details.

averyniceday:
@zhx828 @jjgao I need to look into this PR more but off the top of my head the ASCN data does not going into the annotations column in the mutation table. It is being imported into a new allele_specific_copy_number table.

jjgao:
@averyniceday got it. Thanks!

jjgao:
@zhx828 I am wondering how much is left to complete this?

PR #6458, opened 2019-07-31
Code for posting requests to google analytics

n1zea144:
Interceptor posts request to cBioPortal website that occur outside of website to google analytics collect service.  Interceptor is enabled via following JVM argument:

-Dgoogle.analytics.tracking=ga-tracking-enabled

Requires the following properties set:

Tracking ID (Format UX-XXXX-X)
google.analytics.tracking.code.api=

Client id - it can be any valid UUID style format (e.g. XXXXXXXX-XXXXX-XXXX-XXXX-XXXXXXXXXXXXX)
google.analytics.application.client.id=

Updates to documentation to follow....

### DIFF
--- a/web/src/main/resources/applicationContext-web.xml
+++ b/web/src/main/resources/applicationContext-web.xml
@@ -67,5 +67,10 @@
     <mvc:interceptors>
         <bean id="involvedCancerStudyInterceptor" class="org.cbioportal.web.util.InvolvedCancerStudyExtractorInterceptor"/>
     </mvc:interceptors>
+    <beans profile="ga-api-tracking-enabled">
+      <mvc:interceptors>
+        <bean id="googleAnalyticsInterceptor" class="org.cbioportal.web.util.GoogleAnalyticsInterceptor"/>
+      </mvc:interceptors>
+    </beans>
 
 </beans>
--- a/portal/src/main/webapp/WEB-INF/web.xml
+++ b/portal/src/main/webapp/WEB-INF/web.xml
@@ -23,7 +23,7 @@
     <!-- dbcp profile is defined here to prevent exceptions during war deployment -->
     <context-param>
         <param-name>spring.profiles.active</param-name>
-        <param-value>${dbconnector:dbcp},${authenticate}</param-value>
+        <param-value>${dbconnector:dbcp},${authenticate},${google.analytics.tracking:ga-api-tracking-disabled}</param-value>
     </context-param>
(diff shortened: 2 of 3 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -183,9 +189,6 @@ pathway_commons.url=http://www.pathwaycommons.org/pc2
 # the new API uses the v3 of bitly API, and a java library to make the API call, so you only need to provide the access token
 bitly.access.token=
 
-# google analytics
```
alisman: this property is in use by existing frontend ga tracking.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -87,6 +87,12 @@ filter_groups_by_appname=true
 googleplus.consumer.key=
 googleplus.consumer.secret=
 
+# To enable Google Analytics on the endpoints, add the appropriate Tracking ID (Format UX-XXXX-X)
```
alisman: it may not be necessary to document these properties unless we think other installers will really want to track usage of their api.    it causes confusion with the frontend google analytics config property below (which might well be in use by others) so perhaps we keep it as hidden feature?  especially until we can be sure it doesn't have performance impact.

web/src/main/resources/applicationContext-web.xml
```diff
@@ -67,5 +67,10 @@
     <mvc:interceptors>
         <bean id="involvedCancerStudyInterceptor" class="org.cbioportal.web.util.InvolvedCancerStudyExtractorInterceptor"/>
     </mvc:interceptors>
+    <beans profile="ga-tracking-enabled">
```
alisman: ga-api-tracking-enabled?  i think we should be careful to distinguish api tracking from existing frontend ga implementation

src/main/resources/portal.properties.EXAMPLE
n1zea144: I was wondering about this too.  This is fine with me.

web/src/main/resources/applicationContext-web.xml
n1zea144: Understood.  Easy change.

src/main/resources/portal.properties.EXAMPLE
alisman: @n1zea144 this one needs to be go back in

PR #6273, opened 2019-06-14
Docker Docs Fixes

inodb:
Update image name. Fix JAVA_OPTS parameter for deploying

PR #6464, opened 2019-08-01
Adjust documentation on new Docker image

oplantalech:
* Remove /cbioportal from the Docker example commands and Keycloak
configuration: the new reccomended Docker deployment and the alternative
with webapp-runner do not use it (see
09f473f and #6273)
* Do not enforce to change the word "Role" into "roles" in Keycloak
(less error-prone)


fedde-s:
I think it would be very helpful for future reference if the commit message explained _why_ the `/cbioportal/` prefix should be removed here. A reference to the pull request that changes the documentation on the recommended way to deploy should do.

oplantalech:
@fedde-s The PR changing the documentation is this one: https://github.com/cBioPortal/cbioportal/pull/6046. I can include the reference to the commit.

fedde-s:
I was thinking that references to commit 09f473f6f50fa7b622e76455c7e6368b2b236906 (which recommends the Docker documentation as the primary way to deploy) and pull request #6273 (which makes the same adjustment in other parts of the documentation) would be useful, but either's fine.

oplantalech:
Thanks for the feedback, @fedde-s! I included your changes.

fedde-s:
@jjgao, would anyone else's feedback be needed before merging this? I believe that it's an obvious fix plus an easy improvement.

PR #6475, opened 2019-08-05
spark backend for sample-counts api

doori:
spark backend for api/sample-counts/fetch.

Reviewers:
@kalletlak @n1zea144


persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetWriter.java
```diff
@@ -35,26 +35,76 @@
...
+                df = spark.read()
+                    .format("csv")
+                    .option("delimiter", ":")
+                    .load(inputFile);
+
+                it = df.toLocalIterator();
+                while (it.hasNext()) {
+                    Row row = (Row) it.next();
+                    String colName = row.getString(0);
+                    String listCol = "case".equalsIgnoreCase(typeOfData) ? "case_list_ids" : "gene_list";
```
n1zea144: Does this value ever change?

doori: i see both [case_list_ids](https://docs.cbioportal.org/5.1-data-loading/data-loading/file-formats#case-lists) and [gene_list](https://docs.cbioportal.org/5.1-data-loading/data-loading/import-gene-panels#gene-panel-file-format) is in the documentation.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -36,12 +37,18 @@
     @Qualifier("sampleSparkRepository")
     private SampleRepository sampleSparkRepository;
     @Autowired
+    private CopyNumberSegmentSparkRepository copyNumberSegmentSparkRepository;
```
n1zea144: Is this not implementing the CopyNumberSegmentRepos?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -190,4 +196,30 @@ private void processSamples(List<Sample> samples, String projection) {
             });
         }
     }
+
+    private void processSamplesSpark(List<Sample> samples, String projection) {
```
n1zea144: Is the only difference here wit processSamples() the call to the spark repos?  If so, suggest passing a ref to the repos into a generalized method.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
doori: This is another case where the method returns internalIds and filters based on that. (`List<Integer> samplesWithCopyNumberSeg = copyNumberSegmentRepository.fetchSamplesWithCopyNumberSegments()`) So the workaround was to create a method that returns `List<String> sampleIds` and create another method in SampleSeriviceImpl.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
doori: That is not the only difference. It's the persistence interface returning internalIds which is something we cannot get from Parquet.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetWriter.java
n1zea144: I mean once this routine is invoked, will listCol be assigned the same value each time through the while-loop?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
n1zea144: Why not introduce a method to the interface which returns a List<String> (or return List<Sample> and have the service process internal ids or stable sample ids as required)?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
n1zea144: See comment above.

persistence/persistence-spark/src/main/java/org/cbioportal/persistence/spark/util/ParquetWriter.java
doori: Thanks for clarifying. moved it out of the while-loop.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
doori: thanks for the suggestion. I added a method in the interface that returns a list of Samples.

PR #6490, opened 2019-08-09
Add correction step for CNA metafile

### LINKED ISSUE #5023: CNA metadata file validation issues
Only `DISCRETE` CNA profiles should have the option to set `show_profile_in_analysis_tab: true`, but currently it is possible to set other CNA profiles (like `CONTINUOUS`) to appear in the query page. 

Desired solution:
* validation script should give an `ERROR` for any non-`DISCRETE` CNA profile which has the configuration `show_profile_in_analysis_tab: true`
* or: data loading script should ignore the `show_profile_in_analysis_tab` setting for non-`DISCRETE` CNA profiles, setting it to `false` in DB when loading.

Fixing the query results page to only use `DISCRETE` CNA profile is also desirable, although less urgent. Can be logged in a separate ticket.

SRodenburg:
# What? Why?
Fix #5023 

<s>Add a correction step in the metafile parser, setting `show_profile_in_analysis_tab: false` if the CNA datatype is not `DISCRETE`. Also shows an INFO message if that's the case.</s>

- Validator throws error when `show_profile_in_analysis_tab: true` is encountered in CNA data that is not `DISCRETE`.
- Unit test for this step


SRodenburg:
I think resolved.

SRodenburg:
All set

SRodenburg:
@jjgao Can you merge?

PR #6492, opened 2019-08-09
Add special ehcahce for holding gene data

sheridancbio:
- rename primary cache alias to GeneralRepositoryCache
- new cache named SpecialRepositoryCacheOne .. size approximated to hold genes in use
- disk persistence set to false in all cache configurations
- expiry policy set to none

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -220,13 +220,13 @@ ucsc.build=hg19
 oncoprint.defaultview=patient
 
 # OncoPrint driver mutation annotations
-#   oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
-#   oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
-#   oncoprint.custom_driver_annotation.binary.default=true
-#   oncoprint.custom_driver_annotation.tiers.default=true
-  oncoprint.oncokb.default=true
-  oncoprint.hotspots.default=true
-#  oncoprint.hide_vus.default=true
+# oncoprint.custom_driver_annotation.binary.menu_label=Custom driver annotation
+# oncoprint.custom_driver_annotation.tiers.menu_label=Custom driver tiers
+# oncoprint.custom_driver_annotation.binary.default=true
+# oncoprint.custom_driver_annotation.tiers.default=true
+oncoprint.oncokb.default=true
+oncoprint.hotspots.default=true
+# oncoprint.hide_vus.default=true
 
 # Custom gene sets
 # querypage.setsofgenes.location=file:/<path>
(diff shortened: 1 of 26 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -238,10 +238,17 @@ ehcache.xml.configuration=/ehcache-mixed.xml
 ehcache.enable.statistics=true
 ehcache.persistence.path=
 
-ehcache.RepositoryCache.maxBytesHeap=100
-ehcache.RepositoryCache.maxBytesHeapUnits=MB
-ehcache.RepositoryCache.maxBytesLocalDisk=250
-ehcache.RepositoryCache.maxBytesLocalDiskUnits=MB
+# size allocation preferences for the general persistence cache
```
averyniceday: for the general cache maybe we should default to 1GB heap and X GB disc since testing showed that having heap did improve performance.

web/src/test/java/org/cbioportal/web/CacheStatsControllerTest.java
```diff
@@ -72,32 +73,46 @@ public static CacheStatisticsService cacheStatisticsService() throws CacheNotFou
...
 
     @Test
     public void testGetKeyCountsPerClassInvalidCache() throws Exception {
-        mockMvc.perform(MockMvcRequestBuilders.get("/" + INVALID_CACHE_ALIAS + "/keyCountsPerClass")
+        mockMvc.perform(MockMvcRequestBuilders.get("/" + invalidCacheAlias() + "/keyCountsPerClass")
             .accept(MediaType.APPLICATION_JSON))
             .andExpect(MockMvcResultMatchers.status().isNotFound());
     }
 
+    private String invalidCacheAlias() {
```
averyniceday: Not sure I understand why we changed the valid caches to a list of cache names and added this method to generate an invalid cache name. Wouldn't it be okay to keep the existing tests?

ao508: agreed ^

ao508:
My other suggestion is to rename `Special*` cache to something else that may indicate that this cache is retained/persisted and not refreshed as frequently as the `GeneralRepositoryCache`

src/main/resources/portal.properties.EXAMPLE
sheridancbio: Done ... I set 1GB for heap and 4GB for disk

web/src/test/java/org/cbioportal/web/CacheStatsControllerTest.java
sheridancbio: reverted

sheridancbio:
I talked with @mandawilson and we thought StaticRepositoryDiskCacheOne would be more informative than "Special". Update made.

PR #6502, opened 2019-08-13
Adjustments to Ehcache settings

sheridancbio:
- add getAllGenes() and getAliasesOfGeneByEntrezGeneId() and getAliasesOfGeneByHugoGeneSymbol() to static cache
- stop caching a once-only persistence call getAllAliases()
- increase max-object-graph-size to Long.MAX_VALUE (from 10000000) and include in ehcache-disk-only settings
- increase heap resource for static cache to 30 MB (from 3 MB)
- decrease disk resource for static cache to 32 MB (from 4 GB)

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
@@ -41,7 +41,7 @@ import java.util.List;
 
 public interface GeneRepository {
 
-    @Cacheable("GeneralRepositoryCache")
+    @Cacheable("StaticRepositoryCacheOne")
     List<Gene> getAllGenes(String keyword, String alias, String projection, Integer pageSize, Integer pageNumber, String sortBy, 
                            String direction);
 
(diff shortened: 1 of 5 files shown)


persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
```diff
@@ -57,13 +57,13 @@
...
-    @Cacheable("GeneralRepositoryCache")
+    @Cacheable("StaticRepositoryCacheOne")
     List<String> getAliasesOfGeneByEntrezGeneId(Integer entrezGeneId);
 
-    @Cacheable("GeneralRepositoryCache")
+    @Cacheable("StaticRepositoryCacheOne")
     List<String> getAliasesOfGeneByHugoGeneSymbol(String hugoGeneSymbol);
 
-    @Cacheable("GeneralRepositoryCache")
+    // not cached because this is called only a single time, during @PostConstruct method o GeneServiceImpl
```
ao508: 👍

PR #6505, opened 2019-08-14
fixing null point exception when queried MT genes not found in the re…

### LINKED ISSUE #6504: Null point exception from Expression Web Service when queried mitochondrial genes not found from Reference Genome Genes DB table
Currently Enrichment Expression Web Service throws a null point exception when queried genes such as mitochondrial genes are not found in the reference genome gene table. 
Screen Shot 2019-08-08 at 4 29 33 PM

khzhu:
…ference genome gene DB table

# What? Why?
Fix #6504 

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/ExpressionEnrichmentServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/ExpressionEnrichmentServiceImpl.java
@@ -22,13 +22,13 @@ import org.cbioportal.model.MolecularAlteration;
 import org.cbioportal.model.MolecularProfile;
 import org.cbioportal.model.MolecularProfile.MolecularAlterationType;
 import org.cbioportal.model.MolecularProfileCaseIdentifier;
-import org.cbioportal.model.ReferenceGenomeGene;
+import org.cbioportal.model.Gene;
 import org.cbioportal.model.Sample;
 import org.cbioportal.persistence.MolecularDataRepository;
 import org.cbioportal.service.ExpressionEnrichmentService;
 import org.cbioportal.service.MolecularDataService;
 import org.cbioportal.service.MolecularProfileService;
-import org.cbioportal.service.ReferenceGenomeGeneService;
+import org.cbioportal.service.GeneService;
 import org.cbioportal.service.SampleService;
 import org.cbioportal.service.exception.MolecularProfileNotFoundException;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -48,7 +48,7 @@ public class ExpressionEnrichmentServiceImpl implements ExpressionEnrichmentServ
     @Autowired
     private MolecularDataService molecularDataService;
     @Autowired
-    private ReferenceGenomeGeneService refGeneService;
+    private GeneService geneService;
     @Autowired
     private MolecularDataRepository molecularDataRepository;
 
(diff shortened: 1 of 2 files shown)


khzhu:
@kalletlak , can I please have your review. Thanks!

service/src/main/java/org/cbioportal/service/impl/ExpressionEnrichmentServiceImpl.java
```diff
@@ -137,9 +137,16 @@
                 .collect(Collectors.groupingBy(ReferenceGenomeGene::getEntrezGeneId));
 
         expressionEnrichments.forEach(expressionEnrichment -> {
-            ReferenceGenomeGene gene = geneMapByEntrezId.get(expressionEnrichment.getEntrezGeneId()).get(0);
-            expressionEnrichment.setHugoGeneSymbol(gene.getHugoGeneSymbol());
-            expressionEnrichment.setCytoband(gene.getCytoband());
+            try {
```
kalletlak: Should we remove this(setting Hugo gene symbol and cytoband), since we are getting gene info for all genes in the front-end.

khzhu: let me check if we back fill cytobands for expression in frontend.

khzhu: you are right. We can remove set up for cytobands (it is done in, but we still need to set hugoSymbols.

kalletlak: ok. I think there are couple of ways to do it
1. we could you `hugoSymbols` from the response of `reference-genome-genes` api
2. set `hugoSymbols` similar to what's done for mutation or copy number enrichments api

khzhu: @kalletlak , we have to set hugoSymbols for all alterations in the backend, not in frontend. Please check the function makeEnrichmentDataPromise see if you agree. 
```
                // add cytoband from reference gene
                for (const d of data) {
                    const refGene = params.referenceGenesPromise.result![d.hugoGeneSymbol];

                    if (refGene)
                        d.cytoband = refGene.cytoband;
                }
```

khzhu: @kalletlak , changes made. please have your final review. thanks!

service/src/main/java/org/cbioportal/service/impl/ExpressionEnrichmentServiceImpl.java
```diff
@@ -131,15 +131,21 @@
...
                 .stream()
-                .collect(Collectors.groupingBy(ReferenceGenomeGene::getEntrezGeneId));
+                .collect(Collectors.groupingBy(Gene::getEntrezGeneId));
 
         expressionEnrichments.forEach(expressionEnrichment -> {
-            ReferenceGenomeGene gene = geneMapByEntrezId.get(expressionEnrichment.getEntrezGeneId()).get(0);
-            expressionEnrichment.setHugoGeneSymbol(gene.getHugoGeneSymbol());
-            expressionEnrichment.setCytoband(gene.getCytoband());
+            Gene gene = geneMapByEntrezId.get(expressionEnrichment.getEntrezGeneId()).get(0);
+            if (gene != null) {
```
kalletlak: I think gene won't be null any time, because only validated data would be loaded in to database

kalletlak:
@khzhu I think you need to update tests

service/src/main/java/org/cbioportal/service/impl/ExpressionEnrichmentServiceImpl.java
kalletlak: if its null then you should remove that particular `expressionEnrichment` object from response

PR #6507, opened 2019-08-14
Reduce response size of gene panel data for ProfiledSamplesCounter (fixes issue #6461)

### LINKED ISSUE #6461: performance issue for genie
GENIE study view is slow. Query is also kind of slow. I am wondering if it’s due to all the gene panels in genie. The implementation on gene panels are not optimized.

ao508:
# What? Why?
Fix #6461

ProfiledSamplesCounter was returning detailed gene panel data for each altered gene in the response. Matching gene panels are returned for every altered gene on the study summary page which included every single gene in every matching gene panel found.

This is problematic for large studies with many altered genes, samples, and associated gene panels. The call for mutated genes resulted in very large data transfers (4GB for genie_private study) which ultimately crash the study summary page.

The ProfiledSamplesCounter response now excludes the list of genes linked to a gene panel and instead only includes the meta information (gene panel stable id and description). For the beta genie study (genie_private) this reduced the size of the response from 4GB to 2.4MB and the study summary page now renders as expected.

Co-authored by: Manda Wilson <1458628+mandawilson@users.noreply.github.com>
 

# Checks
- [ ] Runs on Heroku.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/util/ProfiledSamplesCounter.java
+++ b/service/src/main/java/org/cbioportal/service/util/ProfiledSamplesCounter.java
@@ -56,7 +56,11 @@ public class ProfiledSamplesCounter {
                 List<GenePanel> matchingGenePanels = geneGenePanelMap.get(entrezGeneId);
                 for (GenePanel genePanel : matchingGenePanels) {
                     numberOfSamplesProfiled += genePanelDataMap.get(genePanel.getStableId()).size();
-                    allPanels.add(genePanel);
+                    GenePanel genePanelSummary = new GenePanel();
+                    genePanelSummary.setDescription(genePanel.getDescription());
+                    genePanelSummary.setInternalId(genePanel.getInternalId());
+                    genePanelSummary.setStableId(genePanel.getStableId());
+                    allPanels.add(genePanelSummary);
                 }
                 
                 numberOfSamplesProfiled += profiled.stream().filter(g -> g.getGenePanelId() == null).count();


jjgao:
@zhx828 previously fixed this for another endpoint: https://github.com/cBioPortal/cbioportal/pull/6341. I am wondering if there are other endpoints to be fixed? and whether the two fixes are consistent?

zhx828:
@rohangoel96 @sheridancbio  frontend uses the gene panel data in study view to calculate the frequency. And in my fix https://github.com/cBioPortal/cbioportal/pull/6341 we no longer return the panel info anymore except the gene panel id. Then the frontend pulls the gene panel info through gene panel endpoints by using the id. 
I'm so sure how the summary data will be used, but seems like a reasonable addition?

sheridancbio:
I know that @mandawilson is away for this week and will not be able to submit a review rapidly. I have dropped the request for review from her.

A quick comment to respond to @jjgao : @ao508 did search the other endpoints hit by the study view looking for similar data expansions. The only other problematic endpoint (for copy number counts via http://www.cbioportal.org/api/cna-genes/fetch) was also corrected through the modification to this shared service layer utility.

sheridancbio:
@jjgao @zhx828 : I also looked at #6341. My understanding after a quick look is:

in #6341 the per-gene response originally included a list of fully modeled gene panels. That was changed to return a set of strings holding the gene panel identifiers.

The case we are fixing in this PR is similar, so we could alter the model of the response here to also return a simple set of strings holding gene panel identifiers. The frontend code would also need to be changed if the gene panel information is used in the front end for the frequency calculation. It seems as though the numerator and denominator for the frequency calculation may both already be part of the api response in this case ... but I may be wrong.

Maybe it would be good for us to deploy the current fix as a hotfix which does not modify the current api response schema, and then work on a fix similar to #6341 ?  Or does the frontend actually need (currently) the full gene panel membership list ?  We did test the current code in this PR and the study view looked correct (all content seemed the same).

zhx828:
@sheridancbio it does require the full list of gene panel info. A good way to test is to hover the frequency cell of mutated genes table. There is a tooltip to lead to a full gene panel modal popup. And this is way we cannot do a hotfix, instead we sent #6341 to rc(currently in release-3.1)

jjgao:
@ao508 @sheridancbio @zhx828 maybe worth a quick discussion to resolve this?

sheridancbio:
@jjgao @zhx828 @n1zea144 @averyniceday @ao508 @mandawilson Ok .. we think we understand a little better now. The fix in #6341 has already accomplished what this PR was trying to do (reduce the size of the return from the study-view endpoint). The code from #6341 has been merged to rc and to release-3.1.0, but not yet merged to master.

Until this is merged to master, the loading of the larger studies (such as genie-7.x) will not be possible because of the multi-GB response bodies. If we can wait for the public release of release-3.1.0, then we only need to wait. If we want a faster (hotfix) release of this fix we should push the code in #6341 into master and redeploy the production portals.

I guess we can close this PR now.

PR #6508, opened 2019-08-15
Drop MyBatis caching and use EHCache instead

sheridancbio:
# What? Why?
Portals running at MSK have been showing progressive memory use creep and have been crashing during garbage collection sweeps. This has been tracked to unbounded growth of mybatis cache size (which was limited only by count of entries rather than memory use footprint). This PR replaces mybatis caching with Ehcache, and limits the size of cache memory use. Three alternative configuration modes are provided.

- drop MyBatis configurations
- initial configuration options for EHCache, with some portal.properties settings
    - three ehcache settings files available (selectable from portal.properties):
        - ehcache-heap-only.xml
        - ehcache-disk-only.xml
        - ehcache-mixed.xml
    - disk persistence set to false in all cache configurations
    - expiry policy set to none
    - max-object-graph-size set to Long.MAX_VALUE (from 10000000)
- mark repository functions as @Cacheable
- enable cache event logging
- track cache statistics
- two caches available
    - StaticRepositoryCacheOne - holds only some GeneRepository call values and sized approximately to hold all genes in actual use (30MB)
    - GeneralRepositoryCache - holds all other cached repository calls
Add /api/cacheStatistics endpoint for reporting cache content / sizes
- endpoint can be enabled/disabled via portal.properties
- adjust GlobalExceptionHandler to return a 404 NotFound status for requests on an invalid cache alias
- store requested cache name in CacheNotFoundException
- add unit tests for cache statistics controller
    - mock CacheStatisticsService to throw CacheNotFoundException when requests for unknown cache alias are made
    - test keysInCache endpoint and keyCountsPerClass endpoints with valid and invalid cache aliases
Error fixed in cgds.sql (order of DROP statements must follow foreign key constraints)
Other notes
- xalan dependency excluded due to xml parsing errors
- old core/src/main/resources/ehcache.xml config file dropped

Co-authored-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>
Co-authored-by: Benjamin Gross <benjamin.gross@gmail.com>
Co-authored-by: Robert Sheridan <7747489+sheridancbio@users.noreply.github.com>

# Checks
- [ ] Runs on Heroku.


sheridancbio:
Rebasing / squashing to a single commit now

sheridancbio:
🔴 Note : at time of merge, updates to portal.properties must be checked in to the mskcc mercurial repo holding configuration settings.

docs/portal.properties-Reference.md
```diff
@@ -280,3 +281,62 @@ This gene set will add the following in the query box:
...
+
+If the cache is configured to use disk resources, users must make a directory available and set it with the `ehcache.persistence_path` property. Ehcache will create seperate directories under the provided path for each cache defined in the ehcache.xml_configuration file. 
+```
+ehcache.persistence_path=[location on the disk filesystem where Ehcache can write the cache to /tmp/]
+```
+
+Cache size must be set for both heap and disk; Ehcache requires disk size to be greater than heap size. Default values are provided. The general repository cache is specified to use 1021MB for heap and 4GB for disk. The static repository cache is specified to use 30MB for heap and 32MB for disk. For installations with increased traffic or data, cache sizes can be increased to further improve performance. 
+```
+ehcache.general_repository_cache.max_bytes_heap=
+ehcache.general_repository_cache.max_bytes_heap_units=[size unit e.g MB, GB, entries]
```
zhx828: Personally, I prefer less property settings to prevent issues. I don't see anyone would want to set this to B, or TB so maybe default to MB? If they want to increase to GB level, just need to do the math?

docs/portal.properties-Reference.md
```diff
@@ -280,3 +281,62 @@ This gene set will add the following in the query box:
...
+ehcache.general_repository_cache.max_bytes_local_disk=
+ehcache.general_repository_cache.max_bytes_local_disk_units=[size unit e.g MB, GB, entries]
+
+ehcache.static_repository_cache_one.max_bytes_heap=
+ehcache.static_repository_cache_one.max_bytes_heap_units=[size unit e.g MB, GB, entries]
+ehcache.static_repository_cache_one.max_bytes_local_disk=
+ehcache.static_repository_cache_one.max_bytes_local_disk_units[size unit e.g MB, GB, entries]
+```
+
+To configure a system where no caching occurs, use ehache-heap-only.xml with minimal heap sizes (0 is not supported, so set max_bytes_heap=1 and max_bytes_heap_units=B for all caches) and set
```
zhx828: It just seems weird to me if I set the option cache_enabled to false, and I still need to configure the max_bytes_heap and max_bytes_heap_units to the minimum. Does that mean if I disable the cache and max_bytes_heap accidentally sets to 10GB, the ehcache will still work?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CacheEventLogger.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2019 Memorial Sloan-Kettering Cancer Center.
```
zhx828: I don't think these MSK Copyright still applicable anymore. For the newly created files, we should avoid adding them.

persistence/persistence-api/src/main/resources/ehcache-disk-only.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
zhx828: It seems to me we could just have one ehcache.xml configuration file. Can we inject if/else conditions in the xml files that controlled by properties like `turn.on.disk.cache` and `turn.on.heap.cache`? This could avoid some duplications in the configuration. From what I see, they are almost identical.

sheridancbio: There is probably a way to do that. We can look into it.

db-scripts/src/main/resources/cgds.sql
```diff
@@ -97,9 +97,8 @@ DROP TABLE IF EXISTS `type_of_cancer`;
 DROP TABLE IF EXISTS `geneset_hierarchy_leaf`;
 DROP TABLE IF EXISTS `geneset_hierarchy_node`;
 DROP TABLE IF EXISTS `geneset`;
-DROP TABLE IF EXISTS `genetic_entity`;
-DROP TABLE IF EXISTS `reference_genome_gene`;
```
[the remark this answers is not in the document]
sheridancbio: It is true .. this is a small bugfix that helped with our unit testing. When using an actual persistent external test database, the tables must be dropped in the correct order before setting them up for the next run (the foreign key constraints must be respected with innodb). This fix  could be a separate PR, but since it does not alter the schema and does not seem to affect many others it seemed ok to just fix it as part of this work.

docs/portal.properties-Reference.md
```diff
@@ -280,3 +281,62 @@ This gene set will add the following in the query box:
...
 ```
+# Ehcache Settings
+cBioPortal is supported on the backend with Ehcache. The configuration, size, and location of these caches are configurable from within portal.properties through the following properties.
+
+First, select a cache configuration using `ehcache.xml_configuration`. This specifies whether to use a disk-only, heap-only, or hybrid (disk + heap) cache. The default value is `ehcache-mixed.xml` which initializes a hybrid caching system.
+```
+ehcache.xml_configuration=[ehcache-heap-only.xml OR ehcache-mixed.xml OR ehcache-disk-only.xml]
+```
+
+If the cache is configured to use disk resources, users must make a directory available and set it with the `ehcache.persistence_path` property. Ehcache will create seperate directories under the provided path for each cache defined in the ehcache.xml_configuration file.
```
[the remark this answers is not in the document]
sheridancbio: Yes, if the configuration is ehcache-disk-only.xml or ehcache-mixed.xml and there is no path, there will be an exception. If the configuration is ehcache-heap-only, then ehcache.persistence_path can be left blank.

docs/portal.properties-Reference.md
sheridancbio: When the user disables Ehcache (ehcache.cache_enabled=false) they still must give a valid configuration, because cache_enabled only means that items are never put in the cache or retrieved from the cache ... not that the cache does not exist. So in the documentation we wrote, we recommend setting a heap-only configuration and setting the size to something minimal (1B). I guess 1MB is not that much space to waste (and maybe Ehcache does not allocate any memory until items are actually stored in the cache). So we could go with MB as the default unit size if we are willing to possibly waste 1 MB of heap space. Anyone have comments?

docs/portal.properties-Reference.md
sheridancbio: If we are able to do conditional configuration of the ehcache-*.xml file as you suggest below, then we can handle this programatically. When ehcache.cache_enabled is false, it could ignore any of the storage footprint sizes and just configure a 1B heap only cache so that ehcache would not throw an exception on startup.
If we cannot do that, maybe we should just give the deployer a fourth configuration file (ehcache-disabled.xml) file with hardcoded values and tell the user to set ehcache.cache_enabled to false and set ehcache.xml_configuration to "ehcache-disabled.xml" (any other resource settings then are ignored due to hardcoding)

sheridancbio:
Thanks for the review @zhx828 ... these are good suggestions. We will try to work them in.

docs/portal.properties-Reference.md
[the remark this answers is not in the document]
zhx828: Is there a way to check whether the path is specified when building the war file? If so, can we throw an exception at that time?

zhx828:
@sheridancbio thanks for the detailed reply.

sheridancbio:
@zhx828 ... we have now simplified the configuration options:
- all sizes are in MB units
- a mixed (heap/disk) cache is presumed now ... deployers who want to use heap only or disk only must edit the single Ehcache configuration file (ehcache.xml) to drop the disk or heap usage (comments are provided in the file)
- when ehcache.cache_enabled is set to false, all other settings are ignored and a minimal (1 byte) heap configuration is automatically used ... no .xml file needed

Would you like to look over again and see if this PR is ready to be merged now?

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -220,20 +220,35 @@ ucsc.build=hg19
...
+# oncoprint.hide_vus.default=true
 
 # Custom gene sets
 # querypage.setsofgenes.location=file:/<path>
-#
-# Disable cache (for non authorized portals this can sometimes avoid running
-# out of memory with relatively little performance loss)
-# mybatis.cache.enabled=false
+
+# to disable use of Ehcache, set ehcache.cache.enabled to false, use ehcache.xml_configuration=/ehcache-heap-only.xml with ehcache.general_repository_cache.max_bytes_heap=1 with units=B
```
n1zea144: This is a bit confusing - Why not say something like:
```
# to disable use of EhCache, uncomment the following four lines and comment all the other lines starting with ehcache 
#ehcache.cache.enabled=false
#ehcache.xml_configuration=/ehchache-heap-only.xml
#ehcache.general_repository_cach.max_bytes_heap=1
#ehcache.general_repository.max_bytes_heap_units=B
```

n1zea144:
> Thanks so much for the effort! @sheridancbio I'm really impressed by the changes you guys made in a short peroid of time. Just one note, I think in most cases when user uses a docker image to start the application, the disk is not available. cc @inodb for confirmation. Even in the current public portal setup, we don't have a disk available yet until that's specifically configured. Do you think it makes sense to use /ehcache-heap-only.xml as the default?

Good point @zhx828.  We definitely should put some thought into this.  We can avoid having to guess what a good/working default configuration may be if we turned it off.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GeneRepository.java
```diff
@@ -35,30 +35,43 @@
 import org.cbioportal.model.GeneAlias;
 import org.cbioportal.model.meta.BaseMeta;
 
+import org.springframework.cache.annotation.Cacheable;
+
 import java.util.List;
 
 public interface GeneRepository {
 
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
```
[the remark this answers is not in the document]
sheridancbio: thanks ... the explanation (for future reference) is that this was the first static cache we introduced (where we expect no turnover), but we anticipate future "side" caches for performance tuning .. and we thought that we might be adding additional data types into any one of them, so we did not want to give a datatype specific name (like StaticGeneCache).

persistence/persistence-api/src/main/java/org/cbioportal/persistence/GenesetHierarchyRepository.java
```diff
@@ -5,11 +5,16 @@
 import org.cbioportal.model.Geneset;
 import org.cbioportal.model.GenesetHierarchyInfo;
 
+import org.springframework.cache.annotation.Cacheable;
+
 public interface GenesetHierarchyRepository {
 
+    @Cacheable(cacheNames = "GeneralRepositoryCache", condition = "@cacheEnabledConfig.getEnabled()")
```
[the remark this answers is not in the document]
sheridancbio: For now we just included genes (individual and the full list), but we did consider the possibility of adding other related data which would be co-resident. (which is why there is a fairly generic name for this cache). If we see performance issues which point at gene set loading as a problem, we might add it to this cache.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomKeyGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(CustomKeyGenerator.class);
+
+    public Object generate(Object target, Method method, Object... params) {
+        if (!cacheEnabledConfig.isEnabled()) {
+            return "";
+        }
+        String key = target.getClass().getSimpleName() + "_"
+            + method.getName() + "_"
+            + StringUtils.arrayToDelimitedString(params, "_");
```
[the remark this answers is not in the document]
sheridancbio: We do have the cache statistics API endpoint which will report the set of current keys. I don't have statistics, but it must be true that there are some recurring queries which pass a list of tens of thousands of sample identifiers. However, often the response from the persistence layer will be fully modeled objects which also contain the sample ids plus additional information, so in general I would expect the size of the keys to be smaller than the size of the values. Still, it would be good to measure this in a public portal with the cache statistics endpoint turned on.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -253,4 +268,12 @@ oncoprint.defaultview=patient
 
 # MatchMiner AWS url
 matchminer.url=
-matchminer.token=
\ No newline at end of file
+matchminer.token=
+
+# enable/disable hrrs logging
```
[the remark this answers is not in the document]
sheridancbio: apparently not ... or it was accidentally deleted at some point. These settings are not present in the master branch portal.properties.EXAMPLE

src/main/resources/portal.properties.EXAMPLE
sheridancbio: This is no longer necessary. With the latest changes, all a user has to do is to set ehcache.cache_enabled to false --- the rest will then be ignored

src/main/resources/portal.properties.EXAMPLE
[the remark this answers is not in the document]
inodb: hrrs is only in release-3.1.0 branch. So this configuration should prolly be left out of this PR

sheridancbio:
@zhx828 we have made the requested change to a default heap-only configuration and updated the documentation and comments accordingly. Now out of the box, the default config has 5120MB of Heap for caching.

docs/Caching.md
```diff
@@ -0,0 +1,17 @@
+# Backend Caching
```
inodb: To make this show up on docs.cbioportal.org add it to https://github.com/cBioPortal/cbioportal/blob/master/docs/SUMMARY.md and https://github.com/cBioPortal/cbioportal/blob/master/docs/README.md

docs/portal.properties-Reference.md
```diff
@@ -280,3 +281,58 @@ This gene set will add the following in the query box:
...
+```
+ehcache.xml_configuration=
+```
+
+If the cache is configured to use disk resources, users must make a directory available and set it with the `ehcache.persistence_path` property. Ehcache will create separate directories under the provided path for each cache defined in the ehcache.xml_configuration file. 
+```
+ehcache.persistence_path=[location on the disk filesystem where Ehcache can write the cache to /tmp/]
+```
+
+Cache size must be set for heap and/or disk depending which are in use; Ehcache requires disk size to be greater than heap size in a hybrid configuration. Zero is not a supported size and will cause an exception. Units are in megabytes. Default values are provided. The general repository cache is specified to use 5120MB of heap. The static repository cache is specified to use 30MB of heap. For installations with increased traffic or data, cache sizes can be increased to further improve performance.
```
inodb: Nice description! Only minor comment is that 5120MB of heap might be quite big - our current hardware requirements say a minimum of 2GB of RAM is required to run the portal. Maybe the default should be to not use any caching? Like only use if you opt in and know what you're doing kind of thing?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.beans.factory.annotation.Value;
+
+public class CustomEhCachingProvider extends EhcacheCachingProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CustomEhCachingProvider.class);
+
+    @Value("${ehcache.xml_configuration:/ehcache.xml}")
+    private String xmlConfiguration;
+
+    @Value("${ehcache.cache_enabled}")
```
inodb: good to use defaults here with `:false` or `:true` - what if somebody is using an old properties file without these properties defined? Currently it would crash because this is not defined. Unfortunately spring's default error logging is very bad when it comes to this for those unfamiliar with java. It won't give a clear error that a property/argument is missing

persistence/persistence-api/src/main/resources/applicationContext-ehcache.xml
```diff
@@ -0,0 +1,42 @@
...
+    xmlns:cache="http://www.springframework.org/schema/cache"
+    xsi:schemaLocation="
+    http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-4.0.xsd
+    http://www.springframework.org/schema/cache http://www.springframework.org/schema/cache/spring-cache.xsd
+    http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-4.0.xsd">
+
+    <cache:annotation-driven cache-manager="cacheManager" key-generator="customKeyGenerator"/>
+
+    <bean id="cacheEnabledConfig" class="org.cbioportal.persistence.CacheEnabledConfig">
+        <constructor-arg value="${ehcache.cache_enabled}"/>
```
inodb: also here good to set the default (apart from in `portal.properties.EXAMPLE`)

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -220,20 +220,35 @@ ucsc.build=hg19
...
+ehcache.statistics_enabled=true
+# Enable cache statistics endpoint for Ehcache monitoring
+cache.statistics_endpoint_enabled=false
+
+# Properties for specifying cache size constraints
+# - Zero is not an accepted value, will break deployment
+# - disk size (if used) must be greater than heap size
+
+ehcache.general_repository_cache.max_mega_bytes_heap=5120
+ehcache.static_repository_cache_one.max_mega_bytes_heap=30
```
inodb: these defaults should prolly also be set in code. Ideally people should be able to run the new code without having to update their properties file. A good CI test would be to pull the existing EXAMPLE properties file from the branch one is running against, in this case `master`, compile & run with that properties file and see if one is able to boot the portal.

.travis.yml
```diff
@@ -77,6 +77,7 @@ script:
           -Dfinal.war.name=cbioportal \
           -Ddb.user=cbio_user \
           -Ddb.password=somepassword \
+          -Dehcache.xml.configuration=/ehcache.xml \
```
inodb: why does this parameter need to be added - can't the default be e.g. `classpath:/ehcache.xml`?

inodb:
> Thanks so much for the effort! @sheridancbio I'm really impressed by the changes you guys made in a short peroid of time. Just one note, I think in most cases when user uses a docker image to start the application, the disk is not available. cc @inodb for confirmation. Even in the current public portal setup, we don't have a disk available yet until that's specifically configured. Do you think it makes sense to use /ehcache-heap-only.xml as the default?

Second that this is some awesome work :)!

@zhx828 @n1zea144 @sheridancbio  In the public portal when using a docker image in the k8s cluster you can write to `/`, but it's unclear what the limit is in terms of disk space. It will prolly run out of disk space when the sum of all pods running on the node goes beyond the default ec2 disk space.

Therefore for our own purposes we can change the k8s configuration to add a volume with a specified size and write to that one. 

It might be good to add some documentation in this PR to the hardware requirements about the heap caching and disk space usage as well: https://docs.cbioportal.org/2.1-deployment/hardware-requirements

sheridancbio:
Thanks for the review @inodb ... we'll review your requests and try to accommodate. We have discussed some of the tradeoffs of configuration in our group ... we may need to have further discussions, but I think you make good points.

inodb:
Thanks @sheridancbio ! Yeah I don't think they are essential changes but might make upgrading a little smoother for some

docs/Caching.md
sheridancbio: done

docs/portal.properties-Reference.md
sheridancbio: We handled this by having Ehcache be disabled with the default configuration. Now if users want to use caching, the set the property in portal.properties and uncomment the lines setting disk / heap usage. (and maybe adjust)

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
sheridancbio: done

persistence/persistence-api/src/main/resources/applicationContext-ehcache.xml
sheridancbio: done

src/main/resources/portal.properties.EXAMPLE
sheridancbio: done

.travis.yml
sheridancbio: this line is no longer needed for travis testing because the cache is now disabled ... in addition, the default was available too

sheridancbio:
@inodb We think we have implemented all of your requested changes ... if you approve, we will merge.

sheridancbio:
Removing @n1zea144 from list of requested reviewers

src/main/resources/portal.properties.EXAMPLE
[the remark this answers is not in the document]
sheridancbio: removed

docs/portal.properties-Reference.md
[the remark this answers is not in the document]
sheridancbio: we did not add a build-time check, but we did add a default path the supplied "ehcache.xml" file. (Now there is only one file supplied with the default code)

docs/portal.properties-Reference.md
sheridancbio: this is now not required ... if ehcache.cache_enabled is false, other properties don't need to be set.

persistence/persistence-api/src/main/resources/ehcache-disk-only.xml
sheridancbio: done

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomKeyGenerator.java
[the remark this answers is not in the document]
sheridancbio: we plan to have the endpoint turned on for the genie public portal

sheridancbio:
during final testing, we discovered that the attempt to set default values in ehcache.xml for the following properties was not being properly handled by maven filtering or spring properties injection.
- ehcache.statistics_enabled
- ehcache.persistence_path
- ehcache.general_repository_cache.max_mega_bytes_heap
- ehcache.general_repository_cache.max_mega_bytes_local_disk
- ehcache.static_repository_cache_one.max_mega_bytes_heap
- ehcache.static_repository_cache_one.max_mega_bytes_local_disk

 For now, we have dropped the defaults out of ehcache.xml ... if we are able to discover a way to get these default values to work correctly, we will add them back later. For users who use the portal.properties.EXAMPLE file settings, or who omit all of the ehcache settings, they will build a portal where ehcache use is disabled (and hence, ehcache.xml will not be in use at all). This is because ehcache.statistic_enabled is properly injected into the java class EhcacheCachingProvider through the @Value annotation.

PR #6516, opened 2019-08-20
raise logger level to error for unknown gene panels

SRodenburg:
Gene panel IDs are checked with the API whether they are already present. If not, a `warning` was raised and later on the importer crashes if this warning is overridden.

<b>Changes:</b>
- Raise logger `warning` to `error` in validator.


SRodenburg:
@jjgao Can you merge?

PR #6520, opened 2019-08-21
Add portal.properties setting to disable all ehcaching

sheridancbio:
# What? Why?
Ability to disable Ehcache caching requested

- CacheDisableSetting bean added to hold setting
    - provide verbatim setting value, inverse, or boolean
- Add condition attribute to cacheable annotations, which will omit caching depending on the CacheDisableSetting
- Add check for disabled caching to CustomKeyGenerator (no key computed if caching is disabled)

# Checks
- [ ] Runs on Heroku.

### DIFF
--- a/docs/Caching.md
+++ b/docs/Caching.md
@@ -1,5 +1,5 @@
 # Backend Caching 
-cBioPortal provides the option of caching information on the backend as a means for improving performance. Without caching, each query including repeated queries, will query the database, process the returned data, and construct a response. This may lead to performance issues as the entire process is rather costly, especially for queries on larger studies. With caching turned on, the query response will be stored during the initial query. If the same query is made, the response will be taken directly from cache instead of having to be reconstructed. 
+cBioPortal provides the option of caching information on the backend to improve performance. Without caching, every time a request is received by the backend, a query is sent to the database system for information, and the returned data is processed to construct a response. This may lead to performance issues as the entire process can be rather costly, especially for queries on larger studies. With caching turned on, query responses can be taken directly from the cache if they have already been constructed. They would only be constructed for the initial query.
 
 ## Cache Configuration
 The portal is configured to use Ehcache for backend caching; caching configuration is
@@ -7,15 +7,14 @@ specified inside an xml file under `persistence/persistence-api/src/main/resourc
 three different configurations available - `disk-only`, `heap-only`, and `mixed`. The configuration files 
 specify which caches to create. Additional specifications such as cache size and location are set inside `portal.properties` (more information [here](portal.properties-Reference.md#ehcache-settings)).
  
- 
  ## Creating additional caches
-The default configuration initializes two separate caches. However, a user may wish to introduce new caches with different policies (e.g expiration policy) for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the ehcache.xml configuration file. The `@Cacheable` annotation must also be added to function declarations to indicate which functions are to be cached. 
+The default configuration initializes two separate caches. However, you may wish to introduce new caches with different policies (e.g expiration policy) for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the ehcache.xml_configuration file. The `@Cacheable` annotation must also be added (or adjusted) to function declarations to indicate which functions are to be cached. Those might look like this example:
 ``` 
-@Cacheable(ClinicalDataCache)
+@Cacheable(cacheNames = "ClinicalDataCache", condition = "@cacheEnabledConfig.getEnabled()")
 public String getDataFromClinicalDataRepository(String param) {}
 ```
-Additionally, new properties for setting cache sizes should be added to `portal.properties`. Alternatively, values may also be hardcoded directly into the ehcache.xml configuration file. 
+Additionally, new properties for setting cache sizes should be added to `portal.properties`. Alternatively, values may also be hardcoded directly into the ehcache.xml_configuration file. 
 
-For more information on constructing an ehcache.xml configuration file, refer to the documentation [here](https://www.ehcache.org/documentation/3.7/xml.html). 
+For more information on constructing an ehcache.xml_configuration file, refer to the documentation [here](https://www.ehcache.org/documentation/3.7/xml.html). 
 
 For more information on linking caches to functions, refer to the documentation [here](https://spring.io/guides/gs/caching/).
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -230,13 +230,14 @@ oncoprint.hotspots.default=true
 
 # Custom gene sets
 # querypage.setsofgenes.location=file:/<path>
-#
-# Disable cache (for non authorized portals this can sometimes avoid running
-# out of memory with relatively little performance loss)
-# mybatis.cache.enabled=false
-ehcache.xml.configuration=/ehcache-mixed.xml
-ehcache.enable.statistics=true
-ehcache.persistence.path=
+
+# to disable use of Ehcache, set ehcache.cache.enabled to false, use ehcache.xml_configuration=/ehcache-heap-only.xml with ehcache.general_repository_cache.max_bytes_heap=1 with units=B
+ehcache.cache_enabled=true
+ehcache.xml_configuration=/ehcache-mixed.xml
+ehcache.persistence_path=/tmp
+ehcache.statistics_enabled=true
+# Enable cache statistics endpoint for Ehcache monitoring
+cache.statistics_endpoint_enabled=false
 
 # size allocation preferences for the general persistence cache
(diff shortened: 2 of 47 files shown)


persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
```diff
@@ -0,0 +1,40 @@
+package org.cbioportal.persistence;
+
+import org.springframework.beans.factory.annotation.Value;
+
+public class CacheDisableSetting {
```
averyniceday: is there a reason we introduced `inverse`/`booleanValue` as additional data members instead of just using `setting`?

averyniceday: if we have to keep to additional data members `inverse` and `booleanValue`, I would rename them to something for descriptive (especially `inverse`)

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -234,6 +234,7 @@ oncoprint.hotspots.default=true
 # Disable cache (for non authorized portals this can sometimes avoid running
 # out of memory with relatively little performance loss)
 # mybatis.cache.enabled=false
+ehcache.disable_all_caching=false
```
averyniceday: maybe follow the convention for the mybatis property? something like `ehcache.cache.disabled`

docs/Caching.md
```diff
@@ -1,17 +1,16 @@
 # Backend Caching 
-cBioPortal provides the option of caching information on the backend as a means for improving performance. Without caching, each query including repeated queries, will query the database, process the returned data, and construct a response. This may lead to performance issues as the entire process is rather costly, especially for queries on larger studies. With caching turned on, the query response will be stored during the initial query. If the same query is made, the response will be taken directly from cache instead of having to be reconstructed. 
+cBioPortal provides the option of caching information on the backend to improve performance. Without caching, every time a request is received by the backend, a query is sent to the database system for information, and the returned data is processed to construct a response. This may lead to performance issues as the entire process can be rather costly, especially for queries on larger studies. With caching turned on, query responses can be taken directly from the cache if they have alredy been constructed. They only would be constructed for the initial query.
```
ao508: Typo:

> With caching turned on, query responses can be taken directly from the cache if they have alredy been constructed.

With caching turned on, query responses can be taken directly from the cache if they have **already** been constructed.

ao508: Rephrase:

> They only would be constructed for the initial query.

They would only be constructed for the initial query.

docs/Caching.md
```diff
@@ -1,17 +1,16 @@
...
  
- 
  ## Creating additional caches
-The default configuration initializes two separate caches. However, a user may wish to introduce new caches with different policies (e.g expiration policy) for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the ehcache.xml configuration file. The `@Cacheable` annotation must also be added to function declarations to indicate which functions are to be cached. 
+The default configuration initializes two separate caches. However, you may wish to introduce new caches with different policies (e.g expiration policy) for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the ehcache.xml configuration file. The `@Cacheable` annotation must also be added (or adjusted) to function declarations to indicate which functions are to be cached. Those might look like this example:
 ``` 
-@Cacheable(ClinicalDataCache)
+@Cacheable(cacheNames = "ClinicalDataCache", condition = "@cacheDisableSetting.getInverse()")
```
ao508: What does `@cacheDisableSetting.getInverse()` here mean?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Value("${ehcache.disable_all_caching:false}")
+    private String setting;
```
ao508: `setting` here can be automatically parsed as a boolean if you change the declaration to `private Boolean setting;`

ao508: I also think that the member should be renamed to something like `ehcacheDisabled`

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
ao508: Agreed ^ having the additional `inverse` member is a bit confusing.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public String getInverse() {
+        if (inverse == null) {
+            initializeOtherDataMembers();
+        }
+        return inverse;
+    }
+
+    public boolean getBoolean() {
```
ao508: maybe also rename this to `isCachingDisabled()` or something like that which would return the value. That way I think we could get rid of these other methods in here as well. The way that the property setting and how the inverse is determined and returned is a bit more complicated that it needs to be - I recommend simplifying.

docs/Caching.md
ao508: Never mind, I see that it's a method from the `CacheDisableSetting` class

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
```diff
@@ -0,0 +1,40 @@
+package org.cbioportal.persistence;
```
ao508: instead of naming this class `CacheDisableSetting` might I suggest a different name like `DisabledCacheConfig` or similar?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
ao508: or if the method is renamed to something like 

`isCachingEnabled()` then the return statement could be
```
return !ehcacheDisabled
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
sheridancbio: I've changed the setting from disabled to enabled .. so I will remove the inverse member
The booleanValue is to save the backend the time of many times converting string "false" into a boolean false value (or "true" to true)

src/main/resources/portal.properties.EXAMPLE
sheridancbio: I've renamed it to ehcache.cache_enabled and used underscores for other settings like this (trying to standardize) :
ehcache.xml.configuration -> ehcache.xml_configuration
ehcache.enable.statistics -> ehcache.statistics_enabled
cache.statistics.endpoint.enabled -> ehcache.statistics_endpoint_enabled
ehcache.persistence.path -> ehcache.persistence_path

docs/Caching.md
sheridancbio: changes made

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
sheridancbio: sounds good .. renaming

sheridancbio: actually, using CacheEnabledConfig .. to match the portal.properties setting "ehcache.cache_enabled"

ao508: sounds good to me!

src/main/resources/portal.properties.EXAMPLE
sheridancbio: after talking with @averyniceday I went back to cache.statistics_endpoint_enabled

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
sheridancbio: the member is now a simple boolean called "enabled"

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheDisableSetting.java
sheridancbio: I now have two access functions:
- getEnabled()  (returns "true" or "false")
- isEnabled() (returns true or false)

sheridancbio:
Removing @mandawilson from review requests. Also dismissing change requests from @ao508, who has now approved this PR

PR #6572, opened 2019-09-06
Multiple Reference Genome Migration - defaults and warnings

### LINKED ISSUE #6566: Multiple Reference Genome Migration - defaults and warnings
The 3.1.0-release includes multiple reference genome support —  part of this support includes migrating the database to include a reference genome for existing studies. The studies are assigned a reference genome id depending on the NCBI_BUILD present in mutation event records associated with the study.

The migration assumes only one NCBI_BUILD to be returned; however, this is not always the case (potentially due to incorrect data in the database). 

Requested improvements are:

1. make the default reference genome id during migration hg19 (to match the default import behavior when reference genome build is not specified). The current default hg38.

1. modify the import_db.py script so that a check is performed before doing the migration step which introduces multi-reference genome support. The check will examine all mutation_event records (on a study-by-study basis) and see whether the NCBI_BUILD values  are consistent. If not, a warning with details is displayed to the user and the program exits (unless a "--force" command line argument had been provided when script is run). The warning should indicate which genome build would have been set if forced (genome build setting may come from the portal.properties file)

cc: @cBioPortal/importer-pipelines

khzhu:
# What? Why?
Fix #6566 

- make the default reference genome build during migration hg19 (to match the default import behavior when reference genome build is not specified).
- modify the import_db.py script so that a check whether the NCBI_BUILD values are consistent is performed before doing the migration step which introduces multi-reference genome support.

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?

### DIFF
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -165,7 +165,6 @@ def check_reference_genome(portal_properties, cursor, force_migration):
                            join mutation on mutation.MUTATION_EVENT_ID = mutation_event.MUTATION_EVENT_ID 
                            join genetic_profile on genetic_profile.GENETIC_PROFILE_ID = mutation.GENETIC_PROFILE_ID
                            join cancer_study on cancer_study.CANCER_STUDY_ID = genetic_profile.CANCER_STUDY_ID
-                           where NCBI_BUILD not in ('NA','')
                            group by CANCER_STUDY_IDENTIFIER
                        """
         cursor.execute(sql_statement)
@@ -183,9 +182,9 @@ def check_reference_genome(portal_properties, cursor, force_migration):
                         sys.exit(1)
                 elif int(ref_count) > 1:
                     msg = """
-                            WARNING: %s contains %s reference genomes. Please clean up the mutation_event table and ensure it only contains one valid reference genome (%s).
+                            WARNING: %s contains %s reference genomes (%s). Please clean up the mutation_event table and ensure it only contains one valid reference genome (%s).
                             OR use the "--force" option to surpress the warning, then the default reference genome (%s) listed in the portal.properties file will be used.
-                        """%(study, ref_count, ','.join(ALLOWABLE_GENOME_REFERENCES), default_genome_build)
+                        """%(study, ref_count, retrieved_ncbi_build,','.join(ALLOWABLE_GENOME_REFERENCES), default_genome_build)
                     print(msg, file=ERROR_FILE)
                     sys.exit(1)
     except MySQLdb.Error as msg:
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -22,11 +22,13 @@ DATABASE_USER = 'db.user'
 DATABASE_PW = 'db.password'
 VERSION_TABLE = 'info'
 VERSION_FIELD = 'DB_SCHEMA_VERSION'
+GENOME_BUILD = 'ucsc.build'
+ALLOWABLE_GENOME_REFERENCES = ['37', 'hg19', 'GRCh37', '38', 'hg38', 'GRCh38','mm9','GRCm37','mm10','GRCm38']
 
 class PortalProperties(object):
     """ Properties object class, just has fields for db conn """
 
-    def __init__(self, database_host, database_name, database_user, database_pw):
+    def __init__(self, database_host, database_name, database_user, database_pw, genome_build):
         # default port:
         self.database_port = 3306
         # if there is a port added to the host name, split and use this one:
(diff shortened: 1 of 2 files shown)


khzhu:
@sheridancbio , please add any reviewers from your group as needed. Thank you!

sheridancbio:
Thanks Kelsey, this looks good. I'll do a finer pass shortly.

core/src/main/scripts/migrate_db.py
```diff
@@ -150,6 +154,32 @@ def is_version_larger(version1, version2):
...
+def check_reference_genome(portal_properties, cursor, force_migration):
+    """ Retrieve reference genomes from database """
+    try:
+        sql_statement = """
+                           select count(distinct NCBI_BUILD), CANCER_STUDY_IDENTIFIER
+                           from mutation_event 
+                           join mutation on mutation.MUTATION_EVENT_ID = mutation_event.MUTATION_EVENT_ID 
+                           join genetic_profile on genetic_profile.GENETIC_PROFILE_ID = mutation.GENETIC_PROFILE_ID
+                           join cancer_study on cancer_study.CANCER_STUDY_ID = genetic_profile.CANCER_STUDY_ID
+                           where NCBI_BUILD not in ('NA','')
```
sheridancbio: The sql statement in the migration.sql script does not filter out values 'NA' or '' ... so I think the test for problems here should also not filter out these strings. Or maybe the query inside the migration script should also filter out these values when trying to choose a build.

core/src/main/scripts/migrate_db.py
```diff
@@ -150,6 +154,32 @@ def is_version_larger(version1, version2):
...
+                           group by CANCER_STUDY_IDENTIFIER
+                       """
+        cursor.execute(sql_statement)
+        for row in cursor.fetchall():
+            ref_count, study = row
+            if int(ref_count) > 1 and not force_migration:
```
sheridancbio: This test is good for catching whether there are multiple reference genomes present in the mutation event table. But what if there is only a single value for genome build in the mutation event table ... but that value differs from one of our recognized patterns {'37', 'hg19', 'GRCh37', '38', 'hg38', 'GRCh38'}. Let's say everywhere in mutation_event records for a study, the ncbi_build string is "UNKNOWN". I think this could also  generate a warning here, rather than proceeding to import the study with no warning and associate it with genome build hg19 (the default) during migration. Maybe something like:
allowable_genome_references = set( ['37', 'HG19', 'GRCH37', '38', 'HG38', 'GRCH38'] )
if not (retrieved_ncbi_build.upper() in allowable_genome_references):
give warning and exit

this would require changing the query to something like select group_concat(ncbi_build), count(distinct ncbi_build), .... and the later:
retrieved_ncb_build, ref_count, study = row

Maybe this goes beyond what we originally asked for though ... I guess the whole point of having the migration script default to hg37 is that you don't need to check for every possible "other" valid genome specification. So if it is ok that "UNKNOWN" will be silently migrated to hg19, then maybe this comment can be ignored.

khzhu: @sheridancbio , good point! will add logic in. Thanks!

khzhu: @sheridancbio , I've made changes as you suggested and checked in the code. Thanks!

core/src/main/scripts/migrate_db.py
ao508: agreed ^

khzhu: all right, will not filter out 'NA' or ''.

khzhu: done.

khzhu:
@sheridancbio @ao508 ,I've done my corrections. Please let me know if what else needs to be done. Thank you!

sheridancbio:
I was going to make some small fixes during merge, but as the fixes became larger I think having a separate PR would be better. Some issues
- the biggest issue is that the check_reference_genome() function should not be called if the database has already been migrated beyond the multi-reference-genome schema step (2.11.0). This is not currently a problem, because there is nothing yet beyond .. but it will become a problem later. I have a fix for this ready already (will make a follow up PR)
- it would be nice to report which properties were unset in the properties file (rather than "one or more"
- it would be nice to report all the warnings from check_reference_genome() before exiting .. so that all problematic studies could be cleaned up in one pass (rather than fixing one, and then getting the next warning when trying to migrate again)
- it seems like a waste to ask the user if they are sure they want to proceed ... and then to error out if the properties file is not adequate. Error out immediately if the properties are not set.

I will make a follow up PR with those extra fixes. For now, this code is valid until we get new db_schema_version migration steps, so I am squashing/merging.

sheridancbio:
One other issue I noticed while squashing is that although the migrate_db.py script will report that it will use as a default whatever value is in the properties file property 'ucsc.build' ... in truth, the migration.sql script is going to apply the simple rule of IF(`mutation_event`.NCBI_BUILD in ('38', 'hg38','GRCh38') then use hg38 ... otherwise it will be using hg19. So we actually do not need to read ucsc.build when running the migrate_db.py script ... the default is hardcoded to be hg19 in migration.sql, so we should just report that when running the check_reference_genome() function.

PR #6582, opened 2019-09-10
point docs to use latest docker image

### LINKED ISSUE #6558: Cannot find migration.sql while deploying with docker
I am following [this](https://github.com/cBioPortal/cbioportal/tree/master/docs/docker
) link to deploy cbioportal with docker and I am running below command

```
docker run --rm -it --net cbio-net     -v /home/ubuntu/portal.properties:/cbioportal/portal.properties:ro     cbioportal/cbioportal:3.0.1     migrate_db.py -p /cbioportal/portal.properties -s /cbioportal/db-scripts/src/main/resources/migration.sql

```
And I see the following 

```
Unable to find image 'cbioportal/cbioportal:3.0.1' locally
3.0.1: Pulling from cbioportal/cbioportal
741437d97401: Pull complete 
34d8874714d7: Pull complete 
0a108aa26679: Pull complete 
eea0f08c7492: Pull complete 
ca74b3d90c15: Pull complete 
f11a643c4a85: Pull complete 
05eafed73b65: Pull complete 
883f4bb87298: Pull complete 
90499d8233de: Pull complete 
1329833db340: Pull complete 
41b0803e6f1b: Pull complete 
f3ecc6599cdc: Pull complete 
abf7393e8cd6: Pull complete 
52bd7b911bc4: Pull complete 
62a20707052a: Pull complete 
Digest: sha256:596365788aabe6edf129239bca1028a6460096b82dd19f5f7c68bbfeb5f2f2a9
Status: Downloaded newer image for cbioportal/cbioportal:3.0.1
sql file /cbioportal/db-scripts/src/main/resources/migration.sql cannot be found
migrate_db.py --properties-file [portal properties file] --sql [sql migration file]

```

It says `sql file /cbioportal/db-scripts/src/main/resources/migration.sql cannot be found`

Can someone please help?

### LINKED ISSUE #6574: Step 4 of Docker install assumes db-scripts is in container when it is not
## Steps to Reproduce
I was using the [documentation](https://docs.cbioportal.org/2.1.1-deploy-with-docker-recommended/docker) to install the cBioportal. I got to step 4 successfully and ran this: 
```bash
docker run --rm -it --net cbio-net \
    -v /home/luke/code/cbioportal-properties/portal.properties:/cbioportal/portal.properties:ro \
    cbioportal/cbioportal:3.0.1 \
    migrate_db.py -p /cbioportal/portal.properties -s /cbioportal/db-scripts/src/main/resources/migration.sql
```
I expected this to migrate the database. Instead, it reported this error: 

> sql file /cbioportal/db-scripts/src/main/resources/migration.sql cannot be found

## Confirmation
I stepped into the container using this command:
```bash
docker run --rm -it --net cbio-net \
    -v /home/luke/code/cbioportal-properties/portal.properties:/cbioportal/portal.properties:ro \
    cbioportal/cbioportal:3.0.1 \
    /bin/bash
```
and confirmed that the file was in fact not in the container. Furthermore, the `/cbioportal/db-scripts/` directory does not exist.

## Cause
This happens because the 3.0.1 container does not copy db-scripts. This issue was fixed in 3.0.6. Amending the command to reference a newer allows the script to run.

inodb:
Fix #6574 
Fix #6558 

I've also set up a new rule to make "latest" point to the latest tagged commit

### DIFF
--- a/docs/docker/README.md
+++ b/docs/docker/README.md
@@ -93,15 +93,29 @@ db.use_ssl=true
 
 ### Step 4 - Migrate database to latest version ###
 
-Update the seeded database schema to match the cBioPortal version
-in the image, by running the following command. Note that this will
-most likely make your database irreversibly incompatible with older
-versions of the portal code.
+First check what the latest release is of cBioPortal
+[here](https://github.com/cBioPortal/cbioportal/releases/latest). You can
+programmatically get the latest version like this:
+
+```
+LATEST_VERSION=$(curl --silent "https://api.github.com/repos/cBioPortal/cbioportal/releases/latest" \
+    | grep "tag_name" \
+    | cut -d'"' -f4 \
+    | cut -dv -f2)
+echo $LATEST_VERSION
+```
+
+Make note of what version you aim to deploy. Once you finish step 6 you can
+also see in the footer of the homepage what version was deployed.
+
+Update the seeded database schema to match the cBioPortal version in the image,
+by running the following command. Note that this will most likely make your
+database irreversibly incompatible with older versions of the portal code.
 
 ```
 docker run --rm -it --net cbio-net \
     -v /<path_to_config_file>/portal.properties:/cbioportal/portal.properties:ro \
-    cbioportal/cbioportal:3.0.6 \
+    cbioportal/cbioportal:${LATEST_VERSION} \
     migrate_db.py -p /cbioportal/portal.properties -s /cbioportal/db-scripts/src/main/resources/migration.sql
 ```
 
@@ -144,7 +158,7 @@ docker run -d --restart=always \
         -Dsession.service.url=http://cbio-session-service:5000/api/sessions/my_portal/
     ' \
     -p 8081:8080 \
-    cbioportal/cbioportal:3.0.6 \
+    cbioportal/cbioportal:${LATEST_VERSION} \
     /bin/sh -c 'java ${JAVA_OPTS} -jar webapp-runner.jar /cbioportal-webapp'
 ```
 
--- a/docs/docker/README.md
+++ b/docs/docker/README.md
@@ -161,8 +161,25 @@ virtual machine in which all Docker processes are running.
 cBioPortal can now be reached at <http://localhost:8081/>
 
 Activity of Docker containers can be seen with:
+
 ```
 docker ps -a
+```
+
+## A note on versioning ##
+
+For production you might want to deploy a specific docker image tag, instead of
+the `latest` image. The version can be seen in the footer of the home page. The
+various image versions can be found here:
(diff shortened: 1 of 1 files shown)


pieterlukasse:
@inodb should docs point to `latest` instead? Otherwise, who will keep updating the docs?

inodb:
@pieterlukasse yeah i was thinking that, but the problem is that people might always deploy latest and suddenly find themselves on a new version, without specifically changing anything. You can get the latest version like this:

```
curl --silent "https://api.github.com/repos/cBioPortal/cbioportal/releases/latest" | grep "tag_name" | cut -d'"' -f4 | cut -dv -f2
```

Maybe the docs should mention something about that. Like you can get the latest tag like that. But make sure to write down what version you've deployed as to not accidentally upgrade.

inodb:
@pieterlukasse @pvannierop I added some more docs on how to get the latest version, so we don't have to manually update the docs. Let me know what you think

inodb:
@pieterlukasse @pvannierop srry updated again - i think it makes things too complicated, extracting the version etcetera. I assume most people reading the docs won't have an existing cBioPortal running, so maybe better to focus on that use case

pieterlukasse:
@inodb I am fine with the documentation just telling people how to install the latest version. The process of deploying to production requires many more steps, like freezing on a specific release, testing, etc. I don't think that is the scope of this specific part of the documentation, right? A small note mentioning that in real deployments one should use a release version instead of latest can be added though.

pieterlukasse:
Maybe add a small note as simple as: "For real deployments, remember to replace `latest` with a specific release number."

inodb:
@pieterlukasse yeah good idea. I added a small section about versioning. Let me know if that works

docs/docker/README.md
```diff
@@ -162,10 +161,27 @@ virtual machine in which all Docker processes are running.
...
+various image versions can be found here:
+https://hub.docker.com/r/cbioportal/cbioportal/tags. You can also get the
+latest version programmatically like this:
+
+```
+LATEST_VERSION=$(curl --silent "https://api.github.com/repos/cBioPortal/cbioportal/releases/latest" \
+    | grep "tag_name" \
+    | cut -d'"' -f4 \
+    | cut -dv -f2)
+echo $LATEST_VERSION
```
pieterlukasse: I would just remove this part...but that's just because I see no reason why one would like to do this programmatically (except for beta portals that are automatically updated...but these could also just run on :latest)

inodb: not really - just figured that if you deployed `latest`, it's not super straightforward to figure out what tag that corresponds to. Cause if you go to the docker hub tag page you see a zillion images. If you go to the github page with the latest release, you have to remove the `v` prefix to make it work. This way you can copy+paste a command, so seemed easier

PR #6603, opened 2019-09-14
migrate_db.py now avoids no space left on disk error

sheridancbio:
# What? Why?
When migrating very large databases with migrate_db.py, the check for ambiguous or invalid entries for a study in the NCBI_BUILD field of mutation_event was causing errors such as
ERROR 3 (HY000): Error writing file '/tmp/MYcbW7w7' (Errcode: 28 - No space left on device)
(This occurred on a mysql server with 8GB available in the /tmp filesystem)

By avoiding use of keywords GROUP_CONCAT and DISTINCT in the query SELECT clause, and instead using "GROUP BY CANCER_STUDY_IDENTIFIER, NCBI_BUILD", less disk space is used for post-processing query output and the error above was avoided.

- drop GROUP_CONCAT and DISTINCT terms in select clause
    - instead, the GROUP BY clause now includes both CANCER_STUDY_IDENTIFIER and NCBI_BUILD
    - returned records show the count of mutation_event table records which have each NCBI_BUILD value
    - this avoids the generation of large /tmp/xxxxx files on database server to post-process output (the presumed cause of no space left on device error)
- records from query now aggregated into a python map {study_id : {ncbi_build : count}}
- warnings now report the number of mutation_event records linked to each ncbi_build value
- clarified warning message
    - when multiple reference genomes are present, the first encountered one will be used to set the cancer_study genome
    - the default genome (hg19) will be used in place of an unrecognized genome (those not in ALLOWABLE_GENOME_REFERENCES)

# Checks
- [ ] Runs on Heroku.

### DIFF
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -137,11 +137,12 @@ def is_version_larger(version1, version2):
     return False
 
 def print_all_check_reference_genome_warnings(warnings, force_migration):
+    """ Format warnings for output according to mode, and print to ERROR_FILE """
     space =  ' '
     indent = 28 * space
     allowable_reference_genome_string = ','.join(ALLOWABLE_GENOME_REFERENCES)
     clean_up_string = ' Please clean up the mutation_event table and ensure it only contains references to one of the valid reference genomes (%s).' % (allowable_reference_genome_string)
-    use_default_string = 'the default reference genome (%s) will be used.' % (DEFAULT_GENOME_REFERENCE)
+    use_default_string = 'the default reference genome (%s) will be used in place of invalid reference genomes and the first encountered reference genome will be used.' % (DEFAULT_GENOME_REFERENCE)
     use_force_string = 'OR use the "--force" option to override this warning, then %s' % (use_default_string)
     forcing_string = '--force option in effect : %s' % (use_default_string)
     for warning in warnings:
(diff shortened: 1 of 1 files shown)


core/src/main/scripts/migrate_db.py
```diff
@@ -156,24 +156,32 @@ def check_reference_genome(portal_properties, cursor, force_migration):
...
-                    warnings.append(msg)
-            elif int(ref_count) > 1:
-                msg = 'WARNING: Study %s contains mutation_event records with %s NCBI_BUILD values (%s).'%(study, ref_count, retrieved_ncbi_build)
+            if study in study_to_ncbi_to_count:
+                study_to_ncbi_to_count[study][retrieved_ncbi_build] = ref_count
+            else:
+                study_to_ncbi_to_count[study] = {retrieved_ncbi_build : ref_count}
+        for study in study_to_ncbi_to_count:
+            ncbi_to_count = study_to_ncbi_to_count[study]
+            if len(ncbi_to_count) == 1:
```
khzhu: @sheridancbio , I would suggest to pull lines 176 ~ 183 out and into a sub routine let's call it validate_reference_genome_count
```
def validate_reference_genome_count(warnings, ncbi_to_count):
    if len(ncbi_to_count) == 1:
        for retrieved_ncbi_build in ncbi_to_count: # single iteration
            if retrieved_ncbi_build.upper() not in [x.upper() for x in ALLOWABLE_GENOME_REFERENCES]:
                msg = 'WARNING: Study %s contains mutation_event records with unsupported NCBI_BUILD value %s.'%(study, retrieved_ncbi_build)
                warnings.append(msg)
    elif len(ncbi_to_count) > 1:
        msg = 'WARNING: Study %s contains mutation_event records with %s NCBI_BUILD values {ncbi_build:record_count,...} %s.'%(study, len(ncbi_to_count), ncbi_to_count)
        warnings.append(msg)
     return warnings
```

khzhu: then you can iterate through study_to_ncbi_to_count and then call the new sub routine. 
```
        for study in study_to_ncbi_to_count:
            validate_reference_genome_count(warnings, study_to_ncbi_to_count[study])
```

khzhu:
@sheridancbio , done my review. good work!

core/src/main/scripts/migrate_db.py
sheridancbio: Thank you .. I have added the subroutine. I named it validate_reference_genome_values_for_study, because it validates both the count and the acceptability of the values. Also, the function does not need to return a value, because 'warnings' is a pass by reference argument so when the subroutine appends to that object, the modifications are visible in the caller.

sheridancbio:
I have also added function-level comments for the new functions, to match the style in the rest of the script.

PR #6619, opened 2019-09-18
make clinical attributes search/filter case insensitive

### LINKED ISSUE #5853: Summary view filtering bug
Grouping of samples in the pie chart is not case sensitive, but if we apply filter then the code does exact string match.
Example: Local Recurrence = 11 samples, If I apply the filter only 10 samples get selected because one of the sample had a typo: 'Local recurrence' (R is in lower case).

khzhu:
# What? Why?
Addresses backend of #5853 

- a
- b

# Checks
- [ ] Runs on Heroku.

# Any screenshots or GIFs?


khzhu:
Hi @kalletlak , could you please review the change. Thank you!

web/src/main/java/org/cbioportal/web/util/ClinicalDataFilterApplier.java
```diff
@@ -46,13 +46,17 @@ public ClinicalDataFilterApplier(PatientService patientService,
...
+                c.setAttributeId(c.getAttributeId().toUpperCase());
+            });
             clinicalDataList = clinicalDataService.fetchClinicalData(filterClinicalDataType.equals(ClinicalDataType.PATIENT) ?
                     studyIdsOfPatients : studyIds, filterClinicalDataType.equals(ClinicalDataType.PATIENT) ? patientIds : sampleIds,
                 attributes.stream().map(ClinicalDataFilter::getAttributeId).collect(Collectors.toList()),
                 filterClinicalDataType.name(), Projection.SUMMARY.name());
 
             clinicalDataList.forEach(c -> {
-                if (c.getAttrValue().toUpperCase().equals("NAN") || c.getAttrValue().toUpperCase().equals("N/A")) {
+                c.setAttrValue(c.getAttrValue().toUpperCase());
```
kalletlak: why are all attribute values converted to upper case?

kalletlak: value should be converted to upper case just for conditional checking.

web/src/main/java/org/cbioportal/web/util/ClinicalDataFilterApplier.java
```diff
@@ -46,13 +46,17 @@ public ClinicalDataFilterApplier(PatientService patientService,
             List<Patient> patients = patientService.getPatientsOfSamples(studyIds, sampleIds);
             List<String> patientIds = patients.stream().map(Patient::getStableId).collect(Collectors.toList());
             List<String> studyIdsOfPatients = patients.stream().map(Patient::getCancerStudyIdentifier).collect(Collectors.toList());
+            attributes.forEach(c-> {
+                c.setAttributeId(c.getAttributeId().toUpperCase());
```
kalletlak: mysql queries are case insensitive. this won't be required

web/src/main/java/org/cbioportal/web/util/ClinicalDataFilterApplier.java
khzhu: mysql is case in sensitive but not persistent layer. if you try 
```
https://www.cbioportal.org/study/summary?filterAttributeId=SEX&filterValues=Male&id=msk_impact_2017
```
and 
```
 https://www.cbioportal.org/study/summary?filterAttributeId=SEX&filterValues=MALE&id=msk_impact_2017
```
you will get a complete different result.

kalletlak: you might need to update values(to uppercase) in attributes for it to work

khzhu: yes. I am correcting those lines.

web/src/main/java/org/cbioportal/web/util/ClinicalDataFilterApplier.java
khzhu: @kalletlak , I do not think  ClinicalDataEqualityFilterApplier  is a right place to go. It overrides apply method of ClinicalDataFilterApplier and returns a total count instead of SampleIdentifier object.

kalletlak: true, but the check whether the sample attribute value is in the filtered values values is done there

khzhu: right, need some changes in that class too. testing now. thanks!

khzhu:
@kalletlak , corrected. could you please take a final look at the PR. Thanks!

khzhu:
can I have your approval, @zhx828 , thanks!

inodb:
So @kalletlak tested this further (thank you 🙏!) and backend is good. The correct response is returned. Just the originally merged frontend part is actually not working for that particular example ☝️ (this PR fixed case insensitive attribute ids but not attribute values: https://github.com/cBioPortal/cbioportal-frontend/pull/2716)

PR #6653, opened 2019-09-30
Quick Search For Samples

Luke-Sikina:
- Add Sample endpoint for keyword search
    - Added logic to Sample controller, service, and repository, as well as the XML
    - If the keyword parameter is present, samples are matched to it based on STABLE_ID
- Expand Patient keyword search
    - XML: added an internal join to the corresponding samples so that patients with
    matching samples show up in quicksearch

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?
foo


persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/PatientMyBatisRepositoryTest.java
```diff
@@ -34,6 +36,16 @@ public void getAllPatients() throws Exception {
         Assert.assertEquals("TCGA-A1-A0SB", patient.getStableId());
         Assert.assertNull(patient.getCancerStudy());
     }
+    
+    @Test
+    public void getAllPatientsByKeywordMatchingSample() {
```
inodb: nice

PR #6660, opened 2019-10-01
Retrieve cBP version through API (fix https://github.com/cBioPortal/icebox/issues/55)

SRodenburg:
Fix https://github.com/cBioPortal/icebox/issues/55

Validator now retrieves the cBioPortal version through the `/api/info` URI and displays it on the resulting HTML page.

When setting up a cBioPortal instance from a portal dump directory (`metaImport.py -p <dir>`), the validator will set the version to `unknown`, since there is no dump function implemented for the `info` API yet.

**EDIT:**
Turned out the version was already detemined in the logger prior to HTML generation. Now `cbio_version` was removed from the constructor of the Jinja2 HTML logger.

**EDIT:**
In case no version could be retrieved, the reason can be 1. A json format error 2. offline portal instance 3. Portal checks disabled.
The validator now considers these scenarios and displays them when no version could be retrieved.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -169,11 +169,10 @@ class Jinja2HtmlHandler(logging.handlers.BufferingHandler):
 
     """Logging handler that formats aggregated HTML reports using Jinja2."""
 
-    def __init__(self, study_dir, output_filename, cbio_version, max_reported_values, *args, **kwargs):
+    def __init__(self, study_dir, output_filename, max_reported_values, *args, **kwargs):
         """Set study directory name, output filename and buffer size."""
         self.study_dir = study_dir
         self.output_filename = output_filename
-        self.cbio_version = cbio_version
         self.max_reported_values = max_reported_values
         self.max_level = logging.NOTSET
         self.closed = False
@@ -196,8 +195,11 @@ class Jinja2HtmlHandler(logging.handlers.BufferingHandler):
         """Never flush; emit() caps the buffer and close() renders output."""
         return False
 
-    def generateHtml(self):
-        """Render the HTML page for the current content in self.buffer """
+    def generateHtml(self, **kwargs):
+        """Render the HTML page for the current content in self.buffer
+        
+        **kwargs allows to override logger variables to display.
+        """
         # require Jinja2 only if it is actually used
         import jinja2
         j_env = jinja2.Environment(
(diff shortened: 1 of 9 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -4833,13 +4845,13 @@ def validate_study(study_dir, portal_instance, logger, relaxed_mode, strict_maf_
...
-    """
-    pom_path = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))))) + "/pom.xml"
-    return pom_path
+# deprecated function
+# def get_pom_path():
+#     """
+#     Get location of pom.xml. In system and integration test this is mocked.
+#     """
+#     pom_path = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))))) + "/pom.xml"
+#     return pom_path
```
pvannierop: Better to just remove?

SRodenburg: removed and adjusted code

core/src/main/scripts/importer/validateData.py
```diff
@@ -4885,24 +4890,6 @@ def main_validate(args):
     # set default to unknown because validator can be run independently from cBioPortal
     cbio_version = "unknown"
```
pvannierop: It seems the default value `cbio_version` is set at two places (see previous comment). Better to choose one place where default is set.

core/src/main/scripts/importer/validateData.py
```diff
@@ -294,6 +297,12 @@ def __init__(self, cancer_type_dict, hugo_entrez_map, alias_entrez_map, gene_set
         self.__ncbi_build = '37'
         self.__genome_build = 'hg19'
 
+        # set version as attribute
+        if portal_info_dict is not None:
+            self.portal_version = portal_info_dict.get('portalVersion',  'unknown')
+        else:
+            self.portal_version = 'unknown'
```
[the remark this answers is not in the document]
SRodenburg: fixed

core/src/main/scripts/importer/validateData.py
SRodenburg: fixed

SRodenburg:
@inodb 
I think this can finally be merged

PR #6664, opened 2019-10-01
Add preauth filtering of structural variants

sheridancbio:
Fix missing security filtering for StructuralVariant endpoint on authenticated portals

- first commit contains whitespace cleanup and style standardization
- second commit contains the following:
- annotated handler with @PreAuthorize
- added Attributes to handler (extracted by interceptor), and use intercepted body instead of actual body parameter
- make request body parameter optional (because interceptor has consumed it)
- make StructuralVariantFilter model serializable
- added handling of endpoint /structuralvariant/fetch to InvolvedCancerStudyExtractorInterceptor
    - extract involved cancer studies based on SampleMolecularIdentifiers (if present) ... otherwise on MolecularProfileIds (from filter argument)

# Checks
- [ ] Runs on heroku

### DIFF
--- a/web/src/main/java/org/cbioportal/web/StructuralVariantController.java
+++ b/web/src/main/java/org/cbioportal/web/StructuralVariantController.java
@@ -53,14 +53,14 @@ import java.util.ArrayList;
 public class StructuralVariantController {
     @Autowired
     private StructuralVariantService structuralVariantService;
-    
-    @RequestMapping(value = "/structuralvariant/fetch", method = RequestMethod.POST, 
+
+    @RequestMapping(value = "/structuralvariant/fetch", method = RequestMethod.POST,
             consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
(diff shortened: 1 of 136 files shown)


web/src/main/java/org/cbioportal/web/StructuralVariantController.java
```diff
@@ -36,14 +36,17 @@
...
+import org.springframework.web.bind.annotation.RequestAttribute;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.bind.annotation.RequestBody;
+import springfox.documentation.annotations.ApiIgnore;
 
 import javax.validation.Valid;
-import java.util.List;
+import java.util.*;
```
n1zea144: Minor, but you can remove following import java.util.ArrayList;

sheridancbio: done

PR #6666, opened 2019-10-01
Configure Ehcache using runtime properties

### LINKED ISSUE #6631: ehcache parameters don't work as runtime config (need to recompile)
Some ehcache properties currently only work when included during compile time. This should all be runtime config that can be set using both

`-Dproperty_name=something`

and

`$PORTAL_HOME/portal.properties`

mandawilson:
Fixes #6631 

We have switched from an Ehcache XML configuration to a primarily Java configuration so that we can configure Ehcache using runtime properties rather than relying on Maven filtering.  For example, `-Dehcache.cache_type=hybrid`.

- ehcache.xml switched to a more general template instead of initiailizing each individual cache
 - properties now auto-wired in through CustomEhCacheProvider
 - allow for disk-only/heap-only/hybrid config to all be controlled from portal.properties `ehcache.cache_type`
 - update properties docs
 - update caching documentation

### DIFF
--- a/docs/Caching.md
+++ b/docs/Caching.md
@@ -2,16 +2,36 @@
 cBioPortal provides the option of caching information on the backend to improve performance. Without caching, every time a request is received by the backend, a query is sent to the database system for information, and the returned data is processed to construct a response. This may lead to performance issues as the entire process can be rather costly, especially for queries on larger studies. With caching turned on, query responses can be taken directly from the cache if they have already been constructed. They would only be constructed for the initial query.
 
 ## Cache Configuration
-The portal is configured to use Ehcache for backend caching; caching configuration is specified inside an xml file under `persistence/persistence-api/src/main/resources`. The default configuration is a hybrid (disk + heap); however, Ehcache supports both a disk-only and heap-only mode. Refer to comments in `ehcache.xml` to switch between cache configurations. The configuration files also specify which caches to create. Additional specifications such as cache size and location are set inside `portal.properties` (more information [here](portal.properties-Reference.md#ehcache-settings)).
+The portal is configured to use Ehcache for backend caching. Ehcache supports a hybrid (disk + heap), disk-only, and heap-only mode; this, along with additional specifications such as cache size and location, are set inside `portal.properties`(more information [here](portal.properties-Reference.md#ehcache-settings)). 
  
 ## Creating additional caches
-The default configuration initializes two separate caches. However, you may wish to introduce new caches with different policies (e.g expiration policy) for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the Ehcache xml configuration file. The `@Cacheable` annotation must also be added (or adjusted) to function declarations to indicate which functions are to be cached. Those might look like this example:
+Cache initialization is handled inside the [CustomEhCachingProvider](../persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java). The default configuration initializes two seperate caches; however, you may wish to introduce new caches for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), new code must be added to the CustomEhCachingProvider.
+
+Within the `CustomEhCachingProvider`, initialize a new `ResourcePoolsBuilder` for the new cache and set the resources accordingly. 
+```
+ResourcePoolsBuilder clinicalDataCacheResourcePoolsBuilder = ResourcePoolsBuilder.newResourcePoolsBuilder();
+clinicalDataCacheResourcePoolsBuilder = clinicalDataCacheResourcePoolsBuilder.heap(clinicalDataCacheHeapSize, MemoryUnit.MB);
+clinicalDataCacheResourcePoolsBuilder = clinicalDataCacheResourcePoolsBuilder.disk(clinicalDataCacheDiskSize, MemoryUnit.MB);
+```
+After initialzing the `ResourcePoolsBuilder`, create a CacheConfiguration for the new cache using the new `ResourcePoolsBuilder` just created.
+```
+CacheConfiguration<Object, Object> clinicalDataCacheConfiguration = xmlConfiguration.newCacheConfigurationBuilderFromTemplate("RepositoryCacheTemplate", 
+Object.class, Object.class, clinicalDataCacheResourcePoolsBuilder)
+.withSizeOfMaxObjectGraph(Long.MAX_VALUE)
+.withSizeOfMaxObjectSize(Long.MAX_VALUE, MemoryUnit.B)
+.build();
+```
+Finally, add the new `CacheConfiguration` to the map of managed caches with a name for the cache. 
+```
+caches.put("ClinicalDataCache", ClinicalDataCacheConfiguration);
+```
+The `@Cacheable` annotation must also be added (or adjusted) to function declarations to indicate which functions are to be cached. Those might look like this example:
 ``` 
 @Cacheable(cacheNames = "ClinicalDataCache", condition = "@cacheEnabledConfig.getEnabled()")
 public String getDataFromClinicalDataRepository(String param) {}
 ```
-Additionally, new properties for setting cache sizes should be added to `portal.properties`. Alternatively, values may also be hardcoded directly into the Ehcache xml configuration file. 
+Additionally, new properties for setting cache sizes should be added to `portal.properties` and loaded into the [CustomEhCachingProvider](../persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java). Alternatively, values may be hardcoded directly inside [CustomEhCachingProvider](../persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java).
 
-For more information on constructing an Ehcache xml configuration file, refer to the documentation [here](https://www.ehcache.org/documentation/3.7/xml.html). 
+For more information on cache templates and the Ehcache xml configuration file, refer to the documentation [here](https://www.ehcache.org/documentation/3.7/xml.html). 
 
 For more information on linking caches to functions, refer to the documentation [here](https://spring.io/guides/gs/caching/).
--- a/docs/portal.properties-Reference.md
+++ b/docs/portal.properties-Reference.md
@@ -284,12 +284,12 @@ This gene set will add the following in the query box:
 # Ehcache Settings
 cBioPortal is supported on the backend with Ehcache. The configuration, size, and location of these caches are configurable from within portal.properties through the following properties.
 
-Caching is disabled by default. When caching is disabled, no respones will be stored in cache. To turn on caching set `ehcache.cache_enabled` to true.  
+The cache type is set using `ehcache.cache_type`. Valid values are `none`, `heap` (heap-only), `disk` (disk-only), and `hybrid` (disk + heap). By default, `ehcache.cache_type` is set to `none` which disables the cache. When the cache is disabled, no responses will be stored in the cache. 
 ```
-ehcache.cache_enabled=true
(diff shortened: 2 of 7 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -234,7 +234,7 @@ oncoprint.hotspots.default=true
 # to enable use of Ehcache, set ehcache.cache_enabled to true and uncomment all Ehcache properties
 ehcache.cache_enabled=false
```
ao508: we dont need this anymore right?

mandawilson: Sorry we hadn't added that file to the commit.  Thanks!

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -231,10 +231,10 @@ oncoprint.hotspots.default=true
...
 
-# to enable use of Ehcache, set ehcache.cache_enabled to true and uncomment all Ehcache properties
-ehcache.cache_enabled=false
+# to enable use of Ehcache, set ehcache.cache_type to a valid cache type (heap, disk, hybrid)
+ehcache.cache_type=none
 #ehcache.xml_configuration=/ehcache.xml
-#ehcache.statistics_enabled=true
+#ehcache.statistics_enabled=false
 # Enable cache statistics endpoint for Ehcache monitoring
 #cache.statistics_endpoint_enabled=false
```
ao508: there also seems to be some redundancy here - can you confirm if we need the property `cache.statistics_endpoint_enabled=false`

averyniceday: yeah, this is for the actual cacheStats endpoint, the ehcache.statistics_enabled has to do with the actual configuration of the cacheManager. 

It's separate because ideally the cacheStats endpoint should be independent of whether or not we use Ehcache as our caching system.

docs/portal.properties-Reference.md
```diff
@@ -308,9 +308,8 @@ ehcache.static_repository_cache_one.max_mega_bytes_heap=
 ehcache.static_repository_cache_one.max_mega_bytes_local_disk=
 ```
 
-Additional properties can be specified for cache statistics monitoring. To log metrics regarding memory usage, set `ehcache.statistics_enabled` to true. Logged metrics and additional information such as cache size and cached keys are available through an optional endpoint. The optional endpoint is turned off by default but can be turned on by setting `cache.statistics_endpoint_enabled` to true. 
+Logged metrics and additional information such as cache size and cached keys are available through an optional endpoint. The optional endpoint is turned off by default but can be turned on by setting `cache.statistics_endpoint_enabled` to true.
```
ao508: This info should be removed since we are no longer supporting this property.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
```diff
@@ -40,43 +43,104 @@
...
 import org.ehcache.core.config.DefaultConfiguration;
 import org.ehcache.jsr107.EhcacheCachingProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.ehcache.impl.config.persistence.DefaultPersistenceConfiguration;
+import org.cbioportal.persistence.CacheEnabledConfig;
 
 public class CustomEhCachingProvider extends EhcacheCachingProvider {
```
ao508: did you manage to inject properties the spring way without annotating this class?

docs/portal.properties-Reference.md
```diff
@@ -284,12 +284,12 @@ This gene set will add the following in the query box:
...
 ```
-ehcache.cache_enabled=true
+ehcache.cache_type=[none or heap or disk or hybrid]
 ```
 
-When caching is enabled, set a cache configuration using `ehcache.xml_configuration`. This should be the name of an Ehcache xml configuration file; the default provided is `ehcache.xml` which configures a hybrid (disk + heap) cache. To change caching configuration, directly edit `/ehcache.xml`. Alternatively, you can create your own Ehcache xml configuration file, place it under `/persistence/persistence-api/src/main/resources/` and set `ehcache.xml_configuration` to `/[Ehcache xml configuration filename]`.  
+Ehcache initializes caches based off of a template found in an Ehcache xml configuration file. When caching is enabled, set `ehcache.xml_configuration` to the name of the Ehcache xml configuration file. The default provided is `ehcache.xml`; to change the cache template, directly edit this file. Alternatively, you can create your own Ehcache xml configuration file, place it under `/persistence/persistence-api/src/main/resources/` and set `ehcache.xml_configuration` to `/[Ehcache xml configuration filename]`.
```
sheridancbio: "based off of a" -> "using" ?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheEnabledConfig.java
```diff
@@ -1,13 +1,30 @@
...
-    public CacheEnabledConfig(String cacheEnabled) {
-        this.enabled = Boolean.parseBoolean(cacheEnabled);
+    public static ArrayList<String> validCacheTypes = new ArrayList<String>(Arrays.asList(DISK, HEAP, HYBRID));
+
+    public CacheEnabledConfig(String cacheType) {
+        this.enabled = enableCache(cacheType);
+    }
+
+    public static boolean enableCache(String cacheType) {
+        for (String validCacheType : validCacheTypes) {
```
sheridancbio: I think we could do this with `validCacheTypes.contains(cacheType.toLowerCase())` ... searching an array of length 3 is fine with a linear scan (rather than hashing in a set)

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
```diff
@@ -40,43 +43,104 @@
...
         CacheManager toReturn = null;
-         try {
-            if (cacheEnabled != null && cacheEnabled) {
-                LOG.info("Caching is enabled, using '" + xmlConfiguration + "' for configuration");
-                toReturn = this.getCacheManager(getClass().getResource(xmlConfiguration).toURI(),
-                                            getClass().getClassLoader());
+        try {
+            if (CacheEnabledConfig.enableCache(cacheType)) {
+                LOG.info("Caching is enabled, using '" + xmlConfigurationFile + "' for configuration");
+                XmlConfiguration xmlConfiguration = new XmlConfiguration(getClass().getResource(xmlConfigurationFile));
```
sheridancbio: it might be a good idea to catch the case where getResource(xmlConfigurationFile) returns null .. that way we can do something like LOG.error("attempt to load Ehcache xml configuration file " + xmlConfiguration + " failed -- file is not available in the classpath") rather than an NPE

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
```diff
@@ -40,43 +43,104 @@
...
     public CacheManager getCacheManager() {
 
         CacheManager toReturn = null;
-         try {
-            if (cacheEnabled != null && cacheEnabled) {
-                LOG.info("Caching is enabled, using '" + xmlConfiguration + "' for configuration");
-                toReturn = this.getCacheManager(getClass().getResource(xmlConfiguration).toURI(),
-                                            getClass().getClassLoader());
+        try {
+            if (CacheEnabledConfig.enableCache(cacheType)) {
```
sheridancbio: more than once in this try block we are checking the configuration disk / heap / hybrid with string comparisons. We could simplify that by up front creating two boolean variables "cachesUseHeap" and "cachesUseDisk" ... then below there is no need for string comparison and code is cleaner.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -231,10 +231,9 @@ oncoprint.hotspots.default=true
 # Custom gene sets
 # querypage.setsofgenes.location=file:/<path>
 
-# to enable use of Ehcache, set ehcache.cache_enabled to true and uncomment all Ehcache properties
-ehcache.cache_enabled=false
+# to enable use of Ehcache, set ehcache.cache_type to a valid cache type (heap, disk, hybrid)
```
sheridancbio: In the comment, I would add a suffix ... ", to disable use of Ehcache set cache_type to none
That way a user who has enabled the cache will know how to disable it again.
Maybe a simpler comment would be # valid cache types are (heap, disk, hybrid), or use 'none' to disable use of Ehcache

docs/portal.properties-Reference.md
mandawilson: @ao508 We are supporting `cache.statistics_endpoint_enabled`.  We are not supporting `ehcache.statistics_enabled`.  Confusing I know.  The endpoint one is to say whether or not we want the endpoints related to viewing various information about the caches.  The `ehcache.statistics_enabled` I thought was meant to turn on/off `EhCacheStatistics.java` and/or the logging that starts with "CACHE_STATISTICS START", but it doesn't seem to do either of those things.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
mandawilson: Yes they are injected without annotating, I assume because the bean is defined in the XML file.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheEnabledConfig.java
mandawilson: `validCacheTypes.contains(cacheType.toLowerCase())` would throw a null pointer exception because cacheType can be null.  That is what was breaking the Travis tests.  We could of course get around that... but we don't really need to.

docs/Caching.md
```diff
@@ -2,16 +2,36 @@
...
  
 ## Creating additional caches
-The default configuration initializes two separate caches. However, you may wish to introduce new caches with different policies (e.g expiration policy) for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the Ehcache xml configuration file. The `@Cacheable` annotation must also be added (or adjusted) to function declarations to indicate which functions are to be cached. Those might look like this example:
+Cache initialization is handled inside the [CustomEhCachingProvider](../persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java). The default configuration initializes two seperate caches; however, you may wish to introduce new caches for different datatypes. To create additional caches (e.g creating a cache specifically for clinical data), a new cache must be added to the CustomEhCachingProvider.
```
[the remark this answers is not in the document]
mandawilson: Done.

docs/portal.properties-Reference.md
mandawilson: Done.

src/main/resources/portal.properties.EXAMPLE
mandawilson: Done.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/CacheEnabledConfig.java
sheridancbio: 👍

sheridancbio:
Hi @inodb ... unit tests here have now passed. (Or maybe you were referring to end-to-end tests?)

sheridancbio:
@n1zea144 : we may be merging this PR soon. Do you want us to wait to give you a chance to review?

PR #6669, opened 2019-10-02
fix report result html with adding new profiles

yichaoS:


inodb:
included these changes as part of https://github.com/cBioPortal/cbioportal/pull/6661

PR #6694, opened 2019-10-10
Improve maven dependency caching Dockerfile

pvannierop:
Building the docker image from `/docker/web-and-data/Dockerfile` takes longer than needed after code updates because it does not take advantage of docker caching.

### Fix
- We have updated the Dockerfile to cache downloading of mvn dependencies. This means that instead of a build time of 6'30'' after a code update the build time is reduced to 1'30''.
- Also we updated the message on exposure of the .git folder in the image (which does not occur, due to a multilayered build).

### DIFF
--- a/docker/web-and-data/Dockerfile
+++ b/docker/web-and-data/Dockerfile
@@ -8,9 +8,9 @@
 #
 # docker build -f docker/web-and-data/Dockerfile -t cbioportal-container:tag-name .
 #
-# WARNING: Be careful about publishing images generated like this publicly
-# because your .git folder is exposed in the build step. We are not sure if
-# this is a security risk: stackoverflow.com/questions/56278325
+# NOTE: the .git folder is included in the build stage, but excluded 
+# from the final image. No confidential information is exposed.
+# (see: stackoverflow.com/questions/56278325)
 FROM maven:3.5.4 as build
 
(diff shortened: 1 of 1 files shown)


docker/web-and-data/Dockerfile
```diff
@@ -8,16 +8,49 @@
...
+COPY db-scripts/pom.xml                       db-scripts/
+COPY model/pom.xml                            model/
+COPY persistence/pom.xml                      persistence/
+COPY persistence/persistence-api/pom.xml      persistence/persistence-api/
+COPY persistence/persistence-mybatis/pom.xml  persistence/persistence-mybatis/
+COPY scripts/pom.xml                          scripts/
+COPY security/pom.xml                         security/
+COPY security/security-spring/pom.xml         security/security-spring/
+COPY service/pom.xml                          service/
+COPY web/pom.xml                              web/
```
inodb: It would be nice if the pom files weren't hardcoded but there might not be an easy way to do that

pvannierop: The thing is that your would want them on separate lines of the Dockerfile. That would then only be possible when you create a Dockerfile via script.

inodb: I see - thanks

PR #6695, opened 2019-10-10
Skip SV records with invalid exon/ensembl transcript ids

ao508:
Addresses Problem:
SV visualization in beta portal breaks if SV record is missing either site1 or site2 ensembl transcript id.

**Skip SV records with invalid exon/ensembl transcript ids**

Skip records in Structural Variant data file that are missing any of the following required fields:
- site 1 and 2 exon (default is -1, these will be skipped)
- site 1 and 2 ensembl transcript ids (default to NA if missing, these will be skipped)
- site 1 and 2 entrez id/hugo symbol (records missing both of these will be skipped)

# Checks
- [ ] Runs on heroku

# Notify reviewers

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportStructuralVariantData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportStructuralVariantData.java
@@ -27,7 +27,6 @@ import java.util.*;
 import org.mskcc.cbio.maf.*;
 import org.mskcc.cbio.portal.dao.*;
 import org.mskcc.cbio.portal.model.*;
-import org.mskcc.cbio.portal.model.ExtendedMutation.MutationEvent;
 import org.mskcc.cbio.portal.util.*;
 
 /**
@@ -36,65 +35,17 @@ import org.mskcc.cbio.portal.util.*;
  */
 
 public class ImportStructuralVariantData {
-    // Column names structural variant file
-    public static final String SAMPLE_ID = "Sample_ID";
-    public static final String SITE1_ENTREZ_GENE_ID = "Site1_Entrez_Gene_Id";
-    public static final String SITE1_HUGO_SYMBOL = "Site1_Hugo_Symbol";
-    public static final String SITE1_ENSEMBL_TRANSCRIPT_ID = "Site1_Ensembl_Transcript_Id";
-    public static final String SITE1_EXON = "Site1_Exon";
-    public static final String SITE1_CHROMOSOME = "Site1_Chromosome";
-    public static final String SITE1_POSITION = "Site1_Position";
-    public static final String SITE1_DESCRIPTION = "Site1_Description";
-    public static final String SITE2_ENTREZ_GENE_ID = "Site2_Entrez_Gene_Id";
-    public static final String SITE2_HUGO_SYMBOL = "Site2_Hugo_Symbol";
-    public static final String SITE2_ENSEMBL_TRANSCRIPT_ID = "Site2_Ensembl_Transcript_Id";
-    public static final String SITE2_EXON = "Site2_Exon";
-    public static final String SITE2_CHROMOSOME = "Site2_Chromosome";
-    public static final String SITE2_POSITION = "Site2_Position";
-    public static final String SITE2_DESCRIPTION = "Site2_Description";
-    public static final String SITE2_EFFECT_ON_FRAME = "Site2_Effect_On_Frame";
-    public static final String NCBI_BUILD = "NCBI_Build";
-    public static final String DNA_SUPPORT = "DNA_Support";
-    public static final String RNA_SUPPORT = "RNA_Support";
-    public static final String NORMAL_READ_COUNT = "Normal_Read_Count";
-    public static final String TUMOR_READ_COUNT = "Tumor_Read_Count";
-    public static final String NORMAL_VARIANT_COUNT = "Normal_Variant_Count";
-    public static final String TUMOR_VARIANT_COUNT = "Tumor_Variant_Count";
-    public static final String NORMAL_PAIRED_END_READ_COUNT = "Normal_Paired_End_Read_Count";
-    public static final String TUMOR_PAIRED_END_READ_COUNT = "Tumor_Paired_End_Read_Count";
-    public static final String NORMAL_SPLIT_READ_COUNT = "Normal_Split_Read_Count";
-    public static final String TUMOR_SPLIT_READ_COUNT = "Tumor_Split_Read_Count";
-    public static final String ANNOTATION = "Annotation";
-    public static final String BREAKPOINT_TYPE = "Breakpoint_Type";
-    public static final String CENTER = "Center";
-    public static final String CONNECTION_TYPE = "Connection_Type";
-    public static final String EVENT_INFO = "Event_Info";
-    public static final String VARIANT_CLASS = "Class";
-    public static final String LENGTH = "Length";
-    public static final String COMMENTS = "Comments";
-    public static final String EXTERNAL_ANNOTATION = "External_Annotation";
-    public static final String DRIVER_FILTER = "cbp_driver";
-    public static final String DRIVER_FILTER_ANNOTATION = "cbp_driver_annotation";
-    public static final String DRIVER_TIERS_FILTER = "cbp_driver_tiers";
-    public static final String DRIVER_TIERS_FILTER_ANNOTATION = "cbp_driver_tiers_annotation";
-
-    // Other constants
-    public static final String PROTEIN_CHANGE_FUSION = "FUSION";
 
     // Initialize variables
     private File structuralVariantFile;
     private int geneticProfileId;
-    private int mutationGeneticProfileId;
     private String genePanel;
     private Set<String> sampleSet  = new HashSet<>();
-    private HashMap<String, Integer> columnIndexMap;
-
 
     public ImportStructuralVariantData(File structuralVariantFile, int geneticProfileId, String genePanel) throws DaoException {
         this.structuralVariantFile = structuralVariantFile;
         this.geneticProfileId = geneticProfileId;
         this.genePanel = genePanel;
-        this.columnIndexMap = new HashMap<String, Integer>();
     }
 
     public void importData() throws IOException, DaoException {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportStructuralVariantData.java
```diff
@@ -36,134 +35,46 @@
...
-                structuralVariant.setDriverFilter(TabDelimitedFileUtil.getPartString(getColumnIndex(DRIVER_FILTER), parts));
-                structuralVariant.setDriverFilterAnn(TabDelimitedFileUtil.getPartString(getColumnIndex(DRIVER_FILTER_ANNOTATION), parts));
-                structuralVariant.setDriverTiersFilter(TabDelimitedFileUtil.getPartString(getColumnIndex(DRIVER_TIERS_FILTER), parts));
-                structuralVariant.setDriverTiersFilterAnn(TabDelimitedFileUtil.getPartString(getColumnIndex(DRIVER_TIERS_FILTER_ANNOTATION), parts));
+                if (!structuralVariantUtil.hasRequiredStructuralVariantFields(structuralVariant)) {
+                    ProgressMonitor.logWarning("Invalid Site 1 or 2 Ensembl transcript ID or exon found, ignoring structural variant for SV record #" +
+                            recordCount + " (sample, site 1 gene, site 2 gene):  (" +
+                            structuralVariant.getSampleId() + ", " + structuralVariant.getSite1HugoSymbol() +
+                            ", " + structuralVariant.getSite2HugoSymbol() + ")");
+                    continue;
```
khzhu: it might be better rearranging if else to remove "continue" 
```
                if (structuralVariantUtil.hasRequiredStructuralVariantFields(structuralVariant)) {
                     // get sample
                     Sample sample = DaoSample.getSampleByCancerStudyAndSampleId(
                      ....
                }
                else {
                    ProgressMonitor.logWarning("Invalid Site 1 or 2 Ensembl transcript ID or exon found, ignoring structural variant for SV record #" +
                            recordCount + " (sample, site 1 gene, site 2 gene):  (" +
                            structuralVariant.getSampleId() + ", " + structuralVariant.getSite1HugoSymbol() +
                            ", " + structuralVariant.getSite2HugoSymbol() + ")");
                }            
```

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,163 @@
...
+*/
+
+package org.mskcc.cbio.portal.util;
+
+import java.util.HashMap;
+import org.mskcc.cbio.maf.TabDelimitedFileUtil;
+import org.mskcc.cbio.portal.model.StructuralVariant;
+
+/**
+ *
```
khzhu: extra *

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return index;
+    }
+
+    public Boolean hasRequiredStructuralVariantFields(StructuralVariant record) {
+        return !record.getSite1EnsemblTranscriptId().equalsIgnoreCase(TabDelimitedFileUtil.NA_STRING) &&
+                !record.getSite2EnsemblTranscriptId().equalsIgnoreCase(TabDelimitedFileUtil.NA_STRING) &&
+                record.getSite1Exon() != -1 &&
+                record.getSite2Exon() != -1 &&
+                (record.getSite1EntrezGeneId() != Long.MIN_VALUE || !record.getSite1HugoSymbol().equalsIgnoreCase(TabDelimitedFileUtil.NA_STRING)) &&
```
khzhu: do we need both of them be present? we only need either site1_entrez_gene_id1 or site1_hugo_symbol1, right?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportStructuralVariantData.java
ao508:
```
if (...) {
continue
}
```

looks cleaner in my opinion but I think it's just a matter of personal preference. I don't believe there really is any consistency across the codebase.

khzhu: it is okay go with your way. I normally do not use a jump (continue) in the middle of code.

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,162 @@
...
+import org.mskcc.cbio.portal.model.StructuralVariant;
+
+/**
+ * @author ochoaa
+ */
+public class StructuralVariantUtil {
+    private HashMap<String, Integer> columnIndexMap;
+
+    // Column names structural variant file
+    public static final String VARIANT_CLASS = "Class";
```
sheridancbio: Can we alphabetize these? This order looks scrambled.

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,162 @@
...
+    public static final String EXTERNAL_ANNOTATION = "External_Annotation";
+    public static final String TUMOR_SPLIT_READ_COUNT = "Tumor_Split_Read_Count";
+    public static final String SITE1_CHROMOSOME = "Site1_Chromosome";
+    public static final String SITE2_EXON = "Site2_Exon";
+    public static final String LENGTH = "Length";
+    public static final String SITE1_ENSEMBL_TRANSCRIPT_ID = "Site1_Ensembl_Transcript_Id";
+    public static final String SITE2_ENSEMBL_TRANSCRIPT_ID = "Site2_Ensembl_Transcript_Id";
+    public static final String NORMAL_VARIANT_COUNT = "Normal_Variant_Count";
+
+    public StructuralVariantUtil(){}
```
sheridancbio: If an instance is created with the default constructor, it seems like it will be non-functional because the column indices will not have been set and there is no other mechanism to set them aside from the other constructor. Perhaps this should be declared private so that it prevents use of the default constructor.

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,162 @@
...
+
+    public StructuralVariantUtil(String line) {
```
sheridancbio: the line must be the header line --- maybe the parameter name should be "headerLine" so that readers who look at this constructor code will know immediately what is expected.

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,162 @@
...
+        // Find header indices
+        for (int i=0; i<headerParts.length; i++) {
+            // Put the index in the map
+            this.columnIndexMap.put(headerParts[i].toLowerCase(), i);
+        }
+    }
+
+    public StructuralVariant parseStructuralVariantRecord(String[] parts) {
+        StructuralVariant structuralVariant = new StructuralVariant();
+        structuralVariant.setSampleId(TabDelimitedFileUtil.getPartString(getColumnIndex(StructuralVariantUtil.SAMPLE_ID), parts));
```
sheridancbio: I wonder what will happen if a column is missing from the input file ... getColumnIndex will return -1 ... I suppose getPartString(-1, parts) will return an appropriate Null / NA value rather than throwing an ArrayBoundsException?

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
```diff
@@ -0,0 +1,162 @@
...
+    public int getColumnIndex(String colName) {
+        Integer index = this.columnIndexMap.get(colName.toLowerCase());
+        if (index == null) {
+            index = -1;
+        }
+        return index;
+    }
+
+    public Boolean hasRequiredStructuralVariantFields(StructuralVariant record) {
+        return !record.getSite1EnsemblTranscriptId().equalsIgnoreCase(TabDelimitedFileUtil.NA_STRING) &&
```
sheridancbio: This is fine if we want to filter out all records which are not complete with the full set of information for visualization. I think @n1zea144 suggested that we could have a two-level approach though ... where we let through records which have none of these fields filled in; and also let through records which have all of these fields filled in. We would only filter records which have *at least one, but not all* of these fields filled in. This is because the simpler records which were missing transcripts and exons still were being presented in the frontend --- just without the extra visualization showing the exons and the junction between members of the fusion. So these primitive records would just show that there was *some* kind of fusion which happened between to the two genes, but with no details. I think @n1zea144 wanted to allow these primitive fusions to go through ... and also let the fully detailed fusions go through. We would filter the "a little extra detail but not enough" records. (In my mind we could drop the partial extra details and turn it into a primitive record, but I think that is the kind of data fixing we are now trying to avoid)

core/src/test/java/org/mskcc/cbio/portal/util/TestStructuralVariantUtil.java
```diff
@@ -0,0 +1,46 @@
...
+        record.setSite1HugoSymbol("NA");
+        record.setSite1EnsemblTranscriptId("ENST29O3403");
+        record.setSite1Exon(2);
+        record.setSite2EntrezGeneId(-1L);
+        record.setSite2HugoSymbol("ALK1");
+        record.setSite2EnsemblTranscriptId("NA");
+        record.setSite2Exon(1);
+        Assert.assertFalse(structuralVariantUtil.hasRequiredStructuralVariantFields(record));
+    }
+
```
sheridancbio: If we allow the primitive sort of fusions I mentioned above to go through, then we might have three tests --- a valid primitive fusion, a valid detailed fusion, and an invalid "almost detailed" fusion.

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
ao508: @sheridancbio I needed the default constructor for unit testing

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
ao508: yes, exactly

core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
ao508: @sheridancbio Oh I see, basically allow records with all or "nothing" filled in for those key columns. I just want to confirm with @n1zea144 as well since the (now removed) file format documentation for SV data says that the site 1 and 2 exons and ensembl transcript id fields [are required](https://github.com/cBioPortal/cbioportal/blob/da5740b5c450758f0a90e3aa5cd6274c18882c0a/docs/File-Formats.md#structural-variants-data).

PR #6725, opened 2019-10-18
Fix Swagger Scheme Defaults

### LINKED ISSUE #5760: Swagger API default scheme is http
When browsing the API here:

https://www.cbioportal.org/api

It lists http as the default scheme, which gives errors on public portal when clicking on execute. We should maybe only show the https scheme when browsing over https, and http when browsing over http

Luke-Sikina:
Fix #5760
- Get rid of the specification to make swagger (hopefully) default to what the application runs on


inodb:
Tested it on Heroku. That supports both HTTP and HTTPS.  It does indeed use HTTPS when accessing over HTTPS and uses HTTP when accessing over HTTP. Nice work!

PR #6731, opened 2019-10-23
hide fusion gene table in the summary view if no fusion data present …

### LINKED ISSUE #6709: fusion tables
The new fusion tables in study view are a great addition, but I suspect that they are currently misleading.

I visited study view for a random sampling of studies, and most have an empty fusion table:
image
https://www.cbioportal.org/study/summary?id=cesc_tcga

I suspect that for many of these studies, it is not that there are truly no fusions, but rather that fusions weren't called. If true, this is very misleading because the table clearly says "XXX profiled samples".

How are these tables being populated? Do they depend on a case list that doesn't exist yet?

In the meantime, perhaps we can hide these tables by default so they are less obvious/misleading to our users? @schultzn @jjgao  thoughts?

khzhu:
hide the fusion gene table in the summary view if no fusion data present and fix the number of samples for fusion.

Fix #6709 

- [ ] add a new fusion case list
- [ ] update migration.sql script to back fill fusion case lists for studies with fusion data.
- [ ] add a new flag fusionPresent to sample model
- [ ] add new properties numberOfFusionProfiledSamples/numberOfFusionUnprofiledSamples in MolecularProfileSampleCount model
- [ ] update sample-counts web service endpoint to include the number of Profiled/Un-profiled fusion Samples.

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


khzhu:
@zhx828 , in SampleServiceImpl processSample method.

khzhu:
@alisman , here is the backend PR.

khzhu:
Hi @sheridancbio , would you be able to review this PR. It will fix the [problem](https://github.com/cBioPortal/cbioportal/issues/6709) reported by Tali. Thank you!

alisman:
@kalletlak do frontned/backend need to be deployed together?   I.E. will either break if they deployed separately?

zhx828:
@khzhu let me rephrase, where the logic(script) to populate the sample list if the study has the fusion data in the database?

kalletlak:
> @kalletlak do frontned/backend need to be deployed together? I.E. will either break if they deployed separately?

@alisman Yes

khzhu:
Hi @zhx828 , I did not change any existing logic to populate the sample list other than added a new case list **_cases_fusion_** to list all samples that have fusion data.
The following line I added to the SampleServiceImpl class that will mark any samples as **_fusion present_** if the sample is listed in the cases_fusions list. 
```
sample.setFusionPresent(fusionSampleIdsMap.get(sample.getCancerStudyIdentifier()).contains(sample.getStableId()));
```
Hope it answers your question.

zhx828:
@khzhu if this is the case,  @sheridancbio  could you point  Kelsey to the pipeline code where the logic determines whether the fusion list should be populated? @khzhu could you help to make a patch to that code too?

khzhu:
@zhx828 , will do.

sheridancbio:
@zhx828 @khzhu I'm reading up on this now. From a first read-through it looks like you introduce a new standard sample list to hold all samples which underwent fusion profiling. Then the presence or absence of list elements determines whether this study has any fusion data to display or not.

Two very quick thoughts:
- I think there are 4 current standard case lists : _all, _sequenced, _cna, _cnaseq ... The meaning of the case lists seem a little mixed already. "_sequenced" perhaps is all samples which have any genetic profiling performed at all. _cna perhaps means samples for which a detected cna alteration event is present. Adding _fusion would be "those samples which were characterized on a platform which can detect fusion events?" I'm just trying to understand the meaning of our standard case list files to see if this fits in with the other case lists.
- If there is now going to be a file in the import directory such as staging_dir/case_lists/cases_fusion.txt I wonder if it will be a required file or not. If it is missing and we still have fusion events in data_SV.txt or data_mutations_extended.txt I wonder what we should do. Perhaps the validator script needs to be updated for cases like this. When deploying this code we will need to make adjustments to the msk importer code base to generate this _fusion case list ... and I suppose we will need to infer which samples were analyzed for fusions based on whether the mutation sequencing and analysis was a protocol which can detect/report fusions or not (as well as other factors). It may require some approximations.

Just to be clear ... it *is* possible in your minds that there can be samples on the _fusion case list for which no fusions were detected. (in other words .. profiled but nothing found) ? cc: @n1zea144

khzhu:
hi @sheridancbio , thank for your inputs. 
to answer your questions:
> Adding _fusion would be "those samples which were characterized on a platform which can detect fusion events?"

yes, that is right.

> If there is now going to be a file in the import directory such as staging_dir/case_lists/cases_fusion.txt I wonder if it will be a required file or not.

it will not be a required file. 

> If it is missing and we still have fusion events in data_SV.txt or data_mutations_extended.txt I wonder what we should do.

we will not add cases_fusion.txt file for both. cases_fusion.txt will be eventually deprecated once we started to use structural variants.

n1zea144:
@khzhu @sheridancbio 

My two cents - 
I would follow the same pattern that is used to display the mutation x cna plot, mutated genes & cna genes table, etc.  (which I imagine is being done in this PR).  Curious, if it is done by the presence or absence of a case list, how is the rendering of these charts determined on virtual cohort study summary pages?

jjgao:
one minor comment: should we use cases_fusion or cases_sv? (assume we will use the same case list after switch to the sv implementation)

db-scripts/src/main/resources/migration.sql
```diff
@@ -720,3 +720,22 @@ CASE
...
+            JOIN `cancer_study` ON `cancer_study`.CANCER_STUDY_ID = `genetic_profile`.CANCER_STUDY_ID
+        WHERE MUTATION_TYPE = 'fusion';
+REPLACE INTO sample_list_list(list_id, sample_id)
+    SELECT DISTINCT list_id, SAMPLE_ID
+        FROM `mutation_event`
+            JOIN `mutation` ON `mutation`.MUTATION_EVENT_ID = `mutation_event`.MUTATION_EVENT_ID
+            JOIN `genetic_profile` ON `genetic_profile`.GENETIC_PROFILE_ID = `mutation`.GENETIC_PROFILE_ID
+            JOIN `cancer_study` ON `cancer_study`.CANCER_STUDY_ID = `genetic_profile`.CANCER_STUDY_ID
+            JOIN `sample_list` ON `sample_list`.CANCER_STUDY_ID = `cancer_study`.CANCER_STUDY_ID AND `sample_list`.STABLE_ID LIKE '%fusion%'
+        WHERE MUTATION_TYPE = 'fusion';
```
jjgao: My understanding of this code is that a default _fusion sample list will be created which is the list of samples that have fusion data. There is a caveat: some profiled samples may be fusion negative. For example, many MSK-IMPACT samples do not have fusions but they were all profiled.

This might be fine as the default one for migration, but we should generate proper _fusion case list for our studies.

Alternatively, I am wondering if we should just copy _sequenced to _fusion when there are fusion data?

@n1zea144 @sheridancbio thoughts?

khzhu: @n1zea144 @SRodenburg , any inputs on this? thanks!

n1zea144: Hi @jjgao Rob and I were talking about this too.  Using MSK-IMPACT-ARCHER as an example, I'm not sure if its 100% accurate to assume that if a sample was sequenced for mutation calling, that it was also analyzed for fusions (although it is true that with impact, there are some probes looking for canonical fusions).  However, it seems excessive to have a separate list for cases that were checked for SVs, and then another case list for samples in which SVs were found.

khzhu: thanks, @n1zea144 ! 
but how do we know if there are fusions or not at the importing stage to decide if we should carry the sequenced sample list over to fusions?

khzhu: here is my solution:
1. use _fusion case list to add fusion positive samples to the sample list table.
2. SampleServiceImpl populates fusionSampleMap using fusion case list, but uses sequencedSampleIdsMap to actually mark samples **_fusion present_** if fusionSampleMap is not empty.

what do you think, @jjgao , @n1zea144 , @sheridancbio?

khzhu:
> Alternatively, I am wondering if we should just copy _sequenced to _fusion when there are fusion data?

I tried JJ's alternative approach and tested on msk_impact_2017 dataset. And, it worked well.
should we go with this plan, @jjgao @n1zea144 @sheridancbio?
1. for existing datasets that have fusions, we backfill fusion sample case list using samples recorded in the **_mutation_event_** table with mutation_type = 'fusion'
2. for new dataset that has fusions, we will just copy _sequenced to _fusion before uploading.

jjgao:
@khzhu I am not sure if I understand correctly, but I think we should:

1. for existing datasets that have fusions, we backfill fusion sample case list with the _sequenced case list. 
2. for a new dataset that has fusions, if there is _fusions case list, use that; otherwise,  copy _sequenced to _fusion before uploading.

This will ensure that our current database will work after migration, and our current data files will work. Also once we put _fusion case lists, they will also work.

BTW, should we call it _sv case list?

khzhu:
> for existing datasets that have fusions, we backfill fusion sample case list with the _sequenced case list.

will update migration sql statement

>  for a new dataset that has fusions, if there is _fusions case list, use that; otherwise, copy _sequenced to _fusion before uploading.

sounds good.

> BTW, should we call it _sv case list?

go with _sv

khzhu:
Hi @jjgao @n1zea144 @sheridancbio ,
Recap changes made based on inputs from you all.
- [ ] create a new **__sv_** case list and add a new **_all_cases_with_structural_variant_data_** case list category
- [ ] for existing datasets that have fusions, backfill fusion sample case list from **__sequenced_** case list.
- [ ] for new dataset with fusions, if **__sv_** case list supplied, use it; otherwise, create **__sv_** case list from **__sequenced_** case list before uploading.

may I please have your final review on the changes. Thanks!

core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
```diff
@@ -42,6 +42,7 @@
 public enum SampleListCategory {
     ALL_CASES_IN_STUDY("all_cases_in_study"),
     ALL_CASES_WITH_MUTATION_DATA("all_cases_with_mutation_data"),
+	ALL_CASES_WITH_FUSION_DATA("all_cases_with_fusion_data"),
```
zhx828: I guess this is no longer needed?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -173,11 +174,15 @@ private void processSamples(List<Sample> samples, String projection) {
 
         if (projection.equals("DETAILED")) {
             Map<String, Set<String>> sequencedSampleIdsMap = new HashMap<>();
+            Map<String, Set<String>> fusionSampleIdsMap = new HashMap<>();
```
zhx828: Should we call this structuralVariantSampleIdsMap?

core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
ao508: whitespace/alignment inconsistency

core/src/main/scripts/importer/validateData.py
```diff
@@ -4243,6 +4243,7 @@ def processCaseListDirectory(caseListDir, cancerStudyId, logger,
                                 'all_cases_with_mutation_and_cna_data',
```
ao508: Shouldn't `all_cases_with_fusion_data` need to be added here as well since [it is in SampleListCategory](https://github.com/cBioPortal/cbioportal/pull/6731/files#diff-d8591fa1bafcfbe2f3dc97c663e87c28R45)?

db-scripts/src/main/resources/migration.sql
```diff
@@ -720,3 +720,36 @@ CASE
     ELSE 1
 END;
 UPDATE `info` SET `DB_SCHEMA_VERSION`="2.11.0";
+
+##version 2.11.1
+CREATE TEMPORARY TABLE IF NOT EXISTS
+    fusion_studies ( INDEX(CANCER_STUDY_IDENTIFIER) )
+    ENGINE=MyISAM
```
ao508: Does the engine on these temporary tables have to be `MyISAM`? I thought our defaults were `InnoDB` - please correct me if I'm wrong.

model/src/main/java/org/cbioportal/model/Sample.java
```diff
@@ -119,11 +120,13 @@ public void setSequenced(Boolean sequenced) {
...
-    }
+    public Boolean getCopyNumberSegmentPresent() { return copyNumberSegmentPresent; }
 
     public void setCopyNumberSegmentPresent(Boolean copyNumberSegmentPresent) {
         this.copyNumberSegmentPresent = copyNumberSegmentPresent;
     }
+
+    public Boolean getFusionPresent() { return fusionPresent; }
+
+    public void setFusionPresent(Boolean fusionPresent) { this.fusionPresent = fusionPresent; }
```
ao508: Can you format these methods the same was as the existing methods in this class?

core/src/main/scripts/importer/validateData.py
khzhu: we will replace
```
ALL_CASES_WITH_FUSION_DATA("all_cases_with_fusion_data")
```
with
```
ALL_CASES_WITH_STRUCTURAL_VARIANT_DATA("all_cases_with_structural_variant_data"),
```

db-scripts/src/main/resources/migration.sql
khzhu: Either should work. Only needs to index one field. MyISAM is lighter and faster than InnoDB.

model/src/main/java/org/cbioportal/model/Sample.java
khzhu: done.

core/src/main/scripts/importer/validateData.py
khzhu: all_cases_with_fusion_data removed.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
khzhu: good idea. renamed.

sheridancbio:
One additional change request is that docs/FileFormats.md contains the list of valid case list categories, so this doc should be updated to include the _fusions case list type and we should discuss whether the case list is required, or what should happen when the case list is absent. (perhaps validator should raise an error if there is no case list for fusions, but some fusion events are present in the data for example?)

khzhu:
good point, @sheridancbio. 
will update docs. to include the _sv (renamed _fusion to _sv, as @jjgao suggested) case list type as an optional. Agree with Rob, validator should raise an error if there is no _sv case list but some fusion/SV events are present in the data.

sheridancbio:
This may be a side issue, but while looking through the api endpoints I see that the study model also reports sample counts in some of the key / required categories. For example, the response to curl -X GET "https://www.cbioportal.org/api/studies/msk_impact_2017" -H "accept: application/json" includes counts like:
"allSampleCount": 0,
"cnaSampleCount": 0,
"miRnaSampleCount": 0,
"mrnaRnaSeqSampleCount": 0,
"mrnaRnaSeqV2SampleCount": 0,
"rppaSampleCount": 0,
"sequencedSampleCount": 0,

I am not sure why these all were returned as zero ... If this is a deprecated feature that we no longer support then perhaps we should drop sample counts from the Study model. But if we do still support these data members, it might be appropriate to add structuralVariantSampleCount.
cc:  @n1zea144 @jjgao @ao508

model/src/main/java/org/cbioportal/model/MolecularProfileSampleCount.java
```diff
@@ -9,6 +9,8 @@
 	private Integer numberOfCNAProfiledSamples;
```
sheridancbio: For this work, we have chosen to keep the fusionProfiledSample list parallel to the mutationProfiledSample list ... in other words there can be samples which were profiled for fusions/structural variants and in which no events were found (for any gene). CNA events have a representation in the portal database formatted not as single records representing events (such as mutations and fusions) but as event call vectors ... lists of numbers showing the collection of event (or non-event) codes for every sample in the profile ... one vector per profiled gene. Because of this, data is present inherently for all samples in the profile (even if the data records a non-event). This means that we do not have the same difficulties distinguishing the samples which were not profiled from the samples which were profiled but did not appear in the list of events at all.
However, we have considered moving away from this representation and moving towards a representation which only contains events (and not codes for non-events). If we make any modifications to the CNA representation, we should consider using the cases_cna case_list as the authority for determining which samples were profiled in a way which would detect CNA. So we should look for the first opportunity to make this concept parallel across all molecular_data_profile datatypes.

model/src/main/java/org/cbioportal/model/Sample.java
```diff
@@ -54,6 +54,7 @@ public String toString() {
     private String cancerStudyIdentifier;
     private Boolean sequenced;
     private Boolean copyNumberSegmentPresent;
+    private Boolean fusionPresent;
```
sheridancbio: This property be true even if no fusion events were detected. So I request that this be renamed to profiledForFusions or profiledForStructuralVariants or analyzedForFusions or fusionScanPerformed.

fusionPresent suggests that a fusion was detected in the sample.

model/src/main/java/org/cbioportal/model/Sample.java
sheridancbio: If changing data member name, it will need to be propagated through here.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -173,11 +174,15 @@ private void processSamples(List<Sample> samples, String projection) {
...
             Map<String, Set<String>> sequencedSampleIdsMap = new HashMap<>();
+            Map<String, Set<String>> structuralVariantSampleIdsMap = new HashMap<>();
             List<String> distinctStudyIds = samples.stream().map(Sample::getCancerStudyIdentifier).distinct()
                 .collect(Collectors.toList());
+            
             for (String studyId : distinctStudyIds) {
                 sequencedSampleIdsMap.put(studyId,
                                           new HashSet<String>(sampleListRepository.getAllSampleIdsInSampleList(studyId + SEQUENCED)));
+                structuralVariantSampleIdsMap.put(studyId,
+                                       new HashSet<String>(sampleListRepository.getAllSampleIdsInSampleList(studyId + STRUCTURAL_VARIANT)));
```
sheridancbio: Note for future performance improvements : this seems like a good opportunity for caching ... otherwise every call to the sample service for a DETAILED response is recomputing these lists. I wonder if we could just add a function call in sampleListRepository which returned a HashSet rather than a list. (then it would be cached)

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -193,6 +198,9 @@ private void processSamples(List<Sample> samples, String projection) {
                 sample.setSequenced(sequencedSampleIdsMap.get(sample.getCancerStudyIdentifier())
                     .contains(sample.getStableId()));
                 sample.setCopyNumberSegmentPresent(samplesWithCopyNumberSegMap.contains(sample.getInternalId()));
+                if (!structuralVariantSampleIdsMap.isEmpty()) {
+                    sample.setFusionPresent(structuralVariantSampleIdsMap.get(sample.getCancerStudyIdentifier()).contains(sample.getStableId()));
```
sheridancbio: setProfiledForFusions (or whatever rename is chosen)

db-scripts/src/main/resources/migration.sql
khzhu: okay. will drop the engine reference.

db-scripts/src/main/resources/migration.sql
```diff
@@ -720,3 +720,36 @@ CASE
...
+             JOIN `genetic_profile` ON `genetic_profile`.GENETIC_PROFILE_ID = `mutation`.GENETIC_PROFILE_ID
+             JOIN `cancer_study` ON `cancer_study`.CANCER_STUDY_ID = `genetic_profile`.CANCER_STUDY_ID
+    WHERE MUTATION_TYPE = 'fusion'
+);
+REPLACE INTO sample_list(STABLE_ID, CATEGORY, CANCER_STUDY_ID, NAME, DESCRIPTION)
+SELECT CONCAT(CANCER_STUDY_IDENTIFIER, '_sv'), 'all_cases_with_structural_variant_data', CANCER_STUDY_ID, 'all_cases_with_structural_variant_data','All cases with structural variant data'
+FROM `fusion_studies`;
+CREATE TEMPORARY TABLE IF NOT EXISTS
+    fusion_samples_by_study ( INDEX(cancer_study_id) )
+    ENGINE=MyISAM
```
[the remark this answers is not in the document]
khzhu: will drop it.

core/src/main/scripts/importer/validateData.py
khzhu: to clarify: we are continue to use existing **_all_cases_with_fusion_data_** for fusions and **_all_cases_with_sv_data_** (newly added by PR #6860) for structural variants

db-scripts/src/main/resources/migration.sql
khzhu: done.

db-scripts/src/main/resources/migration.sql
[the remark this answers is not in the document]
khzhu: dropped the engine reference.

model/src/main/java/org/cbioportal/model/Sample.java
khzhu: go with profiledForFusions.

model/src/main/java/org/cbioportal/model/Sample.java
khzhu: adjusted member methods.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
khzhu: good point! we should add a hashSet when we switch to the structural variant table.

khzhu: very good point, @sheridancbio! Let's have a follow up PR on this once we switch to the structural variant table. Thanks!

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
khzhu: done.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -374,6 +374,11 @@
             molecularProfileSampleCount.setNumberOfCNSegmentSamples(Math.toIntExact(sampleService
                 .fetchSamples(studyIds, sampleIds, Projection.DETAILED.name()).stream().filter(
                     s -> s.getCopyNumberSegmentPresent()).count()));
+            molecularProfileSampleCount.setNumberOfFusionProfiledSamples(Math.toIntExact(sampleService
+                .fetchSamples(studyIds, sampleIds, Projection.DETAILED.name()).stream().filter(
+                    s -> s.getFusionPresent()).count()));
```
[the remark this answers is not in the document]
khzhu: done.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -374,6 +374,11 @@
...
+            molecularProfileSampleCount.setNumberOfFusionUnprofiledSamples(sampleCount -
+                molecularProfileSampleCount.getNumberOfFusionProfiledSamples());
```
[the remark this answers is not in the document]
khzhu: Hi @sheridancbio , I've made corrections except for your suggestion on adding an Eh cache to the SampleList Repository which returns a HashSet of all sample Ids. Let's have a follow up PR on that once we switch to the structural variant table. Thanks!

khzhu:
hi @jjgao @n1zea144 @sheridancbio @zhx828 
I added one more sql statement to the migration script that would backfill genetic profiles for the existing fusion data. The fusion Gene table on the summary view will be displayed or hidden depending on if fusion was profiled or not. Details can be found in this [PR](https://github.com/cBioPortal/cbioportal-frontend/pull/2812)
Would you please take a look and let me know your thoughts on this? thanks

core/src/main/scripts/importer/validateData.py
zhx828: If that's the case, I don't think the `all_cases_with_sv_data ` should be deleted?

khzhu: @zhx828 , yes, we keep all_cases_with_sv_data for structural variants. It was added by PR #6860.

jjgao:
@khzhu @alisman @zhx828 what's left for this?

khzhu:
Hi @jjgao , all work has been done and I am waiting for @n1zea144 or @sheridancbio to complete their reviews. Thanks!

inodb:
I created a new PR rebased to release-3.2.0: https://github.com/cBioPortal/cbioportal/pull/6945. Please do further review there

PR #6733, opened 2019-10-23
Allow 'ALL' or 'NONE' SV site 1 and 2 Ensembl transcript ids/exons info for import

ao508:
Allow 'ALL' or 'NONE' SV site 1 and 2 Ensembl transcript ids/exons info for import

     * Structural variant records are valid if:
     *   (1) Neither Site 1 or Site 2 Ensembl transcript IDs and exons are defined in the record.
     *   (2) Both Site 1 and Site 2 Ensmebl transcript IDs and exons are defined in the record.
     *
     * If a structural variant record has a mix of defined and missing values for Site 1 or Site 2
     * Ensembl transcript IDs and/or exons then the structural variant record will not be imported.
     *
     * Example (assuming that site 1 and site 2 hugo symbol and/or entrez id are present):
     *
     * Valid Record:
     *     Site 1 Transcript: EST0000024958
     *     Site 1 Exon: 4
     *     Site 2 Transcript: EST0001651651
     *     Site 1 Exon: 2
     *
     * Valid Record:
     *     Site 1 Transcript: NA
     *     Site 1 Exon: -1
     *     Site 2 Transcript: NA
     *     Site 1 Exon: -1
     *
     * INVALID Record:
     *     Site 1 Transcript: EST0000024958
     *     Site 1 Exon: -1
     *     Site 2 Transcript: EST0001651651
     *     Site 1 Exon: -1

Co-authored by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java
@@ -150,11 +150,49 @@ public class StructuralVariantUtil {
         return index;
     }
 
+    /**
+     * Determines whether the structural variant record meets the minimal requirements for import.
+     *
+     * Structural variant records are valid if:
+     *   (1) Neither Site 1 or Site 2 Ensembl transcript IDs and exons are defined in the record.
+     *   (2) Both Site 1 and Site 2 Ensmebl transcript IDs and exons are defined in the record.
+     *
+     * If a structural variant record has a mix of defined and missing values for Site 1 or Site 2
+     * Ensembl transcript IDs and/or exons then the structural variant record will not be imported.
+     *
+     * Example (assuming that site 1 and site 2 hugo symbol and/or entrez id are present):
+     *
+     * Valid Record:
+     *     Site 1 Transcript: EST0000024958
+     *     Site 1 Exon: 4
+     *     Site 2 Transcript: EST0001651651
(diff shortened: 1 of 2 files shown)


sheridancbio:
@n1zea144 has pointed out that in the original structural variant specification (which has been removed temporarily from the cbioportal code base .. see #6644) has filtering rules that specify three required things for all structural variants (meaning both the simpler and the more specified format) :
- sample id
- site1_entrez_gene_id and/or site1_hugo_symbol
- event info

It also describes the additional required fields for the more specified format (which it calls "fusion events"):
- site2_entrez_gene_id and/or site1_hugo_symbol
- site1_exon
- site2_exon
- site1_ensembl_transcript_id
- site2_ensembl_transcript_id

We need to check that this code conforms with these specifications. Maybe the correct filtering is being done at an earlier stage (during file parsing) for the basic requirements .. but in this code we are combining transcript_id and hugo_symbol rather than entrez_gene_id and hugo_symbol, so we should trace deeper before merging.

ao508:
@sheridancbio Checks for sample id and that the site 1 and site 2 genes are done in the `ImportStructuralVariantData` class. see [here](https://github.com/ao508/cbioportal/blob/sv-unit-tests/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportStructuralVariantData.java#L80)

PR #6738, opened 2019-10-24
Add RefGenomeGene to EHCache static cache

ao508:

# Checks
- [ ] Runs on heroku

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/ReferenceGenomeGeneRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/ReferenceGenomeGeneRepository.java
@@ -2,11 +2,18 @@ package org.cbioportal.persistence;
 
 import org.cbioportal.model.ReferenceGenomeGene;
 import java.util.List;
+import org.springframework.cache.annotation.Cacheable;
 
 public interface ReferenceGenomeGeneRepository {
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getAllGenesByGenomeName(String genomeName);
+    @Cacheable(cacheNames = "GeneralRepositoryCache", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getGenesByHugoGeneSymbolsAndGenomeName(List<String> geneIds, String genomeName);
+    @Cacheable(cacheNames = "GeneralRepositoryCache", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getGenesByGenomeName(List<Integer> geneIds, String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     ReferenceGenomeGene getReferenceGenomeGene(Integer geneId, String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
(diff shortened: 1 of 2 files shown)


persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomEhCachingProvider.java
```diff
@@ -227,13 +227,15 @@ public void detectCacheConfigurationErrorsAndLog() {
                 File persistenceDirectory = new File(persistencePath);
                 boolean accessible = false;
                 try {
-                    if (persistenceDirectory.isDirectory()) {
+                    if (persistenceDirectory.isDirectory() && persistenceDirectory.canWrite()) {
```
averyniceday: 👍

ao508: I added this check to prevent silently failing when a valid persistence directory is provided but the user does not have write permissions to this directory

persistence/persistence-api/src/main/java/org/cbioportal/persistence/ReferenceGenomeGeneRepository.java
```diff
@@ -2,11 +2,17 @@
...
 public interface ReferenceGenomeGeneRepository {
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getAllGenesByGenomeName(String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getGenesByHugoGeneSymbolsAndGenomeName(List<String> geneIds, String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getGenesByGenomeName(List<Integer> geneIds, String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     ReferenceGenomeGene getReferenceGenomeGene(Integer geneId, String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
```
sheridancbio: I'm a little confused by this function. I don't think end users can retrieve geneticEntityId values ... but this is the persistence layer, so it could be that other service layer logic looks up genetic entity ids and then queries for gene metadata through this function. It is OK that this be in the StaticRepositoryCacheOne space .. but it does seem like a waste of cache space if this function could be implemented by delegating/calling another function, such as getReferenceGeneomeGene.  Actually, I wonder what the Integer geneId argument to that function is supposed to accept. But I guess that is outside the scope of this PR.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/ReferenceGenomeGeneRepository.java
```diff
@@ -2,11 +2,17 @@
 
 import org.cbioportal.model.ReferenceGenomeGene;
 import java.util.List;
+import org.springframework.cache.annotation.Cacheable;
 
 public interface ReferenceGenomeGeneRepository {
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
     List<ReferenceGenomeGene> getAllGenesByGenomeName(String genomeName);
+    @Cacheable(cacheNames = "StaticRepositoryCacheOne", condition = "@cacheEnabledConfig.getEnabled()")
```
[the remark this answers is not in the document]
ao508: done!

PR #6746, opened 2019-10-25
Sample Quicksearch nolonger 403s

### LINKED ISSUE #6745: Quick Search Displays No Results w/ Auth
Results are not displayed because the samples endpoint 403s.

Luke-Sikina:
Fix #6745
- moved filtering to service layer to avoid 403

How can we verify that this works? Right now this is unverified.

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/SampleRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/SampleRepository.java
@@ -9,10 +9,10 @@ import java.util.List;
 
 public interface SampleRepository {
     @Cacheable(cacheNames = "GeneralRepositoryCache", condition = "@cacheEnabledConfig.getEnabled()")
-    List<Sample> getAllSamples(String keyword, String projection, Integer pageSize,
+    List<Sample> getAllSamples(String keyword, List<String> studyIds, String projection, Integer pageSize,
                                Integer pageNumber, String sort, String direction);
 
-    BaseMeta getMetaSamples(String keyword);
+    BaseMeta getMetaSamples(String keyword, List<String> studyIds);
 
     @Cacheable(cacheNames = "GeneralRepositoryCache", condition = "@cacheEnabledConfig.getEnabled()")
     List<Sample> getAllSamplesInStudy(String studyId, String projection, Integer pageSize, Integer pageNumber,
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepository.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepository.java
@@ -20,15 +20,19 @@ public class SampleMyBatisRepository implements SampleRepository {
     private OffsetCalculator offsetCalculator;
 
     @Override
-    public List<Sample> getAllSamples(String keyword, String projection, Integer pageSize, Integer pageNumber, String sort, String direction) {
-        
-        return sampleMapper.getSamples(null, null, null, keyword, projection, pageSize,
-            offsetCalculator.calculate(pageSize, pageNumber), sort, direction);
+    public List<Sample> getAllSamples(
+        String keyword, List<String> studyIds, String projection,
+        Integer pageSize, Integer pageNumber, String sort, String direction
+    ) {
+        return sampleMapper.getSamples(
+            studyIds, null, null, keyword, projection, pageSize,
+            offsetCalculator.calculate(pageSize, pageNumber), sort, direction
+        );
     }
 
     @Override
-    public BaseMeta getMetaSamples(String keyword) {
-        return sampleMapper.getMetaSamples(null, null, null, keyword);
+    public BaseMeta getMetaSamples(String keyword, List<String> studyIds) {
+        return sampleMapper.getMetaSamples(studyIds, null, null, keyword);
     }
 
     @Override
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/SampleMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/SampleMapper.xml
@@ -33,9 +33,14 @@
         <where>
             <if test="sampleIds == null and studyIds != null">
                 cancer_study.CANCER_STUDY_IDENTIFIER IN
-                <foreach item="item" collection="studyIds" open="(" separator="," close=")">
-                    #{item}
-                </foreach>
+                <if test="studyIds.isEmpty()">
+                    (NULL)
+                </if>
+                <if test="!studyIds.isEmpty()">
+                    <foreach item="item" collection="studyIds" open="(" separator="," close=")">
+                        #{item}
+                    </foreach>
+                </if>
             </if>
             <if test="sampleIds != null">
                 <if test="@java.util.Arrays@stream(studyIds.toArray()).distinct().count() == 1">
@@ -56,6 +61,7 @@
                 AND patient.STABLE_ID = #{patientId}
             </if>
             <if test="keyword != null">
+                AND
                 <foreach item="item" collection="keyword.split(' ')" open="(" separator=") AND (" close=")">
                     sample.STABLE_ID like CONCAT('%', #{item}, '%')
                 </foreach>
--- a/persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepositoryTest.java
+++ b/persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepositoryTest.java
@@ -340,7 +340,7 @@ public class SampleMyBatisRepositoryTest {
      
     @Test
     public void getSamplesByKeyword() {
-         List<Sample> result = sampleMyBatisRepository.getAllSamples("TCGA-A1-A0SB", "SUMMARY", 10, 0, null, null);
+         List<Sample> result = sampleMyBatisRepository.getAllSamples("TCGA-A1-A0SB", null, "SUMMARY", 10, 0, null, null);
          List<String> actual = result.stream().map((Sample::getStableId)).collect(Collectors.toList());
          List<String> expected = Arrays.asList("TCGA-A1-A0SB-01", "TCGA-A1-A0SB-01", "TCGA-A1-A0SB-02");
 
@@ -349,7 +349,7 @@ public class SampleMyBatisRepositoryTest {
      
     @Test
     public void getMetaSamplesByKeyword() {
-        BaseMeta actual = sampleMyBatisRepository.getMetaSamples("TCGA-A1-A0SB");
+        BaseMeta actual = sampleMyBatisRepository.getMetaSamples("TCGA-A1-A0SB", null);
         Integer expected = 3;
         
         Assert.assertEquals(expected, actual.getTotalCount());
--- a/service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
@@ -13,7 +13,9 @@ import org.cbioportal.service.exception.PatientNotFoundException;
 import org.cbioportal.service.exception.SampleNotFoundException;
 import org.cbioportal.service.exception.StudyNotFoundException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
+import org.springframework.security.access.prepost.PostFilter;
 import org.springframework.stereotype.Service;
 
 import java.util.*;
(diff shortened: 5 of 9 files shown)


service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -36,6 +37,7 @@
     private CopyNumberSegmentRepository copyNumberSegmentRepository;
     
     @Override
+    @PostFilter("hasPermission(filterObject, 'read')")
```
kalletlak: Can you check if `org.cbioportal.security.spring.CancerStudyPermissionEvaluator` `hasPermission` method supports `Sample` target object?

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,43 @@
...
+                )
+                .stream()
+                .map(CancerStudy::getCancerStudyIdentifier)
+                .collect(Collectors.toList());
+        }
+
         if (projection == Projection.META) {
             HttpHeaders httpHeaders = new HttpHeaders();
-            httpHeaders.add(HeaderKeyConstants.TOTAL_COUNT, sampleService.getMetaSamples(keyword).getTotalCount().toString());
+            if (studyIds != null && studyIds.isEmpty()) {
```
Luke-Sikina: I do not like this solution, and I am open to suggestions. The problem I was having was, when the user has permission to access no studies, the search uses an empty list. Depending on my implementation, the SQL query was either malformed, or returned all matching studies for the keyword. What I need it to do is return no samples when the list of study ids is empty, but not null.

There _has_ to be a cleaner way to do this in `SampleMapper.xml`. Help?

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,43 @@
...
+                HeaderKeyConstants.TOTAL_COUNT, 
+                studyIds != null && studyIds.isEmpty() ? 
+                    "0" :
+                    sampleService.getMetaSamples(keyword, studyIds).getTotalCount().toString());
             return new ResponseEntity<>(httpHeaders, HttpStatus.OK);
         }
         return new ResponseEntity<>(
-            sampleService.getAllSamples(keyword, projection.name(), pageSize, pageNumber, sort, direction.name()),
+            studyIds != null && studyIds.isEmpty() ? 
+                new ArrayList<>() :
```
Luke-Sikina: Same with this gross stuff down here.

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,43 @@
...
         if (projection == Projection.META) {
             HttpHeaders httpHeaders = new HttpHeaders();
-            httpHeaders.add(HeaderKeyConstants.TOTAL_COUNT, sampleService.getMetaSamples(keyword).getTotalCount().toString());
+            if (studyIds != null && studyIds.isEmpty()) {
+                httpHeaders.add(HeaderKeyConstants.TOTAL_COUNT, "0");
+            }
+            httpHeaders.add(
+                HeaderKeyConstants.TOTAL_COUNT, 
+                studyIds != null && studyIds.isEmpty() ? 
+                    "0" :
```
Luke-Sikina: This ugliness is part of the mess that I made to overcome this problem.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/SampleMapper.xml
```diff
@@ -31,7 +31,7 @@
 
     <sql id="where">
         <where>
-            <if test="sampleIds == null and studyIds != null">
+            <if test="sampleIds == null and studyIds != null and !studyIds.isEmpty()">
```
kalletlak: instead why not set `IN` clause to `NULL` at line 36
```suggestion
            <if test="sampleIds == null and studyIds != null">
                cancer_study.CANCER_STUDY_IDENTIFIER IN
                <if test="studyIds.isEmpty()">
                    (NULL)
                </if>
                <if test="!studyIds.isEmpty()">
                    <foreach item="item" collection="studyIds" open="(" separator="," close=")">
                        #{item}
                    </foreach>
                </if>
            </if>
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/SampleRepository.java
```diff
@@ -9,10 +9,10 @@
 
 public interface SampleRepository {
     @Cacheable(cacheNames = "GeneralRepositoryCache", condition = "@cacheEnabledConfig.getEnabled()")
-    List<Sample> getAllSamples(String keyword, String projection, Integer pageSize,
+    List<Sample> getAllSamples(String keyword, List<String> studyIds, String projection, Integer pageSize,
```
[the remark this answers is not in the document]
Luke-Sikina: If you want to, we can. I was following the naming convention as seen in `PatientController`, and `GeneController`. In both instances, the method name is `getAll...` despite the fact that it has a keyword parameter.  
I agree this is misleading, and I'm happy to change it, but maybe it would be better to do it in all the places.

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,37 @@
         @RequestParam(defaultValue = "ASC") Direction direction
     ) {
         String sort = sortBy == null ? null : sortBy.getOriginalValue();
-        
+        List<String> studyIds = null;
+        if (usingAuth()) {
+            /*
+             If using auth, filter the list of samples returned using the list of study ids the
+             user has access to. If the user has access to no studies, the endpoint should not 403,
+             but instead return an empty list.
```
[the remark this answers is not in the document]
sheridancbio: I agree .. empty list is appropriate for users who do not have access to any matching samples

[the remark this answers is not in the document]
Luke-Sikina: I'm glad we're all in agreement. That is what I implemented and wrote in the comment, unless I'm missing something?

persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepositoryTest.java
```diff
@@ -340,7 +340,7 @@ public void getSamplesByInternalIds() throws Exception {
      
     @Test
     public void getSamplesByKeyword() {
-         List<Sample> result = sampleMyBatisRepository.getAllSamples("TCGA-A1-A0SB", "SUMMARY", 10, 0, null, null);
+         List<Sample> result = sampleMyBatisRepository.getAllSamples("TCGA-A1-A0SB", null, "SUMMARY", 10, 0, null, null);
```
sheridancbio: here, the unit test is being run in an unauthenticated mode I guess ... where we can pass "null" as the list of available studies and we still get back the expected response ... because all studies in the test database are considered public I guess. This means that this unit test is not testing the study-filtering logic in the SampleMapper. I am assuming the proper behavior was observed by running this code on an authenticated portal where the studies which a user has access to are a factor in the response from the database queries.

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -45,14 +49,25 @@
...
+    private StudyService studyService;
 
     @Autowired
     private UniqueKeyExtractor uniqueKeyExtractor;
 
-    @PreAuthorize("hasPermission(#studyId, 'CancerStudyId', 'read')")
+    @Value("${authenticate:false}")
+    private String authenticate;
+    
+    private boolean usingAuth() {
```
sheridancbio: this is the fourth implementation of logic determining whether the portal is running in an "open to the public" or an authenticated / authorized access mode. We should soon create a shared utility function which can be called here and other places (if possible even during the wiring of applicationContext-security). (see core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java, security/security-spring/src/main/resources/applicationContext-security.xml)

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -45,14 +49,25 @@
...
+        return !authenticate.isEmpty()
+            && !authenticate.equals("false")
+            && !authenticate.contains("social_auth");
+    }
+    
     @RequestMapping(value = "/samples", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
-    @ApiOperation("Get all samples")
-    public ResponseEntity<List<Sample>> getAllSamples (
+    @ApiOperation("Get all samples matching keyword")
+    public ResponseEntity<List<Sample>> getSamplesByKeyword(
```
sheridancbio: 👍 good rename

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,37 @@
...
-        
+        List<String> studyIds = null;
+        if (usingAuth()) {
+            /*
+             If using auth, filter the list of samples returned using the list of study ids the
+             user has access to. If the user has access to no studies, the endpoint should not 403,
+             but instead return an empty list.
+            */
+            studyIds = studyService
+                .getAllStudies(
```
sheridancbio: this call will be PostFiltered according to user authorities

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,37 @@
...
+                .stream()
+                .map(CancerStudy::getCancerStudyIdentifier)
+                .collect(Collectors.toList());
+        }
+
         if (projection == Projection.META) {
             HttpHeaders httpHeaders = new HttpHeaders();
-            httpHeaders.add(HeaderKeyConstants.TOTAL_COUNT, sampleService.getMetaSamples(keyword).getTotalCount().toString());
+            httpHeaders.add(
+                HeaderKeyConstants.TOTAL_COUNT,
```
sheridancbio: as a line continuation indentation, I would indent this 8 rather than 4 (personal preference)

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -75,14 +90,37 @@
...
+                sampleService.getMetaSamples(keyword, studyIds).getTotalCount().toString()
+            );
```
sheridancbio: This is the continuation of line 116. I think our general style is to close function calls on the last line containing arguments rather than pushing the closing parenthesis for the call to sit on its own line. So in this case, the pattern I have seen most often would be to have line 118 end with ".toString());"
If this closing parenthesis is to remain on its own line, the google style requires indentation.
I think we could have a project-wide discussion on style .. I think you made the point that having fewer connecting points in an argument list allows changes which reduce the number of changed lines .. which would be nice. In that case, possibly we should move to starting each line in an argument continuation line with a comma, rather than ending with a comma, and also move the closing parenthesis to a line on its own. But if we adopt that style we should apply it uniformly throughout our codebase.

sheridancbio:
> If we decide to go with way on how auth is added into `SampleController`, for consistency across the codebase `PatientController` needs to be updated.

This is a good catch .. yes, PatientService.getAllPatients() at least should be post-filtered in authenticated portals. And the other endpoints in PatientService should perhaps reduce the set of studies provided in arguments so that only studies the user has access to are passed on in calls to the persistence layer.

sheridancbio:
After checking, I see that PatientService.getAllPatients() _is_ already filtered with `@PostFilter` ... I guess the previous comment meant that the linear filtering of the full list of every patient has the same drawbacks as the linear filtering of the full list of all samples, and that on authenticated portals we should also do the passing of the authorized study list down into the persistence layer so that quicksearch filtering of patients by keywords (I'm not sure if we need that yet, but why not?) could also work correctly and not construct huge temporary structures. @kalletlak

inodb:
@sheridancbio @Luke-Sikina @kalletlak since this addresses a critical issue ill merge it so I can include it in this week's release. You can submit a new PR to follow up on the comments

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepository.java
```diff
@@ -20,15 +20,19 @@
...
 
     @Override
-    public List<Sample> getAllSamples(String keyword, String projection, Integer pageSize, Integer pageNumber, String sort, String direction) {
-        
-        return sampleMapper.getSamples(null, null, null, keyword, projection, pageSize,
-            offsetCalculator.calculate(pageSize, pageNumber), sort, direction);
+    public List<Sample> getAllSamples(
+        String keyword, List<String> studyIds, String projection,
+        Integer pageSize, Integer pageNumber, String sort, String direction
+    ) {
```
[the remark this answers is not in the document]
Luke-Sikina: Ok, so to summarize: indent continuations 4 spaces further, normalize formatting within file. I can do that.

persistence/persistence-mybatis/src/test/java/org/cbioportal/persistence/mybatis/SampleMyBatisRepositoryTest.java
Luke-Sikina: I added two additional tests to verify behavior when the list is empty and populated

web/src/main/java/org/cbioportal/web/SampleController.java
Luke-Sikina: Leaving this off the PR, but yes, agree.

web/src/main/java/org/cbioportal/web/SampleController.java
Luke-Sikina: For now I'll change it to reflect the status quo.

PR #6754, opened 2019-10-29
Gene info parser fix

SRodenburg:
I ran into some problems while building a hg38 seed database:

Sometimes the file `Homo_sapiens.gene_info` contains missing values in the `map_location` and/or `chromosome` columns. This file is downloaded from NCBI and required when updating genes and gene aliases.
Additionally, sometimes a particular symbol from the GenCode annotation file is not in the info file, causing a `NullPointerException`.

Changes:
- When the chromosome can't be parsed from the `map_location` column, try the `chromosome` column. If that also fails, skip gene.
- prevent `NullPointerException` when the gene symbol can't be found, and instead showing a warning, and skipping the gene.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
@@ -96,12 +96,22 @@ public class ImportGeneData extends ConsoleRunnable {
                     String strAliases = parts[4];
                     String strXrefs = parts[5];
                     String cytoband = parts[7];
-                    String chr = cytoband.split("p|q")[0];
+                    String chr = cytoband.split("p|q|;|c")[0]; // various strange characters were found in this column
                     int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeByBuildName(genomeBuild).getReferenceGenomeId();
                     String desc = parts[8];
                     String type = parts[9];
                     String mainSymbol = parts[10]; // use 10 instead of 2 since column 2 may have duplication
                     Set<String> aliases = new HashSet<String>();
+
+                    // try to get chr from other column if needed
+                    if (chr.equals("-")) {
+                        if (!parts[6].equals("-")) {
+                            chr = parts[6];
+                        } else {
+                            // skip line if still unable to parse chr
+                            continue;
+                        }
+                    }
 
                     if (!locusTag.equals("-")) {
                         aliases.add(locusTag);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -96,13 +96,23 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
...
-                    String chr = cytoband.split("p|q")[0];
+                    String chr = cytoband.split("p|q|;|c")[0]; // various strange characters were found in this column
                     int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeByBuildName(genomeBuild).getReferenceGenomeId();
                     String desc = parts[8];
                     String type = parts[9];
                     String mainSymbol = parts[10]; // use 10 instead of 2 since column 2 may have duplication
                     Set<String> aliases = new HashSet<String>();
 
+                    // try to get chr from other column if needed
+                    if (chr.equals("-")) {
```
khzhu: combine two ifs into one, so the second condition will not be evaluated if first one is false.
```
       if (chr.equals("-") && (!parts[6].equals("-")) {
              chr = parts[6];
       } else {
              // skip line if still unable to parse chr
             continue;
      }
```

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -504,6 +514,12 @@ else if (i.contains("gene_name")) {
...
         CanonicalGene gene = daoGeneOptimized.getNonAmbiguousGene(symbol);
+        
+        if (gene==null) {
+            ProgressMonitor.logWarning("Unable to retrieve gene by symbol: " +symbol);
+            return false;
+        }
+
         System.out.println(" --> update reference genome gene:  " + gene.getHugoGeneSymbolAllCaps());
         DaoReferenceGenomeGene daoReferenceGenomeGene = DaoReferenceGenomeGene.getInstance();
         boolean lengthUpdated = false;
```
khzhu: move line 525 before line 518, then line 520 just be
```
return lengthUpdated
```

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
SRodenburg: Done

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
SRodenburg: Although I fail to see the added value of it, I did implement this.
Done.

SRodenburg:
Thanks for reviewing @khzhu. Adjusted your proposed changes, ready to merge?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
SRodenburg: The unittests revealed a problem with this implementation; the code will go straight to `continue` when `chr` does not equal `"-"`, which is obviously not what we want.

A "synonymous" implementation would be:
```
if (chr.equals("-") && parts[6].equals("-")) {
    continue
} 
if (chr.equals("-")) {
    chr = parts[6];
}
```

However, this requires 2 logical tests while my original implemetation for most of the lines requires 1, thus would be better performance-wise.

SRodenburg:
@inodb can you merge?

Issue #6769, opened 2019-11-04
Fetching treatment_response data by using generic_assay endpoints


dippindots:
Fetching treatment_response data by using generic_assay endpoints, need to do the following steps:

- [x] (back-end) merge  `pivot_threshold_value` and `value_sort_order ` into `generic_assay`.
- [x] (back-end) improve `generic_assay` endpoint
- [x] (database) import `TREATMENT_RESPONSE` as `generic_assay` data
- [x] (front-end) change front-end to use `generic_assay` APIs
- [x] (back-end) remove the `TREATMENT_RESPONSE` APIs
- [x] (database) remove `TREATMENT_RESPONSE` data in database (e.g. treatment table - but try to move data in migration.sql instead of having to reimport it.  If this is tedious, we will instruct people to reimport treatment data after running migration script).

CC: @n1zea144 @pvannierop

PR #6789, opened 2019-11-07
Change generic assay endpoint

dippindots:
Bring back an endpoint in generic_assay (support sample list)

the second step in https://github.com/cBioPortal/cbioportal/issues/6769
Note: merge this one after #6776

### DIFF
--- a/service/src/main/java/org/cbioportal/service/GenericAssayService.java
+++ b/service/src/main/java/org/cbioportal/service/GenericAssayService.java
@@ -15,4 +15,12 @@ public interface GenericAssayService {
     List<GenericAssayData> getGenericAssayDataInMultipleMolecularProfiles(List<String> molecularProfileIds, 
                                                                         List<String> sampleIds, List<String> genericAssayStableIds, String projection)
         throws MolecularProfileNotFoundException;
+
+    List<GenericAssayData> getGenericAssayData(String molecularProfileId, String sampleListId, 
+                                            List<String> genericAssayStableIds, String projection) 
+        throws MolecularProfileNotFoundException;
+
+    List<GenericAssayData> fetchGenericAssayData(String molecularProfileId, List<String> sampleIds, 
+                                            List<String> genericAssayStableIds, String projection) 
+        throws MolecularProfileNotFoundException;  
 }
--- a/service/src/main/java/org/cbioportal/service/GenericAssayService.java
+++ b/service/src/main/java/org/cbioportal/service/GenericAssayService.java
@@ -12,15 +12,15 @@ public interface GenericAssayService {
     List<GenericAssayMeta> getGenericAssayMetaByStableIdsAndMolecularIds(List<String> stableIds, List<String> molecularProfileIds, String projection)
         throws GenericAssayNotFoundException;   
 
-    List<GenericAssayData> getGenericAssayDataInMultipleMolecularProfiles(List<String> molecularProfileIds, 
-                                                                        List<String> sampleIds, List<String> genericAssayStableIds, String projection)
-        throws MolecularProfileNotFoundException;
-
     List<GenericAssayData> getGenericAssayData(String molecularProfileId, String sampleListId, 
                                             List<String> genericAssayStableIds, String projection) 
         throws MolecularProfileNotFoundException;
 
     List<GenericAssayData> fetchGenericAssayData(String molecularProfileId, List<String> sampleIds, 
                                             List<String> genericAssayStableIds, String projection) 
-        throws MolecularProfileNotFoundException;  
+        throws MolecularProfileNotFoundException;
+
+    List<GenericAssayData> fetchGenericAssayData(List<String> molecularProfileIds, List<String> sampleIds, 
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/cbioportal/service/GenericAssayService.java
```diff
@@ -15,4 +15,12 @@
     List<GenericAssayData> getGenericAssayDataInMultipleMolecularProfiles(List<String> molecularProfileIds, 
                                                                         List<String> sampleIds, List<String> genericAssayStableIds, String projection)
         throws MolecularProfileNotFoundException;
+
+    List<GenericAssayData> getGenericAssayData(String molecularProfileId, String sampleListId, 
+                                            List<String> genericAssayStableIds, String projection) 
+        throws MolecularProfileNotFoundException;
```
pvannierop: Is it not nicer to create an overloaded method `getGenericAssayData` that takes either a single or multitple molecularProfileId(s)?

```
List<GenericAssayData> getGenericAssayData(String molecularProfileId, ...
List<GenericAssayData> getGenericAssayData(List<String> molecularProfileIds, 
```

service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -155,4 +158,77 @@
...
+                                                    List<String> genericAssayStableIds, String projection)
+        throws MolecularProfileNotFoundException {
+        
+        validateMolecularProfile(molecularProfileId);
+        List<String> sampleIds = sampleListRepository.getAllSampleIdsInSampleList(sampleListId);
+        if (sampleIds.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return fetchGenericAssayData(molecularProfileId, sampleIds, genericAssayStableIds, projection);
+    }
```
pvannierop: I smell code duplication here but I am not able to spot this. In essence I expect that a call to `getGenericAssayData` would call method `getGenericAssayDataInMultipleMolecularProfiles` like so:
```
public List<GenericAssayData> getGenericAssayData(String molecularProfileId, String sampleListId,
                                                    List<String> genericAssayStableIds, String projection)
        throws MolecularProfileNotFoundException {
       String[] molecularProfileIds = { molecularProfileId };
       return getGenericAssayDataInMultipleMolecularProfiles(molecularProfileIds, ....);
    }
```

Either that, or I would expect method `getGenericAssayDataInMultipleMolecularProfiles` to make use of the `fetchGenericAssayData`.

I have not dived into the code to deep, but glancing this code I would like to discuss whether this code may be optimized.

dippindots: Thanks for your comments! Just updated the solution, could you take a look again?

service/src/main/java/org/cbioportal/service/GenericAssayService.java
n1zea144: Yeah, I'm not sure what the difference in method names (get vs. fetch) serves here.

service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -155,4 +158,33 @@
...
+
+    @Override
+    public List<GenericAssayData> fetchGenericAssayData(String molecularProfileId, List<String> sampleIds,
+            List<String> genericAssayStableIds, String projection) throws MolecularProfileNotFoundException {
+        return fetchGenericAssayData(Arrays.asList(molecularProfileId), sampleIds, genericAssayStableIds, projection);
```
n1zea144: I think you should validate the molecular profile here.

service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -155,4 +158,33 @@
...
+    public List<GenericAssayData> getGenericAssayData(String molecularProfileId, String sampleListId,
+                                                    List<String> genericAssayStableIds, String projection)
+        throws MolecularProfileNotFoundException {
+        
+        validateMolecularProfile(molecularProfileId);
+        List<String> sampleIds = sampleListRepository.getAllSampleIdsInSampleList(sampleListId);
+        if (sampleIds.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return fetchGenericAssayData(molecularProfileId, sampleIds, genericAssayStableIds, projection);
```
n1zea144: I would do as @pvannierop suggests - convert molecularProfileId to list here and call fetchGenericAssayData directly (not getGenericAssayDataInMultipleMolecularProfiles).

service/src/main/java/org/cbioportal/service/GenericAssayService.java
dippindots: @n1zea144 I just referenced the [MolecularDataService](https://github.com/cBioPortal/cbioportal/blob/master/service/src/main/java/org/cbioportal/service/MolecularDataService.java). Do you want me to change to other names?

PR #6795, opened 2019-11-08
Allow values other than 'Fusion' for 'Event_Info' in SV data

ao508:
Structural Variant validator only accepts value "Fusion" for "Event_Info" field. This PR removes this restriction. Records with "Event_Info" values other than "Fusion" will skip the step to validate the fusion event information like exons and ensembl transcript ids.

Signed-off-by: Angelica Ochoa 15623749+ao508@users.noreply.github.com

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -2971,13 +2971,9 @@ class StructuralVariantValidator(Validator):
         site2_entrez_gene_id = checkPresenceValue('Site2_Entrez_Gene_Id', self, data)
         self.checkGeneIdentification(site2_hugo_symbol, site2_entrez_gene_id)
 
-        # Validate fusion events
+        # Validate fusion events if Event_Info is 'Fusion'
         if data[self.cols.index('Event_Info')] == 'Fusion':
             checkFusionValues(self, data)
-        else:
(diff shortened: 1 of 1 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -3086,6 +3079,9 @@ def validateExonsInTranscript(self, transcript, exon, line_number):
             # For every fusion event, check if exon is part of the transcript
             validateExons(self)
 
+            # List all transcripts
```
averyniceday: was this moved because there are no transcripts if self.transcirpt_set is <= 0? Haven't looked at the function in detail or where self.transcript_set is set.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2971,13 +2971,9 @@ def checkExonPresence(self, exon, column_name):
         site2_entrez_gene_id = checkPresenceValue('Site2_Entrez_Gene_Id', self, data)
         self.checkGeneIdentification(site2_hugo_symbol, site2_entrez_gene_id)
 
-        # Validate fusion events
```
averyniceday: looks good

core/src/main/scripts/importer/validateData.py
ao508: @averyniceday correct

PR #6798, opened 2019-11-11
Persistent Cache Timestamp API

### LINKED ISSUE #6635: reference genome genes endpoint is a bottleneck
This endpoint shows up in monitoring as a performance drain on db. 

https://www.cbioportal.org/api/reference-genome-genes/hg19

It returns all ~143,000 genes in public portal.   Is it necessary to return all of them or could we instead return just genes which are queried?

@jjgao @inodb

Luke-Sikina:
Fix #6635

- New endpoint
    - For every table that has endpoints that are being persistently
    cached the endpoint returns a table name timestamp pair of the
    last time the table was updated
    - If the table has never been updated, it instead returns
    the creation date of the table
- This feature requires SQL 5.7.2+ because the InnoDb engine did not
update this column correctly before then
    - updated the requirements in the docs to require 5.7 rather than 5.x

# Checks
- [ ] Runs on heroku

### DIFF
--- a/docs/Deploy-Using-Docker.md
+++ b/docs/Deploy-Using-Docker.md
@@ -29,7 +29,9 @@ docker network create "cbio-net"
 
 ## 2. Database Setup
 
-As of this writing, the cBioPortal software runs properly on MySQL version 5.0.x. The software can be found and downloaded from the [MySQL website](https://www.mysql.com).
+The cBioPortal software should run properly on MySQL version 5.7.x. Versions higher than 5.7.x can cause an issue
+while loading the database schema. Minor versions lower than 5.7.x will cause issues with persistent cache invalidation.
+The software can be found and downloaded from the [MySQL website](http://www.mysql.com/).
 
 There are two options to set up the cBioPortal Database:    
 2.1 Run MySQL on the host.    
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/StaticDataTimestampMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/StaticDataTimestampMapper.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
+<mapper namespace="org.cbioportal.persistence.mybatis.StaticDataTimestampMapper">
+    <select id="getTimestamps" resultType="org.cbioportal.model.TableTimestampPair">
+        <if test="dbName != null">
+            SELECT
+                TABLE_NAME AS tableName,
+                COALESCE(UPDATE_TIME, CREATE_TIME) AS updateTime
+            FROM information_schema.tables
+            WHERE
+                TABLE_SCHEMA=#{dbName}
+                <if test="tables != null and !tables.isEmpty()">
+                    AND TABLE_NAME IN
+                    <foreach item="item" collection="tables" open="(" separator="," close=");">
+                        #{item}
+                    </foreach>
+                </if>    
+        </if>
+        
+    </select>
+</mapper>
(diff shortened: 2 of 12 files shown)


persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/StaticDataTimestampMapper.xml
```diff
@@ -0,0 +1,19 @@
...
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
+<mapper namespace="org.cbioportal.persistence.mybatis.StaticDataTimestampMapper">
+    <select id="getTimestamps" resultType="org.cbioportal.model.TableTimestampPair">
+        SELECT
+            TABLE_NAME AS tableName,
+            COALESCE(UPDATE_TIME, CREATE_TIME) AS updateTime
+        FROM information_schema.tables
+        WHERE
+            TABLE_SCHEMA='cbioportal'
```
inodb: @Luke-Sikina i believe the schema should be `db.portal_db_name`

service/src/main/java/org/cbioportal/service/impl/StaticDataTimestampServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.List;
+
+@Service
+public class StaticDataTimestampServiceImpl implements StaticDataTimestampService {
+    @Autowired
+    private StaticDataTimeStampRepository staticDataTimeStampRepository;
+
+    @Override
+    public List<TableTimestampPair> getTimestamps(List<String> tables) {
```
inodb: @Luke-Sikina should we limit this to only gene table and perhaps some others? Not sure if there's anything against being able to get last update time for any table. Can't think of anything atm so maybe it's fine

Luke-Sikina: I thought about it and couldn't come up with anything either. For what it's worth, that's part of why I stuck with a hard coded list of tables in the controller. I thought that letting a user request the update date of _any_ table could be problematic.

docs/Deploy-Using-Docker.md
```diff
@@ -29,7 +29,9 @@ docker network create "cbio-net"
 
 ## 2. Database Setup
 
-As of this writing, the cBioPortal software runs properly on MySQL version 5.0.x. The software can be found and downloaded from the [MySQL website](https://www.mysql.com).
+The cBioPortal software should run properly on MySQL version 5.7.x. Versions higher than 5.7.x can cause an issue
```
Luke-Sikina: From https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-2.html
> InnoDB: Beginning with MySQL 5.7.2, UPDATE_TIME displays a timestamp value for the last UPDATE, INSERT, or DELETE performed on InnoDB tables. Previously, UPDATE_TIME displayed a NULL value for InnoDB tables. For MVCC, the timestamp value reflects the COMMIT time, which is considered the last update time. Timestamps are not persisted when the server is restarted or when the table is evicted from the InnoDB data dictionary cache.

web/src/main/java/org/cbioportal/web/StaticDataTimestampController.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Collections;
+import java.util.List;
+
+@InternalApi
+@RestController
+@Validated
+@Api(tags = "Timestamps", description = " ")
+public class StaticDataTimestampController {
+    private static final List<String> tables = Collections.singletonList("gene");
```
n1zea144: Whats cached on the frontend?  Should you add reference_genome_gene table here?  That contains chr, cytoband, and some other props.

Luke-Sikina: Oh, you're right. Looking at the mapper, the fields come entirely from `reference_genome_gene`, but there is an inner join with `gene` on `entrez_gene_id`. I think the answer is that the cache should be invalidated if either change, so I'll include both tables. 

The frontend caches the full superagent response object that it makes when it calls `/api/reference-genome-genes/hg19`

PR #6802, opened 2019-11-11
Attach auth header when using oncokb service

zhx828:
fixes: #6950 
What the workflow is trying to do:
- Add an additional property to specify the oncokb token when using www.oncokb.org
- If the token is not specified, use a legacy instance(legacy.oncokb.org) and give a warning when building war file, and when deploy to the server(this is particularly important for docker instance)

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -39,6 +39,7 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Component;
 import org.springframework.util.ResourceUtils;
+import org.springframework.util.StringUtils;
 
 import java.io.*;
 import java.nio.file.Files;
@@ -142,7 +143,6 @@ public class GlobalProperties {
     public static final String STUDY_VIEW_MDACC_HEATMAP_URL = "mdacc.heatmap.study.url";
     public static final String STUDY_VIEW_MDACC_HEATMAP_META_URL = "mdacc.heatmap.study.meta.url";
 
-    public static final String ONCOKB_PUBLIC_API_URL = "oncokb.public_api.url";
     public static final String SHOW_ONCOKB = "show.oncokb";
 
     private static String sessionServiceURL;
@@ -856,28 +856,6 @@ public class GlobalProperties {
     public static String getSessionServicePassword()
     {
         return sessionServicePassword;
-    }
-
-    public static String getOncoKBPublicApiUrl()
-    {
-        String oncokbApiUrl = portalProperties.getProperty(ONCOKB_PUBLIC_API_URL);
-        String showOncokb = portalProperties.getProperty(SHOW_ONCOKB);
-        
-        if(showOncokb == null || showOncokb.isEmpty()) {
-                    showOncokb = "true";
-        }
-        
-        // Empty string should be used if you want to disable the OncoKB annotation.
-        if(oncokbApiUrl == null || oncokbApiUrl.isEmpty()) {
-            oncokbApiUrl = "oncokb.org/api/v1";
-        }
-        
-        if(showOncokb.equals("true")) {
-           return oncokbApiUrl;
-        } else {
-           return "";
-        }
-        
     }
 
     public static String getCivicUrl() {
--- a/portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
@@ -9,7 +9,6 @@ window.legacySupportFrontendConfig = {
     appVersion : '<%=GlobalProperties.getAppVersion()%>',
     maxTreeDepth : <%=GlobalProperties.getMaxTreeDepth()%>,
     showOncoKB : <%=GlobalProperties.showOncoKB()%>,
-    oncoKBApiUrl : '<%=GlobalProperties.getOncoKBPublicApiUrl()%>',
     genomeNexusApiUrl : '<%=GlobalProperties.getGenomeNexusApiUrl()%>',
     showCivic : <%=GlobalProperties.showCivic()%>,
     showHotspot : <%=GlobalProperties.showHotspot()%>,
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -159,6 +159,10 @@ segfile.url=http://cbio.mskcc.org/cancergenomics/gdac-portal/seg/
 # Enable OncoKB annotation (true, false)
 show.oncokb=true
 
+# OncoKB API Token, for more information please check https://www.oncokb.org/dataAccess
+# The token is required if you want to use OncoKB service
+oncokb.token=
+
 # Enable Chang's hotspot list (true, false)
 show.hotspot=true
 hotspots.url=https://www.cancerhotspots.org/api/
--- a/web/src/main/java/org/cbioportal/proxy/ProxyController.java
+++ b/web/src/main/java/org/cbioportal/proxy/ProxyController.java
@@ -1,10 +1,11 @@
 package org.cbioportal.proxy;
 
-import org.springframework.http.HttpEntity;
-import org.springframework.http.HttpHeaders;
-import org.springframework.http.HttpMethod;
-import org.springframework.http.MediaType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.*;
 import org.springframework.http.converter.StringHttpMessageConverter;
+import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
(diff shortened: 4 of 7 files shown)


web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -1,38 +1,74 @@
...
 
         HttpHeaders httpHeaders = new HttpHeaders();
         String contentType = request.getHeader("Content-Type");
         if (contentType != null) {
             httpHeaders.setContentType(MediaType.valueOf(contentType));
         }
+        if (!StringUtils.isEmpty(this.oncokbToken)) {
+            httpHeaders.add("Authorization", "Bearer " + this.oncokbToken);
+        }
+//        else if(StringUtils.isEmpty(this.oncokbToken) && this.oncokbPublicApiUrl.equalsIgnoreCase("legacy.oncokb.org")){
```
zhx828: @sheridancbio I commented out this section since I don't want to log a warning everytime the proxy is used. What's the best way to print a warning message after instance is deployed?

web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -1,38 +1,74 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.*;
 import org.springframework.http.converter.StringHttpMessageConverter;
+import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.client.RestTemplate;
 
+import javax.annotation.PostConstruct;
```
zhx828: This is not used. I will clean it up.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -157,6 +157,10 @@ segfile.url=http://cbio.mskcc.org/cancergenomics/gdac-portal/seg/
 # Enable OncoKB annotation (true, false)
 show.oncokb=true
 
+# OncoKB API Token, for more information please check https://beta.oncokb.org/dataAccess
```
zhx828: This needs to be updated to oncokb.org

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -869,7 +869,7 @@ public static String getOncoKBPublicApiUrl()
         
         // Empty string should be used if you want to disable the OncoKB annotation.
         if(oncokbApiUrl == null || oncokbApiUrl.isEmpty()) {
-            oncokbApiUrl = "oncokb.org/api/v1";
+            oncokbApiUrl = "www.oncokb.org/api/v1";
```
sheridancbio: The comment above the if statement is incorrect ... you can never disable OncoKB annotation because of this code. I would add a symbolic static final string at the top of the class (DEFAULT_ONCO_KB_API_URL)

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -157,6 +157,10 @@ segfile.url=http://cbio.mskcc.org/cancergenomics/gdac-portal/seg/
...
+# The token is required if you want to use OncoKB service
+oncokb.token=
+
```
sheridancbio: shouldn't the oncokb public URL property oncokb.public_api.url be here too? (and I would add a comment that if not provided, it default to the value from global properties)

web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -1,38 +1,74 @@
...
+
+    private Logger LOG = LoggerFactory.getLogger(ProxyController.class);
 
     @RequestMapping("/**")
     public String proxy(@RequestBody(required = false) String body, HttpMethod method, HttpServletRequest request)
         throws URISyntaxException {
+        HttpHeaders httpHeaders = new HttpHeaders();
+        String contentType = request.getHeader("Content-Type");
+        if (contentType != null) {
+            httpHeaders.setContentType(MediaType.valueOf(contentType));
```
sheridancbio: It took some thinking for me to understand this --- the content type of the proxied request is being set to the same content type as the received request.

web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -1,38 +1,74 @@
...
+        }
+        return exchangeData(body, buildUri(request.getPathInfo(), request.getQueryString(), false), method, httpHeaders).getBody();
+    }
 
-        String queryString = request.getQueryString();
-        URI uri = new URI("http:/" + request.getPathInfo() + (queryString == null ? "" : "?" + queryString));
+    @RequestMapping("/oncokb/**")
```
sheridancbio: Doesn't this provide a route for anybody to make a request for data from oncokb ?  Basically this proxy service allows anybody to make requests to oncokb using the cbioportal server's token. If there is a path like that, doesn't it undermine the effort to add access control to oncokb?

web/src/main/java/org/cbioportal/proxy/ProxyController.java
sheridancbio: Code can be added that will run at application startup using the interface "ServletContextListener"
https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletContextListener.html
https://stackoverflow.com/questions/4366074/java-web-application-initialization-and-shutdown
you need to add a reference to the implementing class in web.xml

web/src/main/resources/applicationContext-proxy.xml
```diff
@@ -9,6 +9,21 @@
      http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
      http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc.xsd">
 
+    <bean id="propertyPlaceholderConfigurer" class="org.springframework.beans.factory.config.PropertyPlaceholderConfigurer">
```
sheridancbio: I wasn't familiar with this file before.

src/main/resources/portal.properties.EXAMPLE
zhx828: @sheridancbio it's never our intention to let other instances to specify oncokb link. This is just a property for our usage, so it's not documented

web/src/main/java/org/cbioportal/proxy/ProxyController.java
zhx828: It does and it's a loophole. We don't have a better solution at the moment to address the issue. Portal allows cross-domain access, so there is no way to prevent it. We are planning to make some changes to only return the data portal needs, but the way frontend uses oncokb is totally depending on the clients generated through swagger. Which means, if we want to fix the issue, that involves too many changes. In a limited time, we are ok to let the cbioportal has a proxy like this.

web/src/main/resources/applicationContext-proxy.xml
zhx828: I borrowed the ideas from other application context. From what I understand, the web does not rely on the core module so the properties will not be injected unless you specify the property files in the web xml

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -858,28 +858,6 @@ public static String getSessionServicePassword()
         return sessionServicePassword;
     }
 
-    public static String getOncoKBPublicApiUrl()
```
zhx828: We no longer provide oncokb api url to frontend. The website will disable the cross domain.

web/src/main/java/org/cbioportal/web/config/WebServletContextListener.java
```diff
@@ -0,0 +1,113 @@
...
+            resourceInputStream.close();
+        }
+        catch (IOException e) {
+            if (LOG.isErrorEnabled()) {
+                LOG.error("Error loading properties file: " + e.getMessage());
+            }
+        }
+
+        return properties;
+    }
```
zhx828: I copied these code from `GlobalProperties`. Since we don't include core as the dependency of the web. Not sure how to what the convention is to share the code.

web/src/main/java/org/cbioportal/web/config/WebServletContextListener.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void afterPropertiesSet() throws Exception {
+    }
+
+    private void checkOncokbInfo() {
+        System.out.println("------------------------");
+        if(StringUtils.isEmpty(this.oncokbToken)) {
+            // oncokb.org is deprecated, www.oncokb.org should be used
+            System.out.println("OncoKB requires token to access the latest data. Please review oncokb terms(https://www.oncokb.org/terms) and register an account accordingly. For now, we will redirect the requests to a legacy instance which will not accept any new data changes. Thank you.");
+            LOG.warn("OncoKB requires token to access the latest data. Please review oncokb terms(https://www.oncokb.org/terms) and register an account accordingly. For now, we will redirect the requests to a legacy instance which will not accept any new data changes. Thank you.");
```
zhx828: No sure what I did wrong, but the LOG.warn does not print out the message. Only the `System.out.println` works.

src/main/resources/portal.properties.EXAMPLE
inodb: @zhx828 so people can't deploy their own instance of OncoKB anymore? This might be a limitation for Dana Farber. They have some policy to not allow external API calls with mutation data

web/src/main/java/org/cbioportal/web/config/WebServletContextListener.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Override
+    public void contextDestroyed(ServletContextEvent arg0) {
+        System.out.println("ServletContextListener destroyed");
+    }
+
+    //Run this before web application is started
+    @Override
+    public void contextInitialized(ServletContextEvent arg0) {
+        Properties properties = loadProperties(getResourceStream("portal.properties"));
```
inodb: @zhx828 i think this should be done in another way, because it doesn't support setting these properties on the command line with `-D`

inodb: can use something like this from `GlobalProperties.java`: https://github.com/cBioPortal/cbioportal/blob/d786e77a59a880fc331e608276586a0b5f0b4c78/core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java#L350-L353

web/src/main/java/org/cbioportal/web/config/WebServletContextListener.java
inodb: might need to change log4j properties to print WARN to stdout. With default props I think it outputs to the log file. We might want to change the default `log4j.properties`. This is an example for my debug config:

```
# Change INFO to DEBUG, if you want to see debugging info on underlying libraries we use.
log4j.rootLogger=INFO, a

# Change INFO to DEBUG, if you want see debugging info on our packages and spring security packages.
log4j.category.org.mskcc=INFO
log4j.logger.org.springframework.security=INFO

# Use the JVM option, e.g.: "java -DPORTAL_HOME=/pathto/portal_homedir",
# or - "java -DPORTAL_HOME=$PORTAL_HOME", where PORTAL_HOME is shell (environment) variable.

## IMPORTANT - THRESHOLD SHOULD NOT BE DEBUG FOR PRODUCTION, CREDENTIALS CAN BE DISPLAYED!
log4j.appender.a=org.apache.log4j.ConsoleAppender
log4j.appender.a.Target=System.out
log4j.appender.a.layout=org.apache.log4j.PatternLayout

log4j.logger.org.cbioportal.persistence.mybatis.StudyMapper=TRACE
log4j.logger.org.cbioportal.persistence.mybatis=DEBUG
src/main/resources/log4j.properties.DEBUG (END)
```

src/main/resources/portal.properties.EXAMPLE
zhx828: @inodb technically, they still can. I will just write a document on how to do that.

web/src/main/java/org/cbioportal/web/config/WebServletContextListener.java
zhx828: yeah... I thought I put it in, will update.

web/src/main/java/org/cbioportal/web/config/WebServletContextListener.java
zhx828: @inodb this prints out the warning not log.debug. So technically in the default which prints INFO would also include the WARN.

web/src/main/java/org/cbioportal/proxy/ProxyController.java
alisman: @sheridancbio the only way to address this issue in long run is to do rate limiting or invisible catcha on the api (or proxy) so that people can't abuse it.  there is no other solution

portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
```diff
@@ -9,7 +9,6 @@ window.legacySupportFrontendConfig = {
     appVersion : '<%=GlobalProperties.getAppVersion()%>',
     maxTreeDepth : <%=GlobalProperties.getMaxTreeDepth()%>,
     showOncoKB : <%=GlobalProperties.showOncoKB()%>,
-    oncoKBApiUrl : '<%=GlobalProperties.getOncoKBPublicApiUrl()%>',
```
alisman: this is cleanup. this property does nothing

onursumer:
Updates: 

- Rebased the PR onto the latest master
- No longer supporting the legacy API if `oncokb.token` is not provided
- Changes in this PR won't affect any build. Until the frontend code is also modified, builds will continue to use the legacy OncoKB API.

PR #6805, opened 2019-11-12
Treatment database migration

dippindots:
The last step of https://github.com/cBioPortal/cbioportal/issues/6769

migrate the existing treatment data to generic_assay, and delete the treatment table.


db-scripts/src/main/resources/migration.sql
```diff
@@ -762,3 +762,34 @@ CREATE TABLE `generic_entity_properties` (
...
+-- insert URL into generic_entity_properties
+INSERT INTO generic_entity_properties (GENETIC_ENTITY_ID, NAME, VALUE)
+SELECT
+    GENETIC_ENTITY_ID,
+    "URL",
+    LINKOUT_URL
+FROM `treatment`;
+-- update genetic_entity and genetic_profile
+UPDATE genetic_entity INNER JOIN treatment ON genetic_entity.ID = treatment.GENETIC_ENTITY_ID SET genetic_entity.STABLE_ID = treatment.STABLE_ID, genetic_entity.ENTITY_TYPE = "GENERIC_ASSAY";
+UPDATE genetic_profile SET GENERIC_ASSAY_TYPE = "TREATMENT_RESPONSE" WHERE genetic_profile.GENETIC_ALTERATION_TYPE = "GENERIC_ASSAY";
```
n1zea144: This works because we know at this point in time there is no GENERIC_ASSAY data other than TREATMENT_RESPONSE

dippindots: Yes, that's true! Should we update the db_version from `2.12.1` to `2.12.2`? Maybe we shouldn't?

PR #6817, opened 2019-11-14
Fixed validation issue with different geneset_def_version

dianab0:
The current validation process of cBioPortal does not throw an error when the property geneset_def_version is different in the study from the one in the database. This leads to an error throwing when cBioPortal tries to load the study.
Fixed it by first adding the geneset version from database to the API and also to PortalInfo, then retrieving it in the validateData.py script and comparing it to geneset_def_version in the study.

- Updated the API to include the geneset version at /api/genesets/version
- Updated the DumpPortalInfo script to get the geneset from API
- Updated the validateData.py to get the geneset version from either portal info or API, compare it to the geneset_def_version in the study and return error if different.

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


dianab0:
Pushed a second commit to include the necessary changes to the test_data folder so the test scripts will work.

PR #6836, opened 2019-11-18
Update docs/docker/README.md

Spenca:


inodb:
Thanks so much @Spenca !

PR #6839, opened 2019-11-20
On the road to end2end localdb tests

### LINKED ISSUE #3850: Run frontend localdb end2end tests in backend repo
Run the e2e localdb test in backend repo: https://github.com/cBioPortal/cbioportal-frontend/blob/master/.circleci/config.yml#L178. Currently in the frontend repo when you open a PR to the master branch, that frontend code is tested against a backend docker container running the latest master code. In the backend we would like the same thing but then testing the backend changes, that is: 

1. build a docker image for this backend (or get a war in some other way)
2. find the frontend commit in `pom.xml` (https://github.com/cBioPortal/cbioportal/blob/master/pom.xml#L273)
3. check that out and run the localdb e2e tests against the image from 1. No need to spin up a frontend separately, since that's bundled in the war from 1. 

A previous attempt was made here: https://github.com/cBioPortal/cbioportal/pull/4124/files#diff-1d37e48f9ceff6d8030570cd36286a61R20. This include downloading the jar from jitpack which is no longer necessary, since we have a Docker container now for the backend. The previous attempt stalled, since it proved tricky to get the tests to pass. The remote e2e tests expected quite a heavy machine, with the CircleCI backend machine the same set of e2e tests would result in failures. This should no longer be a problem with the localdb tests, since in the frontend all the containers are running on CirlceCI infrastructure. 

More comprehensive documentation about running the localdb tests e2e can be found here: https://github.com/cBioPortal/cbioportal-frontend#run-of-local-e2e-tests

n1zea144:
Fix #3850

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,170 @@
- version: 2
- jobs:
-   build:
-     docker:
-       - image: circleci/ruby:2.4.1
-     steps:
-       - checkout
-       - run: echo "A first hello"
+defaults: &defaults
+    docker:
+        - image: circleci/node:8.12.0-browsers
+
+version: 2
+jobs:
+    pull_frontend_codebase:
+        <<: *defaults
+        working_directory: /tmp/repos/cbioportal
+        steps:
+            - checkout
+            - run:
+                name: Pull frontend code
+                command: |
+                    export FRONTEND_VERSION=$(grep 'frontend\.version' pom.xml | sed 's/<frontend\.version>//g' | sed 's|</frontend\.version>||' | tr -d '[:blank:]') && \
+                    cd .. && \
+                    git clone https://github.com/$CIRCLE_PROJECT_USERNAME/cbioportal-frontend.git && \
+                    cd cbioportal-frontend && \
+                    git fetch --tags && \
+                    git checkout ${FRONTEND_VERSION}
+            - persist_to_workspace:
+                root: /tmp/repos
+                paths:
+                    - cbioportal-frontend
+
+    install_yarn:
+        <<: *defaults
+        working_directory: /tmp/repos/cbioportal-frontend
+        steps:
+            - attach_workspace:
+                at: /tmp/repos
+            - run:
+                name: "Install yarn at specific version"
+                command:
+                    sudo npm install --global yarn@1.13.0
+            - run:
+                name: "Show yarn and node versions"
+                command: |
+                    node --version
+                    yarn --version
+            # cache build artifacts. Use concatenation of all source files as cache
+            # key. If there are no changes to src/ and yarn.lock, no need to rebuild
+            - run:
+                name: "Concatenate all source files to use as hash key for caching dist folder"
+                command: "cat yarn.lock $(find src/ -type f | sort) webpack.config.js vendor-bundles.webpack.config.js > has_source_changed"
+            - restore_cache:
+                keys:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,8 +1,171 @@
...
+                    git clone https://github.com/$CIRCLE_PROJECT_USERNAME/cbioportal-frontend.git && \
+                    cd cbioportal-frontend && \
+                    git fetch --tags && \
+                    git checkout ${FRONTEND_VERSION}
+            - persist_to_workspace:
+                root: /tmp/repos
+                paths:
+                    - cbioportal-frontend
+
+    install_yarn:
```
inodb: Once the e2e tests are using the backend of this PR the `install_yarn` job prolly won't be necessary anymore because the frontend artifacts are already in the war. No need to recompile them

n1zea144: Hi @inodb Unless I'm misunderstanding something, it is already using the  [backend](https://github.com/cBioPortal/cbioportal/blob/54d2812bc57e99410981564c60ea3d51d6de05c1/.circleci/config.yml#L95).

inodb: Good point - so it looks like it's pretty much working then 🎉. My point was mostly regarding some minor optimization. I think currently the localdb test tries to mount the frontend inside the war. That's why you need these steps of yarn run build etc. On the frontend e2e tests that makes sense cause you need to test against that frontend code in the PR. On the backend it's redundant b/c the war already contains the compiled frontend. 

Another minor thing is that in the localdb tests the backend code gets pulled from GitHub again and the docker image is made. An optimization would be to create the image before running these tests and using it subsequently in the localdb tests ( to avoid the extra pull). You can then extend it with things like: only rebuild the image when the code has changed.

But yeah we can skip these optimizations for now I think, better to have something working in master branch.

Maybe just double check the backend of the PR is actually being used e.g. by introducing an error or something. And check it is properly creating the new backend image when you add a new commit in the pr to fix it.

n1zea144: It is true that the only use of the backend checkout right now is to get the version of the frontend to checkout.  I explored the idea of creating an image from the checkout step done by the pull_frontend_codebase job.  In one scenario, it may just involve the modification of [build_portal_image.sh](https://github.com/cBioPortal/cbioportal-frontend/blob/master/end-to-end-test/local/docker/build_portal_image.sh), maybe with some additional args passed to [setup_docker_containers.sh](https://github.com/cBioPortal/cbioportal-frontend/blob/master/end-to-end-test/local/docker/setup_docker_containers.sh) (called by the circle-ci config).  @pvannierop any thoughts on this?

I did see from the circle-ci page that my hash was being used during the Setup docker images and containers step.  I went ahead and introduced an error as you suggested to confirm the error and fix on revert.  This looks good.

n1zea144: @inodb I have a further thought on this - maybe I could add a check to see what files have changed after checkout and then short circuit the rest of the script when code files haven't changed.

inodb:
Small note: we should prolly update this once this one is fixed: https://github.com/cBioPortal/cbioportal/issues/6844

.circleci/config.yml
n1zea144: @inodb based on what I've discovered and communicated to you during our conversation this morning, I'm going to drop the effort to optimize the current configuration.   I'll merge this PR once you give the approval.  cc @pvannierop

PR #6818, opened 2019-11-14
Remove treatment implementation

dippindots:
The fifth step of https://github.com/cBioPortal/cbioportal/issues/6769.

- Removed treatment controller, services, tests, importer and related files.
- Keep the data file, transfer the data file to the generic_assay file.
- Remove treatment validator

PR #6841, opened 2019-11-20
Add treatment validation

### LINKED ISSUE #6826: Treatment data validation
The validation of `treatment` data has been removed when migrating to `generic_assay` data. Maybe we should bring back the validation just for the file which `generic_assay_type` is `TREATMENT_RESPONSE`.

CC: @pvannierop @n1zea144 @inodb 

validation should be added:
1. multiple files validations (all files should include the same columns and same entities)
https://github.com/cBioPortal/cbioportal/pull/6818#discussion_r346649928
2. validation tests
https://github.com/cBioPortal/cbioportal/pull/6818#discussion_r346649820

dippindots:
Fix #6826 

- a add one endpoint (`GET /generic_assay_meta/{genericAssayType}`) to support because the validation relies on the API support
- b bring all treatments validation back


web/src/main/java/org/cbioportal/web/GenericAssayController.java
```diff
@@ -48,6 +51,25 @@
...
+        @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_SIZE) Integer pageSize,	
+        @ApiParam("Page number of the result list")	
+        @Min(PagingConstants.MIN_PAGE_NUMBER)	
+        @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_NUMBER) Integer pageNumber) throws GenericAssayNotFoundException {	
+
+        return new ResponseEntity<>(	
+            genericAssayService.getGenericAssayMetaByGenericAssayType(genericAssayType, projection.name(), pageSize, pageNumber), HttpStatus.OK);
+    }
+
     @PreAuthorize("hasPermission(#molecularProfileId, 'MolecularProfileId', 'read')")
```
n1zea144: I noticed that the implementation of this routine is not quite right.  You'll need to introduce a interceptedGenericAssayMetaFilter.  Just a note - PreAuthorizing this method seems pointless in light of the fact that we are introducing /generic_assay_meta/{genericAssayType} without any preauthorization.  As stated above,  I guess the idea is that this information is just like gene information - which is freely available.  However, the fact that the data is requested via a molecular profile id is reason enough to keep preauthorization here.

web/src/main/java/org/cbioportal/web/GenericAssayController.java
```diff
@@ -48,6 +51,25 @@
     @Autowired
     private GenericAssayService genericAssayService;
 
+    @RequestMapping(value = "/generic_assay_meta/{genericAssayType}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
n1zea144: @dippindots @pvannierop What is the point of this endpoint?

Having said that, I suppose we don't need permissions on this because this data can be though of as gene information - freely available.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/GenericAssayMapper.xml
```diff
@@ -9,6 +9,28 @@
                 <result property="value" column="VALUE"/> 
     </resultMap>
 
+    <select id="getGenericAssayMetaByGenericAssayType" resultType="org.cbioportal.model.meta.GenericAssayMeta">	
+        SELECT DISTINCT genetic_entity.STABLE_ID as stableId,
+        genetic_entity.ENTITY_TYPE as entityType
+        FROM genetic_entity
+        <if test="genericAssayType != null">	
+                LEFT JOIN	
+            genetic_alteration AS ga ON ga.GENETIC_ENTITY_ID = genetic_entity.ID
```
n1zea144: I'm not following this - why join genetic_alteration (has data) or genetic_profile below.  Don't you need to join generic_entity_properties table instead?

dippindots:
Hi @pvannierop, the base is based on the previous pr you have already reviewed: https://github.com/cBioPortal/cbioportal/pull/6818. This is the only new commit to bring the validator back, you can just review this commit at here: https://github.com/cBioPortal/cbioportal/pull/6841/commits/cd57f5967e1b8c6280a65053a37a23f2ed3e38cb. Thanks!

web/src/main/java/org/cbioportal/web/GenericAssayController.java
dippindots: @n1zea144 It's because the treatment validation will need to use this endpoint to get all the `treatment` entities from the database to validate the data (like `geneset` or `gene-panels`).https://github.com/cBioPortal/cbioportal/commit/cd57f5967e1b8c6280a65053a37a23f2ed3e38cb#diff-fee5ab3c72d7aa124ff0d2d3b199c68bR4614-R4615

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenericAssayEntity.java
```diff
@@ -162,103 +160,54 @@ public static void importData(File dataFile, GeneticAlterationType geneticAltera
...
-                    for (String columnName : columnNameList) {
-                        int indexAdditionalField = getColIndexByName(headerNames, columnName);
-                        if (indexAdditionalField != -1) {
-                            propertiesMap.put(columnName, parts[indexAdditionalField]);
-                        }
+            // extract fields; replace optional fields with the Stable ID when not set
+            String stableId = parts[indexStableIdField];
+            HashMap<String, String> propertiesMap = new HashMap<>();
+            if (additionalProperties != null) {
+                String[] columnNameList = additionalProperties.trim().split(",");
```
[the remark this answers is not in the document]
dippindots: We don't have an `additionalProperties generator`. Should I just create a generator names `additionalPropertiesGenerator` in the `core/src/main/java/org/mskcc/cbio/portal/util`?

dippindots:
> Can you please change the base? There are many files that were updated.

Hi @pvannierop, the base is based on the previous pr you have already reviewed: https://github.com/cBioPortal/cbioportal/pull/6818. Maybe the easier way to do this is to review the only new commit to bring the validator back, you can just review this commit at here: https://github.com/cBioPortal/cbioportal/pull/6841/commits/cd57f5967e1b8c6280a65053a37a23f2ed3e38cb. Or if you think I should rebase to some specific branch, please let me know. Thanks!

n1zea144:
Hi @pvannierop I've been discussing this PR with @dippindots, specifically the bringing back of treatment validation code and the introduction of GET /generic_assay_meta/{genericAssayType to support it.  I have some concerns about this implementation and I'm wondering if we can setup a call so that I can understand more about these checks.  I'll reach out on slack to set something up.

dippindots:
Close since we agree to not add validation back for now.

PR #6849, opened 2019-11-22
Frontend Config Additions

### LINKED ISSUE #6635: reference genome genes endpoint is a bottleneck

Luke-Sikina:
Fix #6635

- query_product_limit: Limits the size of gene queries
    - Didn't this happen a while ago? Yes. This was released in
    3.1.1, but I didn't know I needed to change this. Sorry.
- persistent_cache: Enables the frontend persistent cache

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -278,3 +278,13 @@ ehcache.cache_type=none
 
 # ensembl URL template for transcript lookup in Mutations tab. Default is http://grch37.ensembl.org/homo_sapiens/Transcript/Summary?t=<%= transcriptId %>
 #ensembl.transcript_url=http://ensembl.org/homo_sapiens/Transcript/Summary?t=<%= transcriptId %>
+
+# Enable/disable Persistent Cache (true, false)
+# The Persistent Cache is a frontend feature that is used to cache semi static
+# data produced by APIs. It uses IndexedDB as a means for persistent storage.
+# This feature is disabled by default.
+# enable_persistent_cache=true
+
+# Limit the size of gene queries (number). gene count * sample count < query_product_limit
+# This limit is enforced on the frontend.
+query_product_limit=1000000


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -279,3 +279,9 @@ matchminer.token=
 # see: https://github.com/vy/hrrs/blob/master/README.md
 #hrrs.logging.filepath=
 #hrrs.enable.logging=false
+
+# Enable/disable Persistent Cache (true, false)
+# persistent_cache=true
+
+# Limit the size of gene queries (number). gene count * sample count < query_product_limit
+query_product_limit=1000000
```
inodb: is this the current default? Should this only be for the public portal?

Luke-Sikina: This is the current default in the front end. I don't know if this should be public portal only. What determines whether or not to make a config value public portal only, and how does that change what you put in portal.properties?

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -277,4 +277,10 @@ ehcache.cache_type=none
 #hrrs.enable.logging=false
 
 # ensembl URL template for transcript lookup in Mutations tab. Default is http://grch37.ensembl.org/homo_sapiens/Transcript/Summary?t=<%= transcriptId %>
-#ensembl.transcript_url=http://ensembl.org/homo_sapiens/Transcript/Summary?t=<%= transcriptId %>
\ No newline at end of file
+#ensembl.transcript_url=http://ensembl.org/homo_sapiens/Transcript/Summary?t=<%= transcriptId %>
+
+# Enable/disable Persistent Cache (true, false)
```
Luke-Sikina: Add better comment.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -277,4 +277,14 @@ ehcache.cache_type=none
...
+# The Persistent Cache is a frontend feature that is used to cache semi static
+# data produced by APIs. It uses IndexedDB as a means for persistent storage.
+# This feature is disabled by default.
+# enable_persistent_cache=true
+
+# Limit the size of gene queries (number). gene count * sample count < query_product_limit
+# This limit is enforced on the frontend.
+query_product_limit=1000000
```
inodb: @alisman so following same logic as patient filter property, i guess for this case it might make sense to comment it out as well

alisman: agreed. @Luke-Sikina sorry to ask you to change this again.  we've been thinking about configuration today and decided that many settings (e.g. this one) really aren't meant to be managed by installers except in edge cases.  by leaving these commented in the config examples, we can maintain control over the defaults in our own codebase.

PR #6859, opened 2019-11-25
Merge master to rc

dippindots:
Merge master to rc

Fixed conflicts:
- core/src/main/scripts/importer/validateData.py
- docs/File-Formats.md
- service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java


sheridancbio:
I noticed this PR when I was thinking about creating a PR to resolve conflicts merging master to rc. This PR must be reviewed  before being merged, but when I look at it I see 49 commits which appear to come from the master branch, but then changeset c6cfcd57ad9412674c3f3e8b8dee6c4febf58224 seems to introduce widespread changes to 46 different files. If this PR was only resolving conflicts that occur during a merge, and was not introducing new functionality I would have expected only a few files would need conflict resolution .. so I'm not sure how to proceed.

It seems like master and rc are continuing to drift apart. I would look at this and possibly merge it if @dippindots can give me direction on how to interpret the large changeset and why so much conflict resolution was needed. Alternatively, I may try to take a stab at this myself and create a fresh PR which merges master into rc.

Maybe this PR needs to be brought to somebody's attention? Currently there are no requested reviews, and the PR is assigned to @dippindots (the author) ... if we want this merged we need to get somebody else to look at it. cc: @n1zea144 @inodb

PR #6860, opened 2019-11-25
add a new all_cases_with_structural_variant_data to the validation process

### LINKED ISSUE #6741: Fusion tab crashes on mixed structural variant types
The Fusion tab of the results page appears to operate in two modes ... an enhanced visualization mode which renders exon structure diagrams to show fused transcripts, and a basic mode which renders a pie chart with a textual list of fused gene pairings.

Two levels of detail are permitted for the structural variant file format, one which requires exon numbers and transcript ids for site1 and site2 ... and one which does not. The problem occurs when some records are imported with the enhanced level of detail and some records are imported with the non-enhanced level. The frontend page was tested with two fusion records involving a common site1 gene. One was in the enhanced format and one was not. Either record alone would be successfully rendered, but the fusion page would crash when both events were present in the database for the test study.

We would like the frontend to be able to render a mixture of both record detail levels. cc: @n1zea144

khzhu:
Fix #6741 

- add a new all_cases_with_structural_variant_data to the SampleListCategory class
- add a new all_cases_with_structural_variant_data to processCaseListDirectory in the Validate script

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
@@ -52,6 +52,7 @@ public enum SampleListCategory {
     ALL_CASES_WITH_MUTATION_AND_CNA_DATA("all_cases_with_mutation_and_cna_data"),
     ALL_CASES_WITH_MUTATION_AND_CNA_AND_MRNA_DATA("all_cases_with_mutation_and_cna_and_mrna_data"),
     ALL_CASES_WITH_GSVA_DATA("all_cases_with_gsva_data"),
+    ALL_CASES_WITH_SV_DATA("all_cases_with_structural_variant_data"),
     OTHER("other");
 
     // Init the look up map.
--- a/core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
@@ -52,7 +52,7 @@ public enum SampleListCategory {
     ALL_CASES_WITH_MUTATION_AND_CNA_DATA("all_cases_with_mutation_and_cna_data"),
     ALL_CASES_WITH_MUTATION_AND_CNA_AND_MRNA_DATA("all_cases_with_mutation_and_cna_and_mrna_data"),
     ALL_CASES_WITH_GSVA_DATA("all_cases_with_gsva_data"),
-    ALL_CASES_WITH_SV_DATA("all_cases_with_structural_variant_data"),
+    ALL_CASES_WITH_SV_DATA("all_cases_with_sv_data"),
     OTHER("other");
 
     // Init the look up map.
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -4253,6 +4253,7 @@ def processCaseListDirectory(caseListDir, cancerStudyId, logger,
                                 'all_cases_with_mutation_and_cna_data',
                                 'all_cases_with_mutation_and_cna_and_mrna_data',
                                 'all_cases_with_gsva_data',
+                                'all_cases_with_structural_variant_data',
                                 'other']
 
             # If the case list category is invalid, the importer will crash.
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -4253,7 +4253,7 @@ def processCaseListDirectory(caseListDir, cancerStudyId, logger,
                                 'all_cases_with_mutation_and_cna_data',
                                 'all_cases_with_mutation_and_cna_and_mrna_data',
                                 'all_cases_with_gsva_data',
-                                'all_cases_with_structural_variant_data',
+                                'all_cases_with_sv_data',
                                 'other']
 
             # If the case list category is invalid, the importer will crash.


sheridancbio:
I've approved, but I wonder if this fix alone will fix #6741. Have we imported a file with both fully specified structural variants (Site1 and Site2 exon number and transcript id provided), and also non-specified structural variants (Site1 and Site2 exon number and transcript id blank or NA)? Is it properly visualized?

khzhu:
do you have any test data for the second use case @sheridancbio? thanks!

n1zea144:
@khzhu The file in data hub for for sv in msk-2017-impact should contain both types of data.  Does it not?  cc: @ritikakundra 

will follow-up on slack

n1zea144:
@ritikakundra @rmadupuri Can you make sure that you are using the same caselist name in the sv caselist for msk-2017-impact as @khzhu is using here?

Also, can you confirm data_SV.txt file on datahub contains our full set of SV data (both sv and fusion record types - see @sheridancbio comment above).

Thanks.

khzhu:
Thanks, @n1zea144!  
I checked the data_sv.txt file and found some events that either do not have site1_exome or site2_exome. Those cases were all filtered by the importer script and did not make their way into the database. Should we relax importing rules to let them go through? 
We do not have any cases which site1_exome and site2_exom are both null included in data_sv.txt.

core/src/main/java/org/mskcc/cbio/portal/model/SampleListCategory.java
```diff
@@ -52,6 +52,7 @@
     ALL_CASES_WITH_MUTATION_AND_CNA_DATA("all_cases_with_mutation_and_cna_data"),
     ALL_CASES_WITH_MUTATION_AND_CNA_AND_MRNA_DATA("all_cases_with_mutation_and_cna_and_mrna_data"),
     ALL_CASES_WITH_GSVA_DATA("all_cases_with_gsva_data"),
+    ALL_CASES_WITH_SV_DATA("all_cases_with_structural_variant_data"),
```
rmadupuri: Hi @khzhu, we are using `all_cases_with_sv_data` as the category name instead of `all_cases_with_structural_variant_data`.

core/src/main/scripts/importer/validateData.py
```diff
@@ -4253,6 +4253,7 @@ def processCaseListDirectory(caseListDir, cancerStudyId, logger,
                                 'all_cases_with_mutation_and_cna_data',
                                 'all_cases_with_mutation_and_cna_and_mrna_data',
                                 'all_cases_with_gsva_data',
+                                'all_cases_with_structural_variant_data',
```
rmadupuri: Same here, should be `all_cases_with_sv_data`

ritikakundra:
@n1zea144 @khzhu Datahub has the full SV file (including both versions).

core/src/main/scripts/importer/validateData.py
khzhu: okay, will make corrections. thanks @rmadupuri!

khzhu:
thanks, @ritikakundra! I will modify the import script so that those events without exome numbers not being rejected.

sheridancbio:
> Thanks, @n1zea144!
> I checked the data_sv.txt file and found some events that either do not have site1_exome or site2_exome. Those cases were all filtered by the importer script and did not make their way into the database. Should we relax importing rules to let them go through?
> We do not have any cases which site1_exome and site2_exom are both null included in data_sv.txt.

Hi Kelsey. This is just a partial answer to your question.

There are three categories of SV records during import. There are 4 key fields:
- site 1 exon number
- site 2 exon number
- site 1 ensembl transcript id
- site 2 ensembl transcript id

There are basically 3 categories of records:
1. records with all 4 pieces of information (fully specified)
1. records with 1-3 pieces of information (partially specified)
1. records with 0 pieces of information (unspecified)

In the (now hidden) documentation here:
https://github.com/cBioPortal/cbioportal/blob/cbdd9d2657fbe90f01d6d258798028b097ad33ae/docs/File-Formats.md#structural-variants-data

the fully specified records are considered valid "fusion" events
all records in the structural variant data file are "structural variant" records (this includes the "fusion" events and any other less specified events)

The frontend fails to render the studies (and maybe crashes - I cannot recall the exact details) when a genetic profile contains any partially specified records ... so that is why the filter is in place to remove these events during import. We cannot remove that filter until we write some kind of code to either handle these properly in the frontend, or to "downgrade" these records to remain as unspecified records. We have basically postponed this work / decision.

The other issue is that although the frontend can display genetic profiles made up entirely of unspecified records .. or fully specified records ... it cannot render a mixture and the visual display crashes when opening the fusions tab.

If the data_SV.txt file from mskimpact_2017 study contains no unspecified records then for testing you can simply blank out (or make NA) the fields for some of the partially specified records which would demote them to become unspecified records and allow them to be imported. Then importing the profile will give a mixture of unspecified records and fully specified records ... and the frontend currently cannot render such a mixture. We are hoping the frontend can me massaged to render both of these categories at once.

I'll need to respond with more details later.

khzhu:
thanks, @sheridancbio , for detailed information!
I will adjust the importing script to let category 2 go through. Hope that can help reproduce the error (fusion tab crash).
Does not seem like we have the third category: **_records with 0 pieces of information (unspecified)_**
in data_sv.txt downloaded from the data hub.

core/src/main/scripts/importer/validateData.py
khzhu: done.

khzhu:
Hi @sheridancbio , I've made changes to use **_all_cases_with_sv_data_** instead. The PR is ready to merge. 

We will have a new PR for 
- importing the profile with a mixture of unspecified records and fully specified records
- the frontend can render both of these categories:
       - records with all 4 pieces of information (fully specified)
       - records with 0 pieces of information (unspecified)

sheridancbio:
I am canceling the request for review from @ritikakundra because she was part of the conversation about all_cases_with_structural_variant_data v.  all_cases_with_sv_data with @rmadupuri yesterday and also agreed.

sheridancbio:
I just noticed that I failed to squash the two changesets into one while merging. Apologies.

PR #6861, opened 2019-11-26
Backend Routing Improvements

### LINKED ISSUE #6768: Error showing message for non existing page
https://www.cbioportal.org/results/network

Expected behavior:
We see the 404 Page Does Not Exist route (routes.tsx)

Actual behavior:
If you go this page, you'll see a blank page and an error message in console.  

image

Luke-Sikina:
Fix #6768

- Problems:
    - Routes with `//` in them were 500ing when they should really 400
    - Valid frontend routes to tabs other than the main app tab returned
    a 404 response code, even though they are valid routes with
    important information.
        - This seems to be causing these routes to not be indexed by
        search engines
- Solutions:
    - Added an exception handler for the RequestRejectedException that
    serves up a little error page.
    - Added a generic servlet mapping that catches all requests and
    serves index.jsp
        - This way any request not covered by any other mapping doesn't
        404
        - Api requests still 404 correctly, and even give the underlying
        tomcat error message instead of the frontend 404 page
    - Added mappings for the default servlet for the directories of
    static files we still need served the old way

### DIFF
--- a/portal/src/main/webapp/WEB-INF/web.xml
+++ b/portal/src/main/webapp/WEB-INF/web.xml
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <web-app version="3.0" xmlns="http://java.sun.com/xml/ns/javaee"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns:mvc="http://www.springframework.org/schema/mvc"
+         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd">
 
     <!-- general settings -->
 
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/error.html
```diff
@@ -0,0 +1,13 @@
+<html>
```
Luke-Sikina: Apparently the RequestRejectedException is a way of handling potentially dangerous urls. As a result, I felt that the url itself was potentially dangerous, so I didn't want to use it to reference any other paths. That's why I chose to not pull in any css / images or use an existing jsp.

alisman:
@n1zea144 this issue is unrelated.  it has to do with the fact that we never cleaned up for the formerly existant route /results/network.

n1zea144:
Ok, I saw #6768 referenced at the top of this PR.

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -212,7 +213,18 @@
 
 
     <!-- servlet definitions & mappings -->
-
+    <servlet-mapping>
```
n1zea144, line 216: @Luke-Sikina Can we consolidate this mapping with the mappings on [line 589](https://github.com/cBioPortal/cbioportal/blob/9f6344eaf3323d828f41e5cfa53e454ca7fb62f6/portal/src/main/webapp/WEB-INF/web.xml#L589)

Luke-Sikina: Yes, fixed.

portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -568,6 +575,69 @@
...
+    <!--To stop them from being directed to the SPA jsp instead-->
+    <servlet>
+        <servlet-name>LoginJSP</servlet-name>
+        <jsp-file>login.jsp</jsp-file>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>LoginJSP</servlet-name>
+        <url-pattern>login.jsp</url-pattern>
+    </servlet-mapping>
+    <servlet>
```
alisman, line 588: network kill it

Issue #6876, opened 2019-12-03
Matchminer api being called on instances where it is not configured


alisman:
From user:

I have updated our local cBioPortal to version 3.1.6 and also added a Keycloak instance for authentication (which however is currently switched off during testing). For deployment I use docker-compose.
Unfortunately the patient view doesn't load anymore but gives a status 500 error. It seems to be related to the matchminer api (matchminer.url and matchminer.token are left blank in portal.properties).
For testing I'm using own mock study data, but I also tried a public dataset (prad_mpcproject_2018). Downgrading cBioPortal gradually to version 3.1.0 did not change anything.

alisman:
This was caused by matchminer configuration settings being added to example portal.properties file.  That has been corrected.

PR #6877, opened 2019-12-03
Quick fix for deployment without matchminer

victoria34:
Fix https://github.com/cBioPortal/cbioportal/issues/6876

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -268,10 +268,6 @@ ehcache.cache_type=none
 # Enable/Disable quick search (currently in beta, default is false)
 # quick_search.enabled=true
 
-# MatchMiner AWS url
-matchminer.url=
-matchminer.token=
-
 # enable/disable hrrs logging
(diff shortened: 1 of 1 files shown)


victoria34:
> We should also handle when the token/url is not specified. In this case, the service should be disabled instead of giving 500 code.

That's a good point. I'll implement the test later.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -268,9 +268,9 @@ ehcache.cache_type=none
 # Enable/Disable quick search (currently in beta, default is false)
 # quick_search.enabled=true
 
-# MatchMiner AWS url
-matchminer.url=
-matchminer.token=
+# MatchMiner setting: Used for matching patients with OncoKB curated trials.
```
alisman: @victoria34 i think we should just get rid of these from this file for now as it is not a public feature and we don't want to anyone to know about it yet.

victoria34: Removed

PR #6878, opened 2019-12-03
Re-enable support for posted query data

alisman:
Portal queries often contain too much data to represent in URL.  In the past the portal allowed users to post query data to the results page, which would treat posted parameters the same as if they were in URL.   We stopped support for this when we switched to SPA.  This PR (and an associated frontend PR) will resume this feature.


alisman:
@inodb yes.  this is actually only the first step.  frontend needs to then process this.  so, i will put in a test when i wire that up in frontend PR.

pvannierop:
@alisman @inodb I think that we have to make 2 tests:

1. Test correct response of backend to POST request.
2. Correct handling of results of POST request by the frontend.

I think that 1. should be part of this PR. I will look at this today.

PR #6888, opened 2019-12-05
Update generic assay importer (provide updateInfo option when import)

dippindots:
Provide override option when import generic assay data

related [document](https://docs.google.com/document/d/1f3JFG7tPiYXcQsJr5GHRzOK-ko_epBhOjDJa1ppgdsA/edit?disco=AAAADg3rhvo) updated here.

Changes in this pr:
1. add Provide override option (`updateInfo`) when import generic assay data (set `false` as default).
2. log the import results after import
3. Add one more method in DaoGenericAssay to support update of the entities.
4. Do not delete entities if shared with other studies.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
@@ -540,8 +540,6 @@ public final class DaoCancerStudy {
     public static void deleteCancerStudy(int internalCancerStudyId) throws DaoException {
         String[] deleteStudyStatements = {
                 "DELETE FROM sample_cna_event WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?)",
-                "DELETE FROM generic_entity_properties WHERE GENETIC_ENTITY_ID IN (SELECT GENETIC_ENTITY_ID FROM genetic_alteration WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?))",
-                "DELETE FROM genetic_entity WHERE ID IN (SELECT GENETIC_ENTITY_ID FROM genetic_alteration WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=? AND GENETIC_ALTERATION_TYPE='GENERIC_ASSAY'))",
                 "DELETE FROM genetic_alteration WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?)",
                 "DELETE FROM genetic_profile_samples WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?)",
                 "DELETE FROM sample_profile WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?)",
@@ -577,6 +575,7 @@ public final class DaoCancerStudy {
         ResultSet rs = null;
         try {
             con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            deleteGenericAssayMeta(internalCancerStudyId);
             for (String statementString : deleteStudyStatements) {
                 pstmt = con.prepareStatement(statementString);
                 if (statementString.contains("?")) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenericAssayEntity.java
```diff
@@ -117,8 +121,11 @@ public void run() {
                 "'entityType' argument required");
             }
             
-            // Check options
-            boolean updateInfo = options.has("update-info");
+            // Check options, set default as true
+            boolean updateInfo = true;
```
n1zea144: Good - by default, entities get replaced.

jjgao:
@dippindots is this still relevant?

dippindots:
@jjgao Yes, it is still relevant.

jjgao:
@dippindots what are the next steps to get this merged?

dippindots:
@jjgao No, this is the one we left behind, I just rebased to the latest database and it's ready to review now.

jjgao:
Thanks, @dippindots . I added it to candidate of pipeline scrum. @n1zea144

dippindots:
@n1zea144 Thanks for reviewing this pr! I remember one thing about the generic assay implementation: the generic assay meta is not stored in study level, which means the replace operation may effect others data which shares the same `stable_id`. So for this pr, maybe we should not replace the entities by default?

n1zea144:
Yeah, you are correct, by default we should not overwrite genetic entities.  We should only give the user the option by using the flag.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenericAssay.java
```diff
@@ -85,4 +85,29 @@ public static void deleteGenericEntityPropertiesByStableId(String stableId) thro
...
     }
+
+    public static boolean shouldGenericAssayMetaBeDeleted(int cancerStudyId) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+
+        try {
+            con = JdbcUtil.getDbConnection(DaoGeneticEntity.class);
+            pstmt = con.prepareStatement("SELECT DISTINCT CANCER_STUDY_ID FROM genetic_profile WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_alteration WHERE GENETIC_ENTITY_ID IN (SELECT GENETIC_ENTITY_ID FROM genetic_alteration WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?)))");
```
n1zea144, line 96: Seems a bit circular, but the true test is below, that the result set size == 1.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenericAssay.java
```diff
@@ -85,4 +85,29 @@ public static void deleteGenericEntityPropertiesByStableId(String stableId) thro
             JdbcUtil.closeAll(DaoGeneticEntity.class, con, pstmt, rs);
         }
     }
+
+    public static boolean shouldGenericAssayMetaBeDeleted(int cancerStudyId) throws DaoException {
```
n1zea144, line 89: We are testing that genetic entities only exist in this cancer study.  Maybe we should rename this routine - geneticEntitiesOnlyExistInSingleStudy

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -622,6 +621,37 @@ public static void purgeUnreferencedRecordsAfterDeletionOfStudy() throws DaoExce
...
+        String[] deleteGenericAssayStatements = {
+                "DELETE FROM generic_entity_properties WHERE GENETIC_ENTITY_ID IN (SELECT GENETIC_ENTITY_ID FROM genetic_alteration WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=?))",
+                "DELETE FROM genetic_entity WHERE ID IN (SELECT GENETIC_ENTITY_ID FROM genetic_alteration WHERE GENETIC_PROFILE_ID IN (SELECT GENETIC_PROFILE_ID FROM genetic_profile WHERE CANCER_STUDY_ID=? AND GENETIC_ALTERATION_TYPE='GENERIC_ASSAY'))"
+                };
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            if (DaoGenericAssay.shouldGenericAssayMetaBeDeleted(internalCancerStudyId)) {
```
n1zea144, line 638: Shouldn't this test be made inside DaoCancerStudy.deleteCancerStudy() to avoid this call entirely?  (note, I think nested calls into the dao classes are problematic and I would avoid them if possible).

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenericAssay.java
```diff
@@ -85,4 +85,29 @@ public static void deleteGenericEntityPropertiesByStableId(String stableId) thro
...
+            return studies.size() == 1;
+        } catch (SQLException e) {
+            e.printStackTrace();
+        } finally {
+            JdbcUtil.closeAll(DaoGeneticEntity.class, con, pstmt, rs);
+        }
+        // do not update if there is an error
+        return false;
+    }
 }
```
n1zea144, line 113: Please add newline here.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -577,6 +575,7 @@ public static void deleteCancerStudy(int internalCancerStudyId) throws DaoExcept
         ResultSet rs = null;
         try {
             con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            deleteGenericAssayMeta(internalCancerStudyId);
```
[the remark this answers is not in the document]
dippindots: I checked the database scheme, `genetic_alteration` has a constrain **FOREIGN KEY (`GENETIC_ENTITY_ID`) REFERENCES `genetic_entity` (`ID`) ON DELETE CASCADE**. I think generic assay related data can be deleted before we delete the data in `genetic_alteration` table.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
dippindots: Do you think it's better to check at here: `DaoCancerStudy.deleteGenericAssayMeta`?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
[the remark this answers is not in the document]
n1zea144: I was thinking the check should be wrapped around the call here to prevent the method call (and grabbing of db connection - https://github.com/cBioPortal/cbioportal/blob/9c6c3f7d69ede2f7c5e8ee9bbea626dc833b10a7/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java#L578

dippindots:
@n1zea144 All comments addressed!

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -577,6 +575,10 @@ public static void deleteCancerStudy(int internalCancerStudyId) throws DaoExcept
         ResultSet rs = null;
         try {
             con = JdbcUtil.getDbConnection(DaoCancerStudy.class);
+            // check whether should delete generic assay meta
+            if (DaoGenericAssay.geneticEntitiesOnlyExistInSingleStudy(internalCancerStudyId)) {
```
n1zea144, line 579: I'm sorry I wasn't clearer here - we should move this statement up before allocating a db connection to minimize the number of connections in use at any one time.

Issue #6891, opened 2019-12-06
issue with "not profiled" annotation in patient view


tmazor:
It is possible for a gene to have an event called, even though it is not part of the gene panel. I suspect this will most often be the case for fusion partners, but it can also arise for other reasons.

The problem is that if the gene is profiled, the event is not marked as being present in any sample.

![image](https://user-images.githubusercontent.com/10508276/70341489-067c0580-1821-11ea-81ad-24b5982c3f3d.png)
https://www.cbioportal.org/patient?studyId=msk_impact_2017&caseId=P-0000377#navCaseIds=msk_impact_2017:P-0000377

Clearly those two 2 fusions were called in at least one sample, because otherwise they wouldn't be listed. But it is now impossible to know where they were called.

We need to show the sample number anytime an event is called, even if the gene wasn't profiled in that sample. Perhaps the sample number can be slightly transparent, or in some other way indicate that it wasn't profiled?

FWIW, we do this nicely in OncoPrint. If I query just this same patient, I get this:

![image](https://user-images.githubusercontent.com/10508276/70341666-75595e80-1821-11ea-88ff-c6ae0b321ab3.png)
http://bit.ly/2qxm0K4

Sjoerd-van-Hagen:
@tmazor @jjgao this should be fixed when the fusions are not loaded as mutations any more. I assume that they will also be shown separately in the patient view when that happens?

jjgao:
@Sjoerd-van-Hagen yes, this will be fixed in the proper fusion implementation. But I think we should put some code in here to show the sample number when we see mutation/cna in the sample regardless of whether the gene is in the panel.

Sjoerd-van-Hagen:
@jjgao I am not necessarily against doing that but I would prefer that we simply prevent incorrect data from entering the system in the first place. Having mutations in a gene that is not profiled should raise an alarm and prompt the data loader to have a look, don't you think?

jjgao:
@Sjoerd-van-Hagen I completely agree that data should be fixed. I think ideally the logic should be implemented in the validator. But I also think that given noisy data, we should try our best to visualize it in the frontend. The current implementation would be confusing (the user would not know which samples carry the mutation) when the data is incomplete like above.

onursumer:
@jjgao do we still want to show sample numbers with regular colors and opacity for `not profiled` genes, or do we prefer something different, like reduced opacity or different color?

onursumer:
[...]

Issue #6890, opened 2019-12-06
change default behavior of gene panel filter in patient view


tmazor:
The new filter to include genes profiled by any or all samples is great.

However, I think the default behavior should be to show all genes. The user can then chose to filter to just the genes profiled in all samples if they want. But currently we essentially hide those genes, and a user could easily not notice that they are filtered out. So I think it's better to show all and let the user actively select the filter.

For this patient, 3 genes are just hidden by default, including a likely loss-of-function mutation in a known tumor suppressor:
![image](https://user-images.githubusercontent.com/10508276/70341074-1b0bce00-1820-11ea-8242-b4063055c461.png)

![image](https://user-images.githubusercontent.com/10508276/70341110-3971c980-1820-11ea-9db0-215bcc2c1d06.png)

https://www.cbioportal.org/patient?studyId=msk_impact_2017&caseId=P-0000377#navCaseIds=msk_impact_2017:P-0000377

Sjoerd-van-Hagen:
@jjgao this can be set in portal properties!

Sjoerd-van-Hagen:
@alisman ^

alisman:
Ok, will look for property

alisman:
# setting controlling the default setting for filtering genes in patient view
skin.patientview.filter_genes_profiled_all_samples=true

alisman:
@Sjoerd-van-Hagen @jjgao @tmazor I am updating this property on public portal right now.  Does everyone agree we should reverse the default for ALL portals though?

tmazor:
yes, i think all portals should default to "show all"  @alisman

jjgao:
Maybe we don't even need this property and always show all genes?

Sjoerd-van-Hagen:
We do need this property, which is why we implemented it.

Sjoerd-van-Hagen:
@alisman I think you may be misreading the property's intention. 'filter genes profiled (in) all samples' means that the genes are shown that are profiled in all samples, so setting it to true should remove all the genes that are not profiled in all samples. Setting true means filtering out genes, setting false means keeping them all. If it works differently, that may be a bug.

I suppose this confusion means we should rename this property.

@pvannierop can you confirm that I am correct?

alisman:
@Sjoerd-van-Hagen @pvannierop @jj i think what happened here is that in the portal.properties.EXAMPLE file, the example was set to true.    We actually use this examples file as a default (overridden by kubernetes configuration).

[...]

PR #6901, opened 2019-12-10
Add News Item Gene Panel Patient View

inodb:
See rich diff: https://github.com/cBioPortal/cbioportal/pull/6901/files?short_path=6f95322#diff-6f953229832059bab3fe229d4af08b52

### DIFF
--- a/docs/News.md
+++ b/docs/News.md
@@ -1,3 +1,8 @@
+## Dec 10, 2019
+* **New Feature**: Show gene panel information on the _Patient View_ page. This helps to distinguish between whether a mutation in a particular gene was not present in the sample or not profiled ([example](https://www.cbioportal.org/patient?caseId=P-0001613&studyId=lung_msk_2017)):
+
+    ![gene_panel_news_item4](https://user-images.githubusercontent.com/1334004/70575351-58e85980-1b74-11ea-8c40-0d93ecb810eb.gif)
+
 ## Nov 29, 2019
 * **Enhancement**: Support group comparison for custom charts in _Study View_ page
 * **Enhancement**: Performance improvement of Co-Expression analysis.


docs/News.md
```diff
@@ -1,3 +1,8 @@
+## Dec 10, 2019
+* **New Feature**: Show gene panel information on the _Patient View_ page. This helps to distinguish between whether a mutation in a particular gene was not present in the sample or not profiled ([example](https://www.cbioportal.org/patient?caseId=P-0001613&studyId=lung_msk_2017)):
```
jjgao: Wondering if we can find an example with a mutation in one example but not sequenced in another?

inodb: that is in this one right or did I misunderstand? By default it's filtered:

Screen Shot 2019-12-10 at 7 14 44 PM

inodb: it might make sense to show mutations profiled in any sample by default?

tmazor: I definitely agree we should show all mutations by default - I actually made an issue about this the other day :) (#6890 )  If possible, would be great to change this before we make the announcement.

Also how about this case as an example: https://www.cbioportal.org/patient?caseId=P-0002208&studyId=msk_impact_2017
The gene that isn't on one of the panels has a likely oncogenic mutation which I think makes for a nice example.

tmazor: For what it's worth, I also found another issue (#6891) - not sure if it's practical to fix that before this news announcement, but if possible that would be great.

inodb: Thanks Tali! I like that example as well. I think we can wait with the announcement until we make those fixes. @jjgao do u agree to change the default behavior?

jjgao: I agree with @tali.

inodb:
Added here: https://github.com/cBioPortal/cbioportal/pull/6987

PR #6902, opened 2019-12-10
Move loadFrontend invocation into body to fix 3rd party library depen…

alisman:
A library used in frontend project (measure-text) requires that document.body be available at parse time (immediately after js is loaded into page).  Recent changes to our loading js packaging  strategy (having to do with monorepo) moved the library to a different bundle that gets loaded sooner in lifecycle.  For this reason, we need to move the loadFrontend call into `<body>` tag.

### DIFF
--- a/portal/src/main/webapp/index.jsp
+++ b/portal/src/main/webapp/index.jsp
@@ -95,10 +95,6 @@
         });
     </script>
     
-    <script>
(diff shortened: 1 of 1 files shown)


inodb:
Not really sure how to label this btw, "refactoring" ?

PR #6903, opened 2019-12-11
Impl. non-authorized ServerStatus (/health) endpoint

pvannierop:
# Background
For instances that use authorization, the health of cBioPortal cannot be probed since endpoints are shielded.

# Fix
This PR will implement an open `/health` endpoint that does not require authentication. It will return `{"status": "UP"}` when the cancer types can be retrieved from the database. If not, it will return `{"status": "DOWN"}` (or nothing when cBioPortal is down entirely :) ).

### DIFF
--- a/portal/src/main/webapp/WEB-INF/web.xml
+++ b/portal/src/main/webapp/WEB-INF/web.xml
@@ -251,6 +251,7 @@
     <servlet-mapping>
         <servlet-name>api</servlet-name>
         <url-pattern>/api/*</url-pattern>
+        <url-pattern>/health</url-pattern>
     </servlet-mapping>
 
     <servlet>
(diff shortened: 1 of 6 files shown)


inodb:
@pvannierop there is one bean wiring exception in the core tests - any chance you've seen that before?

averyniceday:
@pvannierop thanks for this PR, this looks great! 

I've run into this bean wiring issue in the test classes. It has to do with how the tests for the controllers are set up. I'm a little hazy on exactly why, but it requires all the services to have a bean available. This could be easily done by mocking the ServerStatusService and annotating it as a `@Bean` in any of the controller test classes.

Right now all these service beans are declared across different controller tests classes. Ideally all bean declarations would be organized in one central test configuration class and just autowired into the controller test classes that need them, but that doesn't have to be done here. cc: @inodb

n1zea144:
Hi @pvannierop Nice idea!  Its good to make an explicit endpoint for checking server health that works both with/without the security layers.  Its nice to to have the extra layer of database access checking, but as you point out - if the backend is down, /health will not respond.  @inodb I think checking cancer types is not bad, if only because they exist independent of any study data and get imported via the seed database, but maybe hitting the info table is "purer".

I like the use of ServerStatusMessage in case we want to extend the information that is provided in the future.

inodb:
superseded by https://github.com/cBioPortal/cbioportal/pull/6910 to fix tests (the tickbox "allow edits from maintainers" didn't seem to be enabled, so we couldn't change this PR)

Issue #6306, opened 2019-06-21
Add link to NG-CHM Compendium on Heatmap tab


ChrisWakefield:
(Based on conversation with Rehan Akbani and Niki Schultz:)
insert a fourth button in the Heatmap drop-down. Clicking on the button would show a disclaimer that the user is linking out to an external website and their gene and sample selections may not be the only ones represented in the map. Once the user acknowledges the disclaimer, NG-CHM would open up (perhaps in a new browser tab).

ChrisWakefield:
You can assign this to me.

ChrisWakefield:
Here I am months later.
@jjgao @zhx828 Is there (what is) a preferred way to display a modal confirmation dialog?  Recall that I am to show a disclaimer message before opening the NGCHM.

I've found some existing use of <Modal ...> and been trying to get that to do what I need, but I'm trying to avoid creating whole new files and new classes just for this.  It seems like it should be simple, but my inexperience with the react-js stack, etc is leaving me puzzled.

jjgao:
adding @alisman to this and see he has suggestions.

ChrisWakefield:
@jjgao @zhx828 @alisman Thanks.  I think I've found a solution that should be ok.  It did require more head-banging.

ChrisWakefield:
@jjgao or @zhx828 @alisman 

I'm attempting to test my changes against the rc branch (export BRANCH_ENV=rc).
Even though everything works with the same changes applied to master, I'm getting this error from rc.cbioportal.org well before getting to any of my code:

{"req":{"method":"GET","url":"https://rc.cbioportal.org/api/studies/blca_tcga_pub/molecular-profiles","headers":{"accept":"application/json"}},"xhr":{},"text":"<!doctype html><html lang=\"en\"><head><title>HTTP Status 500 – Internal Server Error</title>
...
that eventually gets to:
bad SQL grammar []; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Unknown column &#39;genetic_profile.GENERIC_ASSAY_TYPE&#39;
...

Any suggestions?  Do something different? Wait?  Submit PR anyway?
Thanks, Chris

PR #6904, opened 2019-12-11
Add configuration option for mdacc ngchm button in oncoprint heatmap menu

alisman:
https://github.com/cBioPortal/cbioportal/issues/6306

Add setting to enable/disable NG-CHM link in oncoprint heatmap menu.

### DIFF
--- a/docs/portal.properties-Reference.md
+++ b/docs/portal.properties-Reference.md
@@ -215,6 +215,13 @@ The CIViC API url is set to https://civic.genome.wustl.edu/api/ by default. It c
 civic.url=
 ```
 
+# MDACC Heatmap Integration
+
+MDACC Heatmap integration (button in OncoPrint heatmap dropdown and tab on Study page can be turned on or off by setting the following property:
+```
+show.mdacc.heatmap=true
+```
+
 # OncoPrint
 
 The default view in OncoPrint ("patient" or "sample") can be set with the following option. The default is "patient".
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -51,6 +51,7 @@
             "mdacc.heatmap.patient.url",
             "mdacc.heatmap.study.meta.url",
             "mdacc.heatmap.study.url",
+            "show.mdacc.heatmap",
             "oncoprint.custom_driver_annotation.tiers.menu_label",
             "priority_studies",
             "show.hotspot",
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -142,8 +142,9 @@ dat.jwt.secret_key=
 # study view settings
 # always show studies with this group
 always_show_study_group=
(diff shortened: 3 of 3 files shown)


alisman:
@ChrisWakefield this is unfortunately necessary to pass settings in portal.properties to frontend.

inodb:
We should also add it to properties reference: https://github.com/cBioPortal/cbioportal/blob/master/docs/portal.properties-Reference.md

ChrisWakefield:
@inodb @alisman 
> can we name this more consistent with other props
> there's also various
> `show.x` properties

I don't mind changing, but I purposefully did not make it "show" since I think that would be misleading - It's actually only shown if it's both enabled and there are compendium heatmaps to be shown.  Therefore, "enable".
I realize now I am inconsistent with existing props that begin "mdacc.xxx".

alisman:
@inodb @ChrisWakefield i changed it to 
```oncoprint.enable_mdacc_ngchm_button```
@inodb i will also update .md file after chat with you

**Comment by jjgao** [comment]
@alisman I think we should change it to `show.mdacc.heatmap` and when it's `false`, we should not show it in the OncoPrint and study view.

**Comment by alisman** [comment]
@jjgao i will change this.  this will have affect of turning OFF feature on installations where it is currently on b/c there is currently no configuration (for study view tab)

File: docs/portal.properties-Reference.md [hunk 1]
```diff
@@ -215,6 +215,13 @@ The CIViC API url is set to https://civic.genome.wustl.edu/api/ by default. It c
 civic.url=
 ```
 
+# MDACC Heatmap Integration
+
+MDACC Heatmap integration (button in OncoPrint heatmap dropdown and tab on Study page can be turned on or off by setting the following property:
+```
+show.mdacc.heatmap=true
```
inodb: If we don't have these documented yet:

```
            "mdacc.heatmap.meta.url",
            "mdacc.heatmap.patient.url",
            "mdacc.heatmap.study.meta.url",
            "mdacc.heatmap.study.url",
```

It might make sense to add those as well? B/c the mdacc heatmap won't work unless you have all these other properties right?

ChrisWakefield: @inodb @alisman Thank you both for all your help!

The first two are now unused, since they were for the patient view, and are only referenced in getMDAndersonHeatMapMetaUrl and getMDAndersonHeatmapPatientUrl, respectively.  Neither of these are called in the current frontend.  The patient view feature does work if we ever want to reinstate it.  i.e. The two properties could be left out without harm.

The next two are used by the study view tab, but they are effectively constants.
The new button in the OncoPrint does not depend on any of these, as it currently has hard-coded urls that will be changing (at our end) which will prompt minor code changes at some point.

Please let me know what you need from me.

inodb: Thanks for clarifying @ChrisWakefield ! Then maybe ok to ignore those properties in the docs for now. Another option would be to have some section "internal use only". That way people browsing the docs can be made aware of the integration, but that it's not working yet / under development

PR #6906, opened 2019-12-11
Frontend load: CSS link tags should be written to head not body

alisman:
Moving link tag write from <head> to <body> caused small changes in styling.

### DIFF
--- a/portal/src/main/webapp/js/src/load-frontend.js
+++ b/portal/src/main/webapp/js/src/load-frontend.js
@@ -27,7 +27,7 @@ function showFrontendPopup(url) {
 window.loadReactApp = function(config) {
     // Set frontend route to /patient
     window.defaultRoute = '/' + config.defaultRoute;
-    
+
     if (window.frontendConfig.frontendUrl === undefined) {
         // this should never happen
         if (console.error) {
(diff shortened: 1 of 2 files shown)


portal/src/main/webapp/js/src/load-frontend.js
```diff
@@ -39,14 +39,17 @@ window.loadReactApp = function(config) {
     if (window.localdev || window.localdist || localStorage.heroku || localStorage.netlify) {
         showFrontendPopup(window.frontendConfig.frontendUrl);
     }
+    document.write('<script src="' + window.frontendConfig.frontendUrl + 'reactapp/common.bundle.js?'+ window.frontendConfig.appVersion +'"></scr' + 'ipt>');
+    document.write('<script src="' + window.frontendConfig.frontendUrl + 'reactapp/main.app.js?'+ window.frontendConfig.appVersion +'"></scr' + 'ipt>');
+
+};
+
+window.loadAppStyles = function() {
```
inodb: where is this `loadAppStyles` called?

onursumer: this should be called somewhere in the `head`, right?

alisman: nice catch guys.  make the edit to a file in target directory. updated.

PR #6920, opened 2019-12-13
/health moves to /api/health for consistency with no-security portals

### LINKED ISSUE #6916: both /health and /api/health endpoints exist (should be only /api/health)
Not sure if this was on purpose, but prolly makes sense to only have one of `/health` and `/api/health`? `/api/health` currently doesn't work on auth portals, only `/health`. I'm open to either, but maybe having only `/api/health` makes more sense? Since all the other `/something` paths are routes to pages? CC: @pvannierop @averyniceday  

See:

https://www.cbioportal.org/health

https://www.cbioportal.org/api/health

n1zea144:
For consistency with unprotected portals.

fixes #6916

### DIFF
--- a/security/security-spring/src/main/resources/applicationContext-security.xml
+++ b/security/security-spring/src/main/resources/applicationContext-security.xml
@@ -99,7 +99,7 @@
     <http pattern="/reactapp/**" security="none"/>
     <http pattern="/api/swagger-resources/**" security="none"/>
     <http pattern="/api/swagger-ui.html" security="none"/>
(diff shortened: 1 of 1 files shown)


jjgao:
fixes #6916

inodb:
Thanks Ben! This fixes the authenticated part, but the `/health/` page is still available: https://cbioportal-slash-health-ea3x5v.herokuapp.com/health. And https://cbioportal-slash-health-ea3x5v.herokuapp.com/api/health

PR #6944, opened 2019-12-23
Update default of skin.patientview.filter_genes_profiled_all_samples …

alisman:
This property should be false by default so that we don't hide mutations which only occur in one sample


alisman:
https://github.com/cBioPortal/cbioportal/pull/6944 
this will reverse that

PR #6946, opened 2019-12-23
add access permission on generic assay meta endpoint

dippindots:
add access permission on generic assay meta endpoint


web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -103,6 +104,7 @@
     public static final String TREATMENT_FETCH_PATH = "/treatments/fetch";
     public static final String STRUCTURAL_VARIANT_FETCH_PATH = "/structuralvariant/fetch";
     public static final String GENERIC_ASSAY_DATA_MULTIPLE_STUDY_FETCH_PATH = "/generic_assay_data/fetch";
+    public static final String GENERIC_ASSAY_META_FETCH_PATH = "/generic_assay_meta/fetch";
```
inodb: @dippindots just noticed these `gneric_assay` paths have underscores in the name while other paths do not. Prolly good to be consistent?

inodb: we can fix it later but just FYI

inodb:
This is a temp fix to address auth issues, will be properly fixed later

dippindots:
@inodb Just want to mention that this is not a temp fix, already updated properly.

PR #6947, opened 2019-12-23
New configuration option for showing mskcc slide viewer

alisman:
@inodb I'm not going to put this in example file or documentation because it's really internal use only.

### DIFF
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -92,6 +92,7 @@
             "skin.show_web_api_tab",
             "skin.show_tweet_button",
             "skin.patientview.filter_genes_profiled_all_samples",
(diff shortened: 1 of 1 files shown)


portal/src/main/webapp/config_service.jsp
```diff
@@ -92,6 +92,7 @@
             "skin.show_web_api_tab",
             "skin.show_tweet_button",
             "skin.patientview.filter_genes_profiled_all_samples",
+            "skin.patientview.show_mkscc_slide_viewer",
```
inodb: @alisman just noticed a typo here, will create an issue

inodb: https://github.com/cBioPortal/cbioportal/issues/6969

PR #6951, opened 2019-12-23
update generic assay api path

dippindots:
Update generic assay endpoints by using `-` instead of `_`


dippindots:
Move discussion to https://github.com/cBioPortal/cbioportal/issues/6955

PR #6954, opened 2019-12-24
Fix migration 3.2.0

inodb:
- rename `all_cases_with_structural_variant_data`  -> `all_cases_with_sv_data` (that's what the backend code looks for)


khzhu:
Hi @inodb , 
i got this error when running your fixing, did it work for you? 
```
1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '@genetic_alteration_fk_name' at line 1, Time: 0.001000s
```

inodb:
@khzhu it seemed to be working for me after i followed this example: https://stackoverflow.com/questions/23757105/dropping-foreign-key-using-variable-as-name. But i've reverted it for now. We'll manually fix the foreign key name when applying the migration instead

khzhu:
great, thanks for the updates!

PR #6945, opened 2019-12-23
Fixes fusion gene table (visibility + counts)

inodb:
- Hide fusion table if no fusion data present
- Fix fusion counts
- backfill genetic profiles for fusions  stored in the mutation_event
table
- add a new filter for fusion data
- add a new case list for sv data

Supersedes https://github.com/cBioPortal/cbioportal/pull/6731

PR #6968, opened 2020-01-02
Structural variant fixes for 3.2.0

### LINKED ISSUE #6966: Populate structural variant (SV) case list after importing mutation data
Before fully switching to using the structural_variant table for fusions (https://github.com/cBioPortal/cbioportal/issues/6611) we will have an in-between phase where some studies might have proper structural variant data in the new format and some studies have structural variant data in the mutations table. This will be in 3.2.0.

The current issue in the release 3.2.0 branch is the following. As part of the db migration we provide some sql statements to create a structural variant case lists with those cases that have a fusion in the mutation event table: https://github.com/cBioPortal/cbioportal/blob/release-3.2.0/db-scripts/src/main/resources/migration.sql#L797-L831.  However, on reimporting mutation data, the case list does not get filled again. To get around this issue we propose to run something similar right after importing mutation data that recomputes the structural variant case lists if there are fusion events in the mutation table. The same SQL as mentioned before could be used, but instead of running on all studies, run it on a single study (the study that is currently being imported). We already have some step where we compute the mutation count clinical attribute after import, so we could add another step after that were we compute the structural variant case lists:

https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java#L476-L483

We will have to remove this implementation later once we completely move over to the structural_variant table. Might be good to add some comments in the code what parts can be removed.

khzhu:
Fix #6966

- [x] modify the existing GeneticProfileReader DAO object to add a structural variant profile when importing a study with old style fusion data
- [x] modify existing Sample service to mark samples profiled by Fusion and supports any studies with or without a _sv case list
- [x] update File-Formats.md to include a new structural variant/fusion case list which will be used to calculate the percentage of samples with fusions.
- [x] use sequenced case list to calculate the percentage of samples with fusions if no structural variant/fusion case list suplied

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


khzhu:
Hi @jjgao @n1zea144 @inodb @sheridancbio @alisman , the PR is ready for the review. Please advise. Your feedback would be greatly appreciated.

khzhu:
looped Gaofei in. Thank you!

khzhu:
Hi @inodb and @dippindots ,
I got this error from end_to_end_tests_localdb in CI. Any suggestions for how to get it fixed, as I could not see how the test is related to the changes proposed in this PR. Thank you!
```
1) patient view page gene panel modal toggles gene panel modal from genomic tracks:
false == true
running
AssertionError [ERR_ASSERTION]: false == true
    at Context.it (/cbioportal-frontend/end-to-end-test/local/specs/core/patientview.spec.js:267:17)
    at new Promise (<anonymous>)
    at new F (/cbioportal-frontend/end-to-end-test/node_modules/core-js/library/modules/_export.js:35:28)
```

dippindots:
@khzhu Hi Kelsey, thanks for making this pr! This localdb test will fail sometimes, we can ignore this for now.

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
```diff
@@ -184,7 +184,7 @@ public void testLoadStudyEs0() throws Throwable {
             geneticProfileStableIds.add("study_es_0_fusion");
```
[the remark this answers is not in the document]
khzhu: hi @inodb, thank you for the review. Let me see if I can explain well.
Since fusions are currently stored in the mutation_event table, fusions have to use the mutation profile in order to get the data inserted into the database, which was okay during the time when we did not have the fusion gene table in the summary view. But now, since we added the fusion gene table in the summary view, it always shows up regardless if a study has fusion data or not, because the frontend uses genetic profiles to control the visibility of gene tables (please see [issue](https://github.com/cBioPortal/cbioportal/issues/6709) reported by Tali). To fix the issue, we decided to give fusion data its own genetic profile, which is a **_STRUCTURAL_VARIANT_** genetic type profile with a **__fusion_** stableId
```
mysql> select distinct genetic_alteration_type from genetic_profile;
+-------------------------+
| genetic_alteration_type |
+-------------------------+
| COPY_NUMBER_ALTERATION  |
| STRUCTURAL_VARIANT      |
| MUTATION_EXTENDED       |
| PROTEIN_LEVEL           |
| METHYLATION             |
| MRNA_EXPRESSION         |
+-------------------------+
6 rows in set (0.00 sec)
```
```
mysql> select * from genetic_profile where stable_id like '%_fusion';
+--------------------+------------------------+-----------------+-------------------------+----------+---------+-------------+------------------------------+-----------------+------------+--------------------+
| GENETIC_PROFILE_ID | STABLE_ID              | CANCER_STUDY_ID | GENETIC_ALTERATION_TYPE | DATATYPE | NAME    | DESCRIPTION | SHOW_PROFILE_IN_ANALYSIS_TAB | PIVOT_THRESHOLD | SORT_ORDER | GENERIC_ASSAY_TYPE |
+--------------------+------------------------+-----------------+-------------------------+----------+---------+-------------+------------------------------+-----------------+------------+--------------------+
|                 45 | msk_impact_2017_fusion |              16 | STRUCTURAL_VARIANT      | FUSION   | Fusions | Fusions.    |                            1 |            NULL | NULL       | NULL               |
|                122 | breast_msk_2018_fusion |              39 | STRUCTURAL_VARIANT      | FUSION   | Fusions | Fusions.    |                            1 |            NULL | NULL       | NULL               |
+--------------------+------------------------+-----------------+-------------------------+----------+---------+-------------+------------------------------+-----------------+------------+--------------------+
2 rows in set (0.00 sec)
```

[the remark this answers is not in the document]
khzhu: if you take a look at the migration script we used **__sv_** STABLE_ID for fusion data (we should use _fusion instead to be consistent). The STABLE_ID does not really matter, what is important is that we give fusion data a **_GENETIC_ALTERATION_TYPE_** genetic profile. I used **__fusion_** and **_structural_variants_** STABLE_ID to distinguish old style fusion data and new structural variant data. msk-impact-2017 is the only study that has both fusions and structural variants.
```
mysql> select cs.cancer_study_identifier, gp.stable_id, gp.genetic_alteration_type, gp.datatype, gp.name from genetic_profile gp join cancer_study cs on gp.CANCER_STUDY_ID = cs.CANCER_STUDY_ID where gp.CANCER_STUDY_ID = 16 and GENETIC_ALTERATION_TYPE = 'STRUCTURAL_VARIANT';
+-------------------------+-------------------------------------+-------------------------+----------+-----------------------+
| cancer_study_identifier | stable_id                           | genetic_alteration_type | datatype | name                  |
+-------------------------+-------------------------------------+-------------------------+----------+-----------------------+
| msk_impact_2017         | msk_impact_2017_structural_variants | STRUCTURAL_VARIANT      | SV       | mskimpact2017 SV Data |
| msk_impact_2017         | msk_impact_2017_fusion              | STRUCTURAL_VARIANT      | FUSION   | Fusions               |
+-------------------------+-------------------------------------+-------------------------+----------+-----------------------+
2 rows in set (0.00 sec)
```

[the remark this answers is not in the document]
khzhu: good catch! changed.

khzhu:
> // Is there a separate fusion profile? -> false

good catch! changed to **_true_**.

khzhu:
I've resolved all comments from @inodb, thank you!

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -82,6 +82,12 @@ public static GeneticProfile loadGeneticProfile(File file) throws IOException, D
                throw new RuntimeException("Error: genetic_profile record found with same Stable ID as the one used in your data:  "
                        + existingGeneticProfile.getStableId() + ". Remove the existing genetic_profile record first.");
             } else if (geneticProfile.getDatatype().equals("FUSION")) {
+                // populate the structural variant genetic profile for fusions
+                GeneticProfile gp = new GeneticProfile(geneticProfile);
+                gp.setGeneticAlterationType(GeneticAlterationType.STRUCTURAL_VARIANT);
+                gp.setStableId(gp.getStableId().replace("mutations","fusion"));
```
inodb: I guess this relies on this part? https://github.com/cBioPortal/cbioportal/blob/7ae3828ac87d2f88703d8f9f3f60d5d756085a4b/core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java#L290-L301

it's a little funky but I guess we can remove all the `FUSION` conditional once we remove `FUSION` data and only have `STRUCTURAL_VARIANT` GeneticAlterationType:

https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/model/GeneticAlterationType.java#L42-L43

core/src/test/java/org/mskcc/cbio/portal/scripts/TestIntegrationTest.java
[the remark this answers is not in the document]
inodb: Thanks @khzhu - that explains a lot

db-scripts/src/main/resources/migration.sql
```diff
@@ -806,7 +806,7 @@ AS (
     WHERE MUTATION_TYPE = 'fusion'
 );
 INSERT INTO genetic_profile(STABLE_ID, CANCER_STUDY_ID, GENETIC_ALTERATION_TYPE, DATATYPE, NAME, DESCRIPTION, SHOW_PROFILE_IN_ANALYSIS_TAB)
-SELECT CONCAT(CANCER_STUDY_IDENTIFIER, '_structural_variants'), CANCER_STUDY_ID, 'STRUCTURAL_VARIANT','SV','Fusions','Fusions',0
+SELECT CONCAT(CANCER_STUDY_IDENTIFIER, '_fusion'), CANCER_STUDY_ID, 'STRUCTURAL_VARIANT','SV','Fusions','Fusions',0
```
inodb: Thanks for updating this! Just one issue: we have already applied this migration to one production database. 

I guess I can rerun this thing?

From: https://github.com/cBioPortal/cbioportal/pull/6945#issuecomment-568592112

```
delete from genetic_profile where STABLE_ID like '%_structural_variants%'
delete from sample_list_list where list_id in (select list_id from sample_list where CATEGORY = 'all_cases_with_sv_data')
delete from sample_list where CATEGORY = 'all_cases_with_sv_data'
```

Another thing - now that you've updated the sample service to support studies with a case list. Should we just remove this whole part of filling in case lists? I think it would be more comprehensible if running the migration on old studies with FUSION datafiles results in the same state of the database as importing the data. Right now from what I understand the migration creates a case list, but importing a study with FUSION data only adds a new profile.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -193,13 +201,17 @@ private void processSamples(List<Sample> samples, String projection) {
...
                     .contains(sample.getStableId()));
                 sample.setCopyNumberSegmentPresent(samplesWithCopyNumberSegMap.contains(sample.getInternalId()));
-                if (!structuralVariantSampleIdsMap.isEmpty()) {
-                    sample.setProfiledForFusions(structuralVariantSampleIdsMap.get(sample.getCancerStudyIdentifier()).contains(sample.getStableId()));
+                if (studiesProfiledWithSVs.contains(sample.getCancerStudyIdentifier())) {
+                    if (!structuralVariantSampleIdsMap.get(sample.getCancerStudyIdentifier()).isEmpty()) {
+                        sample.setProfiledForFusions(structuralVariantSampleIdsMap.get(sample.getCancerStudyIdentifier()).contains(sample.getStableId()));
+                    } else {
+                        sample.setProfiledForFusions(sequencedSampleIdsMap.get(sample.getCancerStudyIdentifier()).contains(sample.getStableId()));
+                    }
```
inodb: might be good to have a comment here. Something like:

```
/*
TODO: Eventually all studies with STRUCTURAL_VARIANT data should have case lists, so there should always be an entry in `structuralVariantSampleIdsMap`. This case is to support old `FUSION` data in the mutations table that don't have case lists. In that case we assume any sample that has been sequenced to have been profiled for fusions as well
*/
```
Makes sense?

inodb: It would be good to create an issue as well pointing to these various parts in the code where FUSION data should be cleaned up. I think by having `FUSION` specifically as a comment it will be easy to find for whoever tries to remove all `FUSION` related parts from the code.

service/src/test/java/org/cbioportal/service/impl/SampleServiceImplTest.java
```diff
@@ -259,6 +263,15 @@ public void fetchSamplesDetailed() throws Exception {
...
             Mockito.anyListOf(String.class), Mockito.anyString())).thenReturn(expectedInternalIdList);
+
+        List<MolecularProfile> expectedMolecularProfileList = new ArrayList<>();
+        MolecularProfile molecularProfile = new MolecularProfile();
+        molecularProfile.setCancerStudyIdentifier(STUDY_ID);
+        molecularProfile.setMolecularAlterationType(MolecularProfile.MolecularAlterationType.STRUCTURAL_VARIANT);
+        expectedMolecularProfileList.add(molecularProfile);
+
+        Mockito.when(molecularProfileRepository.getMolecularProfilesInStudies(Arrays.asList(STUDY_ID), "DETAILED"))
+            .thenReturn(expectedMolecularProfileList);
```
inodb: nice test

db-scripts/src/main/resources/migration.sql
inodb: UPDATE: added note. Now that I think more about it. We could maybe get rid of the required migration and changes to the importer altogether if we change the service layer to return a `_fusion` profile if there are fusions in the mutation table. I don't think it would affect performance too much to find those. The advantage of that approach is that we're only touching the service layer, so:

- no migration necessary
- easier to remove the old FUSION related code since it's limited to only service layer

core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
khzhu: That is right, @inodb, till we completely deprecate the Fusion genetic type, it might take a while.

db-scripts/src/main/resources/migration.sql
khzhu: you do not have to remove case list @inodb, just update the genetic_profile.STABLE_ID using following statement 
```
UPDATE genetic_profile
    INNER JOIN cancer_study ON cancer_study.CANCER_STUDY_ID = genetic_profile.CANCER_STUDY_ID
SET genetic_profile.STABLE_ID = CONCAT(CANCER_STUDY_IDENTIFIER, '_fusion')
WHERE genetic_profile.STABLE_ID like '%_structural_variants%' AND genetic_profile.datatype = 'FUSION' AND genetic_profile.genetic_alteration_type = 'STRUCTURAL_VARIANT'
```

khzhu: > Another thing - now that you've updated the sample service to support studies with a case list. Should we just remove this whole part of filling in case lists?

good catch, @inodb! we do not need backfilling case lists for fusions any more, but it will not hurt to leave the way as it is now as well.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
khzhu: good idea, @inodb!

db-scripts/src/main/resources/migration.sql
inodb: @khzhu agreed that keeping the case list migration wouldn't affect anything for the user, but for sake of clarity i think it would be cleaner to remove the backfilling. Because the state of the database will differ when you import the study from when you perform the migration. If e.g. somebody works on this code in the future they might assume the case lists to be there based on the migration code

What are your thoughts on removing the migration code & importer code altogether and changing the service layer instead to return a fusion profile if there's fusion data in the mutations table? Avoiding a database migration and importer update might make the rollout process easier

inodb: There is ofc some performance cost to this, but it might be minor

khzhu: hi @inodb , sorry for a late comment. 

> Because the state of the database will differ when you import the study from when you perform the migration. If e.g. somebody works on this code in the future they might assume the case lists to be there based on the migration code

agree with you

> What are your thoughts on removing the migration code & importer code altogether and changing the service layer instead to return a fusion profile if there's fusion data in the mutations table? Avoiding a database migration and importer update might make the rollout process easier

probably not a good idea I would say. Genetic profile is an ORM object mapped to the genetic_profile entity in the database. Each genetic profile is corresponding to a row in the genetic profile table.

inodb: @khzhu no worries! Thanks so much for still replying to this 😃 ! 

Let's change the migration script to not generate case lists then and leave creation of the fusion profile in there

khzhu: my pleasure, @inodb . Will make changes. thanks again for the review.

khzhu: Hi @inodb , migration script updated. please let me know if there is anything else need to be addressed. Thank you!

PR #6976, opened 2020-01-03
fusion profile migration (temporary solution for fusion data)

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/6966

Run scripts after imported fusion data.

CC: @inodb @alisman @n1zea144 @sheridancbio @khzhu

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -1602,4 +1602,58 @@ public final class DaoMutation {
         }
         return !(sqlQueryResult > 0);
     }
+
+    public static void createStructuralVariantData(int cancerStudyId) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+
+        String[] statements = {
+            "INSERT INTO genetic_profile (STABLE_ID, CANCER_STUDY_ID, GENETIC_ALTERATION_TYPE, DATATYPE, NAME, DESCRIPTION, SHOW_PROFILE_IN_ANALYSIS_TAB) " +
+            "SELECT DISTINCT CONCAT(CANCER_STUDY_IDENTIFIER, '_structural_variants'), cancer_study.CANCER_STUDY_ID, 'STRUCTURAL_VARIANT','SV','Fusions','Fusions',0 " +
+            "FROM `mutation_event` " +
+            "JOIN `mutation` ON `mutation`.MUTATION_EVENT_ID = `mutation_event`.MUTATION_EVENT_ID " +
+            "JOIN `genetic_profile` ON `genetic_profile`.GENETIC_PROFILE_ID = `mutation`.GENETIC_PROFILE_ID " +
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportSampleList.java
```diff
@@ -132,6 +132,11 @@ public static void importSampleList(File dataFile) throws IOException, DaoExcept
 
       sampleList = daoSampleList.getSampleListByStableId(stableId);
 
+      if (stableId.contains("_sequenced")) {
+         // Temporary create new format of SV data from fusion data
+         DaoMutation.createStructuralVariantData(cancerStudyId);
+      }
+
```
inodb: Not sure if the check should be in import sample list. I think it should be after import of mutation data. I'm not sure if we can assume the sample list gets imported after the mutation data

khzhu: agree with Ino, we should not make an assumption on the order in which the data got inserted.
Also, this would not work for studies that have the structural variants case list included. For example msk-impact-2017 study. 
in general, we should support two use cases:
1. importing a study that has a _sv case list 
2. importing a study without a _sv case list supplied.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -1602,4 +1602,58 @@ public static boolean hasDuplicateMutationEvents() throws DaoException {
...
+        String[] statements = {
+            "INSERT INTO genetic_profile (STABLE_ID, CANCER_STUDY_ID, GENETIC_ALTERATION_TYPE, DATATYPE, NAME, DESCRIPTION, SHOW_PROFILE_IN_ANALYSIS_TAB) " +
+            "SELECT DISTINCT CONCAT(CANCER_STUDY_IDENTIFIER, '_structural_variants'), cancer_study.CANCER_STUDY_ID, 'STRUCTURAL_VARIANT','SV','Fusions','Fusions',0 " +
+            "FROM `mutation_event` " +
+            "JOIN `mutation` ON `mutation`.MUTATION_EVENT_ID = `mutation_event`.MUTATION_EVENT_ID " +
+            "JOIN `genetic_profile` ON `genetic_profile`.GENETIC_PROFILE_ID = `mutation`.GENETIC_PROFILE_ID " +
+            "JOIN `cancer_study` ON `cancer_study`.CANCER_STUDY_ID = `genetic_profile`.CANCER_STUDY_ID " +
+            "WHERE MUTATION_TYPE = 'fusion' " +
+            "AND NOT EXISTS (SELECT * FROM genetic_profile WHERE STABLE_ID=CONCAT(CANCER_STUDY_IDENTIFIER, '_structural_variants') AND CANCER_STUDY_ID = cancer_study.CANCER_STUDY_ID)" +
+            "AND cancer_study.CANCER_STUDY_ID = ?;",
```
khzhu: insert a new genetic_profile that should be delegated to the GeneticProfileReader dao object.

inodb:
Superseded by: https://github.com/cBioPortal/cbioportal/pull/6968

PR #7015, opened 2020-01-17
added a config parameter to indicate whether OncoKB token is defined

onursumer:


portal/src/main/webapp/config_service.jsp
```diff
@@ -154,7 +154,9 @@
                 enableOncoKBandHotspots = "\"custom\"";
         }
              
-        obj.put("oncoprintOncoKbHotspotsDefault",enableOncoKBandHotspots);    
+        obj.put("oncoprintOncoKbHotspotsDefault",enableOncoKBandHotspots);
```
onursumer: strange that this line shows up as modified

inodb: trailing spaces removed

inodb:
@onursumer maybe good to add docs as well about this new property at https://docs.cbioportal.org/2.3-customization/portal.properties-reference & maybe the `properties.EXAMPLE`. Not sure if that is addressed somewhere as well. We can follow up later

alisman:
@inodb @onursumer we shouldn't document it until we merge frontend that uses it.

onursumer:
@inodb I think `oncokb.token` is already added into `portal.properties.EXAMPLE`, but not sure if it has been documented yet. 

This specific boolean property is derived from `oncokb.token` and only intended for front end use.

inodb:
@alisman As you can see - doubting if this should be in the prototype section or hidden completely ☝️. In some ways it's useful for ourselves to skim-read the release notes whenever we find a bug. When it doesn't show up there you'll have to go through the commits itself which are much harder to read. So was thinking if we make the section in the release notes like

# Prototype (Internal use only)

Then perhaps it's clear enough for people deploying that they can ignore it

PR #7040, opened 2020-01-23
Github action to test performance of genie private portal

mandawilson:
- Tests genie-private POST /api/mutated-genes/fetch endpoint
- Fails if:
  1) curl returns a non-zero exit code
  2) request returns a response code other than 200
- On failure or success sends a message to slack
- Runs at 3am EST

NOTE this test will initially fail until genie private performance improves.

We plan to add more performance tests like this in the future.

### DIFF
--- a/.github/workflows/genie-performance-test.yml
+++ b/.github/workflows/genie-performance-test.yml
@@ -1,0 +1,44 @@
+name: genie-performance-test
+
+on:
+  schedule:
+    - cron:  '0 8 * * *'
+
+jobs:
+  probe-genie-endpoint:
+    
+    name: Genie private endpoint probe
+
+    runs-on: ubuntu-latest
+    
+    env:
+      SLACK_URL_SECRET: ${{ secrets.SLACK_URL_SECRET }}
+      SLACK_CHANNEL: backend-performance-status 
+
+    steps:
+    - name: GET request to genie private page
+      env:
+        GENIE_DATA_ACCESS_TOKEN_SECRET: ${{ secrets.GENIE_DATA_ACCESS_TOKEN_SECRET }}
+        MAX_WAIT_TIME_IN_SECONDS: 300
+        URL: "https://genie-private.cbioportal.org/api/mutated-genes/fetch"
+        OK_RESPONSE_CODE: 200
+      # GET request to URL with a MAX_WAIT_TIME_IN_SECONDS second limit
+      #   if curl fails, automatically exit with curl's non-zero exit code
+      #   else if we did not get a OK_RESPONSE_CODE exit with a non-zero exit code
+      #   else we are OK!
+      run: |
+        curl --max-time $MAX_WAIT_TIME_IN_SECONDS -s -o /dev/null -w "%{http_code}" -X POST -H "Authorization:Bearer $GENIE_DATA_ACCESS_TOKEN_SECRET" "$URL" -H "accept:application/json" -H "Content-Type:application/json" -d "{\"studyIds\":[\"genie_private\"]}" > curl.out
+        RESPONSE_CODE=`cat curl.out`
+        if [ "$RESPONSE_CODE" != "$OK_RESPONSE_CODE" ]; then
+          echo "Request failed with response code '$RESPONSE_CODE', expected '$OK_RESPONSE_CODE'"
+          exit 1
+        else
+          echo "Request succeeded with expected response code '$OK_RESPONSE_CODE'"
+          exit 0
+        fi
+    - name: Failure
+      if: failure()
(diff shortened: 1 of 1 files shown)


.github/workflows/genie-performance-test.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  probe-genie-endpoint:
+    
+    name: Genie private endpoint probe
+
+    runs-on: ubuntu-latest
+    
+    env:
+      SLACK_URL_SECRET: ${{ secrets.SLACK_URL_SECRET }}
+      SLACK_CHANNEL: backend-performance-status
```
inodb: nice! so cool! I love that it's easy to see in code where to sign up for this alert. One question: is this channel private? I can't seem to join - maybe ok if this channel is public?

.github/workflows/genie-performance-test.yml
```diff
@@ -0,0 +1,44 @@
...
+        GENIE_DATA_ACCESS_TOKEN_SECRET: ${{ secrets.GENIE_DATA_ACCESS_TOKEN_SECRET }}
+        MAX_WAIT_TIME_IN_SECONDS: 300
+        URL: "https://genie-private.cbioportal.org/api/mutated-genes/fetch"
+        OK_RESPONSE_CODE: 200
+      # GET request to URL with a MAX_WAIT_TIME_IN_SECONDS second limit
+      #   if curl fails, automatically exit with curl's non-zero exit code
+      #   else if we did not get a OK_RESPONSE_CODE exit with a non-zero exit code
+      #   else we are OK!
+      run: |
+        curl --max-time $MAX_WAIT_TIME_IN_SECONDS -s -o /dev/null -w "%{http_code}" -X POST -H "Authorization:Bearer $GENIE_DATA_ACCESS_TOKEN_SECRET" "$URL" -H "accept:application/json" -H "Content-Type:application/json" -d "{\"studyIds\":[\"genie_private\"]}" > curl.out
```
inodb: @mandawilson Nice! Would it be possible to also log the time the response took somewhere? Maybe it can be sent both to stdout and to the slack channel? That way even if it maybe doesn't go over the threshold we would have some idea of whether response time has changed over time.

mandawilson:
@inodb We like the reporting the response time idea, I will try to add it tomorrow.

.github/workflows/genie-performance-test.yml
```diff
@@ -0,0 +1,44 @@
...
+    
+    env:
+      SLACK_URL_SECRET: ${{ secrets.SLACK_URL_SECRET }}
+      SLACK_CHANNEL: backend-performance-status 
+
+    steps:
+    - name: GET request to genie private page
+      env:
+        GENIE_DATA_ACCESS_TOKEN_SECRET: ${{ secrets.GENIE_DATA_ACCESS_TOKEN_SECRET }}
+        MAX_WAIT_TIME_IN_SECONDS: 300
```
sheridancbio: Maybe this is ok just to get the test in place ... but I think the max wait time in seconds for the mutated-genes fetch should be something like 5-10 seconds

.github/workflows/genie-performance-test.yml
```diff
@@ -0,0 +1,44 @@
...
+        URL: "https://genie-private.cbioportal.org/api/mutated-genes/fetch"
+        OK_RESPONSE_CODE: 200
+      # GET request to URL with a MAX_WAIT_TIME_IN_SECONDS second limit
+      #   if curl fails, automatically exit with curl's non-zero exit code
+      #   else if we did not get a OK_RESPONSE_CODE exit with a non-zero exit code
+      #   else we are OK!
+      run: |
+        curl --max-time $MAX_WAIT_TIME_IN_SECONDS -s -o /dev/null -w "%{http_code}" -X POST -H "Authorization:Bearer $GENIE_DATA_ACCESS_TOKEN_SECRET" "$URL" -H "accept:application/json" -H "Content-Type:application/json" -d "{\"studyIds\":[\"genie_private\"]}" > curl.out
+        RESPONSE_CODE=`cat curl.out`
+        if [ "$RESPONSE_CODE" != "$OK_RESPONSE_CODE" ]; then
```
sheridancbio: maybe add : rm curl.out  (or better : rm $CURL_OUTPUT_FILENAME)
I'm not sure where this is running though --- if it is an ephemeral circle-ci kind of machine, maybe cleanup is unnecessary.

inodb: it's indeed ephemeral so cleanup is not really necessary

inodb:
@mandawilson i think we should merge this, right? Been working ok as far as I know

PR #7082, opened 2020-01-30
Fixed OncoKB proxy return type, and updated default URL

onursumer:
- Fix the incorrect response content type for the OncoKB proxy.
- Update the default URL to the latest OncoKB API endpoint.

# Checks
- [ ] Runs on heroku

### DIFF
--- a/web/src/main/java/org/cbioportal/proxy/ProxyController.java
+++ b/web/src/main/java/org/cbioportal/proxy/ProxyController.java
@@ -18,12 +18,14 @@ import java.nio.charset.StandardCharsets;
 
 @RestController
 public class ProxyController {
+    private static final String DEFAULT_ONCOKB_URL = "https://www.oncokb.org/api/v1";
+    
     @Value("${show.oncokb:true}")
     private Boolean showOncokb;
     @Value("${oncokb.token:}")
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -18,24 +18,32 @@
...
+            buildUri(request.getPathInfo(), request.getQueryString(), false),
+            method,
+            httpHeaders,
+            String.class
+        ).getBody();
     }
 
     @RequestMapping("/oncokb/**")
-    public ResponseEntity<String> proxyOncokb(@RequestBody(required = false) String body, HttpMethod method, HttpServletRequest request)
+    public ResponseEntity<Object> proxyOncokb(@RequestBody(required = false) String body, HttpMethod method, HttpServletRequest request)
```
kalletlak: @onursumer what was the error by using `String` return type?

onursumer: @kalletlak content is attached to `response.text` instead of `response.body` if we use `String`, and that causes OncoKB frontend API client to always return an empty response.

PR #7088, opened 2020-01-31
Performance optimization for ProfiledCasesCounter

### LINKED ISSUE #7044: GENIE performance issue
GENIE portal is really slow currently.

- [ ] identify the baseline performance
- [ ] optimize sql
- [ ] other optimization

### LINKED ISSUE #7072: Optimization of StudyViewController & MutationServiceImpl
GOAL: try to recover the performance of study view page when we first enabled Ehcache ... page renders in seconds.

As part of the effort to improve GENIE study view performance, we will optimize code recently introduced behind the /mutated-genes/fetch endpoint.

In particular, the sorting/iteration code within [StudyViewController.fetchMutatedGenes](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/StudyViewController.java#L255)

Note, this also affects
[StudyViewController.fetchFusion](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/StudyViewController.java#L292)
[StudyViewController.fetchCNAGenes](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/StudyViewController.java#L329)

and additional logic doing data manipulation within [MutationServiceImpl.getSampleCountInMultipleMolecularProfiles](https://github.com/cBioPortal/cbioportal/blob/master/service/src/main/java/org/cbioportal/service/impl/MutationServiceImpl.java#L111)

averyniceday:
Fix #7072  (see https://help.github.com/en/articles/closing-issues-using-keywords)

Left some print statements in there in case we want to do additional optimizations. Will remove after discussing whether or not this can be merged without additional code changes. 

Issue:
- Study Summary Page took ~3-5 minutes to run when loading for a large study (e.g GENIE)
- Profiled code and identified segment inside the ProfiledCasesCounter that processes data AFTER retrieving records from DB (meaning current EHCache placement will not help improve performance). 
- This segment of code took around 90-120 seconds to run.
- Portion of code was recalculating cases for gene panels.

- Cache/store gene panel cases and reuse instead of recalculating
- Other cleanup related to creating maps to store reused objects

Involved endpoint is the `mutated-genes/fetch` endpoint. Current runtime is now ~5 seconds w/ EHCache turned on. Can test login and test [here](https://dashi-dev.cbio.mskcc.org:8080/security-test/api/mutated-genes/fetch) w/ `genie_private` if you have permissions.
# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
+++ b/service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
@@ -50,6 +50,7 @@ public class ProfiledCasesCounter {
                 // there can be duplicate patient or sample id, append study id
                 .map(x -> x.getStudyId() + (countByPatients ? x.getPatientId() : x.getSampleId()))
                 .collect(Collectors.toSet());
+        int profiledCasesCount = profiledCases.size(); 
 
         Set<String> casesWithoutPanelData = profiled
                 .stream()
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -281,7 +281,6 @@
                 });
             }
         }
-
```
ao508: let's revert this change here @averyniceday

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
```diff
@@ -58,32 +59,42 @@ public void calculate(List<String> molecularProfileIds, List<String> sampleIds,
...
-                    allPanels.add(genePanel);
-                });
-                
-                profiledCasesForGene.addAll(casesWithoutPanelData);
+                        genePanelToCases.put(genePanel.getStableId(), casesWithPanelData);
+                    }
+                }
+                // for every gene panel associated containing the gene, use the sum of unique cases
+                // as well as cases without panel data
+                for (GenePanel genePanel : geneGenePanelMap.get(entrezGeneId)) {
```
sheridancbio: I think this loop could be combined with the previous loop (the iteration is the same), although I am not positive. If the three inner lines were appended to the inner lines of the previous loop would the data still be valid?
Looping twice is perhaps not too much overhead, if it is helpful for code clarity.

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
```diff
@@ -58,32 +59,42 @@ public void calculate(List<String> molecularProfileIds, List<String> sampleIds,
...
             List<GenePanel> allPanels = new ArrayList<>();
             
+            Set<String> totalProfiledCases = new HashSet<String>();
+            Set<String> allMatchingGenePanelIds = new HashSet<String>();
+            // different calculations depending on if gene is linked to gene panels 
             if (geneGenePanelMap.containsKey(entrezGeneId)) {
-                geneGenePanelMap.get(entrezGeneId).forEach(genePanel -> {
-                    Set<String> casesWithPanelData = genePanelDataMap
+                // calculate and store cases for each gene panel
+                for (GenePanel genePanel : geneGenePanelMap.get(entrezGeneId)) {
```
kalletlak: Updated `genePanelToCases` can be completely outside of block iterating `alterationCounts` (line 63).

may be
```
        Map<String, Set<String>> genePanelToCases = genePanelDataMap
                .entrySet().stream()
                .collect(Collectors
                        .toMap(
                                e -> e.getKey(),
                                e -> {
                                    return e.getValue()
                                            .stream()
                                            .map(x -> x.getStudyId() + (countByPatients ? x.getPatientId() : x.getSampleId()))
                                            .collect(Collectors.toSet());
                                    }
                                ));
```

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
```diff
@@ -58,32 +59,42 @@ public void calculate(List<String> molecularProfileIds, List<String> sampleIds,
                 .map(x -> x.getStudyId() + (countByPatients ? x.getPatientId() : x.getSampleId()))
                 .collect(Collectors.toSet());
 
+        
+        Map<String, Set<String>> genePanelToCases = new HashMap<String, Set<String>>();
+        
         for (AlterationCountByGene alterationCountByGene : alterationCounts) {
             final Set<String> profiledCasesForGene = new HashSet<String>();
```
kalletlak: never used. cleanup

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
```diff
@@ -58,32 +59,42 @@ public void calculate(List<String> molecularProfileIds, List<String> sampleIds,
...
             Integer entrezGeneId = alterationCountByGene.getEntrezGeneId();
             List<GenePanel> allPanels = new ArrayList<>();
```
kalletlak: never used. cleanup

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
sheridancbio: I'm still trying to intuitively understand these constructs.
My interpretation of your suggested code:
GenePanelData records are linkages between a sample/patient/profile/gene_panel/study and a boolean flag for whether or not the sample was "profiled". I don't fully understand how a sample would be part of a genetic profile and yet not be profiled ... although I guess I know that there are some samples which are part of a mutation profile and which were profiled but for which no mutation events were found. I still think there is some constraint in this collection of objects which I do not understand.
Anyway, what we fetch via fetchGenePanelDataInMultipleMolecularProfiles is a big aggregate of all of these membership records, and then we group these into buckets based on the gene panel id. So the map genePanelDataMap has one bucket of sample memberships per panel. These can span across profiles and studies.
So the suggested code will loop through every entry in genePanelDataMap (each panel) and insert a mapping from the panel_id to a set of "sample/study" identifier strings, which will be constructed by looping through each of the membership records which were in the bucket for that gene panel, and aggregating / uniqueifying the cases.
Once this is done, we would not need to check within the loop which examines all genes whether or not the genePanelToCases entry for this gene panel needs to be computed or not --- because we have precomputed all of them.

Ok, I think I understand.

inodb:
@averyniceday @sheridancbio @ao508 @kalletlak super nice work! I just tested this on the public genie study and even there that endpoint went from ~30s to ~2-3s. Sweet!

PR #7141, opened 2020-02-10
Add option to remove studies giving study Id

oplantalech:
This PR extends the functionality of removing studies from cBioPortal with the option to delete by passing one or multiple study Ids.

### DIFF
--- a/core/src/main/scripts/importer/cbioportalImporter.py
+++ b/core/src/main/scripts/importer/cbioportalImporter.py
@@ -371,7 +371,7 @@ def usage():
                            '--command [%s] --study_directory <path to directory> '
                            '--meta_filename <path to metafile>'
                            '--data_filename <path to datafile>'
-                           '--study_ids <cancer study ids, comma separated>'
+                           '--study_ids <cancer study ids for remove-study command, comma separated>'
                            '--properties-filename <path to properties file> ' % (COMMANDS)), file=OUTPUT_FILE)
 
 def check_args(command):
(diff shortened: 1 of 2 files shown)


oplantalech:
@inodb @sheridancbio I have added you as a reviewers since I discussed this issue with you last Friday, feel free to appoint somebody else as a reviewer if necessary

core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -357,6 +371,7 @@ def usage():
                            '--command [%s] --study_directory <path to directory> '
                            '--meta_filename <path to metafile>'
                            '--data_filename <path to datafile>'
+                           '--study_ids <cancer study ids, comma separated>'
```
sheridancbio: maybe we should mention to the user that this is used in conjunction with the remove-study command? Otherwise some may think this must be supplied during an import.

inodb: second that. Really `command` should be a proper subcommand instead of an option. Each command should have it's own options

inodb: I created an issue for this: https://github.com/cBioPortal/cbioportal/issues/7201. It would be good tho to mention like Rob said that this is only for the remove-study command

sheridancbio: Refactoring this script so that options are context-sensitive to the command could be an issue with the "good first issue" tag. But we also are hoping to replace the importer and scripts package entirely with a standalone code base (this year?)

inodb: Thanks for the context @sheridancbio ! Since it would be a pretty minor change I think it can be ok for a first issue even tho we might decide to retire the code altogether, but that's def good to know!

sheridancbio: My suggestion would be to change this line to:
'--study_ids <cancer study ids for remove-study, comma separated>'

pvannierop: @inodb @sheridancbio I have updated the user help lines in the script as per your suggestion.

sheridancbio:
I believe all requested changes have been made. Merging now.

PR #7149, opened 2020-02-11
Init commit of GitHub actions dev branch. 

ao508:
Added actions.yml file.
Updated dependencies in Dockerfile that are compatible with GitHub `ubuntu-latest` runner.

### DIFF
--- a/docker/web-and-data/Dockerfile
+++ b/docker/web-and-data/Dockerfile
@@ -42,10 +42,11 @@ FROM openjdk:8-jre
 # download system dependencies first to take advantage of docker caching
 RUN apt-get update; apt-get install -y --no-install-recommends \
         build-essential \
-        libmysql-java \
+        default-mysql-client \
         default-libmysqlclient-dev \
         python3 \
         python3-setuptools \
+        libpython-dev \
         python3-dev \
         python3-pip \
 	&& rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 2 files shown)


docker/web-and-data/Dockerfile
```diff
@@ -42,10 +42,11 @@ FROM openjdk:8-jre
 # download system dependencies first to take advantage of docker caching
 RUN apt-get update; apt-get install -y --no-install-recommends \
         build-essential \
-        libmysql-java \
+        default-mysql-client \
```
inodb: Thanks so much for sending this PR! I needed this one line, saved me some work debugging why the docker build was failing 🙏

docker/web-and-data/Dockerfile
```diff
@@ -64,7 +65,8 @@ COPY --from=build /cbioportal/requirements.txt /cbioportal/requirements.txt
 
 # install build and runtime dependencies
 # ignore update failures
-RUN  pip3 install -r /cbioportal/requirements.txt
+RUN pip3 install awscli
```
inodb: I guess this is still a draft PR but just trying to avoid potential issues early: why add `awscli` to the cBioPortal image?

Usually one tries to keep the images as small as possible with one specific purpose. The container running `awscli` should prolly be separate.

To add some context. This image was already too bloated and a good example of what not to do :). The importer should prolly have a separate docker image as well as the validator.

ao508: @inodb yeah I'm not sure this is going to stay actually, mostly just figuring out how to get the github actions runner to cooperate. I dont think that this `awscli` is necessary but for some reason the `pip3 install -r /cbioportal/requirements` complained about this package missing. I imagine it might be a different issue just being masked as this error.

PR #7166, opened 2020-02-13
Increase heading levels to fix FAQ header bugs

### LINKED ISSUE #6186: faq headers are lost
The highest level headers in the FAQ hierarchy appear in the TOC at the top, but not on the page below.

This is particularly an issue for the last entry in the TOC (what if I have other questions?):
image

The text for the answer to this gets lost in the answer to the previous question, because the question itself never appears on the page:
image

ecerami:
# Fix #6186

The bug in #6186 is due to the fact that our default stylesheet hides h1 tags on markdown pages.  This hiding is necessary to prevent two headers from appearing on each markdown page (one from React and one from Markdown).

To fix the issue, I just increased the heading level throughout the FAQ markdown page.  We still have one h1 tag at the top (which is hidden), but after that every h1 tag is now h2, every h2 tag is now h3, etc.  With that, all the headers now appear and are also clickable.

I verified that this fixes the issue by deploying the markdown to surge at:  http://naive-ornament.surge.sh/FAQ.md, and then temporarily changing my local front end to point to surge.  With that change, I was able to verify that these markdown changes do indeed fix the issue.

# Any screenshots or GIFs?

If you go to:  https://www.cbioportal.org/faq, and scroll to the bottom of the page, you will currently see:
screenshot_215

Note that the header text:  "What if I have other questions or comments?" does not appear.

With the revised markdown, you now see missing header/text!

screenshot_217

# Notify reviewers
@inodb @alisman or @jjgao

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -78,108 +78,108 @@
         * [What are the values of the box and whiskers in a boxplot?](#what-are-the-values-of-the-box-and-whiskers-in-a-boxplot)
 * [What if I have other questions or comments?](#what-if-i-have-other-questions-or-comments)
 
-# General Questions
-### What is the cBioPortal for Cancer Genomics?
+## General Questions
+#### What is the cBioPortal for Cancer Genomics?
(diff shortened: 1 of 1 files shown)


alisman:
@ecerami i tried out your idea just manipulating the DOM using browser debugger and it worked as you say.

PR #7173, opened 2020-02-14
Do not return ResponseEntity

zhx828:
Return the body as String.
The suspicion is that the Object.class cannot convert the data properly


web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -1,5 +1,7 @@
 package org.cbioportal.proxy;
 
+import com.mongodb.util.JSON;
+import org.json.simple.JSONObject;
```
onursumer: we don't need these anymore, right?

PR #7197, opened 2020-02-18
Allow the validator to pass for NUMERIC attributes starting with > and < symbols

rmadupuri:
- Numeric clinical attributes for example Age sometimes are given as a range. Currently since the validator fails on these, they are imported as STRINGs to maintain the PHI. Since the front-end accepts these symbols for Numeric attributes we should also let the validator pass on these.

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -2461,7 +2461,7 @@ class ClinicalValidator(Validator):
                 pass
             elif data_type == 'NUMBER':
                 if not self.checkFloat(value):
-                	if value.startswith('>') or value.startswith('<'):
+                	if (value[0] in ('>','<')) and value[1:].isdigit():
                 		pass
                 	else:
                 		self.logger.error(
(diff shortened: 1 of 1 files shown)


core/src/main/scripts/importer/validateData.py
```diff
@@ -2461,7 +2461,10 @@ def checkLine(self, data):
                 pass
             elif data_type == 'NUMBER':
                 if not self.checkFloat(value):
-                    self.logger.error(
+                	if (value[0] in ('>','<')) and value[1:].isdigit():
+                		pass
+                	else:
+                		self.logger.error(
```
inodb: @zhx828 @kalletlak I think this is prolly ok to merge, but maybe we should create a new issue to indicate that that there should be a NUMBER_RAGE data_type or something along those lines. What do you think?

also how is this handled on the study view for barcharts like e.g.:

Screen Shot 2020-02-19 at 6 11 13 PM

inodb: so @rmadupuri validated that this works by importing some studies like this to triage. But yeah still curious what y'all think @zhx828 @kalletlak

zhx828: @inodb this kind of data has been in the portal for quite while now. SV handles it by treating it as the max/min value then group the rest of data together. I'm ok with NUMBER_RANGE, but no sure how frequent it will be used though.

PR #7215, opened 2020-02-24
zscore doc update: two profiles

yichaoS:
Updating the zscore documentation to:
- clean up
- add new profile (all samples as base population)

### DIFF
--- a/docs/Z-Score-normalization-script.md
+++ b/docs/Z-Score-normalization-script.md
@@ -1,8 +1,8 @@
 ## Introduction
-For some data types, when uploading to cBioPortal, it is currently necessary to also provide a z-score transformed version of your input file. The z-score data is essential for the oncoprint functionality. The oncoprint shows over- and underexpression of the data, based on the threshold the user sets when selecting the genomic profile. 
+For some data types, when uploading to cBioPortal, it is currently necessary to also provide a z-score transformed version of your input file. The z-score data is essential for the oncoprint functionality. The oncoprint shows high or low mRNA expression of the genes, based on the threshold the user sets when selecting the genomic profile. 
 
 :warning: 
-Please keep in mind that the z-scores are calculated using only patient data. Hence, overexpressed in this case implies higher expression than the average patient. Also, the source data on which the z-score data is based does not necessarily follow the normal distribution. If your data does not follow the normal distribution, the z-score threshold is less reliable and will result in more false positives or false negatives. You can consider log transforming your value before calculating z-scores to improve this. However, even the logged values may not follow the normal distribution, especially if the data is bimodal.
+Please keep in mind that the z-scores are calculated using only patient data. Hence, 'mRNA High' in this case implies higher expression than the average patient. Also, the source data on which the z-score data is based does not necessarily follow the normal distribution. If your data does not follow the normal distribution, the z-score threshold is less reliable and will result in more false positives or false negatives. You can consider log transforming your value before calculating z-scores to improve this. However, even the logged values may not follow the normal distribution, especially if the data is bimodal.
 
 Currently, the cBioPortal team is woking on a solution that will allow the user to decide whether they want to use zscores, or log values or zscore of log values in the frontend. Furthermore, the team is also working on functionality to be able to use healthy samples in the calculations. 
 
@@ -144,5 +144,5 @@ python NormalizeExpressionLevels_allsampleref.py -i <expression_file> -o <output
 #### Example:
 
 ```
-python NormalizeExpressionLevels.py -i path/to/your/study/data_expression_median.txt -o path/to/your/study/data_mRNA_median_Zscores.txt -l
+python NormalizeExpressionLevels_allsampleref.py -i path/to/your/study/data_expression_median.txt -o path/to/your/study/data_mRNA_median_Zscores.txt -l
 ```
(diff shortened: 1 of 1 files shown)


jjgao:
@yichaoS are we going to push out the zscores relative to all samples soon?

rmadupuri:
@jjgao yes, we have added the new profiles for provisional and pancan studies to datahub. We will be importing soon. 
For the remaining studies, we wanted to confirm from the publications that the profiles are not log-transformed already (36 studies).

docs/Z-Score-normalization-script.md
```diff
@@ -5,31 +5,54 @@ For some data types, when uploading to cBioPortal, it is currently necessary to
...
 
 ```
 (r - mu)/sigma
 ```
 
-where `r` is the raw expression value, and `mu` and `sigma` are the mean and standard deviation of the unaltered distribution, respectively.
+where `r` is the raw expression value, and `mu` and `sigma` are the mean and standard deviation of the base population, respectively.
+
+### How to proceed
+cBioPortal expects z-score normalization to take place per gene. You can produce this extra z-score file yourself, or let cBioPortal do it for your input files by using `convertExpressionZscores.pl` (Method 1, diploid samples as base population) OR `NormalizeExpressionLevels.py` (Method 2, all samples as base population). Examples of the calculation and running the programs are below.
```
tmazor: I worry that this wording might confuse a user into thinking that z-scores will be generated as part of the import process. How about rephrasing the second sentence to:
You can calculate z-scores with your own preferred method, or use one of the cBioPortal provided approaches:
- `convertExpressionZscores.pl` applies Method 1 (diploid samples as base population)
- `NormalizeExpressionLevels.py` applies Method 2 (all samples as base population)

docs/Z-Score-normalization-script.md
```diff
@@ -5,31 +5,54 @@ For some data types, when uploading to cBioPortal, it is currently necessary to
...
 
 Currently, the cBioPortal team is woking on a solution that will allow the user to decide whether they want to use zscores, or log values or zscore of log values in the frontend. Furthermore, the team is also working on functionality to be able to use healthy samples in the calculations. 
- 
-### How to proceed
-cBioPortal expects z-score normalization to take place per gene. You can produce this extra z-score file yourself, or let cBioPortal do it for your input files by using `convertExpressionZscores.pl`. Most information that now follows was taken from the comments in `convertExpressionZscores.pl`. We've also added an example of the calculation and example of running the program below.
+
 
 ## The cBioPortal convertExpressionZscores method
```
tmazor: Given the edits in this PR, this header should probably change to something more generic like "Calculating Z-scores"

docs/Z-Score-normalization-script.md
```diff
@@ -5,31 +5,54 @@ For some data types, when uploading to cBioPortal, it is currently necessary to
...
-## convertExpressionZscores Transformation Algorithm 
+### `NormalizeExpressionLevels.py`
+
+`-i <expression_file> -o <output_file>`
+
+- `<expression_file>` : the [expression (exp) data file](File-Formats.md#expression-data).
+- `<output_file>` : the output file to be generated
+- We require the mRNA expression data to be log transformed before calculating zscores on them. If the data is not already log tranformed use the `-l` option to calculate log base 2 of the expression values.
+
+## Transformation Algorithm
```
tmazor: This header confused me at first. Maybe call it something like "Algorithms & Implementation"? Also you have both algorithm & implementation sections for convertExpressionZscores but only one for NormalizeExpressionLevels

rmadupuri:
Branch rebased to master and squashed.

docs/Z-Score-normalization-script.md
rmadupuri: Hi @tmazor, thank you for the review. This is a good catch. We have updated and reorganized the doc a little bit. Do you mind taking another quick look? Thank you!

docs/Z-Score-normalization-script.md
rmadupuri: We now have a common header `The cBioPortal Z-Score calculation method` that introduces the two methods and each method is then explained in detail.

docs/Z-Score-normalization-script.md
rmadupuri: Merged algorithm and implementation sections.

docs/Z-Score-normalization-script.md
```diff
@@ -5,22 +5,32 @@ For some data types, when uploading to cBioPortal, it is currently necessary to
 Please keep in mind that the z-scores are calculated using only patient data. Hence, overexpressed in this case implies higher expression than the average patient. Also, the source data on which the z-score data is based does not necessarily follow the normal distribution. If your data does not follow the normal distribution, the z-score threshold is less reliable and will result in more false positives or false negatives. You can consider log transforming your value before calculating z-scores to improve this. However, even the logged values may not follow the normal distribution, especially if the data is bimodal.
```
tmazor: Sorry I didn't catch this the first time, but in OncoPrint we have shifted to using "high" and "low" rather than "over/under-expressed" (see: http://bit.ly/2wnevrA). It might be good to use that same terminology in this document.

docs/Z-Score-normalization-script.md
tmazor: The reorganization is a great improvement! I did add one new comment about using high/low rather than over/under-expressed, but besides that, this looks great!

docs/Z-Score-normalization-script.md
rmadupuri: Thanks @tmazor. Updated the introduction to use high/low.

rmadupuri:
@jjgao can we merge this?

PR #7216, opened 2020-02-24
add mutation exclusion to OQL documentation

tmazor:


adamabeshouse:
looks good, thanks! can you also add that you can exclude all mutations by position as well? e.g. `MUT != V600`

tmazor:
@adamabeshouse updated!

adamabeshouse:
thanks!

PR #7259, opened 2020-03-03
updating database schema for resource feature

### LINKED ISSUE #7234: Update DB schema to handle resources
Add following tables:

**resource_definition**
RESOURCE_ID   
DISPLAY_NAME  
DESCRIPTION (optional) 
RESOURCE_TYPE (STUDY | PATIENT | SAMPLE) 
OPEN_BY_DEFAULT (boolean)
PRIORITY (defines order in table/in tabs)
CANCER_STUDY_ID

**resource_sample**
ID (auto_increment)
INTERNAL_ID
RESOURCE_ID
URL

**resource_patient**
ID (auto_increment)
INTERNAL_ID
RESOURCE_ID
URL

**resource_study**
ID (auto_increment)
INTERNAL_ID
RESOURCE_ID
URL

dippindots:
Fix #7234


db-scripts/src/main/resources/cgds.sql
```diff
@@ -865,5 +869,46 @@ CREATE TABLE `info` (
   `DB_SCHEMA_VERSION` varchar(24),
   `GENESET_VERSION` varchar(24)
 );
+
+-- --------------------------------------------------------
+CREATE TABLE `resource_definition` (
+  `RESOURCE_ID` varchar(255) NOT NULL,
+  `DISPLAY_NAME` varchar(255) NOT NULL,
+  `DESCRIPTION` varchar(2048) NOT NULL,
+  `RESOURCE_TYPE` varchar(255) NOT NULL COMMENT 'STUDY, PATIENT or SAMPLE',
```
kalletlak: may be declare it as enum
`RESOURCE_TYPE ENUM('STUDY', 'PATIENT', 'SAMPLE') NOT NULL,`

db-scripts/src/main/resources/cgds.sql
```diff
@@ -865,5 +869,46 @@ CREATE TABLE `info` (
...
+  `OPEN_BY_DEFAULT` BOOLEAN NOT NULL,
+  `PRIORITY` int(11) NOT NULL,
+  `CANCER_STUDY_ID` int(11) NOT NULL,
+  PRIMARY KEY (`RESOURCE_ID`,`CANCER_STUDY_ID`),
```
kalletlak: @adamabeshouse just want to confirm that there won't be a duplicate `RESOURCE_ID` for different types within a study

kalletlak:
@adamabeshouse @dippindots should we create feature branch for this as there are going to be more pull requests for the same feature

adamabeshouse:
one thing I just thought of @dippindots I think that display name should be optional - in that case we would use the URL

db-scripts/src/main/resources/cgds.sql
```diff
@@ -865,5 +869,46 @@ CREATE TABLE `info` (
...
   `GENESET_VERSION` varchar(24)
 );
+
+-- --------------------------------------------------------
+CREATE TABLE `resource_definition` (
+  `RESOURCE_ID` varchar(255) NOT NULL,
+  `DISPLAY_NAME` varchar(255) NOT NULL,
+  `DESCRIPTION` varchar(2048) NOT NULL,
+  `RESOURCE_TYPE` varchar(255) NOT NULL COMMENT 'STUDY, PATIENT or SAMPLE',
+  `OPEN_BY_DEFAULT` BOOLEAN NOT NULL,
```
kalletlak: set default value to false(0)

dippindots:
@kalletlak @adamabeshouse Can I get your review again on the updated schema, thanks!

db-scripts/src/main/resources/cgds.sql
adamabeshouse: correct, that's what I think it should be. thats how it works with clinical attributes, right? (clinical attr, study) is unique? @kalletlak

db-scripts/src/main/resources/migration.sql
```diff
@@ -812,3 +812,41 @@ WHERE NOT EXISTS (SELECT * FROM genetic_profile WHERE  STABLE_ID=CONCAT(`fusion_
     AND CANCER_STUDY_ID = `fusion_studies`.CANCER_STUDY_ID);
 DROP TEMPORARY TABLE fusion_studies;
 UPDATE `info` SET `DB_SCHEMA_VERSION`="2.12.3";
+
+##version: 2.12.4
+CREATE TABLE `resource_definition` (
+  `RESOURCE_ID` varchar(255) NOT NULL,
+  `DISPLAY_NAME` varchar(255) DEFAULT NULL,
+  `DESCRIPTION` varchar(2048) NOT NULL,
```
kalletlak: @dippindots @adamabeshouse any specific reason why `DISPLAY_NAME` could be null and `DESCRIPTION` not to be?

db-scripts/src/main/resources/cgds.sql
kalletlak: yes

kalletlak:
> one thing I just thought of @dippindots I think that display name should be optional - in that case the frontend would use the URL as the name

I think `DESCRIPTION` should even be optional

adamabeshouse:
I agree with @kalletlak , lets make description optional as well

adamabeshouse:
I'm sorry but on further development I realize that we need the display name to be non null because it would be the name of the open tab @kalletlak @dippindots

dippindots:
@adamabeshouse No problem! I will update it.

db-scripts/src/main/resources/cgds.sql
```diff
@@ -865,5 +869,52 @@ CREATE TABLE `info` (
...
+  `OPEN_BY_DEFAULT` BOOLEAN DEFAULT 0,
+  `PRIORITY` int(11) NOT NULL,
+  `CANCER_STUDY_ID` int(11) NOT NULL,
+  PRIMARY KEY (`RESOURCE_ID`,`CANCER_STUDY_ID`),
+  FOREIGN KEY (`CANCER_STUDY_ID`) REFERENCES `cancer_study` (`CANCER_STUDY_ID`) ON DELETE CASCADE
+);
+
+-- --------------------------------------------------------
+CREATE TABLE `resource_sample` (
+  `ID` INT(11) NOT NULL auto_increment,
```
kalletlak: Since this `ID` is never used at all, why. not make (`INTERNAL_ID`, `RESOURCE_ID`, `URL`) as primary key?

adamabeshouse: technially someone could have a duplicate entry though

kalletlak: I think it should be detected and blocked while importing

dippindots: Thanks @kalletlak @adamabeshouse. Updated!

PR #7270, opened 2020-03-05
Resources importer

### LINKED ISSUE #7235: Write importer code for resources
- [x] update database schema: https://github.com/cbioportal/cbioportal/issues/7234
- [x] update DaoCancerStudy to remove values when deleting a study
- [x] create Daos of Resources and tests
- [x] prepare sample files in the study_es_0
- [x] create resources importer and tests
- [x] data validator
- [x] tests
- [x] update `File-Formats.md`

dippindots:
Fix #7235  

Reference [RFC:  Links to resources](https://docs.google.com/document/d/1_Re7ThLGYcyzvpEH9eMrdi-vbOx_He5Txkgt_L0onGA/edit#)
- add new file format `Resource` and `Resource Definition` (meta, data)
- add ImportResourcedata.java and ImportResourceDefinition.java to import resource data and resource definition data
- validator for both resource data and resource definition data
- File-Format add a new `Resource Data` format


adamabeshouse:
I think we need to do the data format differently, the way it is in the RFC. Sorry I didn't make this more clear, that it's not going to be exactly the same as clinical. 

JJ recommended this data format in the RFC:

image

We need to do it this way because, for example, a sample might have more than one resource data for one resource definition, but only one data for a different definition, and there's no way to represent that in the clinical data input format. 

Let me know if that makes sense. Sorry again for the misunderstanding.

dippindots:
@adamabeshouse Thanks for clarifying this, sorry I didn't notice the difference between them. But now the differences I see are:
1. in the resource table (sample/patient/study), we should not use RESOURCE_ID and INTERNAL_ID as the primary key, change schema like this (take `resource_sample` as an example):
```
CREATE TABLE `resource_sample` (
  `ID` INT(11) NOT NULL auto_increment,
  `INTERNAL_ID` int(11) NOT NULL,
  `RESOURCE_ID` varchar(255) NOT NULL,
  `URL` varchar(255) NOT NULL,
  UNIQUE (`INTERNAL_ID`, `RESOURCE_ID`, `URL`),
  PRIMARY KEY (`ID`),
  FOREIGN KEY (`INTERNAL_ID`) REFERENCES `sample` (`INTERNAL_ID`) ON DELETE CASCADE
);
```
2. Is is ok that we add the resource_defination data into the header of each type of resource file like what we do in clinical_attribute (add to data_resource_sample.txt / data_resource_patient.txt / data_resource_study.txt)? 

If we just need multiple resources mapping to a single pair of (INTERNAL_ID and RESOURCE_ID). I think maybe we can still use a similar implementation and change some validation.

adamabeshouse:
1. good catch. yes I don't know about how thats handled but it sounds like you've got it. 
2. I think @jjgao had indicated he thought we should have a different file, like the first table in the RFC screenshot above

image

Because the same format as in clinical doesnt work in the same way with the data table:

image

What do you think? @gaofe

docs/File-Formats.md
```diff
@@ -1375,3 +1376,88 @@ Example with 2 generic entities and 3 samples:
 <tr><td>mut_sig_mean</td><td>name of mean</td><td>description of mean</td><td>0.370</td><td>0.010</td><td>0.005</td></tr>
 <tr><td>mut_sig_confidence</td><td>name of confidence</td><td>description of confidence</td><td>0.653</td><td>0.066</td><td>0.050</td></tr>
 </table>
+
+## Resource Data
+
+The resource data is used to capture resouce data in patients, samples and studies.
```
inodb: would be good to elaborate a bit more what resources are e.g. similar to the RFC:

```
Resources are URIs that can point to:

Files: pdf, txt, png, json, etc.
Web links: non-file links e.g. URLs to other systems
```

docs/File-Formats.md
```diff
@@ -1375,3 +1376,88 @@ Example with 2 generic entities and 3 samples:
 <tr><td>mut_sig_mean</td><td>name of mean</td><td>description of mean</td><td>0.370</td><td>0.010</td><td>0.005</td></tr>
 <tr><td>mut_sig_confidence</td><td>name of confidence</td><td>description of confidence</td><td>0.653</td><td>0.066</td><td>0.050</td></tr>
 </table>
+
+## Resource Data
+
+The resource data is used to capture resouce data in patients, samples and studies.  
+
+the resource file is split into a resource difinition file, sample resource file, patient resource file and study resource file. All data files are required to have a matching meta file.
```
inodb: good to run a spellchecker e.g. `difinition` -> `definition`

inodb: a command line checker is `aspell` https://www.howtoforge.com/linux-aspell-command/

inodb: we could maybe add that one as part of CI at some point ☝️

core/src/test/scripts/test_data/study_es_0/result_report.html
```diff
@@ -1279,6 +1279,218 @@ <h4 class="panel-title">data_mutations_extended.maf</h4>
         </div>
       </div>
 
+      <div class="panel panel-success">
```
inodb: still a bit weird i think that the whole report is checked in as part of testing - we should prolly test specific things inside of the report after parsing it. But this works for now

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,618 @@
...
+
+    private File resourceDataFile;
+    private CancerStudy cancerStudy;
+    private ResourceTypes resourceType;
+    private boolean relaxed;
+    private Set<String> patientIds = new HashSet<String>();
+
+    public static enum MissingResourceValues {
+        NOT_APPLICABLE("Not Applicable"), NOT_AVAILABLE("Not Available"), PENDING("Pending"),
+        DISCREPANCY("Discrepancy"), COMPLETED("Completed"), NULL("null"), MISSING(""), NA("NA");
```
inodb: are these missing values different from some other enum already in the code? This seems like something that's not resource specific

docs/File-Formats.md
dippindots: This is pretty cool, good to know this!

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
dippindots: No, I borrow the same definition from the `Clinical Data`. I can import it from that class also.

dippindots:
Thanks @inodb , I will try to address them soon.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
inodb: yeah there can maybe be some shared class with common N/A values. I do think some of these don't really look like N/A to me, e.g. `Discrepancy`, `Pending` and `Completed`. That seems very specific to some pipeline that we have set up

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,618 @@
...
+        stablePatientId = StableIdUtil.getPatientId(stablePatientId);
+        int resourceIdIndex = findResourceIdColumn(headerNames);
+        int urlIndex = findURLColumn(headerNames);
+        int internalSampleId = -1;
+        int internalPatientId = -1;
+
+        // create resource_id set
+        Set<String> patientResourceIdSet = resources.stream().filter(resource -> resource.getResourceType().equals(ResourceDefinition.PATIENT_RESOURCE_TYPE)).map(resource -> resource.getResourceId()).collect(Collectors.toSet());
+        Set<String> sampleResourceIdSet = resources.stream().filter(resource -> resource.getResourceType().equals(ResourceDefinition.SAMPLE_RESOURCE_TYPE)).map(resource -> resource.getResourceId()).collect(Collectors.toSet());
+        Set<String> studyResourceIdSet = resources.stream().filter(resource -> resource.getResourceType().equals(ResourceDefinition.STUDY_RESOURCE_TYPE)).map(resource -> resource.getResourceId()).collect(Collectors.toSet());
```
onursumer: I would break these lines into multiple lines

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,618 @@
...
+                throw new RuntimeException("Error: Sample " + stableSampleId
+                        + " was previously linked to another patient, and not to " + stablePatientId);
+            }
+            numSamplesProcessed++;
+        }
+
+        // if the resource id or url matches one of the missing values, skip this resource:
+        if ((resourceIdIndex != -1 && MissingResourceValues.has(fields[resourceIdIndex])) || (urlIndex != -1 && MissingResourceValues.has(fields[urlIndex]))) {
+            numEmptyResourcesSkipped++;
+        } else {
```
onursumer: this `else` block seems to complex to review

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
dippindots: Yeah, that would be better.

dippindots:
@onursumer Yes, we still use Daos to import data. We are using MyBatis in repositories.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
dippindots: @onursumer I updated the solution, could you take a look if that's better?

dippindots:
@inodb @onursumer Thanks for reviewing my pr. I addressed your comments, could you take a look again, thanks!

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
dippindots: @inodb I haven't found common null values in the project, so I created one `ValuesUtils.java`

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
onursumer: @dippindots looks better, thanks!

core/src/main/scripts/importer/validateData.py
```diff
@@ -3502,7 +3502,7 @@ class ResourceDefinitionValidator(Validator):
     # 'RESOURCE_ID', 'RESOURCE_TYPE', 'DISPLAY_NAME' are required
     REQUIRE_COLUMN_ORDER = False
     REQUIRED_HEADERS = ['RESOURCE_ID', 'RESOURCE_TYPE', 'DISPLAY_NAME']
-    NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]", "[completed]", "[null]", "", "na"]
+    NULL_VALUES = ["[not applicable]", "[not available]", "[null]", "", "na"]
```
inodb: @dippindots i imagine `pending/discrepancy/completed` were prolly there for a reason. Did someone confirm they were no longer necessary? Otherwise I would leave them. I mostly meant that when writing the resource importer/validator code u prolly don't want to include them, since they are kind of weird NULL_VALUES. Like if I were to load data I would be surprised that these type of values are considered NULL. For backwards compatibility you would like to leave them but when introducing new code I would leave them out. 

So maybe you have something like `NULL_VALUES` that make sense for all kinds of data and `OTHER_CLINICAL_ATTRIBUTE_NULL_VALUES` which are in addition to the NULL_VALUES and specific to clinical attributes only

Anyway this is not super important but just wanted to clarify. You can have a single NULL_VALUES list but then to be safe I would include the previous values `"[pending]", "[discrepancy]", "[completed]"`

inodb:
@dippindots Thanks for updating! Looks good! One minor comment about removing existing NULL VALUES

core/src/main/scripts/importer/validateData.py
dippindots: Hi @inodb, I don't think I totally understand what you mean. I am not sure whether we want users to have NULL_VALUES in their data file?

dippindots: Do you mean that we can remove the current NULL_VALUES definition?

dippindots:
@inodb Thanks again! I updated the solution!

core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -219,6 +219,8 @@ def process_directory(jvm_args, study_directory):
     study_meta_dictionary = {}
     cancer_type_filepairs = []
     sample_attr_filepair = None
+    sample_resource_filepair = None
+    resource_defination_filepair = None
```
onursumer: it should be `definition` not `defination` :)

dippindots: I do need a spell checker!!

dippindots: updated!

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    private void validateAddDatum(int internalId, String stableId, String resourceId, String resourceURL, String resourceType, Set<String> resourceSet, MultiKeyMap resourceMap) throws Exception {
+        // throw exception if resource definition is not exist in the database
+        if (!resourceSet.contains(resourceId)) {
+            throw new RuntimeException("Error: " + resourceType.toLowerCase() + " " + stableId
+            + " with resource " + resourceId
+            + " does not have matching resources information in database, please make sure to include resource definition in the resource definition file");
+        }
+        // The resourceMap makes sure a pari of (internalId/resource_id/url) is unique
```
ao508: typo `pari` --> `pair`

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceDefinition.java
```diff
@@ -0,0 +1,330 @@
...
+            if (descriptionIndex != -1 && isValueNotMissing(fieldValues[descriptionIndex])) {
+                description = fieldValues[descriptionIndex];
+            }
+            // get resourceType (must be value of ResourceTypes)
+            if (resourceTypeIndex != -1 && isValidResourceType(fieldValues[resourceTypeIndex])) {
+                resourceType = fieldValues[resourceTypeIndex];
+            }
+            // get openByDefault (optional)
+            if (openByDefaultIndex != -1 && isValidOpenByDefault(fieldValues[openByDefaultIndex])) {
+                openByDefault = fieldValues[openByDefaultIndex].equalsIgnoreCase("true");
```
ao508: here you just want to parse the boolean value right? and if it's not "true" or "false" then the default would be false?

the `Boolean` java docs describes a built in method for parsing a boolean from a string and I think it's the exact behavior you're looking for too

https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html

core/src/main/java/org/mskcc/cbio/portal/util/ValuesUtils.java
```diff
@@ -0,0 +1,36 @@
+package org.mskcc.cbio.portal.util;
+
+public class ValuesUtils {
+    public static enum MissingValues {
+        NOT_APPLICABLE("Not Applicable"), NOT_AVAILABLE("Not Available"), PENDING("Pending"),
```
ao508: for consistency would you mind putting these enum types on their own separate lines?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceDefinition.java
```diff
@@ -0,0 +1,330 @@
...
+
+    private int findAndValidateOpenByDefaultColumn(String[] headerNames) {
+        return findAndValidateColumnIndexInHeaders(OPEN_BY_DEFAULT_COLUMN_NAME, headerNames);
+    }
+
+    private int findAndValidatePriorityColumn(String[] headerNames) {
+        return findAndValidateColumnIndexInHeaders(PRIORITY_COLUMN_NAME, headerNames);
+    }
+
+    private int findAndValidateColumnIndexInHeaders(String columnHeader, String[] headerNames) {
```
kalletlak: could we send `Map<String,Integer>` of `headerNames` instead of array? checking in map is more efficient than iterating an array

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceDefinition.java
dippindots: Thanks! Good to know this!

core/src/main/java/org/mskcc/cbio/portal/dao/DaoResourceData.java
```diff
@@ -0,0 +1,219 @@
...
+
+import java.sql.*;
+import java.util.*;
+
+/**
+ * Data Access Object for `resource` tables
+ */
+public final class DaoResourceData {
+
+    public static final String SAMPLE_TABLE = "resource_sample";
```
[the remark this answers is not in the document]
dippindots: I will rename them to `RESOURCE_SAMPLE_TABLE` and `RESOURCE_PATIENT_TABLE`

core/src/main/java/org/mskcc/cbio/portal/model/ResourceDefinition.java
```diff
@@ -0,0 +1,93 @@
+package org.mskcc.cbio.portal.model;
+
+public class ResourceDefinition {
```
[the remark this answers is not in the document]
dippindots: Yes, definitely, I will rebase and use that one.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,574 @@
...
+                .map(resource -> resource.getResourceId())
+                .collect(Collectors.toSet());
+        Set<String> studyResourceIdSet = resources
+                .stream()
+                .filter(resource -> resource.getResourceType().equals(ResourceDefinition.STUDY_RESOURCE_TYPE))
+                .map(resource -> resource.getResourceId())
+                .collect(Collectors.toSet());
+
+        // check if sample is not already added:
+        Sample sample = DaoSample.getSampleByCancerStudyAndSampleId(cancerStudy.getInternalId(), stableSampleId, false);
```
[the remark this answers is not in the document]
dippindots: I don't think I can fetch data by resource type since the sample and patient can have no association with resources before we adding the resource data into the database.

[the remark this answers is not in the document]
kalletlak: I mean to say that for a study resource you don't need to fetch Sample. and similarly for patient resource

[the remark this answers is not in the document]
dippindots: I see! I will update this!

dippindots:
Hi, @kalletlak @ao508 I have updated my pr and addressed all the comments, could I get your review again? Thank you!

core/src/main/java/org/mskcc/cbio/portal/util/ValuesUtils.java
```diff
@@ -0,0 +1,41 @@
+package org.mskcc.cbio.portal.util;
+
+public class ValuesUtils {
+    public static enum MissingValues {
```
kalletlak: why not MissingValues enum directly instead of ValuesUtils

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceDefinition.java
```diff
@@ -0,0 +1,337 @@
...
+        buff.close();
+
+        if (MySQLbulkLoader.isBulkLoad()) {
+            MySQLbulkLoader.flushAll();
+            MySQLbulkLoader.relaxedModeOff();
+        }
+    }
+
+    private String[] splitFields(String line) throws IOException {
+        String[] fields = line.split(DELIMITER, -1);
```
kalletlak:
```suggestion
       return line.split(DELIMITER, -1);
```

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceDefinition.java
```diff
@@ -0,0 +1,337 @@
...
+
+        // now iterate over lines and trim each value:
+        for (int i = 0; i < fieldValues.length; i++) {
+            fieldValues[i] = fieldValues[i].trim();
+        }
+        return fieldValues;
+    }
+
+    private Boolean isValueNotMissing(String value) {
+        if (MissingValues.has(value)) {
```
kalletlak:
```suggestion
        return !MissingValues.has(value);
```

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,559 @@
...
+    private int findURLColumn(Map<String, Integer> headerIndexMap) {
+        return findColumnIndexInHeaders(URL_COLUMN_NAME, headerIndexMap);
+    }
+
+    private int findSampleTypeColumn(Map<String, Integer> headerIndexMap) {
+        return findColumnIndexInHeaders(SAMPLE_TYPE_COLUMN_NAME, headerIndexMap);
+    }
+
+    private int findColumnIndexInHeaders(String columnHeader, Map<String, Integer> headerIndexMap) {
+        if (headerIndexMap.containsKey(columnHeader)) {
```
kalletlak: may be
```suggestion
        return headerIndexMap.getOrDefault(columnHeader, -1);
```

model/src/main/java/org/cbioportal/model/ResourceType.java
```diff
@@ -1,7 +1,17 @@
 package org.cbioportal.model;
 
 public enum ResourceType {
-    STUDY,
-    SAMPLE,
-    PATIENT
+    STUDY("STUDY"),
```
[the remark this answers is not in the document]
dippindots: I want to get the string value from the `ResourceType`, I didn't have a way to do that, so I overrode this. 
Is there a way to get the value of the `ResourceType`? For example https://github.com/cBioPortal/cbioportal/pull/7270/files#diff-5041ba5290d6be4231695bde8b85883bR28.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoResourceData.java
```diff
@@ -0,0 +1,214 @@
...
+
+            return toReturn;
+        } catch (SQLException e) {
+            throw new DaoException(e);
+        } finally {
+            JdbcUtil.closeAll(DaoResourceData.class, con, pstmt, rs);
+        }
+    }
+
+    public static ResourceBaseData getDatum(String cancerStudyId, String patientId, String resourceId) throws DaoException {
```
kalletlak: looks like this method is never used

model/src/main/java/org/cbioportal/model/ResourceType.java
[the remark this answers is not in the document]
kalletlak: there is `name()` to get string value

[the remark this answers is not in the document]
dippindots: Thanks! Got it!

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,555 @@
...
+                patient = new Patient(cancerStudy, patientId);
+                internalPatientId = DaoPatient.addPatient(patient);
+            }
+        }
+        return internalPatientId;
+    }
+
+    private int addSampleToDatabase(String sampleId, String[] fields, Map<String, Integer> headerIndexMap)
+            throws Exception {
+        int sampleTypeIndex = findSampleTypeColumn(headerIndexMap);
```
kalletlak: Are we going to have `sampleType`(`SAMPLE_TYPE` column) in data file?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,555 @@
...
+            if (sampleType != null && sampleType.isNormal()) {
+                return internalSampleId;
+            }
+            String stablePatientId = getStablePatientId(sampleId, fields, headerIndexMap);
+            if (validPatientId(stablePatientId)) {
+                Patient patient = DaoPatient.getPatientByCancerStudyAndPatientId(cancerStudy.getInternalId(),
+                        stablePatientId);
+                if (patient == null) {
+                    addPatientToDatabase(stablePatientId);
+                    patient = DaoPatient.getPatientByCancerStudyAndPatientId(cancerStudy.getInternalId(),
```
kalletlak: `addPatientToDatabase` already return patient internal id. do we still need this query?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
```diff
@@ -0,0 +1,555 @@
...
+            // want to try and capture normal sample types based on value for SAMPLE_TYPE
+            // if present in resource data
+            if (sampleType != null && sampleType.isNormal()) {
+                return internalSampleId;
+            }
+            String stablePatientId = getStablePatientId(sampleId, fields, headerIndexMap);
+            if (validPatientId(stablePatientId)) {
+                Patient patient = DaoPatient.getPatientByCancerStudyAndPatientId(cancerStudy.getInternalId(),
+                        stablePatientId);
+                if (patient == null) {
```
kalletlak: I think `addSampleToDatabase` should be executed only when patient is added. I prefer passing patient object in the method parameter. this would reduce the number of queries made to database to fetch patient(`getPatientByCancerStudyAndPatientId`)

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportResourceData.java
dippindots: Yes, IMO, if someone introduces a new sample in the resource data file, `SAMPLE_TYPE` can be an optional column in the data file.

PR #7272, opened 2020-03-06
Create stale.yml

jjgao:
Install [Stale App](https://github.com/apps/stale) to clean up issues.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,19 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - critical
+  - priority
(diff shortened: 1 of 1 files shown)


inodb:
@jjgao I like the idea of using the stale app! we had some discussions about using this before, but we didn't want to use it then because many issues we have that are open are still relevant even though we are not actively working on them. Could you elaborate why we want to give it a try now? Just give it a shot?

jjgao:
Just wanted to give it a try. At the time when a issue is labeled as stale/wontfix, we should take some action if it's still relevant.

.github/stale.yml
```diff
@@ -0,0 +1,19 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - critical
+  - priority
```
inodb: maybe we should add Epic here as well?

PR #7300, opened 2020-03-16
Add news pathwaymapper

inodb:
The pathways tab has been out for almost two weeks, this adds the news item

### DIFF
--- a/docs/News.md
+++ b/docs/News.md
@@ -1,3 +1,8 @@
+## March 3, 2020
+* **New Feature**: Add Pathways tab to the _Results View_ page which visualizes alteration frequencies of genes in pathways of interest. The pathways are pulled from https://www.pathwaymapper.org. One can add new pathways and edit existing ones there. For more information see the [tutorial](https://www.cbioportal.org/tutorials#pathways).
+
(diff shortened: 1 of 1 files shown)


ugurdogrusoz:
How about a little rewording like this:
"The pathways are pulled from https://www.pathwaymapper.org and shown in a read only view. One can edit these pathways in the PathwayMapper editor as well."
As it is people might think that they can actually add pathways of their own to the pathway list available in cBioPortal.

inodb:
@ugurdogrusoz good point! Updated with a minor change to wording (removed `as well`)

Curious: what's the process for adding new pathways that would show up in cBioPortal?

ugurdogrusoz:
@inodb Pathways in PathwayMapper were previously drawn and saved on **server side**. The PathwayMapper editor allows one to create and save pathways locally and upload them for use later but the cBioPortal edition (viewer) can currently only display these pre-defined pathways on server side. If we get a student accepted for GSoC, we should see how we can accommodate users to make use of their own pathways within cBioPortal as well.

inodb:
Thanks @ugurdogrusoz !

PR #7330, opened 2020-03-25
Customize citation field

oplantalech:
Added a new property into config_service for controlling the citation field on frontend


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -57,6 +57,9 @@ skin.right_nav.whats_new_blurb=<p> &bull;<a href="news.jsp"> <b>New data and fea
 # setting controlling the blurb
 skin.blurb=The cBioPortal for Cancer Genomics provides <b>visualization</b>, <b>analysis</b> and <b>download</b> of large-scale cancer genomics data sets.  <p>Please adhere to <u><a href="http://cancergenome.nih.gov/abouttcga/policies/publicationguidelines"> the TCGA publication guidelines</a></u> when using TCGA data in your publications.</p> <p><b>Please cite</b> <a href="http://www.ncbi.nlm.nih.gov/pubmed/23550210">Gao et al. <i>Sci. Signal.</i> 2013</a> &amp;  <a href="http://cancerdiscovery.aacrjournals.org/content/2/5/401.abstract">Cerami et al. <i>Cancer Discov.</i> 2012</a> when publishing results based on cBioPortal.</p>
 
+# setting controlling the citation below the blurb
+skin.citation_rule_text=Please cite: <a href="http://cancerdiscovery.aacrjournals.org/content/2/5/401.abstract" target="_blank">Cerami et al., 2012</a> &amp; <a href="http://www.ncbi.nlm.nih.gov/pubmed/23550210" target="_blank">Gao et al., 2013</a>
```
alisman: i think we should comment this out so that the default will be in frontend and this will only be set here when user truly intends to do so

oplantalech: @alisman fixed!

Issue #7327, opened 2020-03-24
Support GRCh38 in mutation mapper tool


leexgh:
Sub-issue of https://github.com/genome-nexus/genome-nexus/issues/306
Trying to support GRCh38 in mutation mapper. Make tweaks to frontend to allow using either hg19 or hg38
- [x] frontend changes on support GRCh38
- [x] add the new GRCh38 property to the backend
- [x] add tests
- [x] figure out what to do when GRCh38 instance is not defined (there is default now, you can use the `show.mutation_mapper_tool.grch38` config to disable showing grch38

PR #7348, opened 2020-04-01
add genome nexus grch38 to config

leexgh:
Fix backend https://github.com/cBioPortal/cbioportal/issues/7327

add genome nexus grch38 to config

### DIFF
--- a/docs/portal.properties-Reference.md
+++ b/docs/portal.properties-Reference.md
@@ -233,6 +233,25 @@ The CIViC API url is set to https://civic.genome.wustl.edu/api/ by default. It c
 civic.url=
 ```
 
+# Genome Nexus Integration
+Genome Nexus provides annotations of mutations in cBioPortal. The mutations tab relies heavily on the Genome Nexus API, therefore that tab won't work well without it. By default cBioPortal will use the public Genome Nexus API, such that no extra configuration is necessary.
+
+## Genome Build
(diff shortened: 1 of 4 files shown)


docs/portal.properties-Reference.md
```diff
@@ -379,4 +379,18 @@ For more information on how Ehcache is implemented in cBioPortal refer to the [C
...
+#### GRCh37
+GRCh37 is the default setting in cBioPortal. Genome Nexus API url is set to https://v1.genomenexus.org/. It can be overridden using the following property:
+```
+genomenexus.url=
+```
+#### GRCh38
+GRCh38 Genome Nexus API is set to https://grch38.genomenexus.org/ by default. It can be overridden using the following property:
+```
+genomenexus.url.grch38=
 ```
```
inodb, line 396: This Genome build section is specifically about Genome Nexus. It looks like for OncoKB and CIVIC we have separate sections for those integrations. I think it would be good to make one for Genome Nexus as well. This PR just made me realize it seems there is none at the moment

The Genome Build section can then be a subheader of this new `Genome Nexus Integration` header smiilar to OncoKB: https://docs.cbioportal.org/2.3-customization/portal.properties-reference#oncokb-integration


Maybe something like this:

```
# Genome Nexus Integration
Genome Nexus provides annotations of mutations in cBioPortal. The mutations tab relies heavily on the Genome Nexus API, therefore that tab won't work well without it. By default cBioPortal will use the public Genome Nexus API, such that no extra configuration is necessary.

## Genome Build
Genome Nexus supports both GRCh37 and GRCh38, but support for the latter is limited. Several annotation sources served by Genome Nexus might not have official GRCh38 support yet i.e. [OncoKB](https://www.oncokb.org/), [CIViC](https://civicdb.org/), [Cancer Hotspots](https://www.cancerhotspots.org/), [My Cancer Genome](https://www.mycancergenome.org/) and [3D structures](https://g2s.genomenexus.org/). Although most of the time the canonical transcript for a gene will be the same between GRCh37 and GRCh38 there might be some that cause issues. In addition the complete integration of cBioPortal with Genome Nexus' GRCh38 is not complete yet. That is cBioPortal currently only connects to one Genome Nexus API by default (the GRCh37 one), so it's not possible to have multiple genome builds in one instance of cBioPortal and get the correct annotations from Genome Nexus for both. Currently only the [mutation mapper tool page](https://www.cbioportal.org/mutation_mapper) is able to handle both.

## Properties
By default the Genome Nexus API url is set to https://v1.genomenexus.org/, which uses GRCh37. It can be overridden using the following property:

genomenexus.url=

The mutation mapper tool page can annotate GRCh38 coordinates. By default it uses https://grch38.genomenexus.org. It can be overridden by setting:

```
genomenexus.url.grch38=
```

PR #7352, opened 2020-04-02
fix resource data validator bug

### LINKED ISSUE #7345: resources validation bug
- [x] the validator should allow patient/sample that has multiple URLs (https://github.com/cBioPortal/cbioportal/blob/release-3.3.0/core/src/main/scripts/importer/validateData.py#L3733)

- [x] the validator should validate that all RESOURCE_ID in (study/sample/patient) data is valid.

Thanks @pieterlukasse for finding these issues! Please also feel free to add more comments or change the current content.

dippindots:
Fix #7345 

Update resource data validator logic 
- data definitions should be included in the data definition file, throw an error if not.
- resource data should not duplicate (e.g. the combination of RESOURCE_ID and URL should be unique)

Update resource data tests:
- check resource definition existing for all types of data.
- check resource duplication for all types of data.


dippindots:
Hi @pieterlukasse, thanks a lot for noticing this issue on the community call! I changed the logic and added some tests for the validator, could you take a look? thanks!

core/src/main/scripts/importer/validateData.py
```diff
@@ -3717,42 +3741,63 @@ def __init__(self, *args, **kwargs):
         """Initialize a PatientResourceValidator with the given parameters."""
         super(PatientResourceValidator, self).__init__(*args, **kwargs)
         self.patient_id_lines = {}
+        self.defined_resources = {}
 
     def checkLine(self, data):
```
inodb, line 3746: normally methods use underscores in python, but i see your following the existing convention in this file so that's fine. Just FYI

core/src/main/scripts/importer/validateData.py
```diff
@@ -3764,6 +3809,44 @@ class StudyResourceValidator(ResourceValidator):
     def __init__(self, *args, **kwargs):
         """Initialize a StudyResourceValidator with the given parameters."""
         super(StudyResourceValidator, self).__init__(*args, **kwargs)
+        self.defined_resources = {}
+
+    def checkLine(self, data):
```
inodb, line 3814: normally methods use underscores in python, but i see your following the existing convention in this file so that's fine. Just FYI

core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -2566,7 +2566,7 @@ def _resetMultipleFileHandlerClassVars():
         c.prior_validated_feature_ids = None
         c.prior_validated_header = None
 
-# --------------------------- resource definition wise test ------------------------------
+# --------------------------- resources wise test ------------------------------
```
inodb, line 2569: i looked a bit in this file - im not entirely sure why these tests are often called `wise` test and not e.g. ResourceTest. Again this is following existing convention so prolly fine, but that wording strikes me as a bit odd

dippindots:
Thanks @inodb, I merged it.

PR #7374, opened 2020-04-09
Add github action for dev docker images

inodb:
Move from building on docker hub to building using github actions (hopefully faster, cause it's in parallel)


inodb:
gonna close this for now until i am more certain this works well for genome nexus

PR #7399, opened 2020-04-14
Allow specifying the entityBaseURL.

holtgrewe:
- allow specifying `entityBaseUrl` for `MetadataGenerator`; this appears to be required when deploying cBioPortal behind a reverse proxy that does the SSL handling.
- setting this to `#{null}` by default to keep the old behaviour

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/security/security-spring/src/main/resources/applicationContext-security.xml
+++ b/security/security-spring/src/main/resources/applicationContext-security.xml
@@ -349,6 +349,7 @@
         <b:constructor-arg>
             <b:bean class="org.springframework.security.saml.metadata.MetadataGenerator">
                 <b:property name="entityId" value="${saml.sp.metadata.entityid}"/>
+                <b:property name="entityBaseURL" value="${saml.sp.metadata.entitybaseurl}"/>
                 <b:property name="extendedMetadata">
                     <b:bean class="org.springframework.security.saml.metadata.ExtendedMetadata">
                         <b:property name="idpDiscoveryEnabled" value="true"/>
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -100,6 +100,7 @@ microsoftlive.consumer.secret=
 
 ## SAML settings
 saml.sp.metadata.entityid=
+# change this url if behind reverse proxy that handles SSL, see docs/Authenticating-Users-via-SAML.md
 saml.sp.metadata.entityBaseURL=#{null}
 saml.idp.metadata.location=
 saml.idp.metadata.entityid=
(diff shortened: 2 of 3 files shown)


holtgrewe:
@n1zea144 could you review this? You last edited the area around my change in `applicationProperties-security.xml`.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -100,6 +100,7 @@ microsoftlive.consumer.secret=
 
 ## SAML settings
 saml.sp.metadata.entityid=
+saml.sp.metadata.entityBaseURL=#{null}
```
inodb, line 103: @holtgrewe Thanks for the PR! Why should this one be `#{null}`? Is that the same as empty string? Like why should this value have to be different from the other saml example values?

inodb: ah i see never mind, you put it in the description, for keeping the old behavior. Makes sense

inodb: maybe good to put a comment in the example file to clarify? Something like
```suggestion
# change this url if behind reverse proxy that handles SSL, see docs/Authenticating-Users-via-SAML.md
saml.sp.metadata.entityBaseURL=#{null}
```

holtgrewe:
@inodb thanks! Would it be possible to get a Docker container just for this patch? Somehow when I build them locally they do not startup properly (cannot connect to MySQL).

security/security-spring/src/main/resources/applicationContext-security.xml
```diff
@@ -349,6 +349,7 @@
         <b:constructor-arg>
             <b:bean class="org.springframework.security.saml.metadata.MetadataGenerator">
                 <b:property name="entityId" value="${saml.sp.metadata.entityid}"/>
+                <b:property name="entityBaseURL" value="${saml.sp.metadata.entitybaseurl}"/>
```
inodb, line 352: curious, would this btw work?
```suggestion
                <b:property name="entityBaseURL" value="${saml.sp.metadata.entitybaseurl:#{null}}"/>
```

inodb:
@holtgrewe how are you building the container?

This should work from the root dir (best to clone the repo fresh):
```
 docker build . -t cbioportal/cbioportal:3.2.13 -f docker/web-and-data/Dockerfile
```

Make sure you don't have an existing `portal.properties` file before running it (mvn will use the default `portal.properties.EXAMPLE` in that case):

```
rm src/main/resources/portal.properties
```

You can also build directly with Docker (not 100% sure if this command works but this is approx what we use in our e2e tests):

```
docker build https://github.com/holtgrewe/cbioportal.git#ee865928db66a53a4e70a237c851a2a8b6a668b9 -f docker/web-and-data/Dockerfile -t holtgrewe/cbioportal:saml-entity-url
```

We release a new one every week on Thursday (so just missed this one). If that's too long you could also temporarily use the dev image `cbioportal/cbioportal:master` after this is merged.

Issue #7395, opened 2020-04-14
Oncokb token is not accessible if deploy via docker 


leexgh:
Similar to this issue: https://groups.google.com/forum/#!topic/cbioportal/H-5G97YFpX4

Properties are set to:
`show.oncokb=true`
`oncokb.token=real token here`
Got 403 error if deploy with docker:
![image](https://user-images.githubusercontent.com/16869603/79245604-5198cb80-7e46-11ea-8c3e-178dbb592fc5.png)

According to @zhx828, using local tomcat app is fine.

inodb:
@leexgh for the production portal we provide it like this (with `-D` flag to java): https://github.com/knowledgesystems/knowledgesystems-k8s-deployment/blob/master/cbioportal/cbioportal_spring_boot.yaml#L125. Is it only not working when using docker and the token is provided in a `portal.properties` file?

leexgh:
@inodb yes, but I didn't try other approaches

inodb:
@leexgh 👌 we should def fix this one, because we officially recommend people to deploy with docker and using a portal.properties file

inodb:
@leexgh i do wonder if it this is specific to docker? Maybe it doesn't read the variable from `portal.properties` when supplied dynamically at runtime using `PORTAL_HOME` variable. Note that the properties file can be supplied at compile time and dynamically at runtime using `PORTAL_HOME` variable. The latter is what we do in the docker approach. But for production on internal servers we compile a war with a portal.properties file and then deploy it to tomcat. So maybe it's the dynamically reading of portal.properties that is not working

leexgh:
@inodb other variables in `portal.properties` are accessible, for example `show.oncokb` is always fine. I tried to set `show.oncokb=false` and the error was gone, so for me it seems like only `oncokb.token` is not read properly.

PR #7400, opened 2020-04-15
fix oncokb proxy controller

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/7395
oncokb proxy controller should read properties from PORTAL_HOME/portal.properties


onursumer:
Ideally this should be fixed at the spring configuration level, but I don't know how easy it would be to implement.

In general, implementing low level IO operations inside a controller class indicates that something is not right with the architecture/configuration.

inodb:
I'm gonna try to release a new version of cBioPortal today. I'll merge this so that we have a solution for the users that ran into this problem. Filed a new issue for cleaning it up: https://github.com/cBioPortal/cbioportal/issues/7404

PR #7401, opened 2020-04-15
Make metaImport.py use subcommands instead of -c

### LINKED ISSUE #7201: cbioportalImporter.py should use proper sub commands
Currently `--command` is an option to the `cbioportalImporter.py` script. We should change this such that each possible command is a proper subcommand with its own options. E.g. you can do

`cbioportalImporter.py remove-study --help`

`cbioportalImporter.py import-cancer-type --help`

For each of the possible commands: 

https://github.com/cBioPortal/cbioportal/blob/c545a4ad1593f2778bd88d821e3d4363c1ef66cf/core/src/main/scripts/importer/cbioportalImporter.py#L50-L54

mdhvprasad:
Close #7201 

The cbioportalImporter.py script no longer requires a -c flag to run a command. Each command is now a subcommand that can be run with its own flags.

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/core/src/main/scripts/importer/cbioportalImporter.py
+++ b/core/src/main/scripts/importer/cbioportalImporter.py
@@ -403,26 +403,30 @@ def add_parser_args(parser):
                         help='Path to meta file')
     parser.add_argument('-data', '--data_filename', type=str, required=False,
                         help='Path to Data file')
-    parser.add_argument('-id', '--study_ids', type=str, required=False,
-                        help='Cancer Study IDs for `remove-study` command, comma separated')
 
 def interface():
-    parser = argparse.ArgumentParser(description='cBioPortal meta Importer')
-    subparsers = parser.add_subparsers(title='commands', dest='command',
+    parent_parser = argparse.ArgumentParser(description='cBioPortal meta Importer')
+    add_parser_args(parent_parser)
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(title='subcommands', dest='command',
                           help='Command for import. Allowed commands: import-cancer-type, '
                           'import-study, import-study-data, import-case-list or '
                           'remove-study')
-    import_cancer_type = subparsers.add_parser('import-cancer-type')
-    import_study = subparsers.add_parser('import-study')
-    import_study_data = subparsers.add_parser('import-study-data')
-    import_case_list = subparsers.add_parser('import-case-list')
-    remove_study = subparsers.add_parser('remove-study')
-
-    add_parser_args(import_cancer_type)
-    add_parser_args(import_study)
-    add_parser_args(import_study_data)
-    add_parser_args(import_case_list)
-    add_parser_args(remove_study)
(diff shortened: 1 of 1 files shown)


core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -410,6 +405,25 @@ def interface():
                         help='Path to Data file')
     parser.add_argument('-id', '--study_ids', type=str, required=False,
                         help='Cancer Study IDs for `remove-study` command, comma separated')
```
inodb, line 407: this command only applies to remove-study command so i would suggest to only add this to the remove_study subparser

core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -410,6 +405,25 @@ def interface():
                         help='Path to Data file')
```
inodb, line 405: all these commands above are shared among the various subcommands. I think the way to handle that in argparse is to add them to the parent command as described here: https://stackoverflow.com/questions/7498595/python-argparse-add-argument-to-multiple-subparsers

core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -394,12 +394,7 @@ def check_dir(study_directory):
         print('Study cannot be found: ' + study_directory, file=ERROR_FILE)
         sys.exit(2)
 
-def interface():
-    parser = argparse.ArgumentParser(description='cBioPortal meta Importer')
-    parser.add_argument('-c', '--command', type=str, required=False,
```
inodb, line 399: It would be nice to offer backwards compatibility for the `-c` parameter. I.e. if someone supplies that, call the relevant command accordingly. Maybe add in the help text that the recommended way is to use the command directly and that this is only to provide backwards compatibility

mdhvprasad:
Thanks for the quick response @inodb , I've incorporated the changes you asked for. In case you are wondering why I create a parser and a parent_parser, it's to get around a recursion bug that happens when you add a subparser to a parser and set the parent of that subparser to the same parser. Let me know if you have any questions or if anything looks off, thanks!

inodb:
@mdhvprasad Thanks for updating! Looks good!

Could you check if all these work:

- when `-c` is supplied use the command from there (`metaImport.py -c import-cancer-type`)
- when no `-c` is supplied run the command given (`metaImport.py import-cancer-type`)
- when both are supplied throw an error (`metaImport.py -c import-cancer-type import-cancer-type`). Alternatively if argparse already handles this one some other way, that works too, but just curious to know what happens in that case
- error if an incorrect command is given (`metaImport.py nonexistentcommand`)

mdhvprasad:
@inodb I believe 1, 2, and 4 work and I've just added a patch for 3. Let me know if there's anything else you want me to change, or if it's not working for you. Thank you!

PR #7456, opened 2020-04-29
Publish tags to docker hub

### LINKED ISSUE #7442: Generate tagged image using github actions
We currently build dev images for various branches here as implemented here: https://github.com/cBioPortal/cbioportal/pull/7410. This is much faster than our current setup of using docker hub. It would be nice to generate tagged images in the same manner to save time. The `master` image would usually already be available after tagging, so really it could add another tag for the existing master image. That might be a bit more complicated, so alternatively one could copy the rules of the current dev image generation (https://github.com/cBioPortal/cbioportal/blob/master/.github/workflows/dockerimage.yml)

inodb:
Fix #7442. Publish docker images for each tag on github

Note that we can't reuse the existing master image as proposed in the issue, because the version in the pom is determined on the fly from git i.e. the latest master version shows a different version in the api/info response than a proper tagged version even though the code might be completely the same otherwise

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -5,6 +5,7 @@ on:
       - master
       - release-*
       - rc
+    tags: '*'
 
 jobs:
   build_and_publish_web_and_data:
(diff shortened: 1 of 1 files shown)


inodb:
localdb testing fails but there were no changes to the code, so gonna merge

PR #7458, opened 2020-04-30
Customize Sentry client to send x-forwarded ip address

alisman:
Customize Sentry client to send x-forwarded for ip address as user address instead of ip address of load balancer


alisman:
Note that we need to enable this on instance with configuration 
sentry.factory=org.cbioportal.logging.SentryClientFactory

PR #7459, opened 2020-04-30
add slides webinar 1

inodb:
We should also add them to the google drive: https://drive.google.com/drive/u/0/folders/0B9KTQJAGhFhhRi1qaTdUWmpLQTA


tmazor:
@inodb great point - webinar slides are now in the google drive tutorial folder

Issue #7344, opened 2020-03-31
survival data migration tool


dippindots:
- [x] write a tool to migrate current survival data files to `0 / 1` format.
- [x] supply mapping file, so other institutions can supply their own

PR #7468, opened 2020-05-05
survival data migration tool

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/7344

Use this tool to convert old format of survival data into new format, there are five arguments:
1. input file path: path to the data file
2. vocabularies file path: path to the vocabularies file, there is one default vocabularies file comes with the migration tool.
2. (optional) override data file: true or false, default is false, will create a new_data_clinical_patient.txt by default
3. (optional) output file path: customize path and filename to generate the new file
4. (optional) additional vocabularies: there might be some vocabularies that are not captured in our observation, additional vocabularies will be added into existing vocabularies.


scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,178 @@
...
+import pandas as pd
+import numpy as np
+import os
+import sys
+import argparse
+import shutil
+
+PARENT_PARSER_NAME = 'survival data migration tool'
+NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]", "[completed]", "[null]", "", "na"]
+ZERO_MAPPING_FILE_PATH = os.getcwd() + '/survivalStatusVocabulariesMappingToZero.txt'
```
n1zea144, line 11: I would combine this into one file with two columns.  If there is a many to one mapping in either direction, I would allow for multiple values in a field.

dippindots: @n1zea144 do we still allow user to provide a single vocabulary that will map to 0 or 1? Or should I put an empty string for not provided value?

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,191 @@
...
+import pandas as pd
+import numpy as np
+import os
+import sys
+import argparse
+import shutil
+
+PARENT_PARSER_NAME = 'survival data migration tool'
+NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]", "[completed]", "[null]", "", "na"]
+MAPPING_FILE_PATH = os.getcwd() + '/survivalStatusVocabularies.txt'
```
n1zea144, line 11: I don't know why I didn't catch this before, but we shouldn't hard code this filename and accept an optional file - just take one file from the command line - a user of this tool can choose to add to it or not.

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,197 @@
...
+import pandas as pd
+import numpy as np
+import os
+import sys
+import argparse
+import shutil
+
+PARENT_PARSER_NAME = 'survival data migration tool'
+NULL_VALUES = ["[not applicable]", "[not available]", "[pending]", "[discrepancy]", "[completed]", "[null]", "", "na"]
+DEFAULT_MAPPING_FILE_PATH = os.getcwd() + '/survivalStatusVocabularies.txt'
```
n1zea144, line 11: I still don't understand why you have to "hardcode" this name into the script.  Why not always require a vocabulary file?

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,196 @@
...
+        return "0:" + value
+    else:
+        sys.exit("cannot find the mapping for vocabulary {} , please provide the the mapping rules with option -a".format(value))
+        return value
+
+def generateVocabularies(vocabulariesMappingToOne, vocabulariesMappingToZero, vocabulariesFile):
+    df = pd.read_csv(vocabulariesFile, sep='\t')
+    oneValues = df[MAPPING_TO_ONE_COLUMN].tolist()
+    for unparsedValue in oneValues:
+        for parsedValue in unparsedValue.split(","):
```
rmadupuri, line 29: Thanks for the tool @dippindots! Here are a couple of issues that came up when testing it on files. If one additional vocabulary is passed, say for example `1:LIVE` to OS_STATUS, it is added to survivalStatusVocabularies.txt file as new line `live\tnan`. Running the script the second time, throws an AttributeError on split() for `nan`. Check if value exists before splitting? Also, at line 33.

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,196 @@
...
+    # write files
+    input = open(clinicalFile, 'rb')
+    output = open(newFile, 'wb')
+    tmp_output = open(tmpFile, 'wb')
+
+    writer = csv.writer(output)
+    tmpwritter = csv.writer(tmp_output)
+    for row in csv.reader(input):
+        if row[0].startswith('#'):
+            writer.writerow(row)
```
rmadupuri, line 105: If the script terminates because it couldn't find a value in vocabulary, the intermediate data and tmp file created exists. Should they be deleted? I think its better to save comments to a variable here and write to file at the end?

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,196 @@
...
+        else:
+            tmpwritter.writerow(row)
+    input.close()
+    output.close()
+    tmp_output.close()
+
+    # read file and replace nan
+    df = pd.read_csv(tmpFile, sep='\t')
```
rmadupuri, line 113: Remove the tmp file here after reading to dataframe?

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # find attributes candidates
+    candidates = list(filter(lambda column: column.endswith('_STATUS'), df.columns))
+    targetAttributes = list(filter(lambda candidate: (candidate[:candidate.index("_STATUS")] + '_MONTHS') in df.columns, candidates))
+
+    for targetAttribute in targetAttributes:
+        df[targetAttribute] = df.apply(lambda row: migrateAttibuteValue(row[targetAttribute], vocabulariesMappingToOne, vocabulariesMappingToZero), axis=1)
+
+
+    df.to_csv(newFile, mode='a', index=False, sep='\t', header=True)
```
rmadupuri, line 124: Even if I don't have the survival related attributes, the output is written to a new file. Write both comments and dataframe to output file only if length of `targetAttribute` >=1? else print a message?

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,196 @@
...
+
+    print("vocabularies will be mapping to one: ")
+    print(vocabulariesMappingToOne)
+    print("vocabularies will be mapping to zero: ")
+    print(vocabulariesMappingToZero)
+
+    # write files
+    input = open(clinicalFile, 'rb')
+    output = open(newFile, 'wb')
+    tmp_output = open(tmpFile, 'wb')
```
rmadupuri, line 99: Should the files here be read and written in text mode? Got this error `_csv.Error: iterator should return strings, not bytes (did you open the file in text mode?)` when trying to run the script.

scripts/survival_migration/survivalDataMigration.py
rmadupuri: The new line character seems to be stripped from each line? The header and data records are written to single line (when opened in textwrangler. Seems to be fine in excel or textedit). See the snippet below. 
Screen Shot 2020-05-11 at 11 45 48 AM

scripts/survival_migration/survivalDataMigration.py
```diff
@@ -0,0 +1,196 @@
...
+            print("vocabulary {} has been added to zero mapping".format(zeroUpdateString))
+
+    df.to_csv(vocabulariesFile, index=False, sep='\t', header=True)
+
+def migrate_file(args):
+    clinicalFile = args.clinicalFile
+    vocabulariesFile = args.vocabulariesFile
+    shouldOverrive = args.override
+    newFile = os.path.dirname(args.clinicalFile) + "/new_data_clinical_patients.txt"
+    tmpFile = os.path.dirname(args.clinicalFile) + "/tmp_data_clinical_patients.txt"
```
rmadupuri, line 81: Should the out and tmp file follow the same naming pattern as input file with a suffix of migrated or something like that? Not sure if everyone uses `data_clinical_patients.txt`

scripts/survival_migration/survivalDataMigration.py
dippindots: Since we also want to remove the tmp file and the result file if program exit, should we log this error in console instead of the result file?

scripts/survival_migration/survivalDataMigration.py
rmadupuri: Seems to be my textwrangler issue. Using 4.5.12. Let me upgrade and try again.

dippindots:
close this since migration tool has been merged into datahub: https://github.com/cBioPortal/datahub-study-curation-tools/pull/15

Issue #7505, opened 2020-05-15
study not shown after docker restart container


likelet:
after import my study by following the code from the documentation
```
docker run -it --rm --net cbio-net \
    -v /<path_to_config_file>/portal.properties:/cbioportal/portal.properties:ro \
    -v "<path_to_study_directory>:/study:ro" \
    -v "<path_to_report_folder>:/report" \
    cbioportal/cbioportal:latest \
    metaImport.py -u http://cbioportal-container:8080 -s /study --html=/report/report.html
```
I restart the container by `docker restart ` command. However, the study was not shown in my web page. Did i do something wrong?
thansk for the help 
[report.html.zip](https://github.com/cBioPortal/cbioportal/files/4633215/report.html.zip)

n1zea144:
Hi @likelet I think the importer will not import data if there are warnings without using the -o "override_warning" flag to the metaImport.py script.  Have you check the validation report?  If so, and the warnings are acceptable to you, try using the -o flag.

likelet:
@n1zea144 hi, I add a `-o` parameter in the command, and it does throw an error related to the data format. I will fix the format and let you know once it works.
 BTW, I think this information should be added in the documentation for beginners.  :)

likelet:
by adding  `-o` parameter, it returns the error as follows 
```
#######################################################################
Overriding Warnings. Importing study now
#######################################################################

Data loading step using /cbioportal/scripts/target/scripts-3.3.2-dirty-SNAPSHOT.jar

Checked DB schema version: (expected: 2.12.4) (found: 2.12.4)
Loading cancer types...
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.apache.ibatis.reflection.Reflector (file:/cbioportal/scripts/target/scripts-3.3.2-dirty-SNAPSHOT.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.apache.ibatis.reflection.Reflector
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
--> Loaded 0 new cancer types.
Done.

Warnings / Errors:
-------------------
0.  Cancer type with id 'stad' already exists. Skipping.; 1x

Warnings / Errors:
-------------------
0.  Cancer type with id 'stad' already exists. Skipping.; 1x
Done.
Total time:  6088 ms

WARNING: An illegal reflective access operation has occurred
[...]

PR #7514, opened 2020-05-18
add gene panel command in examlpe_commands.md

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/7505
add gene panel example command in `examlpe_commands.md`.


dippindots:
@inodb I think we also have that in our example commands: https://github.com/dippindots/cbioportal/blob/example-docker-command/docs/docker/example_commands.md#importing-data

Issue #7550, opened 2020-05-28
should log error for populateGenericAssayStableIdToMolecularProfileIdMap in CacheMapUtil


dippindots:
Studies with a record in `genetic_profile` table with type `GENERIC_ASSAY` but without any linked records in the `genetic_alteration` table will resulted in the function being called with an empty list which broke the deployment.

We should throw exceptions when encountered error.

Related functions:
https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java#L102

https://github.com/cBioPortal/cbioportal/blob/master/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenericAssayMyBatisRepository.java#L31

PR #7554, opened 2020-05-29
add checking and add warning in GenericAssayMyBatisRepository

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/7550

- add checking (do not pass in an empty list)
- log error for cases returns an empty list (when a previous function returns an empty list, we stop and log the error)


dippindots:
@averyniceday Thanks a lot! Can we merge this? I saw the `travis-ci` is passing but cannot see the result on github.

PR #7557, opened 2020-06-01
Revert spring saml version

inodb:
Having problems deploying to internal portals. Might work to revert back the spring security saml version


inodb:
I tested this on cbioportal.mskcc.org/beta/ and it seems to have worked

PR #7563, opened 2020-06-01
fix link broken in docker import data document

dippindots:
fix link broken in docker import data document


dippindots:
Not sure why the link will be different:

on the github (**three "-"**): https://github.com/dippindots/cbioportal/blob/update-doc/docs/docker/import_data.md#step-1---import-gene-panels
on the document page (**one "-"**): https://docs.cbioportal.org/2.1.1-deploy-with-docker-recommended/import_data#step-1-import-gene-panels

PR #7610, opened 2020-06-11
Update endpoint name from generic-assay-meta to generic-assay-entity

### LINKED ISSUE #6955: change endpoint path of GENERIC_ASSAY
Change generic endpoints to `/generic-assay-data/` and `/generic-assay-entity/`

------------Previous discussion--------------

Thanks for @inodb brought up this, we want to have a discussion here to re-evaluate the current endpoint path of GENERIC_ASSAY (`/generic_assay_data/` and `/generic_assay_meta/`). 

We think it would be good to avoid using `-data` or `-meta` in the endpoint because everything is data, so the word doesn't convey any extra information. Maybe better to use something like `generic-assays-molecular-data` and `/generic-assays/`? 

Could I get some ideas from you @n1zea144 @pvannierop? Thanks!

dippindots:
Fix #6955


inodb:
Thanks for updating this! I've created a new issue for this #7647. I think we should think a bit more about the naming and the structure before updating. We can re-open this one later

PR #7617, opened 2020-06-15
Accelerate generic assay endpoint response speed

### LINKED ISSUE #7448: Accelerate the response speed of generic assay meta endpoint
This issue has been revealed in this ticket: https://github.com/cBioPortal/cbioportal/issues/7427. We found out that the `/generic_assay_meta/fetch` endpoint is slower than expected, we should find a way to accelerate the response speed of this endpoint.

dippindots:
Fix #7448  
Result: take `ccle_broad_2019` as an example, response time dropped from `9s-10s` to `0.3-0.5s`.

This pr will address the issue that causes the slowness of generic assay endpoint.
The issue is: the mybatis mapper will be called (2 * number of stable ids) times. So the response speed is very slow.

- call mapper just one time
- update related unit tests

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
@@ -18,6 +18,7 @@ import org.cbioportal.model.Sample;
 import org.cbioportal.model.MolecularProfile.MolecularAlterationType;
 import org.cbioportal.model.MolecularProfileSamples;
 import org.cbioportal.model.meta.GenericAssayMeta;
+import org.cbioportal.model.GenericAssayAdditionalProperty;
 import org.cbioportal.persistence.GenericAssayRepository;
 import org.cbioportal.persistence.MolecularDataRepository;
 import org.cbioportal.persistence.SampleListRepository;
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -74,14 +75,19 @@
...
+                    List<GenericAssayAdditionalProperty> filteredAdditionalProperties = additionalProperties
+                            .stream()
+                            .filter(additionalProperty -> additionalProperty.getStableId().equals(stableId))
+                            .collect(Collectors.toList());
+                    for (GenericAssayAdditionalProperty additionalProperty : filteredAdditionalProperties) {
+                        map.put(additionalProperty.getName(), additionalProperty.getValue());
                     }
-                    metaResults.add(new GenericAssayMeta(meta.getEntityType(), meta.getStableId(), map));
+                    meta.setGenericEntityMetaProperties(map);
+                    metaResults.add(meta);
```
dippindots, line 90: Main change is at here: we call function `getGenericAssayAdditionalproperties` once instead of calling N times `getGeneticEntityIdByStableId` and `getGenericAssayMetaPropertiesMap` (N is the number of stable ids).

PR #7622, opened 2020-06-15
Update stale.yml: set daysUntilStale to 180

jjgao:


jjgao:
no worries, it's good to go thru the issue every quarter...

PR #7639, opened 2020-06-19
Auth integration tests

pvannierop:
This is a draft PR for integration test for cBioPortal. Two test types are included:
- security tests that test security layer at Controller level.
- integration level test for saml authentication against live keycloak IDP.
- integration level test against live Google IDP (WIP - requires test Google account)


core/src/test/resources/applicationContext-dao-test.xml
```diff
@@ -0,0 +1,156 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sheridancbio, line 1: same comment about expanding the core package : here we are coupling the core to services provided by the persistence layer

core/src/test/resources/applicationContext-web-test.xml
```diff
@@ -0,0 +1,33 @@
...
+        <property name="ignoreResourceNotFound" value="true" />
+        <property name="locations">
+            <list>
+                <value>file:///#{systemEnvironment['PORTAL_HOME']}/core/src/test/resources/portal.properties</value>
+                <value>file:///#{systemEnvironment['PORTAL_HOME']}/core/src/test/resources/maven.properties</value>
+            </list>
+        </property>
+    </bean>
+
+    <context:component-scan base-package="org.cbioportal.service" />
```
sheridancbio, line 26: We scan the service package for components here. (to make beans available for autowiring in the web config)

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -50,7 +50,7 @@
  *
  * @author Benjamin Gross
  */
-class CancerStudyPermissionEvaluator implements PermissionEvaluator {
+public class CancerStudyPermissionEvaluator implements PermissionEvaluator {
```
sheridancbio, line 53: good .. this should have been done at the beginning of the development of this class.

security/security-spring/src/main/java/org/cbioportal/security/spring/CancerStudyPermissionEvaluator.java
```diff
@@ -220,7 +220,6 @@ private CancerStudy getRelevantCancerStudyFromTarget(Object targetDomainObject)
      * Helper function to determine if given user has access to given cancer study.
      *
      * @param cancerStudy cancer study to check for
-     * @param user Spring Authentication of the logged-in user.
```
sheridancbio, line 223: 👍

web/src/main/java/org/mskcc/cbio/portal/swagger/ApiServiceSwaggerConfig.java
```diff
@@ -17,7 +17,7 @@
 public class ApiServiceSwaggerConfig {
     @Bean
     public Docket api() {
-        return new Docket(DocumentationType.SWAGGER_2)
+        return new Docket(DocumentationType.SWAGGER_2).groupName("legacy")
```
sheridancbio, line 20: Is this a change that has been discussed? I believe elimination of the legacy api controller is desired, so it may be moot how we group it in swagger anyway.  cc: @n1zea144

sheridancbio (commented):
Overall, I think this approach is great. My main concern expressed in the comments is that the tests are developed in the core module (which is a module we are trying to shrink and eliminate). If it is possible to relocate the testing code into a different module or into a new module I would prefer that (so that we don't have extra work dissecting the core package in the future).

I also foresee trouble doing actual google authentication because of the difficulty of registering the appropriate redirect-uri in the google project for cbioportal.

But again, I think this is an excellent approach to security/authorization testing.

jjgao:
@pvannierop what's the next step for this?

pvannierop:
@jjgao I still need to make some changes here. I will discuss with the backend devs whether the feature is still desired.

PR #7659, opened 2020-06-25
Treatments in study view

Luke-Sikina:


web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
```diff
@@ -21,7 +22,9 @@
     private List<String> studyIds;
     private List<ClinicalDataFilter> clinicalDataFilters;
     private List<GeneFilter> geneFilters;
-    private List<List<String>> genomicProfiles;
+    private AndedSampleTreatmentFilters sampleTreatmentFilters;
+    private AndedPatientTreatmentFilters patientTreatmentFilters;
+	private List<List<String>> genomicProfiles;
```
dippindots, line 27: should be consistent (all space).

web/src/main/java/org/cbioportal/web/parameter/filter/SampleTreatmentFilter.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Map;
+
+public class SampleTreatmentFilter {
+    private String treatment;
+    private TemporalRelation time;
+
+    /**
+     * A sample should be included if the treatment row that corresponds
+     * to the treatment and time in this filter contains that sample.
+     * @param treatments key is SampleTreatmentRow::calculateKey
```
dippindots, line 16: Better to add the same comment `* @param sampleId sample.STABLE_ID` as `PatientTreatmentFilter.java`

web/src/test/java/org/cbioportal/web/util/StudyViewFilterApplierTest.java
```diff
@@ -63,6 +63,8 @@
     private SampleListService sampleListService;
     @Mock
     private GenericAssayService genericAssayService;
+    @Mock
+    private TreatmentService treatmentService;
```
dippindots, line 67: Is this necessary? Will you create more test cases for `StudyViewFilterApplierTest`?

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/SampleListMapper.xml
```diff
@@ -32,11 +32,16 @@
...
         <if test="studyIds != null">
-			WHERE cancer_study.CANCER_STUDY_IDENTIFIER IN
-			<foreach item="item" collection="studyIds" open="("
-				separator="," close=")">
-				#{item}
-			</foreach>
+            <if test="studyIds.isEmpty()">
+                WHERE FALSE
+            </if>
+            <if test="!studyIds.isEmpty()">
```
dippindots, line 38: Nice! We have lots of places are require non-empty study id list or non-empty sample/patient list, maybe we can do the same thing for those mapper methods.

web/src/main/java/org/cbioportal/web/parameter/filter/PatientTreatmentFilter.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * A sampleId should be included if the treatment row that corresponds
+     * to the treatment and time in this filter contains that sampleId.
+     * @param sampleId sample.STABLE_ID
+     * @param treatments key is PatientTreatmentRow::calculateKey
+     */
+    public boolean filter(SampleIdentifier sampleId, Map<String, PatientTreatmentRow> treatments) {
+        PatientTreatmentRow row = treatments.get(treatment);
+        return row != null && row.getSamples().contains(sampleId.getSampleId());
```
kalletlak, line 19: I think check for sample in TreatmentRow samples list might be an expensive operation(if there are many samples and/or if there are many treatments in selection). And there can be same sample id for multiple studies. I think you need to check for both sampleId and studyId together.

I propose to change replace sample and studies in both `PatientTreatmentRow` and `SampleTreatmentRow` to a `Map<SampleIdentifier, Boolean>` if not to a List<SampleIdentifier> -> to support duplicate sampleIds.

This goes same with `SampleTreatmentFilter.java`

PR #7663, opened 2020-06-30
update Frequently Asked Questions for import_data.md

### LINKED ISSUE #7606: Update Import data with Docker document
We recently received two questions from users, which could be common issues. We should update [Frequently Asked Questions](https://docs.cbioportal.org/2.1.1-deploy-with-docker-recommended/import_data#frequently-asked-questions).

dippindots:
Fix #7606

### DIFF
--- a/docs/docker/import_data.md
+++ b/docs/docker/import_data.md
@@ -54,7 +54,7 @@ please follow the first step to import gene panels (e.g. import `data_gene_panel
 
 ### Error occurred during validation step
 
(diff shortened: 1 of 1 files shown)


dippindots:
@inodb I just updated the Frequently Asked Questions in `import_data.md`. There were two posts related to this document, but the second one related issue was fixed in our backend, so just added the question from the first post. Can I get your review on this? Thanks!

PR #7670, opened 2020-07-07
Document google analytics profile id config setting

alisman:


docs/Customizing-your-instance-of-cBioPortal.md
```diff
@@ -217,6 +217,19 @@ Below you can find the complete list of all the available skin properties.
             <td>false</td>
             <td>true / false</td>
         </tr>
+      <tr>
+            <td>google_analytics_profile_id</td>
```
alisman, line 221: @inodb it's sad that we have so little consistency in the format of these params (dots, skin, underscores etc).  worth changing now?

jjgao: @alisman I think we should do that separately if we want. And I also think it's ok as long as we document them correctly.

inodb: At this stage maybe not really worth changing manually unless we come up with a better way of enforcing the consistency. It seems to me that the manual review of the introduced properties isn't really working. If we want to keep things consistent we should prolly have some scripts to enforce it. E.g.:

- a script that validates the properties file (maybe this can also help indicate renamed property names)
- check if every property is documented

PR #7681, opened 2020-07-09
Add show.transcript_dropdown to properties

### LINKED ISSUE #7479: Support transcript switch on result view
Have a transcript dropdown next to the lollipop
image

### New design
**Enabling transcript dropdown:**
- [x] query one gene per time instead of query all genes at once
- [x] figure out why mutations are missing if transcript dropdown is enabled
Update: 
Mutation mapper can't handle chromosome 23 and 24, so it causes missing mutations when transcript dropdown is enabled. This issue is fixed in https://github.com/cBioPortal/cbioportal/issues/7504
- [x] add a flag to enable or disable transcript dropdown 

**Mutation table**
- [x] always show fusions
- [x] canonical transcript always uses data from database, non-canonical uses data from GN
- [x] only show mutations belongs to the selected transcript(might need to filter out some mutations) plus `Fusion`
- [x] make sure `Protein Change` and `Mutation Type` are updated at the same time
- [x] update hgvsc column
- [x] If the mutation is categorized as `Other`, show original data in Mutation Type column

**Mutation mapper**

- [x] make sure filter function on mutation count panel works
- [x] split out `Other` to `Other` and ` Fusion`

**Adding selection to urlWrapper**
- [x] adding gene to urlWrapper
- [x] adding transcript id to urlWrapper

TODO
- [x] Fixing protein change with `*` (like `* 280 *`)
- [x] backend PR to add enable_transcript_dropdown to config file

leexgh:
Fix #7479 
Part of the epic, setting `show.transcript_dropdown` to true will add transcript switch dropdown on results view page.

- Add `show.transcript_dropdown` to properties, default is false

### DIFF
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -60,6 +60,7 @@
             "show.civic",
             "show.genomenexus",
             "show.mutation_mappert_tool.grch38",
+            "show.transcript_dropdown",
             "skin.documentation.about",
             "skin.documentation.baseurl",
             "skin.example_study_queries",
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -187,6 +187,9 @@ show.civic=false
 # Link to My Cancer Genome. Please disable (set to false) when using cBioPortal with patient identifiable data due My Cancer Genome license restrictions.
 mycancergenome.show=true
 
+# Enable transcript switch dropdown (true, false)
+# show.transcript_dropdown=false
+
 # igv bam linking
 igv.bam.linking=
 # colon delimited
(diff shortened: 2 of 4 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -187,6 +187,9 @@ show.civic=false
 # Link to My Cancer Genome. Please disable (set to false) when using cBioPortal with patient identifiable data due My Cancer Genome license restrictions.
 mycancergenome.show=true
 
+# Enable transcript switch dropdown (true, false)
+show.transcript_dropdown=false
```
inodb, line 191:
```suggestion
# show.transcript_dropdown=false
```
Good to comment this one out. In case we want to make it enabled by default at some point people don't have to update their properties file

leexgh: @inodb good point, updated

portal/src/main/webapp/config_service.jsp
```diff
@@ -60,6 +60,7 @@
             "show.civic",
             "show.genomenexus",
             "show.mutation_mappert_tool.grch38",
```
inodb, line 62: @leexgh just noticed typo here: `show.mutation_mappert_tool.grch38`. Should be `show.mutation_mapper_tool.grch38`. Can be addressed in separate PR (good to update docs as well)

PR #7695, opened 2020-07-16
Add script that allows for update to existing gene panel 

### LINKED ISSUE #7286: There is no script for updating an existing gene panel.
[The documentation](https://docs.cbioportal.org/5.1-data-loading/data-loading/import-gene-panels#reimport-existing-gene-panel) suggests that I delete the existing gene panel, then reimport the gene panel, then reimport the corresponding studies. That's a lot when you just want to add a gene.

We should make a script that modifies the existing gene panel by adding to the `gene_panel_list` table.

I think this is a good first issue, so I am tagging it as such.

jtquach1:
Fix #7286 

- Update functionality
- Add or remove genes from gene panel depending on list of genes in incoming vs. current panel

# Any screenshots or GIFs?
Before:
sql_before

After: 
sql_after


jtquach1:
These are some sketches I made regarding edge cases in comparing incoming with original genes. 
[updateGenePanel_cases.pdf](https://github.com/cBioPortal/cbioportal/files/4934427/updateGenePanel_cases.pdf)

So far, I've tested intersection case 1 on pages 1 and 2. 
image
image

The first image corresponds to the Before and After in the PR's top comment. For the second image, I tested this using the TESTPANEL2 text file against TESTPANEL2 (unchanged).

jtquach1:
I've just tested cases 1-7, and the only changes I made were for case 1, where having an empty gene list was disallowed for importing. I changed that so case 1 allows the user to delete all genes from the original gene panel when importing an empty gene panel.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -264,63 +258,55 @@ private static void updateGenePanelGeneList(Integer internalId, Set<CanonicalGen
                 return;
             }
 
-            else {
```
jtquach1: moved the conditions out of the else block

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -185,6 +207,35 @@ public UpdateGenePanel(String[] args) {
      */
     public static void main(String[] args) {
         ConsoleRunnable runner = new UpdateGenePanel(args);
-        runner.runInConsole();
+        Scanner s = new Scanner(System.in);
+        boolean confirmed = false;
+
+        System.err.println("WARNING: This script is ONLY for updating a gene panel, not for adding a new gene panel. To add a new gene panel, run the ImportGenePanel script.");
```
jtquach1, line 213: System.err.println() displays red text in Eclipse console. 
System.out.println() displays black text.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -199,114 +197,137 @@ public static void updateGenePanel(GenePanel genePanel, Set<CanonicalGene> canon
...
+
+        // 5. Intersect but neither are supersets of one other
+        else {
+            // Add A - AnB
+            toAdd = diffTwoLists(new ArrayList<>(incoming), new ArrayList<>(AnB));
+            // Remove B - AnB
+            toRemove = diffTwoLists(new ArrayList<>(original), new ArrayList<>(AnB));
+        }
+
+        return new Object[] { toAdd, toRemove };
```
jtquach1, line 254: not sure if I should've used a Pair instead, but the type casting messages to Set<CanonicalGene> don't seem to crash anything

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -34,7 +34,10 @@
 
 import java.sql.*;
 import java.util.*;
+import java.util.stream.Collectors;
```
Luke-Sikina, line 37: The existing methods in this file all interact with the database in some way. Lets move your utility methods into a seperate class in `core/src/main/java/org/mskcc/cbio/portal/util`. Put all methods that don't interact with the database directly there.

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,241 @@
...
+	            	confirmed = true;
+	            	break;
+				default:
+	            	System.err.println("Unrecognized input, try again. (Y/N)");
+	            	break;
+            }
+        }
+    }
+
+    private static String getPropertyValue(String propertyName, Properties properties, boolean noSpaceAllowed) throws IllegalArgumentException {
```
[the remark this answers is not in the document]
jtquach1: The functions I changed from ImportGenePanel were importData and main, so getPropertyValue was part of the original implementation. 
Good to note though!

core/src/test/resources/seed_mini.sql
```diff
@@ -491,6 +491,22 @@ INSERT INTO "sample_profile" ("SAMPLE_ID","GENETIC_PROFILE_ID","PANEL_ID") VALUE
 INSERT INTO "sample_profile" ("SAMPLE_ID","GENETIC_PROFILE_ID","PANEL_ID") VALUES (14,2,NULL);
 INSERT INTO "sample_profile" ("SAMPLE_ID","GENETIC_PROFILE_ID","PANEL_ID") VALUES (14,4,NULL);
 
+-- gene_panel
+DELETE FROM "gene_panel" WHERE "STABLE_ID"="TESTPANEL3";
+INSERT INTO "gene_panel" ("INTERNAL_ID", "STABLE_ID", "DESCRIPTION") VALUES (3, "TESTPANEL3", "Gene panel used for unit testing in seed_mini.sql");
+
+-- gene_panel_list
```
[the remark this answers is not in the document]
jtquach1: ah, I wasn't sure about including it since I'm not sure how to run seed_mini.sql yet (or if I am)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -1,184 +1,384 @@
...
-            pstmt.executeUpdate();
-            genePanelMap.remove(genePanel.getStableId());
-        } catch (SQLException e) {
-            throw new DaoException(e);
-        } finally {
-            JdbcUtil.closeAll(DaoGenePanel.class, con, pstmt, null);
-        }
-    }
-}
+/*
```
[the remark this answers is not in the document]
jtquach1: I think so too, as I get this message from doing commits:

`jtquach@DESKTOP-IST3H2B:~/cbioportal_windows10$ git add  core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
warning: LF will be replaced by CRLF in core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java.
The file will have its original line endings in your working directory`

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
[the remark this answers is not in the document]
jtquach1: I think I might delete these methods since they're static and already defined in ImportGenePanel, although I'm not sure if that's a good programming practice?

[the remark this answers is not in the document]
jtquach1: Never mind, getPropertyValue and getGenes are private (not public)

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
jtquach1: just to clarify: are you referring to the union/intersection/difference methods that use streams?

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,241 @@
...
+        Scanner s = new Scanner(System.in);
+        boolean confirmed = false;
+
+        System.err.println("WARNING: This script is ONLY for updating a gene panel, not for adding a new gene panel. To add a new gene panel, run the ImportGenePanel script.");
+        System.err.println("If you are making changes to an existing gene panel due to a miscount of genes, proceed.");
+        System.err.println("Otherwise, if you would like to add a new version of a gene panel, please exit this script and import the gene panel with a new stable ID!");
+        System.err.println("Will you proceed to update an existing gene panel? (Y/N)");
+
+        // Prompt user for Y/N
+        while (!confirmed) {
```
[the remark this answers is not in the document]
jtquach1: Done! I added a public method promptUser which abstracts the logic

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -55,25 +55,22 @@ public void run() {
             String description = "Update gene panel files.";
 
             OptionParser parser = new OptionParser();
-            OptionSpec<String> data = parser.accepts( "data",
```
jtquach1: I have the auto formatter setting on in VS code, hence why so many lines are changed

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -92,46 +89,51 @@ public void importData() throws Exception {
...
+                        + " already exists in database but is unused. Proceeding with gene panel update!");
             }
-            s.close();
+            DaoGenePanel.updateGenePanel(genePanel, canonicalGenes);
         }
+        s.close();
     }
 
-    private static String getPropertyValue(String propertyName, Properties properties, boolean noSpaceAllowed) throws IllegalArgumentException {
+    public static boolean promptUser(Scanner s) {
```
jtquach1, line 114: This is the real change of the commit

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        if (noSpaceAllowed && propertyValue.contains(" ")) {
+            throw new IllegalArgumentException(propertyName + " cannot contain spaces: " + propertyValue);
+        }
+
+        return propertyValue;
+    }
+
+    private static Set<CanonicalGene> getGenes(String propertyName, Properties properties) {
```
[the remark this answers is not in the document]
jtquach1: It looks like this is the [history](https://github.com/cBioPortal/cbioportal/commits/master/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGenePanel.java) for ImportGenePanel, with the last commit from Jun 2 showing the latest implementation to getGenes. 

I'm also wondering if I should put my username as part of the "author" comment in UpdateGenePanel..?

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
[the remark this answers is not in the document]
jtquach1: Moved getPropertyValue and getGenes to GenePanelUtil.java (new file) in core\src\main\java\org\mskcc\cbio\portal\util

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
jtquach1: Just moved getPropertyValue and getGenes to GenePanelUtil.java (new file) in core\src\main\java\org\mskcc\cbio\portal\util

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,241 @@
...
+        String propertyValue = properties.getProperty(propertyName).trim();
+        if (propertyValue == null || propertyValue.length() == 0) {
```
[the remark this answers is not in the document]
jtquach1: IntelliJ says `Condition 'propertyValue == null' is always 'false'` and `'Simplify provertyValue == null' to false`

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,241 @@
...
+        while (!confirmed) {
+            String answer = s.nextLine();
+            
+            switch (answer) {
+	            case "Y":
+	            case "y":
+	            	System.err.println("Proceeding with changes.");
+	            	confirmed = true;
+	                // Updating should be done exclusively in the case where the gene panel exists and is already being used.         
+	                if (genePanel != null) {
```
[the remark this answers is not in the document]
jtquach1: The null check has been removed now (well earlier, but I forgot to respond here)

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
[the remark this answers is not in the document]
jtquach1: changed to public as part of GenePanelUtil.java

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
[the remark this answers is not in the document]
jtquach1: acknowledged and changed!

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -204,10 +202,6 @@ public static void updateGenePanel(GenePanel genePanel, Set<CanonicalGene> canon
     
     /**
      * Prints the genes to be added and removed from a gene panel in the database.
-     *
```
jtquach1: I just removed the Java doc comments (since I'm not sure what their purpose is)

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -66,7 +66,7 @@
     public static String getPropertyValue(String propertyName, Properties properties, boolean noSpaceAllowed) throws IllegalArgumentException {
         String propertyValue = properties.getProperty(propertyName).trim();
 
-        if (propertyValue == null || propertyValue.length() == 0) {
```
jtquach1: since this always computes to false; same below

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -55,7 +64,7 @@ public static String getPropertyValue(String propertyName, Properties properties
         }
 
         String[] genes = propertyValue.split("\t");
-        Set<CanonicalGene> canonicalGenes = new HashSet<CanonicalGene>();
```
jtquach1: based on IntelliJ code suggestions, just removed references to Generic types

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -65,7 +74,7 @@ public static String getPropertyValue(String propertyName, Properties properties
                     canonicalGenes.add(canonicalGene);
                 }
                 else {
-                    ProgressMonitor.logWarning("Could not find gene in the database: " + String.valueOf(geneId));
```
jtquach1: since long geneId is automatically converted to String

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -137,6 +146,20 @@ public Pair(Set<CanonicalGene> add, Set<CanonicalGene> remove) {
             this.add = add;
             this.remove = remove;
         }
+
+        @Override
+        public boolean equals(Object o) {
```
jtquach1, line 151: needed to implement equals so assertEquals within the unit tests won't throw errors

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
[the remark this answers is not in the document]
jtquach1: update: removed `propertyValue == null`

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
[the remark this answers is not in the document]
jtquach1: Now I understand what you mean; I think getting rid of `propertyName` and just using `gene_list` would be clearer too given the method name

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -0,0 +1,165 @@
...
+        // 5. Intersect but neither are supersets of one other
+        else {
+            // Add A - AnB
+            toAdd = difference(new ArrayList<>(incoming), new ArrayList<>(AnB));
+            // Remove B - AnB
+            toRemove = difference(new ArrayList<>(original), new ArrayList<>(AnB));
+        }
+
+        return new Pair(toAdd, toRemove);
+    }
```
Luke-Sikina, line 141: This seems too complex for what it was trying to do. I replaced it with this and it passed all the tests:
```suggestion
    public static Pair getAddRemove(Set<CanonicalGene> incoming, Set<CanonicalGene> original) {
        Set<CanonicalGene> add = incoming.stream()
            .filter(e -> !original.contains(e))
            .collect(Collectors.toSet());
        Set<CanonicalGene> remove = original.stream()
            .filter(e -> !incoming.contains(e))
            .collect(Collectors.toSet());
        
        return new Pair(add, remove);
    }
```

Are these functionally equivalent, or am I missing something?

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,313 @@
+package org.mskcc.cbio.portal.util;
```
Luke-Sikina, line 1: We'll circle back to these in the next review. Don't bother with them for now.

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    public void importData() throws Exception {
+        ProgressMonitor.setCurrentMessage("Reading data from:  " + genePanelFile.getAbsolutePath());
+        Properties properties = new Properties();
+        properties.load(new FileInputStream(genePanelFile));
+
+        String stableId = getPropertyValue("stable_id", properties, true);
+        Set<CanonicalGene> canonicalGenes = extractGenes("gene_list", properties, true);
+        GenePanel genePanel = DaoGenePanel.getGenePanelByStableId(stableId);
```
Luke-Sikina, line 64: Let's have some logic here that explicitly checks to make sure the gene panel is not null, and errors and exits if it is null

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -181,4 +180,92 @@ public static void deleteGenePanel(GenePanel genePanel) throws DaoException {
...
+            JdbcUtil.closeAll(DaoGenePanel.class, con, pstmt, rs);
+        }
+    }
+    
+    /**
+     * Prints the genes to be added and removed from a gene panel in the database.
+     */
+    public static void updatePreview(GenePanel genePanel, Set<CanonicalGene> incoming) throws DaoException {
+        Pair pair = getAddRemove(incoming, extractGenePanelGenes(genePanel.getInternalId()));
+        Set<CanonicalGene> toAdd = (Set<CanonicalGene>) pair.add;
```
Luke-Sikina, line 208: Get rid of casts

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -0,0 +1,165 @@
...
+                    // for a single gene symbol found in the data file 
+                    canonicalGenes.add(canonicalGenesList.get(0));
+                }
+                else {
+                    ProgressMonitor.logWarning("Could not find gene in the database: " + panelGene);
+                }
+            }
+        }
+
+        return (canonicalGenes.size() == genes.length) ? canonicalGenes : null;
```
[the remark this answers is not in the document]
jtquach1: existing behavior, but can document in @return

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -181,4 +180,92 @@ public static void deleteGenePanel(GenePanel genePanel) throws DaoException {
...
+        System.out.print("ADD:");
+        for (CanonicalGene canonicalGene : toAdd) {
+            System.out.print("\t" + canonicalGene);
+        }
+        System.out.println();
+        System.out.print("REMOVE:");
+        for (CanonicalGene canonicalGene : toRemove) {
+            System.out.print("\t" + canonicalGene);
+        }
+        System.out.println();
```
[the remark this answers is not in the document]
jtquach1: Wow, streams are pretty nifty! Does the `::` map/apply `toString` to `CanonicalGene`?

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public void run() {
+        try {
+            String progName = "UpdateGenePanel";
+            String description = "Update gene panel files.";
+
+            OptionParser parser = new OptionParser();
+            OptionSpec<String> data = parser.accepts("data", "gene panel file").withRequiredArg()
+                    .describedAs("data_file.txt").ofType(String.class);
```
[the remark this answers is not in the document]
jtquach1: ah, I may have to find a Prettier plugin for Eclipse/IntelliJ

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,142 @@
...
+        Set<CanonicalGene> canonicalGenes = extractGenes("gene_list", properties, true);
+        GenePanel genePanel = DaoGenePanel.getGenePanelByStableId(stableId);
+
+        // Add user prompt here about changes to be made to gene panel
+        Scanner s = new Scanner(System.in);
+        System.err.println("WARNING: You are about to make the following changes to gene panel " + stableId + ".");
+        DaoGenePanel.updatePreview(genePanel, canonicalGenes);
+        System.err.println("Proceed? This action is irreversible. (Y/n)");
+        boolean confirmed = promptUser(s);
+        if (confirmed) {
```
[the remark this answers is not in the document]
jtquach1: put both calls to promptUser inside if condition

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
jtquach1: I never thought of it that way... I think they are functionally equivalent!

core/src/main/java/org/mskcc/cbio/portal/util/GenePanelUtil.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Utilities for operations on sets and importing/updating gene panels.
+ *
+ * @author jtquach1
+ */
+public class GenePanelUtil {
+
+    // A ^ B
```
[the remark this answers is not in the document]
jtquach1: Removed union, intersection, difference since the below suggestion for getAddRemove (committed) makes them obsolete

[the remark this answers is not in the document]
jtquach1: Added javadoc comments to all public functions in DaoGenePanel.java and GenePanelUtil.java

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,142 @@
...
+            } else {
+                ProgressMonitor.logWarning("Gene panel " + stableId
+                        + " already exists in database but is unused. Proceeding with gene panel update!");
+            }
+            DaoGenePanel.updateGenePanel(genePanel, canonicalGenes);
+        }
+        s.close();
+    }
+
+    private static boolean promptUser(Scanner s) {
```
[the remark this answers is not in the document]
jtquach1: turns out after closing a scanner once, that breaks InputStream for the rest of the class forever
(fix was to make the Scanner a class variable)

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
jtquach1: Done with 
```java
        if (genePanel == null) {
            ProgressMonitor.logWarning("Gene panel " + stableId + " does not exist in the database! Exiting.");
            return;
        }
```

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,217 @@
...
+        Pair expected = new Pair(g4, g5);
+        Pair actual = getAddRemove(g1, g2);
+        assertEquals(expected, actual);
+        expected = new Pair(g5, g4);
+        actual = getAddRemove(g2, g1);
+        assertEquals(expected, actual);
+    }
+    
+    @Test
+    public void testGetAddRemove() {
```
Luke-Sikina, line 211: Delete

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    private CanonicalGene makeGene(int geneticEntityId, long entrezGeneId, String hugoGeneSymbol, Set<String> aliases) {
+        CanonicalGene gene = new CanonicalGene(geneticEntityId, entrezGeneId, hugoGeneSymbol, aliases);
+        gene.setType("protein-coding");
+        return gene;
+    }
+
+    @Test
+    public void testGetAddRemoveAEqualsB() {
```
Luke-Sikina, line 154: We don't need the duplicate examples for each test case. It just makes this harder to read. It doesn't really add much in terms of additional verification.

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,217 @@
...
+    Properties p5;
+    Properties p6;
+    Set<CanonicalGene> g1;
+    Set<CanonicalGene> g2;
+    Set<CanonicalGene> g3;
+    Set<CanonicalGene> g4;
+    Set<CanonicalGene> g5;
+    Set<CanonicalGene> g6;
+
+    CanonicalGene otor = makeGene(13321, 56914, "OTOR", makeAliases("FDP", "MIAL1"));
```
Luke-Sikina, line 27: Make these class variables, remove the gene sets, create gene sets as local variables in each test case.

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,145 @@
...
+        GenePanel genePanel = DaoGenePanel.getGenePanelByStableId(stableId);
+
+        if (genePanel == null) {
+            ProgressMonitor.logWarning("Gene panel " + stableId + " does not exist in the database! Exiting.");
+            return;
+        }
+
+        System.err.println("WARNING: You are about to make the following changes to gene panel " + stableId + ":");
+        DaoGenePanel.updatePreview(genePanel, canonicalGenes);
+        System.err.println("Proceed? This action is irreversible. (Y/n)");
```
Luke-Sikina, line 79: Couldn't they just run the script again with the old gene panel file? Seems pretty reversible to me.
```suggestion
        System.err.println("Proceed? (Y/n)");
```

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,145 @@
...
+                    + " already exists in database and is being used. Proceeding with gene panel update!");
+            } else {
+                ProgressMonitor.logWarning("Gene panel " + stableId
+                    + " already exists in database but is unused. Proceeding with gene panel update!");
+            }
+            DaoGenePanel.updateGenePanel(genePanel, canonicalGenes);
+        }
+    }
+
+    private static boolean promptUser() {
```
Luke-Sikina, line 93: promptUser is a little vague (I may have told you to call the method this). Let's call it `confirmContinue`

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,145 @@
...
+            return;
+        }
+
+        System.err.println("WARNING: You are about to make the following changes to gene panel " + stableId + ":");
+        DaoGenePanel.updatePreview(genePanel, canonicalGenes);
+        System.err.println("Proceed? This action is irreversible. (Y/n)");
+
+        if (promptUser()) {
+            if (DaoSampleProfile.sampleProfileMappingExistsByPanel(genePanel.getInternalId())) {
+                ProgressMonitor.logWarning("Gene panel " + stableId
```
Luke-Sikina, line 83: Shouldn't this be logDebug instead of logWarning?
```suggestion
                ProgressMonitor.logDebug("Gene panel " + stableId
```

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,145 @@
...
+                    + " already exists in database and is being used. Proceeding with gene panel update!");
+            } else {
+                ProgressMonitor.logWarning("Gene panel " + stableId
```
Luke-Sikina, line 86: ditto

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,145 @@
...
+     * @param args the arguments given on the command line
+     */
+    public static void main(String[] args) {
+        ConsoleRunnable runner = new UpdateGenePanel(args);
+        System.err.println(
+            "WARNING: This script is ONLY for updating a gene panel, not for adding a new gene panel. To add a new gene panel, run the ImportGenePanel script.");
+        System.err.println("If you are making changes to an existing gene panel due to a miscount of genes, proceed.");
+        System.err.println(
+            "Otherwise, if you would like to add a new version of a gene panel, please exit this script and import the gene panel with a new stable ID!");
+        System.err.println("Will you proceed to update an existing gene panel? (Y/n)");
```
Luke-Sikina, line 138: This is super picky, but I don't like the multiple method calls here. Let's just make this one big block. Also, I added a few more new lines. I like to stick to about 80 columns for console output. Oh, also, why is this going to standard err instead of standard out? This isn't an error message.
```suggestion
        System.out.print(
            "WARNING: This script is ONLY for updating a gene panel, not for adding a new gene panel.\n" + 
            "To add a new gene panel, run the ImportGenePanel script.\n" +
            "If you are making changes to an existing gene panel due to a miscount of genes, proceed.\n" +
            "Otherwise, if you would like to add a new version of a gene panel, please exit this script\n" +
            "and import the gene panel with a new stable ID!\n" +
            "Will you proceed to update an existing gene panel? (Y/n)\n"
        );
```

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,242 @@
...
+    CanonicalGene adamts20 = makeGene(15009, 80070, "ADAMTS20", makeAliases("ADAM-TS20", "GON-1", "ADAMTS-20"));
+    CanonicalGene kat2a = makeGene(2154, 2648, "KAT2A", makeAliases("hGCN5", "GCN5", "GCN5L2", "PCAF-b"));
+    CanonicalGene myb = makeGene(3674, 4602, "MYB", makeAliases("c-myb_CDS", "Cmyb", "c-myb", "efg"));
+    CanonicalGene npipb15 = makeGene(24049, 440348, "NPIPB15", makeAliases("A-761H5.4", "NPIPL2"));
+    CanonicalGene dtnb = makeGene(1492, 1838, "DTNB", null);
+    CanonicalGene ablim1 = makeGene(3239, 3983, "ABLIM1", makeAliases("LIMAB1", "abLIM-1", "ABLIM", "LIMATIN"));
+    CanonicalGene piezo1 = makeGene(7609, 9780, "PIEZO1", makeAliases("FAM38A", "Mib", "LMPH3", "DHS"));
+    CanonicalGene fgfr3 = makeGene(1827, 2261, "FGFR3", makeAliases("JTK4", "HSFGFR3EX", "ACH", "CEK2", "CD333"));
+
+    public TestGenePanelUtil() {
```
Luke-Sikina, line 30: Delete

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,242 @@
...
+        Pair actual = getAddRemove(
+            makeGeneSet(adamts20, ablim1),
+            makeGeneSet(piezo1, fgfr3));
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testGetAddRemoveAOrBEmpty() {
+        Pair expected = new Pair(
+            new HashSet<>(),
```
Luke-Sikina, line 167: Just to make things _even easier_ to read, I'd do `makeGeneSet()` instead of `new HashSet<>()`

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
```diff
@@ -0,0 +1,242 @@
...
+            new HashSet<>()
+        );
+        Pair actual = new Pair(
+            makeGeneSet(ablim1, fgfr3, adamts20, piezo1),
+            new HashSet<>()
+        );
+        assertEquals(expected, actual);
+    }
+    
+    private HashSet<String> makeAliases(String... aliases) {
```
Luke-Sikina, line 229: Remind me tomorrow and I'll show you a fun way to generify this and then I'll explain why its a bad idea.

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
jtquach1: I reckoned that the red text from the System.err.print looks more alerting than the black text from System.out.print, but yes it is not an error message perse

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
jtquach1: Oh that's true, what I mean is that for that specific panel, if you end up removing some genes and you try to import them back, the genes won't have an entry in the `sample_profile` (I think)

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
jtquach1: Do you mean the class constructor or the genes? The genes are used in the tests but deleting the constructor still lets the tests run and pass

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
jtquach1: hm, I suppose so since it's not a "warning" the user has to be worried about on their end, but I'm not sure whether to use setCurrentMessage over logDebug

core/src/test/java/org/mskcc/cbio/portal/util/TestGenePanelUtil.java
jtquach1: for all instances of `new HashSet<>()`?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -181,4 +198,103 @@ public static void deleteGenePanel(GenePanel genePanel) throws DaoException {
...
+
+    /**
+     * Prints the genes to be added and removed from a gene panel in the database.
+     *
+     * @param genePanel
+     * @param incoming
+     * @throws DaoException
+     */
+    public static void updatePreview(GenePanel genePanel, Set<CanonicalGene> incoming) throws DaoException {
+        Pair pair = getAddRemove(incoming, extractGenePanelGenes(genePanel.getInternalId()));
```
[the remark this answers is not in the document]
jtquach1: According to the [Java 1.5 documentation](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html), one can call static methods without their class name. I didn't know this since IntelliJ suggested this import style.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -181,4 +198,103 @@ public static void deleteGenePanel(GenePanel genePanel) throws DaoException {
...
+     * Inserts and deletes genes from a gene panel represented by an internalId in
+     * the database.
+     */
+    private static void updateGenePanelGeneList(Integer internalId, Set<CanonicalGene> incoming) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+
+        try {
+            con = JdbcUtil.getDbConnection(DaoGenePanel.class);
```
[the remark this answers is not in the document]
jtquach1: Thanks for the note! I've moved line 270 after the if statement

core/src/main/java/org/mskcc/cbio/portal/scripts/UpdateGenePanel.java
```diff
@@ -0,0 +1,147 @@
...
+        Properties properties = new Properties();
+        properties.load(new FileInputStream(genePanelFile));
+
+        ProgressMonitor.setCurrentMessage("Retrieving matching gene panel from database.");
+        String stableId = extractPropertyValue("stable_id", properties, true);
+        Set<CanonicalGene> canonicalGenes = extractGenes(properties, true);
+        GenePanel genePanel = DaoGenePanel.getGenePanelByStableId(stableId);
+
+        if (genePanel == null) {
+            ProgressMonitor.logWarning("Gene panel " + stableId + " does not exist in the database! Exiting.");
```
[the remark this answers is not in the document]
jtquach1: According to Luke, the import and update functionality should be separate from each other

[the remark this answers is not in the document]
Luke-Sikina: @n1zea144 Leaving them as separate scripts makes it a bit harder to make the mistake of updating an existing gene panel when you instead meant to upload a new version of a gene panel for new samples. I think this is an easy enough mistake- the user copies the gene panel file, vims it, updates the genes, but forgets to change the name of the panel before running the script. By keeping the two scripts separate, we make it so a user that ignores warning messages and just presses enter a bunch cannot end up making this mistake.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenePanel.java
```diff
@@ -181,4 +198,103 @@ public static void deleteGenePanel(GenePanel genePanel) throws DaoException {
...
+            Set<CanonicalGene> toRemove = pair.remove;
+
+            // Add and remove genes from specified gene panel
+            for (CanonicalGene canonicalGene : toAdd) {
+                pstmt = con.prepareStatement("INSERT INTO gene_panel_list (`INTERNAL_ID`, `GENE_ID`) VALUES (?,?)");
+                pstmt.setInt(1, internalId);
+                pstmt.setLong(2, canonicalGene.getEntrezGeneId());
+                pstmt.executeUpdate();
+            }
+            for (CanonicalGene canonicalGene : toRemove) {
```
n1zea144, line 287: It looks like its possible to remove all genes from the panel.  Should we check this and if so, should we prevent this from happening?

jtquach1: In UpdateGenePanel.java, the user prompt in importData() verifies the changes that will be made to the gene panel, like so:
image
Other than that, I asked Luke if importing an empty gene panel should allow for the removal of all existing genes, and he stated that behavior sounds reasonable. I suppose if the user wanted to add back the genes, they could pull up an older version of their input file with the gene names. 
I'll ping Luke about whether after all genes are removed, that the gene panel should be removed from gene_panel_list as well

jtquach1: ah, so removal of all genes means stop the update script and don't make the deletions

n1zea144: It's not meaningful to have a panel in the database, which samples refer to in sample_profile, that contain no genes.  I'm also not sure what the affect is on the web page.  One approach is to remove the the panel and subsequent records in sample_profile (via cascade), but thats a bit messy and could happen by mistake.

n1zea144: Spoke with @Luke-Sikina via Slack.  Best approach is not to prevent an update which results in empty panel in database.

jtquach1: Oh, does that mean that within UpdateGenePanel, checking if the incoming set of genes is empty and aborting the script is not needed?

jtquach1: I've just added the check on lines 77-80 on UpdateGenePanel, after the gene panel null check:
```java
        if (genePanel == null) {
            ProgressMonitor.logWarning("Gene panel " + stableId + " does not exist in the database! Exiting.");
            return;
        }

        if (canonicalGenes == null || canonicalGenes.isEmpty()) {
            System.err.println("Incoming gene panel is empty, which would result in the removal of all genes from gene panel " + stableId + ". Exiting.");
            return;
        }
```

n1zea144: Yeah, checking/aborting when the incoming gene list from the panel file is empty should suffice, thanks for the adjustment.

jtquach1: Thank you for the reviews!

jtquach1: I've just updated the markdown file again to reflect that trying to import an empty gene panel will abort the script.

PR #7708, opened 2020-07-21
remove leftover api-legacy references

inodb:
Current v3.4.4 gives some error stack trace b/c of leftover api-legacy references I believe:

```
13:49:52 [localhost-startStop-1] ERROR org.springframework.web.servlet.DispatcherServlet - Context initialization failed                                       
org.springframework.beans.factory.BeanDefinitionStoreException: IOException parsing XML document from class path resource [applicationContext-weblegacy.xml]; n
ested exception is java.io.FileNotFoundException: class path resource [applicationContext-weblegacy.xml] cannot be opened because it does not exist 
```

### DIFF
--- a/security/security-spring/src/main/resources/applicationContext-security.xml
+++ b/security/security-spring/src/main/resources/applicationContext-security.xml
@@ -107,11 +107,6 @@
         <csrf disabled="true"/>
         <custom-filter ref="tokenAuthenticationFilter" before="FORM_LOGIN_FILTER"/>
     </http>
-    <http use-expressions="true" pattern="/api-legacy/**" entry-point-ref="restAuthenticationEntryPoint" create-session="never">
-        <intercept-url pattern="/**" access="isAuthenticated()"/>
-        <csrf disabled="true"/>
-        <custom-filter ref="tokenAuthenticationFilter" before="FORM_LOGIN_FILTER"/>
-    </http>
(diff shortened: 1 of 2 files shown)


security/security-spring/src/main/resources/applicationContext-security.xml
```diff
@@ -629,7 +624,7 @@
         <!-- create-session is never so spring security will not create the session,
             this relies on the app creating the session -->
         <http use-expressions="true" entry-point-ref="restAuthenticationEntryPoint" create-session="never">
-            <intercept-url pattern="/api-legacy/proxy/session-service/**" access="@sessionSecurity.checkWrite(request)"/>
+            <intercept-url pattern="/api/proxy/session-service/**" access="@sessionSecurity.checkWrite(request)"/>
```
inodb, line 627: @kalletlak @mandawilson do any of you know if this is still necessary? Thanks!

inodb:
Thanks @n1zea144  - looks like it was indeed re-introduced later in this routing PR: https://github.com/cBioPortal/cbioportal/pull/6861/files#diff-7b32dcb779649a9a5cf3a2fb4410b7e6L213-R236. Prolly was based off of an old pom file, but looks like there were no other references to api-legacy included

security/security-spring/src/main/resources/applicationContext-security.xml
inodb: Thanks @kalletlak! I updated it

Issue #4065, opened 2018-04-04
Add a optional warning to all pages


jjgao:
MSK users sometimes put internal urls in their publications. e.g.

http://ascopubs.org/doi/full/10.1200/JCO.2017.75.3384

![image](https://user-images.githubusercontent.com/840895/38285686-64a7c988-378f-11e8-98f9-959401de61a4.png)

We need to be able to add a warning and ask them not to do that. 

- [ ] add a new property in frontend or backend: `study_view_warning`
- [ ] append the warning at the end of the description of a study 
- [ ] add a warning for all our private portal: "Please do NOT include the URL to this study or any other pages on this portal in your publication. If you need to make a study for your publication in the public cBioPortal, please contact cbioportal@cbio.mskcc.org."

jjgao:
this is still happening. But maybe the message I put above is a bit too long.

thoughts, @schultzn ?

schultzn:
How about just "Study URL is temporary - do NOT include in publications."

inodb:
This applies to all studies in MSKCC/triage portals right? If so, maybe we can make it more general i.e. allow to add some blurb to the description of all studies in a portal. For example a `skin.study_view.blurb`. Similar to `skin.blurb`: https://docs.cbioportal.org/2.3-customization/portal.properties-reference#control-the-content-of-specific-sections.

jjgao:
@inodb sounds good. Yes, to all studies.

inodb:
we should do this

Leowisd:
![image](https://user-images.githubusercontent.com/32425638/88334859-fce3d700-ccf7-11ea-9478-1ec3ad7b2547.png)

Leowisd:
![image](https://user-images.githubusercontent.com/32425638/89047863-a4729200-d314-11ea-8003-fb4a6c5818a2.png)
![image](https://user-images.githubusercontent.com/32425638/89047899-afc5bd80-d314-11ea-8b2d-6bcee558c026.png)
![image](https://user-images.githubusercontent.com/32425638/89047926-bb18e900-d314-11ea-9205-612459473528.png)

jjgao:
Thanks, @Leowisd. It looks good!

For the last one, could you add "transferred to the public cBioPortal"?

Is this message configurable? e.g. we should have a message in the private portal too.

@schultzn @n1zea144 comments?

Leowisd:
Okay.

alisman:
[...]

PR #7718, opened 2020-07-24
Add option to show a study view blurb

zhx828:
This is a part of fixes for https://github.com/cBioPortal/cbioportal/issues/4065

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -78,6 +78,9 @@ skin.right_nav.show_whats_new=true
 # settings controlling what to show in the right navigation bar
 skin.study_view.link_text=To build your own case set, try out our enhanced Study View.
 
+# setting controlling the blurb in study view
+# skin.study_view.blurb=
+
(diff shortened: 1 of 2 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -78,6 +78,9 @@ skin.right_nav.show_whats_new=true
 # settings controlling what to show in the right navigation bar
 skin.study_view.link_text=To build your own case set, try out our enhanced Study View.
 
+# setting controlling the blurb in study view
+skin.study_view.blurb=
```
inodb, line 82: @zhx828 does it work when this one is commented out as well? Would prefer to not set default values in EXAMPLE properties file whenever possible

zhx828: @inodb it works as well. I commented out.

zhx828:
Close this ticket.
After further discussion with the group, we need to support more complex text blurb for portal, potentially markdown. The feature will temporary be hard coded for MSK portal until next step is established.

Issue #7756, opened 2020-08-03
adding back mutation assessor


jjgao:
Mutation Assessor is back online.

PR #7767, opened 2020-08-06
add genome nexus annotation sources to portal.properties

leexgh:
Partial Fix https://github.com/cBioPortal/cbioportal/issues/7756

`show.genome_nexus.annotation_sources` will set which sources to show on the frontend, list all sources name with comma-separated.

For now we only support `mutation_assessor`.

By default;
```
show.genome_nexus.annotation_sources=mutation_assessor
```
TODO: support more sources like clinvar and gnomad

### DIFF
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -59,6 +59,7 @@
             "show.oncokb",
             "show.civic",
             "show.genomenexus",
+            "show.genomenexus.annotation_sources",
             "show.mutation_mappert_tool.grch38",
             "show.transcript_dropdown",
             "skin.documentation.about",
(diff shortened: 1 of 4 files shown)


portal/src/main/webapp/config_service.jsp
```diff
@@ -59,6 +59,7 @@
             "show.oncokb",
             "show.civic",
             "show.genomenexus",
+            "show.genome_nexus.annotation_sources",
```
inodb, line 62: @leexgh Question re consistency. In the above property (`show.genomenexus`) we spell `genomenexus` without underscore. Are there other properties where we use `genome_nexus`? Otherwise maybe better to keep it consistent and make it `show.genomenexus.annotation_sources`?

leexgh: @inodb good point, I'll change to `genomenexus` to be consistent with others

PR #7789, opened 2020-08-12
increase genes endpoint page size

SRodenburg:
The genes API endpoint pageSize was restricted to 100K entries, while there are seemingly more genes in the DB (many many phosphoproteins).

The page size in the gene controller is increased and the validator is adjusted as appropriate

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -4885,11 +4885,8 @@ def validate_data_relations(validators_by_meta_type, logger):
 def request_from_portal_api(server_url, api_name, logger):
     """Send a request to the portal API and return the decoded JSON object."""
 
-    if api_name in ['info', 'cancer-types', 'genesets', 'gene-panels']:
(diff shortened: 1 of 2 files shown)


SRodenburg:
These would be 2 separate solutions to the same problem.

First is to increase the page size until all genes are returned in one go. This is the easiest to do IMO. Second, is to iterate over smaller pages until all entries are retrieved.

SRodenburg:
@inodb what about this?

rmadupuri:
Hi @SRodenburg @inodb, just wanted to bring to your attention that the validator is failing to fetch the data from genes endpoint. See the screenshot below. 

Screen Shot 2020-08-18 at 11 45 48 AM

Also, the url (https://www.cbioportal.org/api/genes?pageSize=9999999) shows the message that `pageSize must be less than or equal to 100000`. Thanks!

inodb:
@rmadupuri sorry it isn't deployed yet. Will deploy now

inodb:
@rmadupuri looks like it is working now

rmadupuri:
Perfect. Thanks @inodb!

PR #7796, opened 2020-08-13
add chrom size data and script

SRodenburg:
The validator previously depended on the UCSC portal to download chromosome sizes, and several times this causes crashes when the UCSC was under maintance.
I propose to make use of a pre-downloaded json file instead -- Chromosome sizes should be stable between builds.


SRodenburg:
Just wait with the merge until I fixed the tests

SRodenburg:
ready to go

PR #7842, opened 2020-08-28
Add support for custom driver annotations for CNA data and import of OncoKB annotations as custom driver annotations

pvannierop:
# Custom driver annotations for CNA data and automated import of OncoKB annotations as custom driver annotations

Note:
- This PR is part of RFC56
- This PR covers the following sub-tasks in RFC56:
   1. [Custom PD annotations for CNA alteration type](https://docs.google.com/document/d/1TLRIOurdRK1HE7nDPqk6d4zhfjpgchBd9tU6PQpbwxw/edit#heading=h.z9ypreqxi1s0)
   2. [Include OncoKB annotation in loaded study](https://docs.google.com/document/d/1TLRIOurdRK1HE7nDPqk6d4zhfjpgchBd9tU6PQpbwxw/edit#heading=h.yjoap52lpp73)

The two sub-tasks are submitted as consecutive commits in this PR and can be reviewed independently

## Custom PD annotations for CNA alteration type

This commit will:
- Update the database schema to hold custom driver annotations for mutations, discrete copy number alterations and structural variants to the _alteration_driver_annotation_
image
- Add support in validator for custom driver annotations for discrete CNA data
- Update import code for for custom driver annotations for discrete CNA data
- Update MyBatis mappers to reflect schema update
- Update API for discrete CNA data to return custom driver annotations

## Include OncoKB annotation in loaded study

This commit will:
- Implement addition of OncoKB annotations for mutations and discrete copy number annotations as custom annotations to study files (structural variant not supported)
- Implement integration test for import mechanism
- Fix some minor issues in local files used for testing of data validation and import scripts


jjgao:
I am wondering if it is worth adding another new table "genetic_alteration" similar to "genetic_entity"? (maybe renaming the current alteration table to profile_data to avoid confusion.)

@pvannierop @Sjoerd-van-Hagen @pieterlukasse @n1zea144

Sjoerd-van-Hagen:
You mean as a supertype to mutation, CNA and fusion?

jjgao:
@Sjoerd-van-Hagen yes.

pvannierop:
@jjgao Perhaps, but I do not really see the connection with this PR. Or am I missing your point?

zhx828:
Thank you for initiating the efforts. On top of @inodb 's comment, my main concern is the data discrepancy between patient view and study view. If we save the annotation in the database for all mutations, then we should use the same annotation for patient view as well instead of pulling from oncokb in real time.

Another concern would be the license. I'm not so sure whether you can distribute the portal data dump once you have the oncokb annotation in the database. I will circle back to the oncokb team for more clarification. cc @jjgao @schultzn

jjgao:
> @jjgao Perhaps, but I do not really see the connection with this PR. Or am I missing your point?

@pvannierop My thought was triggered by your proposed `alteration_dirver_annotation` table, which is connecting to three tables (`sample_cna_event`, `structural_variant`, and `mutation`). I was thinking that it might be useful to create an `alteration` table that connects to the three tables, and `alteration_dirver_annotation` table can connect to `alteration` table. The `alteration` table may also be useful to address other use cases, e.g. "Alteration" tab in comparison (so that we can pull from one table instead of three in the backend).

Or maybe `alteration_dirver_annotation` is already the `alteration` table since it'll contain all the alterations in the three tables?

pvannierop:
@jjgao Ah, I understand you now. What you propose makes a lot of sense and is a good idea in my view. What is perhaps not apparent from the design is that the `alteration_driver_annotation` table is only populated when an annotation is linked to an alteration. In that sense it does not act as a true parent table at the moment. Moreover, refactoring this table to become a true parent of cna, mutation and structural variant involved updates of all SQL queries, tests and test data. This is not something done easily. I propose to continue discussion on scheduling of this work on Slack.

jjgao:
Copy over some discussions from Slack.
image

jjgao:
Inspired by @pvannierop's design and the discussion above, here is one option of db design. I don't think I have thought it thoroughly, so this is for triggering some discussion and see if it is possible to take this opportunity to have a more general design.

This design addresses several points:
- Metadata of annotation sources are supported by `annotation_source`, including version
- One alteration can be annotated by multiple annotation sources (e.g. OncoKB v1, OncoKB v2, Civic)
- `alteration` contains all alterations (mutations, CNAs, and SVs) - no need to query the individual tables. This makes it possible to construct SQL for study view page without pulling out sample-level data. (Currently, you would have to pull sample-level data from the three tables and combine them together).
- Connecting to `genetic_entity` makes it flexible for future extension, e.g. other alteration types.

More discussion points:
- Maybe we can even remove `driver` from `alteration_driver_annotation`, e.g. make it `alteration_annotation`, and `FILTER`, etc. This table can potentially host all kinds of annotation.
- We need some examples of `DRIVER_FILTER`/`ANNOTATION`, and see if we can improve it, e.g. maybe annotation can be moved to annotation_source?

image

@pvannierop @Sjoerd-van-Hagen @inodb @n1zea144

Sjoerd-van-Hagen:
@jjgao can you create a separate discussion for this so we can move forward with merging this pull requests and implementing the filtering in the front end?

pvannierop:
I close this PR because we decided to split this PR into multiple chunks.

PR #7843, opened 2020-08-28
Fix portaldump json format to match new genes API format

SRodenburg:
There were several issues with the format of the portaldump JSON files. The portaldump JSON files should roughly match the format of the API responses, because it has to go through the same validator.

A summary of the changes:
- adjust the JsonProperties in some classes to return CamelCase JSON attributes
- change name of cancer-types API endpoint
- make gene panels return a DETAILED JSON file, including genes to validate
- change Entrez IDs to integers in tests.


core/src/main/scripts/importer/validateData.py
```diff
@@ -4765,7 +4765,7 @@ def validateStudyTags(tags_file_path, logger):
             extra={'filename_': tags_file_path})
         with open(tags_file_path, 'r') as stream:
             try:
-                parsedYaml = yaml.load(stream)
+                parsedYaml = yaml.load(stream, Loader=yaml.FullLoader)
```
pvannierop, line 4768: I think this correction does not work in CI. CI runs still on python 3.4. I propose to remove this for now.

pvannierop (changes requested):

core/src/main/scripts/importer/validateData.py
SRodenburg: hmm in my other PR I also adjusted this in the travis file to 3.6

pvannierop: But here you did? Or did I miss something.

pvannierop: But here you did? Or did I miss something.

SRodenburg: please see next commit :)

inodb (approved):
LGTM

pvannierop (approved):

PR #7853, opened 2020-08-31
validation update: pancan normal exp addition

yichaoS:
- added new data type


core/src/main/scripts/importer/validateData.py
```diff
@@ -4640,6 +4640,7 @@ def processCaseListDirectory(caseListDir, cancerStudyId, logger,
                                 'all_cases_with_mutation_and_cna_and_mrna_data',
                                 'all_cases_with_gsva_data',
                                 'all_cases_with_sv_data',
+                                'all_normal_cases_with_mrna_rnaseq_data',
```
jjgao, line 4643: We don't need a case list for normal cases. The normal expression data will only be used for reference and additional comparison. Users will not be able to select them for query.

jjgao, line 4643: We don't need a case list for normal cases. The normal expression data will only be used for reference and additional comparison. Users will not be able to select them for query.

yichaoS: removed

jjgao:
@kalletlak could you also review this?

kalletlak:
@yichaoS @jjgao should be add a new datatype for normals(to differentiate them from regular tumor expression data)?

yichaoS:
@kalletlak @jjgao

For now we are adding the expression values from the normal samples just like another genomic profile under associated tumor samples (e.g. if the exp value for `TCGA-BL-A13J-11` is `19.9125` for gene `A1BG`, we would add a new profile named `rna_seq_v2_mrna_median_normals` under 'TCGA-BL-A13J-01' containing this value, instead of creating a new sample instance of `TCGA-BL-A13J-11`). 

Therefore I only added 3 new profiles under the `allowed-data-type.txt`

If I understand correctly, for now, we don't want these normal samples to actually exist in the studies, but  only have their expression values be seen under the `plots` tab. 

Please let me know what you think? Thanks!

jjgao:
@yichaoS let's leave it as is for now. I'll discuss with @kalletlak and see how the proposed features can be implemented.

jjgao:
I think this is good to go. @kalletlak any more comments?

kalletlak (approved):
LGTM

sheridancbio:
Unit test failures seem to be due to a travis stall while building core. I'm not sure if this is just travis flakiness. Maybe we can re-trigger the unit tests?

Also, there are two end-to-end test failures reported:
shows_gsva_heatmap_track - .oncoprint_tab

sheridancbio:
This PR only modifies the datatypes allowed to be imported .. and expands a warning message for feedback. I don't see how this could have resulted in the CI failures. Does anyone object to merging this as-is?

sheridancbio:
Travis configuration seems to be too slow / under-resourced. Timeout errors reported in these runs:

Issue #7841, opened 2020-08-28
(Generic Assay) meta endpoint bug


dippindots:
meta endpoint fetching by `genericAssayStableIds` is not working.

![image](https://user-images.githubusercontent.com/15748980/91520569-97b96d80-e8c3-11ea-943e-73b9be0a38c7.png)

PR #7856, opened 2020-09-01
Fix generic meta endpoint error when fetching by stable ids only

dippindots:
Fix https://github.com/cbioportal/cbioportal/issues/7841

When fetching generic assay meta by only giving the stable ids, since the allStableIdMap is empty, then given stable ids will be filtered out, and will get empty result.
Resolve this by using following logic:
1. have molecular profiles ids and stable ids: use the shared stable ids
2. only has molecular profiles ids: use stable ids derived from molecular profiles ids
3. only has stable ids: use the given stable ids

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
@@ -58,16 +58,22 @@ public class GenericAssayServiceImpl implements GenericAssayService {
                 List<String> stableIdsInMolecularProfiles = genericAssayRepository.getGenericAssayStableIdsByMolecularIds(distinctMolecularProfileIds);
                 allStableIds.addAll(stableIdsInMolecularProfiles);
             }
-        } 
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -58,16 +58,22 @@
...
-
-			allStableIds = stableIds
-					.stream()
-					.filter(stableId -> allStableIdMap.containsKey(stableId))
-					.collect(Collectors.toSet());
+            // if stableIds and molecularProfileIds both exist, find the common
+            if (stableIds != null) {
+                Map<String, String> allStableIdMap = allStableIds
+                        .stream()
+                        .collect(Collectors.toMap(stableId -> stableId, stableId -> stableId));
```
kalletlak, line 65: Not important, you could use identity function here
`.collect(Collectors.toMap(Function.identity(), Function.identity()));`

kalletlak (approved):

PR #7859, opened 2020-09-02
Fix bug in data access token download.

ao508:
Argument for 'allowRemovalOfExistingTokens' has been disconnected or
removed in some way from the frontend. 

When a user attempts to download a token from the
login dropdown menu, if the option to revoke existing tokens was
enabled then the user would be presented with an error if they had an
existing token instead of being issued a brand new one.

We have decided to simply remove this functionality and always issue a
new token on request.


sheridancbio (changes requested):
Looks good. Two test cases should be dropped because the expected exception will no longer occur. Also, if there are no other references in the code base to MaxNumberTokensExceededException, then that class should be dropped as well. Plus there is a function now that serves no purpose and should be dropped.

Overall comment: we are being expedient here and sacrificing a feature which is not known to be needed in order to resolve the current bug created (most likely) when the oauth2 tokens were introduced. The price we are paying by doing this is that now users will have no way to protect their tokens from being deleted when they are at their quota (under uuid token method). Even a valid (unexpired) token will be deleted at the moment that a user requests a new token when they are at their quota. It may be (I don't remember how the frontend is coded) that every time a user chooses the dropdown menu in the header to retrieve their token, they will cause a current valid token to be invalidated (deleted) and a fresh new token will replace it. Maybe that is the desired behavior, although I suspect some users will use the dropdown menu as a way to "check" what their current token is, which will instead result in invalidation of their current token and issuance of a new one. A different behavior could be coded in the frontend later, where current tokens were retrieved and displayed using the GET /data-access-tokens endpoint ... and then the user could choose to delete tokens or create new ones through some kind of UI which used the POST and DELETE methods available.

web/src/main/java/org/cbioportal/web/DataAccessTokenController.java
```diff
@@ -129,13 +125,8 @@ private String getAuthenticatedUser(Authentication authentication) {
...
     }
 
-    private DataAccessToken createDataAccessToken(String userName, Boolean myAllowRevocationOfOtherTokens) {
-        DataAccessToken token;
-        if (myAllowRevocationOfOtherTokens != null) {
-            token = tokenService.createDataAccessToken(userName, myAllowRevocationOfOtherTokens);
-        }  else {
-            token = tokenService.createDataAccessToken(userName);
-        }
+    private DataAccessToken createDataAccessToken(String userName) {
```
sheridancbio, line 128: This function no longer serves any purpose. I suggest we drop this function and replace calls to it with a call to tokenService.createDataAccessToken(userName) instead.

ao508:
@sheridancbio Addressed your comments, PR has been updated. Thanks!

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -141,7 +141,6 @@ dat.unauth_users=
 dat.method=none
 dat.ttl_seconds=2592000
 dat.uuid.max_number_per_user=1
-dat.uuid.revoke_other_tokens=true
```
sheridancbio, line 144: great .. thanks for adding this change

web/src/main/java/org/cbioportal/web/error/GlobalExceptionHandler.java
```diff
@@ -151,12 +151,6 @@
         return new ResponseEntity<>(response, HttpStatus.UNAUTHORIZED);
     }
 
-    @ExceptionHandler(MaxNumberTokensExceededException.class)
```
sheridancbio, line 154: nice cleaning this up too

sheridancbio (approved):
👍

sheridancbio:
@ao508 because this is a change to the cbioportal API definition we should have at least one reviewer from another institution before merging. Also .. we should remember to update the generated java / js api clients in other repos based on the change this will reflect in the swagger spec.

pvannierop (approved):
@ao508 Ah I think I have broken this functionality when the token download mechanism was changed. I approve you changes since The Hyve does not rely on this. But I would like to put out there the possibility of not dropping this functionality. I repair the frontend and have it send this parameter to the backend again. Your PR will not be needed in that case. I am willing to help out here. Just let me know what you prefer.

ao508:
@pvannierop Just finished dicussing with Ben and I've confirmed that we're okay dropping this functionality. Thanks for the review!

ao508:
@n1zea144 This is okay to merge now

ao508:
@inodb I see the "labeler" and "end_to_end_tests_localdb" checks failing for multiple prs, is this already a known issue that can be ignored for now ?

pvannierop:
@inodb @ao508 @sheridancbio I will fix the e2e-tests today. We made modifications study_es_0 and this broke some tests. Does not affect functionality.

Issue #7280, opened 2020-03-09
Possibility to share created groups when sharing the study


zhx828:
For both physical study and virtual study, should we provide a way to share groups when user is trying to share the study? This request is coming from Niki and wondering whether that's easy to implement.

We tested and it's not supported at the moment.

jjgao:
It can be potential confusing, e.g. if I have a list of groups in msk_impact_2017 and than @schultzn send me his version, would his groups be added to mine? Should I be notified and have a choice? How can Niki decide whether to send the groups or not? What if there is a name clash? 

@schultzn could you elaborate your use case a little more?

schultzn:
I tried to share a virtual cohort with a set of groups pre-defined with somebody the other day. It would have been nice if the shared virtual cohort had also contained my groups. But I understand now that this is potentially difficult. 

I am sure it will come up again, but maybe it is ok to not support this for the moment.

jjgao:
Maybe we can add another button "Share" after "Filter"?

![image](https://user-images.githubusercontent.com/840895/76541722-5db1f780-645a-11ea-94e2-daeb2f0b15c3.png)

The "Share" button will create a link to the study with the selected groups to share, e.g.

https://www.cbioportal.org/study/summary?id=msk_impact_2017#sharedGroups=group1,group2

When a user open a link like this, the two groups will be added to their groups.

Naming clash is probably fine as long as they all have different IDs...

Any technical thoughts, @kalletlak @zhx828 ?

jjgao:
related: https://groups.google.com/forum/#!topic/cbioportal/Caix7Pw3quo

kalletlak:
How should we handle
1. invalid groups?
2. groups belonging to other study(studies) which are not part queried studies?

jjgao:
@kalletlak 

1. For invalid group, I think we can ignore them. How are we dealing with invalid group when a group is loaded now?

2. I think if the URL was constructed by the UI, this scenario is not possible - is that right? Of course, when a user manually change the URL, this could happen. In that case, I think we should just ignore the group when opening.

kalletlak:
> @kalletlak
> 
> 1. For invalid group, I think we can ignore them. How are we dealing with invalid group when a group is loaded now?

Currently we have no mechanism to to deal with them as they are not used directly.

[...]

PR #7862, opened 2020-09-07
Add 'color' attribute for group comparison groups received from session service

pvannierop:
This is part of [RFC55](https://docs.google.com/document/d/1nK7Z0qxg64fTVzP1RquuGry1u4oXwnjA9kB7bos0ihk/edit).

This PR will add a group color attribute to the model for for group comparison groups received from session service.

## Discuss
As far as I can imagine this does not break the backend when this prop is not yet returned by the session service. Am I correct in this? If so, this PR can be merged immediately.


kalletlak:
@pvannierop, from the RFC looks like group are editable, edit/add color to group. With the recent changes to support sharing groups(https://github.com/cBioPortal/cbioportal/issues/7280), we decided to treat groups as immutable. Looks like we already support default colors(color of pie slices) to group in comparison session. Instead we could use this color chooser only when creating new comparison session.

CC: @jjgao

jjgao:
@pvannierop should we reopen this? What's the next step?

pvannierop:
@jjgao We perhaps need to reopen when we have decided on a final implementation of this feature. When we are going to use the user session to persist the group colors, this PR may be obsolete. I propose to leave it closed for now.

PR #7895, opened 2020-09-22
Optimize mutation count by keyword calculation

ao508:
Improves runtime for calculation of mutation count by keyword during import.

Current implementation of this calculation is extremely time consuming
for studies with large and/or multiple MAFs.

With the changes in this PR the runtime for the calculation is reduced
from greater than 60 minutes to approximately 18 minutes for the
mskimpact study.

In short the changes in this PR implement the use of the bulk loader as
well as a step that pre-calculates the gene counts and mutation keyword
counts for the current study being imported before creating the insert
statements with the bulk loader and executing the import.

# Checks
- [ ] Runs on heroku

# Notify reviewers

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -40,6 +40,9 @@ import org.apache.commons.lang.StringUtils;
 import java.sql.*;
 import java.util.*;
 import java.util.regex.*;
+import org.apache.commons.collections.MapIterator;
+import org.apache.commons.collections.keyvalue.MultiKey;
+import org.apache.commons.collections.map.MultiKeyMap;
 
 /**
  * Data access object for Mutation table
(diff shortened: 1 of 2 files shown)


ao508:
there are basically two queries in one happening int he original statement - I've split it into 2 separate statements so that the calculation isn't not happening concurrently with the data insertion into `mutation_count_by_keyword`.

The first statement [here](https://github.com/cBioPortal/cbioportal/pull/7895/files?diff=unified&w=1#diff-6ee8e8336dc2fa1f8a3892830854031fR225) selects the mutation event keyword, entres gene id, and count of distinct samples with matching mutation event keywords and gene ids if the keyword is not null or  this value is set as 0 if the mutation event keyword *is* null. This makes a multikey map:

- key 1 = mutation event keyword
- key 2 = gene id 
- value = count of distinct sample ids for the (mutation event keyword, gene id) pair


The second statement [below](https://github.com/cBioPortal/cbioportal/pull/7895/files?diff=unified&w=1#diff-6ee8e8336dc2fa1f8a3892830854031fR251) is doing a simple count of distinct sample ids for each gene for the specified genetic profile id. This is a simple map of gene id -> count of distinct samples

When creating the records that will be inserted into `mutation_count_by_keyword` we are mapping the appropriate values form the second statement to the corresponding mutation event keyword/gene id pairs from the first statement.

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -186,27 +189,74 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
...
+        ResultSet rs = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoMutation.class);
+            pstmt = con.prepareStatement(
+                "SELECT ENTREZ_GENE_ID AS `ENTREZ_GENE_ID`, COUNT(DISTINCT(SAMPLE_ID)) AS `GENE_COUNT`" +
+                    " FROM mutation WHERE GENETIC_PROFILE_ID = ? " +
+                    "GROUP BY ENTREZ_GENE_ID;");
+            pstmt.setInt(1, geneticProfileId);
+            rs = pstmt.executeQuery();
+            while(rs.next()) {
```
divyamadala30, line 256: Add a space between 'while' and '('?

divyamadala30 (approved):
👍

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
ao508: fixed, thanks!

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -186,27 +189,74 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
...
-                    "GROUP BY g1.`GENETIC_PROFILE_ID` , m1.`ENTREZ_GENE_ID`) AS GENE_COUNT " +
-                    "FROM `mutation` AS m2 , `genetic_profile` AS g2 , `mutation_event` " +
-                    "WHERE m2.`GENETIC_PROFILE_ID` = g2.`GENETIC_PROFILE_ID` " +
-                    "AND m2.`MUTATION_EVENT_ID` = mutation_event.`MUTATION_EVENT_ID` " +
-                    "AND g2.`GENETIC_PROFILE_ID`=? " +
-                    "GROUP BY g2.`GENETIC_PROFILE_ID` , mutation_event.`KEYWORD` , m2.`ENTREZ_GENE_ID`;");
-            pstmt.setInt(1, profileId);
-            return pstmt.executeUpdate();
+                "SELECT ENTREZ_GENE_ID AS `ENTREZ_GENE_ID`, COUNT(DISTINCT(SAMPLE_ID)) AS `GENE_COUNT`" +
+                    " FROM mutation WHERE GENETIC_PROFILE_ID = ? " +
```
sheridancbio, line 252: yes ... here we can simply filter mutation table by genetic_profile_id ... rather than joining to genetic_profile first and then selecting records which match by genetic profile id (as before)

sheridancbio (commented):
👍

Issue #7905, opened 2020-09-24
Survival plots in comparison tab/tab: add a configuration/property to set “X-Axis Max” up to 120 months by default


jjgao:
... when the data is more than 120 months.

This default property should only be used for the **private-genie portal only** for now. 

![image](https://user-images.githubusercontent.com/840895/94202624-015d7500-fe8c-11ea-98da-8e290752eef0.png)

https://genie-private.cbioportal.org/comparison/survival?sessionId=5f63deb2e4b0242bd5d209d7

Issue #7805, opened 2020-08-18
Add a table of survival types to Survival tab in Comparison


jjgao:
The AACR GENIE BPC cohort may contain dozens of different types of survival data. The current interface of the survival tab (see below) in Comparison may not work well when showing many survival types. 

Here is a proposal:

- [x] Get rid of the current Survival Types selection section
![image](https://user-images.githubusercontent.com/840895/90559587-c751d280-e16b-11ea-8b44-9d322a31d761.png)
- [x] When there are more 2 or more survival types, add a table to the left of the survival plot (similar to Clinical tab), with the following columns: `Survival Type`, `# Patients`, `p-Value`, `q-Value`
  - note: we may have to pull all data to the frontend upfront since p-values are calculated in the frontend
- [x] Click each row will show the survival plot. (by default, the first one is selected)
- [x] Allow changing x-axis depending on x-range (should be some option), right now all of them have the same length

@cBioPortal/product

current layout:
![image](https://user-images.githubusercontent.com/840895/90559853-3fb89380-e16c-11ea-9e6d-c0220c081c91.png)

jjgao:
I forgot we haven't added p-values for survival analysis with 3 or more groups. @kalletlak has done some research in #7484 and it was not an easy task.

Maybe we can compute the pair-wise p-values and list the most significant ones? (not sure if q-value still holds - maybe we should not try to calculate q-values). We can also list some more data in the table, e.g. number of total cases and number of cases with event.

@kalletlak thoughts?

schultzn:
Why not show all p-values, so the p-value for each pair? Just trying to understand why this isn’t an easy task?

jjgao:
@schultzn we don't have the code to calculate p-values for three or more groups. Currently, the calculation is in frontend. We'll look into this more.

adamabeshouse:
@jjgao as far as the interface, maybe we should keep the current interface when there's only 2 survival types, like most studies? otherwise it will be a lot of wasted space..

jjgao:
> @jjgao as far as the interface, maybe we should keep the current interface when there's only 2 survival types, like most studies? otherwise it will be a lot of wasted space..

Good point. I think it's useful to have the p-values displayed even when there are only 2 types... Let's add the table first and see if we need to tweak when there are only a small number of survival curves (in most cases)?

jjgao:
@adamabeshouse some additional product requests from @schultzn 

[...]

PR #7913, opened 2020-09-28
Add portal configuration properties relating to survival tab

adamabeshouse:
Backend portion of https://github.com/cBioPortal/cbioportal/issues/7905 and point 3 in https://github.com/cBioPortal/cbioportal/issues/7805#issuecomment-698580622


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -199,6 +199,12 @@ mycancergenome.show=true
 # Enable transcript switch dropdown (true, false)
 # show.transcript_dropdown=false
 
+# Show/hide p- and q-values in survival types table (default is true)
+survival.show_p_q_values_in_survival_type_table=
```
alisman, line 203: better to have these commented out than set to nothing

alisman (changes requested):
You can merge once you comment out those two lines

PR #7937, opened 2020-10-06
Reimplemented gene set data test

### LINKED ISSUE #7536: Reimplement gene set data test
One of the tests broken when upgrading the test frameworks to be compatible with Java 11: https://github.com/cBioPortal/cbioportal/pull/7474/. It would be good to reimplement it to work with the new test frameworks

jtquach1:
Fix #7536 

- Fixed broken fetchGenesetData test for sampleListId and added assertions
- Split fetchGenesetData tests based on usage of sampleIds and sampleListId

### DIFF
--- a/web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
+++ b/web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
@@ -64,46 +64,82 @@ public class GenesetDataControllerTest {
     }
 
     @Test
-    public void fetchGeneticDataItems() throws Exception {
-
+    public void fetchGeneticDataItemsWithSampleIds() throws Exception {
         List<GenesetMolecularData> genesetDataItems = createGenesetDataItemsList();
-        Mockito.when(genesetDataService.fetchGenesetData(Mockito.any(), (List) Mockito.any(),
-            Mockito.any())).thenReturn(genesetDataItems);
+
+        String geneticProfileId = PROF_ID;
+        List<String> sampleIds = Arrays.asList(SAMPLE_STABLE_ID_1);
+        List<String> genesetIds = Arrays.asList(GENESET_ID_1, GENESET_ID_2);
+
+        Mockito.when(genesetDataService.fetchGenesetData(
+            geneticProfileId,
+            sampleIds,
+            genesetIds)
+        ).thenReturn(genesetDataItems);
 
         GenesetDataFilterCriteria genesetDataFilterCriteria = new GenesetDataFilterCriteria();
-        genesetDataFilterCriteria.setGenesetIds(Arrays.asList(GENESET_ID_1, GENESET_ID_2));
+        genesetDataFilterCriteria.setSampleIds(sampleIds);
+        genesetDataFilterCriteria.setGenesetIds(genesetIds);
 
-        mockMvc.perform(MockMvcRequestBuilders.post("/genetic-profiles/" + PROF_ID + "/geneset-genetic-data/fetch")
-                .accept(MediaType.APPLICATION_JSON)
-                .contentType(MediaType.APPLICATION_JSON)
-                .content(objectMapper.writeValueAsString(genesetDataFilterCriteria)))
-                .andExpect(MockMvcResultMatchers.status().isOk())
-                .andExpect(MockMvcResultMatchers.content().contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
-                .andExpect(MockMvcResultMatchers.jsonPath("$", Matchers.hasSize(2)))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[0].geneticProfileId").value(PROF_ID))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[0].genesetId").value(GENESET_ID_1))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[0].sampleId").value(SAMPLE_STABLE_ID_1))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[0].value").value(VALUE_1))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[1].geneticProfileId").value(PROF_ID))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[1].genesetId").value(GENESET_ID_2))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[1].sampleId").value(SAMPLE_STABLE_ID_1))
-                .andExpect(MockMvcResultMatchers.jsonPath("$[1].value").value(VALUE_2));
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/cbioportal/web/GenesetDataController.java
```diff
@@ -36,27 +36,39 @@
...
+        @ApiParam(required = true, value = "Genetic profile ID, e.g. gbm_tcga_gsva_scores")
+        @PathVariable String geneticProfileId,
+        @ApiParam(required = true, value = "Search criteria to return the values for a given set of samples and gene set items. "
+            + "genesetIds: The list of identifiers for the gene sets of interest, e.g. HINATA_NFKB_MATRIX. "
+            + "Use one of these if you want to specify a subset of samples:"
+            + "(1) sampleListId: Identifier of pre-defined sample list with samples to query, e.g. brca_tcga_all "
+            + "or (2) sampleIds: custom list of samples or patients to query, e.g. TCGA-BH-A1EO-01, TCGA-AR-A1AR-01")
+        @RequestBody GenesetDataFilterCriteria genesetDataFilterCriteria) throws MolecularProfileNotFoundException, SampleListNotFoundException {
+
+        if (sampleListIdExists(genesetDataFilterCriteria)) {
```
Luke-Sikina, line 51: This is a place where I think Clean Code is wrong. Pulling the predicate out into its own method when it's used in exactly one spot, and it's design is already clear is not worth it to me. I think this increases the mental effort needed to read this code, as you have to navigate to `sampleListIdExists` to understand the logic, and the logic isn't fully covered by the method name `sampleListIdExists`. I would change this back.

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
```diff
@@ -64,46 +64,77 @@ public void setUp() throws Exception {
...
     @Test
-    public void fetchGeneticDataItems() throws Exception {
+    public void fetchGeneticDataItemsWithSampleIds() throws Exception {
+        List<GenesetMolecularData> genesetDataItems = createGenesetDataItemsList();
+
+        String geneticProfileId = PROF_ID;
+        List<String> sampleIds = Arrays.asList(SAMPLE_STABLE_ID_1);
+        List<String> genesetIds = Arrays.asList(GENESET_ID_1, GENESET_ID_2);
+
+        // Argument values don't matter, just shows which fetchGenesetData is called
```
Luke-Sikina, line 74: Why don't they matter? I think they do. We are saying that when genesetDataService gets these args, it produces this value. We are setting up our expectations for this dependency in the context of this unit test.

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
```diff
@@ -64,46 +64,77 @@ public void setUp() throws Exception {
...
+        Mockito.when(genesetDataService.fetchGenesetData(
+            geneticProfileId,
+            sampleIds,
+            genesetIds)
+        ).thenReturn(genesetDataItems);
 
+        // Set sampleIds to ensure the fetchGenesetData variant above is called:
```
Luke-Sikina, line 81: I don't think this comment says anything useful. Trivial comments get ignored, and eventually age to the point that they're confusing or incorrect. I'd delete it.

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
```diff
@@ -64,46 +64,77 @@ public void setUp() throws Exception {
...
+
+        String geneticProfileId = PROF_ID;
+        List<String> sampleIds = Arrays.asList(SAMPLE_STABLE_ID_1);
+        List<String> genesetIds = Arrays.asList(GENESET_ID_1, GENESET_ID_2);
+
+        // Argument values don't matter, just shows which fetchGenesetData is called
+        Mockito.when(genesetDataService.fetchGenesetData(
+            geneticProfileId,
+            sampleIds,
+            genesetIds)
```
Luke-Sikina, line 78: Good!

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
```diff
@@ -64,46 +64,77 @@ public void setUp() throws Exception {
...
+    public void fetchGeneticDataItemsWithSampleListId() throws Exception {
         List<GenesetMolecularData> genesetDataItems = createGenesetDataItemsList();
-        Mockito.when(genesetDataService.fetchGenesetData(Mockito.any(), (List) Mockito.any(),
-            Mockito.any())).thenReturn(genesetDataItems);
 
+        String geneticProfileId = PROF_ID;
+        String sampleListId = SAMPLE_LIST_ID;
+        List<String> genesetIds = Arrays.asList(GENESET_ID_1, GENESET_ID_2);
+
+        // Argument values don't matter, just shows which fetchGenesetData is called
```
Luke-Sikina, line 111: Argument values matter :(

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
```diff
@@ -64,46 +64,77 @@ public void setUp() throws Exception {
...
+        genesetDataFilterCriteria.setSampleListId(sampleListId);
+        genesetDataFilterCriteria.setGenesetIds(genesetIds);
+
+        mockMvc.perform(MockMvcRequestBuilders.post("/genetic-profiles/" + geneticProfileId + "/geneset-genetic-data/fetch")
+            .accept(MediaType.APPLICATION_JSON)
+            .contentType(MediaType.APPLICATION_JSON)
+            .content(objectMapper.writeValueAsString(genesetDataFilterCriteria)))
+            .andExpect(MockMvcResultMatchers.status().isOk())
+            .andExpect(MockMvcResultMatchers.content().contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
+            .andExpect(MockMvcResultMatchers.jsonPath("$", Matchers.hasSize(2)))
```
Luke-Sikina, line 129: Why did this change from 4 to 2? I liked that the two tests verified different results. It helped mentally separate the two tests for me

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
jtquach1: deleted the comment

Luke-Sikina (approved):

web/src/test/java/org/cbioportal/web/GenesetDataControllerTest.java
jtquach1: deleted the comment

nr23730 (approved):
Thanks for solving this after I broke that test!

PR #7949, opened 2020-10-07
Update keycloak auth docs to specify how to get jwtRolesPath

averyniceday:
Had some issues when deploying GENIE with oauth2 DAT method. 

Ended up realizing it was an issue with jwtRolesPath. Thought it would help to have clearer docs explaining how to get the path out of the JWT. 

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


pvannierop (commented):
Correct me if I am wrong. Normally, when the user follows the standard installation instructions, the default value (_resource_access::cbioportal::roles_) would be ok, not not? If so, I think pointing the user to this while following standard setup instructions, is not the best place. I would propose to move your instruction to a new section before `### Troubleshooting` while explaining that this is only needed when deviating from default installation config.

What do you think?

averyniceday:
@pvannierop good idea, I'll move it to a troubleshooting section. 👍 Will try to update by early next week and ping you.

averyniceday:
@pvannierop updated the PR and moved the new stuff into the Troubleshooting section. Lmk if there's anything else! :)

docs/Authenticating-and-Authorizing-Users-via-keycloak.md
```diff
@@ -357,3 +357,32 @@ log4j.logger.org.springframework.security=DEBUG
 ```
 
 Then, rebuild the WAR, redeploy, and try to authenticate again.  Your log file will then include hundreds of SAML-specific messages, even the full XML of each SAML message, and this should help you debug the error.
+
+#### Determining jwtRolesPath for OAuth2 Token
+If the generated token returns a 4XX error when used, confirm that the `jwtRolesPath` property matches the roles path.
```
pvannierop, line 362: The trigger for looking at this section should not be a 4xx error. I propose to remove this line and replace it for something like this:

> By default user-roles are extracted from path `resource_access::cbioportal::roles` in the JWT json. When the configuration of roles at the realm and client level in a Keycloak instance differ from that described in this manual, configure the path can be specified using the `dat.oauth2.jwtRolesPath` property in the _portal.properties_ file.

averyniceday:
updated

pvannierop (approved):

pvannierop:
@averyniceday Approved!

PR #7950, opened 2020-10-09
add integration test to load study_es_0

SRodenburg:
This PR adds an integration test for the loading of `study_es_0` and associated gene panels, based on a local portal dump, and using the `cbioportal-docker-compose` solution.
To ensure the local repository is used instead of the docker image, we override the war file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,10 @@ env:
   - MAVEN_VERSION=3.5.4 TEST=python-validator
   - MAVEN_VERSION=3.5.4 TEST=sanity-checks
   - MAVEN_VERSION=3.5.4 TEST=end-to-end
+  - MAVEN_VERSION=3.5.4 TEST=integration-test-load-study
 install:
 - |
-  if [[ "${TEST}" == core || "${TEST}" == end-to-end ]]
+  if [[ "${TEST}" == core || "${TEST}" == end-to-end  || "${TEST}" == integration-test-load-study ]]
   then
       mkdir -p ~/maven
       test -d ~/maven/$MAVEN_VERSION/bin || { \
@@ -74,7 +75,7 @@ script:
           integration-test
   fi
 - |
-  if [[ "${TEST}" == end-to-end ]]
+  if [[ "${TEST}" == end-to-end || "${TEST}" == integration-test-load-study ]]
   then
       ~/maven/$MAVEN_VERSION/bin/mvn \
           -q \
@@ -96,4 +97,28 @@ script:
       # spot visual regression by comparing screenshots in the repo with
       # screenshots of this portal loaded with the data from the amazon db
       bash test/end-to-end/test_make_screenshots.sh test/end-to-end/screenshots.yml
+  fi
+- |
+  if [[ "${TEST}" == integration-test-load-study ]]
+  then
+      # test for loading study_es_0 on a complete working instance
+      # download compose and init
+      git clone https://github.com/cBioPortal/cbioportal-docker-compose.git && \
+      cd cbioportal-docker-compose && bash init.sh
+
+      # creates cbioportal container with local maven build
+      docker-compose -f docker-compose.yml -f ../test/integration/docker-compose-localbuild.yml up -d
+
+      # wait for up to 15m until all services are up and running
+      for i in {1..30}; do
+          [[ $(curl -sf http://localhost:8080/api/health) ]] && { healthy=1; break; } || echo "Waiting for cBioPortal services..."
+          sleep 30s
+      done
+      [ -z "$healthy" ] && { echo "Error starting cBioPortal services."; exit 1; } || echo "Successful deploy."
+
+      # load gene panels and study_es_0
+      bash ../test/integration/test_load_study.sh
+
+      # back to main dir
+      cd -
   fi
--- a/test/integration/docker-compose-localbuild.yml
+++ b/test/integration/docker-compose-localbuild.yml
@@ -1,0 +1,12 @@
+# this YAML file overrides settings in the cbioportal-docker-compose YAML
+# and maps a local maven build to appropriate locations in the container (which is retrieved by image).
+
+version: '3'
+
+services:
+  cbioportal:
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -96,4 +97,21 @@ script:
...
+  then
+      # test for loading study_es_0 on a complete working instance
+      # download compose and init
+      git clone https://github.com/cBioPortal/cbioportal-docker-compose.git && \
+      cd cbioportal-docker-compose && bash init.sh
+
+      # creates cbioportal container with local maven build
+      docker-compose -f docker-compose.yml -f ../test/integration/docker-compose-localbuild.yml up -d
+
+      sleep 60s
```
pvannierop, line 113: Should we not replace this with a proper wait cycle that checks ready-state?

test/integration/docker-compose-localbuild.yml
```diff
@@ -0,0 +1,9 @@
+version: '3'
+
+services:
+  cbioportal:
+    command: /bin/sh -c "java -Xms2g -Xmx4g -Dauthenticate=noauthsessionservice -Dsession.service.url=http://cbioportal_session:5000/api/sessions/my_portal/ -jar webapp-runner.jar -AmaxHttpHeaderSize=16384 -AconnectionTimeout=20000 --enable-compression /app.war"
+    volumes:
+     - ./portalinfo/:/portalinfo/
+     - ../:/cbioportal
+     - ../portal/target/cbioportal.war:/app.war
```
pvannierop, line 9: Do relative paths work. At least in v2.4 this gives problems. Or am I mistaken?

test/integration/docker-compose-localbuild.yml
```diff
@@ -0,0 +1,9 @@
+version: '3'
+
+services:
```
pvannierop, line 3: I think it would be a good idea to put a remark here on what this extension does. Pointing out that this mounts the current backend in order to test it would for sure help others that come across this code.

test/integration/test_load_study.sh
```diff
@@ -0,0 +1,26 @@
...
+        "$service" bash -c "$@"
+}
+
+# load panels
+run_in_service cbioportal "cd /cbioportal/core/src/main/scripts/ && perl importGenePanel.pl --data \
+                          /cbioportal/core/src/test/scripts/test_data/study_es_0/data_gene_panel_testpanel1.txt"
+run_in_service cbioportal "cd /cbioportal/core/src/main/scripts/ && perl importGenePanel.pl --data \
+                          /cbioportal/core/src/test/scripts/test_data/study_es_0/data_gene_panel_testpanel2.txt"
+
+# dump portal info
```
pvannierop, line 20: Now, you only test the route via portaldump, isn't it? I think it is a good idea to also add the study load option that uses the server url (-u option) so that validation is against a live portal, not the dump of the portal. Or is such a test redundant?

pvannierop (commented):
Lets discuss these comments.

test/integration/docker-compose-localbuild.yml
```diff
@@ -0,0 +1,12 @@
+# this YAML file overrides settings in the cbioportal-docker-compose YAML
+# and maps a local maven build to appropriate locations in the container (which is retrieved by image).
+
+version: '3'
+
+services:
+  cbioportal:
+    command: /bin/sh -c "java -Xms2g -Xmx4g -Dauthenticate=noauthsessionservice -Dsession.service.url=http://cbioportal_session:5000/api/sessions/my_portal/ -jar webapp-runner.jar -AmaxHttpHeaderSize=16384 -AconnectionTimeout=20000 --enable-compression /app.war"
```
inodb, line 8: is there a way to not override the command? Or does one always need to override the entire service?

inodb (commented):
Nice idea!

SRodenburg:
rebased. Let's see about the tests

test/integration/docker-compose-localbuild.yml
pvannierop: Well, the command is overridden since we run the WAR file itself, not the decompressed version. It removes the need for unzipping. The command does not have to be overridden for technical reasons, but I think that running the WAR directly is cleaner since it removes the need for a unzip in the travis.yml file. It is not necessary and may confuse people in the future. Remember that the only reason for running a unzipped version is that this allows to mount custom icons in an instance.

SRodenburg: we run the WAR file to override running the files in the cbioportal image

SRodenburg:
Can be merged I think

test/integration/test_load_study.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
inodb, line 1: this looks good to me, but would be nice to reorganize the test files a bit at some point. We have two copies of  `study_es_0` I believe (one in frontend, one in here). There's also duplication of commands to load this data, since they exist in frontend localdb as well. I guess we could use this script in the `localdb e2e tests` as well now?

test/integration/docker-compose-localbuild.yml
inodb: I see makes sense

inodb (commented):
Thanks for making this test 🙏! This will be super useful to make sure our recommended docker-compose installation keeps working as expected

Something seems to have gone wrong with rebase, but after that is fixed it looks ok to merge to me

Small side note. We mostly use Travis for legacy reasons. I think it's good to go though, but just for future reference. At the time when we evaluated different CI systems we found that CircleCI's container support was better, so it made it a lot easier to reproduce the testing environment locally. The workflow language to indicate dependencies on other steps is pretty useful as well. Would prefer to implement new test suites in circleci or alternatively as a github action (still experimenting with that)

test/integration/test_load_study.sh
pvannierop: @inodb there is no duplication of study_es_0 that I am aware of. E2e-tests of frontend operate on study_es_0 of the backend.

>There's also duplication of commands to load this data, since they exist in frontend localdb e2e tests as well. I guess we could use this script in the localdb e2e tests

Indeed, is one of the things to improve. I save this for when The Hyve has less projects for customers. Feel free to pick up this task if you want to.

pvannierop:
@inodb 

>Would prefer to implement new test suites in circleci or alternatively as a github action (still experimenting with that)

Ok, when you think in this direction we should have a 15 min talk on how to start doing this. We have more travisd integration tests in the pipeline.

SRodenburg:
@inodb can be merged?

PR #7960, opened 2020-10-13
Treatments API Tweaks

### LINKED ISSUE #7928: Should we turn on treatment functionality in study view on all portals?

Luke-Sikina:
Fix #7928

- Add new endpoint for determining if sample treatments should be displayed.
  This needed to be separated because for `gbm_columbia_2019`, sample treatments
  should not be displayed, but patient treatments should.
- Filter out unknown treatments from sample treatments API
- Switch from using internal ID to stable ID for the patient field

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/TreatmentServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/TreatmentServiceImpl.java
@@ -80,7 +80,7 @@ public class TreatmentServiceImpl implements TreatmentService {
      * treatments. Each call will move samples taken 
      */
     private static class TreatmentRowTriplet {
-        private final Set<ClinicalEventSample> pre, post, unknown;
+        private final Set<ClinicalEventSample> pre, post;
         private final String treatment;
 
         TreatmentRowTriplet(List<ClinicalEventSample> samples, String treatment) {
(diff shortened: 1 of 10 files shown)


web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -111,6 +111,7 @@
     public static final String GENERIC_ASSAY_META_FETCH_PATH = "/generic_assay_meta/fetch";
     public static final String TREATMENTS_PATIENT_PATH = "/treatments/patient";
     public static final String TREATMENTS_PATIENT_EXISTS_PATH = "/treatments/display";
```
kalletlak, line 113: May be rename it `display-patient`

service/src/main/java/org/cbioportal/service/impl/TreatmentServiceImpl.java
```diff
@@ -170,4 +169,12 @@ private PatientTreatmentRow createPatientTreatmentRowForTreatment(
     public Boolean containsTreatmentData(List<String> samples, List<String> studies) {
         return treatmentRepository.getTreatmentCount(samples, studies) > 0;
     }
+
+    @Override
+    public Boolean containsSampleTreatmentData(List<String> samples, List<String> studies) {
```
kalletlak, line 174: since `samples` is always null, why not remove it and pass null while invoking `getSampleCount` and `getTreatmentCount` methods?

kalletlak (approved):
LGTM

PR #7975, opened 2020-10-15
add travis_wait

SRodenburg:
Should prevent failing core tests by increasing the reserved time.

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received


SRodenburg:
Alternative we could remove the `-q` in the maven command for screen output

inodb:
I think I solved the travis issue, seems like removing the cache solved it

PR #7984, opened 2020-10-19
Dependency and plugin management

nr23730:
- Introduce dependency management
- Introduce plugin management

These two features by Maven ensure that the same versions of dependencies and plugins are used across the whole project including all submodules. 
When removing the dependencies from the pom.xml in the project root I had to add them to the submodules that require them, so that now it should be more clear which dependencies the submodules have.

# Notify reviewers
@inodb


nr23730:
After some issues with database related tests (I don't have a local test db), these seem to be fixed.
Right now cBioPortal uses versions 1.0.3 and 2.0.0M31 of the `spring-security-saml2-core` library, which seem to be incompatible.  For now I separated those, so still both versions can be included.

This is exactly why I would like to see dependency management here...

inodb (approved):
Thanks so much for sharing this @nr23730 ! This looks good to me. I'm not very familiar with mvn/plugin dependency management, but this does look better to me. Before we were using some variables in the pom to manage consistency of versions across submodules, but this seems like the more proper approach

Also note for people reviewing. You can ignore the whitespace changes with:

Screen Shot 2020-10-26 at 10 38 10 AM

nr23730:
@inodb Is there any chance getting this reviewed? If so, I would do a rebase these days to resolve the conflicts.

kalletlak (approved):
@nr23730 thank! Looks good to me. Yes, need to be rebased and make sure all the tests are passing.

Luke-Sikina (approved):

nr23730:
Done that. Also did upgrade the apache commons dependencies. 
The failing Travis CI tests seem to be related to rate limits so that resources could not be downloaded.
The failing Circle CI test is also failing on the current master branch, so it's probably not related to the changes of this PR.

inodb:
Thanks so much @nr23730 ! Sorry for the delay on this. I'm def no Java expert so appreciate the extra eyes @Luke-Sikina & @kalletlak !

PR #8053, opened 2020-11-11
GetStop doesn't NPE

Luke-Sikina:
Treatment getStop calculation was getting error in genie portal

```
SLF4J: Failed toString() invocation on an object of type [org.redisson.client.RedisConnection]
Reported exception:
	at org.cbioportal.model.Treatment.getStop(Treatment.java:46)
	at org.cbioportal.model.Treatment.hashCode(Treatment.java:67)
	at java.base/java.lang.Object.toString(Object.java:246)
	at java.base/java.lang.String.valueOf(String.java:2951)
```


inodb (approved):
Thanks!

PR #8066, opened 2020-11-16
Treatment SQL performs well on multi-study queries

### LINKED ISSUE #8063: Several count API endpoints very slow on v3.4.21 and v3.4.22
Try e.g. https://master.cbioportal.org/study/summary?id=5dd408f0e4b0f7d2de7862a8. A lot of endpoints take over 40 seconds:

Screen Shot 2020-11-16 at 10 29 19 AM

In production, which is running v3.4.20, the slowest endpoint is gene panel data:

Screen Shot 2020-11-16 at 10 30 36 AM

Luke-Sikina:
Fix #8063

- Filter cancer study before filtering by study, sample id pairs

### DIFF
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/TreatmentMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/TreatmentMapper.xml
@@ -99,7 +99,11 @@
                     </foreach>
                 </if>
                 <if test="@java.util.Arrays@stream(studyIds.toArray()).distinct().count() > 1">
-                    (cancer_study.CANCER_STUDY_IDENTIFIER, sample.STABLE_ID) IN
+                    cancer_study.CANCER_STUDY_IDENTIFIER IN
+                    <foreach item="item" collection="@java.util.Arrays@stream(studyIds.toArray()).distinct().collect(@java.util.stream.Collectors.toList())" open="(" separator="," close=")">
+                        #{item}
+                    </foreach>
+                    AND (cancer_study.CANCER_STUDY_IDENTIFIER, sample.STABLE_ID) IN
                     <foreach index="i" collection="sampleIds" open="(" separator="," close=")">
                         (#{studyIds[${i}]}, #{sampleIds[${i}]})
                     </foreach>
--- a/web/src/main/java/org/cbioportal/web/util/StudyViewFilterApplier.java
+++ b/web/src/main/java/org/cbioportal/web/util/StudyViewFilterApplier.java
@@ -240,14 +240,20 @@ public class StudyViewFilterApplier {
             }
         }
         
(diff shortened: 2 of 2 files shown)


persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/TreatmentMapper.java
```diff
@@ -7,13 +7,13 @@
...
 
 public interface TreatmentMapper {
-    List<Treatment> getAllTreatments(List<String> sampleIds, List<String> studyIds);
+    List<Treatment> getAllTreatments(List<String> sampleIds, List<String> studyIds, List<String> uniqueStudyIds);
 
-    List<ClinicalEventSample> getAllSamples(List<String> sampleIds, List<String> studyIds);
+    List<ClinicalEventSample> getAllSamples(List<String> sampleIds, List<String> studyIds, List<String> uniqueStudyIds);
 
-    Set<String> getAllUniqueTreatments(List<String> sampleIds, List<String> studyIds);
+    Set<String> getAllUniqueTreatments(List<String> sampleIds, List<String> studyIds, List<String> uniqueStudyIds);
```
inodb, line 14: might be cleaner if the unique/distinct step can be done in mysql in some way. Looks slightly odd when studyIds are given as argument twice imo

inodb (approved):
Thanks! This seems like it could work (didn't fully test, but easier to try in master once merged)

I wonder if we should also change in the frontend how we load these counts given that all rendering/display currently waits for all data to be loaded. Maybe with these fixes the latency is minor, but just curious if u had any thoughts on that @Luke-Sikina

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/TreatmentMapper.xml
```diff
@@ -99,7 +99,11 @@
                     </foreach>
                 </if>
                 <if test="@java.util.Arrays@stream(studyIds.toArray()).distinct().count() > 1">
-                    (cancer_study.CANCER_STUDY_IDENTIFIER, sample.STABLE_ID) IN
+                    cancer_study.CANCER_STUDY_IDENTIFIER IN
+                    <foreach item="item" collection="uniqueStudyIds" open="(" separator="," close=")">
```
kalletlak, line 103: Could you try this and see if it works
```
 <foreach item="item" collection="@java.util.Arrays@stream(studyIds.toArray()).distinct().collect(@java.util.stream.Collectors.toList())" open="(" separator="," close=")">
    #{item}
 </foreach>
```

Luke-Sikina: This also works, and this fixes Ino's concerns. Updated to match.

kalletlak (approved):

PR #8099, opened 2020-12-04
Compress problematically large request bodies

### LINKED ISSUE #8026: GENIE performance - upload speed
## Problem
GENIE was being super slow for me today, so I did a little poking around (with some help from @Luke-Sikina ). Perhaps you are already aware of this, but I found that most of the time was spent on the request upload: this is the gene-panel-data API call, which was the slowest component of loading the Cancer Types Summary page
image
(https://bit.ly/38hPy1R)

Several other API calls also took several minutes, and for all of them the time was almost all spent on sending the request.

Granted, my upload speed was particularly terrible at the time I was doing this, but, I still want to flag this as we think about performance improvements for GENIE.

## Proposed Solution
The poor performance that Tali was seeing was due to her low upload speed and the large size of some of our request bodies. 

### Short Term (Will close this issue)
- [x] Create a Filter that will wrap incoming http requests if they are a post and the content encoding is gzipped
  - [x] Simple implementation
  - [x] Verification
  - [x] Make sure you can't zip bomb us too easily
  - [x] Tests
  - [x] Code cleanup (Right now we have to read the whole request input stream, which means we have to parse the args ourselves. I'd rather not do this.)
  - [x] Documentation: portal.properties.example, and any markdown files that seem relevant 
- [x] Figure out how to gzip the body of select request on the frontend
- [x] Make configurable
### Long Term
Why do we have to send such large request bodies? This information originated in the backend, so can't we have a shared, abbreviated representation of the data that we can pass?

Luke-Sikina:
Fix #8026

Problem:
- Some request bodies were so large (20+ Mb) that they resulted in
significantly degraded performance for people with poor upload speeds

Solution:
- Create RequestBodyGZipFilter. It:
  - Catches requests with a `Content-Encoding: gzip` header
  - Reads and inflates the input stream
  - Prevents excessively large requests from consuming too much memory
- Add filter to web.xml
- Add configurations for filter in portal.properties
  - Feature toggle
  - Inflated request body size limit
- Add Content-Encoding header to list of accepted headers for
Cors filter.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/util/RequestBodyGZipFilter.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/util/RequestBodyGZipFilter.java
@@ -1,0 +1,79 @@
+package org.mskcc.cbio.portal.util;
+
+import com.google.common.net.HttpHeaders;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@Component("requestBodyGZipFilter")
+public class RequestBodyGZipFilter implements Filter {
+    /**
+     * We need to limit the size of the gzipped request bodies to avoid denial of memory attacks.
+     * Because gzip can achieve text compression rates of around 1000x, without a common sense limit for
+     * request body size, it would be very easy to send a request body that, when inflated, consumes the
+     * entirety of a server's memory.
+     */
+    @Value("${request_gzip_body_size_bytes:80000000}")
+    private int maxInflatedRequestBodySize;
+    
+    @Value("${enable_request_body_gzip_compression:false}")
+    private boolean enabled;
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        // Nothing to init (Sonar asks you to comment if you leave a method empty)
+    }
+
+    /**
+     * Analyzes servlet request for possible gzipped body.
+     * When Content-Encoding header has "gzip" value and request method is POST, we read all the
+     * gzipped stream and if it has any data unzip it. In case when gzip Content-Encoding header
+     * specified but body is not actually in gzip format we will throw ZipException.
+     *
+     * @param servletRequest  servlet request
+     * @param servletResponse servlet response
+     * @param chain           filter chain
+     * @throws IOException      throws when fails
+     * @throws ServletException thrown when fails
+     */
+    @Override
+    public final void doFilter(
+        final ServletRequest servletRequest,
+        final ServletResponse servletResponse,
+        final FilterChain chain
+    ) throws IOException, ServletException {
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+        HttpServletResponse response = (HttpServletResponse) servletResponse;
+        boolean isGzipped = request.getHeader(HttpHeaders.CONTENT_ENCODING) != null
+            && request.getHeader(HttpHeaders.CONTENT_ENCODING).contains("gzip");
+        boolean requestTypeSupported = "POST".equals(request.getMethod());
+
+        if (isGzipped) {
+            if (!enabled) {
+                throw new IllegalStateException(
+                    "Received gzipped request body, but enable_request_body_gzip_compression is not set or is false"
+                );
+            }
+            if (!requestTypeSupported) {
+                throw new IllegalStateException(
+                    "Received request with a Content-Encoding: gzip header and a request method of: " +
+                    request.getMethod()
+                    + " Only POST requests are supported."
+                );
+            }
+
+            request = new GzippedInputStreamRequestWrapper((HttpServletRequest) servletRequest, maxInflatedRequestBodySize);
+        }
+
+        chain.doFilter(request, response);
+    }
+    
+    @Override
+    public final void destroy() {
+        // Nothing to destroy (Sonar asks you to comment if you leave a method empty)
+    }
+}
--- a/core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
+++ b/core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
@@ -1,0 +1,174 @@
+package org.mskcc.cbio.portal.util;
+
+import com.google.common.net.HttpHeaders;
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.springframework.test.util.ReflectionTestUtils;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletInputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.HttpMethod;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.stream.Collectors;
+
+public class TestRequestBodyGZipFilter {
+    // This is the byte representation of a gzipped json file with the following contents: {"field":"value"}
+    private static final byte[] GZIPPED_JSON_FILE_BYTES = {
+        31, -117, 8, 8, -58, -49, -50, 95, 0, 3, 103, 122, 105, 112, 95, 102, 105, 108, 116, 101, 114, 46, 106, 115,
+        111, 110, 0, -85, 86, 74, -53, 76, -51, 73, 81, -78, 82, 42, 75, -52, 41, 77, 85, -86, -27, 2, 0, 47, 99,
+        109, 9, 18, 0, 0, 0
+    };
+
+    /**
+     * When I filter a request that doesn't have the gzip header,
+     * the request should not be wrapped, and chain.doFilter should
+     * be called on the same request object passed to the method.
+     */
+    @Test
+    public void testDoFilterEnabledNotGZipped() throws Exception {
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 50000000);
+
+        HttpServletRequest request = createHttpServletRequest(null, HttpMethod.POST);
+        HttpServletResponse response = Mockito.mock(HttpServletResponse.class);
+        FilterChain chain = Mockito.mock(FilterChain.class);
+
+        subject.doFilter(request, response, chain);
+        
+        // request should not be wrapped
+        Mockito.verify(chain, Mockito.times(1)).doFilter(request, response);
+    }
+
+    /**
+     * When I filter a request that has the gzip header, but the gzip
+     * feature is disabled, the doFilter method should throw an exception.
+     */
+    @Test(expected = IllegalStateException.class)
+    public void testDoFilterDisabledGZipped() throws Exception {
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(false, 50000000);
+
+        HttpServletRequest request = createHttpServletRequest("gzip", HttpMethod.POST);
+        HttpServletResponse response = Mockito.mock(HttpServletResponse.class);
+        FilterChain chain = Mockito.mock(FilterChain.class);
+
+        subject.doFilter(request, response, chain);
+    }
+
+    /**
+     * When I filter a request that has the gzip header and the gzip
+     * feature is enabled, but the method is something other than POST,
+     * the doFilter method should throw an exception.
+     */
+    @Test(expected = IllegalStateException.class)
+    public void testDoFilterEnabledGZippedBadMethod() throws Exception {
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 50000000);
+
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/mskcc/cbio/portal/util/RequestBodyGZipFilter.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    
+    @Override
+    public final void destroy() {
+        // Nothing to destroy (Sonar asks you to comment if you leave a method empty)
+    }
+
+    static class GzippedInputStreamWrapper extends HttpServletRequestWrapper {
+        private final int maxInflatedRequestBodySize;
+        private final CountingInputStream inputStream;
```
Luke-Sikina, line 87: `CountingInputStream` is marked as beta, but it works well, and I haven't found a more mature solution. Open to suggestions here.

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+     * The BufferedReader for this GzippedInputStreamWrapper should return
+     * the contents of the gzipped file used to create the original input stream.
+     */
+    @Test
+    public void testDoFilterEnabledGZipped() throws Exception {
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 50000000);
+
+        // Not thrilled to be using the file system, but this was the most readable way to do this
+        // that I could think of.
+        // gzip_filter.json.gz is a gzip of a json file that contains the following: {"field":"value"}
```
Luke-Sikina, line 84: Interested if anyone has an idea on how to do this better.

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+
+
+    private RequestBodyGZipFilter createRequestBodyGZipFilter(boolean enabled, int requestBodySize) {
+        RequestBodyGZipFilter subject = new RequestBodyGZipFilter();
+        ReflectionTestUtils.setField(subject, "maxInflatedRequestBodySize", requestBodySize);
+        ReflectionTestUtils.setField(subject, "enabled", enabled);
+        return subject;
+    }
+    
+    private HttpServletRequest createHttpServletRequest(String gzip, String post) {
```
sheridancbio, line 151: the "post" argument name is improper ... sometimes the string coming in is "GET", so rename to "command" or "httpCommand"

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * When I filter a request that doesn't have the gzip header,
+     * the request should not be wrapped, and chain.doFilter should
+     * be called on the same request object passed to the method.
+     */
+    @Test
+    public void testDoFilterEnabledNotGZipped() throws Exception {
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 50000000);
+
+        HttpServletRequest request = createHttpServletRequest(null, "POST");
```
sheridancbio, line 30: The accepted strings for httpMethod can be referenced symbolically here perhaps? (from quick web search):
https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/HttpMethod.html

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+     * feature is enabled and the method is POST, the doFilter method
+     * should wrap the request with a GzippedInputStreamWrapper.
+     * The BufferedReader for this GzippedInputStreamWrapper should return
+     * the contents of the gzipped file used to create the original input stream.
+     */
+    @Test
+    public void testDoFilterEnabledGZipped() throws Exception {
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 50000000);
+
+        // Not thrilled to be using the file system, but this was the most readable way to do this
```
sheridancbio, line 82: If we have not come up with a better way by the time we are ready to merge, we should drop this comment maybe

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
sheridancbio: The total file size is 55 bytes .. perhaps it is reasonable to encode it directly as a byte array in code using escaped byte literals? The file is not human readable in any case (although I guess as a file it can be unzipped at the command line to "look inside" .. but the comment above handles that in the code.

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+        // that I could think of.
+        // gzip_filter.json.gz is a gzip of a json file that contains the following: {"field":"value"}
+        FileInputStream inputStream = new FileInputStream("src/test/resources/gzip_filter.json.gz");
+        TestInputStream stream = new TestInputStream(inputStream);
+        HttpServletRequest request = createHttpServletRequest("gzip", "POST");
+        Mockito.when(request.getInputStream()).thenReturn(stream);
+        
+        HttpServletResponse response = Mockito.mock(HttpServletResponse.class);
+        FilterChain chain = Mockito.mock(FilterChain.class);
+
```
sheridancbio, line 92: don't need 2 blank lines here

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * When I filter a request that has the gzip header and the gzip
+     * feature is enabled and the method is POST, but the unzipped message
+     * body is larger than maxInflatedRequestBodySize, the doFilter method
+     * should wrap the request with a GzippedInputStreamWrapper.
+     * The BufferedReader for this GzippedInputStreamWrapper should throw an
+     * IOException as it is read. This exception gets turned into
+     * an UncheckedIOException by something, so we'll check for that instead.
```
sheridancbio, line 115: mysterious

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+
+        // Request should be wrapped, so the request should now be a GzippedInputStreamWrapper
+        ArgumentCaptor<RequestBodyGZipFilter.GzippedInputStreamWrapper> requestArg =
+            ArgumentCaptor.forClass(RequestBodyGZipFilter.GzippedInputStreamWrapper.class);
+        ArgumentCaptor<HttpServletResponse> responseArg = ArgumentCaptor.forClass(HttpServletResponse.class);
+        Mockito.verify(chain, Mockito.times(1)).doFilter(requestArg.capture(), responseArg.capture());
+        RequestBodyGZipFilter.GzippedInputStreamWrapper wrappedStream = requestArg.getValue();
+
+        //this should cause an IOException
+        wrappedStream.getReader().lines().collect(Collectors.joining());
```
sheridancbio, line 140: since the test is working we can conclude that the Reader returned gets used with only the read() method by .lines()
this construction at least is not using read(byte[])

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 1);
+
+        FileInputStream inputStream = new FileInputStream("src/test/resources/gzip_filter.json.gz");
+        TestInputStream stream = new TestInputStream(inputStream);
+        HttpServletRequest request = createHttpServletRequest("gzip", "POST");
+        Mockito.when(request.getInputStream()).thenReturn(stream);
+
+        HttpServletResponse response = Mockito.mock(HttpServletResponse.class);
+        FilterChain chain = Mockito.mock(FilterChain.class);
+
```
sheridancbio, line 128: double blank line

docs/portal.properties-Reference.md
```diff
@@ -416,3 +417,44 @@ For more information on how Ehcache is implemented in cBioPortal refer to the [C
...
 skin.show_gsva=true
 ```
+
+# Request Body Compression
+
+## Background
+Some REST endpoints that the cBioPortal frontend uses have request bodies that scale as your dataset increases. In
+portals where users commonly query more than 100,000 samples, we found that some of these request bodies could
+get as large as 20 Mb. These large request bodies pose a significant problem for users with poor upload speeds - some
+users experienced upload times of more than five minutes for these requests. Request body compression is our solution to
```
sheridancbio, line 427: maybe "short-term solution" (because when sample lists are 300000 samples we will again see this problem even with compression)

docs/portal.properties-Reference.md
```diff
@@ -416,3 +417,44 @@ For more information on how Ehcache is implemented in cBioPortal refer to the [C
...
+provided:
+  - It is a POST request.
+  - It has a `Content-Encoding: gzip` header.
+
+## Reasons to Enable This Feature
+- You have studies with tens of thousands of samples.
+- You have users with poor upload speeds (< 1mb up).
+
+## Reasons to Disable This Feature
+- It is harder to debug gzipped requests
```
sheridancbio, line 448: good catch / comment

docs/portal.properties-Reference.md
```diff
@@ -416,3 +417,44 @@ For more information on how Ehcache is implemented in cBioPortal refer to the [C
...
+- It is a potential vector for denial of memory attacks.
+  - Any request that has a body that takes significantly more space in memory than it does in the request body is
+    potentially dangerous. We try to mitigate this by limiting the size of the unzipped request body via the
+    `request_gzip_body_size_bytes` property, but at a fundamental level, this is still a concern.
+  - Along these lines, if you do enable this feature, setting `request_gzip_body_size_bytes` to an arbitrarily large
+    number would be unwise.
+- This is not a cure-all for performance issues
+  - Most requests the cBioPortal makes do not have large request bodies, so most requests will not be compressed, and
+    will see no performance improvement.
+  - Users with good upload speeds will see minimal performance improvements, as their upload speed is not a bottleneck.
```
sheridancbio, line 460: right .. we need to find a way to stop sending very long identifier lists if we really want to improve performance for everybody

portal/src/main/webapp/config_service.jsp
```diff
@@ -114,6 +114,7 @@
             "oncoprint.custom_driver_annotation.tiers.default",
             "ensembl.transcript_url",
             "enable_persistent_cache",
+            "enable_request_body_compression",
```
sheridancbio, line 117: I guess this reports the setting of the property to the frontend perhaps?  That way the front end will know whether it can use compression or not. Ok

sheridancbio (approved):
Made a few suggestions and a few requested changes, but no show-stoppers. So approving, but will wait before merging to allow any improvements to be incorporated.

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
```diff
@@ -0,0 +1,174 @@
...
+        RequestBodyGZipFilter subject = createRequestBodyGZipFilter(true, 50000000);
+
+        HttpServletRequest request = createHttpServletRequest(null, "POST");
+        HttpServletResponse response = Mockito.mock(HttpServletResponse.class);
+        FilterChain chain = Mockito.mock(FilterChain.class);
+
+        subject.doFilter(request, response, chain);
+        
+        // request should not be wrapped
+        Mockito.verify(chain, Mockito.times(1)).doFilter(request, response);
```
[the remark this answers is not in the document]
Luke-Sikina: Prior to you asking, I didn't, because I was pretty confident that this worked. Now that you asked though, I have verified that this test fails under that condition. If `doFilter` is not called with those exact `request` and `response` objects, the test fails.

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
Luke-Sikina: I like it. Thanks for the suggestion.

core/src/test/java/org/mskcc/cbio/portal/util/TestRequestBodyGZipFilter.java
Luke-Sikina: This is what happens when you click through Intellij's method refactoring too quickly. I've changed it to `httpMethod`

portal/src/main/webapp/config_service.jsp
Luke-Sikina: Exactly.

PR #8104, opened 2020-12-07
Add schema changes for custom driver annotations

inodb:
This adds several indices for the custom driver annotation implementaton


inodb:
Decided to hold of on merging this until we have tested it more thoroughly

PR #8106, opened 2020-12-08
support a non-default MySQL database port in 'mysqladmin ping` command

dtenenba:

- Added port flag (`-P`) to `mysqladmin ping` command in Docker entrypoint script. This was done because at my institution we would like our production deployment to use a MySQL database that is provided by a DBaaS service that we have which creates databases on non-default ports (instead of using a containerized database).

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/docker/web-and-data/docker-entrypoint.sh
+++ b/docker/web-and-data/docker-entrypoint.sh
@@ -37,11 +37,11 @@ check_db_connection() {
     eval "$(parse_db_params_from_config_and_command_line $@ | sed 's/^db\./db_/g')"
     POTENTIAL_DB_PARAMS=$@
 
-    while ! mysqladmin ping -s -h$(echo ${db_host} | cut -d: -f1) -u${db_user} -p${db_password};
+    while ! mysqladmin ping -s -h$(echo ${db_host} | cut -d: -f1) -P$(echo ${db_host} | cut -d: -f2) -u${db_user} -p${db_password};
     do
         sleep 5s;
         if [ -n "$SHOW_DEBUG_INFO" ] && [ "$SHOW_DEBUG_INFO" != "false" ]; then
-            echo mysqladmin ping -s -h$(echo ${db_host} | cut -d: -f1) -u${db_user} -p${db_password}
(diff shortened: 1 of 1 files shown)


dtenenba:
I believe the check failure in end_to_end_tests is not related to the change I made, as the`mysqladmin ping` command, which is the only thing I changed, completed successfully and the failure occurred after that.

docker/web-and-data/docker-entrypoint.sh
```diff
@@ -37,11 +37,11 @@ check_db_connection() {
     eval "$(parse_db_params_from_config_and_command_line $@ | sed 's/^db\./db_/g')"
     POTENTIAL_DB_PARAMS=$@
 
-    while ! mysqladmin ping -s -h$(echo ${db_host} | cut -d: -f1) -u${db_user} -p${db_password};
+    while ! mysqladmin ping -s -h$(echo ${db_host} | cut -d: -f1) -P$(echo ${db_host} | cut -d: -f2) -u${db_user} -p${db_password};
```
inodb, line 40: Thanks so much for your contribution @dtenenba !

It might be nice to handle one more corner case here. It is possible the db_host does not contain a port (the `:` part). Could you maybe separate this out to a variable `$db_port` that's set to 3306 in case `localhost` does not contain `:`?

inodb: The localdb test might actually be failing correctly, because the host name there doesn't have a `:` in the hostname: https://github.com/cBioPortal/cbioportal-frontend/blob/master/end-to-end-test/local/runtime-config/portal.properties#L5. There are however two subtests in there that are failing now as part of that routine, but it should be able to at least start the tests

dtenenba:
Thanks @inodb for the suggestions and troubleshooting. I have addressed them in the latest commit.
Now it will use `$db_port` if set. If not it will check to see if `$db_host` has a colon and use whatever comes after the colon, otherwise it will use the default mysql port (3306).

inodb (approved):
Thanks so much for your contribution @dtenenba ! Nice work!

PR #8112, opened 2020-12-16
Fix Scripts Jar

Luke-Sikina:
- This is the first time we've made a servlet filter that uses the
application context. Turns out this breaks the scripts jar, because
that jar doesn't know about the Filter class, and when you run the
importer, it will try to wire up your filter anyways
- Added the javax.servlet dependecy to the scripts jar


inodb (approved):
Seems like a good solution to me, thanks!

PR #8220, opened 2020-12-18
Sample Treatments Work Fix

Luke-Sikina:
- Sample treatments still work if the only clinical event for each sample is
a clinical event of type `SEQUENCING`
- In general, sample treatments are more flexible, and will work as long as there is
at least one clinical event for a sample


inodb (approved):
Looks good to me! Tests demonstrate correctness 👍

PR #8234, opened 2020-12-23
Fix slow mutation mapper query

alisman:


Luke-Sikina:
Duplicate. See https://github.com/cBioPortal/cbioportal/pull/8229

PR #8252, opened 2021-01-06
Redis Support for cBioPortal

averyniceday:
Added support for Redis (in addition to existing EhCache system)

- Properties changed to accept a base cache type
- Cache naming now depends on application/portal name
- System supports multiple apps sharing one redis instance for caching
- Updates to caching documentation

Here's a screenshot from grafana showing memory usage with traffic on genie deployed with redis. We can see that the memory usage plateaus at a relatively low level without stairstepping. This is because cached items are now stored on a separate node and redis instance (instead of in-memory like ehcache).

Screen Shot 2020-12-09 at 1 08 52 PM

Here's a screenshot without redis. 

Screen Shot 2020-12-08 at 7 04 04 AM

Related, we have created a new instance group for genie portal and redis to avoid issues where node is overloaded on memory due to multiple pods spiking on same node). Seperate PR will be created in the K8S repo. 

Squash during merge!! 

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


inodb (approved):
Thanks so much @averyniceday ! Great to see this come to fruition! Nice work!

Just for my understanding, both grafana images are for instances with cache enabled?

The top is using redis on separate nodes:

image

The bottom image is using ehcache:

image

Correct?

Second question is regarding the top image with redis on separate nodes. It seems like there is still a slight increase in memory over time? Def pretty minimal so it's prolly fine when there's more traffic as well but just curious if that's what you've observed

Last question is re performance differences. I know you previously benchmarked caching vs no caching but curious if the results are the same. Maybe just having a few stats of genie v8.1 study page time load with caching vs no caching. Could be nice to mention in the docs as well to have a few ballpark numbers

averyniceday:
image

Leaving this here for now. We ran several tests using requests collected from the genie portal through hrrs and replayed them against our own test deployment without cache, with ehcache, and with redis. Redis performed similarly, if not slightly better than ehcache across most of the endpoints.

sheridancbio (approved):
This adds another option for persistence layer caching which is configured to run outside of the JVM. Hopefully this can help us with the memory resource / garbage collection issues we have been having in our backend clusters.

averyniceday:
Forgot to open this as a draft PR first, but opened up another PR from same branch and the e2e localdb and travis tests passed. 

https://github.com/cBioPortal/cbioportal/pull/8261

Screen Shot 2021-01-08 at 12 48 57 PM

PR #8274, opened 2021-01-13
Skip records with empty driver and tiers fields (custom_driver_annotation table)

pvannierop:
## Current behavior
When both DRIVER_FILTER and DRIVER_TIERS_FILTER are empty a record is incorrectly imported into the _custom_driver_annotation_ table.

## Problem
_custom_driver_annotation_ table contains many useless records. This was not according to desired behavior when this feature was created.

## Solution
JAVA importer skips annotations when both DRIVER_FILTER and DRIVER_TIERS_FILTER are 1) null, 2) empty string or 3) 'NA' or 'na'.

Note: I decided not to add this change to migration.sql because the feature is still not announced to the public.


inodb (approved):
Thanks for the quick fix!

PR #8275, opened 2021-01-14
Skip records with empty driver and tiers fields (custom_driver_annotation table) -- UPDATED

pvannierop:
Note: one commit in the previous PR with the same title was not included. This PR corrects this. 

## Current behavior
When both DRIVER_FILTER and DRIVER_TIERS_FILTER are empty a record is incorrectly imported into the _custom_driver_annotation_ table.

## Problem
_custom_driver_annotation_ table contains many useless records. This was not according to desired behavior when this feature was created.

## Solution
JAVA importer skips annotations when both DRIVER_FILTER and DRIVER_TIERS_FILTER are 1) null, 2) empty string or 3) 'NA' or 'na'.

Note: I decided not to add this change to migration.sql because the feature is still not announced to the public.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
@@ -64,8 +64,13 @@ public final class DaoCnaEvent {
                     Integer.toString(cnaEvent.getCnaProfileId())
                     );
 
-            if ((cnaEvent.getDriverFilter() != null && !cnaEvent.getDriverFilter().isEmpty())
-                || (cnaEvent.getDriverTiersFilter() != null && !cnaEvent.getDriverTiersFilter().isEmpty())) {
+            if ((cnaEvent.getDriverFilter() != null
+                && !cnaEvent.getDriverFilter().isEmpty()
+                && !cnaEvent.getDriverFilter().toLowerCase().equals("na"))
+                || 
+                (cnaEvent.getDriverTiersFilter() != null
+                && !cnaEvent.getDriverTiersFilter().isEmpty()
+                && !cnaEvent.getDriverTiersFilter().toLowerCase().equals("na"))) {
                 MySQLbulkLoader.getMySQLbulkLoader("alteration_driver_annotation").insertRecord(
                     Long.toString(eventId),
                     Integer.toString(cnaEvent.getCnaProfileId()),
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -32,17 +32,34 @@
 
 package org.mskcc.cbio.portal.dao;
 
-import org.mskcc.cbio.portal.model.*;
-import org.mskcc.cbio.portal.util.MutationKeywordUtils;
-
-import org.apache.commons.lang.StringUtils;
-
-import java.sql.*;
-import java.util.*;
-import java.util.regex.*;
 import org.apache.commons.collections.MapIterator;
 import org.apache.commons.collections.keyvalue.MultiKey;
 import org.apache.commons.collections.map.MultiKeyMap;
+import org.apache.commons.lang.StringUtils;
+import org.mskcc.cbio.portal.model.CancerStudy;
+import org.mskcc.cbio.portal.model.CanonicalGene;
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
```diff
@@ -64,8 +64,13 @@ public static void addCaseCnaEvent(CnaEvent cnaEvent, boolean newCnaEvent) throw
...
 
-            if ((cnaEvent.getDriverFilter() != null && !cnaEvent.getDriverFilter().isEmpty())
-                || (cnaEvent.getDriverTiersFilter() != null && !cnaEvent.getDriverTiersFilter().isEmpty())) {
+            if ((cnaEvent.getDriverFilter() != null
+                && !cnaEvent.getDriverFilter().isEmpty()
+                && !cnaEvent.getDriverFilter().toLowerCase().equals("na"))
+                || 
+                (cnaEvent.getDriverTiersFilter() != null
+                && !cnaEvent.getDriverTiersFilter().isEmpty()
+                && !cnaEvent.getDriverTiersFilter().toLowerCase().equals("na"))) {
```
inodb, line 73: I believe there might be some class with different potential N/A values that one can use here. Not entirely sure tho-  can look more later

inodb (approved):

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -32,17 +32,34 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
```
inodb, line 62: not sure if people prefer using `*` imports, CC @Luke-Sikina ?

inodb: fine by me to change like this tho, just not sure if we have a preferred way of import statements

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
inodb: can't find any util class that provides this NA checking functionality:

https://github.com/cBioPortal/cbioportal/search?p=5&q=nan

So I think this is ok. Just wondering if it can't also be `N/A` or `NaN`?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
pvannierop: I just had feedback by @sheridancbio on another PR where this subject came up. He expressed the desire to not use * for cbioportal packages, but allow it for others. I just turned off the use of wildcards in my IDE all together. Perhaps we can come up with a clear line of code conventions that we can enforce in an automated way?

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
pvannierop: Well, in FileFormats.md these fields can be '' or 'NA'. This is enforced by the validator. This is the relevant code:

line 1220
```
NULL_DRIVER_TIERS_VALUES = ('', 'NA')
```

line 1326
```
        if value not in self.NULL_DRIVER_TIERS_VALUES:
            self.logger.info('Values contained in the column cbp_driver_tiers that will appear in the "Mutation Color" '
                             'menu of the Oncoprint',
                             extra={'line_number': self.line_number, 'column_number': self.cols.index('cbp_driver_tiers'), 'cause': value})
            self.tiers.add(value)
```

core/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
inodb: yeah that would be nice, maybe soundcloud has some linting options

core/src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
inodb: Thanks!

PR #8300, opened 2021-01-25
Add new Troubleshooting section to docs, with an E2E test subsection

adamabeshouse:


alisman:
LGTM, @inodb do you think it makes sense here?

inodb (approved):
Thanks!

inodb (changes requested):
Actually - I'm not sure if this is the best place since it's really developer doc. Most of the docs.cbioportal.org is focused on people deploying cBioPortal

Wondering if it makes more sense here with the other e2e test comments:

https://github.com/cBioPortal/cbioportal-frontend#writing-e2e-tests

adamabeshouse:
@inodb good idea, I'll add it there. incidentally, that section should probably be reformatted eventually because it's a bit hard to read..

PR #8302, opened 2021-01-26
Fix incorrect passing of CNA type to count endpoint

pvannierop:
For the CNA enrichment service the CNA types were incorrectly passed to the  CNA count service. As a result the service would return counts for all CNA types (argument _Select.all()_ was passed rather than _AMP_ or _HOMDEL_).

I have added a regression test, but its is clear that this class requires more extensive unit tests to be implemented.


inodb (approved):
Thanks for finding this one before it got into the release 🥇

PR #8328, opened 2021-02-03
Allow http content security policy to be configured

alisman:
Installers should be able to allow specified domains to load their portals in iframes.   This PR allows them to do so by setting content security policy.

### DIFF
--- a/portal/src/main/webapp/index.jsp
+++ b/portal/src/main/webapp/index.jsp
@@ -27,6 +27,8 @@
 %>
 
 
+<% response.addHeader("Content-Security-Policy", GlobalProperties.getContentSecurityPolicy()); %>
+
 <!DOCTYPE html>
 <html class="cbioportal-frontend">
 <head>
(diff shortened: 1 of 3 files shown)


portal/src/main/webapp/index.jsp
```diff
@@ -27,6 +27,8 @@
 %>
 
 
+<% response.addHeader("Content-Security-Policy", GlobalProperties.getContentSecurityPolicy()); %>
```
inodb, line 30: @alisman Is this currently set to something else? Ideally one would manage this at the nginx level rather than the app level I imagine

alisman: hey, so i agree that if installers can readily configure this for whatever they are using to run the portal, then that's the best way to do it.    right now, there is no default in our codebase.     when i run the app locally, i do get a x-frame-options meta tag (this is soon to be obsolete in factor of CSP header), but maybe this is a property of Tomcat or whatever I'm using (apprunner?).

alisman: i will respond to user who inquired about this that they should try to configure it themselves

Issue #8338, opened 2021-02-05
Change default logging from file to stdout


inodb:
Currently the default logging goes to a file:

https://github.com/cBioPortal/cbioportal/blob/master/src/main/resources/log4j.properties.EXAMPLE#L13-L16

In Docker world you usually log to stdout and have some logging system that handles the stdout from various places. Should we change the default to be consoleappender (https://howtodoinjava.com/log4j/log4j-console-appender-example/)?

PR #8339, opened 2021-02-05
Make console logging default

inodb:
Previously we recommended using files for logging. That makes a lot of
sense when using a separate tomcat setup. Now with docker the tomcat is
included in the container. It's not very common with containers to log
to a file

Fix https://github.com/cBioPortal/cbioportal/issues/8338


Luke-Sikina (approved):
This is what I was thinking of - looks good to me.

kalletlak (commented):
LGTM

PR #8342, opened 2021-02-08
Delete portal/src/main/webapp/css and portal/src/main/webapp/js

Luke-Sikina:
- There is a lot of old, legacy code in the portal module
- This is an attempt to delete some of the obviously unused portions of that module:
`portal/src/main/webapp/css`, and `portal/src/main/webapp/js`


kalletlak:
@Luke-Sikina there is one page `Login.jsp` which is using deleted js and css files. It might need a cleanup.

PR #8344, opened 2021-02-09
Handle 'noauthsessionservice' in SessionServiceController

pvannierop:
This PR will make the SessionServiceController operate correctly when the `noauthsessionservice` authentication parameter has been set. The session will work under user with name _anonymous_.

Note: there are no unit tests for the SessionServiceController in place. I think we should change this in the future.

### DIFF
--- a/web/src/main/java/org/cbioportal/web/SessionServiceController.java
+++ b/web/src/main/java/org/cbioportal/web/SessionServiceController.java
@@ -1,14 +1,14 @@
 package org.cbioportal.web;
 
-import java.io.IOException;
-import java.util.*;
-
-import javax.servlet.http.HttpServletResponse;
-import javax.validation.constraints.Size;
-
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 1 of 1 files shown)


kalletlak (commented):
I think this would fetch same virtual studies, groups and custom data for all the users when `noauthsessionservice` authentication is used.

pvannierop:
@kalletlak Yes indeed and I see that you do expect strange user experience when the user comes to a portal that has `noauthsessionservice` enables and sees all kinds of customizations that she/he did not create. But the flag is names `noauthsessionservice` meaning there is 1) no authentication and 2) there is session service. In the current implementation there simply is no session service (controller returns no session information). Perhaps we can rename the current flag `noauthsessionservice` to `noauth` and create a new profile `dev-sessionservice` that runs the changes I made in sessionservice controller?

kalletlak:
With `noauthsessionservice` you could save virtual study, custom chart data, groups, comparison and main session, everything except that's related to user session/identity. That was the original design.

We might need product approval to introduce new profiles. Btw. what is the use-case to introduce `dev-sessionservice`?

Also how do we differentiate users when application is running with `dev-sessionservice` profile?

pvannierop:
I see. Well, we also may no longer need it when the e2e tests use authenticated portal. Let's put this on hold for now.

PR #8361, opened 2021-02-17
Update session service to 0.4.0

inodb:


pom.xml
```diff
@@ -412,7 +412,7 @@
       <dependency>
         <groupId>com.github.cbioportal</groupId>
         <artifactId>session-service</artifactId>
-        <version>9dd8da2dd69dc3d78938275dd35f95d96c5d2d9c</version>
+        <version>15a6a0005f822d12cfc498b082ece41849058195</version>
```
Luke-Sikina, line 415: Assuming that this is the commit hash for the 0.4.0 tag, I think this would be a bit clearer.
```suggestion
        <version>0.4.0</version>
```

PR #8363, opened 2021-02-19
Use a specific version of the session service artifact

Luke-Sikina:
- Replaced the commit hash with a version number

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -412,7 +412,7 @@
       <dependency>
         <groupId>com.github.cbioportal</groupId>
         <artifactId>session-service</artifactId>
-        <version>9dd8da2dd69dc3d78938275dd35f95d96c5d2d9c</version>
+        <version>0.3.0</version>
         <exclusions>
           <exclusion>
             <groupId>org.springframework.boot</groupId>


Luke-Sikina:
Closing. This is covered by #8361 and #8357

Luke-Sikina:
Reopening since the switch to 0.4.0 might take a minute.

pom.xml
```diff
@@ -412,7 +412,7 @@
       <dependency>
         <groupId>com.github.cbioportal</groupId>
         <artifactId>session-service</artifactId>
-        <version>9dd8da2dd69dc3d78938275dd35f95d96c5d2d9c</version>
+        <version>0.3.0</version>
```
kalletlak, line 415: I remember it being 
```
<version>master-SNAPSHOT</version>
<classifier>model</classifier>
 ```
 from https://github.com/cBioPortal/cbioportal/pull/5923/files#diff-06c528e8bf4bae607af1c54bfcfbb8f21c0d09a44cddb2851ffa08207dba3052R119
 not sure why this classifier is removed. should we revert this back?
 
 CC: @inodb

inodb: @kalletlak You're right but that actually resulted in a problem. Luke was trying to build an old version of the portal but it pulled in session service 0.4.0 instead of 0.3.0 because of that. I think in general it's better to use hardcoded versions. Even now master branch of session service doesn't work with cbioportal, see e.g.:  https://github.com/cBioPortal/cbioportal/pull/8361

inodb (approved):

PR #8366, opened 2021-02-22
update FAQ about z-scores

tmazor:


tmazor:
Updated the FAQ about z-scores to no longer say that the diploid profile is the default for mRNA. @schultzn @jjgao @ritikakundra are my edits correct?

tmazor:
Also is it still true that only that one prostate study has z-scores relative to normal?

jjgao:
Thanks, @tmazor. Many TCGA PanCancer Atlas studies will also have normal based zscores. They are actually already there in Plots tab but we have to also show them in Query page: https://github.com/cBioPortal/datahub/issues/1348. I have asked @rmadupuri to prioritize that.

tmazor:
ok thanks @jjgao  -  I removed that mention of the prostate study in anticipation of additional studies also having z-scores relative to normal.

PR #8421, opened 2021-03-04
Include files used in login.jsp

Luke-Sikina:


Luke-Sikina:
Doing styles in frontend instead. JS does nothing, so we can drop it.

PR #8423, opened 2021-03-05
Improve gene panel data query

kalletlak:
- Move all genePanelData logic to single sql query
- Avoid multiple calls to fetch genePanelData while querying for alteration counts


Luke-Sikina (approved):

inodb:
Only need to run remote e2e tests for this one to confirm it's working

inodb:
@karthik took a quick look at this one on msk portal. For the MSK-IMPACT study view the gene panel data query on production takes ~5-10s without those changes. It stays pretty much in that ballpark when I deploy those changes. However I did notice that with those changes the first load takes ~50s, which is a lot faster than the >1m i believe we were getting before without caching enabled

@kalletlak do you have any statistics on performance for this PR?

PR #8432, opened 2021-03-09
Better handle discrete data bins

### LINKED ISSUE #8375: Barchart label not always correct
See number of bone scans in https://genie-private.cbioportal.org/study/summary?id=nsclc_genie_bpc

Screen Shot 2021-02-25 at 9 17 58 AM

The data is mostly 0 and sometimes positive but never negative. Either the labels should be 0 and >0, but in this case since there are not that many ticks, maybe it makes sense to show all ticks?

Same for:

image

Prolly better to show all ticks

onursumer:
Fix #8375

# Checks
- [ ] Runs on heroku


web/src/test/java/org/cbioportal/web/util/DataBinnerTest.java
```diff
@@ -48,608 +48,10 @@
 
     @Before
     public void setup() {
-        mockData = initMockData();
+        mockData = DataBinnerMocker.mockData();
         MockitoAnnotations.initMocks(this);
     }
 
-    public Map<String, String[]> initMockData() {
```
onursumer, line 55: Cannot add anymore data into this method due to a compiler limit (https://stackoverflow.com/a/2407930). Moved mock data into `DataBinnerMocker.java` instead.

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -332,4 +332,71 @@ public boolean isAgeAttribute(String attributeId) {
...
+            }
+            // else keep the bin as is
+            else {
+                distinctBins.add(bin);
+            }
+        }
+        
+        // all bins except the outlier bins has to be distinct,
+        // otherwise return the original input bins (no conversion)
+        if (areAllDistinctExceptOutliers(distinctBins)) {
```
kalletlak, line 378: can't we just return `distinctBins`?

kalletlak (approved):

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
onursumer: good point. technically we can but we need to make sure that mixing distinct bins with internal bins doesn't mess up the bar chart layout in the frontend.

inodb (approved):
Perfect! Thank you!

Issue #8440, opened 2021-03-10
Fix localdb tests on backend repo


inodb:
Broke with: https://github.com/cBioPortal/cbioportal-frontend/pull/3650. There is a lot of code duplication in the `config.yml`. So just need to figure out how to update

PR #8441, opened 2021-03-10
Fix localdb tests

inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/8440

Copied over much from: https://github.com/cBioPortal/cbioportal-frontend/pull/3650

What it does differently:

- build the custom war file in a separate job (before even running localdb tests)
- always use a custom backend image (any time you run the test on backend you want to use that commit, this is different from the frontend scenario where you want to reuse the backend image from the branch your PR is pointing to (master/rc/etc))

What could be improved:

- It's a bit messy since it borrows so much from the frontend PR, but yet does a few things quite differently. Would prolly be good to revisit this at some point and think about how we can create jobs that just build docker images. Then have a separate job that runs tests against the images. That might make it slightly easier to re-use jobs between different repos
- Not sure if we really need to run yarn, seems kind of funky this checking out of the frontend repo, running yarn and then running the e2e tests. It has been working like that for a while tho

Other note:

- the one test that's failing for localdb was already failing in other frontend PRs as well (the `toggles_gene_panel_modal_from_patient_header` test)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,12 +119,9 @@ jobs:
                 # with variation for checking out a commit
                 command: |
                     mkdir -p $E2E_WORKSPACE && \ 
-                    cd $E2E_WORKSPACE && \
-                    git fetch --depth 1 origin $BACKEND_BRANCH && \
-                    git checkout $BACKEND_BRANCH && \
-                    cd cbioportal && \
+                    cd /tmp/repos/cbioportal && \
                     mvn -DskipTests clean install && \
-                    unzip portal/target/cbioportal*.war -d portal/target/war-exploded
+                    unzip portal/target/cbioportal*.war -d $E2E_WORKSPACE/cbioportal/portal/target/war-exploded
                 no_output_timeout: 25m
             - run:
                 name: Setup docker compose assets
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,8 +114,17 @@ jobs:
                     ./setup_environment.sh && ./setup_environment.sh >> $BASH_ENV
             - run:
                 name: Build custom backend
+                # Copied from
+                #     $TEST_HOME/docker_compose/build.sh
+                # with variation for checking out a commit
                 command: |
-                    $TEST_HOME/docker_compose/build.sh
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -90,6 +118,11 @@ jobs:
                 name: Setup python libraries
                 command: |
                     pip3 install requests pyyaml
+            - run:
+                name: Install dependencies
+                command: |
+                    sudo apt-get update && \
+                    sudo apt-get install jq
```
inodb, line 125: not sure if we need `jq` - no longer need the unzip dep cause it's part of the maven build image

.circleci/config.yml
```diff
@@ -102,54 +135,70 @@ jobs:
...
-                    cd $TEST_HOME/local/runtime-config && \
+                    source $PORTAL_SOURCE_DIR/env/custom.sh || true && \
+                    cd $TEST_HOME/runtime-config && \
                     ./setup_environment.sh && ./setup_environment.sh >> $BASH_ENV
             - run:
-                name: Build or pull backend image
+                name: Build custom backend
+                command: |
+                    mkdir -p $E2E_WORKSPACE; \
+                    mv /tmp/repos/cbioportal $E2E_WORKSPACE
```
inodb, line 145: this now just copies over the portal repo previously built in the backend build job

inodb:
Going to merge this b/c localdb is not running on backend branch at all now. This way at least it runs. Feel free to continue to review and I can follow up

inodb:
Ugh forgot to squash 🤦‍♂️ - ah well

.circleci/config.yml
pvannierop: @inodb We need `jq` for the config of the services at localdb tests.

pvannierop:
@inodb I have the feeling that you inadvertently removed some functionality in the process. Can we discuss what you did in a call and see if this is or is not the case? For instance, I notice that you always build a custom backend, even when this is not needed. I was wondering why you made this choice.

pvannierop:
@ino Ah never mind. I see this is the repo for the backend :)

PR #8456, opened 2021-03-23
Remove native ldap/ad support (now requires keycloak)

n1zea144:
We have numerous reports that the native LDAP code does properly authenticate users (see this google group [thread](https://groups.google.com/g/cbioportal/c/qoQdzS1OnZc)).  Without any resident users of native LDAP and our ever increasing adoption of Keycloak, there doesn't seem to be any reason to maintain native LDAP support.  This PR removes native LDAP/AD support and advises users to setup Keycloak/LDAP integration.


inodb (approved):
That sounds good to me

PR #8460, opened 2021-03-24
Doc update related HGNC-based gene table update

yichaoS:
### What's changed
- Line 9: Source for human gene is updated to HGNC. See line 9.
- Line 43: `importGene.pl` is updated by @mandawilson and the command line for running it importing human gene is updated. I think the command line for importing mouse genes remain the same? Separated these two scenarios. 
- Line  61: separated the section for command-line for human and mouse. However, I didn't update the command-line itself. @mandawilson Could you please let me know what is the correct command-line for generating the `reference-genome` table with the new `gene_info.txt` format?

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2015 - 2021 Memorial Sloan-Kettering Cancer Center.
  *
  * This library is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
@@ -73,97 +73,95 @@ public class ImportGeneData extends ConsoleRunnable {
                     continue;
                 }
                 
-                    String species = GlobalProperties.getSpecies();
-                    String parts[] = line.split("\t");
-                    int taxonomy = Integer.parseInt(parts[0]);
-                    if (species.equals("human")) {
-                            if (taxonomy!=9606) {
-                                // only import human genes
-                                continue;
-                            }
-                        } else if (species.equals("mouse")) {
-                            if (taxonomy!=10090) {
-                                // only import mouse genes
-                                continue;
-                            }
-                        } else {
-                                throw new Error("Species "+species+" is not supported.");
-                        }
+                String species = GlobalProperties.getSpecies();
+                String parts[] = line.split("\t");
+                int taxonomy = Integer.parseInt(parts[0]);
+                if (species.equals("human")) {
+                    if (taxonomy!=9606) {
+                        // only import human genes
+                        continue;
+                    }
+                } else if (species.equals("mouse")) {
+                    if (taxonomy!=10090) {
+                        // only import mouse genes
+                        continue;
+                    }
+                } else {
+                    throw new Error("Species "+species+" is not supported.");
+                }
                     
-                    int entrezGeneId = Integer.parseInt(parts[1]);
-                    String geneSymbol = parts[2];
-                    String locusTag = parts[3];
-                    String strAliases = parts[4];
-                    String strXrefs = parts[5];
-                    String cytoband = parts[7];
-                    String chr = cytoband.split("[pq;c \\|]")[0]; // various strange characters were found in this column
-                    int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeByBuildName(genomeBuild).getReferenceGenomeId();
-                    String desc = parts[8];
-                    String type = parts[9];
-                    String mainSymbol = parts[10]; // use 10 instead of 2 since column 2 may have duplication
-                    Set<String> aliases = new HashSet<String>();
+                int entrezGeneId = Integer.parseInt(parts[1]);
+                String geneSymbol = parts[2];
+                String locusTag = parts[3];
+                String strAliases = parts[4];
+                String strXrefs = parts[5];
+                String cytoband = parts[7];
+                String chr = cytoband.split("[pq;c \\|]")[0]; // various strange characters were found in this column
+                int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeByBuildName(genomeBuild).getReferenceGenomeId();
+                String desc = parts[8];
+                String type = parts[9];
+                String mainSymbol = parts[10]; // use 10 instead of 2 since column 2 may have duplication
+                Set<String> aliases = new HashSet<String>();
 
-                    // try to get chr from other column if needed
-                    if (chr.equals("-")) {
-                        if (!parts[6].equals("-")) {
-                            chr = parts[6];
-                        } else { 
-                            continue; // skip if both columns are absent
-                        }
+                // try to get chr from other column if needed
+                if (chr.equals("-")) {
+                    if (!parts[6].equals("-")) {
+                        chr = parts[6];
+                    } else { 
+                        continue; // skip if both columns are absent
                     }
-                    if (!locusTag.equals("-")) {
-                        aliases.add(locusTag);
-                    }
-                    if (!strAliases.equals("-")) {
-                        aliases.addAll(Arrays.asList(strAliases.split("\\|")));
-                    }
+                }
+                if (!locusTag.equals("-")) {
+                    aliases.add(locusTag);
+                }
+                if (!strAliases.equals("-")) {
+                    aliases.addAll(Arrays.asList(strAliases.split("\\|")));
+                }
                     
-                    if (geneSymbol.startsWith("MIR") && type.equalsIgnoreCase("miscRNA")) {
-                        line = buf.readLine();
-                        continue; // ignore miRNA; process separately
-                    }
+                if (geneSymbol.startsWith("MIR") && type.equalsIgnoreCase("miscRNA")) {
+                    line = buf.readLine();
+                    continue; // ignore miRNA; process separately
+                }
                     
-                    CanonicalGene gene = null;
-                    ReferenceGenomeGene refGene = null;
-                    if (!mainSymbol.equals("-")) {
-                        //Try the main symbol:
-                        gene = new CanonicalGene(entrezGeneId, mainSymbol, aliases);
-                        refGene = new ReferenceGenomeGene(entrezGeneId, referenceGenomeId, chr, cytoband, 0, 0, 0);
-                        Set<CanonicalGene> genes = genesWithSymbolFromNomenClatureAuthority.get(mainSymbol);
-                        Set<ReferenceGenomeGene> refGenes = refGenesWithSymbolFromNomenClatureAuthority.get(mainSymbol);
-                        
-                        if (genes==null) {
-                            genes = new HashSet<CanonicalGene>();
-                            genesWithSymbolFromNomenClatureAuthority.put(mainSymbol, genes);
-                        }
-                        if (refGenes==null) {
-                            refGenes = new HashSet<ReferenceGenomeGene>();
-                            refGenesWithSymbolFromNomenClatureAuthority.put(mainSymbol, refGenes);
-                        }
-                        genes.add(gene);
-                        refGenes.add(refGene);
-                    } else if (!geneSymbol.equals("-")) {
-                        //there is no main symbol, so import using the temporary/unofficial(?) symbol:
-                        gene = new CanonicalGene(entrezGeneId, geneSymbol, aliases);
-                        refGene = new ReferenceGenomeGene(entrezGeneId, referenceGenomeId, chr, 
-                            cytoband, 0, 0, 0);
-                        Set<CanonicalGene> genes = genesWithoutSymbolFromNomenClatureAuthority.get(geneSymbol);
-                        if (genes==null) {
-                            genes = new HashSet<CanonicalGene>();
-                            genesWithoutSymbolFromNomenClatureAuthority.put(geneSymbol, genes);
-                        }
-                        genes.add(gene);
-                        Set<ReferenceGenomeGene> refGenes = refGenesWithoutSymbolFromNomenClatureAuthority.get(geneSymbol);
-                        if (refGenes==null) {
-                            refGenes = new HashSet<ReferenceGenomeGene>();
-                            refGenesWithoutSymbolFromNomenClatureAuthority.put(geneSymbol, refGenes);
-                        }
-                        refGenes.add(refGene);
+                CanonicalGene gene = null;
+                ReferenceGenomeGene refGene = null;
+                if (!mainSymbol.equals("-")) {
+                    //Try the main symbol:
+                    gene = new CanonicalGene(entrezGeneId, mainSymbol, aliases);
+                    refGene = new ReferenceGenomeGene(entrezGeneId, referenceGenomeId, chr, cytoband, 0, 0, 0);
+                    Set<CanonicalGene> genes = genesWithSymbolFromNomenClatureAuthority.get(mainSymbol);
+                    Set<ReferenceGenomeGene> refGenes = refGenesWithSymbolFromNomenClatureAuthority.get(mainSymbol);
+                    if (genes==null) {
+                        genes = new HashSet<CanonicalGene>();
+                        genesWithSymbolFromNomenClatureAuthority.put(mainSymbol, genes);
                     }
+                    if (refGenes==null) {
+                        refGenes = new HashSet<ReferenceGenomeGene>();
+                        refGenesWithSymbolFromNomenClatureAuthority.put(mainSymbol, refGenes);
+                    }
+                    genes.add(gene);
+                    refGenes.add(refGene);
+                } else if (!geneSymbol.equals("-")) {
+                    //there is no main symbol, so import using the temporary/unofficial(?) symbol:
+                    gene = new CanonicalGene(entrezGeneId, geneSymbol, aliases);
+                    refGene = new ReferenceGenomeGene(entrezGeneId, referenceGenomeId, chr, cytoband, 0, 0, 0);
+                    Set<CanonicalGene> genes = genesWithoutSymbolFromNomenClatureAuthority.get(geneSymbol);
+                    Set<ReferenceGenomeGene> refGenes = refGenesWithoutSymbolFromNomenClatureAuthority.get(geneSymbol);
+                    if (genes==null) {
+                        genes = new HashSet<CanonicalGene>();
+                        genesWithoutSymbolFromNomenClatureAuthority.put(geneSymbol, genes);
+                    }
+                    if (refGenes==null) {
+                        refGenes = new HashSet<ReferenceGenomeGene>();
+                        refGenesWithoutSymbolFromNomenClatureAuthority.put(geneSymbol, refGenes);
+                    }
+                    genes.add(gene);
+                    refGenes.add(refGene);
+                }
                     
-                    if (gene!=null) {
-                        gene.setType(type);
-                    }
+                if (gene!=null) {
+                    gene.setType(type);
+                }
             }
             addGenesToDB(genesWithSymbolFromNomenClatureAuthority, genesWithoutSymbolFromNomenClatureAuthority);
             addReferenceGenomeGenesToDB(refGenesWithSymbolFromNomenClatureAuthority, 
@@ -171,20 +169,20 @@ public class ImportGeneData extends ConsoleRunnable {
         }
     }
         
-        /**
-         * Iterate over the genes found in the given maps and try to add them to the DB.
-         * 
-         * @param genesWithSymbolFromNomenClatureAuthority: genes with official symbol
-         * @param genesWithoutSymbolFromNomenClatureAuthority: genes without official symbol (can happen, some entrez genes
-         * have no official symbol yet, but a temporary symbol).
-         * 
-         * @throws DaoException
-         */
-        private static void addGenesToDB(Map<String, Set<CanonicalGene>> genesWithSymbolFromNomenClatureAuthority,
-                  Map<String, Set<CanonicalGene>> genesWithoutSymbolFromNomenClatureAuthority) throws DaoException {
+    /**
+     * Iterate over the genes found in the given maps and try to add them to the DB.
+     * 
+     * @param genesWithSymbolFromNomenClatureAuthority: genes with official symbol
+     * @param genesWithoutSymbolFromNomenClatureAuthority: genes without official symbol (can happen, some entrez genes
+     * have no official symbol yet, but a temporary symbol).
+     * 
+     * @throws DaoException
+     */
+    private static void addGenesToDB(Map<String, Set<CanonicalGene>> genesWithSymbolFromNomenClatureAuthority,
+              Map<String, Set<CanonicalGene>> genesWithoutSymbolFromNomenClatureAuthority) throws DaoException {
                      
-          DaoGeneOptimized daoGene = DaoGeneOptimized.getInstance();
-          int nrExisting = 0;
+        DaoGeneOptimized daoGene = DaoGeneOptimized.getInstance();
+        int nrExisting = 0;
         // Add genes with symbol from nomenclature authority
         for (Map.Entry<String, Set<CanonicalGene>> entry : genesWithSymbolFromNomenClatureAuthority.entrySet()) {
             Set<CanonicalGene> genes = entry.getValue();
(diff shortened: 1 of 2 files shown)


oplantalech (approved):
@yichaoS Thank you for the documentation update. Regarding mouse, there is a [link](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-8) on HGNC mentioning MGNC for mouse, which might be interesting to take a look when updating the mouse seed database.

SRodenburg (commented):
The changes themselves look fine, but from a general aspect I think some points can be discussed.
- If the procedure for mouse genes diverges from that of human, we might best move that to either a new page or to a new paragraph. Now the human and mouse parts are mixed.
- Should there be a script to automatically update the database (genes)? I don't like the fact that you have to insert MySQL statements and fiddle with tables and keys in order to update the gene tables. If in the future gene table updates are going to be more frequent, this process should be safer and smoother.

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -6,8 +6,8 @@ When loading studies into cBioPortal it is possible for warnings to occur that a
 The cBioPortal scripts package provides a method to update the `gene` and `gene_alias` tables. This requires the latest version of the NCBI Gene Info.
```
mandawilson, line 6: @yichaoS I don't think we need the "This requires the latest version of the NCBI Gene Info." sentence anymore.

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -51,6 +58,11 @@ export JAVA_HOME=<jre_installation_folder>
 3. Use the **gene** table if you query information such as hugo symbols, types of the gene 
 4. Use **reference_genome_gene** table if you query information such as chromosome, cytoband, exonic length, or the start or end of the gene
 5. Load genes only to the **reference_genome_gene** table without updating the **gene** table, please use the following command:
```
mandawilson, line 60: @yichaoS I don't really understand why someone would be doing this - but if they want to this procedure would remain the same as it was before so this is correct.

mandawilson:
@SRodenburg really what has happened is not that mouse and human have diverged.  Anyone could still use either NCBI files or HGNC files for mouse or human.  I guess we have just done the work of switching to HGNC and created the input file for human but not for mouse.  Perhaps that should be made clear.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -72,119 +72,190 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
                 if (line.startsWith("#")) {
                     continue;
                 }
-
```
mandawilson, line 75: I recommend viewing these changes with whitespace changes hidden.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -493,15 +563,15 @@ else if (i.contains("gene_name")) {
...
         ProgressMonitor.setCurrentMessage("Updated length info for " + nrGenesUpdated + " genes\n");
-        
+
         buf.close();
     }
-    
+
     /**
      * This method receives a symbol, a chromosome and a list of loci (should be from the same gene), and with that it retrieves the database gene and it calculates the length
-     * of all its exons contained in loci. If the symbol is non-ambiguous, or the chromosome reported does not match the cytoband of the database gene, then length is not updated.
```
mandawilson, line 502: I think this must have been a typo.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -527,24 +597,30 @@ public static boolean updateLength(String symbol, String chromosome, List<long[]
         ReferenceGenomeGene refGene = daoReferenceGenomeGene.getGene(gene.getEntrezGeneId(), refreneceGenomeId);
         /// If it's not in the database, don't add it
         if (refGene != null) {
-                /// Calc length
+            /// Calc length
             long[] exonic = calculateGeneLength(loci);
-                
+
             /// If there is a cytoband in database, check if cytoband-chr matches input-chr
-            String cbChr = "chr"+refGene.getCytoband();
```
mandawilson, line 534: I believe I am fixing a bug here.  See: https://github.com/cBioPortal/cbioportal/commit/b1b65c78668ebdfbb9d6921ccade66d170c04025#diff-ce623038e7b8c7a5b69d28130fe0729842482421327b51e6caf583deb8af4b30R518
```
String cbChr = "chr"+refGene.getCytoband();
System.out.println(" updateLength--> cytoband of the gene:  " + cbChr);
if (cbChr.equals(chromosome)) { //Update the length only if the chromosome matches
```
Example cytobands are "16q22.1" and "19q13|19q13-qter". An example chromosome in the gtf file (which is what chromosome comes from) is "chr19". Obviously "chr" + "19q13|19q13-qter" will never equal "chr19". You can see the code had been
```
String cbChr = "chr"+cytoband.split("p|q")[0];
```
Which makes more sense to me.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -607,96 +683,105 @@ static void importSuppGeneData(File suppGeneFile, String referenceGenomeBuild) t
...
+                System.out.println("Reading gene data from:  " + geneFile.getAbsolutePath());
+                int numLines = FileUtil.getNumLines(geneFile);
+                System.out.println(" --> total number of lines:  " + numLines);
+                ProgressMonitor.setMaxValue(numLines);
+                MySQLbulkLoader.bulkLoadOn();
+                ImportGeneData.importData(geneFile, (String)options.valueOf("genome-build"));
+                MySQLbulkLoader.flushAll(); //Gene and gene_alias should be updated before calculating gene length (gtf)!
+            }
+
+            if(options.has("hgnc")) {
```
mandawilson, line 732: Maybe this should be something more generic instead of "hgnc" because really there is nothing about the input file that is specific to hgnc.

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: Thx. Removed.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -72,119 +72,190 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
...
+     * Adds the genes parsed from the HGNC file into the database.
+     *
+     * @param geneFile HGNC File with gene information
+     * @throws IOException
+     * @throws DaoException
+     */
+    public static void importHGNCData(File geneFile, String genomeBuild) throws IOException, DaoException {
+        Map<String, Set<CanonicalGene>> canonicalGeneMap = new LinkedHashMap<>();
+        Map<String, Set<ReferenceGenomeGene>> refGeneMap = new LinkedHashMap<>();
+        try (FileReader reader = new FileReader(geneFile)) {
```
sheridancbio, line 182: if geneFile does not exist on the filesystem, we should get a FileNotFound exception here. Hopefully the stacktrace report is adequate for the user to know what went wrong

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -72,119 +72,190 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
...
+            BufferedReader buf = new BufferedReader(reader);
+            String line;
+            buf.readLine(); // skip first (header) line
```
sheridancbio, line 185: because we do not parse the line to distinguish a header / comment / blank line, the documentation should be emphatic that the input file must have a single line header, with no extra blank lines or comments

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -72,119 +72,190 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
...
+            ProgressMonitor.incrementCurValue(); // for header line
+            while ((line = buf.readLine()) != null) {
+                ProgressMonitor.incrementCurValue();
+                ConsoleUtil.showProgress();
+                String parts[] = line.split("\t", -1); // include trailing empty strings
+                if (!parts[0].matches("[0-9]+")) {
+                    ProgressMonitor.logWarning("Skipping gene with invalid entrez gene id '" + parts[1] + "'");
```
sheridancbio, line 192: This may be a mis-formatted line without a second column (no tabs). To avoid an exception perhaps the entire line should be printed out instead of parts[1]

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -72,119 +72,190 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
...
+                String cytoband = parts[3];
+                String chr = cytoband.split("[pq;c \\|]")[0]; // various strange characters were found in this column
+                String type = parts[4];
+                String strAliases = parts[5];
+                Set<String> aliases = new HashSet<String>();
+                int referenceGenomeId = DaoReferenceGenome.getReferenceGenomeByBuildName(genomeBuild).getReferenceGenomeId();
+
+                // try to get chr from chromosome column if needed
+                // if cytoband is mitochondria then use chromosome M
+                if (!chr.matches("[0-9XY]{1,2}")) {
```
sheridancbio, line 206: This accepts chromosomes such as "XY", "X3", "0" and other invalid patterns. Also chromosome ranges may differ between species. We should probably have a chromosome validator utility which worked based on the genomeBuild. [a future improvement]

sheridancbio (commented):
partial review

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -607,96 +683,105 @@ static void importSuppGeneData(File suppGeneFile, String referenceGenomeBuild) t
...
+            }
+
+            if( options.has( help ) ){
+                throw new UsageException(progName, description, parser);
+            }
+
+            ProgressMonitor.setConsoleMode(true);
+
+            if(options.has("genes")) {
+                File geneFile = new File((String) options.valueOf("genes"));
```
sheridancbio, line 722: good

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
sheridancbio: I think hgnc is ok. The general format of the file (number of columns and column content) is appropriate for hgnc processing, and the semantics (such as cytoband expectations) are based on the hgnc content. If we add new possible sources of genes, those may have their own formats and semantics. Possibly we should rename the "-genes" option to "-ncbi" because that file format has semantics specific to the ncbi gene source.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -607,96 +683,105 @@ static void importSuppGeneData(File suppGeneFile, String referenceGenomeBuild) t
...
+            }
+
+            if(options.has("hgnc")) {
+                File geneFile = new File((String) options.valueOf("hgnc"));
+                System.out.println("Reading gene data from:  " + geneFile.getAbsolutePath());
+                int numLines = FileUtil.getNumLines(geneFile);
+                System.out.println(" --> total number of lines:  " + numLines);
+                ProgressMonitor.setMaxValue(numLines);
+                MySQLbulkLoader.bulkLoadOn();
+                ImportGeneData.importHGNCData(geneFile, (String)options.valueOf("genome-build"));
```
sheridancbio, line 739: another possible improvement : validation of the genome-build string. For now we have some expectations of the format of chromosome identifiers, so we can handle "human-like" chromosome names. hg, mm, rn. But this breaks down for some other species (drosophila has "3L" and "3R" for instance, or Caenorhabditis elegans uses roman numerals).

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -3,14 +3,16 @@ This manual is intended for users that have knowledge about the structure of the
...
 
-The cBioPortal scripts package provides a method to update the `gene` and `gene_alias` tables. This requires the latest version of the NCBI Gene Info.
+The cBioPortal scripts package provides a method to update the `gene` and `gene_alias` tables.
+
+## Prepare
 
 ### Human genes
-Homo_sapien.gene_info.gz\
-ftp://ftp.ncbi.nih.gov/gene/DATA/GENE_INFO/Mammalia/Homo_sapiens.gene_info.gz
+Download [gene_info.txt](https://github.com/cBioPortal/datahub-study-curation-tools/blob/master/gene-table-update/build-input-for-importer/gene_info.txt)
```
sheridancbio, line 11: This fixed/constant file reference to the master branch of cBioPortal is a bit of a problem. The master branch may move forward and the gene_info.txt file may get updated, but how will the user notice that they are out of date and need to do a gene update? There is no versioning in the filename. I guess we could have a policy that users will download and import this file periodically whether there has been a change or not ... but that seems a little opaque. I think it would be good if we somehow captured what version of the gene table the users are currently using, and what version they are updating to.

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -3,14 +3,16 @@ This manual is intended for users that have knowledge about the structure of the
...
 
 ### Human genes
-Homo_sapien.gene_info.gz\
-ftp://ftp.ncbi.nih.gov/gene/DATA/GENE_INFO/Mammalia/Homo_sapiens.gene_info.gz
+Download [gene_info.txt](https://github.com/cBioPortal/datahub-study-curation-tools/blob/master/gene-table-update/build-input-for-importer/gene_info.txt)  
+Generated based on latest HGNC release using script [HERE](https://github.com/cBioPortal/datahub-study-curation-tools/tree/master/gene-table-update/build-input-for-importer) 
 
 ### Mouse genes
-Mus_musculus.gene_info.gz\
+Download `Mus_musculus.gene_info.gz` from
```
sheridancbio, line 15: So we are telling users that for human they use our own distribution of hgnc genes with the -hgnc argument, but for mice they use the ncbi download with the -genes argument? It would be better if we could be consistent. We could also show the "ncbi download" method for human, "hgnc" method for human, "ncbi download" method for mouse, and hgnc method for mouse ... so users know all the possibilities.

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -40,17 +42,29 @@ ALTER TABLE `geneset` AUTO_INCREMENT = 1;
...
+export PORTAL_HOME=<cbioportal_configuration_folder>
+export JAVA_HOME=<jre_installation_folder>
+./importGenes.pl --hgnc <gene_info.txt> --genome-build <GRCh37>
+```
+### Mouse genes
 ```
 export PORTAL_HOME=<cbioportal_configuration_folder>
 export JAVA_HOME=<jre_installation_folder>
-./importGenes.pl --genes <ncbi_species.gene_info> --gtf <gencode.v25.annotation.gtf> --genome-build <GRCh37>
+./importGenes.pl --genes <Mus_musculus.gene_info.gz> --gtf <gencode.v25.annotation.gtf> --genome-build <GRCh37>
```
sheridancbio, line 55: does our code really accept a gzip file? I would have thought that the user must unzip the file before running the import (and that the doc page here just omitted that step)

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: I don't understand this mode of operation either (what it does and why)

sheridancbio (changes requested):
A few small requests and suggestions.

I can talk through my comments with @mandawilson and mark work which will not be part of this PR as 'resolved'

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
sheridancbio: @yichaoS : we are not sure where the file format for the hgnc gene file is documented ... maybe you can check if it says somewhere that there *must* be a header line and that it should be only a single line

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -72,119 +72,190 @@ public static void importData(File geneFile, String genomeBuild) throws IOExcept
...
+                            chr = "M";
+                        } else {
+                            ProgressMonitor.logWarning("Skipping gene with chromosome " + chr);
+                            continue; // skip if we don't know chromosome
+                        }
+                    }
+                }
+
+                if (!strAliases.equals("-") && !strAliases.equals("")) {
+                    aliases.addAll(Arrays.asList(strAliases.split("\\|")));
```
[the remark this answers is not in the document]
sheridancbio: @yichaoS : here also, the file format documentation could be strengthened if there is no caution telling users to avoid spaces in the alias line

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
sheridancbio: I agree

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
```diff
@@ -527,24 +597,30 @@ public static boolean updateLength(String symbol, String chromosome, List<long[]
...
+                            daoReferenceGenomeGene.getGene(gene.getEntrezGeneId(), refreneceGenomeId).getExonicLength());
+                    lengthUpdated = true;
+                }
+                else {
+                    ProgressMonitor.logWarning("Cytoband does not match, gene not saved (likely another version of gene in gtf has correct chr and is saved)");
+                }
             }
             else {
+                // TODO is this the right thing to do?  we used to update the gene and mark it as length updated even though length wasn't updated
                 ProgressMonitor.logWarning("Cytoband does not match, gene not saved (likely another version of gene in gtf has correct chr and is saved)");
```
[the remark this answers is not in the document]
sheridancbio: 624 : Cytoband is not available - unable to update the length of this gene ....
619 : Cytoband from reference gene is not on the same chromosome - unable to update the length of this gene ...

Note : the logic here for computing gene lengths appears to overwrite the computed length for every matching case in the gtf file. If we are expecting the length to be set to the *longest* transcript length, then this code seems to have a bug --- we should test a case like that.

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: @yichaoS if there is versioning for the gene_info file, the documentation should explain how the versioning will be visible to users ... how users will be able to know if they are current, or if they need to do an update.

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: @yichaoS maybe we have decided that we are phasing out ncbi updates altogether?  So that would explain why we are hiding the ncbi human update feature from users.

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: @yichaoS was this a mistake?

docs/Updating-gene-and-gene_alias-tables.md
```diff
@@ -40,17 +42,29 @@ ALTER TABLE `geneset` AUTO_INCREMENT = 1;
...
 1. The **reference_genome** table needs to be populated before updating the **gene** table. Further details can be found in [this document](import-reference-genome.md). 
 2. Use **--species** option when importing genes for a species other than human
 3. Use the **gene** table if you query information such as hugo symbols, types of the gene 
 4. Use **reference_genome_gene** table if you query information such as chromosome, cytoband, exonic length, or the start or end of the gene
 5. Load genes only to the **reference_genome_gene** table without updating the **gene** table, please use the following command:
+### Human genes
+```
+./importGenes.pl --gtf <gencode.v27.annotation.gtf> --genome-build <GRCh38>
+```
+### Mouse genes
```
sheridancbio, line 67: this is wrong @yichaoS ... "GRCh38" is a human assembly, not mouse

Also ... is it correct to be using the same gencode gtf file both for human and for mouse?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
yichaoS: @sheridancbio @mandawilson Yes, the hgnc gene file (gene_info.txt) needs to have a header and it's a single line. Here is where it resides and looks like: https://raw.githubusercontent.com/cBioPortal/datahub-study-curation-tools/master/gene-table-update/build-input-for-importer/gene_info.txt

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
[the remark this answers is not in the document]
yichaoS: @sheridancbio Yes, there shouldn't be space in alias. I should add this to the documentation as well. Here is the doc for that: https://github.com/cBioPortal/datahub-study-curation-tools/tree/master/gene-table-update/build-input-for-importer

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: @sheridancbio the most up-to-date input gene file for importer will always be here: https://github.com/cBioPortal/datahub-study-curation-tools/blob/master/gene-table-update/build-input-for-importer, named as `gene_info.txt`; the file name would not include the versioning (the original output would and it would be in the archive folder, we just copy the original and rename it as `gene_info.txt`).

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: @sheridancbio Yes, we are not using NCBI updates altogether.

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: @sheridancbio I don't think so. We didn't update the part for importing mouse gene data at all, and the current doc for mouse gene seems to indicate the input is a gzip file: https://github.com/cBioPortal/cbioportal/blob/master/docs/Updating-gene-and-gene_alias-tables.md#mouse-genes

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: @sheridancbio we didn't update anything related to mouse genes....But was that a mistake we always have then? Not familiar when mouse genes. And for the gtf file not used for human genes anymore (@mandawilson ?) since we don't generate gene length in the new table? Should we just delete that parameter?

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportGeneData.java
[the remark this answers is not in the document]
sheridancbio: ok .. so maintainers can do a "diff" of this master branch to their last installed version of gene info from the archive directory and see if it changes. (or they can just see if a new entry appears in the archive directory). Thanks.

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: okay, so the user can keep track of their last installed version and either do a "diff" to the master branch, or they can look for when new entries appear in the archive folder. Thanks.

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: But we are still using NCBI for mouse. Hopefully this does not confuse users [ncbi for mouse, hgnc for human] ... and hopefully we change over to using hgnc for mouse soon.

docs/Updating-gene-and-gene_alias-tables.md
sheridancbio: @yichaoS ... yes, that is the file to download ... but unless java knows to automatically unzip a file while it is reading it [I don't think so] then the file must be manually unzipped before running the importGenes.pl command. So I think we need to change back to <Mus_musculus.gene_info> and we should probably also add a line in the instructions telling users to unzip the file. So in the "Prepare" section above, after the lines:
```
Download `Mus_musculus.gene_info.gz` from   ftp://ftp.ncbi.nih.gov/gene/DATA/GENE_INFO/Mammalia/Mus_musculus.gene_info.gz
```
I would add this:
```
Unzip the downloaded file with the command `gunzip Mus_musculus.gene_info.gz`
```

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: @sheridancbio agreed. I created this issue here: https://github.com/cBioPortal/cbioportal/issues/8678 and tagged you me and @mandawilson for future reference.

docs/Updating-gene-and-gene_alias-tables.md
yichaoS: @sheridancbio I added that line in this PR as you suggested. I do think tho the whole process of mouse genes import needs to be streamlined at some point. Add this .zip thing in the [issue](https://github.com/cBioPortal/cbioportal/issues/8678) I created as well.

sheridancbio: I'm going to merge this as-is, but I very much doubt that the command will work with the file name Mus_musculus.gene_info.gz. I'll add a comment into the docs separately with my concerns.

sheridancbio (approved):
Ok. Approving the PR without what I believe to be a needed correction to the `Mus_musculus.gene_info.gz` filename. An issue has been created to confirm and resolve this problem and attempt to standardize mouse loading.

PR #8553, opened 2021-04-27
Improve gene-panel api performance

kalletlak:
Solution: if there are more than 30,000 objects(molecular case identifiers) then fetch gene panel data for all the samples/patients for queried molecular profiles and then filter the result

Also removed all the logic for finding whether the sample is profile or not based on sequenced case list.

This improved the performance by alteast 2x(11sec -> 4sec) when tested locally for `/results?cancer_study_list=msk_impact_2017&tab_index=tab_visualize&case_set_id=msk_impact_2017_all&Action=Submit&gene_list=TMPRSS2%250ATP53%250AKRAS%250ACDKN2A%250ACDKN2B`

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
@@ -1,16 +1,17 @@
 package org.cbioportal.service.impl;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
+import java.util.stream.Collectors;
+
 import static java.util.stream.Collectors.*;
 
 import org.apache.commons.math3.util.Pair;
 import org.cbioportal.model.GenePanel;
 import org.cbioportal.model.GenePanelData;
 import org.cbioportal.model.GenePanelToGene;
+import org.cbioportal.model.MolecularProfileCaseIdentifier;
 import org.cbioportal.model.meta.BaseMeta;
 import org.cbioportal.persistence.GenePanelRepository;
 import org.cbioportal.service.GenePanelService;
@@ -23,11 +24,13 @@ import org.springframework.stereotype.Service;
 public class GenePanelServiceImpl implements GenePanelService {
     @Autowired
     private GenePanelRepository genePanelRepository;
+    
+    private final Integer maxCasesCountToIncludeInQuery = 30000;
 
     @Override
     public List<GenePanel> getAllGenePanels(String projection, Integer pageSize, Integer pageNumber, String sortBy, 
                                             String direction) {
-        
+
         List<GenePanel> genePanels = genePanelRepository.getAllGenePanels(projection, pageSize, pageNumber, sortBy, 
             direction);
 
@@ -39,13 +42,13 @@ public class GenePanelServiceImpl implements GenePanelService {
             genePanels.forEach(g -> g.setGenes(genePanelToGeneList.stream().filter(p -> p.getGenePanelId()
                 .equals(g.getStableId())).collect(toList())));
         }
-        
+
         return genePanels;
     }
 
     @Override
     public BaseMeta getMetaGenePanels() {
-        
+
         return genePanelRepository.getMetaGenePanels();
     }
 
@@ -56,16 +59,16 @@ public class GenePanelServiceImpl implements GenePanelService {
         if (genePanel == null) {
             throw new GenePanelNotFoundException(genePanelId);
         }
-        
-        genePanel.setGenes(genePanelRepository.getGenesOfPanels(Arrays.asList(genePanelId)));
+
+        genePanel.setGenes(genePanelRepository.getGenesOfPanels(Collections.singletonList(genePanelId)));
         return genePanel;
     }
 
     @Override
-	public List<GenePanel> fetchGenePanels(List<String> genePanelIds, String projection) {
+    public List<GenePanel> fetchGenePanels(List<String> genePanelIds, String projection) {
 
         List<GenePanel> genePanels = genePanelRepository.fetchGenePanels(genePanelIds, projection);
-        
+
         if (projection.equals("DETAILED")) {
 
             List<GenePanelToGene> genePanelToGeneList = genePanelRepository.getGenesOfPanels(genePanels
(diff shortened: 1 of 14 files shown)


service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
```diff
@@ -76,118 +80,177 @@ public GenePanel getGenePanel(String genePanelId) throws GenePanelNotFoundExcept
...
             .collect(toMap(
-                    datum -> new Pair(datum.getMolecularProfileId(),
+                datum -> new Pair(datum.getMolecularProfileId(),
                     datum.getSampleId()), Function.identity()));
 
-        return genePanelData.stream().map(datum -> {
-            String molecularProfileId = datum.getMolecularProfileId();
+        for (int i = 0; i < molecularProfileIds.size(); i++) {
+            String molecularProfileId = molecularProfileIds.get(i);
+            String sampleId = sampleIds.get(i);
```
n1zea144, line 153: Similar question here - is sampleIds the same size as molecularProfileIds?

n1zea144 (commented):
Is it true that this code will only work until fusions are moved to structural variants?

Is it possible to combine the setting of variables distinctMolecularProfileIds and  hasFusionProfileIdsInQuery in one pass over molecularProfileIds (and maybe updatedMolecularProfileIds and updatedSampleIds in the same pass)?

Also, have you considered implementing fetchGenePanelDataInMultipleMolecularProfilesByPatientIds by converting patient ids to samples ids then calling fetchGenePanelDataInMultipleMolecularProfiles.  Maybe its cost prohibitive, but seems like very similar code.

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
```diff
@@ -76,118 +80,177 @@ public GenePanel getGenePanel(String genePanelId) throws GenePanelNotFoundExcept
...
-        for (int i = 0; i < molecularProfileIds.size(); i++) {
-            String molecularProfileId = molecularProfileIds.get(i);
-            if(molecularProfileId.endsWith("_fusion")) {
-                updatedMolecularProfileIds.add(molecularProfileId.replace("_fusion", "_mutations"));
-                updatedSampleIds.add(sampleIds.get(i));
+            for (int i = 0; i < molecularProfileIds.size(); i++) {
+                String molecularProfileId = molecularProfileIds.get(i);
+                if(molecularProfileId.endsWith("_fusion")) {
+                    updatedMolecularProfileIds.add(molecularProfileId.replace("_fusion", "_mutations"));
+                    updatedSampleIds.add(sampleIds.get(i));
```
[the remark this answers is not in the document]
kalletlak: They are paired. they are separated from MolecularCaseIdentifiers. May be I should refactor complete gene-panel api

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
kalletlak: yes

kalletlak:
> Is it true that this code will only work until fusions are moved to structural variants?

There are two parts in this
1. As you said, checking if there are any fusions.
2. Checking for number of queried object(if there are more than 30,000 then query for just based on molecular profiles and then do the filtering)
> 
> Is it possible to combine the setting of variables distinctMolecularProfileIds and hasFusionProfileIdsInQuery in one pass over molecularProfileIds (and maybe updatedMolecularProfileIds and updatedSampleIds in the same pass)?

Yes, we could do it
> Also, have you considered implementing fetchGenePanelDataInMultipleMolecularProfilesByPatientIds by converting patient ids to samples ids then calling fetchGenePanelDataInMultipleMolecularProfiles. Maybe its cost prohibitive, but seems like very similar code.

True, but in order to do that I need to make two queries to sql. One for fetching study id based on molecular profile and then samples from patient ids and study ids. I can give a try

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
```diff
@@ -23,11 +25,13 @@
 public class GenePanelServiceImpl implements GenePanelService {
     @Autowired
     private GenePanelRepository genePanelRepository;
+    
+    private final Integer maxCasesCountToIncludeInQuery = 30000;
```
inodb, line 29: @kalletlak this number works now but it might be nice to get some idea of how we should update it based on max cohort size. I imagine if you have 1M samples in a cohort it might be better to increase the threshold? So it's somewhat study size dependent, correct?

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
```diff
@@ -76,118 +80,215 @@ public GenePanel getGenePanel(String genePanelId) throws GenePanelNotFoundExcept
...
+        }
 
-    @Override
-    public List<GenePanelData> fetchGenePanelDataInMultipleMolecularProfilesByPatientIds(
-            List<String> molecularProfileIds, List<String> patientIds) {
+        Map<Pair<String, String>, Sample> sampleSetWithSequencedInfo;
+
+        if (updatedSampleIds.size() < QUERY_LIMIT) {
+            sampleSetWithSequencedInfo = sampleService
+                .fetchSamples(filteredStudyIds, filteredSampleIds, "SEQUENCED")
```
sheridancbio, line 149: using "SEQUENCED" in the projection field seems contrary to expectations. The four members of the Projection enum are [ ID, SUMMARY, DETAILED, META ], so adding meaning to this argument in the service layer in order to process "SEQUENCED" differently worries me. I think it should be discussed at least.

Maybe adding a named argument to the service layer function will help make the meaning more clear .. such as 
public List<Sample> fetchSamples(
    List<String> studyIds,
    List<String> sampleIds,
    Boolean computeSampleSequencedAttributeByProfile,
    String projection)
This way it is kept clear which mode (Projection) the original api request from the user had, and the behavior sub-service layer interactions are requesting.
Also ... I think that all api endpoints which accept the Projection argument should behave correctly for all 4 possible projections supplied by the user. So I would test that this code functions correctly if the original API call is made with projection ID, META, SUMMARY, DETAILED {they should all show consistent sample counts}

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -222,6 +222,21 @@ private void processSamples(List<Sample> samples, String projection) {
...
+            List<MolecularProfile> molecularProfiles = molecularProfileRepository.getMolecularProfilesInStudies(distinctStudyIds, projection);
+
+            for (String studyId : distinctStudyIds) {
+                sequencedSampleIdsMap.put(studyId,
+                    new HashSet<String>(sampleListRepository.getAllSampleIdsInSampleList(studyId + SEQUENCED)));
+            }
+
+            samples.forEach(sample -> {
+                sample.setSequenced(sequencedSampleIdsMap.get(sample.getCancerStudyIdentifier())
+                    .contains(sample.getStableId()));
```
sheridancbio, line 238: Because sample stable ids are not globally unique, wouldn't we run into a problem in a case like this:
cancer_study `studyX` has molecular_profile `X_mutations` has patient `p24` has sample `s6`
cancer_study `studyX` has molecular_profile `X_mutations` has patient `p24` has sample `s8`
cancer_study `studyY` has molecular_profile `Y_mutations` has patient `p57` has sample `s3`
cancer_study `studyY` has molecular_profile `Y_mutations` has patient `p57` has sample `s8`
If the query list includes [
{"molecular profile" : "X_mutations", "sample_stable_id" : "s8"} , 
{"molecular profile" : "Y_mutations", "sample_stable_id" : "s3"}
]
Is it possible we are going to be returning/counting 3 samples ... X_mutations s8, Y_mutations s3, Y_mutations s8 ?

sheridancbio (commented):
A couple of comments/concerns.

I'm leaving a comment rather than approval/change request because my understanding is not complete. In particular I didn't have time to digest the database operations sufficiently to know the logical validity of the changes. However, I did not see any obvious logical problems in the SQL.

kalletlak:
@sheridancbio thanks for reviewing. After discussing with @jjgao we decided to remove logic for finding whether the sample is profile or not based on case list. So all the logic related to `SEQUENCED` projection is removed.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
kalletlak: c

service/src/main/java/org/cbioportal/service/impl/GenePanelServiceImpl.java
kalletlak: Not required anymore

kalletlak:
These changes are included in #8574. Closing this.

PR #8562, opened 2021-05-03
Check inequality of stable_id & source_stable_id

oplantalech:
Add validation check so that `stable_id` and `source_stable_id` are not referencing the same profile in the meta Z-score expression file.


inodb (approved):
LGTM

PR #8564, opened 2021-05-03
Avoid redundant filter and clinical data fetch calls

### LINKED ISSUE #8584: Improve clinical data bin counts endpoint
- [x] Avoid hitting DB for filtered API calls
- [x] Refactor and split the existing logic into multiple methods
- [x] Move the refactored methods out of the controller layer to the service/util layer
- [x] Add tests for the new methods

onursumer:
Fix #8584 

- Avoid redundant database queries by implementing in-memory clinical data filtering
- Avoid applying the same filter twice
- Move out clinical data bin count endpoint logic to the new `ClinicalDataBinUtil` class.
- Add unit tests for the new util class

### DIFF
--- a/web/src/main/java/org/cbioportal/web/StudyViewController.java
+++ b/web/src/main/java/org/cbioportal/web/StudyViewController.java
@@ -15,19 +15,15 @@ import org.cbioportal.service.exception.StudyNotFoundException;
 import org.cbioportal.service.util.ClinicalAttributeUtil;
 import org.cbioportal.web.config.annotation.InternalApi;
 import org.cbioportal.web.parameter.ClinicalDataBinCountFilter;
-import org.cbioportal.web.parameter.ClinicalDataBinFilter;
 import org.cbioportal.web.parameter.ClinicalDataCountFilter;
 import org.cbioportal.web.parameter.ClinicalDataFilter;
-import org.cbioportal.web.parameter.ClinicalDataType;
 import org.cbioportal.web.parameter.DataBinMethod;
 import org.cbioportal.web.parameter.GenericAssayDataBinCountFilter;
 import org.cbioportal.web.parameter.GenomicDataBinCountFilter;
 import org.cbioportal.web.parameter.Projection;
 import org.cbioportal.web.parameter.SampleIdentifier;
 import org.cbioportal.web.parameter.StudyViewFilter;
-import org.cbioportal.web.util.DataBinner;
-import org.cbioportal.web.util.StudyViewFilterApplier;
-import org.cbioportal.web.util.StudyViewFilterUtil;
+import org.cbioportal.web.util.*;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.context.ApplicationContext;
@@ -69,6 +65,8 @@ public class StudyViewController {
     @Autowired
     private ClinicalDataService clinicalDataService;
     @Autowired
+    private ClinicalDataFetcher clinicalDataFetcher;
+    @Autowired
     private MolecularProfileService molecularProfileService;
     @Autowired
     private AlterationCountService alterationCountService;
@@ -81,8 +79,6 @@ public class StudyViewController {
     @Autowired
     private SignificantCopyNumberRegionService significantCopyNumberRegionService;
     @Autowired
-    private DataBinner dataBinner;
-    @Autowired
     private StudyViewFilterUtil studyViewFilterUtil;
     @Autowired
     private ClinicalAttributeService clinicalAttributeService;
@@ -92,6 +88,8 @@ public class StudyViewController {
     private SampleListService sampleListService;
     @Autowired
     private StudyViewService studyViewService;
+    @Autowired
+    private ClinicalDataBinUtil clinicalDataBinUtil;
 
     private static final List<CNA> CNA_TYPES_AMP_AND_HOMDEL = Collections.unmodifiableList(Arrays.asList(CNA.AMP, CNA.HOMDEL));
 
@@ -141,17 +139,10 @@ public class StudyViewController {
         @ApiIgnore // prevent reference to this attribute in the swagger-ui interface. this attribute is needed for the @PreAuthorize tag above.
         @Valid @RequestAttribute(required = false, value = "interceptedClinicalDataBinCountFilter") ClinicalDataBinCountFilter interceptedClinicalDataBinCountFilter
     ) {
-
-        List<ClinicalDataBinFilter> attributes = interceptedClinicalDataBinCountFilter.getAttributes();
-        StudyViewFilter studyViewFilter = interceptedClinicalDataBinCountFilter.getStudyViewFilter();
-
-        if (attributes.size() == 1) {
-            studyViewFilterUtil.removeSelfFromFilter(attributes.get(0).getAttributeId(), studyViewFilter);
-        }
-
+        StudyViewFilter studyViewFilter = clinicalDataBinUtil.removeSelfFromFilter(interceptedClinicalDataBinCountFilter);
         boolean singleStudyUnfiltered = studyViewFilterUtil.isSingleStudyUnfiltered(studyViewFilter);
         List<ClinicalDataBin> clinicalDataBins = 
-            instance.cachableFetchClinicalDataBinCounts(dataBinMethod, attributes, studyViewFilter, singleStudyUnfiltered);
+            instance.cachableFetchClinicalDataBinCounts(dataBinMethod, interceptedClinicalDataBinCountFilter, singleStudyUnfiltered);
 
         return new ResponseEntity<>(clinicalDataBins, HttpStatus.OK);
     }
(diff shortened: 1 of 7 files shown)


web/src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
```diff
@@ -153,4 +156,117 @@ public Integer getFilteredCountByDataEquality(List<ClinicalDataFilter> attribute
...
+        List<String> conflictingPatientAttributes
+    ) {
+        List<ClinicalData> combinedResult = new ArrayList<>();
+        
+        // create look up for faster filtering
+        Map<String, String> sampleIdToStudyId = mapCaseToStudy(sampleIds, studyIds);
+        Map<String, String> patientIdToStudyId = mapCaseToStudy(patientIds, studyIdsOfPatients);
+        Map<String, Boolean> sampleAttributeIdLookup = listToMap(sampleAttributeIds);
+        Map<String, Boolean> patientAttributeIdLookup = listToMap(patientAttributeIds);
+        Map<String, Boolean> conflictingPatientAttributeIdLookup = listToMap(conflictingPatientAttributes);
```
onursumer, line 196: Ideally we should not create indexes for every single API call. If we find a way to cache indexes like we cache the DB query responses, we can further improve the performance of filtering.

onursumer: Profiling shows that the time required to create indexes is insignificant compared to the time required for the rest of the operations. Let's keep it this way for now.

web/src/test/java/org/cbioportal/web/util/ClinicalDataBinUtilTest.java
```diff
@@ -0,0 +1,791 @@
...
+
+        studyIds.add(STUDY_ID);
+        studyIds.add(STUDY_ID);
+        studyIds.add(STUDY_ID);
+        studyIds.add(STUDY_ID);
+        studyIds.add(STUDY_ID);
+        studyIds.add(STUDY_ID);
+        studyIds.add(STUDY_ID);
+
+        return studyIds;
```
dippindots, line 489: Since we have a lot this kind of code, maybe we should create a function `populateStudyIds`?

dippindots (approved):
@onursumer This pr looks good to me, I just have one minor suggestion, it's ok to leave it as is.

web/src/test/java/org/cbioportal/web/util/ClinicalDataBinUtilTest.java
onursumer: Good point. Seems like there is already a utility method called `nCopies`. So something like this should work:

```Java
return Collections.nCopies(7, STUDY_ID);
```

Luke-Sikina (approved):

web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
```diff
@@ -0,0 +1,351 @@
...
+
+        Map<String, ClinicalDataType> attributeDatatypeMap = constructAttributeDataMap(
+            unfilteredSampleAttributeIds,
+            unfilteredPatientAttributeIds,
+            unfilteredConflictingPatientAttributeIds
+        );
+
+        List<ClinicalData> unfilteredClinicalDataForSamples = clinicalDataFetcher.fetchClinicalDataForSamples(
+            unfilteredStudyIds,
+            unfilteredSampleIds,
```
kalletlak, line 87: Since we are fetching data for all sample, can't we just use study ids

web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
```diff
@@ -0,0 +1,351 @@
...
+            new ArrayList<>(unfilteredSampleAttributeIds)
+        );
+
+        List<ClinicalData> unfilteredClinicalDataForPatients = clinicalDataFetcher.fetchClinicalDataForPatients(
+            studyIdsOfUnfilteredPatients,
+            unfilteredPatientIds,
```
kalletlak, line 93: same here

web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
```diff
@@ -0,0 +1,351 @@
...
+        List<SampleIdentifier> filteredSampleIdentifiers =
+            studyViewFilterUtil.shouldSkipFilterForClinicalDataBins(studyViewFilter) ?
+                unfilteredSampleIdentifiers : studyViewFilterApplier.apply(studyViewFilter);
+
+        List<String> filteredUniqueSampleKeys;
+        List<String> filteredUniquePatientKeys;
+        List<ClinicalData> filteredClinicalData;
+
+        // if filtered and unfiltered samples are exactly the same, no need to fetch clinical data again
+        if (filteredSampleIdentifiers.equals(unfilteredSampleIdentifiers)) {
```
kalletlak, line 119: may be replace this by `studyViewFilterUtil.shouldSkipFilterForClinicalDataBins(studyViewFilter)` and calculate `filteredSampleIdentifiers` inside else block

web/src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
```diff
@@ -153,4 +154,117 @@ public Integer getFilteredCountByDataEquality(List<ClinicalDataFilter> attribute
...
+    }
+    
+    private <T> Map<T, Boolean> listToMap(List<T> list) {
+        return list.stream().collect(Collectors.toMap(s -> s, s -> true, (s1, s2) -> s1));
+    }
+
+    private  Map<String, String> mapCaseToStudy(List<String> caseIds, List<String> studyIds) {
+        Map<String, String> caseToStudy = new HashMap<>();
+        
+        for (int i =0; i < caseIds.size(); i++) {
```
kalletlak, line 264: there is a possibility for duplicate caseId. combination of case id and study id should be the key

web/src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
kalletlak: Looks like these Maps are required depending on attributes. may be they can be moved inside `if` block

web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
onursumer: Not sure if it is okay to ignore sample ids because, even if we call it unfiltered, actually we still filter by `sampleIdentifiers` here: https://github.com/onursumer/cbioportal/blob/f5e85a8c33181a8d13f166e36cab05118299945b/web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java#L49

I thought we were doing this for a reason so didn't want to change the existing logic.

web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
onursumer: In theory, `filteredSampleIdentifiers` may still be equal to `unfilteredSampleIdentifiers` even after applying the filter (in case provided filter does not filter anything out after we apply it). So ideally we should always check for equality for both lists. I am not sure if this ever happens in practice though.

web/src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
```diff
@@ -0,0 +1,351 @@
...
+                    .collect(Collectors.toList());
+
+                clinicalDataBins.addAll(dataBins);
+            }
+        }
+
+        return clinicalDataBins;
+    }
+
+    public List<ClinicalDataBin> calculateDynamicDataBins(
```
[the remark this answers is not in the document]
onursumer: Actually we never use dynamic data bins in the frontend. We always use the initial bins created for the unfiltered data. The endpoint still support dynamic binning though. Maybe we add tests for this method in a follow up PR?

web/src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
onursumer: Good catch! A single patient/sample can be found in more than one study, right?

kalletlak: yes

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -196,147 +187,15 @@ private void init() {
     )
     public List<ClinicalDataBin> cachableFetchClinicalDataBinCounts(
```
Luke-Sikina, line 188: You need an annotation here to enable caching. 
```suggestion
    @Cacheable(
        cacheResolver = "staticRepositoryCacheOneResolver",
        condition = "@cacheEnabledConfig.getEnabled() && #singleStudyUnfiltered"
    )
    public List<ClinicalDataBin> cachableFetchClinicalDataBinCounts(
```

PR #8595, opened 2021-05-12
fix 8594 cache samples when given sampleListIds in the static cache

### LINKED ISSUE #8594: Apply caching to samples endpoint when we receive sampleListId
We want to cache samples when we are passed sampleListIds.  There is a finite set of these and thus we can put them into static cache.

Use the static cache instead of the general cache.  See the resolver is different than in the gene-panel example:
```
  @Cacheable(
        condition = "@cacheEnabledConfig.getEnabled()"
    )
```

jagnathan:
fix #8594 cache samples when given sampleListIds in the static cache

- a New function in SampleController for which results can be cached in static cache
- b Singleton Instance to store reference to the SampleController object

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -258,6 +270,20 @@ private boolean usingAuth() {
...
+    @Cacheable(
+        cacheResolver = "staticRepositoryCacheOneResolver",
+        condition = "@cacheEnabledConfig.getEnabled()"
+    )
+    public List<Sample> fetchSamplesInner(
+        List<String> sampleListIds, String projection
+    ) {
+        List<Sample> samples;
+        samples = sampleService.fetchSamples(sampleListIds, projection);
+        return samples;
```
Luke-Sikina, line 282:
```suggestion
        return sampleService.fetchSamples(sampleListIds, projection);
```

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -244,7 +256,7 @@ private boolean usingAuth() {
             List<Sample> samples;
 
             if (interceptedSampleFilter.getSampleListIds() != null) {
-                samples = sampleService.fetchSamples(interceptedSampleFilter.getSampleListIds(), projection.name());
+                samples = instance.fetchSamplesInner(interceptedSampleFilter.getSampleListIds(), projection.name());
```
Luke-Sikina, line 259: This is fine, but why not cache this for meta projections as well? (Line 244)

Luke-Sikina (approved):

web/src/main/java/org/cbioportal/web/SampleController.java
Luke-Sikina: Ok, never mind, meta is super fast.

jagnathan: meta projections is fast, so improvement by caching this could be minimal, as of now.

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -258,6 +270,18 @@ private boolean usingAuth() {
...
     }
 
+    @Cacheable(
+        cacheResolver = "staticRepositoryCacheOneResolver",
+        condition = "@cacheEnabledConfig.getEnabled()"
+    )
+    public List<Sample> fetchSamplesInner(
+        List<String> sampleListIds, String projection
+    ) {
+        return sampleService.fetchSamples(sampleListIds, projection);
```
alisman, line 280: Couple of things.  One, I think we should fetch for samplelistids one by one so that we cache at the list level.  That maskes it more reusable for different combinations of studies.  This will happen a lot.  @jagnathan this would mean fetchSamplesInner would accept on one listId.   @kalletlak points out that caching at the service level here is very redundant because it's just a pass through to repository query, which is itself cached.  This is a good point, however, we want this data to be cached with STATIC cache because the list of possible items is finite.  As we know from experience, when things like this are allowed to be evicted, users really suffer.  Do we have the static option at the repository level of caching?  @Luke-Sikina @kalletlak?

alisman (changes requested):
see discussion started in comment

web/src/main/java/org/cbioportal/web/SampleController.java
jagnathan: I can cache it with a single SampleListId and then add a loop to combine the results for a list of SampleListId.

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -258,6 +273,18 @@ private boolean usingAuth() {
         }
     }
 
+    @Cacheable(
```
alisman, line 276: remove cacheable b/c it's already at repositoyr level

alisman: change the resolver in the repository cache decorstor to "staticRepositoryCacheOneResolver"

Issue #8571, opened 2021-05-05
incorrect number of driver SV/Fusion


jjgao:
https://bit.ly/3hbuHkS

![image](https://user-images.githubusercontent.com/840895/117187350-d20bd080-ada9-11eb-87bb-76d0322f1b9f.png)

inodb:
Also oncoprint annotations seem to be missing oncokb:

![Screen Shot 2021-05-13 at 10.43.25 AM.png](https://images.zenhubusercontent.com/58337ef1b9dc3c622022519c/a859c5f6-449e-4f16-9cfc-fd2eadcbc66a)

inodb:
Te issue is that the annotation only works if the first gene is ALK

jjgao:
@leexgh all of them should be drivers as annotated in the mutations tab, but I guess they were not correctly annotated in OncoPrint... @zhx828 could you help look into this?

inodb:
Solved 🙌 - thanks all!

![Screen Shot 2021-05-21 at 4.58.00 PM.png](https://images.zenhubusercontent.com/58337ef1b9dc3c622022519c/6e4c7957-0fd9-4402-aeb6-58345b71396b)

PR #8602, opened 2021-05-14
Pick the proper fusion gene partner

zhx828:
This is related to https://github.com/cBioPortal/cbioportal/issues/8571

### DIFF
--- a/service/src/main/java/org/cbioportal/service/util/MutationMapperUtils.java
+++ b/service/src/main/java/org/cbioportal/service/util/MutationMapperUtils.java
@@ -113,10 +113,20 @@ public class MutationMapperUtils {
                                     structuralVariant.setSite2EntrezGeneId(fusion.getEntrezGeneId());
                                     structuralVariant.setSite2HugoSymbol(fusion.getGene().getHugoGeneSymbol());
                                 } else {
-                                    Gene site2Gene = getGene(site2GeneSymbol);
-                                    if (site2Gene != null) {
-                                        structuralVariant.setSite2EntrezGeneId(site2Gene.getEntrezGeneId());
-                                        structuralVariant.setSite2HugoSymbol(site2Gene.getHugoGeneSymbol());
+                                    Gene site1Gene = getGene(site1GeneSymbol);
+                                    // we need to pick the right gene since the fusion gene has been assigned to structural variant site1
+                                    Gene pickedGene = null;
+                                    if (site1Gene != null && !site1Gene.getEntrezGeneId().equals(fusion.getEntrezGeneId())) {
+                                        pickedGene = site1Gene;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/cbioportal/service/util/MutationMapperUtils.java
```diff
@@ -113,10 +113,19 @@
...
                                     Gene site2Gene = getGene(site2GeneSymbol);
-                                    if (site2Gene != null) {
-                                        structuralVariant.setSite2EntrezGeneId(site2Gene.getEntrezGeneId());
-                                        structuralVariant.setSite2HugoSymbol(site2Gene.getHugoGeneSymbol());
+                                    // we need to pick the right gene since the fusion gene has been assigned to structural variant site1
+                                    Gene pickedGene = null;
+                                    if (site1Gene != null && !site1Gene.getEntrezGeneId().equals(fusion.getEntrezGeneId())) {
+                                        pickedGene = site1Gene;
+                                    }
+                                    if (site2Gene != null && !site2Gene.getEntrezGeneId().equals(fusion.getEntrezGeneId())) {
```
kalletlak, line 123: shouldn't this be in else block or `pickedGene == null` needs to be included in if caluse

zhx828: @kalletlak updated

service/src/main/java/org/cbioportal/service/util/MutationMapperUtils.java
```diff
@@ -113,10 +113,18 @@
                                     structuralVariant.setSite2EntrezGeneId(fusion.getEntrezGeneId());
                                     structuralVariant.setSite2HugoSymbol(fusion.getGene().getHugoGeneSymbol());
                                 } else {
+                                    Gene site1Gene = getGene(site1GeneSymbol);
                                     Gene site2Gene = getGene(site2GeneSymbol);
-                                    if (site2Gene != null) {
-                                        structuralVariant.setSite2EntrezGeneId(site2Gene.getEntrezGeneId());
-                                        structuralVariant.setSite2HugoSymbol(site2Gene.getHugoGeneSymbol());
+                                    // we need to pick the right gene since the fusion gene has been assigned to structural variant site1
```
kalletlak, line 118: `site2Gene` is used if the first condition fails. may be move this inside `else` block

kalletlak (approved):
LGTM

service/src/main/java/org/cbioportal/service/util/MutationMapperUtils.java
zhx828: @kalletlak updated

PR #8665, opened 2021-06-04
Impl. ConditionalOnProperty classs for conditional bean creation

pvannierop:
This PR will implement a new module named _utils_. This module has a new class _ConditionalOnProperty_ that is used in other modules to conditionally instantiate beans. Were applicable it replaces class _ConditionalOnDatMethod_ with similar function but with much narrower scope. This class replaces the mechanism in which conditional beans were created from spring profiles (e.g. as for _persistence.cache_type_). Therefore, passing of jvm parameter `persistence.cache_type` is no longer needed.

The _ConditionalOnProperty_ class is modelled after the class with identical name from Spring Boot ().

# Usage

Only instantiate when property `my_prop` has value `correct_value`:

```
@ConditionalOnProperty(name = "my_prop", havingValue = "correct_value")
```

Only instantiate when property `my_prop` does not have value `correct_value`:

```
@ConditionalOnProperty(name = "my_prop", havingValue = "correct_value", isNot = true)
```

Only instantiate when property `my_prop` has value `correct_value_1` or  `correct_value_2` :

```
@ConditionalOnProperty(name = "my_prop", havingValue = { "correct_value_1", "correct_value_2"} )
```

Only instantiate when property `my_prop` does not have value `correct_value_1` and not value  `correct_value_2` :

```
@ConditionalOnProperty(name = "my_prop", havingValue = { "correct_value_1", "correct_value_2"}, isNot = true)
```

# Tests

Unit tests are included.


inodb (approved):
This looks great to me but I don't know much about bean wiring :)

kalletlak (approved):

Luke-Sikina (approved):
I'm going to defer to Karthik for the code review, but I did run this locally and I can confirm that you can now set the cache type after the jar has been built. Nice job!

Luke-Sikina:
@inodb What else has to be done before we can merge this?

PR #8667, opened 2021-06-07
Remove the need to restart Tomcat after data loading via cache-clearing endpoint

pvannierop:
# Problem
When loading a new study the cBioPortal tomcat server needs to be restarted. This breaks the user experience of users logged in the running portal instance. The need to restart comes from the need to flush two levels of caching used by cBioPortal. The first is by the _CacheManager_ used by Spring (can be MyBatis/EHcache or redis). The second is by the _CacheMapUtils_ class that references lists of studies, samples and profiles for fast lookup during evaluation of user permissions.

# Solution
This PR adds a new endpoint `/api/cache` that triggers flushing of the Spring-managed and CacheMapUtils caches when a DELETE http-request is received. The CacheMapUtils cache is re-initialized for handling of user permissions. Calling this endpoint removes the need to restart the cBioPortal tomcat server/docker service after loading of a new study (or other data).

## Rationale for choosing endpoint over script 'local' script
I opted for an endpoint over a mechanism that involved calling a local script for flushing the caches because in certain deployment environments (e.g., kubernetes) triggering scripts in a running instance is not/less convenient. A REST endpoint can be easily called on all running instances in a system irrespective of the deployment method. I consider the ability to turn off the endpoint in _portal.properties_  and the use of a secret API key to sufficiently cover the added security risk. 

## Access to the endpoint
The endpoint is not handled by the user authN/authZ method. Flushing of the caches can only be performed when passing an API key configured in the _portal.properties_ in the `X-API-KEY` header. When a API key is not provided the endpoint returns a BAD REQUEST response. When the wrong API key is passed, the endpoint returns an UNAUTHORIZED response. 

## Additions to _portal.properties_
These properties are added:

```
# Turn cache management endpoint on or off (default)
# cache.endpoint.enabled=false

# API key for access to cache management endpoint
# cache.endpoint.api-key=fd15f1ae-66f2-4b8a-8d54-fb899b03557e
```

## Tests
Unit tests for the controller and service classes were added.

## Documentation
Where applicable, instructions to restart the cBioPortal tomcat server or restart of the docker container were amended with instructions to call the `/api/cache` endpoint.

### DIFF
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
@@ -67,10 +67,10 @@ public class CacheMapUtil {
     private static final int REPOSITORY_RESULT_OFFSET = 0; // retrieve all entries (do not skip any)
 
     // maps used to cache required relationships - in all maps stable ids are key
-    private Map<String, MolecularProfile> molecularProfileCache;
-    private Map<String, SampleList> sampleListCache;
-    private Map<String, CancerStudy> cancerStudyCache;
-    private Map<String, String> genericAssayStableIdToMolecularProfileIdCache;
+    private static Map<String, MolecularProfile> molecularProfileCache;
+    private static Map<String, SampleList> sampleListCache;
+    private static Map<String, CancerStudy> cancerStudyCache;
+    private static Map<String, String> genericAssayStableIdToMolecularProfileIdCache;
 
     public Map<String, MolecularProfile> getMolecularProfileMap() {
         return molecularProfileCache;
@@ -89,11 +89,17 @@ public class CacheMapUtil {
     }
 
     @PostConstruct
-    private void initializeCacheMemory() {
+    private void init() {
+        initializeCacheMemory();
+    }
+
+    public void initializeCacheMemory() {
+        
         // CHANGES TO THIS LIST MUST BE PROPAGATED TO 'GlobalProperties'
         this.cacheEnabled = (!authenticate.isEmpty() 
                 && !authenticate.equals("false") 
                 && !authenticate.contains("social_auth"));
+        
         if (cacheEnabled) {
             LOG.debug("creating cache maps for authorization");
             populateMolecularProfileMap();
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
@@ -162,7 +162,7 @@ public class CacheMapUtil {
             null,
             "ASC")) {
             // Only select GENERIC_ASSAY profiles
-            if (mp.getMolecularAlterationType().toString() == EntityType.GENERIC_ASSAY.toString()) {
+            if (EntityType.GENERIC_ASSAY.toString().equals(mp.getMolecularAlterationType().toString())) {
                 List<String> molecularId = new ArrayList<String>();
                 molecularId.add(mp.getStableId());
                 List<String> stableIds = genericAssayRepository.getGenericAssayStableIdsByMolecularIds(molecularId);
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
@@ -97,7 +97,7 @@ public class CacheMapUtil {
         initializeCacheMemory();
     }
 
-    public void initializeCacheMemory() {
+    public synchronized void initializeCacheMemory() {
         
         // CHANGES TO THIS LIST MUST BE PROPAGATED TO 'GlobalProperties'
         this.cacheEnabled = (!authenticate.isEmpty() 
@@ -114,9 +114,7 @@ public class CacheMapUtil {
     }
 
     private void populateMolecularProfileMap() {
-        if (molecularProfileCache == null) {
-            molecularProfileCache = new HashMap<String, MolecularProfile>();
-        }
+        molecularProfileCache = new HashMap<String, MolecularProfile>();
         for (MolecularProfile mp : molecularProfileRepository.getAllMolecularProfiles(
                 "SUMMARY",
                 REPOSITORY_RESULT_LIMIT,
@@ -129,9 +127,7 @@ public class CacheMapUtil {
     }
 
     private void populateSampleListMap() {
-        if (sampleListCache == null) {
-            sampleListCache = new HashMap<String, SampleList>();
-        }
+        sampleListCache = new HashMap<String, SampleList>();
         for (SampleList sl : sampleListRepository.getAllSampleLists(
                 "SUMMARY",
                 REPOSITORY_RESULT_LIMIT,
@@ -144,9 +140,7 @@ public class CacheMapUtil {
     }
 
     private void populateCancerStudyMap() {
-        if (cancerStudyCache == null) {
-            cancerStudyCache = new HashMap<String, CancerStudy>();
-        }
+        cancerStudyCache = new HashMap<String, CancerStudy>();
         for (CancerStudy cs : studyRepository.getAllStudies(
                 null,
                 "SUMMARY",
@@ -160,9 +154,7 @@ public class CacheMapUtil {
     }
 
     private void populateGenericAssayStableIdToMolecularProfileIdMap() {
-        if (genericAssayStableIdToMolecularProfileIdCache == null) {
-            genericAssayStableIdToMolecularProfileIdCache = new HashMap<String, String>();
-        }
+        genericAssayStableIdToMolecularProfileIdCache = new HashMap<String, String>();
         for (MolecularProfile mp : molecularProfileRepository.getAllMolecularProfiles(
             "SUMMARY",
             REPOSITORY_RESULT_LIMIT,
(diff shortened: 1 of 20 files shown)


persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
```diff
@@ -67,10 +67,10 @@
     private static final int REPOSITORY_RESULT_OFFSET = 0; // retrieve all entries (do not skip any)
 
     // maps used to cache required relationships - in all maps stable ids are key
-    private Map<String, MolecularProfile> molecularProfileCache;
-    private Map<String, SampleList> sampleListCache;
-    private Map<String, CancerStudy> cancerStudyCache;
-    private Map<String, String> genericAssayStableIdToMolecularProfileIdCache;
+    private static Map<String, MolecularProfile> molecularProfileCache;
```
Luke-Sikina, line 70: I know why you made this change, but I don't think anyone else does. Can you add a comment here explaining why these need to be static?

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -334,3 +334,8 @@ persistence.cache_type=no-cache
 # Installation map URL for installation map iframes. 
 # If blank or commented out, no map will be shown and that page will be hidden.
 # installation_map_url=https://installationmap.netlify.app/
+
+# Turn cache management endpoint on (default) or off
+# cache.endpoint.enabled=false
+# API key for access to cache management endpoint
+# cache.endpoint.api-key=fd15f1ae-66f2-4b8a-8d54-fb899b03557e
```
Luke-Sikina, line 341: This works as a means to authenticate cache clearing requests, but I don't like that it does not use spring security. I think in a perfect world,  we have a role for clearing the cache, and we use that for security. My guess is that you didn't do that because that pattern fails for portals that do not use authentication. 

I think this all comes down to the fact that portals without authentication still need a way to keep and verify secrets, and as of now, we don't have a standardized way of supporting that. This PR doesn't deserve to be scuttled while we talk about how to standardize this, but we should have that conversation.

web/src/main/java/org/cbioportal/web/CacheController.java
```diff
@@ -0,0 +1,43 @@
...
+    @Value("${cache.endpoint.enabled:true}")
+    private boolean cacheEndpointEnabled;
+
+    @RequestMapping(value = "/cache", method = RequestMethod.DELETE, produces = MediaType.TEXT_PLAIN_VALUE)
+    @ApiOperation("Flush and reinitialize caches")
+    public ResponseEntity<String> clearAllCaches(@RequestHeader("X-API-KEY") String providedApiKey) {
+        if (!cacheEndpointEnabled)
+            return new ResponseEntity<>("Cache endpoint is disabled for this instance.", HttpStatus.NOT_FOUND);
+        if ("not set".equals(requiredApiKey) || ! requiredApiKey.equals(providedApiKey))
+            return new ResponseEntity<>("", HttpStatus.UNAUTHORIZED);
```
Luke-Sikina, line 39: Bracket all `if`s

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
```diff
@@ -89,11 +89,17 @@
...
     @PostConstruct
-    private void initializeCacheMemory() {
+    private void init() {
+        initializeCacheMemory();
+    }
+
+    public void initializeCacheMemory() {
+        
         // CHANGES TO THIS LIST MUST BE PROPAGATED TO 'GlobalProperties'
         this.cacheEnabled = (!authenticate.isEmpty()
```
Luke-Sikina, line 99: I know you didn't change this, but is it even possible for `authenticate` to be empty? The annotation is `@Value("${authenticate:false}")`, so I think this is always true.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
```diff
@@ -89,11 +89,17 @@
     }
 
     @PostConstruct
-    private void initializeCacheMemory() {
+    private void init() {
+        initializeCacheMemory();
+    }
+
+    public void initializeCacheMemory() {
```
Luke-Sikina, line 96: I think this should be synchronized, especially since these maps aren't thread safe.
```suggestion
    public synchronized void initializeCacheMemory() {
```

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
```diff
@@ -89,11 +89,17 @@
...
+        
         // CHANGES TO THIS LIST MUST BE PROPAGATED TO 'GlobalProperties'
         this.cacheEnabled = (!authenticate.isEmpty() 
                 && !authenticate.equals("false") 
                 && !authenticate.contains("social_auth"));
+
```
Luke-Sikina, line 102: This comment is actually about the checks farther down that look like this:
```java
        if (molecularProfileCache == null) {
            molecularProfileCache = new HashMap<String, MolecularProfile>();
        }
```

Why do we need those? Shouldn't we always be clearing the map prior to populating? Seems to me that this setup lets deleted studies remain in the cache even after it's cleared. I'd just always make a new map.

Luke-Sikina (commented):
Great job! I am super excited to see this merge. This has been a major cause for frustration, and it's great to see that the solution is so straightforward. Most of my comments are pretty straight forward; I'll try to bring up my authentication concerns during planning today.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
```diff
@@ -89,11 +89,17 @@
...
 
     @PostConstruct
-    private void initializeCacheMemory() {
+    private void init() {
+        initializeCacheMemory();
+    }
+
+    public void initializeCacheMemory() {
+        
         // CHANGES TO THIS LIST MUST BE PROPAGATED TO 'GlobalProperties'
```
Luke-Sikina, line 98: To what list?

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
```diff
@@ -89,11 +89,17 @@
...
         this.cacheEnabled = (!authenticate.isEmpty() 
                 && !authenticate.equals("false") 
                 && !authenticate.contains("social_auth"));
+        
         if (cacheEnabled) {
```
Luke-Sikina, line 103: Line 169:
```java
if (mp.getMolecularAlterationType().toString() == EntityType.GENERIC_ASSAY.toString()) {
```
How is this working? Are we just getting lucky with the GC's string deduplication here? This should probably be changed to 
```java
if (EntityType.GENERIC_ASSAY.toString().equals(mp.getMolecularAlterationType().toString())) {
```

src/main/resources/portal.properties.EXAMPLE
pvannierop: That crossed my mind, but it will complicate automated import scripts that perform a cache flush at the end. But this might be my ignorance. We can discuss solutions here.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
pvannierop: Done.

web/src/main/java/org/cbioportal/web/CacheController.java
pvannierop: Done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
pvannierop: Done

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
pvannierop: yes, good point. Updated

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
pvannierop: This is not my comment. And I also did not make changed to that list :) So, I considered this to be not relevant to me.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
pvannierop: I did not write this code, but this indeed seems strange. I guess we have to ask @dippindots ?

dippindots: @pvannierop @Luke-Sikina This is definitely not correct, I think we should go with @Luke-Sikina's suggestion.

pvannierop: Ok, updated to code suggested by Luke.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapUtil.java
pvannierop: I think in theory this would be possible when portal.properties contains this:

```
authenticate=
```

Spring will interpret this as `authenticate` being intialized to empty string.

Luke-Sikina (approved):

PR #8677, opened 2021-06-09
Remove NA value from Generic Assay Data endpoints

dippindots:
Fix https://github.com/cbioportal/cbioportal/issues/8695

Currently, if data is not available, we require to put a NA in the data file, leave it blank is not allowed. However we should allow it to be blank since we won't expose those data anyway. So the task is:

- [x] Validator: Allow blank (empty) cell
- [x] Controller: Filter out blank (empty) values before send response
- [x] Documentation: Update documentation

### DIFF
--- a/web/src/main/java/org/cbioportal/web/GenericAssayController.java
+++ b/web/src/main/java/org/cbioportal/web/GenericAssayController.java
@@ -3,11 +3,13 @@ package org.cbioportal.web;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
+import org.apache.commons.lang3.StringUtils;
 import springfox.documentation.annotations.ApiIgnore;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.validation.Valid;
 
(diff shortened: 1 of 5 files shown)


dippindots:
@kalletlak @Luke-Sikina Could I get your review on this pr? 

This pr fixes two issues:
- `generic-assay-data` endpoints are not filtering out `NA` or empty values before return the results. See how we treat this in https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/MolecularDataController.java#L119-L172
- Previously, no blank cell is allowed in data file, empty value must be represented by `NA`, we should allow blank cell.

web/src/main/java/org/cbioportal/web/GenericAssayController.java
```diff
@@ -148,4 +150,17 @@ private void extractMolecularProfileAndSampleIds(GenericAssayDataMultipleStudyFi
...
+        List<GenericAssayData> result = new ArrayList<>();
+        genericAssayDataList.forEach(g -> {
+            String value = g.getValue();
+            // filter out empty string or "NA"
+            if (StringUtils.isNotEmpty(value) && !value.equals("NA")) {
+                result.add(g);
+            }
+        });
+
+        return result;
```
Luke-Sikina, line 164: We can make this a bit cleaner

```suggestion
        return genericAssayDataList.stream()
            .filter(g -> StringUtils.isNotEmpty(g.getValue()) && !g.getValue().equals("NA"))
            .collect(Collectors.toList());
```

web/src/main/java/org/cbioportal/web/GenericAssayController.java
```diff
@@ -121,16 +123,16 @@
 
         List<GenericAssayData> result;
         if (interceptedGenericAssayDataMultipleStudyFilter.getMolecularProfileIds() != null) {
-            result = genericAssayService.fetchGenericAssayData(
+            result = filterEmptyGenericAssayData(genericAssayService.fetchGenericAssayData(
```
Luke-Sikina, line 126: This genericAssayService method is also used in the `StudyViewFilterApplier`. Do we want to filter out blank and na values there as well? If yes, I think you should move this filtering logic to either service layer, or, if it's simple enough, to the SQL itself. If no, I think it's probably fine to keep the logic here.

kalletlak (commented):
LGTM. btw any specific reason to filter them?

dippindots:
@kalletlak we should do the filtering just like molecular data endpoint, filtering out them since we don't want a `NA` group shows on the plots tab. And plots tab won't know what's is our `Nullable` value, so we should filter out those empty data before send the response from backend.

Besides that, all of our implementation knows how to represent `Nullable` data, so we can just filter out those `Nullable` value.

web/src/main/java/org/cbioportal/web/GenericAssayController.java
dippindots: @Luke-Sikina Currently, we don't have this need to filter out blank and na values for study view, so I'd suggest to keep this as same as molecularDataController(https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/MolecularDataController.java#L119-L172)

dippindots:
@Luke-Sikina Thanks for your review, I updated this pr.

Luke-Sikina (approved):

PR #8681, opened 2021-06-11
Export namespace columns through Mutations endpoint

### LINKED ISSUE #8050: Namespaces feature does not work
Adding columns using namespaces as described [here](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#mutation-data) does not work because the validator rejects the study.

This is the error message:
Study has been validated with errors: Validation failed. Error code: 1; ERROR: data_mutations_extended.txt: ns namespace defined but MAF does not have any matching columns
ERROR: data_mutations_extended.txt: Invalid column header, file cannot be parsed

pvannierop:
Fix #8050

# Problem
Custom columns in the extended MAF file format imported through the _namespace_ mechanism are not returned by the mutations endpoint.

# Solution
This PR will export the custom columns from the mutations endpoint. For instance, when namespace `ZYGOSITY` is defined in the meta file and the data file has column `ZYGOSITY.status` with value `Homozygous` for a mutation row, the API will return the following JSON record for this mutation (only relevant fields are shown):

```shell script
{
    "namespaceColumns": {
        "zygosity": {
            "status": "Homozygous"
    },
}
```

# Tests
Tests for the MutationsController were added.

# Documentation
Documentation of the namespace import feature was updated with the example shown above.


Sjoerd-van-Hagen:
@inodb this is the PR to fix https://github.com/cBioPortal/cbioportal/issues/8050

Issue #8690, opened 2021-06-15
filtering by genes is broken in v3.6.19


inodb:
<img width="1050" alt="Screen Shot 2021-06-15 at 1 53 26 PM" src="https://user-images.githubusercontent.com/1334004/122100401-2363a280-cde1-11eb-91d6-5f81c67e8896.png">

Can see the issue in master env here: https://bit.ly/3xtvrXL

inodb:
@pvannierop @kalletlak - not entirely sure what the cause of this bug is but I'm guessing it's either https://github.com/cBioPortal/cbioportal/pull/8647 or https://github.com/cBioPortal/cbioportal/pull/8671 - any clue?

pvannierop:
@inodb @kalletlak The strange thing is that at first load the page errors out. However, when pressing reload the page loads fine.

PR #8692, opened 2021-06-17
Allow NA for MUTATION_STATUS field in /filtered-samples endpoint

pvannierop:
Fixes cBioPortal/cbioportal#8690

# Problem
With certain studies (e.g. _msk_impact_2017_) study view shows no samples after gene-based filtering.

# Cause
Studies with this problem have _NA_ in the MUTATION_STATUS column of the _mutation_ table. This possibility was not considered in the SQL statements used for driver filtering.

#  Solution
- SQL statement for driver filtering was updated to include NA.
- Regression tests are included.


kalletlak (approved):

inodb:
@pvannierop there's some discussion about this on slack. I think the proposal is for Mutation_Status do `Germline == Germline` and have everything else be `Somatic`. Would that work for your use case?

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MutationMapper.xml
```diff
@@ -216,16 +218,18 @@
...
                     <if test="geneFilterQuery.getIncludeGermline()">
                         OR LOWER(mutation.MUTATION_STATUS) = 'germline'
                     </if>
                     <if test="geneFilterQuery.getIncludeSomatic()">
-                        OR LOWER(mutation.MUTATION_STATUS) = 'somatic'
+                        OR LOWER(mutation.MUTATION_STATUS) = 'somatic' 
                     </if>
                     <if test="geneFilterQuery.getIncludeUnknownStatus()">
-                        OR LOWER(mutation.MUTATION_STATUS) IN ('unknown', '')
+                        OR LOWER(mutation.MUTATION_STATUS) NOT IN ('somatic', 'germline')
```
inodb, line 232: this logic should work for now but before enabling this on public portal we might want to play around with it to handle these cases:

Screen Shot 2021-06-17 at 12 58 59 PM

Another option might be to just leave mutation_status as is and not try to group them in somatic/germline/unknown. Then we would need to fix the data

inodb (approved):

inodb:
@pvannierop Thanks for fixing this!

PR #8702, opened 2021-06-21
Cache Generic Assay Meta endpoint and improvements

### LINKED ISSUE #8698: generic-assay-meta endpoint improvement
Recent added profiles introduced `20000` entities, this revealed the weakness of `generic-assay-meta` endpoint.

A few improvements can be made:
- [x] We can enable endpoint level cache since `generic-assay-meta` are not likely to change
- [x] Service layer improvement

dippindots:
Fix #8698 

- Cache Generic Assay Meta related endpoints
- Service layer improvement

### DIFF
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenericAssayMyBatisRepository.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenericAssayMyBatisRepository.java
@@ -5,6 +5,7 @@ import org.cbioportal.model.GenericAssayAdditionalProperty;
 import org.cbioportal.persistence.GenericAssayRepository;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Repository;
+import org.springframework.cache.annotation.Cacheable;
 import java.util.ArrayList;
 import java.util.List;
 import org.slf4j.Logger;
@@ -19,17 +20,20 @@ public class GenericAssayMyBatisRepository implements GenericAssayRepository {
     private static final Logger LOG = LoggerFactory.getLogger(GenericAssayMyBatisRepository.class);
 
     @Override
+    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public List<GenericAssayMeta> getGenericAssayMeta(List<String> stableIds) {
 
         return genericAssayMapper.getGenericAssayMeta(stableIds);
     }
 
     @Override
+    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public List<GenericAssayAdditionalProperty> getGenericAssayAdditionalproperties(List<String> stableIds) {
         return genericAssayMapper.getGenericAssayAdditionalproperties(stableIds);
     }
 
     @Override
+    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public List<String> getGenericAssayStableIdsByMolecularIds(List<String> molecularProfileIds) {
         
         List<Integer> molecularProfileInternalIds = genericAssayMapper.getMolecularProfileInternalIdsByMolecularProfileIds(molecularProfileIds);
--- a/service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
@@ -48,8 +48,10 @@ public class GenericAssayServiceImpl implements GenericAssayService {
         if (molecularProfileIds != null) {
             List<String> distinctMolecularProfileIds = molecularProfileIds.stream().distinct().sorted().collect(Collectors.toList());
             if (distinctMolecularProfileIds.size() > 0) {
-                List<String> stableIdsInMolecularProfiles = genericAssayRepository.getGenericAssayStableIdsByMolecularIds(distinctMolecularProfileIds);
-                allStableIds.addAll(stableIdsInMolecularProfiles);
+                // fetch one profile at a time to improve cache performace for multiple profiles query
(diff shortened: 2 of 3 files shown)


service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -80,15 +80,15 @@
                 }
             } else {
                 List<GenericAssayAdditionalProperty> additionalProperties = genericAssayRepository.getGenericAssayAdditionalproperties(distinctStableIds);
+                Map<String, List<GenericAssayAdditionalProperty>> additionalPropertiesGroupedByStableId =
```
Luke-Sikina, line 83: Line 77: `if (projection == "ID") {` - that can't be right, can it?

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenericAssayMyBatisRepository.java
```diff
@@ -19,17 +20,20 @@
...
     }
 
     @Override
+    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public List<GenericAssayAdditionalProperty> getGenericAssayAdditionalproperties(List<String> stableIds) {
         return genericAssayMapper.getGenericAssayAdditionalproperties(stableIds);
     }
 
     @Override
+    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
```
alisman, line 36: It would seem to me the cache keys here would be the set of molecular profile ids.  I would think we want to cache at the profileId level so that 1) different combinations of ids can still leverage cache, 2) we have a finite set of cache items, allowing us to use the static cache (instead of general).     
We can achieve thisby making an inner function that fetches by individual id and cache at that level.

However, another question: if these service level methods are just pass throughs to the data access layer (mapeprs), and there is no significant processing we save by caching here, then perhaps we should just cache at the data layer.   If that is so, then also need to evaluate the granularity of the cache there, again to make sure we cache at the id level, and not the group-of-ids level.

service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
```diff
@@ -80,15 +80,15 @@
...
+                    additionalProperties.stream().collect(Collectors.groupingBy(GenericAssayAdditionalProperty::getStableId));
```
Luke-Sikina, line 84: `additionalProperties` is only being used in the next line, so you might as well clean this up a bit:
```suggestion
                Map<String, List<GenericAssayAdditionalProperty>> additionalPropertiesGroupedByStableId =
                    genericAssayRepository.getGenericAssayAdditionalproperties(distinctStableIds).stream()
                        .collect(Collectors.groupingBy(GenericAssayAdditionalProperty::getStableId));```

File: service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java [hunk 4]
```diff
@@ -80,15 +80,15 @@
...
-                            .filter(additionalProperty -> additionalProperty.getStableId().equals(stableId))
-                            .collect(Collectors.toList());
-                    for (GenericAssayAdditionalProperty additionalProperty : filteredAdditionalProperties) {
-                        map.put(additionalProperty.getName(), additionalProperty.getValue());
+                    if (additionalPropertiesGroupedByStableId.containsKey(stableId)) {
+                       for (GenericAssayAdditionalProperty additionalProperty : additionalPropertiesGroupedByStableId.get(stableId)) {
+                           map.put(additionalProperty.getName(), additionalProperty.getValue());
+                       }
                     }
                     meta.setGenericEntityMetaProperties(map);
```
Luke-Sikina, line 93: I'm not going to make you change this because I'd rather keep this pull request small, but in general, you should declare map types as `Map<`, rather than `HashMap<` even when the field is guaranteed to be initialized as a HashMap. If you aren't using any features specific to HashMap, this just makes your types a bit more honest about their requirements.

Luke-Sikina: Same goes for List / ArrayList, Set / HashSet

Luke-Sikina (approved):
Just a few tiny comments.

service/src/main/java/org/cbioportal/service/impl/GenericAssayServiceImpl.java
dippindots: This is not correct... updated

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenericAssayMyBatisRepository.java
dippindots: @alisman updated to fetch data for each profile, I don't really understand the second part, do you mean I should enable cache on mappers instead of repositories?

alisman (approved):

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/GenericAssayMyBatisRepository.java
```diff
@@ -19,17 +20,20 @@
...
         return genericAssayMapper.getGenericAssayMeta(stableIds);
     }
 
     @Override
+    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public List<GenericAssayAdditionalProperty> getGenericAssayAdditionalproperties(List<String> stableIds) {
         return genericAssayMapper.getGenericAssayAdditionalproperties(stableIds);
     }
 
     @Override
```
alisman, line 35: problem is the cachekey will still be the particular combination of ids

alisman (changes requested):
Still need to work on caching

alisman (approved):

PR #8745, opened 2021-07-12
Add documentation for structural variants

inodb:
We haven't previously added the structural variants format documentation. I believe this should also deprecate the fusion format afaik, so indicated those as deprecated. This was mostly an attempt to capture all the columns from this [example file](https://github.com/cBioPortal/datahub/blob/master/public/msk_impact_2017/data_fusions.txt) as well as those defined in the [sql schema](https://github.com/cBioPortal/cbioportal/blob/master/db-scripts/src/main/resources/cgds.sql#L389-L431). I used some of the info from [this doc](https://docs.google.com/document/d/1BF2ewnDHl81nSS9xWE_1TjkdnPIcmjQ6_YgKCRRsm3s/edit#) for reference.

TODO:

- [x] Make it more clear which columns are optional

Comments:

- There isn't a clear mapping for each column header in the file to a database field afaik. Might be good to investigate some of these, but maybe ok to add those later since they seem to be optional (e.g. NCBI_BUILD, DNA_SUPPORT, RNA_SUPPORT)

### DIFF
--- a/docs/File-Formats.md
+++ b/docs/File-Formats.md
@@ -905,13 +905,15 @@ A structural variant data file is a tab-delimited file with one structural varia
 | Field                        | Allowed/Example values         | Comments |
 | -- | -- | -- |
 | Sample_ID                    | Sample_1     | This is the sample ID  |
-| Site1_Hugo Symbol  | ERG | A [HUGO](https://www.genenames.org/) gene symbol.  |
+| Site1_Hugo_Symbol  | ERG | A [HUGO](https://www.genenames.org/) gene symbol.  |
+| Site1_Entrez_Gene_Id  | Entrez gene id of ERG |  |
 | Site1_Ensembl_Transcript_Id  | ENST00000288319 | An [Ensembl Transcript](https://useast.ensembl.org/Help/View?id=151#:~:text=Each%20transcript%20ID%20includes%20a,ENSMUST%20defines%20a%20mouse%20transcript).) identifier.  |
 | Site1_Exon                   | 4 | |
 | Site1_Chromosome             | 21 |      |
 | Site1_Position               | 39842043 |                |
 | Site1_Description           | |               |
 | Site2_Hugo Symbol  | TMPRSS2 | A [HUGO](https://www.genenames.org/) gene symbol.  |
+| Site2_Entrez_Gene_Id  | Entrez gene id of TMPRSS2 |  |
 | Site2_Ensembl_Transcript_Id  | ENST00000288319 | An [Ensembl Transcript](https://useast.ensembl.org/Help/View?id=151#:~:text=Each%20transcript%20ID%20includes%20a,ENSMUST%20defines%20a%20mouse%20transcript).) identifier.  |
 | Site2_Exon                   | 2 | |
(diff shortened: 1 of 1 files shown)


docs/File-Formats.md
```diff
@@ -868,7 +869,74 @@ AKT1 AKT2 10000|AKT<TAB>0.17071492725<TAB>0.264067254391
...
+data_filename: data_SV.txt
+```
+
+### Data file
+A structural variant data file is a tab-delimited file with one structural variant per row.  For each structural variant (row) in the data file, the following tab-delimited values are required:
+
+| Field                        | Allowed/Example values         | Comments |
+| -- | -- | -- |
+| Sample_ID                    | Sample_1     | This is the sample ID  |
+| Site1_Hugo Symbol  | ERG | A [HUGO](https://www.genenames.org/) gene symbol.  |
```
n1zea144, line 908: Missing "_" between Site1_Hugo and Symbol.

docs/File-Formats.md
```diff
@@ -868,7 +869,74 @@ AKT1 AKT2 10000|AKT<TAB>0.17071492725<TAB>0.264067254391
...
+datatype: SV
+stable_id: structural_variants
+show_profile_in_analysis_tab: true
+profile_name: mskimpact2017 SV Data
+profile_description: Structural Variant Data for mskimpact2017
+data_filename: data_SV.txt
+```
+
+### Data file
+A structural variant data file is a tab-delimited file with one structural variant per row.  For each structural variant (row) in the data file, the following tab-delimited values are required:
```
n1zea144, line 903: There are missing fields here as compared to [cgds.sql](https://github.com/cBioPortal/cbioportal/blob/8d4c7fb40a996ebcfffbcc172e4e03eb6b7cb671/db-scripts/src/main/resources/cgds.sql#L389), like SITE2_EFFECT_ON_FRAME, ENTREZ_GENE_IDs, DNA/RNA_SUPPORT, EXTERNAL_ANNOTATION.  Is this intentional?

n1zea144 (commented):
I found the following utility class in the codebase which describes minimum requirements, like Site1_Hugo_Symbol and Site2_Hugo_Symbol are required (or Entrez_Id).  If I understand this correctly, transcript ids and exon information has to be provided together for both sites:
https://github.com/cBioPortal/cbioportal/blob/8d4c7fb40a996ebcfffbcc172e4e03eb6b7c[…]main/java/org/mskcc/cbio/portal/util/StructuralVariantUtil.java

cc @ao508

docs/File-Formats.md
inodb: I added them in. I was not entirely sure about whether this should really be in the model, but we can leave them for now and just let the docs describe the state as is

n1zea144: Yeah agreed.  I think they came out of one or two discussions that involved Niki & Trevor.  It sounded like something we might want to have, but I don't think we were entirely sure...it did make it into the model.

n1zea144 (approved):

inodb:
Approved verbally by JJ & Karthik as well

Issue #8680, opened 2021-06-10
[SECURITY] Denial of service because of unsafe regex processing


edvraa:
I have tried to contact you by cbioportal@cbio.mskcc.org and asked for any other email in https://github.com/cBioPortal/cbioportal/issues/8658. Nobody replied.

The cBioPortal is vulnerable to regex injection that may lead to Denial of Service.

User controlled `heatmap` and `alteration` are used to build and run a regex expression:
https://github.com/cBioPortal/cbioportal/blob/e086d40f17ac037161677d5478a5dc5ec710bbd8/core/src/main/java/org/mskcc/cbio/portal/servlet/ProteinArraySignificanceTestJSON.java#L104-L106
The value end up in getAlteredCases https://github.com/cBioPortal/cbioportal/blob/e086d40f17ac037161677d5478a5dc5ec710bbd8/core/src/main/java/org/mskcc/cbio/portal/servlet/ProteinArraySignificanceTestJSON.java#L279-L282

Since the attacker controls the string and the regex pattern he may cause a ReDoS by regex catastrophic backtracking on the server side.

inodb:
CC @adamabeshouse

inodb:
Thanks for reporting! I don't think this code is used anymore actually? Maybe we can delete it?

edvraa:
Hi,
Do you plan to release a GitHub security advisory and/or request CVE number?

edvraa:
Oh, did you just close the issue without fixing the code? Even if the parameters are not used the code is still callable. Do I miss something?

edvraa:
I thought since the issue was so brutally closed without explanation maybe my code analysis is wrong and it is not expoitable. Thus I have followed the instructions from https://docs.cbioportal.org/2.1.1-deploy-with-docker-recommended/docker and ran a local instance of cbioportal in container. I have a proof of concept when just a single request makes server cpu to consume 100% indefinetely. Please create a [security advisory](https://docs.github.com/en/code-security/security-advisories/creating-a-security-advisory) where you could invite me and discuss it in private if you have any questions.

It makes me sad that such a noble project makes it hard to responsibly disclose a security issue that may potentially lead to Denial of Service. Please respond in 24 hours.

jjgao:
[...]

PR #8751, opened 2021-07-15
Delete unused ProteinArraySignificanceTest endpoint with security issue

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8680


sheridancbio (changes requested):
If this servlet is being removed, I think we also need to drop the servlet beans from portal/src/main/webapp/WEB-INF/web.xml  (lines 341 - 348). Otherwise I expect we would see a BeanNotFound Exception during startup ... or maybe when the servlet is first accessed.

From the issue comments it seems like there is already approval to drop this service from the core. If that is so, this is a first step towards our larger cleanup (jettisoning the core).

edvraa:
If I may suggest. Keep the class. Make `processRequest` return empty result early if it is really not used. That would be the quickest less intrusive security fix without making too much refactoring.

n1zea144 (approved):

n1zea144:
@adamabeshouse Why not just include the update to web.xml as Rob suggests?  Seems like an easy update to the PR.

adamabeshouse:
@n1zea144 @sheridancbio done! thanks

n1zea144 (approved):

PR #8769, opened 2021-07-20
Identify genes outside the sequencing panel - for profiles in gene matrix file

rmadupuri:
- Reports an ERROR message if the gene identified in the genomic profile (maf, cna) is not from the sequencing panel.
- We previously had this check for mutations profile, and this PR extends this functionality to copy number profiles. 
- Currently, the gene matrix files list only mutations and cna panels and this check can be extended to other profiles in the future.


dippindots (commented):
@rmadupuri Looks good to me! It would be great if we have have some small tests for these new changes. Could you add some unit tests for it at here: https://github.com/cBioPortal/cbioportal/blob/master/core/src/test/scripts/unit_tests_validate_data.py?

rmadupuri:
This PR is on hold, as we noticed most of the public studies with panels fail the validation if this piece of code is introduced. 
We are investigating what needs to be done in such cases (Remove the off panel gene related data from profiles? Reach out to the authors? Or change the ERROR to WARNING in validator?)

PR #8779, opened 2021-07-26
Update fetchMolecularProfileSampleCounts endpoint to support patient level count

dippindots:
Backend support of  https://github.com/cBioPortal/cbioportal-frontend/pull/3772

We are getting samples/patient count from backend endpoint `fetchMolecularProfileSampleCounts`, so we should update this function to support patient level molecular profiles.

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/StudyViewServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/StudyViewServiceImpl.java
@@ -34,16 +34,21 @@ public class StudyViewServiceImpl implements StudyViewService {
         List<MolecularProfileCaseIdentifier> molecularProfileSampleIdentifiers =
             molecularProfileService.getMolecularProfileCaseIdentifiers(studyIds, sampleIds);
 
-        Map<String, Integer> molecularProfileSampleCountSet = genePanelService
+        List<MolecularProfile> molecularProfiles = molecularProfileService
+            .getMolecularProfilesInStudies(new ArrayList<>(new HashSet<>(studyIds)),"SUMMARY");
+        Map<String, MolecularProfile> molecularProfileMap = molecularProfiles
+            .stream()
+            .collect(Collectors.toMap(MolecularProfile::getStableId, Function.identity()));
+
+        Map<String, Integer> molecularProfileCaseCountSet = genePanelService
             .fetchGenePanelDataInMultipleMolecularProfiles(molecularProfileSampleIdentifiers)
             .stream()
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/cbioportal/service/impl/StudyViewServiceImpl.java
```diff
@@ -56,7 +65,7 @@
                     .getValue()
                     .stream()
                     .mapToInt(molecularProfile ->
-                        molecularProfileSampleCountSet.getOrDefault(molecularProfile.getStableId(), 0))
+                        BooleanUtils.isTrue(molecularProfile.getPatientLevel()) ? molecularProfilePatientCountSet.getOrDefault(molecularProfile.getStableId(), 0) : molecularProfileSampleCountSet.getOrDefault(molecularProfile.getStableId(), 0))
```
Luke-Sikina, line 68: Just a thought - rather than having this `isTrue` boilerplate everywhere, why not add that logic and an `@NotNull` annotation to the getter in MolecularProfile. The current approach seems like a bug waiting to happen, especially when you consider line 908 of StudyViewFilterApplier.

Luke-Sikina (approved):
Don't know if you still need my review or not. I had a few small comments, but it seems reasonable.

service/src/main/java/org/cbioportal/service/util/PatientLevelProfileUtil.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+@Component
+public class PatientLevelProfileUtil {
+    @SafeVarargs
+    public static <T> Predicate<T> distinctByKeys(Function<? super T, ?>... keyExtractors)
```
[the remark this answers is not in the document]
dippindots: Thanks @Luke-Sikina! @kalletlak gave me a suggestion which will not need this component anymore, so I will remove it. Thanks for your suggestion on this.

service/src/main/java/org/cbioportal/service/impl/StudyViewServiceImpl.java
dippindots: This is a good idea and it's true, I will add this annotation!

Luke-Sikina (approved):

kalletlak (approved):

PR #8793, opened 2021-07-29
Optimize coexpression endpoint

### LINKED ISSUE #8775: Improve Co-Expression Performance
It is slow.

Rajat-Sirohi:
Fix #8775

Speedup `CoExpressionService` endpoint via multithreading and faster SQL logic. The multithreading capability is accessible throughout the service layer simply by adding an `@Async` annotation to some method and ensuring that it is called from somewhere outside its own class.

Unit tests and the portal configuration property `multithreading.core_pool_size` (with default value 16) have also been added.

Rates of speedup vary across queries but generally hover somewhere around a 2x improvement.

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/MolecularDataRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/MolecularDataRepository.java
@@ -18,12 +18,18 @@ public interface MolecularDataRepository {
     @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
     Map<String, MolecularProfileSamples> commaSeparatedSampleIdsOfMolecularProfilesMap(Set<String> molecularProfileIds);
 
-    @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
+    // Not caching when entrezGeneIds is null or empty because the large response size sometimes crashes the cache
+    @Cacheable(cacheResolver = "generalRepositoryCacheResolver",
+               condition = "@cacheEnabledConfig.getEnabled() && #entrezGeneIds != null && #entrezGeneIds.size() != 0")
     List<GeneMolecularAlteration> getGeneMolecularAlterations(String molecularProfileId, List<Integer> entrezGeneIds,
                                                               String projection);
 
     Iterable<GeneMolecularAlteration> getGeneMolecularAlterationsIterable(String molecularProfileId, List<Integer> entrezGeneIds,
                                                                           String projection);
+
+    // Same as getGeneMolecularAlterationsIterable above, except assumes that
+    // entrezGeneIds is null or empty AND projection is "SUMMARY"
+    Iterable<GeneMolecularAlteration> getGeneMolecularAlterationsIterableFast(String molecularProfileId);
 
     @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
     List<GeneMolecularAlteration> getGeneMolecularAlterationsInMultipleMolecularProfiles(Set<String> molecularProfileIds,
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
@@ -91,6 +91,19 @@
         </where>
     </select>
 
+    <!-- This routine is an abbreviated copy of getGeneMolecularAlterationsIter above. The two should be kept in sync. -->
+    <select id="getGeneMolecularAlterationsIterFast" resultType="org.cbioportal.model.GeneMolecularAlteration">
+        SELECT
+        gene.ENTREZ_GENE_ID AS entrezGeneId,
+        genetic_alteration.VALUES AS "values"
+        FROM genetic_alteration
+        INNER JOIN genetic_profile ON genetic_alteration.GENETIC_PROFILE_ID = genetic_profile.GENETIC_PROFILE_ID
+        INNER JOIN gene ON genetic_alteration.GENETIC_ENTITY_ID = gene.GENETIC_ENTITY_ID
+        <where>
+            genetic_profile.STABLE_ID = #{molecularProfileId}
+        </where>
+    </select>
+
     <select id="getGeneMolecularAlterationsInMultipleMolecularProfiles" resultType="org.cbioportal.model.GeneMolecularAlteration">
         SELECT
         gene.ENTREZ_GENE_ID AS entrezGeneId,
--- a/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
@@ -1,27 +1,27 @@
 package org.cbioportal.service.impl;
 
-import org.apache.commons.lang3.math.NumberUtils;
-import org.apache.commons.math3.linear.Array2DRowRealMatrix;
-import org.apache.commons.math3.linear.RealMatrix;
-import org.apache.commons.math3.stat.correlation.SpearmansCorrelation;
 import org.cbioportal.model.*;
 import org.cbioportal.persistence.MolecularDataRepository;
 import org.cbioportal.persistence.SampleListRepository;
 import org.cbioportal.service.*;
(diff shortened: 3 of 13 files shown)


service/src/main/java/org/cbioportal/service/util/AppConfig.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.concurrent.Executor;
+
+@Configuration
+@EnableAsync
+public class AppConfig implements AsyncConfigurer {
+
+    @Override
+    public Executor getAsyncExecutor() {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(16);
```
Luke-Sikina, line 17: We should probably make this configurable.

service/src/main/java/org/cbioportal/service/util/CoExpressionAsyncMethods.java
```diff
@@ -0,0 +1,68 @@
...
+    @Async
+    public CompletableFuture<CoExpression> computeCoExpression(String entityId, List<String> valuesA, List<String> valuesB,
+            Boolean isMolecularProfileBOfGenesetType, Double threshold, String molecularProfileId) throws GenesetNotFoundException, GeneNotFoundException {
+
+        List<String> valuesACopy = new ArrayList<>(valuesA);
+        List<String> valuesBCopy = new ArrayList<>(valuesB);
+
+        List<Integer> valuesToRemove = new ArrayList<>();
+        for (int i = 0; i < valuesBCopy.size(); i++) {
+            if (!NumberUtils.isNumber(valuesBCopy.get(i)) || !NumberUtils.isNumber(valuesACopy.get(i))) {
```
Luke-Sikina, line 30: `isNumber` is deprecated.
```suggestion
            if (!NumberUtils.isCreatable(valuesBCopy.get(i)) || !NumberUtils.isCreatable(valuesACopy.get(i))) {
```

service/src/main/java/org/cbioportal/service/util/CoExpressionAsyncMethods.java
```diff
@@ -0,0 +1,68 @@
...
+                valuesToRemove.add(i);
+            }
+        }
+
+        for (int i = 0; i < valuesToRemove.size(); i++) {
+            int valueToRemove = valuesToRemove.get(i) - i;
+            valuesBCopy.remove(valueToRemove);
+            valuesACopy.remove(valueToRemove);
+        }
```
Luke-Sikina, line 39: I know you just copied this, but this whole section seems not great. I think this code is weird looking because we're trying to remove elements from the collection we're iterating over _and_ we're iterating over two collections simultaneously, so there's a nasty side effect. That said, we can still clean this up. Here's a helper function I came up with that accomplishes the same thing. You can use it if you'd like, just make sure you benchmark it if you do.

(Note the non short circuit or - I think this is the first time I've had to write one of those in like 2 years!)

```java
    private void removeNonNumbersFromBothLists(ArrayList<String> valuesACopy, ArrayList<String> valuesBCopy) {
        Iterator<String> itA = valuesACopy.listIterator();
        Iterator<String> itB = valuesBCopy.listIterator();
        while (itA.hasNext() && itB.hasNext()) {
            if (!NumberUtils.isCreatable(itA.next()) | !NumberUtils.isCreatable(itB.next())) {
                itA.remove();
                itB.remove();
            }
        }
    }
```

service/src/main/java/org/cbioportal/service/util/AppConfig.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.scheduling.annotation.AsyncConfigurer;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+import java.util.concurrent.Executor;
+
+@Configuration
+@EnableAsync
+public class AppConfig implements AsyncConfigurer {
```
Luke-Sikina, line 13: Looks good to me. I'm going to add Ben as a reviewer because of this though. I think it's correct, but he should know when we make a significant addition to the backend config.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -91,6 +91,19 @@
         </where>
     </select>
 
+    <!-- This routine is an abbreviated copy of getGeneMolecularAlterationsIter above. The two should be kept in sync. -->
```
Luke-Sikina, line 94: So what's the difference here? No projection checks? How much does this actually speed things up?

service/src/main/java/org/cbioportal/service/impl/CoExpressionServiceImpl.java
```diff
@@ -176,20 +176,19 @@
...
-            CoExpression ce = computeCoExpressions(entityId, values, includedQueryValues, isMolecularProfileBOfGenesetType, threshold, molecularProfileId);
-            if (ce != null) {
-                toReturn.add(ce);
-            }
+            CompletableFuture<CoExpression> future = asyncMethods
+                .computeCoExpression(entityId, values, includedQueryValues, isMolecularProfileBOfGenesetType, threshold, molecularProfileId);
+            returnFutures.add(future);
         }
-        return toReturn;
+        return returnFutures.stream().filter(Objects::nonNull).map(CompletableFuture::join).collect(Collectors.toList());
```
Luke-Sikina, line 191: Couldn't you stream all of this, instead of starting off with a for loop? I think you have to remove the exception declarations from computeCoExpression, but it's not like we're handling those elegantly anyways.

```java
        return StreamSupport.stream(maItr.spliterator(), false)
            .filter(ma -> queryGeneticEntityId.equals(ma.getStableId()))
            .map(ma -> {
                String entityId = ma.getStableId();
                List<String> internalValues = new ArrayList<>(Arrays.asList(ma.getSplitValues()));
                List<String> values = includedIndexes.stream().map(index -> internalValues.get(index)).collect(Collectors.toList());
                return asyncMethods.computeCoExpression(entityId, values, includedQueryValues, isMolecularProfileBOfGenesetType, threshold, molecularProfileId);
            })
            .map(CompletableFuture::join)
            .filter(Objects::nonNull)
```

service/src/test/java/org/cbioportal/service/util/CoExpressionAsyncMethodsTest.java
```diff
@@ -0,0 +1,107 @@
...
+        valuesA3.add("1"); valuesA3.add("4"); valuesA3.add("0");
+        allValuesA.add(valuesA3);
+        return allValuesA;
+    }
+
+    private List<String> createValuesB() {
+        List<String> valuesB = new ArrayList<>();
+        valuesB.add("2.1"); valuesB.add("3"); valuesB.add("3");
+        return valuesB;
+    }
```
Luke-Sikina, line 106: This seems unnecessary. I'd rather you inline something like this:

```suggestion
    new ArrayList<>(Arrays.asList("2.1", "3", "3"));
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/MolecularDataRepository.java
```diff
@@ -18,13 +18,18 @@
     @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
     Map<String, MolecularProfileSamples> commaSeparatedSampleIdsOfMolecularProfilesMap(Set<String> molecularProfileIds);
 
-    @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
+    @Cacheable(cacheResolver = "generalRepositoryCacheResolver",
+               condition = "@cacheEnabledConfig.getEnabled() && #entrezGeneIds != null && #entrezGeneIds.size() != 0")
```
Luke-Sikina, line 22: Good. Let's add a comment here explaining why we're not caching those requests.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularDataMapper.xml
Rajat-Sirohi: No projection check or gene id nullity/empty check. Although it seems trivial, I was actually getting non-negligible speedups (not huge, but better than nothing). I think the reason is just because the table is quite large, so skipping those checks for every single row makes a difference

Luke-Sikina (approved):

PR #8795, opened 2021-07-29
Check and remove invalid foreign key in migrate_db.py

dippindots:
This is a temporary solution to fix a issue related to database migration:

Issue description:
If someone has a very old database: version `<= 1.4.0`, and try to migrate to a version `>= 2.12.1`

There will be a invalid and unwanted foreign key exist in the database: `genetic_alteration_fk_2`
We need to check and delete that key.


inodb (approved):
Neat - thank you!

PR #8815, opened 2021-08-05
Add modal to HTML report to visualize all error/warning values

oplantalech:
Currently, the report does not allow to visualize more than 3 genes/values that fall under the same warning or error.

In this PR a link is introduced in each log that has more than 3 values. The link opens a modal displaying the full content of the log. The validator/importer also allowed you to change this threshold value of 3, which has now been removed for simplicity.

Image of the new report with a modal opened:
Screenshot from 2021-08-05 16-31-43

### DIFF
--- a/core/src/main/scripts/importer/cbioportal_common.py
+++ b/core/src/main/scripts/importer/cbioportal_common.py
@@ -554,8 +554,16 @@ class CollapsingLogMessageHandler(logging.handlers.MemoryHandler):
                                  record.getMessage())
             if identifying_tuple not in grouping_dict:
                 grouping_dict[identifying_tuple] = []
-            grouping_dict[identifying_tuple].append(record)
 
+            # add a tuple that keeps the relation between the line number,
+            # the column number and the cause for the HTML report
+            record.__dict__['value'] = (
+                getattr(record, 'line_number', None),
+                getattr(record, 'column_number', None),
+                getattr(record, 'cause', None))
+
+            grouping_dict[identifying_tuple].append(record)
+        
         aggregated_buffer = []
         # for each list of same-message records
         for record_list in list(grouping_dict.values()):
--- a/core/src/main/scripts/importer/validation_report_template.html.jinja
+++ b/core/src/main/scripts/importer/validation_report_template.html.jinja
@@ -128,6 +128,7 @@
                 <th>Column Number</th>
                 <th>Message</th>
                 <th>Value Encountered</th>
+                <th></th>
               </tr>
             </thead>
             <tbody>
@@ -144,11 +145,8 @@
               <tr class="{{ record_class }}">
                 {% macro format_aggregated(record, attr_name) %}
                     {% if record[attr_name + '_list'] is defined -%}
-                      {% if (record['show_all_values'] is not defined or record['show_all_values'] == False) and (record[attr_name + '_list']|length) > max_reported_values -%}
-                              {{ (record[attr_name + '_list'][:max_reported_values]|join(', ') +
-                                  ', (' +
-                                  record[attr_name + '_list'][max_reported_values:]|length|string +
-                                  ' more)')|e }}
+                      {% if (record['show_all_values'] is not defined or record['show_all_values'] == False) and (record[attr_name + '_list']|length) > 3 -%}
+                              {{ record[attr_name + '_list'][:3]|join(', ')|e }}{{ '&hellip;' }}
                           {%- else -%}
                                   {{ (record[attr_name + '_list']|join(', '))|e }}
                           {%- endif %}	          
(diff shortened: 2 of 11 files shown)


core/src/main/scripts/importer/cbioportal_common.py
```diff
@@ -554,8 +554,22 @@ def flush(self):
                                  record.getMessage())
             if identifying_tuple not in grouping_dict:
                 grouping_dict[identifying_tuple] = []
-            grouping_dict[identifying_tuple].append(record)
 
+            #Add a tuple that keeps the relation between the line number,
```
fedde-s, line 558: The first thing that distracted me was the different comment style from the ones above and below, but I should probably try not to let that distract me.

core/src/main/scripts/importer/cbioportal_common.py
```diff
@@ -554,8 +554,22 @@ def flush(self):
...
+            #the column number and the cause for the HTML report
+            line_number = None
+            column_number = None
+            cause = None
+            if 'line_number' in record.__dict__:
```
fedde-s, line 563: This block of code looks overcomplicated, talking about internal instance dicts and looking for the attribute twice; I'd recommend making the triple in this way instead:
```python
(
    getattr(record, 'line_number', None),
    getattr(record, 'column_number', None),
    getattr(record, 'cause' None))
```

core/src/test/scripts/test_data/study_es_0/result_report.html
```diff
@@ -36,7 +36,7 @@
   <!--[if lt IE 9]>
     <script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
   <![endif]-->
-  <title>cBioPortal validation report: study in 'test_data/study_es_0/'</title>
+  <title>cBioPortal validation report: study in '/study'</title>
```
fedde-s, line 39: Is this correct? And is it odd that the only test you touched was about an HTML report without warnings or errors?

fedde-s (commented):
I found a few small things that I had some doubts about, and also wondered: do any tests turn red if the new functionality does not work?

SRodenburg (approved):

core/src/main/scripts/importer/cbioportal_common.py
oplantalech: @fedde-s @SRodenburg Thanks for the tip

core/src/main/scripts/importer/validation_report_template.html.jinja
```diff
@@ -167,6 +166,60 @@
...
+                        {% set num_values = record[attr_name + '_list']|length|string %}
+                        {% set modal_name = "modal" + filename.rsplit('.', 1)[0] + num_record + attr_name %}
+                        <a href="#showValues" data-toggle="modal" data-backdrop="false" data-target=".{{ modal_name }}">See all {{ num_values }} values </a>
+
+                        <div class="modal fade {{ modal_name }}" tabindex="-1" role="dialog" aria-labelledby="myLargeModalLabel">
+                          <div class="modal-dialog modal-lg" role="document">
+                            <div class="modal-content">
+                              <div class="modal-header">
+                                <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+                                <h4 class="modal-title" id="gridSystemModalLabel">Listing all values for the message: {{ record.getMessage() }}</h4>
```
fedde-s, line 182: Aren't you placing the message into the HTML without escaping characters like `<` here? The `record.getMessage()` call above this new macro definition does apply `e` filter, suggesting that the return value is  _not_ supposed to be parsed as HTML.

core/src/main/scripts/importer/validation_report_template.html.jinja
```diff
@@ -167,6 +166,60 @@
...
+                                    <tr>
+                                      <th>Line Number</th>
+                                      <th>Column Number</th>
+                                      <th>Value Encountered</th>
+                                    </tr>
+                                  </thead>
+                                  <tbody>
+                                    {% for value in record[attr_name + '_list'] %}
+                                      <tr>
+                                        <td>{% if value[0] is not none -%}
```
fedde-s, line 196: It's been a long time since I wrote Jinja2 code, but these `if` blocks look a bit drawn out and repetitive. Do you think something like [inline if expressions](https://jinja.palletsprojects.com/en/2.11.x/templates/#if-expression) or a macro could make this block less cumbersome to read?

In Python even indexing tuples with integers like this is usually not the most readable way to do things, but I imagine that Jinja2 isn't as fancy in that regard, it's designed for templating and not for logic.

fedde-s (changes requested):

core/src/main/scripts/importer/validation_report_template.html.jinja
fedde-s: As I said, too long ago, but I imagine that ‘undefined’ in the context of the [`default()`](https://jinja.palletsprojects.com/en/2.11.x/templates/#default) filter does not include ‘having a value of `none`’?

core/src/main/scripts/importer/validation_report_template.html.jinja
oplantalech: Thanks @fedde-s for spotting this!

core/src/main/scripts/importer/validation_report_template.html.jinja
oplantalech: Thanks again for suggesting the inline if - the code looks much better now. And as you mention, unfortunately the `default()` filter does not include `none` as an undefined value.

core/src/main/scripts/importer/validation_report_template.html.jinja
fedde-s: The problem seems to have been solved for this particular line, but still exists for every other insertion of data from the input file into the HTML code.

fedde-s (dismissed):
The issues I noticed, both for code readability/maintainability and for input handling, look resolved.

SRodenburg (approved):

fedde-s:
Don't wait for me; unless one of you is aware of something major, I don't imagine I would have much to add with a re-review.

PR #8382, opened 2021-02-25
Clean up of unused legacy database fields

jagnathan:
Remove several fields from a variety of tables:

- Fix #7820; Table Type_of_cancer remove field CLINICAL_TRIAL_KEYWORDS
- Fix #6375; Table Cancer_study remove field SHORT_NAME (requires https://github.com/cBioPortal/cbioportal-frontend/pull/3632)
- Fix #3088; Table Sample remove field TYPE_OF_CANCER_ID

PR #8819, opened 2021-08-06
Fix validator unit tests matching database cleanup

oplantalech:
Fix validator tests that have been broken due to changes in the database fields (PR #8382).


inodb (approved):
Perfect! Thanks so much!

PR #8821, opened 2021-08-06
Fix integration tests

oplantalech:
Fix integration tests that were failing due to the recent update of the seed database:

- Gene ACPP is now ACP3
- microRNA gene MIR4714 is not in the seed database.
- Use public database for the OncoKB import integration test


test/integration/integration_test_oncokb_import.sh
```diff
@@ -6,7 +6,7 @@
 cd "$PORTAL_SOURCE_DIR"/core/src/main/scripts/importer
 python3 ./metaImport.py \
   --study_directory="$PORTAL_SOURCE_DIR"/core/src/test/scripts/test_data/study_oncokb_import \
-  --portal_info_dir="$PORTAL_SOURCE_DIR"/core/src/test/scripts/test_data/api_json_system_tests \
+  --url_server="https://www.cbioportal.org" \
   --import_oncokb \
```
inodb, line 10: @oplantalech should we update the test_data as well?

inodb (approved):
@oplantalech Thanks so much for fixing this!

test/integration/integration_test_oncokb_import.sh
inodb: Like this one `"$PORTAL_SOURCE_DIR"/core/src/test/scripts/test_data/api_json_system_tests`

inodb: and if not - do we just delete that folder?

oplantalech: @inodb we need to address this with the datahub team. We use some genes `A1CF` (Entrez ID: `29974`), `A2M` (Entrez ID: `2`) and `A2ML1` (Entrez ID: `144568`) that are not in the new seed but they are in the public database. I don't know if it is a mistake or not, we need to discuss it.

Ideally then we should update indeed these test_data

PR #8825, opened 2021-08-06
Implement molecular data endpoint to use one db query per gene, for caching reasons

adamabeshouse:
Questions for backend team:
- Do you see any reason *not* to do this?
- As long as we're here, should we do this for the rest of the molecular data endpoints too? For mutation/structural variant endpoints too?

### DIFF
--- a/service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
@@ -189,8 +189,13 @@ public class MolecularDataServiceImpl implements MolecularDataService {
             samples = sampleService.fetchSamples(studyIds, sampleIds, "ID");
         }
 
-        List<GeneMolecularAlteration> molecularAlterations = molecularDataRepository
-            .getGeneMolecularAlterationsInMultipleMolecularProfiles(distinctMolecularProfileIds, entrezGeneIds, projection);
+        // query each entrezGeneId separately so they can be cached
+        List<GeneMolecularAlteration> molecularAlterations = entrezGeneIds.stream()
+            .flatMap(gene -> molecularDataRepository.getGeneMolecularAlterationsInMultipleMolecularProfiles(
+                    distinctMolecularProfileIds, Collections.singletonList(gene), projection
+                ).stream()
+            )
+        .collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/cbioportal/service/impl/MolecularDataServiceImpl.java
```diff
@@ -189,8 +189,15 @@ public Integer getNumberOfSamplesInMolecularProfile(String molecularProfileId) {
...
-            .getGeneMolecularAlterationsInMultipleMolecularProfiles(distinctMolecularProfileIds, entrezGeneIds, projection);
+        List<GeneMolecularAlteration> molecularAlterations = new ArrayList<>();
+        // query each entrezGeneId separately so they can be cached
+        for (Integer entrezGeneId: entrezGeneIds) {
+            List<Integer> singleGene = Arrays.asList(new Integer[]{ entrezGeneId });
+            molecularAlterations.addAll(
+                molecularDataRepository
+                    .getGeneMolecularAlterationsInMultipleMolecularProfiles(distinctMolecularProfileIds, singleGene, projection)
+            );
+        }
```
Luke-Sikina, line 200: This is fundementally fine, but there's a few small things we can improve:
`Arrays.asList(new Integer[]{ entrezGeneId });`-> `Arrays.asList(entrezGeneId);` Java has varargs. You are right that an array of the same type is equivalent, but this looks a bit cleaner.

`Arrays.asList` vs `Collections.singletonList`. Use the former for lists with multiple elements. Use the latter for singleton lists. 

Iterating using a for loop vs streaming. Java streams are pretty nice. They're lazily evaluated, so you can split things into easy to read sections without taking a performance hit. Generally, if I can solve something in a stream without any side effects, I do. This is just a flatmap and a collect.

Putting all those things together, here's what I would recommend:
```suggestion
        // query each entrezGeneId separately so they can be cached
        List<GeneMolecularAlteration> molecularAlterations = entrezGeneIds.stream()
            .flatMap(gene -> molecularDataRepository.getGeneMolecularAlterationsInMultipleMolecularProfiles(
                    distinctMolecularProfileIds, Collections.singletonList(gene), projection
                ).stream()
            )
            .collect(Collectors.toList());
```

Luke-Sikina (approved):
Looks functionally correct. Just some code style comments.

Luke-Sikina (approved):

PR #8827, opened 2021-08-09
fix 8823 foreign key with multiple names

### LINKED ISSUE #8823: sample table foreign key with multiple names causes migration issue
In sample table, Foreign key sample_ibfk_2 can sometimes be switched with sample_ibfk_1. This needs to be resolved during migration of database.

jagnathan:
foreign key with multiple names in sample table;  using the script to check for either name with reference to table and remove the one that exists

Fix #8823

- use the script to check for either name with reference to table and remove the one that exists
- alter migration.sql to remove the drop foreign key constraint for sample table


db-scripts/src/main/resources/migration.sql
```diff
@@ -944,8 +944,9 @@ ALTER TABLE `cancer_study` DROP COLUMN `SHORT_NAME`;
 -- 2.12.9 for all changes
 
 -- WARNING: this will drop column TYPE_OF_CANCER_ID from table sample
-ALTER TABLE `sample`
-DROP FOREIGN KEY `sample_ibfk_2` ;
+-- the foreign key can exist as sample_ibfk_1 or sample_ibfk_2
+-- so removing the fk in check_and_remove_foreign_keys_sample_ibfk
+-- once the fk is removed, column can be dropped
```
inodb, line 949:
```suggestion
-- the foreign key on TYPE_OF_CANCER_ID in the sample table is removed with the migrate_db.py script. This is because the name can be either sample_ibfk_1  or sample_ibfk_2
```

core/src/main/scripts/migrate_db.py
```diff
@@ -237,6 +238,48 @@ def check_and_remove_invalid_foreign_keys(cursor):
                 print('Invalid foreign key has been deleted.', file=OUTPUT_FILE)
     except MySQLdb.Error as msg:
         print(msg, file=ERROR_FILE)
+#since the foreign key can exist as sample_ibfk_1 or sample_ibfk_2
+#using the script to check for either name with reference to table and remove the one that exists
+def check_and_remove_foreign_keys_sample_ibfk(cursor):
```
inodb, line 243:
```suggestion
def check_and_remove_type_of_cancer_id_foreign_key(cursor):
"""The TYPE_OF_CANCER_ID foreign key in the sample table can be either sample_ibfk_1 or sample_ibfk_2. Figure out which one it is and remove it"""
```

inodb (commented):
Thanks for fixing this @jagnathan ! This looks good to me - just some minor comments on wording, but logic looks good

core/src/main/scripts/migrate_db.py
```diff
@@ -377,6 +420,8 @@ def main():
         if is_version_larger(MULTI_REFERENCE_GENOME_SUPPORT_MIGRATION_STEP, db_version):
             #retrieve reference genomes from database
             check_reference_genome(portal_properties, cursor, parser.force)
+        if not is_version_larger(SAMPLE_FK_MIGRATION_STEP, db_version):
+            check_and_remove_foreign_keys_sample_ibfk(cursor)
```
inodb, line 424:
```suggestion
            check_and_remove_type_of_cancer_id_foreign_key(cursor)
```

core/src/main/scripts/migrate_db.py
inodb: in the future there might be other foreign keys in sample, so maybe good to be slightly more explicit

inodb: maybe even better: `check_and_remove_type_of_cancer_id_foreign_key_in_sample_table`

inodb:
@jagnathan Thanks for making the updates! Seems like something goes wrong with running the localdb tests. My guess is that maybe the migration is not working as expected?

inodb (approved):
Thanks for fixing this 🙏 !

PR #8842, opened 2021-08-14
Fix coexpression optimization multithreading

Rajat-Sirohi:
There was a problem with PR #8793 which mitigated its performance improvements. I only noticed this after merging, so this PR redresses that problem by fixing the broken multithreading.


inodb (approved):
Going to merge this b/c we are also rolling out another backend fix. The other backend changes to coexp are already in master, so if anything shows up we can always revert both

Feel free to add any post merge review comments tho

Issue #8844, opened 2021-08-16
Backend support for custom gene list


dippindots:
- [x] (session service) Update session type in session service repo https://github.com/cBioPortal/session-service/pull/39
- [ ] (cBioPortal backend) Update cBioPortal backend to use latest session service after first pr merged
- [ ] (cBioPortal backend) Create new endpoints in [SessionServiceController](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SessionServiceController.java), functions needed: `add`, `fetch`, `delete`
- [ ] (cBioPortal frontend) Update `sessionServiceAPI`

PR #8846, opened 2021-08-16
Update session service version

dippindots:
Merge this after https://github.com/cBioPortal/session-service/pull/39
Related to https://github.com/cbioportal/cbioportal/issues/8844

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -412,7 +412,7 @@
       <dependency>
         <groupId>com.github.cbioportal</groupId>
         <artifactId>session-service</artifactId>
-        <version>0.3.0</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -412,7 +412,7 @@
       <dependency>
         <groupId>com.github.cbioportal</groupId>
         <artifactId>session-service</artifactId>
-        <version>0.3.0</version>
```
dippindots, line 415: Any reason that we are not using latest version `0.4.0` at here?

kalletlak: @dippindots don't remember. could you try to update and test it?

PR #8847, opened 2021-08-16
RFC 59 - show studies for which a user is not authorized

dianab0:
Worked on Rfc 59 (https://docs.google.com/document/d/1NkNhBG1sk0ZvCOwk3XtiuRT9a_tKkFCgXC9I3XT-Ye4/edit#heading=h.6c2l76kgw1ta) to retrieve all studies (authorized and unauthorized), each having a boolean variable isAuthorized, when the skin.show_unauthorized_studies is true in portal.properties (default is false). 

This will be used in frontend to show also the unauthorized studies, but grayed out and with a lock icon as in image below.
Screenshot from 2021-08-17 12-24-03

Frontend PR is https://github.com/cBioPortal/cbioportal-frontend/pull/3932

### DIFF
--- a/docs/Customizing-your-instance-of-cBioPortal.md
+++ b/docs/Customizing-your-instance-of-cBioPortal.md
@@ -224,6 +224,18 @@ Below you can find the complete list of all the available skin properties.
 			<td>0.5</td>
 		</tr>
 		<tr>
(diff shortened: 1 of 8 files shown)


dianab0:
Another PR will be created for this work.

PR #8862, opened 2021-08-23
Create session-service-working.md

kalletlak:
Add documentation how to update session-service and integrating to cBioPortal

### DIFF
--- a/docs/session-service-working.md
+++ b/docs/session-service-working.md
@@ -8,7 +8,7 @@
 ###  2. cBioPortal Backend
 1. Update session-service dependency version in [pom.xml](https://github.com/cBioPortal/cbioportal/blob/master/pom.xml).
 2. Add/Update api's in [SessionServiceController.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SessionServiceController.java)  to support new session type.
-3. Sometime we might need to defined model for session. Check how session class is extended in VirtualStudy and how it is consumed in [SessionServiceController.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SessionServiceController.java).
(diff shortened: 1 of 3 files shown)


docs/session-service-working.md
```diff
@@ -0,0 +1,15 @@
...
+###  1. Steps to add new Session type through session-service
+
+1.  Define new type at [SessionType.java](https://github.com/cBioPortal/session-service/blob/master/src/main/java/org/cbioportal/session_service/domain/SessionType.java).
+2. Add necessary tests if required.
+3. Update [session-service](https://github.com/cBioPortal/session-service#valid-type) documentation.
+
+###  2. cBioPortal Backend
+1. Update session-service dependency version in [pom.xml](https://github.com/cBioPortal/cbioportal/blob/master/pom.xml).
+2. Add/Update api's in [SessionServiceController.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SessionServiceController.java)  to support new session type.
+3. Sometime we might need to defined model for session. Check how session class is extended in VirtualStudy and how it is consumed in [SessionServiceController.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SessionServiceController.java).
```
dippindots, line 11: We should also add link for VirtualStudy:
```suggestion
3. Sometime we might need to defined model for session. Check how session class is extended in [VirtualStudy](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/parameter/VirtualStudy.java) and how it is consumed in [SessionServiceController.java](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SessionServiceController.java).
```

dippindots (commented):
Looks good to me! Just had one minor suggestion.

Besides this document, could you also write some simples lines for local development? I am thinking these maybe useful.

- How to expose database when using docker
- How to use cruel for testing
- What properties do we need to set in `portal.properties`

dippindots (approved):

PR #8870, opened 2021-08-25
Add time to study import timestamp

pvannierop:
# Problem

The field `IMPORT_DATE` (type _datetime_) of the `cancer_study` table reflects the date, but not the time of study import. For example:

```
2021-08-25 00:00:00
```

This makes this field not suitable to detect study updates that are made on the same day.

# Solution

The DaoCancerStudy class places a timestamp in the `IMPORT_DATE` field, like so:

```
2021-08-25 07:42:29
```

# Remark
- Timestamp is resolved in the INSERT statement (so by the MySQL server) and no longer by the Java code. The rationale for this is that this way there is a single point of truth for the timestamp. This defends against time differences between different systems that perform the study import.


sheridancbio (commented):
👍

PR #8879, opened 2021-09-01
Impl. Spring-managed CacheMapUtil

pvannierop:
# Problem

At present cBioPortal has two separate cache implementations for study data. First is a Spring-managed cache used for `@Cacheable` annotations and for MyBatis persistence.  Second cache is _CacheMapUtil_ used for evaluation of user permissions. When new studies are loaded both caches need to be evicted/reinitialized. Cache eviction can be triggered via the `/api/cache` endpoint. The cache current cache structure is problematic in scenarios where caches should be evicted on a cBioPortal instance that consists of multiple containers behind a load balancer. Cache eviction is simple when Spring-managed cache is implemented where all containers use the same Redis instance; triggering of the `/api/cache` endpoint  in a single container will invalidate the cache for all container of the instance. The _CacheMapUtil_ cache, however,  can now only be reinitialized by calling the `/api/cache` endpoint in each individual container. This makes cache eviction more complicated than needed (e.g. requiring access to Kubernetes API for external services ).

# Solution

This PR allows to configure the _CacheMapUtil_ cache to be managed by Spring cache solution (e.g., Redis or EHCache) based on the new application property `cache.cache-map-utils.springmanaged`. As described in the problem statement, this solution might be helpful in certain architectures. It comes at the cost of additional delays in responses to data requests caused by additional calls to the Spring-managed cache.

# Tests

No tests are included for this feature.


core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -483,12 +484,19 @@ public static boolean wantIGVBAMLinking() {
...
-    // CHANGES TO THIS LIST MUST BE PROPAGATED TO 'CacheMapUtil'
+    
+	/** 
+     * @deprecated
+     * This function is replaced by the {@link PortalSecurityConfig#userAuthorizationEnabled()}
+     * method in the org.cbioportal.utils package.
+     * 
+     * When updating logic make sure to also update {@link PortalSecurityConfig#userAuthorizationEnabled()}.
+     */
+    @Deprecated
```
pvannierop, line 495: @n1zea144 Is it correct that the _org.mskcc.cbio.portal.util_ lib will be removed in the future? Does it make sense to mark it as @deprecated?  I started to create functions that have identical behavior in a new _utils_ module in package `org.cbioportal.utils`. Do you agree with migration of (some of) the GlobalProperties logic elsewhere?

n1zea144: The main goal is to decompose the core module into more transparent "top-level" modules and consolidate code when possible (like the model package in core with the top level model package used by the web service).  If you've created functions with identical behavior in the new utils module, I think its ok to deprecated these.  Will this PR or a follow-up replace calls to these methods with the new ones?

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -58,14 +59,9 @@
...
 
-    @Value("${authenticate:false}")
-    private String authenticate;
-    
-    private boolean usingAuth() {
-        return !authenticate.isEmpty()
-                && !authenticate.equals("false")
-                && !authenticate.contains("social_auth");
-    }
+    @Autowired
```
Luke-Sikina, line 62: Thank you for the cleanup

Luke-Sikina (approved):
Just a few tiny changes. Nice PR, and thank you for keeping these pull requests smaller-  they're so much easier to review!

Oh, and don't forget to squash your commits. I don't care if you have more than 1, but each commit should be a working, high quality version of the code base.

core/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
pvannierop: @n1zea144 I am not looking for a replacement of the entire GlobalProperties class. I merely would like to clarify to new developers that functionality that is present in the _org.mskcc.cbio.portal_ package should not be used in _org.cbioportal_ packages. Revisiting this topic I think it is better not to add the @Deprecased annotation since it is not actually replaced.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+    private SampleListRepository sampleListRepository;
+
+    @Autowired
+    private GenericAssayRepository genericAssayRepository;
+
+    private static final int REPOSITORY_RESULT_LIMIT = Integer.MAX_VALUE; // retrieve all entries (no limit to return size)
+    private static final int REPOSITORY_RESULT_OFFSET = 0; // retrieve all entries (do not skip any)
+    
+    public Map<String, MolecularProfile> buildMolecularProfileMap() {
+        HashMap<String, MolecularProfile> molecularProfileCache = new HashMap<String, MolecularProfile>();
```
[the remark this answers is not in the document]
pvannierop: Removed.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/CacheMapBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+        for (MolecularProfile mp : molecularProfileRepository.getAllMolecularProfiles(
+            "SUMMARY",
+            REPOSITORY_RESULT_LIMIT,
+            REPOSITORY_RESULT_OFFSET,
+            null,
+            "ASC")) {
+            molecularProfileCache.put(mp.getStableId(), mp);
+        }
```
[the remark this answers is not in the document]
pvannierop: Updated where possible.

service/src/test/java/org/cbioportal/service/impl/CacheServiceImplTest.java
```diff
@@ -25,7 +25,7 @@
     private CacheManager cacheManager;
 
     @Mock
-    private CacheMapUtil cacheMapUtil;
+    private StaticRefCacheMapUtil cacheMapUtil;
     private Cache mockCache;
 
     @Before
```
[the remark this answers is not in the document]
pvannierop: Updated

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/util/InactiveCacheMapUtil.java
```diff
@@ -0,0 +1,72 @@
...
+@Component
+// This implementation of the CacheMapUtils is instantiated on portals where all uses can access any study.
+@ConditionalOnExpression("! @portalSecurityConfig.userAuthorizationEnabled()")
+public class InactiveCacheMapUtil implements CacheMapUtil {
+
+    // Since user-permission evaluation is not needed when this bean is present, throw an error when it is accessed.
+    
+    @Override
+    public Map<String, MolecularProfile> getMolecularProfileMap() {
+        throw new RuntimeException("CacheMapUtils are contacted on a portal where studies are accessible to all users.");
```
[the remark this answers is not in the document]
pvannierop: Reworded.

web/src/main/java/org/cbioportal/web/SampleController.java
pvannierop: Sure ! :)

Luke-Sikina:
This is failing to complete the `Start cbioportal and other services` step for e2e tests. We need that to be fixed before this can merge.

(This isn't news to anyone, just documenting why this is stagnating)

PR #8881, opened 2021-09-01
Update fetchGenericAssayDataCounts endpoint

dippindots:
Update `/generic-assay-data-counts/fetch` endpoint to correct data count for single chart filtering operation

In the study view page, we fetch single chart for calling `/generic-assay-data-counts/fetch` again, and at this time, we do not want to get filtered data count from that endpoint. We want to get original data count without filtering, because we want to show unselected portion of the chart, this change it to update this logic.

Note: this endpoint is in development, there is no frontend usage right now.

### DIFF
--- a/web/src/main/java/org/cbioportal/web/StudyViewController.java
+++ b/web/src/main/java/org/cbioportal/web/StudyViewController.java
@@ -508,6 +508,12 @@ public class StudyViewController {
 
         List<GenericAssayDataFilter> gaFilters = interceptedGenericAssayDataCountFilter.getGenericAssayDataFilters();
         StudyViewFilter studyViewFilter = interceptedGenericAssayDataCountFilter.getStudyViewFilter();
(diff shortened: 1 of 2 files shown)


Luke-Sikina (approved):

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -508,6 +508,12 @@ private void init() {
 
         List<GenericAssayDataFilter> gaFilters = interceptedGenericAssayDataCountFilter.getGenericAssayDataFilters();
         StudyViewFilter studyViewFilter = interceptedGenericAssayDataCountFilter.getStudyViewFilter();
+        // when there is only one filter, it means study view is doing a single chart filter operation
```
alisman, line 511: why is this not also necessary when there are more than one filters?

PR #8888, opened 2021-09-10
Study-specific cache eviction

pvannierop:
# Problem
Calls to the `/api/cache/` endpoint clear all cached data of a running portal. When database changes relate to a specific study, cache eviction at present evicts more caches than needed.

# Solution
This PR implements a new `/api/cache/{studyId}` endpoint which limits the caches evicted when a study is imported/deleted/updated to those possibly affected by the database change.

This section was added to the Caching.md docs:

> #### How does study-specific cache eviction work?
> 
> ##### Structure of cache keys
> 
> The caching solutions integrated with cBioPortal (EHCache and Redis) store data as key-value pairs. Each key represents
> a method call signature that contains the Java class name, the method name and a serialized representation of all method
> arguments. For instance this is the key for a call to the CancerTypMyBatisRepository.getAllCancerTypes() method with
> arguments _[ SUMMARY, 10000000, 0, null, ASC]_:
> 
> ```
> CancerTypeMyBatisRepository_getAllCancerTypes_"SUMMARY"_10000000_0_null_"ASC"
> ```
> 
> Cached data that relates to a specific study can be recognized by the occurrence of the cancer study identifier anywhere
> in the method arguments. The study identifier can occur in the method arguments because it is passed as argument itself,
> like here for a study with identifier _study_es_0_:
> 
> ```
> SignificantCopyNumberRegionMyBatisRepository_getSignificantCopyNumberRegions_"study_es_0"_"SUMMARY"_null_null_null_null
> ```
> 
> Alternatively the study identifier is present as the prefix of referenced study entities. For example, this is the
> request for all molecular profiles:
> 
> ```
> MolecularProfileMyBatisRepository_getMolecularProfiles_["study_es_0_mrna_median_Zscores","study_es_0_log2CNA","study_es_0_fusion","study_es_0_treatment_ec50","study_es_0_mutational_signature","study_es_0_gsva_pvalues","study_es_0_mutations","study_es_0_methylation_hm27","study_es_0_gistic","study_es_0_generic_assay_patient_test","study_es_0_treatment_ic50","study_es_0_mrna","study_es_0_gsva_scores"]_"SUMMARY""
> ```
> 
> #### Cache eviction rules
> 
> When a study is added, deleted or updated, all caches are evicted where the respective key meets any of these
> requirements:
> 
> 1. The key contains the cancer study identifier of the study that is added, deleted or updated.
> 2. The key does not contain the cancer study identifier of any study present in the database.
> 
> The rationale behind rule 1. is that when a key references data for the affected study it points to potentially outdated
> data and its associated cache should be evicted. The rationale behind rule 2. is that any key that does not reference
> data for any study potentially points to data derived from all studies in the database, including the affected study,
> and its associated cache should be evicted. Because not every key without study identifiers necessarily points to study
> related data, this rule is overly broad. At the moment of this writing we were unable to implement reliable methods that
> would further specify such keys. This might be a start-off point for future optimizations.
> 
> 
> 
>


persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/EhCacheUtils.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public List<String> getKeys(String cacheName) throws IllegalArgumentException {
+        javax.cache.Cache<String, Object> cache = cacheManager.getCache(cacheName);
+        if (cache == null) {
+            throw new IllegalArgumentException("Cannot find cache with name '" + cacheName + "'");
+        }
+        List<String> keysInCache = new ArrayList<String>();
+        Iterator<javax.cache.Cache.Entry<String, Object>> iterator = cache.iterator();
+        while (iterator.hasNext()) {
+            javax.cache.Cache.Entry<String, Object> entry = iterator.next();
```
Luke-Sikina, line 36: You can make this a for each if you want
```suggestion
        for (javax.cache.Cache.Entry<String, Object> entry : cache) {
            keysInCache.add(entry.getKey());
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/EhCacheUtils.java
```diff
@@ -0,0 +1,49 @@
...
+        this.cacheManager = customEhcachingProvider.getCacheManager();
+    }
+    
+    @Override
+    public List<String> getKeys(String cacheName) throws IllegalArgumentException {
+        javax.cache.Cache<String, Object> cache = cacheManager.getCache(cacheName);
+        if (cache == null) {
+            throw new IllegalArgumentException("Cannot find cache with name '" + cacheName + "'");
+        }
+        List<String> keysInCache = new ArrayList<String>();
```
Luke-Sikina, line 33:
```suggestion
        List<String> keysInCache = new ArrayList<>();
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/EhCacheUtils.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return keysInCache;
+    }
+
+    @Override
+    public void evictByPattern(String cacheName, String pattern) {
+        javax.cache.Cache<String, Object> cache = cacheManager.getCache(cacheName);
+        this.getKeys(cacheName).stream()
+            .filter(key -> key.matches(pattern))
+            .forEach(key -> cache.remove(key));
```
Luke-Sikina, line 47:
```suggestion
            .forEach(cache::remove);
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomKeyGenerator.java
```diff
@@ -32,26 +32,31 @@
...
 import org.springframework.cache.interceptor.KeyGenerator;
 import org.springframework.util.DigestUtils;
 
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
 
 public class CustomKeyGenerator implements KeyGenerator {
     public static final String DELIMITER = "_";
```
Luke-Sikina, line 52: I just realized that we have two public constants named `DELIMITER`. That's my fault, but let's fix it. Can we rename this? Maybe `CACHE_KEY_PARAM_DELIMITER`?

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/RedisCacheUtils.java
```diff
@@ -0,0 +1,40 @@
...
+    
+    @Override
+    public List<String> getKeys(String cacheName) {
+        Cache cache = cacheManager.getCache(cacheName);
+        Object nativeCache = cache.getNativeCache();
+        if (nativeCache instanceof RedissonClient) {
+            return ((RedissonClient)nativeCache)
+                .getKeys()
+                .getKeysStream()
+                .filter(k -> k.startsWith(cache.getName()))
```
Luke-Sikina, line 28:
```suggestion
                .filter(k -> k.startsWith(cache.getName() + DELIMITER))
```

service/src/main/java/org/cbioportal/service/exception/CacheOperationException.java
```diff
@@ -0,0 +1,7 @@
+package org.cbioportal.service.exception;
+
+public class CacheOperationException extends Exception {
```
Luke-Sikina, line 3: Love properly named exceptions instead of just using RuntimeException everywhere. Nice job.

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
```diff
@@ -1,27 +1,40 @@
...
 import org.springframework.stereotype.Component;
 
+import java.util.ArrayList;
+import java.util.List;
+
 @Component
 public class CacheServiceImpl implements CacheService {
-
+    
+    private static Log LOG = LogFactory.getLog(CacheServiceImpl.class);
```
Luke-Sikina, line 19:
```suggestion
    private static final Log LOG = LogFactory.getLog(CacheServiceImpl.class);
```

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
```diff
@@ -31,4 +44,61 @@ public void clearCaches(boolean clearSpringManagedCache) {
...
+            LOG.error("Error while evicting cache." + e.getMessage());
+            throw new CacheOperationException("Error while evicting cache.", e);
+        }
+    }
+    
+    // Regex that selects keys that match id of deleted study
+    // or lacking any study id completely. For example:
+    // ^(?=.*study_id_1).*|^(?!.*study_id_1)(?!.*study_id_2)(?!.*study_id_3).*
+    // https://stackoverflow.com/a/8240998/11651683
+    private String buildRegex(String studyId, List<String> allStudyIds) {
```
Luke-Sikina, line 79:
```suggestion
    private String buildEvictionRegex(String studyId, List<String> allStudyIds) {
```

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
```diff
@@ -31,4 +44,61 @@ public void clearCaches(boolean clearSpringManagedCache) {
...
+        
+        // make sure imported studyId is in all studies.
+        List<String> all = new ArrayList<>();
+        all.add(studyId);
```
Luke-Sikina, line 83: Wait- why? If the regex is "contains `studyId`, or does not contain `any study Id`, haven't you already ensured that the string the regex is matching against does _not_ contain `studyId` by the time the matcher reaches the second part of the regex?

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
```diff
@@ -31,4 +44,61 @@ public void clearCaches(boolean clearSpringManagedCache) {
...
+        stringBuilder.append("|");
+        stringBuilder.append("^");
+        
+        all.forEach(id -> {
+            stringBuilder.append("(?!.*");
+            stringBuilder.append(id);
+            stringBuilder.append(")");
+        });
+        stringBuilder.append(".*");
+        return stringBuilder.toString();
```
Luke-Sikina, line 101: Not a huge fan of all this StringBuilder stuff - I don't think we need to make regex any harder to read. Here's what this looks like if you use `String.format`. You can take it or leave it.

```suggestion
        String allIdsRegex = all.stream()
            .map(id -> "(?!.*" + id + ")")
            .collect(Collectors.joining(""));
        
        return String.format("^(?=.*%s).*|^%s.*", studyId, allIdsRegex);
```

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
```diff
@@ -31,4 +44,61 @@ public void clearCaches(boolean clearSpringManagedCache) {
...
+    }
+
```
Luke-Sikina, line 103: If I'm right about not having to add `studyId` and you like the String.format option, I think this is what the method ends up being. Again, up to you here:

```java
    private String buildEvictionRegex(String studyId, List<String> allStudyIds) {
        String allIdsRegex = allStudyIds.stream()
            .map(id -> "(?!.*" + id + ")")
            .collect(Collectors.joining(""));
        
        return String.format("^(?=.*%s).*|^%s.*", studyId, allIdsRegex);
    }
```

Luke-Sikina: This might be clearer than format. Idk, Java string interpolation leaves something to be desired.

```java
return "^(?=.*" + studyId + ").*|^" + allIdsRegex + ".*";
```

web/src/main/java/org/cbioportal/web/CacheController.java
```diff
@@ -45,7 +55,38 @@
...
+            return new ResponseEntity<>("Cache endpoint is disabled for this instance.", HttpStatus.NOT_FOUND);
+        }
+        if ("not set".equals(requiredApiKey) || ! requiredApiKey.equals(providedApiKey)) {
+            return new ResponseEntity<>("", HttpStatus.UNAUTHORIZED);
+        }
+
+        List<String> allStudyIds = studyRepository.getAllStudies(null, "SUMMARY", null, null, null, null)
+            .stream()
+            .map(study -> study.getCancerStudyIdentifier())
+            .collect(Collectors.toList());
```
Luke-Sikina, line 84: How do you determine whether a service method receives a piece of information as an argument, or gets it itself by talking to another service? I think I generally choose the second option if I'm confident that the service method in question will always know how to request that information.

For example, in this case, I think `allStudyIds` will always be `studyRepository.getAllStudies`, so I'd do it within the CacheServiceImpl. Interested to hear how you make this sort of determination.

Luke-Sikina (approved):
Seems pretty good. Just a bunch of small questions and suggestions.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomKeyGenerator.java
```diff
@@ -78,9 +83,15 @@ private String exceptionlessWrite(Object toSerialize) {
...
-                // hash long keys
-                return DigestUtils.md5DigestAsHex(json.getBytes());
+            if (json.length() > PARAM_LENGTH_HASH_LIMIT) {
+                // To allow study-specific cache eviction, extract relevant
+                // study identifiers and add these to the cache keys.
+                String matchedStudyIds = studyRepository.getAllStudies(null, "SUMMARY", null, null, null, null)
+                    .stream()
+                    .map(study -> study.getCancerStudyIdentifier())
+                    .filter(studyId-> json.contains(studyId))
+                    .collect(Collectors.joining(DELIMITER));
```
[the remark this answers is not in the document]
pvannierop: Updated. I did not know that it was possible to do:

```
                    .filter(json::contains)
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomRedisCache.java
```diff
@@ -99,24 +98,31 @@ public ValueWrapper putIfAbsent(Object key, @Nullable Object value) {
...
+    public void evict(Object pattern) {
+        evictIfPresent(pattern);
     }
 
     @Override
-    public boolean evictIfPresent(Object key) {
-        // no op: Redis handles evictions
+    public boolean evictIfPresent(Object pattern) {
+        // Pattern is expected to be a regular expression
+        if (pattern instanceof String) {
```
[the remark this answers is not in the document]
pvannierop: Updated.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomRedisCache.java
```diff
@@ -99,24 +98,31 @@ public ValueWrapper putIfAbsent(Object key, @Nullable Object value) {
...
+            String[] keys = redissonClient.getKeys().getKeysStream()
+                .filter(key -> key.startsWith(name))
+                .filter(key -> key.matches((String) pattern))
+                .toArray(String[]::new);    
+            return redissonClient.getKeys().delete(keys) > 0;
+        }
```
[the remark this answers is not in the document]
pvannierop: Lol thank you. I do not see anything obvious that I would change there :)

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomRedisCache.java
```diff
@@ -99,24 +98,31 @@ public ValueWrapper putIfAbsent(Object key, @Nullable Object value) {
...
-    public void evict(Object key) {
-        // no op: Redis handles evictions
+    public void evict(Object pattern) {
+        evictIfPresent(pattern);
     }
 
     @Override
-    public boolean evictIfPresent(Object key) {
-        // no op: Redis handles evictions
+    public boolean evictIfPresent(Object pattern) {
```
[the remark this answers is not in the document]
pvannierop: Done.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/EhCacheUtils.java
pvannierop: Updated to:

```
        List<String> keysInCache = new ArrayList<>();
        cache.forEach(entry -> keysInCache.add(entry.getKey()));
        return keysInCache;
```

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/EhCacheUtils.java
pvannierop: Yes, updated.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/EhCacheUtils.java
pvannierop: Updated.

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomKeyGenerator.java
pvannierop: Done

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/RedisCacheUtils.java
pvannierop: Updated.

service/src/main/java/org/cbioportal/service/exception/CacheOperationException.java
pvannierop: Thanks :)

pvannierop: Btw I reworked the exception handling somewhat. While working on Spring Boot cBioPortal I came across class GlobalExceptionHandler that has the responsibility to handle problems. This now includes:

```
    @ExceptionHandler(CacheOperationException.class)
    public ResponseEntity<ErrorResponse> handleCacheOperationException(CacheNotFoundException ex) {
        ErrorResponse response = new ErrorResponse("Error evicting caches. Please try again or validate correct operation of your configured caching implementation.");
        return new ResponseEntity<>(response, HttpStatus.INTERNAL_SERVER_ERROR);
```

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
pvannierop: Added

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
pvannierop: Updated.

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
pvannierop: No, the rule is delete the key when 1) the key contains `studyId`, or 2) the key contains no study id at all (including not containing `studyId`). The part in the regex before the `|` character relates to 1) and the part after relates to 2).

```
^(?=.*study_id_1).*|^(?!.*study_id_1)(?!.*study_id_2)(?!.*study_id_3).*
```

`^(?=.*study_id_1).*` means: any key that contains _study_id_1_.

`^(?!.*study_id_1)(?!.*study_id_2)(?!.*study_id_3).*` means does not contain any study id.

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
pvannierop: Oooooh much nicer! I used StringBuilder since I remember building strings is more memory efficient this way. But your solution is much nicer. Thanks I will remember this.

service/src/main/java/org/cbioportal/service/impl/CacheServiceImpl.java
pvannierop: Yes, included.

web/src/main/java/org/cbioportal/web/CacheController.java
pvannierop: Ah good question. I do not have a clear rationale I guess. I like your proposal better I think. I will move the method.

Luke-Sikina (approved):

PR #8889, opened 2021-09-10
Correct cache-assignment in Repository calls

pvannierop:
# Problem
At several places @Cacheable annotations point to non-existing cache names.

# Solution
This PR will correct @Cacheable annotations to reference _cacheResolvers_ as used at all other placed in the code.


dippindots (approved):
Thanks for fixing this! LGTM.

PR #8900, opened 2021-09-15
Downgrade mongo driver to fix strange error when calling session service

alisman:
After upgrade, at least 2 users experienced an error when calling session service

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -413,7 +413,7 @@
       <dependency>
         <groupId>org.mongodb</groupId>
         <artifactId>mongo-java-driver</artifactId>
-        <version>3.12.10</version>
+        <version>3.12.9</version>
       </dependency>
       <dependency>
         <groupId>org.mskcc.cbio</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -413,7 +413,7 @@
       <dependency>
         <groupId>org.mongodb</groupId>
         <artifactId>mongo-java-driver</artifactId>
-        <version>3.12.10</version>
+        <version>3.12.9</version>
```
inodb, line 416: it seems like this driver version might be compatible with mongodb version 3.7 (https://github.com/cBioPortal/cbioportal-docker-compose/blob/a8759d8751ecc6ba172a67b06e1b40cd4ac5b8bf/docker-compose.yml#L49):  https://docs.mongodb.com/drivers/java/sync/current/compatibility/. For some reason mongodb version 3.7 is not specifically listed (maybe it's not LTS) or something

inodb (approved):

pom.xml
inodb: anyway it's just a patch version change so i guess it should be fine

PR #8905, opened 2021-09-16
Remote legacy js and css from login jsp

alisman:
Remove legacy js and css from login.jsp (404s were making page load slowly)


alisman:
@ino gonna test this on master to make sure login functionality still works

inodb (approved):
@alisman Neat! It fixes the slow load time

Note that login itself is still broken

PR #8914, opened 2021-09-21
clinical density plot api: support log scale for MUTATION_COUNT, and return correlations

adamabeshouse:


model/src/main/java/org/cbioportal/model/DensityPlotData.java
```diff
@@ -0,0 +1,39 @@
+package org.cbioportal.model;
+
+import javax.validation.constraints.NotNull;
+import java.io.Serializable;
+import java.math.BigDecimal;
+import java.util.List;
+
+public class DensityPlotData implements Serializable {
+    @NotNull
+    private List<DensityPlotBin> bins;
```
Luke-Sikina, line 10: In most places where we use `@NotNull`, the parent object is never directly instantiated. Instead, it either comes from the API, in which case Jackson makes it, or it comes from the database, in which case MyBatis makes it. This all makes it much easier for the `@NotNull` assertion to always be true.

In your case, you actually use the constructor directly, so you do have DensityPlatData objects that have null bins. I think the best way to address this is to assign an empty arraylist to bins when you declare it. I realize we don't do this elsewhere, but again, we don't instantiate most of our objects directly.

```suggestion
    private List<DensityPlotBin> bins = new ArrayList<>();
```

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -453,6 +489,13 @@ private void init() {
                 densityPlotBin.setMaxY(yValueBigDecimal);
             }
         }
+        
+        result.setPearsonCorr(
+            new PearsonsCorrelation().correlation(xValues, yValues)
+        ); 
+        result.setSpearmanCorr(
+            new SpearmansCorrelation().correlation(xValues, yValues)
+        );
```
Luke-Sikina, line 498: What do you have your column count set to? On the backend, we don't enforce this strictly, but I'd say use 120. This looks like 80.

```suggestion
        result.setPearsonCorr(new PearsonsCorrelation().correlation(xValues, yValues)); 
        result.setSpearmanCorr(new SpearmansCorrelation().correlation(xValues, yValues));
```

web/src/test/java/org/cbioportal/web/StudyViewControllerTest.java
```diff
@@ -583,69 +583,158 @@ public void fetchClinicalDataDensityPlot() throws Exception {
...
-            .andExpect(MockMvcResultMatchers.jsonPath("$[7].maxY").doesNotExist())
-            .andExpect(MockMvcResultMatchers.jsonPath("$[7].count").value(0))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].binX").value(0.6666666666666666))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].binY").value(272.0))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].minX").value(1.0))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].maxX").value(1.0))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].minY").value(400.0))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].maxY").value(400.0))
-            .andExpect(MockMvcResultMatchers.jsonPath("$[8].count").value(1));
+            .andExpect(MockMvcResultMatchers.jsonPath("$.bins[0].binX").value(0.0))
```
Luke-Sikina, line 586: quite the test.

Luke-Sikina (approved):
Tiny changes. LGTM

PR #8943, opened 2021-10-01
Add CLI arg to scrpits to override server detection

Luke-Sikina:
- The docker-compose setup doesn't give correct console output
because the tomcat jar is included in core
- To get proper setup, introduce an override that asserts
that this script is not running on a server

Before:
```bash
/cbioportal/core/src/main/scripts# ./importGenePanel.pl --data /data_gene_panel_RMG.txt 
Done.
Total time:  2017 ms
```

After:
```bash

/cbioportal/core/src/main/scripts# ./importGenePanel.pl --data /data_gene_panel_RMG.txt --notaserver
Reading data from:  /data_gene_panel_RMG.txt
Extracting genes

Warnings / Errors:
-------------------
0.  Gene panel RMG already exists in the database but is not being used. Overwriting old gene panel data.; 1x
Done.
Total time:  2115 ms
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -17,10 +17,22 @@
     <dependency>
       <groupId>org.mskcc.cbio</groupId>
       <artifactId>persistence-connections</artifactId>
+        <exclusions>
+          <exclusion>
+            <artifactId>catalina</artifactId>
+            <groupId>org.apache.catalina</groupId>
+          </exclusion>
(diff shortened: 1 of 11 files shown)


Luke-Sikina:
Closing. Better solution here:  https://github.com/cBioPortal/cbioportal/pull/8946

PR #8977, opened 2021-10-14
Add validation for Arm-level CNA data and update docs

### LINKED ISSUE #8926: Add validation rules for specific Generic Assay files
Main purpose of this is to validate Categorical Generic Assay data. When we modify multiple studies (e.g. tcga studies), we will need to change more than 30 studies in a row. And it's likely to miss out one or two studies, so we should have a validation mechanism when we import those data.

dippindots:
Fix #8926

- In `validateData.py`, update validation logic for `Generic Assay Categorical Data`, if we are validating predefined types (e.g. Arm-Level CNA data), then use predefined list of values to validate data files.
- Update `File-Format.md` with new predefined data type and links to example files. [Document preview](https://github.com/cBioPortal/cbioportal/blob/e07ee866b4af9cd184157954805bc12c9e2ea106/docs/File-Formats.md) 
image
image

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -4544,12 +4544,42 @@ class GenericAssayCategoricalValidator(GenericAssayWiseFileValidator):
 
     """ Validator for files containing generic assay categorical values.
     """
+
+    # Global validation for specific generic assay types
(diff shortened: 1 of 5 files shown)


n1zea144:
@dippindots, @rmadupuri maintains the validator for us.  I've added her in the list of reviewers.

PR #9026, opened 2021-11-02
Remove filter validation from generic-assay-meta endpoint

### LINKED ISSUE #9025: patient view does not work because it's calling generic assay meta endpoint with empty query
patient view does not work because it's calling generic assay meta endpoint with empty query

**Details about this issue:**
This issue is introduced from this pr: https://github.com/cBioPortal/cbioportal/pull/8991
In this pr, we removed access control for `/generic-assay-meta/fetch` endpoint, and this result in the query need to pass a validation which were not actually been used before: https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/parameter/GenericAssayMetaFilter.java#L9-L12.

**Why it's not an issue before?**
It's because we had access control for `/generic-assay-meta/fetch` endpoint previously, and request goes into access control layer directly and did not went through validation.

**Solution proposal:** 
This could be a frontend issue or backend issue, it depends on how we want to validate requests:
1. If we want to allow user/frontend send empty requests and get empty response, then this is a backend issue, and we can modify the validation rules for this endpoint. https://github.com/cBioPortal/cbioportal-frontend/pull/4019
2. If we don't want to allow user/frontend send empty requests, then this is a frontend issue, and we should all related frontend code to make sure no empty requests will be sent. https://github.com/cBioPortal/cbioportal/pull/9026

dippindots:
Fix #9025 

- Remove filter validation from `generic-assay-meta/fetch` endpoint to allow empty request

### DIFF
--- a/web/src/main/java/org/cbioportal/web/GenericAssayController.java
+++ b/web/src/main/java/org/cbioportal/web/GenericAssayController.java
@@ -8,6 +8,7 @@ import springfox.documentation.annotations.ApiIgnore;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


dippindots:
We selected and merged frontend approach, close this one.

PR #9015, opened 2021-10-29
Update frontend version (run more localdb tests)

onursumer:
Fix #9012 

# Checks
- [ ] Runs on heroku
- [ ] Has tests or has a separate issue that describes the types of test that should be created. If no test is included it should explicitly be mentioned in the PR why there is no test.
- [ ] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!

PR #9027, opened 2021-11-03
Fixes Structural Variant Genes Table disappearing issue

### LINKED ISSUE #9011: Structural Variant Genes Table does not show up for archer
https://cbioportal.mskcc.org/study/summary?id=mskarcher
Screen Shot 2021-10-28 at 11 48 20 AM

There is a SQL error:

```
### The error may exist in URL [jar:file:/target/tomcat.8888/webapps/expanded/WEB-INF/lib/persistence-mybatis-0-unknown-version-SNAPSHOT.jar!/org/cbioportal/persistence/mybatis/AlterationCountsMapper.xml]
### The error may involve defaultParameterMap
### The error occurred while setting parameters
### SQL: SELECT              ENTREZ_GENE_ID AS entrezGeneId,             HUGO_GENE_SYMBOL AS hugoGeneSymbol,             COUNT(*) AS totalCount,             COUNT(DISTINCT(CASE_ID)) AS numberOfAlteredCases         FROM         (                        ) as JoinedTable                    GROUP BY ENTREZ_GENE_ID, HUGO_GENE_SYMBOL;
### Cause: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') as JoinedTable

        GROUP BY ENTREZ_GENE_ID, HUGO_GENE_SYMBOL' at line 9
; bad SQL grammar []; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') as JoinedTable

        GROUP BY ENTREZ_GENE_ID, HUGO_GENE_SYMBOL' at line 9
```

ao508:
Fix #9011 

Addresses issue where structural variant genes table was not rendering for MSKARCHER.

For similar reasons the Oncoprint, Mutation Tab, and the Patient/Sample-view mutation tables were also not displaying fusion events for ARCHER.

## Before fix:
image
image

----
## After fix:

image
image

# Checks
- [ ] Runs on heroku

# Notify reviewers
@kalletlak  @n1zea144 @inodb 

### DIFF
--- a/model/src/main/java/org/cbioportal/model/MolecularProfile.java
+++ b/model/src/main/java/org/cbioportal/model/MolecularProfile.java
@@ -80,10 +80,11 @@ public class MolecularProfile implements Serializable {
     }
 
     public MolecularAlterationType getMolecularAlterationType() {
-        // Alteration type is always set to STRUCTURAL_VARIANT when importing fusion profile
+        // TODO: remove this logic once all the fusions are migrated to structural variants
+        // most fusion profiles are imported as SV data but this also handles the archer
+        // study case where the fusions are imported under a mutations profile instead
         // https://github.com/cBioPortal/cbioportal/blob/8704058562c386afeac3082e50f39c1097d47983/core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java#L93
         // But somehow there was no migration for existing data. To resolve it replace FUSION alteration type by STRUCTURAL_VARIANT.
-        // TODO: remove this logic once all the fusions are migrated to structural variants
         return (molecularAlterationType.equals(MolecularAlterationType.FUSION) ||
                 (molecularAlterationType.equals(MolecularAlterationType.MUTATION_EXTENDED)
                 && this.datatype.equals("FUSION")))
@@ -162,7 +163,7 @@ public class MolecularProfile implements Serializable {
     public void setGenericAssayType(String genericAssayType) {
         this.genericAssayType = genericAssayType;
     }
-    
+
     public Boolean getPatientLevel() {
         return patientLevel;
     }
--- a/model/src/main/java/org/cbioportal/model/MolecularProfile.java
+++ b/model/src/main/java/org/cbioportal/model/MolecularProfile.java
@@ -84,7 +84,9 @@ public class MolecularProfile implements Serializable {
         // https://github.com/cBioPortal/cbioportal/blob/8704058562c386afeac3082e50f39c1097d47983/core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java#L93
         // But somehow there was no migration for existing data. To resolve it replace FUSION alteration type by STRUCTURAL_VARIANT.
         // TODO: remove this logic once all the fusions are migrated to structural variants
-        return molecularAlterationType.equals(MolecularAlterationType.FUSION)
+        return (molecularAlterationType.equals(MolecularAlterationType.FUSION) ||
+                (molecularAlterationType.equals(MolecularAlterationType.MUTATION_EXTENDED)
+                && this.datatype.equals("FUSION")))
                 ? MolecularAlterationType.STRUCTURAL_VARIANT
                 : molecularAlterationType;
     }
@@ -120,7 +122,8 @@ public class MolecularProfile implements Serializable {
     public Boolean getShowProfileInAnalysisTab() {
         // TODO: remove this logic once the data is fixed (when all the fusions are migrated to structural variants)
         return showProfileInAnalysisTab
-                || (getMolecularAlterationType().equals(MolecularAlterationType.STRUCTURAL_VARIANT)
+                || ((molecularAlterationType.equals(MolecularAlterationType.STRUCTURAL_VARIANT)
+                        || molecularAlterationType.equals(MolecularAlterationType.MUTATION_EXTENDED))
                         && datatype.equals("FUSION"));
     }
 
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/AlterationMyBatisRepository.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/AlterationMyBatisRepository.java
@@ -62,7 +62,8 @@ public class AlterationMyBatisRepository implements AlterationRepository {
         // if fusions were imported as a "mutations" profile then replace STRUCTURAL_VARIANT in
         // groupedIdentifiersByProfileType map with MUTATION_EXTENDED
         for (MolecularProfile profile : molecularProfileRepository.getMolecularProfiles(molecularProfileIds, "SUMMARY")) {
-            if (profile.getStableId().endsWith("mutations") && profile.getDatatype().equals("FUSION")) {
+            if (profile.getStableId().endsWith("mutations") && profile.getDatatype().equals("FUSION") &&
+                    groupedIdentifiersByProfileType.get(MolecularAlterationType.STRUCTURAL_VARIANT) != null) {
                 groupedIdentifiersByProfileType.put(MolecularAlterationType.MUTATION_EXTENDED,
                         groupedIdentifiersByProfileType.get(MolecularAlterationType.STRUCTURAL_VARIANT));
                 groupedIdentifiersByProfileType.remove(MolecularAlterationType.STRUCTURAL_VARIANT);
--- a/service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
@@ -41,7 +41,7 @@ public class SampleServiceImpl implements SampleService {
     private CopyNumberSegmentRepository copyNumberSegmentRepository;
     @Autowired
     private MolecularProfileRepository molecularProfileRepository;
-    
+
     @Override
     public List<Sample> getAllSamples(String keyword, List<String> studyIds, String projection,
                                       Integer pageSize, Integer pageNumber, String sort, String direction) {
@@ -54,7 +54,7 @@ public class SampleServiceImpl implements SampleService {
     public BaseMeta getMetaSamples(String keyword, List<String> studyIds) {
         return sampleRepository.getMetaSamples(keyword, studyIds);
     }
-    
+
     @Override
     public List<Sample> getAllSamplesInStudy(String studyId, String projection, Integer pageSize, Integer pageNumber,
                                              String sortBy, String direction) throws StudyNotFoundException {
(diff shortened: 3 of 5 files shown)


ao508:
@dippindots @alisman Where can I find the logic encoded in the Query button the study view page?

I'm still running into trouble with the Oncoprint rendering the fusion events for archer samples, although I've worked out the issues for rendering the structural variant genes table on the study view page and the mutation tables on the patient/sample view pages. 

Understanding what the frontend is expecting or executing when clicking that Query button will probably help shed more light on what exactly the backend is failing to do specifically for the Oncoprint.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -174,31 +174,52 @@ public BaseMeta fetchMetaSamples(List<String> sampleListIds) {
...
+    }
 
+    private Boolean hasFusionsAsMutations(MolecularProfile p) {
+        return (p.getMolecularAlterationType().equals(MolecularProfile.MolecularAlterationType.MUTATION_EXTENDED) &&
+                p.getDatatype().equals("FUSION"));
+    }
+
+    private void processSamples(List<Sample> samples, String projection) {
         if (projection.equals("DETAILED")) {
+            // working in patient/sample view but not oncoprint
```
inodb, line 189: @ao508 is this still true?

ao508: nope, removing

model/src/main/java/org/cbioportal/model/MolecularProfile.java
```diff
@@ -84,7 +84,9 @@ public MolecularAlterationType getMolecularAlterationType() {
         // https://github.com/cBioPortal/cbioportal/blob/8704058562c386afeac3082e50f39c1097d47983/core/src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java#L93
         // But somehow there was no migration for existing data. To resolve it replace FUSION alteration type by STRUCTURAL_VARIANT.
```
dippindots, line 85: Maybe we can also add some comment at here to mention that "fusion data could also come from a mutation profile"

model/src/main/java/org/cbioportal/model/MolecularProfile.java
```diff
@@ -84,7 +84,9 @@ public MolecularAlterationType getMolecularAlterationType() {
...
         // TODO: remove this logic once all the fusions are migrated to structural variants
-        return molecularAlterationType.equals(MolecularAlterationType.FUSION)
+        return (molecularAlterationType.equals(MolecularAlterationType.FUSION) ||
+                (molecularAlterationType.equals(MolecularAlterationType.MUTATION_EXTENDED)
```
n1zea144, line 89: Since MSK piggy-backs fusions in mutation profile using FUSION as datatype.

model/src/main/java/org/cbioportal/model/MolecularProfile.java
```diff
@@ -120,7 +122,8 @@ public void setDescription(String description) {
     public Boolean getShowProfileInAnalysisTab() {
         // TODO: remove this logic once the data is fixed (when all the fusions are migrated to structural variants)
         return showProfileInAnalysisTab
-                || (getMolecularAlterationType().equals(MolecularAlterationType.STRUCTURAL_VARIANT)
+                || ((molecularAlterationType.equals(MolecularAlterationType.STRUCTURAL_VARIANT)
+                        || molecularAlterationType.equals(MolecularAlterationType.MUTATION_EXTENDED))
```
n1zea144, line 127: Are the parenthesis correct here?  Should && only apply to datatype FUSION and MUTATION_EXTENDED?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -174,31 +174,51 @@ public BaseMeta fetchMetaSamples(List<String> sampleListIds) {
...
                                           new HashSet<String>(sampleListRepository.getAllSampleIdsInSampleList(studyId + SEQUENCED)));
-                structuralVariantSampleIdsMap.put(studyId,
-                                       new HashSet<String>(sampleListRepository.getAllSampleIdsInSampleList(studyId + STRUCTURAL_VARIANT)));
+                // get sv samples from sequenced case list if fusions imported as mutations
+                Set<String> svSamples = new HashSet<>();
+                if (studiesProfiledWithFusionsAsMutations.contains(studyId)) {
+                    // use sample list that has already been fetched
+                    svSamples = new HashSet<>(sequencedSampleIdsMap.get(studyId));
+                } else {
+                    svSamples = new HashSet<String>(sampleListRepository.getAllSampleIdsInSampleList(studyId + STRUCTURAL_VARIANT));
```
dippindots, line 211: I guess we can also use `new HashSet<>` to replace `new HashSet<String>`

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/AlterationMyBatisRepository.java
```diff
@@ -62,7 +62,8 @@
         // if fusions were imported as a "mutations" profile then replace STRUCTURAL_VARIANT in
         // groupedIdentifiersByProfileType map with MUTATION_EXTENDED
         for (MolecularProfile profile : molecularProfileRepository.getMolecularProfiles(molecularProfileIds, "SUMMARY")) {
-            if (profile.getStableId().endsWith("mutations") && profile.getDatatype().equals("FUSION")) {
+            if (profile.getStableId().endsWith("mutations") && profile.getDatatype().equals("FUSION") &&
+                    groupedIdentifiersByProfileType.get(MolecularAlterationType.STRUCTURAL_VARIANT) != null) {
```
n1zea144, line 66: Keeps from clobbering value keyed by MUTATION_EXTENDED.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -174,31 +174,51 @@ public BaseMeta fetchMetaSamples(List<String> sampleListIds) {
         return sampleRepository.getSamplesByInternalIds(internalIds);
     }
 
-    private void processSamples(List<Sample> samples, String projection) {
+    private Boolean isProfiledWithSV(MolecularProfile p) {
+        return (p.getMolecularAlterationType().equals(MolecularProfile.MolecularAlterationType.STRUCTURAL_VARIANT) ||
+                hasFusionsAsMutations(p));
```
n1zea144, line 179: Curious why you added hasFusionsAsMutations in this clause.

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
```diff
@@ -174,31 +174,51 @@ public BaseMeta fetchMetaSamples(List<String> sampleListIds) {
...
+    }
 
+    private Boolean hasFusionsAsMutations(MolecularProfile p) {
```
n1zea144, line 182: We should add the usual:
TODO: remove this logic once all the fusions are migrated to structural variants

dippindots (commented):
@ao508 Thanks for fixing this! Looks good to me! Just two minor comments.

I also would like to confirm that the issue was caused by this reason:
We were not able to find related samples because those samples are come from `mutation` profile, but we cannot fetch those data by using this method
```
sampleListRepository.getAllSampleIdsInSampleList(studyId + STRUCTURAL_VARIANT)
```
we need to fetch samples from sequenced samples like this:
```
sequencedSampleIdsMap.get(studyId)
```

Is that correct?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
n1zea144, line 211: Is this hit when the profile alteration type is STRUCTURAL_VARIANT?

service/src/main/java/org/cbioportal/service/impl/StructuralVariantServiceImpl.java
```diff
@@ -141,23 +141,29 @@
             String molecularProfileId = molecularProfileIds.get(i);
             if (molecularProfileId.endsWith(molecularProfileUtil.STRUCTURAL_VARIANT_PROFILE_SUFFIX)) {
                 structuralVariantMolecularProfileIds.add(molecularProfileId);
-                structuralVariantSampleIds.add(sampleIds.get(i));
+                if (CollectionUtils.isNotEmpty(sampleIds))  {
```
n1zea144, line 144: Nice.

n1zea144 (changes requested):

model/src/main/java/org/cbioportal/model/MolecularProfile.java
ao508: the logic before was SV && FUSION and since our case is mutations && FUSION I decided to consolidate the logic a little

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
ao508: so that the sample status `isProfiledWithSV` can be set to true in the mskarcher cases

the lambda function was a little.. ugly with everything included right there so I moved it to a function but I can move it back here instead if that's preferred

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
ao508: yes

ao508:
> @ao508 Thanks for fixing this! Looks good to me! Just two minor comments.
> 
> I also would like to confirm that the issue was caused by this reason: We were not able to find related samples because those samples are come from `mutation` profile, but we cannot fetch those data by using this method
> 
> ```
> sampleListRepository.getAllSampleIdsInSampleList(studyId + STRUCTURAL_VARIANT)
> ```
> 
> we need to fetch samples from sequenced samples like this:
> 
> ```
> sequencedSampleIdsMap.get(studyId)
> ```
> 
> Is that correct?

Yep, that helped resolve the study view SV genes table rendering issue as well as the patient/sample page displaying those fusion events.

model/src/main/java/org/cbioportal/model/MolecularProfile.java
n1zea144: Oh, the datatype is FUSION for SV alteration type?  Is that always true?

service/src/main/java/org/cbioportal/service/impl/SampleServiceImpl.java
n1zea144: I just noticed you are calling both below, so didn't think it was necessary.

n1zea144 (approved):

inodb (commented):
Thanks for fixing this!

We are currently not running the full suite of tests for the localdb nor are we running the tests for the backend. It would be good to run these thru at least the full set of localdb tests before merging. This is the PR that re-introduces many of the localdb tests: https://github.com/cBioPortal/cbioportal/pull/9015. Let's wait until that is merged

dippindots:
Tests looks good, I am going to merge this into master.

PR #9046, opened 2021-11-15
Fixed issue where /api/cache/<study_id> returned Unauthorized

dianab0:
I was able to call the /api/cache endpoint, but not the /api/cache/<study_id>, as it was returning an Unauthorized error.

To fix this the following have been modified:
- **portal/src/main/webapp/WEB-INF/web.xml:** Removed <url-pattern>/health</url-pattern> and <url-pattern>/cache</url-pattern> as both were already covered by /api/* url pattern.

- **security/security-spring/src/main/resources/applicationContext-security.xml:** Changed url to api/cache/** instead of only /api/cache.


pvannierop (approved):
Great thanks for solving this.

dippindots (approved):
Looks good to me

PR #9047, opened 2021-11-15
Fix errors when importing data with empty Start_Position or End_Position

nr23730:

The current [file format description for mutation data](https://docs.cbioportal.org/5.1-data-loading/data-loading/file-formats#mutation-data) lists `Start_Position` and `End_Position` as optional attributes. 

If those columns are not present at all the import of data without start/end position works just fine. However, **if these columns are present in the header and are left empty** you will get an error message stating that `''` is  not an integer value. 
By looking at the source code of the validator I identified `NA` as an alternative option. The usage of `NA` however leads to an error message in the `checkAlleleMAFFormat` function where it tries to cast `NA` to a float. 

Therefore I propose to gor for an empty string for `Start_Position` and `End_Position` as this successfully passes the validator and is much more intuitive.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -1695,7 +1695,7 @@ class MutationsExtendedValidator(CustomDriverAnnotationValidator):
             tumor_seq_allele2 = data[self.cols.index('Tumor_Seq_Allele2')].strip()
         if set(necessary_columns_check10).issubset(self.cols):
             # Set positions to False if position are empty, so checks will not be performed
-            if data[self.cols.index('Start_Position')] == "" or data[self.cols.index('End_Position')] == "":
+            if data[self.cols.index('Start_Position')] == '' or data[self.cols.index('Start_Position')] == 'NA' or data[self.cols.index('End_Position')] == '' or data[self.cols.index('End_Position')] == 'NA':
                 positions = False
             else:
                 positions = True
@@ -2202,7 +2202,7 @@ class MutationsExtendedValidator(CustomDriverAnnotationValidator):
 
     def checkStartPosition(self, value):
         """Check that the Start_Position value is an integer."""
-        if value == 'NA':
+        if value == '' or value == 'NA':
             self.logger.warning(
                 'The start position of this variant is not '
                     'defined. The chromosome plot in the patient view '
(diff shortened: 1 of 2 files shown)


inodb (commented):
This seems fine to me but prolly good to have @oplantalech take another look

One thing to note is that a lot of functionality doesn't work when start/end positions are missing (e.g. changing transcripts  on the mutations tab as well as several of the annotations provided thru Genome Nexus)

core/src/main/scripts/importer/validateData.py
```diff
@@ -2199,7 +2199,7 @@ def checkSwissProt(self, value):
 
     def checkStartPosition(self, value):
         """Check that the Start_Position value is an integer."""
-        if value == 'NA':
```
ao508, line 2202: should this perhaps be checking if `NA` or empty string?

ao508 (commented):
It's been a while since I've looked at the validator but I am curious if the check we're doing should be for both `NA` or empty string because I can definitely see data being submitted with `NA` values for the positions or other fields with numeric values instead of empty strings.

nr23730:
Thank you for having a look at this so quickly! 
First: I'm well aware of the consequences of letting these fields empty. 

Second: I tried to import the data when setting `Start_Position` and `End_Position` to `NA` - it failed. This is because [L1699](https://github.com/cbioportal/cbioportal/blob/master/core/src/main/scripts/importer/validateData.py#L1699) & [L1700](https://github.com/cbioportal/cbioportal/blob/master/core/src/main/scripts/importer/validateData.py#L1700) try to convert `NA` String to a float type. So if you'd like to enable both `""` and `"NA"` we'll have to adjust those two lines too.

nr23730:
@ao508 @inodb 
I updated this to the most recent master branch and modified it so that now `NA` and empty values are allowed.

oplantalech:
Sorry for replying to this so late. It is a very good catch and I think we need to solve these reported issues.

I just have one question: the error is still there if the value is empty but contains spaces, I guess a strip is missing somewhere?

oplantalech:
@nr23730 I agree with allowing `NA` or leaving empty spaces, but what I mean is that we should either accept whitespaces as a valid value or catch the error so that the validator can finish the process. Right now, with whitespaces the cryptic error "`'' is not an integer value`" is thrown and the script does not finalize with `INFO: -: Validation complete` (and then either `Validation of study succeeded.` or `Validation of study failed.` depending on what we decide to do).

oplantalech:
Closing this PR since it is included in #9799.

PR #9089, opened 2021-12-02
Introduced threads around calculateStatic/DynamicDataBin computation

n1zea144:
In an effort to improve backend performance via call to endpoint /clinical-data-bin-counts/fetch, threading was introduced to parallelize calls to ClinicalDataBinUtil.calculateStaticDataBins() which through profiling appeared to be one bottleneck in backend servicing of these requests. Threading mechanism was patterned matched against implementation used to improve co-expression computation.

The optimization was tested against the MSK IMPACT Study (~75k samples) using a laptop running the codebase connecting to a remote MySQL via vpn and deploying the war with a cold cache.

An example call to ClinicalDataBinUtil.calculateStaticDataBins before optimization:

image

An example call to ClinicalDataBinUtil.calculateStaticDataBins after optimization:

image

### DIFF
--- a/web/src/main/java/org/cbioportal/web/util/ClinicalDataBinAsyncMethods.java
+++ b/web/src/main/java/org/cbioportal/web/util/ClinicalDataBinAsyncMethods.java
@@ -1,0 +1,58 @@
+package org.cbioportal.web.util;
+
+import org.cbioportal.model.ClinicalData;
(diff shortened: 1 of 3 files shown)


n1zea144:
Oops, need to fix some unit tests.

n1zea144:
Oops, need to fix a failing unit test.

PR #9092, opened 2021-12-06
Fix validation of resource data for patient

inodb:
For some reason this gives an error when using a single patient with
associated resource data. It is e.g. checking whether a particular
patient has samples, but in the case where a resource is attached to a
patient directly it does not matter it the patient has samples at all.
The variable `RESOURCE_PATIENTS_WITH_SAMPLES` is not defined. Not
entirely sure how to best check for that case, so I just removed the
validation altogether

Btw: sorry for whitespace changes but you can ignore them with:
Screen Shot 2021-12-06 at 5 12 52 PM

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -727,7 +727,7 @@ class Validator(object):
                        'cause': sample_id})
             return False
         return True
-    
+
     def checkPatientId(self, patient_id, column_number):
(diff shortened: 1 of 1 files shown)


dippindots (approved):
Looks good to me, the `RESOURCE_PATIENTS_WITH_SAMPLES` is not necessary since we don't need related sample for patient resource data.

PR #9102, opened 2021-12-09
fix bug in importer and revisit ref genome in validator

SRodenburg:
The validator and importer Java backend code are rather confusing on what syntax to use to indicate an alternative genome build (e.g. GRCh38/hg38 or GRCm38/mm10 for  mouse). There is a mixed and ambiguous usage of the terms "genome build" and "reference genome" throughout the code.

- Validator now checks that _meta_study.txt_ in (optional) field `reference_genome` contains one of `hg19`, `hg38`, `mm10` and **else** uses the values specified in `portal.properties`.
- Validator now checks for MAF files that the column `NCBI_Build` contains a value that is consistent with the specified `reference_genome` value, thus `GRCh37` for `hg19` studies etc.
- some code cleaning to make this logic better comprehensible.

Already some time ago I looked into this but we never sorted this out: https://github.com/cBioPortal/cbioportal-frontend/pull/3073

There is also an accommodating PR from @pvannierop  in frontend. (https://github.com/cBioPortal/cbioportal-frontend/pull/4084)

This partially fixes this issue as well: https://github.com/cBioPortal/icebox/issues/115


SRodenburg:
@inodb happy new year! Can we merge this one?

inodb:
@SRodenburg happy new year! Can this be merged independently of the other PRs mentioned in the description (e.g. the frontend)

PR #9109, opened 2021-12-10
Turn off aspectj logging by default to improve request processing time.

n1zea144:
Improves response time /clinical-data-bin-counts by removing "cross-cutting" calls to logging for every entry/exit to methods in service-impl, web, and mybatis. Also, brings clarity to business logic that takes most processing time.

Before:
image

After:
image


dippindots (approved):
LGTM!

PR #9116, opened 2021-12-13
Update log4j dependencies

### LINKED ISSUE #9114: Upgrade to Log4j v2.16.0
cBioPortal's log4j is not affected by `CVE-2021-44228`, since it's using an old version of log4j

image (41)

However, as mentioned [here](https://www.lunasec.io/docs/blog/log4j-zero-day/#affected-apache-log4j-versions) this version is open to other exploits. We should upgrade to ~v2.15.0 therefore~ v2.16.0 (see: https://github.com/advisories/GHSA-7rjr-3q55-vv33)

(Thanks for digging @pvannierop !)

Luke-Sikina:
Fix #9114
- Update log4j dependencies. We don't have any RCE issues, they're just super out of date.

### TODO:
Please verify that the MSK / public log4j configs still work with this update. It worked locally for me, but you probably have different configs than me.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -196,8 +196,8 @@
       <artifactId>hrrs-servlet-filter-base64</artifactId>
     </dependency>
     <dependency>
-      <groupId>log4j</groupId>
-      <artifactId>apache-log4j-extras</artifactId>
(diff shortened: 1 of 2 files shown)


inodb (approved):
LGTM - thanks!

PR #9119, opened 2021-12-14
Log missing molecular profile IDs instead of NPEing

Luke-Sikina:
- AlterationRepository was NPEing because it was trying to group by null
if a molecular profile internal identifier was mismatched between cache and
database
- This changes that behavior to instead warn, and still try and complete the
groupby with any profile ids that are present

### DIFF
--- a/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/AlterationMyBatisRepository.java
+++ b/persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/AlterationMyBatisRepository.java
@@ -12,6 +12,8 @@ import org.cbioportal.model.MolecularProfile.MolecularAlterationType;
 import org.cbioportal.model.util.Select;
 import org.cbioportal.persistence.AlterationRepository;
 import org.cbioportal.persistence.MolecularProfileRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Repository;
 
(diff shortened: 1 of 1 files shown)


persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/AlterationMyBatisRepository.java
```diff
@@ -56,7 +59,15 @@
...
-            .collect(Collectors.groupingBy(e -> profileTypeByProfileId.getOrDefault(e.getMolecularProfileId(), null)));
+            .peek(e -> {
+                // Collectors.groupingBy does not allow null keys (it NPEs). The keys for this groupBy should never be null,
+                // but if they are, we filter them out and log a warning about them, rather than letting this whole function break.
+                if (!profileTypeByProfileId.containsKey(e.getMolecularProfileId())) {
+                    LOG.warn("Could not find internal molecular profile id {} - you probably need to clear your cache or restart your server.", e.getMolecularProfileId());
+                }
+            })
+            .filter(e -> profileTypeByProfileId.containsKey(e.getMolecularProfileId()))
+            .collect(Collectors.groupingBy(e -> profileTypeByProfileId.get(e.getMolecularProfileId())));
```
Luke-Sikina, line 70:
```suggestion
            .collect(Collectors.groupingBy(e -> profileTypeByProfileId.get(e.getMolecularProfileId())));
```

@inodb  we could just do this instead. This will still NPE, but at least we aren't explicitly trying to group by null

inodb: I like the explicit message that indicates what to do, but maybe exception is better? Just because I guess the numbers would be wrong and people might draw wrong conclusions

Issue #8710, opened 2021-06-22
Settings for multiple vs single OncoKB icon in annotation column 


leexgh:
We want to have a setting for choosing between single or multiple oncokb icons:

- [ ] at the app level we want to control the default column visibility maybe something like `show.default_annotation_columns="oncokb_drug,oncokb_diagnostics,etc.."`. That way we can enable the mutiple oncokb icons on the MSK portal to have it be consistent with the clinical PDF report that shows the same icons. On the public portal we can show a single item
- [x] a user should be able to choose themselves what columns to show (e.g. only one oncokb icon or more)
- [x] ideally, a user could persist their choice between sessions (either thru localStorage or session-service) by clicking on some button that indicates "make current column visibility my default" and e.g. "reset to default column visibility". Maybe we can use a similar UX flow for study view (where it also might be nice to opt in to persist the current chart selection rather than always saving it as we do now)

Note: there might be some overlap with other tickets here in terms of saving user settings, so make sure to avoid duplicate work

tmazor:
@jjgao what's the timeline for going back to a single icon for oncokb? Does it require this settings feature to be developed first, or can we go back to a single icon and then add these setting later?

inodb:
Let's leave the first item here undone until someone requests this

PR #9145, opened 2021-12-17
Add config param to customize single vs multi oncokb icons

onursumer:
Related to #8710

# Checks
- [ ] Runs on heroku

### DIFF
--- a/portal/src/main/webapp/config_service.jsp
+++ b/portal/src/main/webapp/config_service.jsp
@@ -45,6 +45,7 @@
             "oncoprint.hide_vus.default",
             "mycancergenome.show",
             "oncokb.public_api.url",
+            "oncokb.merge_icons_by_default",
             "digitalslidearchive.iframe.url",
             "digitalslidearchive.meta.url",
             "mdacc.heatmap.meta.url",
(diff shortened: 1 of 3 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -191,6 +191,8 @@ oncokb.public_api.url=https://public.api.oncokb.org/api/v1
 # Your OncoKB Token which an be found under https://www.oncokb.org/account/settings
 oncokb.token=
 
+# Enable merging of OncoKB icons
+oncokb.merge_icons=true
```
inodb, line 195: i think it's better to comment out this one. The default of `true` should be done in code rather than in the properties file. Issue is that when someone upgrades their cbioportal version they wouldn't have this property set in their existing properties file

portal/src/main/webapp/config_service.jsp
```diff
@@ -45,6 +45,7 @@
             "oncoprint.hide_vus.default",
             "mycancergenome.show",
             "oncokb.public_api.url",
+            "oncokb.merge_icons",
```
inodb, line 48: maybe `merge_icons_by_default`? If we ever would want to give the option to not split the icons at all thru config then this is more explicit

inodb (approved):
Thanks!

PR #9149, opened 2021-12-20
Throws exceptions for missing sample list id

dippindots:
Fix https://github.com/cbioportal/cbioportal/issues/9138

**Issue Description:**
Sometimes, Results view page will try to get samples from the sample list id, but an outdated query can have a sample list id that does not exist anymore.
In this case, we want to throw a 404 exception because we cannot find that sample list id in our database.

**Describe changes proposed in this pull request:**
- Update MolecularProfileMapper.xml (studyIds can be an empty list, this update will fix this issue)
- Throw SampleListNotFoundException for the case when we have a missing sample list id in `/samples/fetch` endpoint
- Renaming `fetchSamples` method to `fetchSampleBySampleListIds`, and have it iteratively call `fetchSampleBySampleListId` method that is cached.

### DIFF
--- a/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularProfileMapper.xml
+++ b/persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularProfileMapper.xml
@@ -39,9 +39,14 @@
         INNER JOIN reference_genome ON cancer_study.REFERENCE_GENOME_ID = reference_genome.REFERENCE_GENOME_ID
         <if test="studyIds != null">
             WHERE cancer_study.CANCER_STUDY_IDENTIFIER IN
-            <foreach item="item" collection="studyIds" open="(" separator="," close=")">
-                #{item}
-            </foreach>
+            <if test="studyIds.isEmpty()">
+                (NULL)
+            </if>
+            <if test="!studyIds.isEmpty()">
+                <foreach item="item" collection="studyIds" open="(" separator="," close=")">
+                    #{item}
+                </foreach>
+            </if>
         </if>
         <if test="sortBy != null and projection != 'ID'">
             ORDER BY ${sortBy} ${direction}
--- a/web/src/main/java/org/cbioportal/web/SampleController.java
+++ b/web/src/main/java/org/cbioportal/web/SampleController.java
@@ -4,10 +4,13 @@ import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import org.cbioportal.model.CancerStudy;
+import org.cbioportal.model.SampleList;
 import org.cbioportal.model.meta.BaseMeta;
 import org.cbioportal.model.Sample;
+import org.cbioportal.service.SampleListService;
 import org.cbioportal.service.StudyService;
 import org.cbioportal.service.exception.PatientNotFoundException;
+import org.cbioportal.service.exception.SampleListNotFoundException;
 import org.cbioportal.service.exception.SampleNotFoundException;
 import org.cbioportal.service.exception.StudyNotFoundException;
 import org.cbioportal.service.SampleService;
@@ -55,6 +58,9 @@ public class SampleController {
     
     @Autowired
     private SampleService sampleService;
+
+    @Autowired
+    private SampleListService sampleListService;
     
     @Autowired
     private StudyService studyService;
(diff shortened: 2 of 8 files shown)


persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularProfileMapper.xml
```diff
@@ -39,9 +39,14 @@
         INNER JOIN reference_genome ON cancer_study.REFERENCE_GENOME_ID = reference_genome.REFERENCE_GENOME_ID
         <if test="studyIds != null">
             WHERE cancer_study.CANCER_STUDY_IDENTIFIER IN
-            <foreach item="item" collection="studyIds" open="(" separator="," close=")">
-                #{item}
-            </foreach>
+            <if test="studyIds.isEmpty()">
```
Luke-Sikina, line 42: Good.

web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -261,7 +267,12 @@
     
     public List<Sample> fetchSamplesInner(
         List<String> sampleListIds, String projection
-    ) {
+    ) throws SampleListNotFoundException {
+        for (String sampleListId : sampleListIds) {
```
Luke-Sikina, line 271: Hmm. Not a huge fan of iteratively calling service methods. I'm approving this, but can you just verify that this isn't a performance concern? Maybe just call it with ~ 100 valid sample list IDs.

Luke-Sikina (approved):
Approved, but lmk about the performance of that check.

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/MolecularProfileMapper.xml
dippindots: Thanks for the advice

web/src/main/java/org/cbioportal/web/SampleController.java
dippindots: @Luke-Sikina I don't think this will cause a performance issue, the reason is that `fetchSamplesInner` is an internally used function in `SampleController`, and the only use case is like [this](https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/SampleController.java#L246-L248):
```
for (String sampleListId : sampleListIds) {
    samples.addAll(fetchSamplesInner(Arrays.asList(sampleListId), projection.name()));
 }
```
So this `fetchSamplesInner` function will be called multiple times if there are more than one `sampleListIds`, but each time it will only accept a single sample list id.

However this is not very good since the function signature of `fetchSamplesInner` indicate that this function will accept a list of sample list ids. Maybe I can change `fetchSampleInner` to this?
```
public List<Sample> fetchSamplesInner(
    String sampleListId, String projection
) throws SampleListNotFoundException {
    // check sampleList
    // if sampleList not found, throw sampleListNotFound exception
    sampleListService.getSampleList(Arrays.asList(sampleListId));
    return sampleService.fetchSamples(sampleListIds, projection);
}
```

jagnathan:
hi @dippindots @Luke-Sikina, Is it possible that the loop for fetching the data for each samplelistID was implemented to help caching the data in the redis? Since the list of SampleListIDs can vary, so the cache will quickly fill up for each variation in the list. But storing data in cache for individual SampleListID, the cache data will be for each SampleListID  and thus limited in data size.

Luke-Sikina:
Wait, so we're always sending `fetchSamples` 1 element arrays so we can organize our cache better? That's going to be really hard to enforce. I think a better solution would be to leave `fetchSamples` uncached, and have it iteratively call a `fetchSample` method that is cached. That way you have the organized cache, but without having to reimplement this weird code every time you call `fetchSamples`.

@dippindots @jagnathan

Luke-Sikina:
Something like this:

```java
    @Cacheable(cacheResolver = "staticRepositoryCacheOneResolver", condition = "@cacheEnabledConfig.getEnabled()")
    List<Sample> fetchSamples(List<String> sampleListIds, String projection);

    List<Sample> fetchSampleBySampleListId(String sampleListIds, String projection);
    ```

Luke-Sikina (approved):

dippindots:
@Luke-Sikina I noticed that the `fetchSamplesBySampleListIds` method in `SampleMyBatisRepository.java` is not correct. The reason is: when there are overlapping sample ids associated with multiple sample list IDs, now we will keep all duplicated samples, but we shouldn't. I found this because I saw a test error here: https://github.com/cBioPortal/cbioportal/runs/4706712710?check_suite_focus=true.

Here is my fix based on my understanding:
https://github.com/cBioPortal/cbioportal/pull/9149/commits/21541f08d3cf8c69ffec42d2ccfdb3c90c3c7cd1

Luke-Sikina (approved):

dippindots:
All tests are passing except local e2e tests. But the current local e2e tests are expected to fail and it's same as the master branch, so I will merge this pr.

dippindots:
@alisman I remember we discussed the frontend should be able to handle exceptions like 404 and so on, I think maybe you already have some ideas about it, do you mind creating a ticket for that?

PR #9156, opened 2021-12-23
Fix ASCN namespace import

pvannierop:
# Problem
Namespace import of ASCN columns from MAF file was broken in previous commit because the namespace match is now case sensitive.

# Solution
Correct case for namespace name used to decide whether to import ASCN columns.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
@@ -73,7 +73,7 @@ public class ImportExtendedMutationData{
     private Set<String> filteredMutations = new HashSet<String>();
     private Set<String> namespaces = new HashSet<String>();
     private Pattern SEQUENCE_SAMPLES_REGEX = Pattern.compile("^.*sequenced_samples:(.*)$");
-    private final String ASCN_NAMESPACE = "ascn";
+    private final String ASCN_NAMESPACE = "ASCN";
 
     /**
      * construct an ImportExtendedMutationData.


onursumer (approved):
Would be nice to have some tests to prevent these types of regression

pvannierop:
@onursumer I agree. But I think this is out of scope for this PR. I was not involved in implementation of the ASCN feature. Perhaps we can make an issue? However, it is good to merge this PR asap since it is resulting in errors in frontend e2e-localdb tests.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -73,7 +73,7 @@
     private Set<String> filteredMutations = new HashSet<String>();
     private Set<String> namespaces = new HashSet<String>();
     private Pattern SEQUENCE_SAMPLES_REGEX = Pattern.compile("^.*sequenced_samples:(.*)$");
-    private final String ASCN_NAMESPACE = "ascn";
+    private final String ASCN_NAMESPACE = "ASCN";
```
alisman, line 76: @averyniceday can see any problem with this viz-a-viz legacy data?  i.e. it was working before with lowercasing, now it won't?

pvannierop: No, not really. The manual describes that ASCN should be represented with capitals in the meta_mutations_extended.txt file. The lowercase conversion happens in the backend and ASCN_NAMESPACE matched this converted value. In a previous PR I have removed this conversion to lowercase. This fix will correct the match to the original case present in the meta_mutations_extended.txt file.

TLDR Legacy data is only affected when using lowecase _ascn_ in the metafile. This is not as described in the documentation and I consider this as making use a unsupported feature.

PR #9170, opened 2022-01-03
Fixes divide by zero events on the study view page

n1zea144:
PR fixes scenarios in which off panel genes have events that are not capture by the gene panel.  Prior to PR, if an event took place in a gene that was not on the panel (corner cases in studies like MSK-IMPACT or GENIE), the number of profiled samples calculated would be WES/WGS samples, which could be 0...and cause a divide by zero error (see image).   Code now uses total profiled samples (all profiled cases) for genes that are not on panel (this was behavior prior to optimization PR).

image


inodb (approved):
Tested on https://master.cbioportal.org/study/summary?id=msk_impact_2017 and there it seems to be fixed

PR #9198, opened 2022-01-11
Additional binning functions for genetic data

pvannierop:
# Description
This PR will add three new binning functions for histograms in Study View. It adds:
- Quartiles
- Split around median
- Generate bins based on bin size and anchor value. Creates bins upward from and downward from the _anchor value_. The size of each bin is the _bins size_. Bins are generated to cover the entire range of the data. 

# Backwards compatibility
The PR is compatible with the current frontend version. After update of PR https://github.com/cBioPortal/cbioportal-frontend/pull/4102 the _DataBinFilter_ class should be updated to remove this backward compatibility.

# Tests
Tests are added for all new binning functionality.

### DIFF
--- a/web/src/main/java/org/cbioportal/web/parameter/DataBinFilter.java
+++ b/web/src/main/java/org/cbioportal/web/parameter/DataBinFilter.java
@@ -5,11 +5,28 @@ import java.math.BigDecimal;
 import java.util.List;
 
 import javax.validation.constraints.AssertTrue;
+import org.springframework.validation.annotation.Validated;
 
+@Validated
 public class DataBinFilter implements Serializable {
 
+    public enum BinMethod {
+        MEDIAN,
+        QUARTILE,
+        CUSTOM,
+        GENERATE;
+    }
+
     private Boolean disableLogScale = false;
-    private List<BigDecimal> customBins;
+    private List<BigDecimal> customBins;                    // needed for 'Custom bins' frontend option
+    
+    // FIXME: Code added for backwards compatibility.
+    // Replace by commented out line after merge of PR:
+    // https://github.com/cBioPortal/cbioportal-frontend/pull/4102
+    private BinMethod binMethod = BinMethod.CUSTOM;
+    //private BinMethod binMethod;                           // needed for 'Median split' and 'Quartile' frontend options
+    
+    private BinsGeneratorConfig binsGeneratorConfig;          // needed for 'Generate Bins' frontend option
     private BigDecimal start;
     private BigDecimal end;
 
@@ -43,6 +60,23 @@ public class DataBinFilter implements Serializable {
 
     public void setEnd(BigDecimal end) {
         this.end = end;
+    }
+
+    public BinMethod getBinMethod() {
+        return binMethod;
+    }
+
+    public void setBinMethod(BinMethod binMethod) {
+        this.binMethod = binMethod;
+    }
+
+    public BinsGeneratorConfig getBinsGeneratorConfig() {
+        return binsGeneratorConfig;
+    }
+
+    public void setBinsGeneratorConfig(
+        BinsGeneratorConfig binsGeneratorConfig) {
+        this.binsGeneratorConfig = binsGeneratorConfig;
     }
 
     // TODO: make this work
--- a/web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
+++ b/web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
@@ -419,7 +419,7 @@ public class DataBinHelper {
         // Assumes that elements are sorted in ascending order
         BigDecimal minValue = sortedNumericalValues.get(0);
         BigDecimal maxValue = sortedNumericalValues.get(sortedNumericalValues.size()-1);
-        Assert.isTrue(minValue.compareTo(maxValue) != 1, "minValue larger than maxValue. Input is not sorted in ascending order!");
+        Assert.isTrue(minValue.compareTo(maxValue) < 1, "minValue larger than maxValue. Input is not sorted in ascending order!");
         
         List<BigDecimal> bins = new ArrayList<>();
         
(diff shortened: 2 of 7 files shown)


web/src/main/java/org/cbioportal/web/parameter/DataBinFilter.java
```diff
@@ -5,11 +5,49 @@
...
 public class DataBinFilter implements Serializable {
 
+    public enum BinMethod {
+        MEDIAN,
+        QUARTILE,
+        CUSTOM,
+        GENERATE;
+    }
+
+    public static class GenerateBinsConfig implements Serializable {
```
onursumer, line 20: It doesn't feel right to me when a class name sounds more like a function name. Wouldn't it be better if we name this class `DataBinConfig` or `DataBinnerConfig`?. Also, should we move this class to a separate file together with `BinMethod`? I guess `BinMethod` can be defined inside `DataBinConfig` class instead of `DataBinFilter`.

pvannierop: @onursumer I have added the following changes:
- Renamed `GenerateBinsConfig` to `BinsGeneratorConfig`. The _Generate_ in the name distinguishes it from _Custom_ bins configuration. So your proposed name `DataBinnerConfig` does not sufficiently express that the backend will generate bins depending on the provided parameters.
- Move the `BinsGeneratorConfig` class to a dedicated file.

Moving BinMethod into the `BinsGeneratorConfig` class since the Generator represents on of the options represented by the BinMethod enum. I can move the BinMethod enum to a separate file if you want.

web/src/main/java/org/cbioportal/web/parameter/DataBinFilter.java
```diff
@@ -5,11 +5,28 @@
...
+import org.springframework.validation.annotation.Validated;
 
+@Validated
 public class DataBinFilter implements Serializable {
 
+    public enum BinMethod {
+        MEDIAN,
+        QUARTILE,
+        CUSTOM,
+        GENERATE;
```
onursumer, line 17: Should we name this method `ANCHORED` or something like that to make it more clear?

pvannierop: No, the _generate_ approach takes two params the _anchor_ and the _bin size_. So `ANCHORED` does not sufficiently describe it.

onursumer: For the `MEDIAN` and `QUARTILE` methods the bins are also generated, right? So we generate bins for all methods other than `CUSTOM`. At least that is how I see it. If you think `GENERATE` describes this method the best I am okay with that though.

pvannierop: These names reflect the UI elements in the frontend that control them. Would it not be a good idea to match these?

dippindots (commented):
Overall It looks very good to me, just have one question on how to calculate the median.

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -399,4 +405,45 @@ public Boolean areAllIntegers(Set<BigDecimal> uniqueValues) {
...
     }
+
+    public List<BigDecimal> generateBins(List<BigDecimal> sortedNumericalValues, BigDecimal binSize, BigDecimal anchorValue) {
+
+        Assert.notNull(sortedNumericalValues, "sortedNumerical values is null!");
+        Assert.notNull(binSize, "binSize values is null!");
+        Assert.notNull(anchorValue, "anchorValue values is null!");
+
+        BigDecimal minValue = min(sortedNumericalValues);
+        BigDecimal maxValue = max(sortedNumericalValues);
```
Luke-Sikina, line 416: If these values are sorted, aren't min and max just first and last?

pvannierop: Ah yes, I updated the logic, but added checks that the order is as we expect.

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -146,6 +147,11 @@ public BigDecimal calcQ1(List<BigDecimal> sortedValues) {
             sortedValues.get((int) Math.floor(sortedValues.size() / 4.0)) : null;
     }
 
+    public BigDecimal calcMedian(List<BigDecimal> sortedValues) {
+        return sortedValues.size() > 0 ?
+            sortedValues.get((int) Math.ceil((sortedValues.size() * (1.0 / 2.0)))) : null;
+    }
+
```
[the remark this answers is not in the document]
pvannierop: I have update the calculation to @Luke-Sikina 's suggestion.

Luke-Sikina (approved):

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -399,4 +405,47 @@ public Boolean areAllIntegers(Set<BigDecimal> uniqueValues) {
...
+        Assert.notNull(anchorValue, "anchorValue values is null!");
+        
+        if (sortedNumericalValues.isEmpty()) {
+            return null;
+        }
+
+        // Assumes that elements are sorted in ascending order
+        BigDecimal minValue = sortedNumericalValues.get(0);
+        BigDecimal maxValue = sortedNumericalValues.get(sortedNumericalValues.size()-1);
+        Assert.isTrue(minValue.compareTo(maxValue) != 1, "minValue larger than maxValue. Input is not sorted in ascending order!");
```
Luke-Sikina, line 422:
```suggestion
        Assert.isTrue(minValue.compareTo(maxValue) < 0, "minValue larger than maxValue. Input is not sorted in ascending order!");
```

That should make sonar happy

dippindots (approved):

PR #9203, opened 2022-01-11
Add endpoint for clinical x vs y violin plot data

adamabeshouse:


web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+        ClinicalViolinPlotData result = new ClinicalViolinPlotData();
+        result.setAxisStart(Double.POSITIVE_INFINITY);
+        result.setAxisEnd(Double.NEGATIVE_INFINITY);
+        result.setRows(new ArrayList<>());
+
+        List<String> studyIds = new ArrayList<>();
+        List<String> sampleIds = new ArrayList<>();
+        studyViewFilterUtil.extractStudyAndSampleIds(studyViewFilterApplier.apply(interceptedStudyViewFilter), studyIds, sampleIds);
+        //DensityPlotData result = new DensityPlotData();
+        //result.setBins(new ArrayList<>());
```
dippindots, line 612: we could clean up a little bit here

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+            );
+        }
+
+        List<ClinicalData> clinicalDataList = clinicalDataFetcher.fetchClinicalData(
+            studyIds, sampleIds, patientIds, studyIdsOfPatients, sampleAttributeIds, patientAttributeIds, null
+        );
+
+        List<ClinicalData> sampleClinicalDataList;
+        // put all clinical data into sample form
+        if (CollectionUtils.isNotEmpty(patientAttributeIds)) {
```
dippindots, line 647: If this line if the same as line 631, maybe we can put the contents into the same if clause?

dippindots: I see... There are some common stuff in the first `if` at line 631, maybe that's the reason not combining them together.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+            int valuesIndex = 0;
+            for (ClinicalData d: data) {
+                Double value = useLogScale ? StudyViewController.logScale(Double.parseDouble(d.getAttrValue())) : Double.parseDouble(d.getAttrValue());
+                values[valuesIndex] = value;
+                min = Math.min(value, min);
+                max = Math.max(value, max);
+                valuesIndex += 1;
+            }
+                
+            //
```
dippindots, line 722: We can remove this.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+        List<ClinicalViolinPlotRowData> rows = result.getRows();
+        nonOutliers.forEach((category, data)->{
+            ClinicalViolinPlotRowData row = new ClinicalViolinPlotRowData();
+            row.setCategory(category);
+            row.setNumSamples(data.size() + outliers.get(category).size());
+            row.setBoxData(boxData.get(category).limitWhiskers(result));
+            
+            List<ClinicalData> _individualPoints = new ArrayList<>();
+            
+            if (row.getNumSamples() <= 7) {
```
dippindots, line 800: Could we add some comments on what is this means?

dippindots (commented):
@adamabeshouse Overall it looks good to me! One possible improvement is that maybe we can move some computations into a util function because the controller function is around 300 lines (a little bit larger than it should be).

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+        }
+
+        // clinicalDataMap is a map sampleId->studyId->data
+        Map<String, Map<String, List<ClinicalData>>> clinicalDataMap = sampleClinicalDataList.stream().collect(Collectors.groupingBy(ClinicalData::getSampleId,
+            Collectors.groupingBy(ClinicalData::getStudyId)));
+        
+        // Group data by category
+        Map<String, List<ClinicalData>> groupedDetailedData = new HashMap<>();
+        boolean useLogScale = logScale && StudyViewController.isLogScalePossibleForAttribute(numericalAttributeId);
+        clinicalDataMap.forEach((k, v) -> v.forEach((m, n) -> {
```
Luke-Sikina, line 683: `m` and `n` don't seem like good variable names.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+            if (
+                // n.size() == 2 means we have clinical data for the sample for both of the queried attributes
+                n.size() == 2 &&
+                    // check if the second data is numerical
+                    NumberUtils.isCreatable(n.get(1).getAttrValue())) {
```
Luke-Sikina, line 688: I don't think we really need the second comment, which lets us put that first comment above the loop, and clean up the indentation.
```suggestion
            // n.size() == 2 means we have clinical data for the sample for both of the queried attributes
            if (n.size() == 2 && NumberUtils.isCreatable(n.get(1).getAttrValue())) {                    
```

web/src/main/java/org/cbioportal/web/StudyViewController.java
Luke-Sikina, line 722: Delete

Luke-Sikina (commented):
Lots of changes, but they're all very small. Seems like a reasonable Pr - nice job.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -571,6 +574,271 @@ private static double parseValueLinear(ClinicalData c) {
...
+        if (CollectionUtils.isNotEmpty(patientAttributeIds)) {
+            sampleClinicalDataList = new ArrayList<>();
+            for (ClinicalData d: clinicalDataList) {
+                if (d.getSampleId() == null) {
+                    // null sample id means its a patient data, 
+                    //  we need to distribute the value to samples
+                    List<Sample> samplesForPatient = patientToSamples.get(
+                        d.getPatientId()
+                    ).get(d.getStudyId());
+                    if (samplesForPatient != null) {
```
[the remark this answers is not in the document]
adamabeshouse: honestly I don't even know why this check is here. would samplesForPatient ever be null?

model/src/main/java/org/cbioportal/model/ClinicalViolinPlotBoxData.java
```diff
@@ -0,0 +1,55 @@
+package org.cbioportal.model;
+
+public class ClinicalViolinPlotBoxData {
+    private double whiskerLower;
```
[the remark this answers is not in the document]
adamabeshouse: what makes you prefer boxed primitives? some googling is telling me they're less efficient

web/src/main/java/org/cbioportal/web/StudyViewController.java
adamabeshouse: its' because I want to call the clinical data fetcher regardless

web/src/main/java/org/cbioportal/web/StudyViewController.java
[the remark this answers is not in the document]
Luke-Sikina: I don't know. If you don't think that's a valid case, you should probably 400 in the controller method if you don't have a patient.

model/src/main/java/org/cbioportal/model/ClinicalViolinPlotBoxData.java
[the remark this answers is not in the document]
Luke-Sikina: There are trade offs to both approaches. `double` doesn't extend Object, and can't be used in collection (so you can't do `List<double>`. It also can't be null, which is nice if you can guarantee that, but it can also cause NPEs if you're marshaling from something where that value is null.

There are some very minor performance differences between the two because of the boxing/unboxing logic. In general, I wouldn't worry about it unless you expect to be doing something several thousand times. 

For this case, if it isn't valid for `whiskerLower` to be null, I would make it `double`. If it is valid, I would make it `Double`

Luke-Sikina (approved):
Approved. Just address that last comment, and then I think this can merge.

model/src/main/java/org/cbioportal/model/ClinicalViolinPlotBoxData.java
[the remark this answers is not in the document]
adamabeshouse: okay, then I will leave it since it's not null

model/src/main/java/org/cbioportal/model/ClinicalViolinPlotIndividualPoint.java
```diff
@@ -0,0 +1,31 @@
+package org.cbioportal.model;
+
+public class ClinicalViolinPlotIndividualPoint {
+    private String sampleId;
+    private String studyId;
+    private double value;
```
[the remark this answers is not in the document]
adamabeshouse: based on the discussion above I will leave it since it's not nullable

dippindots:
@Luke-Sikina @adamabeshouse The core tests are failing, I can't find what's wrong, any clue?

Luke-Sikina:
@dippindots Yup. We just need to wire up a mock for the new service we created for the controller test suite. Right now the controller tests are failing because they spin up all the controllers, and wire all the beans they depend on, but there's no VoilinPlotService bean in the test ApplicationContext.

I have a fix for it. I'll go over it with Adam tomorrow.

PR #9231, opened 2022-01-18
Fix import of ASCN columns

pvannierop:
Was incompletely fixed in previous commit. This caused e2e-localdb tests to fail.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/maf/ValueTypeUtil.java
+++ b/core/src/main/java/org/mskcc/cbio/maf/ValueTypeUtil.java
@@ -4,7 +4,7 @@ public class ValueTypeUtil {
 
     private ValueTypeUtil() {
         throw new IllegalStateException("This is a utility class. Do not instantiate.");
-    } 
+    }
 
     public static boolean isInt(String value) {
         if (value == null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mskcc/cbio/portal/model/AlleleSpecificCopyNumber.java
```diff
@@ -32,9 +32,13 @@
 
 package org.mskcc.cbio.portal.model;
 
+import org.mskcc.cbio.maf.ValueTypeUtil;
```
onursumer, line 35: Do we need this? The static import statement below should suffice, right?

```Java
import static org.mskcc.cbio.maf.ValueTypeUtil.*;
```

core/src/main/java/org/mskcc/cbio/maf/ValueTypeUtil.java
```diff
@@ -1,5 +1,7 @@
 package org.mskcc.cbio.maf;
 
+import sun.jvm.hotspot.utilities.Assert;
```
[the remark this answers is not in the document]
pvannierop: Ah well spotted. I remove this.

core/src/main/java/org/mskcc/cbio/portal/model/AlleleSpecificCopyNumber.java
pvannierop: Import were updated.

onursumer (approved):

PR #9259, opened 2022-01-27
New oncoprint property to configure tracks that are shown by default (backend)

dianab0:
Added a new portal property named oncoprint.clinical_tracks.show_by_default that points to a JSON file (on the classpath) that configures tracks that should be visible by default. Example:

oncoprint.clinical_tracks.show_by_default =classpath:/oncoprint-default-tracks.json

The configuration JSON contains an array of objects with stable ids of clinical variables, and optionally their (nested) sort order and/or gap on option. Example:

[
      {
        "stableId": "CANCER_TYPE",
        "sortOrder": "ASC",
        "gapOn": true
      },
      {
        "stableId": "MUTATION_COUNT",
        "sortOrder": "DESC",
        "gapOn": false
      },
      {
        "stableId": "AGE",
        "sortOrder": null,
        "gapOn": null
      }
 ]

When the backend exposes this new property, the cbioportal-frontend should render the tracks when initializing. For example, on the configuration json above:

Oncoprint will show the CANCER_TYPE, MUTATION_COUNT and AGE tracks when first showing, and will sort the samples on oncoprint first on CANCER_TYPE (top track) in ascending order and subsequently on MUTATION_COUNT in descending order (middle track). The AGE track will be shown as the bottom track (not sorted).

Oncoprint will ignore any clinical variable that is configured in the configuration JSON that is not available in the study that is analyzed at that moment in Results View.


dippindots (commented):
@dianab0 It looks good to me, I'm not very familiar with reading JSON stored in classPath, we should get @pvannierop's opinion too.
A question is where should we put this file? `src/main/resources`?

pvannierop:
@dippindots I do not think this is needed. There are similar cases (e.g. custom gene groups on query page) that use a similar JSON file, but we do not provide an example for. I vote for not adding it. The docs are sufficient here to configure correctly.

dippindots (approved):

PR #9281, opened 2022-02-01
this commit should fail

Luke-Sikina:


web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -696,4 +696,20 @@ private static double parseValueLinear(ClinicalData c) {
 
         return new ResponseEntity<>(studyViewFilterApplier.getDataBins(dataBinMethod, interceptedGenericAssayDataBinCountFilter), HttpStatus.OK);
     }
+    private String getFirstName() {
```
Luke-Sikina, line 699: I pretty much copy pasted from their example: https://sonarcloud.io/organizations/cbioportal/rules?languages=java&open=java%3AS1698

PR #9282, opened 2022-02-02
Do the monkey!

onursumer:
No monkeys have been harmed in the making of this PR.

### DIFF
--- a/web/src/main/java/org/cbioportal/proxy/ProxyController.java
+++ b/web/src/main/java/org/cbioportal/proxy/ProxyController.java
@@ -1,28 +1,30 @@
 package org.cbioportal.proxy;
 
+import org.apache.commons.codec.binary.Base64;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -44,14 +46,100 @@ public String proxy(@RequestBody(required = false) String body, HttpMethod metho
...
+        return exchangeOncokbData(
+            body,
+            request.getPathInfo().replaceFirst("/oncokb", ""),
+            request.getQueryString(),
+            method,
+            getOncokbHeaders(request)
+        );
+    }
+    
+    @RequestMapping("/monkeykb/**")
```
onursumer, line 65: Tentative name for the new endpoint. We can make this one non human readable as well.

alisman (approved):

PR #9335, opened 2022-02-24
Only evict Redis cache if keys present

pvannierop:
# Problem
When evicting the Redis cache when no cache keys are present, the Redisson library throws an error.

# Solution
Only evict the cache when there are keys.


alisman (approved):

persistence/persistence-api/src/main/java/org/cbioportal/persistence/util/CustomRedisCache.java
```diff
@@ -110,7 +110,9 @@ public boolean evictIfPresent(Object pattern) {
                 .filter(key -> key.startsWith(name))
                 .filter(key -> key.matches((String) pattern))
                 .toArray(String[]::new);
-            return redissonClient.getKeys().delete(keys) > 0;
+            // Calling delete() with empty array causes an error in the Redisson client.
```
alisman, line 113: what is the error out of curiosity?

PR #9350, opened 2022-03-01
Implement caching on samples fetch endpoint

alisman:
Sample endpoint should be cached at service level and with granularity at sample list id.   Only when sample list id is passed.

Fixes https://github.com/cbioportal/cbioportal/issues/9351

### DIFF
--- a/web/src/main/java/org/cbioportal/web/SampleController.java
+++ b/web/src/main/java/org/cbioportal/web/SampleController.java
@@ -249,11 +249,22 @@ public class SampleController {
             if (interceptedSampleFilter.getSampleListIds() != null) {
                 List<String> sampleListIds = interceptedSampleFilter.getSampleListIds();
                 
+                samples = new ArrayList<Sample>();
+                
                 for (String sampleListId : sampleListIds) {
                     // check that all sample lists exist (this method throws an exception if one does not)
                     sampleListService.getSampleList(sampleListId);
                 }
-                samples = sampleService.fetchSamples(sampleListIds, projection.name());
+                
+                for (String sampleListId : sampleListIds) {
+                    // fetch by sampleId so that we get cache at level of id instead list of ids
+                    samples.addAll(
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/cbioportal/web/SampleController.java
```diff
@@ -249,11 +249,22 @@
...
+                samples = new ArrayList<Sample>();
+                
                 for (String sampleListId : sampleListIds) {
                     // check that all sample lists exist (this method throws an exception if one does not)
                     sampleListService.getSampleList(sampleListId);
                 }
-                samples = sampleService.fetchSamples(sampleListIds, projection.name());
+                
+                for (String sampleListId : sampleListIds) {
+                    // fetch by sampleId so that we get cache at level of id instead list of ids
```
onursumer, line 260: The comment is a bit confusing to me. Here we still fetch by `sampleListId` not `sampleId`, right? We still invoke the same method. We just invoke it multiple times instead of once. Is that what makes caching at level of id work?

Luke-Sikina: The idea is that you fetch each one individually, so for `n` sampleListIds, there are `n` potential cache entries rather than `n!` potential entries.

The implementation has some flaws - it should probably be done under the hood a bit more, and I don't like that it's just useless extra work for anyone running an uncached portal. That said, it does accomplish its goal well enough to be worth merging.

Luke-Sikina (approved):

onursumer (approved):

web/src/main/java/org/cbioportal/web/SampleController.java
alisman: @onursumer yes.  we do it one by one.  @Luke-Sikina i agree i just wanted to accomplish this with as little refactoring as possible.   i think as far as "extra work" my rejoinder would be that anyone not using caching is most likely dealing with small data sets, otherwise, they would/should be using caching.   and if data is small then all this is just negligible.

PR #9354, opened 2022-03-01
Treatment Target

Luke-Sikina:
Treatment Target

# Function
Create two new Study View treatment charts for treatment targets. Treatment targets are the biological function that a drug might target; a good example is `Tyrosine Kinase`. The two charts reflect the sample / patient paradigm already used by the treatment and treatment group charts.

# Implementation
This PR adds the following to the codebase:
- Backend
  - Expanded ClinicalEventKeyCode
    - Delimited value types
    - Added AGENT_TARGET
  - StudyViewFilterApplier
    - Tweeked StudyViewFilterSubApplier hierarchy
    - Added PatientTreatmentTargetSubFilter and SampleTreatmentTargetSubFilter
    - Simplified wiring of StudyViewSubFIlterAppliers
  - Treatment Controller / Service / Repository / Mapper
    - Added support for comma delimited clinical event values
- Frontend
  - API: add treatment target to clinical event key enum
  - StudyViewPageStore:
    - Treatment target data methods
    - Add treatment charts to spaghetti


web/src/main/java/org/cbioportal/web/util/appliers/TreatmentRowExtractor.java
```diff
@@ -1,19 +1,17 @@
...
 @Service
 public class TreatmentRowExtractor {
     public Set<String> extractSamples(TreatmentRow row) {
-        Set<String> samples = new HashSet<>();
-        row.getSamples().forEach(sample -> {
-            samples.add(sample.key());
-        });
-
-        return samples;
+        return row.getSamples().stream()
```
Luke-Sikina, line 13: This logic hasn't changed in practice. The collection is immutable now, but that doesn't matter. No idea why I made this code so gross to begin with.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/TreatmentMyBatisRepository.java
```diff
@@ -43,8 +63,17 @@
...
+        return treatmentMapper.getAllUniqueTreatments(sampleIds, studyIds, key.getKey())
+            .stream()
+            .flatMap(treatment -> {
+                if (key.isDelimited()) {
+                    return Arrays.stream(treatment.split(key.getDelimiter()));
+                } else {
+                    return Stream.of(treatment);
+                }
+            })
+            .collect(Collectors.toSet());
```
dippindots, line 76: I have one question about this. I know `treatmentMapper.getAllUniqueTreatments` can promise give you unique treatments, but how can we know the result is still unique after we do `treatment.split(key.getDelimiter())`?

web/src/main/java/org/cbioportal/web/util/appliers/PatientTreatmentTargetFilterApplier.java
```diff
@@ -0,0 +1,20 @@
...
+public class PatientTreatmentTargetFilterApplier extends AbstractPatientTreatmentFilter {
+
+    @Override
+    protected AndedPatientTreatmentFilters getFilters(StudyViewFilter filter) {
+        return filter.getPatientTreatmentTargetFilters();
+    }
+
+    @Override
+    protected ClinicalEventKeyCode getCode() {
+        return ClinicalEventKeyCode.AgentTarget;
```
dippindots, line 18: So taking patient data as an example
`Agent` -> `PatientTreatment`
`AgentClass` -> `PatientTreatmentGroup`
`AgentTarget` -> `PatientTreatmentTarget`
Is this correct?

web/src/test/resources/springfox.properties
```diff
@@ -0,0 +1 @@
+springfox.documentation.swagger.v2.path:/api-docs
```
dippindots, line 1: What's this for?

dippindots (commented):
Thanks @Luke-Sikina! Overall it looks good to me. I have some questions, I put them in the comments. And I see the core tests are failing, I rerun it but still get error, so I think it's related to this pr, could you take a look?

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/TreatmentMyBatisRepository.java
Luke-Sikina: We don't, but `toSet` will enforce uniqueness, so you'll still end up with the correct output.

web/src/main/java/org/cbioportal/web/util/appliers/PatientTreatmentTargetFilterApplier.java
Luke-Sikina: Yup

web/src/test/resources/springfox.properties
Luke-Sikina: Oop, think that's a holdover from another PR. Deleted.

persistence/persistence-mybatis/src/main/java/org/cbioportal/persistence/mybatis/TreatmentMyBatisRepository.java
dippindots: I ignored that, you are right

PR #9367, opened 2022-03-04
Fix Generic Assay entities deletion error and profile property error

### LINKED ISSUE #9358: (Importer) Generic Assay entities don't get deleted correctly
Generic Assay entities don't get deleted correctly when the profile that gets deleted is the only one profile related.

When deleting a study, we should do as follows:

- Find all Generic Assay profiles in this study
- Go through each profile one by one, try to determine if we need to delete entities per profile instead of per study

We should also add some test cases for deleting Generic Assay entities.

- [x] Change deletion logic to determine if we need to delete entities per profile

### LINKED ISSUE #9415: (Importer) Generic Assay profile property error
When importing a Generic Assay profile, the property `pivot_threshold_value` and `value_sort_order` should be imported separately.

dippindots:
Fix #9358
Fix #9415

This pr address two things:

- Generic Assay entities don't get deleted correctly when the profile that gets deleted is the only one profile related. When deleting a study, we should do as follows: 
    - Find all Generic Assay profiles in this study. 
    - Go through each profile one by one, try to determine if we need to delete entities per profile instead of per study

- When importing a Generic Assay profile, the property pivot_threshold_value and value_sort_order should be imported separately.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenericAssay.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenericAssay.java
@@ -2,10 +2,14 @@ package org.mskcc.cbio.portal.dao;
 
 import java.sql.*;
 import java.util.*;
+import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.cbioportal.model.GenericEntityProperty;
 import org.cbioportal.model.GeneticEntity;
 import org.cbioportal.model.meta.GenericAssayMeta;
+import org.mskcc.cbio.portal.model.GeneticAlterationType;
+import org.mskcc.cbio.portal.model.GeneticProfile;
 
 public class DaoGenericAssay {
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/DaoGenericAssay.java
```diff
@@ -143,4 +147,42 @@ public static boolean geneticEntitiesOnlyExistInSingleStudy(int cancerStudyId) t
...
+     * check and delete generic assay meta records profile by profile
+     * @throws DaoException
+     */
+    public static void checkAndDeleteGenericAssayMetaInStudy(int internalCancerStudyId) throws DaoException {
+        List<GeneticProfile> geneticProfiles = DaoGeneticProfile.getAllGeneticProfiles(internalCancerStudyId);
+        List<GeneticProfile> genericAssayProfiles = geneticProfiles.stream().filter(profile -> profile.getGeneticAlterationType() == GeneticAlterationType.GENERIC_ASSAY).collect(Collectors.toList());
+        List<Pair<Integer, String>> deleteGenericAssayStatements = new ArrayList<>();
+
+        for (GeneticProfile profile : genericAssayProfiles) {
+            if (DaoGenericAssay.geneticEntitiesOnlyExistInSingleStudy(profile.getGeneticProfileId())) {
```
n1zea144, line 162: Profiles can have different generic entities so this has to be checked each iteration through the loop.

n1zea144 (approved):

PR #9374, opened 2022-03-08
Customize NA category in comparison page

### LINKED ISSUE #9372: Customize "NA" values in GENIE clinical comparison page
Customize "NA" category values in GENIE (for example "unknown" and "unspecified"), so when selecting or deselecting "Show NA", all "NA" categories change accordingly.
- [x] Show NA by default 
- [x] Support "NA" category value customize on backend, calculate p-value and q-value without "NA" 
- [x] Hide or show all "NA" categories on the plot when clicking "Show NA"

leexgh:
Fix #9372 

- [x] Support "NA" category value customize on backend, calculate p-value and q-value without "NA"

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -393,4 +393,4 @@ persistence.cache_type=no-cache
 # enable_treatment_groups=false
 
 # Set comparison categorical "NA" value (not case sensitive)
-comparison.categorical_na_value=NA,unknown
+# comparison.categorical_na_value=NA,unknown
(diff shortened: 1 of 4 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -391,3 +391,6 @@ persistence.cache_type=no-cache
 
 # Feature flag for treatment groups
 # enable_treatment_groups=false
+
+# Set comparison categorical "NA" value (not case sensitive)
+comparison.categorical_na_value=NA,unknown
```
inodb, line 396: @leexgh default should be commented out - that is to avoid us from being able to change the default value when people upgrade to a new version of the portal

PR #9404, opened 2022-03-22
lengthen sample stable id field

sheridancbio:
- sample stable ids now can be up to 63 characters
- add check for maximum stable id to validator
- migration script update (schema 2.12.13)
- remove wildcard from search for cgds.sql during unit tests

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers


core/src/main/scripts/importer/validateData.py
```diff
@@ -2667,6 +2670,14 @@ def checkLine(self, data):
                                'column_number': col_index + 1,
                                'cause': value})
                     continue
+                if len(value.strip()) > MAX_SAMPLE_STABLE_ID_LENGTH:
```
Luke-Sikina, line 2673: Thank you! We've run into this a few times.

Luke-Sikina (approved):
Looks good to me

dippindots (approved):
LGTM, thanks!

inodb (changes requested):
There seems to be a test failing: https://github.com/cBioPortal/cbioportal/runs/5648963463?check_suite_focus=true. Looks like we missed updating the db.version in `db-scripts/src/main/resources/cgds.sql`?

```
Making sure all db versions are the same in cgds.sql, pom.xml and migration.sql
pom.xml db version is 2.12.13
db-scripts/src/main/resources/cgds.sql db version is 2.12.12
db-scripts/src/main/resources/migration.sql db version is 2.12.13
db versions mismatch
```

sheridancbio:
Excellent ... I'm so glad that test was added! .. thanks @inodb
I've fixed cgds.sql and want to do a little testing of the validator code ... maybe adding a unit test for the validator itself. So let's hold off merging a bit. I'll post a comment here when testing is complete.

sheridancbio:
@inodb I have completed my system testing of the study validator. I used the study luad_tcga_pub from datahub. I changed one of the study ids to a length 63 string throughout the datafiles and the study successfully validated. Then I made that sample id length 64 and validation failed with this message in the logs:

DEBUG: data_clinical_sample.txt: Starting validation of file
ERROR: data_clinical_sample.txt: line 65: column 2: SAMPLE_ID too long (63 character maximum); value encountered: 'abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij1234'
INFO: data_clinical_sample.txt: Validation of file complete
INFO: data_clinical_sample.txt: Read 235 lines. Lines with warning: 0. Lines with error: 1

I was unable to get the unit tests for the validator script to run smoothly in my environment, so I did not add a unit test for this case into for example core/src/test/scripts/unit_tests_validate_data.py or core/src/test/scripts/system_tests_validate_data.py

But I think we can move forward with this now without the validator unit tests (or, if anyone can give guidance on running the unit tests, I'd be happy to put it in)

dippindots:
@sheridancbio If this is the error you get when run `/unit_tests_validate_data.py`
```
ModuleNotFoundError: No module named 'importer'
```
Ramya gave me this suggestion before, and it worked great for me:
You can create a copy of the importer folder in the `cbioportal/core/src/test/scripts/` folder, So the module is in the same path to run the unit tests locally. Then you should be able to run that test locally.

PR #9407, opened 2022-03-25
update importCosmicData: flexible reading for new version of cosmic

egarcialara:
Fix: import cosmic data with new COSMIC versions. Tested for v.94

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
@@ -28,17 +28,21 @@
  *
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
-*/
+ */
 
 package org.mskcc.cbio.portal.scripts;
 
 import org.mskcc.cbio.portal.dao.*;
 import org.mskcc.cbio.portal.util.*;
 import org.mskcc.cbio.portal.model.*;
+import org.springframework.util.Assert;
 
 import java.io.*;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.regex.*;
+import java.util.stream.IntStream;
 
 public class ImportCosmicData {
     private File file;
@@ -47,9 +51,17 @@ public class ImportCosmicData {
         this.file = file;
     }
 
+    private final static String geneEntryName = "GENE";
+    private final static String strandEntryName = "STRAND";
+    private final static String cdsEntryName = "CDS";
+    private final static String aaEntryName = "AA";
+    private final static String cntEntryName = "CNT";
+
     public void importData() throws IOException, DaoException {
         DaoGeneOptimized daoGeneOptimized = DaoGeneOptimized.getInstance();
-        Pattern p = Pattern.compile("GENE=([^;]+);STRAND=(.);CDS=([^;]+);AA=p\\.([^;]+);CNT=([0-9]+)");
+        // Pattern: must have gene, strand, cds, aa and cnt in any order, with other strings possibly in between.
+        Pattern p = Pattern.compile("(?=.*GENE=[^;]+.*)(?=.*STRAND=(.).*)(?=.*CDS=[^;]+.*)(?=.*AA=p\\.[^;]+.*)(?=.*CNT=[0-9]+.*)");
+
         MySQLbulkLoader.bulkLoadOn();
         FileReader reader = new FileReader(file);
         BufferedReader buf = new BufferedReader(reader);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
```diff
@@ -89,10 +107,13 @@ public void importData() throws IOException, DaoException {
...
                     if (keyword == null) {
+                        System.out.println("Mutation keyword in COSMIC not recognized: "+aa);
                         continue;
                     }
-                    
-                    int count = Integer.parseInt(m.group(5));
+
+                    String count_field = fields[inxCntField];
+                    count_field = removePrefix(count_field);
+                    int count = Integer.parseInt(count_field);
```
pvannierop, line 116:
```suggestion
                    int count = Integer.parseInt(removePrefix(fields[inxCntField]));
```

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
```diff
@@ -65,18 +67,34 @@ public void importData() throws IOException, DaoException {
...
-//                    if (gene.contains("_ENST")) {
-//                        gene = gene.substring(0,gene.indexOf("_ENST"));
-//                    }
+                    // Remove fields that won't be used, so that they don't interfere with column indices
+                    // If field in last place, it won't be removed, but it won't interfere either
+                    infoColumnValue = infoColumnValue.replaceAll("LEGACY_ID=[^;]+;","")
+                                                        .replaceAll("HGVSP=[^;]+;","")
+                                                        .replaceAll("HGVSG=[^;]+;","")
+                                                        .replaceAll("HGVSC=[^;]+;","")
+                                                        .replaceAll("SNP;","");
```
pvannierop, line 85: What is the reason for removing these fields?

pvannierop (changes requested):
One question and a change request.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
egarcialara: It is common that in the same file the fields HGVSP, HGVSG, HGVSC and SNP are present only in some rows. This affects the selection after the first row.

pvannierop: But the presence of these fields does not interfere with the function of the script, does it? At least I do not see how that is a problem.

egarcialara: It does when the column 7 looks like:
```
GENE=; STRAND=; CDS=; HGVSP=; AA=; CNT=
GENE=; STRAND=; CDS=; HGVSP=; AA=; CNT=
GENE=; STRAND=; CDS=; AA=; CNT=
GENE=; STRAND=; CDS=; HGVSPC; HGVSP=; AA=; CNT=
GENE=; STRAND=; CDS=; HGVSP=; HGVSP=; SNP; AA=; CNT=
```
In this case only rows 1 and 2 would be correct.

pvannierop (approved):

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
```diff
@@ -65,18 +77,22 @@ public void importData() throws IOException, DaoException {
...
-//                    if (gene.contains("_ENST")) {
-//                        gene = gene.substring(0,gene.indexOf("_ENST"));
-//                    }
+                    Map<String, String> fieldValues = evaluateFieldValues(infoColumnValue);
+
+                    String gene = fieldValues.get(geneEntryName);
+                           
+                    if (gene.contains("_ENST")) {
+                        gene = gene.substring(0,gene.indexOf("_ENST"));
+                    }
```
dippindots, line 95: I noticed that we comment this part out currently, could you tell me why do we need to add this back?

dippindots (commented):
Thanks @egarcialara. It looks good to me, I just have one question regarding the `gene` field.

core/src/main/java/org/mskcc/cbio/portal/scripts/ImportCosmicData.java
egarcialara: Hi @dippindots, thank you for reviewing the PR!

I have uncommented it because the new cosmic files have genes with that prefix (e.g. `GENE=PLCH2_ENST00000378488`), and this code allows to insert the record instead of getting the message `Gene symbol in COSMIC not recognized: PLCH2_ENST00000378488`.

dippindots (approved):

dippindots:
@egarcialara @pvannierop Last question before we merge it, is this pr backward compatible? If we merge it, can we still import the old version of cosmic data?

egarcialara:
> @egarcialara @pvannierop Last question before we merge it, is this pr backward compatible? If we merge it, can we still import the old version of cosmic data?

Hi @dippindots, yes this can import the old version. We just made the reading more flexible.

dippindots:
@egarcialara Thanks, got it. Could you please squash all the commit into one and rebase to the latest master? I will merge after that.

egarcialara:
@dippindots Done!

PR #9427, opened 2022-03-31
Downgrade to Java 8

Luke-Sikina:
- There's a bad RCE involving Spring + Java 9 + Tomcat
- Downgrading to Java 8 keeps us safe and can be released quickly
since we already support Java 8

### DIFF
--- a/docker/web-and-data/Dockerfile
+++ b/docker/web-and-data/Dockerfile
@@ -11,7 +11,7 @@
 # NOTE: the .git folder is included in the build stage, but excluded 
 # from the final image. No confidential information is exposed.
 # (see: stackoverflow.com/questions/56278325)
-FROM maven:3-openjdk-11 as build
+FROM maven:3-openjdk-8 as build
 
 # download maven dependencies first to take advantage of docker caching
 COPY pom.xml                                     /cbioportal/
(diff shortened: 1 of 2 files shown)


docker/web/Dockerfile
```diff
@@ -13,13 +13,13 @@
 #
 # WARNING: the shendoah image is a nightly, untested, experimental build. If
 # you want to use an official openjdk image instead use the web-and-data image.
```
inodb, line 15: maybe remove this comment since this switches to official openjdk image

docker/web-and-data/Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN mvn dependency:go-offline --fail-never
 COPY $PWD /cbioportal
 RUN mvn -DskipTests clean install
 
-FROM openjdk:11-jre-slim
```
dippindots, line 40: The reason we are using `openjdk-shenandoah:8` instead of `openjdk:8-jre-slim` is: we want to keep using shenandoah GC, and `openjdk:8-jre-slim` doesn't support it.

dippindots (approved):

dippindots:
We tested it on beta portal and decided to merge this. We should do a follow-up discussion for upgrading to higher java version and ZGC/shenandoah.

dippindots:
Related discussion topics: https://github.com/cBioPortal/cbioportal/issues/9442

PR #9434, opened 2022-04-01
Fix query string decoder

### LINKED ISSUE #9432: Proxy fails when there are query parameters
This seems to happen because of a redundant `?` in the beginning of the decoded query string.

onursumer:
Fix #9432


web/src/main/java/org/cbioportal/proxy/Monkifier.java
```diff
@@ -0,0 +1,54 @@
...
+        return UriComponentsBuilder
+            .newInstance()
+            .queryParams(CollectionUtils.toMultiValueMap(decodedQueryParams))
+            .build()
+            // encode to avoid potentially invalid query strings
+            .encode()
+            .toString()
+            // remove ? from the beginning of the query string,
+            // we only want the parameters (e.g: param1=value1&param2=value2)
+            .replaceFirst("\\?", "");
```
onursumer, line 52: This is the actual fix.

web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -90,38 +88,7 @@ public String proxyEncodedOncokb(
...
-                    e -> this.decodeBase64(e.getKey()),
-                    e -> Arrays.stream(e.getValue()).map(this::decodeBase64).collect(Collectors.toList())
-                )
-            );
-        
-        return UriComponentsBuilder
-            .newInstance()
-            .queryParams(CollectionUtils.toMultiValueMap(decodedQueryParams))
-            .build()
-            .toString();
```
onursumer, line 124: Moved to `Monkifier.java`

zhx828 (approved):
LGTM 🐒

PR #9449, opened 2022-04-08
Treatment Performance Improvements

### LINKED ISSUE #9447: treatment/samples endpoint causes database transfer which sometimes brings down portal

Luke-Sikina:
- Treatment endpoints were tanking on large multi study queries
- Filter down study list to only studies w/ treatments prior to
performing treatment queries

Fix #9447

### DIFF
--- a/persistence/persistence-api/src/main/java/org/cbioportal/persistence/TreatmentRepository.java
+++ b/persistence/persistence-api/src/main/java/org/cbioportal/persistence/TreatmentRepository.java
@@ -1,13 +1,13 @@
 package org.cbioportal.persistence;
 
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
 import org.cbioportal.model.ClinicalEventKeyCode;
 import org.cbioportal.model.ClinicalEventSample;
 import org.cbioportal.model.Treatment;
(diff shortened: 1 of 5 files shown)


persistence/persistence-api/src/main/java/org/cbioportal/persistence/TreatmentRepository.java
```diff
@@ -27,4 +27,7 @@
 
     @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public Integer getSampleCount(List<String> studies);
+
+    @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
+    public Set<String> getFilteredStudyIds(List<String> studyIds, ClinicalEventKeyCode key);
```
alisman, line 32: can we change this to cache at single study level?  i know this requires then querying at single study level but it's a tradeoff.

Luke-Sikina: Sure. Fixed

alisman (approved):

Issue #9413, opened 2022-03-29
add NA terms to genie private portal


inodb:
Update our configuration to also treat the following items as NA:

- Not documented in medical record
- Test ordered, results not in chart
- Test not done

For both genie private and genie public portals:

https://github.com/knowledgesystems/knowledgesystems-k8s-deployment/blob/master/cbioportal/cbioportal_backend_genie_private.yaml#L153-L156

For testing one might use https://genie-private.cbioportal.org/study/summary?id=nsclc_genie_bpc

leexgh:
Testing link: https://genie-private.cbioportal.org/study/summary?id=crc_genie_bpc

PR #9462, opened 2022-04-14
Change comparison page na terms delimiter

leexgh:
Part of: https://github.com/cBioPortal/cbioportal/issues/9413
Change na delimiter to "|" in backend

### DIFF
--- a/docs/portal.properties-Reference.md
+++ b/docs/portal.properties-Reference.md
@@ -194,9 +194,9 @@ skin.show_settings_menu=
 ```
 
 ## Customize "NA" values in group comparison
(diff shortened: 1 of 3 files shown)


docs/portal.properties-Reference.md
```diff
@@ -194,9 +194,9 @@ skin.show_settings_menu=
 ```
 
 ## Customize "NA" values in group comparison
-Customize "NA" category in group comparison, for example "NA" and "unknown", all values are treated as "NA" and could be shown or hidden together on comparison page.
+Customize "NA" category in group comparison, for example "NA" and "unknown", all values are treated as "NA" and could be shown or hidden together on comparison page. Values are separated by "|".
```
leexgh, line 197: Talk with Brooke and decided to change to `|` instead of `,`

inodb (approved):

PR #9487, opened 2022-04-22
Require API token for the plain proxy

### LINKED ISSUE #9411: Retire oncokb proxy endpoint
We should remove `proxy/oncokb` endpoint and update the projects/packages depending on it.

onursumer:
Fix #9411 

Proxy now requires a valid token.

### DIFF
--- a/web/src/main/java/org/cbioportal/proxy/ProxyController.java
+++ b/web/src/main/java/org/cbioportal/proxy/ProxyController.java
@@ -43,20 +43,22 @@ public class ProxyController {
             String.class
         ).getBody();
     }
-
-    // TODO remove this endpoint after external projects switch to the new proxy
+    
     @RequestMapping("/oncokb/**")
     public String proxyOncokb(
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -43,20 +43,22 @@ public String proxy(@RequestBody(required = false) String body, HttpMethod metho
...
-
-    // TODO remove this endpoint after external projects switch to the new proxy
+    
     @RequestMapping("/oncokb/**")
     public String proxyOncokb(
         @RequestBody(required = false) String body,
         HttpMethod method,
         HttpServletRequest request
     ) throws URISyntaxException {
+        String token = request.getHeader("X-Proxy-User-Agreement");
```
onursumer, line 53: We can also use a different header name, but that requires adding the new header into the CORS allowed headers list.

inodb (approved):

PR #9514, opened 2022-05-09
Fix calculation of quartile values for study view data binning

pvannierop:
# Problem
Functions used for calculation of quartile values for numeric data do not take into account low complexity series where Q1, Q2 and or Q3 are identical. This resulted in in correct values where low complexity applied.

# Solution
Strategy for low complexity data series was added to quartile calculation. Tests were added to document this behavior.

### DIFF
--- a/web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
+++ b/web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
@@ -1,7 +1,6 @@
 package org.cbioportal.web.util;
 
 import com.google.common.collect.Range;
-import org.apache.commons.lang3.ObjectUtils;
 import org.cbioportal.model.DataBin;
 import org.springframework.stereotype.Component;
 
@@ -66,6 +65,28 @@ public class DataBinHelper {
 
         return dataBin;
     }
+    
+    public List<BigDecimal> calcQuartileBoundaries(List<BigDecimal> sortedValues) {
+        // Edge case: some of Q1, Q2, and Q3 are the same value.
+        // Solution: reduce bins to represent unique values only.
+        // Note: cannot use stream.distinct() because BigDecimal does
+        // not play nice with this (e.g., "2.5E+2" is not identical to "250"
+        // when using Object.equals())
+        final BigDecimal q1 = calcQ1(sortedValues);
+        final BigDecimal q2 = calcMedian(sortedValues);
+        final BigDecimal q3 = calcQ3(sortedValues);
+        List<BigDecimal> boundaries = new ArrayList<>();
+        boundaries.add(q1);
+        // Check Q1 smaller than Q2
+        if (q1.compareTo(q2) < 0) {
+            boundaries.add(q2);
+        }
+        // Check Q2 smaller than Q3
+        if (q2.compareTo(q3) < 0) {
+            boundaries.add(q3);
+        }
+        return boundaries;
+    }
 
     public Range<BigDecimal> calcBoxRange(List<BigDecimal> sortedValues) {
         if (sortedValues == null || sortedValues.size() == 0) {
@@ -75,7 +96,7 @@ public class DataBinHelper {
         // Find a generous IQR. This is generous because if (values.length / 4)
         // is not an int, then really you should average the two elements on either
         // side to find q1 and q3.
-        Range<BigDecimal> interquartileRange = calcInterquartileRange(sortedValues);
+        Range<BigDecimal> interquartileRange = calcInterquartileRangeApproximation(sortedValues);
 
         BigDecimal q1 = interquartileRange.lowerEndpoint();
         BigDecimal q3 = interquartileRange.upperEndpoint();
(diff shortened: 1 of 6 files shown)


onursumer (approved):
LGTM

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -145,16 +145,17 @@ public Range<BigDecimal> calcInterquartileRange(List<BigDecimal> sortedValues) {
...
     public BigDecimal calcQ1(List<BigDecimal> sortedValues) {
         return sortedValues.isEmpty() ?
-            null : sortedValues.get((int) Math.floor(sortedValues.size() / 4.0));
+            null : sortedValues.get((int) Math.floor(sortedValues.size() * 0.25));
     }
 
     public BigDecimal calcMedian(List<BigDecimal> sortedValues) {
-        return ObjectUtils.median(sortedValues.toArray(new BigDecimal[0]));
+        return sortedValues.isEmpty() ?
+            null : sortedValues.get((int) Math.floor(sortedValues.size() * 0.50));
```
dippindots, line 153: What if we have an even number of elements? It doesn't look like a median. I think we should change it, or rename it if we are not calculating `median` at here.
image

dippindots (commented):
I have one comment related to the `median` calculation, I don't think that's exactly same as our previous implementation. Others look good to me.

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
```diff
@@ -145,16 +145,17 @@ public Range<BigDecimal> calcInterquartileRange(List<BigDecimal> sortedValues) {
 
     public BigDecimal calcQ1(List<BigDecimal> sortedValues) {
         return sortedValues.isEmpty() ?
-            null : sortedValues.get((int) Math.floor(sortedValues.size() / 4.0));
+            null : sortedValues.get((int) Math.floor(sortedValues.size() * 0.25));
```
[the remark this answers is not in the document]
pvannierop: @Luke-Sikina No, just to bring the implementations more inline.

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
pvannierop: @dippindots @Luke-Sikina  Well, that existing implementation is wrong! Looking at the implementation of that one, it will cast the data to a series of unigue values (note the **TreeSet**) before evaluating the median:

```
    @SafeVarargs
    public static <T extends Comparable<? super T>> T median(T... items) {
        Validate.notEmpty(items);
        Validate.noNullElements(items);
        TreeSet<T> sort = new TreeSet();
        Collections.addAll(sort, items);
        T result = (Comparable)sort.toArray()[(sort.size() - 1) / 2];
        return result;
    }
```

This is incorrect. The median should take multiple identical observations into account.

web/src/main/java/org/cbioportal/web/util/DataBinner.java
```diff
@@ -308,11 +309,13 @@ public <T extends DataBinFilter> Collection<DataBin> calcNumericalDataBins(
             } else if (DataBinFilter.BinMethod.MEDIAN == binMethod) {
                 // NOOP - handled later
             } else if (DataBinFilter.BinMethod.QUARTILE == binMethod) {
-                List<BigDecimal> bins = Arrays.asList(
+                // Edge case: some of Q1, Q2, and Q3 are the same value.
+                // Solution: reduce bins to represent unique values only.
+                List<BigDecimal> bins = Stream.of(
```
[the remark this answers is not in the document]
pvannierop: @Luke-Sikina Well, the existing implementation only worked when Q1, Q2 and Q3 were all different values. I practice this is not always the case. In this data series [10, 250, 250, 250, 250, 250, 250, 250] Q1, Q2 and Q3 are all 250. This would in effect mean that there is a single bin, not three. The included tests describe these situations quite ok I think.

Btw the break of logic was also caused by incorrect login for evaluating the median (see comment by @dippindots below).

web/src/main/java/org/cbioportal/web/util/DataBinHelper.java
pvannierop: @dippindots @Luke-Sikina 
>What if we have an even number of elements?
The issue here is that the calcQ1 and calcQ3 functions also do not calculate true first and third quartile boundaries. I did this by extensive testing. The problem here is that there are no tests included for any of these functions. I think you have a valid point here, but I do not think that it is in the scope of this PR.

pvannierop: It turns out that the legacy calcQ1 and calcQ3 functions were used to calculate the outer-whiskers of (I presume) a box plot which used some very funky approximations.  I have renamed these functions to valueCloseToQ1 and valueCloseToQ3, respectively, to indicate that these legacy functions do actually calculate quartiles as suggested in their name. I have (re-)implemented the calcQ1, calcMedian, and calcQ3 functions to calculate actual quartile boundaries. These new functions are used for binning of histograms. I have added plenty of tests that should document behavior of all involved functions. Can you re-review is you have time

dippindots (approved):

Luke-Sikina (approved):

PR #9527, opened 2022-05-11
implement gene alias resolve in oncokb scripts

SRodenburg:
The OncoKB enrichment scripts would crash for unknown gene symbols.
They now check the genes endpoint whether they can extract the Entrez by symbol alias in case of unknown symbols.

Also, small changes to the validator. The `dumpPortalInfo.pl` script does in fact export a file `genesaliases.json` but there is no dedicated API endpoint for aliases. Iterative queries to the genes API endpoint for aliases is not viable. With the current changes the validator loads aliases in offline mode, otherwise defaults to empty set.

I made a few small changes to the JSON format of the `genesaliases.json` file, that is created with a portal dump, to match the API format.


inodb (approved):
LGTM - thanks!

PR #9537, opened 2022-05-13
Document Logback config, improve usage

Luke-Sikina:
Fix https://github.com/cBioPortal/cbioportal/issues/9538. Describe changes proposed in this pull request:
- update poms with new excludes
- update main pom to copy logback config
- update README with logback info


docs/Authenticating-Users-via-SAML.md
```diff
@@ -228,15 +228,26 @@ If this does not happen, see the Troubleshooting Tips  below.
...
 
 ```
-# Change INFO to DEBUG, if you want to see debugging info on underlying libraries we use.
-log4j.rootLogger=DEBUG, a
-
-# Change INFO to DEBUG, if you want see debugging info on our packages and spring security packages.
-log4j.category.org.mskcc=DEBUG
-log4j.logger.org.springframework.security=DEBUG
+# Change warn to debug, if you want to see debugging info on underlying libraries we use.
+<root level="warn">
```
Luke-Sikina, line 236: /s/warn/debug

docs/Authenticating-and-Authorizing-Users-via-keycloak.md
```diff
@@ -345,15 +345,25 @@ More information on configuration of the cBioPortal backend can be found in [Aut
...
 
 ```
-# Change INFO to DEBUG, if you want to see debugging info on underlying libraries we use.
-log4j.rootLogger=DEBUG, a
-
-# Change INFO to DEBUG, if you want see debugging info on our packages and spring security packages.
-log4j.category.org.mskcc=DEBUG
-log4j.logger.org.springframework.security=DEBUG
+# Change warn to debug, if you want to see debugging info on underlying libraries we use.
+<root level="warn">
```
Luke-Sikina, line 352: /s/warn/debug/

sheridancbio (approved):
I'm approving, although we should not rely on me alone. My understanding now is that a switch in logging was made months ago to logback configuration, but the configuration files in the build setup (pom.xml) fell out of sync. This PR seems to give updated guidance to deployers on how to configure logback (dropping guidance on log4j config files), and has the logic for including the logback.xml-EXAMPLE file in the build when the user does not provide a custom logback.xml file of their own.

The fact that the pom.xml reference to logback.xml is in an <exclude> element is a  little confusing to me --- because I think we want them included during build. But maybe this is to exclude them from getting bundled into sub-modules as they are deployed at the root level.

👍

inodb:
Changed the title and description slightly b/c I think the critical piece is that this changes the default Docker setup to no longer do debug logging by default. Feel free to change tho

PR #9562, opened 2022-05-20
Upgrade to Java 17

Luke-Sikina:
- Upgrade dependencies where needed for Java 17
- Add dependencies that were removed from JVM so that ehcache works
- Fix CI workflows
- Fix integration test with temporary reference to Java 17 image
  - Once we move the cbioportal-docker-compose repo to a 5.x cbioportal
  image, we can revert this
- Switch docker images over to Java 17 base images, move away from
Oracle openjdk images

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ version: 2
 jobs:
     build_backend:
         docker:
-            - image: maven:3-openjdk-11
+            - image: maven:3-openjdk-17
         working_directory: /tmp/repos/cbioportal
         steps:
             - checkout
@@ -21,7 +21,11 @@ jobs:
                 name: Build and unzip war
                 command: |
                     mvn -DskipTests clean install && \
-                    unzip portal/target/cbioportal*.war -d portal/target/war-exploded
+                    mkdir portal/target/war-exploded && \
+                    mv portal/target/cbioportal*.war portal/target/war-exploded/ && \
+                    cd portal/target/war-exploded/ && \
+                    jar -xvf cbioportal*.war && \
+                    cd ../../..
             - save_cache:
                 paths:
                   - ~/.m2
@@ -180,6 +184,7 @@ jobs:
             - run:
                 name: Start cbioportal and other services
                 command: |
+                    export DOCKER_IMAGE_CBIOPORTAL='cbioportal/cbioportal:demo-java-17' && \
                     $TEST_HOME/docker_compose/start.sh
             - run:
                 name: Change owner of keycloak MySQL database files (needed by cache)
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -17,10 +17,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: ./cbioportal
-      - name: 'Set up JDK 11'
+      - name: 'Set up JDK 17'
         uses: actions/setup-java@v1
         with:
-          java-version: 11
+          java-version: 17
       - name: 'Cache Maven packages'
         uses: actions/cache@v1
         with:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -8,6 +8,7 @@ jobs:
       PORTAL_SOURCE_DIR: /home/runner/work/cbioportal/cbioportal/cbioportal
       PORTAL_COMPOSE_DIR: /home/runner/work/cbioportal/cbioportal/cbioportal-docker-compose
       PORTAL_INFO_DIR: /home/runner/work/cbioportal/cbioportal/portalInfo
+      DOCKER_IMAGE_CBIOPORTAL: cbioportal/cbioportal:demo-java-17
     steps:
       - name: 'Checkout cbioportal repo'
         uses: actions/checkout@v2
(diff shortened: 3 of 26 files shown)


.circleci/config.yml
```diff
@@ -21,7 +21,11 @@ jobs:
                 name: Build and unzip war
                 command: |
                     mvn -DskipTests clean install && \
-                    unzip portal/target/cbioportal*.war -d portal/target/war-exploded
+                    mkdir portal/target/war-exploded && \
```
Luke-Sikina, line 24: unzip isn't in the image, nor is apt/yum/etc. This seems to be the equivalent command

.github/workflows/integration-test.yml
```diff
@@ -8,6 +8,7 @@ jobs:
       PORTAL_SOURCE_DIR: /home/runner/work/cbioportal/cbioportal/cbioportal
       PORTAL_COMPOSE_DIR: /home/runner/work/cbioportal/cbioportal/cbioportal-docker-compose
       PORTAL_INFO_DIR: /home/runner/work/cbioportal/cbioportal/portalInfo
+      DOCKER_IMAGE_CBIOPORTAL: cbioportal/cbioportal:demo-java-17
```
Luke-Sikina, line 11: TODO: revert this once everything is merged and cbioportal-docker-compose is updated.

pom.xml
```diff
@@ -570,7 +575,7 @@
       <dependency>
         <groupId>org.mybatis</groupId>
         <artifactId>mybatis</artifactId>
-        <version>3.5.7</version>
+        <version>3.5.10-SNAPSHOT</version>
```
Luke-Sikina, line 578: TODO: Before merging, switch from 3.5.10-SNAPSHOT to 3.5.10

.github/workflows/core-test.yml
```diff
@@ -17,10 +17,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: ./cbioportal
-      - name: 'Set up JDK 11'
+      - name: 'Set up JDK 17'
         uses: actions/setup-java@v1
         with:
-          java-version: 11
+          java-version: `17`
```
onursumer, line 23: minor but looks like we have the version without quotes in other workflows.
```suggestion
          java-version: 17
```

onursumer (approved):
LGTM!

dippindots (approved):
Thanks! Looks good to me

docker/web/Dockerfile
```diff
@@ -13,13 +13,14 @@
...
-FROM maven:3-openjdk-11 as build
+FROM maven:3-openjdk-17 as build
 COPY $PWD /cbioportal
 WORKDIR /cbioportal
 ARG MAVEN_OPTS=-DskipTests
 RUN mvn ${MAVEN_OPTS} clean install
 
-FROM shipilev/openjdk:11
+# We shouldn't use openjdk images. See here for explanation: https://hub.docker.com/_/openjdk
+FROM eclipse-temurin:17
```
inodb, line 23: Does this offer Shenandoah as well? We will need to double check the parameters and confirm the behavior is still the same for production since we are using that garbage collector there CC @dippindots. We could switch to another low pause GC if preferred but we would need to test its performance

inodb (approved):
It looks good overall, thanks! Not entirely sure about switching the image for the web docker container since we are using the Shenandoah GC in production. It will require some monitoring in production to confirm its behavior is similar

docker/web/Dockerfile
Luke-Sikina: Good question - I checked and it looks like we're all set. Here are some relevant log statements:

```
cbioportal-container           | Running: /bin/sh -c java -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC -XX:ShenandoahUncommitDelay=1000 -XX:ShenandoahGuaranteedGCInterval=10000 -Dauthenticate=noauthsessionservice -Dsession.service.url=http://cbioportal-session:5000/api/sessions/my_portal/ -jar webapp-runner.jar -AmaxHttpHeaderSize=16384 -AconnectionTimeout=20000 --enable-compression /cbioportal-webapp
cbioportal-container           | Connector attributes
cbioportal-container           | property: maxHttpHeaderSize - 16384(OK)
cbioportal-container           | property: connectionTimeout - 20000(OK)
...
cbioportal-container           | May 31, 2022 2:07:51 PM org.apache.coyote.AbstractProtocol start
cbioportal-container           | INFO: Starting ProtocolHandler ["http-nio-8080"]
```

averyniceday:
Just tried propagating this build through the rest of our applications (genome-nexus-annotation-pipeline and importer) and a couple things need to be resolved before this can be merged.

1. Can you add a jitpack.yml file to the repo as part of the PR? Jitpack defaults to building with Java8 which is not compatible with the new codebase. A config file can be used to override the default and have Jitpack build with Java17

2. Requires upgrading several packages including (vintage-junit, spring boot, spring version) inside Genome Nexus

3. Need to install Java 17 across our servers for build

averyniceday:
are the automated core unit tests running here @Luke-Sikina ? 

Don't see it in the list of checks -- running it on commandline in our local environment with Java17 fails (mvn clean install).

averyniceday:
Screen Shot 2022-06-06 at 12 26 01 PM

looks like the quotes are invalid in the yaml and the test isn't running.

mandawilson:
@Luke-Sikina I fixed the error you see in the unit tests by updating jacoco-maven-plugin to:

<groupId>org.jacoco</groupId>
<artifactId>jacoco-maven-plugin</artifactId>
<version>0.8.8</version>

See: https://www.jacoco.org/jacoco/trunk/doc/changes.html

But then I ran into other problems.  I was running mvn clean install locally.

Luke-Sikina:
Closing - bundled with spring-boot

PR #9627, opened 2022-06-21
Move new documentation site into docs folder and update to latest changes 

### LINKED ISSUE #9602: Documentation site is out of sync
We currently have two places for documentation:

- https://github.com/cBioPortal/cbioportal/tree/master/docs
- https://github.com/cBioPortal/documentation/tree/master/src (shown on docs.cbioportal.org)

We should have a single place for the documentation. Previously we used `/docs` in the main repo to be able to update both the docs and code in a single PR. Also the release notes helps keep track of documentation updates which is a nice benefit. However we could def choose to change this model to only update the documentation repo if it makes things easier, but we should prolly remove `/docs/` in that case

alisman:
Fixes #9602


inodb (commented):
@alisman Thanks! Looks like there are still two folders in this PR, both `docs` and `documentation`. Shouldn't `documentation` be removed?

inodb (approved):
LGTM - thanks!

pieterlukasse:
@alisman @inodb thanks for all the efforts towards maintaining good documentation! 

Just wanted to bring to your attention that this PR did break some links to `portal.properties-Reference.md`. If you do a global search for `portal.properties-Reference.md` (e.g. in VS Code) you will find them. E.g.

image

PR #9640, opened 2022-06-28
Remove hotspots url property as dead code

ruslan-forostianov:
Clean up the code. The `hotspots.url` is a dead config option and all hotspot info comes from genomenexus atm.


inodb (approved):
LGTM - thank you!

PR #9680, opened 2022-07-18
Do not run studyIdHasTreatments() more than neccesary

pvannierop:
# Problem
The `/api/treatments/patient` endpoint becomes very slow for large studies. This is caused by redundant calls to the _studyIdHasTreatments()_ method.

# Solution
Reduce the number of calls by calling for unique studyIds, rather than for all studyIds.


Luke-Sikina (approved):

onursumer (approved):
🙌

PR #9692, opened 2022-07-19
Remove NCBI-Build from minimal MAF format

rmadupuri:
- NCBI-Build is currently added as a required column in the minimal MAF whereas it is not really needed. 
The MAF is considered GRCh37 build by default.


inodb (approved):
Thank you!!

PR #9709, opened 2022-07-26
Rename oncoprint clinical tracks config property

BasLee:
Rename the oncoprint clinical tracks configuration property from `oncoprint_clinical_tracks_show_by_default` to `oncoprint_clinical_tracks_config_json`, as suggested in [frontend PR 4152]( https://github.com/cBioPortal/cbioportal-frontend/pull/4152#discussion_r929804188)

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -292,7 +292,7 @@ oncoprint.defaultview=patient
 oncoprint.oncokb.default=true
 oncoprint.hotspots.default=true
 # oncoprint.hide_vus.default=true
-# oncoprint.clinical_tracks.show_by_default=classpath:/oncoprint-default-tracks.json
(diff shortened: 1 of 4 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -292,7 +292,7 @@ oncoprint.defaultview=patient
 oncoprint.oncokb.default=true
 oncoprint.hotspots.default=true
 # oncoprint.hide_vus.default=true
-# oncoprint.clinical_tracks.show_by_default=classpath:/oncoprint-default-tracks.json
+# oncoprint.clinical_tracks.config_json=classpath:/oncoprint-default-tracks.json
```
alisman, line 295: We can merge this but we should think about whether we want this to be separate json file or perhaps part of a larger json configuration file.  @inodb don't we already have that?

alisman (approved):

PR #9722, opened 2022-08-01
Add new alteration counts endpoint

onursumer:
Related to #9305

- Add an endpoint to fetch alteration counts by gene
- Request parameters are entrez gene ids, molecular profile case identifiers, and an alteration filter.
- Fetching by study or cohort is not supported yet.

### DIFF
--- a/web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
+++ b/web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
@@ -1,0 +1,110 @@
+package org.cbioportal.web.util;
+
+import org.apache.commons.math3.util.Pair;
+import org.cbioportal.model.*;
+import org.cbioportal.model.util.Select;
+import org.cbioportal.service.AlterationCountService;
+import org.cbioportal.service.GenePanelService;
+import org.cbioportal.web.parameter.AlterationCountFilter;
+import org.cbioportal.web.parameter.Projection;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component
+public class AlterationCountUtil {
+    @Autowired
+    private AlterationCountService alterationCountService;
+    @Autowired
+    private GenePanelService genePanelService;
+    
+    public Pair<List<AlterationCountByGene>, Long> getAlterationCounts(
+        AlterationCountFilter alterationCountFilter,
+        AlterationCountType alterationCountType,
+        List<Integer> entrezGeneIds
+    ) {
+        // select all genes if no list of entrez gene ids provided
+        Select<Integer> geneSelect = (entrezGeneIds == null || entrezGeneIds.isEmpty()) ?
+            Select.all(): Select.byValues(entrezGeneIds);
+        List<MolecularProfileCaseIdentifier> molecularProfileCaseIdentifiers =
+            alterationCountFilter.getMolecularProfileCaseIdentifiers();
+        boolean includeFrequency = true;
+        boolean includeMissingAlterationsFromGenePanel = false;
+        AlterationFilter alterationFilter = alterationCountFilter.getAlterationFilter();
+
+         return alterationCountType.equals(AlterationCountType.SAMPLE) ?
+            alterationCountService.getSampleAlterationCounts(
+                molecularProfileCaseIdentifiers,
+                geneSelect,
+                includeFrequency,
+                includeMissingAlterationsFromGenePanel,
+                alterationFilter
+            ) :
+            alterationCountService.getPatientAlterationCounts(
+                molecularProfileCaseIdentifiers,
+                geneSelect,
+                includeFrequency,
+                includeMissingAlterationsFromGenePanel,
+                alterationFilter
+            );
+    }
+    
+    public Set<GenePanelToGene> getProfiledGenes(
+        AlterationCountFilter alterationCountFilter,
+        AlterationCountType alterationCountType,
+        List<Integer> entrezGeneIds
+    ) {
+        List<MolecularProfileCaseIdentifier> molecularProfileCaseIdentifiers =
+            alterationCountFilter.getMolecularProfileCaseIdentifiers();
+        Set<GenePanelToGene> profiledGenes;
+        List<GenePanelData> genePanelDataList = alterationCountType.equals(AlterationCountType.SAMPLE) ?
+            genePanelService.fetchGenePanelDataInMultipleMolecularProfiles(molecularProfileCaseIdentifiers) :
+            genePanelService.fetchGenePanelDataInMultipleMolecularProfilesByPatientIds(molecularProfileCaseIdentifiers);
+
+        if (genePanelDataList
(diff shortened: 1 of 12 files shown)


web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
```diff
@@ -0,0 +1,109 @@
...
+
+        if (genePanelDataList
+            .stream()
+            .anyMatch(
+                genePanelData -> genePanelData.getProfiled() && genePanelData.getGenePanelId() == null
+            )
+        ) {
+            // if there is at least one sample/patient with all genes profiles,
+            // then all genes are profiled for the current data set
+            profiledGenes = null;
```
Luke-Sikina, line 75: I don't get this logic. I would assume that a method called `getProfiledGenes` would return all the genes profiled by the molecular profiles listed in the alt count filter. This logic of _if there exists a gene panel for these molecular profiles that was profiled, then return `null`_ just makes no sense to me. What is `null` representing here?

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
```diff
@@ -0,0 +1,109 @@
...
+            alterationCountFilter.getMolecularProfileCaseIdentifiers();
+        Set<GenePanelToGene> profiledGenes;
+        List<GenePanelData> genePanelDataList = alterationCountType.equals(AlterationCountType.SAMPLE) ?
+            genePanelService.fetchGenePanelDataInMultipleMolecularProfiles(molecularProfileCaseIdentifiers) :
+            genePanelService.fetchGenePanelDataInMultipleMolecularProfilesByPatientIds(molecularProfileCaseIdentifiers);
+
+        if (genePanelDataList
+            .stream()
+            .anyMatch(
+                genePanelData -> genePanelData.getProfiled() && genePanelData.getGenePanelId() == null
```
Luke-Sikina, line 70: Remind me again - why do we allow gene panel data objects that don't have a gene panel? What does the object even represent at that point?

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
```diff
@@ -0,0 +1,109 @@
...
+            )
+        ) {
+            // if there is at least one sample/patient with all genes profiles,
+            // then all genes are profiled for the current data set
+            profiledGenes = null;
+        } else {
+            List<String> profiledPanelIds =  genePanelDataList
+                .stream()
+                .filter(GenePanelData::getProfiled)
+                .map(GenePanelData::getGenePanelId)
```
Luke-Sikina, line 80: How come you're not doing the null check on genePanelId here?

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
```diff
@@ -0,0 +1,109 @@
...
+                .collect(Collectors.toList());
+
+            if (profiledPanelIds.isEmpty()) {
+                profiledGenes = Collections.emptySet();
+            } else {
+                List<GenePanel> panels = genePanelService.fetchGenePanels(
+                    profiledPanelIds,
+                    Projection.DETAILED.name()
+                );
```
Luke-Sikina, line 89:
```suggestion
                List<GenePanel> panels = genePanelService.fetchGenePanels(profiledPanelIds, Projection.DETAILED.name());
```
Exactly 120 columns :)

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
```diff
@@ -0,0 +1,109 @@
...
+
+                Stream<GenePanelToGene> genePanelGenes = panels
+                    .stream()
+                    .map(GenePanel::getGenes)
+                    .filter(Objects::nonNull)
```
Luke-Sikina, line 94: The gene panel API is breaking me yet again. Why can this be null? What use is a gene panel that does not refer to a list of genes? Can we extend the gene panel service API to include some methods with better guarantees about gene panels, because this is super counter intuitive

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
```diff
@@ -0,0 +1,109 @@
...
+                    .flatMap(Collection::stream);
+
+                // if entrezGeneIds is not empty, further filter by given entrez gene ids
```
Luke-Sikina, line 97: This seems like a bit of a code smell. The hidden behavior for an empty gene list could have untended consequences for consumers of this API. Why can't these be two separate methods?

Luke-Sikina: Or better yet, just do the gene filtration logic outside of this method.

Luke-Sikina (commented):
Mostly just complaining about gene panels and asking you to write some tests.

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
onursumer: I agree that returning `null` a bit confusing. We should return something like `all genes are profiled`, but not sure how to do that with a method that returns a set. This logic is inline with the front end code: https://github.com/cBioPortal/cbioportal-frontend/blob/3a8e6c89e43c1bbec138b36ba6bb62a910f1db44/src/shared/lib/GenePanelUtils.ts#L78-L108. We have the same assumption there. No `genePanelId` means `all genes profiled`. We should try refactoring some gene panel related stuff, but I think it is out of this PR's scope.

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
onursumer: Good point. Seems like we should do null check.

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
onursumer: I agree. We need some refactoring effort for these kind of things. But I think this is out of this PR's scope.

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
Luke-Sikina: Ok, lets at least put this in the method's javadoc. I have to revive my gene panel PR eventually; when I do, I'll chat with you about what we can do to clean this up

web/src/main/java/org/cbioportal/web/util/AlterationCountUtil.java
Luke-Sikina: I think we can add it to sonar. I do like the autonomy of doing it myself though. Have you considered just adding a ruler to your IDE?

onursumer:
Looks like changes introduced by https://github.com/cBioPortal/cbioportal/commit/ad8f6f5ae59afd4ffe35b781236c447af7498a76 causing discrepancy between the frontend and the backend when calculating the number of profiled cases for multiple studies.

We can solve the issue by reverting it back and use `casesWithoutPanelData`, but we probably should do it when the method is called only for this endpoint.

PR #9767, opened 2022-08-24
Add the base path to controller redirects

Luke-Sikina:
- Regex for base path
- Tests

### DIFF
--- a/web/src/main/java/org/cbioportal/web/DocRedirectController.java
+++ b/web/src/main/java/org/cbioportal/web/DocRedirectController.java
@@ -7,6 +7,8 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes;
 import org.springframework.web.servlet.view.RedirectView;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/cbioportal/web/DocRedirectController.java
```diff
@@ -23,7 +28,8 @@ public RedirectView docRedirect(
         HttpServletRequest request
     ) {
         redirectAttributes.addAttribute("group", group);
-        return new RedirectView("/api/v2/api-docs");
+        Matcher m = docsPattern.matcher(request.getRequestURI());
```
onursumer, line 31: should we add a method like this?
```Java
String findBaseURI(Pattern pattern, HttpServletRequest request) {
     Matcher m = pattern.matcher(request.getRequestURI());
     return m.find() ? m.group(1) : "";
}
```

then we can call it like

```Java
    return new RedirectView(findBaseURI(uiPattern, request) + ...)
```

onursumer (approved):
:raised_hands:

PR #9768, opened 2022-08-25
Fix another XSS

Luke-Sikina:
- Springfox doesn't have a fix, so we're just going to block that url param

Note: `configUrl` is used to point to another origin of the [api-docs](https://www.cbioportal.org/api/v2/api-docs?group=default). We don't use it so fine to disable


alisman (approved):

inodb (approved):
LGTM - thank you!

PR #9776, opened 2022-08-29
Update resource file formats

tmazor:
data import fails when these fields are not present, so updating the docs to show them as required


dippindots (approved):
Thanks @tmazor, I created another issue as a follow-up of this. This is a great temporary solution, but I still want to investigate and change it to meet our original design. Here is the new issue: https://github.com/cBioPortal/cbioportal/issues/9781.

dippindots:
@tmazor I will merge this now. I may have time next week, I will put this in the next week's sprint.

PR #9791, opened 2022-09-06
Update minor version description

dippindots:


inodb (approved):
LGTM - thanks!

I think we prolly have to work out the transcript change example a bit, but we can maybe elaborate in the description of the PR itself

PR #9793, opened 2022-09-06
Always load frontend in https when localdev is true

alisman:
This prevents us from having to start webpack dev server in ssl and not ssl to test against local instance vs live instance

### DIFF
--- a/portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
+++ b/portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
@@ -98,10 +98,13 @@ if (/localdist=true/.test(window.location.href)) {
 }
 window.localdist = localStorage.getItem("localdist") === "true";
 
-if (window.localdist || window.localdev) {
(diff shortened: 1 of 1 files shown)


portal/src/main/webapp/WEB-INF/jsp/global/frontend_config.jsp
```diff
@@ -98,10 +98,13 @@ if (/localdist=true/.test(window.location.href)) {
 }
 window.localdist = localStorage.getItem("localdist") === "true";
 
-if (window.localdist || window.localdev) {
-	window.frontendConfig.frontendUrl = '//localhost:3000/'
+if (window.localdist) {
+    window.frontendConfig.frontendUrl = '//localhost:3000/'
```
inodb, line 102: do we still need ssl for `localdist`? I thought that one also always goes over ssl these days:

https://github.com/cBioPortal/cbioportal-frontend/blob/master/scripts/serve_dist.sh#L20-L27

inodb (approved):
Thanks!

PR #9799, opened 2022-09-12
Fix errors when importing data with empty Start_Position or End_Position

oplantalech:
Continuation of #9047, finishing the PR adding empty spaces as allowed values.

> The current [file format description for mutation data](https://docs.cbioportal.org/5.1-data-loading/data-loading/file-formats#mutation-data) lists Start_Position and End_Position as optional attributes.
> If those columns are not present at all the import of data without start/end position works just fine. However, if these columns are present in the header and are left empty you will get an error message stating that '' is not an integer value.
> By looking at the source code of the validator I identified NA as an alternative option. The usage of NA however leads to an error message in the checkAlleleMAFFormat function where it tries to cast NA to a float.
> 
> Therefore I propose to go for an empty string for Start_Position and End_Position as this successfully passes the validator and is much more intuitive.


inodb (approved):
Thanks so much @oplantalech !

PR #9803, opened 2022-09-15
If invalid Entrez gene id in SV file replace id with null

mandawilson:
If a SV record has an invalid Entrez gene id log a warning and replace the invalid id with NULL.  If the record's other gene is valid, the record is inserted into the database.

Previously what happened was the invalid Entrez gene id was left in the record and then when the record was inserted into the database there was a foreign key exception.


averyniceday (approved):
great!

PR #9854, opened 2022-10-13
Validator: clarify SV error

oplantalech:
Change this SV error message:
```
No Entrez gene id or gene symbol provided for site 1 and site 2. This record will not be loaded
```

By:
```
No Entrez gene id or gene symbol provided for site 1 and site 2
```

Reason: avoid confusion with warnings that contain the message `This record will not be loaded` ([example](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/scripts/importer/validateData.py#L917)).


inodb (approved):
Thanks! LGTM!

PR #9858, opened 2022-10-20
Raise warning when all Site 1 and Site 2 SV Entrez Gene Ids or Hugo Symbols are not known by cBioPortal

oplantalech:
Raise a warning instead of an error when at least one of Site1 or Site2 have an Entrez Gene Id or a Hugo Symbol but it is not known by the cBioPortal database.

The error is only raised when Site1 and Site2 have not an Entrez Gene Id and a Hugo Symbol.


rmadupuri (approved):

rmadupuri:
Thanks for adding the check @oplantalech. Looks good to me!

PR #9882, opened 2022-11-09
Add validation for the new CNA Long Format (with custom namespace columns)

oplantalech:
Validation part for #9847


core/src/test/scripts/unit_tests_validate_data.py
```diff
@@ -1012,7 +1013,198 @@ def test_missing_row_gsva(self):
...
+        self.assertEqual(1, record_list[1].line_number)
+    def test_missing_value_column(self):
+        """Test when a file do not have the Value column."""
+        record_list = self.validate('data_cna_long_genecol_no_value.txt',
+                                    validateData.CNADiscreteLongValidator,
+                                    extra_meta_fields={'namespaces': 'MyNamespace'})
+        # two errors after the info: the first makes the file unparsable
+        self.assertEqual(3, len(record_list))
+        for record in record_list[1:]:
+            self.assertEqual(logging.ERROR, record.levelno)
```
dippindots, line 1084: I can see that there are some tests that check error messages. It's better to apply it to all tests that expect error-level logs.

dippindots (commented):
@oplantalech Thanks for implementing this, overall it looks good to me, just have some minor comments, please comment below.
It's also good to have documentation in [File-Format file](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md) to give everyone an idea of how to use this new file format.

core/src/main/scripts/importer/validateData.py
```diff
@@ -1392,6 +1393,149 @@ def printDataInvalidStatement(self, value, col_index):
...
+
+    def checkHeader(self, cols):
+        num_errors = super(CNADiscreteLongValidator, self).checkHeader(cols)
+        if ('Hugo_Symbol' not in cols and
+                'Entrez_Gene_Id' not in cols):
+            self.logger.error('Hugo_Symbol or Entrez_Gene_Id column needs to be present in the file.',
+                              extra={'line_number': self.line_number})
+            num_errors += 1
+
+        return num_errors
```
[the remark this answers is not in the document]
oplantalech: They cannot be added to `REQUIRED_HEADERS` because at least only one of them is required, but not both.

oplantalech:
@dippindots Regarding when to fix the File-Formats page, I've discussed it with @pvannierop and we thing it is better to have it done after the functionality has been merged and it is working.

core/src/main/scripts/importer/validateData.py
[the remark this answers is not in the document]
dippindots: I see, thanks

core/src/main/scripts/importer/validateData.py
```diff
@@ -1392,6 +1393,151 @@ def printDataInvalidStatement(self, value, col_index):
...
+        'Entrez_Gene_Id',
+        'cbp_driver',
+        'cbp_driver_annotation',
+        'cbp_driver_tiers',
+        'cbp_driver_tiers_annotation'
+    ]
+    REQUIRE_COLUMN_ORDER = False
+    ALLOW_BLANKS = True
+
+    ALLOWED_CNA_VALUES = ['-2', '-1.5', '-1', '0', '1', '2']
```
dippindots, line 1436: `ALLOWED_CNA_VALUES` should includes `NULL_VALUES`, right? Because you are checking values in 1528, and `NA` is also an allowed value.

dippindots: It's also better to have some `NA` value in your test files, in that case, we can cover all possible allowed values.

dippindots (commented):
Thanks for the quick update, I have two more comments, after we fixed these two, I think we can merge it.

core/src/main/scripts/importer/validateData.py
oplantalech: Thank you for rechecking this. I have added a test and made a fix to reallow `NA` values.

dippindots (approved):
@oplantalech It looks good to me, thanks for fixing all the comments.

PR #9884, opened 2022-11-09
Edit pd annot validation: remove None element from entrez ids set 

### LINKED ISSUE #9883: Validating CNA Annotations: 'None' not present in CNA data
## The issue
During the validation of CNA pd annotation files, the validator throws a confusing error when the Hugo Symbol and Entrez gene ids are present but don't match.

When it happens, the error is:
```
Following Entrez_Gene_Id are present in pd annotation file, but not in CNA data [None]
```

However, for other data formats, when the Hugo symbol and Entrez gene ids don't match, the validation throws instead the following warning:
```
Entrez gene id and gene symbol do not match. The gene symbol will be ignored. Might be wrong mapping or recycled gene symbol.
```

## The wanted behaviour
When the Hugo and Entrez gene ids don't match, get a warning during the validation similar to other files. And allow the data to be loaded.

## The changes
Remove the `None` value from the `entrez_gene_ids` set (from PD annot) before it's compared to the set of ids of the CNA file.

egarcialara:
Fix #9883 

- Remove the None value from the entrez_gene_ids set (from PD annot) before it's compared to the set of ids of the CNA file.

# Checks
- [x] Runs on heroku


inodb (approved):
Thanks!

Issue #9829, opened 2022-09-29
docker compose breaks on macbook m1


inodb:
Getting some osx thread issues and platform not supported on latest macbook m1. Had to create a `docker-compose.override.yml` like this:

```
version: '3'

services:
  cbioportal:
    image: cbioportal/cbioportal:5.2.4
  cbioportal-database:
    image: biarms/mysql:5.7
  cbioportal-session:
    image: cbioportal/session-service:0.6.0
  cbioportal-session-database:
    image: mongo:4.2
```

EDIT: Updated the above with pre-released versions of cbioportal and session service

ndfriedman:
@inodb how should I add these changes?  Replace the original "docker-compose.yml" with this file?  Or add this file to the docker-compose file and point to it?  Or replace the specific lines above inside the original "docker-compose.yml"?

inodb:
@ndfriedman you add a file in the same folder called `docker-compose.override.yml`. This currently seems to work:

```
version: '3'

services:
  cbioportal:
    image: cbioportal/cbioportal:demo-support-multi-platform-images
  cbioportal-database:
    image: biarms/mysql:5.7
  cbioportal-session:
    image: cbioportal/session-service:test-arm64
  cbioportal-session-database:
    image: mongo:4.2
```

inodb:
I fixed this in: https://github.com/cBioPortal/cbioportal-docker-compose/commit/80bafc33eb1a3788f938dba49c661530a9919323

PR #9890, opened 2022-11-11
Docker image upgrade (eclipse-temurin-11) and arm64 support

zhx828:
This is the effort to make all apps running on both amd64 cpu and arm64 cpu. This is related to https://github.com/oncokb/oncokb/issues/3271

**Changes included in this PR:**
- Use docker/build-push-action to build multi-platform images. We support amd64 and arm64.
- Update default docker JDK/MavenJDK to eclipse-temurin-11. OpenJDK 11 is already retired https://togithub.com/docker-library/openjdk/issues/505

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -31,3 +31,8 @@ trim_trailing_whitespace = false
 # Formats for HTML files only
 [**.html]
 insert_final_newline = false
+
+[{*.yml, *.yaml}]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -13,13 +13,14 @@
 #
 # WARNING: the shendoah image is a nightly, untested, experimental build. If
 # you want to use an official openjdk image instead use the web-and-data image.
-FROM maven:3-openjdk-11 as build
+FROM maven:3-eclipse-temurin-11 as build
 COPY $PWD /cbioportal
 WORKDIR /cbioportal
 ARG MAVEN_OPTS=-DskipTests
-RUN mvn ${MAVEN_OPTS} clean install
+# Use quite output so we can see the build steps easily
+RUN mvn ${MAVEN_OPTS} clean install -q
 
-FROM shipilev/openjdk:11
+FROM eclipse-temurin:11
 
 ENV PORTAL_WEB_HOME=/cbioportal-webapp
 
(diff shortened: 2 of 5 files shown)


.editorconfig
```diff
@@ -31,3 +31,8 @@ trim_trailing_whitespace = false
 # Formats for HTML files only
 [**.html]
 insert_final_newline = false
+
+[{*.yml, *.yaml}]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
```
zhx828, line 38: I hope this is ok. Seems to be the style we use fro yaml files.

docker/web/Dockerfile
```diff
@@ -29,4 +30,4 @@ COPY --from=build /cbioportal/portal/target/dependency/webapp-runner.jar /webapp
 RUN mkdir -p $PORTAL_WEB_HOME
 RUN cd $PORTAL_WEB_HOME && jar -xvf /app.war
 
-CMD ["/bin/sh", "-c", "/usr/bin/java ${JAVA_OPTS} -jar /webapp-runner.jar ${WEBAPP_OPTS} ${PORTAL_WEB_HOME}"]
```
zhx828, line 32: `/usr/bin/java` is no longer the default java location. We should let the system decides rather.

inodb:
Thanks so much for doing this @zhx828 ! This is awesome!

One minor comment is that the build is slower without using the caching action:

- web: [~5m](https://github.com/cBioPortal/cbioportal/actions/runs/3429354607/usage) -> [12m](https://github.com/cBioPortal/cbioportal/actions/runs/3451633912/usage)
- web-and-data: [~8m](https://github.com/cBioPortal/cbioportal/actions/runs/3429354607/usage) -> [>25m](https://github.com/cBioPortal/cbioportal/actions/runs/3451633912/usage)

Prolly ok tho if build times are longer while we wait for [whoan's action to support multibuild](https://github.com/whoan/docker-build-with-cache-action/issues/82#issuecomment-1307686793), since multiarch is more important

zhx828:
> 

Yeah, I noticed the significant build time increase. I played around with the few caching options but didn't not seem to help unfortunately.

inodb (approved):
@zhx828 Thanks again! This seems to work for me

I tried it like this with the cbioportal-docker-compose setup:

https://github.com/cBioPortal/cbioportal/issues/9829#issuecomment-1332767653

I upgraded the session-service image as well to use the same action (just java 11 upgrade didn't work seamlessly there)

Issue #9702, opened 2022-07-22
Clinical Data Tab Empty for MSK study


inodb:
https://cbioportal.mskcc.org/study/clinicalData?id=mskimpact_test_june

All table data shows up empty:
![Screen Shot 2022-07-22 at 2 46 42 PM](https://user-images.githubusercontent.com/1334004/180504102-c862f244-1b9c-4598-8bd9-95cee1531fec.png)

Follow up:
This is a chrome browser issue, we need to find a way to better address this.
Currently we can have a warning and ask users to limit the patients/samples size based on `samples x attributes`

pvannierop:
@inodb I have seen this before. It happens with studies with large sample counts. When the download for sample or patient clinical data exceeds ~500Mb Chrome silently drops the response. It shows HTTP code 200 in DevTools, but the frontend code receives _null_. This appears to be a Chrome limitation that is not reported much (but see [here](https://support.google.com/chrome/a/thread/28233217/chrome-80-seems-to-enforce-a-new-limit-on-post-response-size-of-around-300mb-is-this-by-design?hl=en)). Opening Clincal Tab in FireFox worked fine in my case.

pvannierop:
I have been discussing solutions to this problem with a client involving e.g. pagination. There are more area's that break/become unusable when you  process large studies btw.

inodb:
@pvannierop Thanks for letting us know! Do you happen to have some list of other places where this is a problem? Sounds like this would require quite a big rework of management of the clinical data

PR #9896, opened 2022-11-16
Implement /clinical-data-table/fetch endpoint for Study View

pvannierop:
Part of fix: #9702

# Problem 
With studies with large sample counts or a lot of clinical attributes for patients/samples the table in clinical tab on Study View would not show any data on Chrome browsers.

# Cause
When the download for sample or patient clinical data exceeds ~500Mb Chrome silently drops the response. It shows 
HTTP code 200 in DevTools, but the frontend code receives null. This appears to be a Chrome limitation that is not reported much (but see [here](https://support.google.com/chrome/a/thread/28233217/chrome-80-seems-to-enforce-a-new-limit-on-post-response-size-of-around-300mb-is-this-by-design?hl=en)). Because the Clinical Tab downloads __all__ clinical data before displaying, the 500Mb limit is exceeded for large studies and no data is displayed.

# Solution
This PR implements a new `/clinical-data-table/fetch` endpoint (Internal API). This endpoint supports pagination, filtering with a search term, and sorting. As such it prepares for server side data and lazy loading in the frontend.

Optimization that are included:

- Instead of lists with sample identifiers, the endpoint receives the current Study View filter and resolves sample identifiers from that. This bypasses the upload of large amounts of (compressed) data in the request body.
-  The endpoint returns sample and patient level clinical attributes in the same request. For this, a ClinicalDataCollection model is introduced.

# Note on filtering and pagination
Pagination, sorting and filtering act on the level of samples of which data is returned (rows in the clinical tab), not on the individual clinical data objects returned by the endpoint.


web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -838,4 +841,69 @@ public ResponseEntity<List<GenericAssayDataBin>> fetchGenericAssayDataBinCounts(
...
+            sortBy != null ? sortBy.getOriginalValue() : null,
+            direction.name());
+            
+        // Return empty when possible.
+        if (sampleClinicalData.isEmpty()) {
+            return new ResponseEntity<>(new ClinicalDataCollection(), HttpStatus.OK);
+        }
+
+        // Resolve for which patient clinical data should be included.
+        final List<ImmutablePair<String, String>> patientIdentifiers = sampleClinicalData.stream()
```
Luke-Sikina, line 891: Why are you explicitly declaring function scoped, effectively final variables as final? Not really a problem, just curious.

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -838,4 +841,69 @@ public ResponseEntity<List<GenericAssayDataBin>> fetchGenericAssayDataBinCounts(
...
+            .map(d -> new ImmutablePair<>(d.getStudyId(), d.getPatientId()))
+            .distinct()
+            .collect(Collectors.toList());
+        List<String> patientStudyIds = patientIdentifiers.stream().map(p -> p.getLeft()).collect(Collectors.toList());
+        List<String> patientIds = patientIdentifiers.stream().map(p -> p.getRight()).collect(Collectors.toList());
+        
+        
+        List<String> searchAllAttributes = null;
```
Luke-Sikina, line 899: This is just to make the next line more readable?

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -37,16 +36,22 @@
...
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @InternalApi
 @RestController
 @Validated
 @Api(tags = "Study View", description = " ")
 public class StudyViewController {
+
+    public static final int CLINICAL_TAB_MAX_PAGE_SIZE = 1000000;
```
Luke-Sikina, line 53: Curious why you're making your own constant instead of using `MAX_PAGE_SIZE`

persistence/persistence-mybatis/src/main/resources/org/cbioportal/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -214,4 +221,37 @@
         GROUP BY clinical_patient.ATTR_ID, clinical_patient.ATTR_VALUE
     </select>
 
+    <select id="getSampleClinicalDataClinicalTable" resultType="org.cbioportal.model.ClinicalData">
+        SELECT
+        <include refid="selectSample">
+            <property name="prefix" value=""/>
+        </include>
+        <include refid="fromSample"/>
+        <include refid="whereSampleSearchAndPagination"/>
```
Luke-Sikina, line 230: Huh - I wouldn't have thought to do it this way. I like how this allows you to reuse the select / from logic.

Luke-Sikina (approved):

PR #9903, opened 2022-11-21
Move maf util to own artifact

Luke-Sikina:
We move MAFUtil to its own artifact so that we can import it without importing all of core


Luke-Sikina:
dup

PR #9906, opened 2022-11-22
Move maf util to own artifact

Luke-Sikina:


pom.xml
```diff
@@ -331,6 +331,7 @@
     <module>db-scripts</module>
     <module>security</module>
     <module>utils</module>
+      <module>maf</module>
```
onursumer, line 334: Indentation seems off here

maf/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>master</artifactId>
+        <groupId>org.mskcc.cbio</groupId>
+        <version>0-unknown-version-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>maf</artifactId>
+    <name>Portal utils</name>
+    <description>Utility classes used in different modules</description>
```
onursumer, line 13: We should probably have MAF specific name and description here

maf/pom.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+    <project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>master</artifactId>
+        <groupId>org.mskcc.cbio</groupId>
+        <version>0-unknown-version-SNAPSHOT</version>
```
BasLee, line 8: Is there a reason to name a version like this?

onursumer (approved):

maf/pom.xml
inodb: good question: we have some maven plugin that automatically updates all the version labels based on the github tag/commits. This is just a place holder. I picked that name at some point, should've prolly used something like `0-this-version-is-automtatically-set-by-plugin-x-SNAPSHOT`, might be good to still change

PR #9909, opened 2022-11-28
Update documentation on test logging

BasLee:
Add some documentation on how to configure logging during tests.


Luke-Sikina (commented):
I don't think this is correct; we should be using logback for our test logging now. Can you rebase this PR and try following the instructions to see if you can configure test output using slf4j.SimpleLogger? I would expect there to be no change in log output.

pvannierop (changes requested):
Please update after merge of PR #9949 that will implement logback logging for all Maven modules.

pvannierop:
[Logging during tests: improve documentation](https://trello.com/c/yCYO05Zi/11-logging-during-tests-improve-documentation)

PR #9912, opened 2022-12-01
Add properties to show selected columns in mutations table on init

oplantalech:
An organization has added 150 custom namespace columns to the MAF file for studies. They want to visualize two of these are  immediately when a user opens Patient View, together with `Gene` and `Protein Change`, while the rest of the columns are hidden in the Columns selection menu.

To do so, we propose to introduce two properties that will allow to define the columns that are shown on initialization for the mutation table, one for the results view and the other for the patient view. If the property is not defined, we keep the default columns that are currently shown in the view. Also, the property works together with the `skin.mutation_table.namespace_column.show_by_default` property; if it is set to true, it will show all the namespace columns together with the other defined columns in the properties added in this PR.

### DIFF
--- a/docs/deployment/customization/portal.properties-Reference.md
+++ b/docs/deployment/customization/portal.properties-Reference.md
@@ -191,6 +191,15 @@ Namespace columns are custom columns in the MAF file that can be shown in Mutati
 skin.mutation_table.namespace_column.show_by_default=
 ```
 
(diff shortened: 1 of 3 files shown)


pvannierop (approved):
Thanks, this PR is needed to use a frontend feature that is already merged in master.

PR #9913, opened 2022-12-02
Implementation of RFC64: 'mutational signatures' in the backend

MatthijsPon:
**Implementation of RFC#64 in the backend**

- Addition of a standalone scripts which takes in a study path and generates mutational signature files following RFC64.
- Mutational matrices are constructed from a maf file in the given study path.
- Create mutational signature contributions and P-values per COSMIC signature using MSKCC's tempoSig. 

TODO's:
- Possible implementation of other mutational signature algorithms
- Convert count matrices to generic assay files and create accompanying meta files
- No test created yet, validateData.py needs to be extended

# Checks
- [ ] Runs on heroku


core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    """
+    with open(file_name, "r") as infile:
+        data = infile.readlines()
+    data[0] = "\t".join(data[0].split("\t")[1:])
+    with open(file_name, "w") as outfile:
+        outfile.writelines(data)
+    return None
+
+
+def run_temposig(temposig_loc, inf_matrix, outf, cosmic=2, nperm=1000):
```
SRodenburg, line 242: instead of `cosmic` i'd suggest to call it `version`

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    :param inf_matrix: str, path to input mutational matrix
+    :param outf: str, path to output file
+    :param cosmic: int, cosmic version to use (2, 3)
+    :param nperm: int, number of permutations for p-value estimation (default: 1000)
+    :return: None
+    """
+    command = f"{temposig_loc} --cosmic_v{cosmic} --pvalue --nperm {nperm} --cbio {inf_matrix} " \
+              f"{outf}_contribution_v{cosmic}.txt --pv.out {outf}_pvalue_v{cosmic}.txt"
+    subprocess.run(command, shell=True)
+    return None
```
SRodenburg, line 255: .

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    print("-" * 60)
+    if algorithm == "tempoSig":
+        algorithm_string = f"{algorithm} and COSMIC version {args.cosmic}"
+        if not args.temposig_loc:
+            temposig_loc = check_install_temposig()
+        else:
+            temposig_loc = args.temposig_loc
+        # TODO check if TempoSig can handle DNS and INDEL
+        # These substitution matrix filenames are consistent
+        for substitution_matrix, matrix_type in [("output/SBS/tmp.SBS96.all", "SNS")]:
```
SRodenburg, line 278: OK - to be extended with DNS and INDEL

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    out_lines = f"cancer_study_identifier: {study_id}\n" \
+                f"genetic_alteration_type: GENERIC_ASSAY\n" \
+                f"generic_assay_type: MUTATIONAL_SIGNATURE\n" \
+                f"datatype: LIMIT-VALUE\n" \
+                f"stable_id: {meta_file}\n" \
+                f"profile_name: {profile_name}\n" \
+                f"profile_description: profile for {file_type} value of mutational signatures, generated using " \
+                f"{algorithm_string}\n" \
+                f"data_filename: data_{meta_file}.txt\n" \
+                f"show_profile_in_analysis_tab: {'false' if file_type != 'contribution' else 'true'}\n" \
```
SRodenburg, line 313:
```suggestion
                f"show_profile_in_analysis_tab: {str(file_type == 'contribution').lower()}\n" \
```

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    # Create matrix
+    run_matrix_generator(genome_build, args.tmp_dir)
+    # TODO keep the relevant matrices and port them into a cBioPortal format for visualisation
+    # Run desired algorithm (and possible pre-processing) for mutational signatures
+    matrices_ran, algorithm_string = wrapper_mut_sig(args.algorithm, args.tmp_dir, args.out_dir, args)
+    # Create associated meta files
+    wrapper_meta_files(args, matrices_ran, study_id, algorithm_string)
+    # cleanup temporary files
+    remove_tmp_files(args.tmp_dir)
+    return None
```
SRodenburg, line 365: .

SRodenburg (changes requested):
Very nice start, I proposed some minor restructuring to make the code more intuitive.

I feel like we should move the temposig outside of the script, and let the user give an installation directory. 

Another thing, I discourage the removal of directories in the script. Just make clear in the names of the generated tmp files that they are temporary.

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+
+    :param file_name: str, name of matrix file to alter
+    :return: None, file is altered on host machine
+    """
+    with open(file_name, "r") as infile:
+        data = infile.readlines()
+    data[0] = "\t".join(data[0].split("\t")[1:])
+    with open(file_name, "w") as outfile:
+        outfile.writelines(data)
+    return None
```
egarcialara, line 239: Agree with comment above
```suggestion
    df = pd.read_csv(file_name, sep="\t", comment="#", dtype="str")
    df = df.iloc[1: , :]
    df.to_csv(file_name, sep="\t", index=False, na_rep="")
```

SRodenburg: not sure about the `na_rep=""` but we will find out

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    
+    :param file: object, open yaml file 
+    :return: dict, yaml key-value pairs
+    """
+    yaml_cont = {}
+    for line in file:
+        split_line = line.strip().split(":")
+        key = split_line[0].strip()
+        val = "".join(split_line[1:]).strip()
+        yaml_cont[key] = val
```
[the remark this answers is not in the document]
egarcialara: Why not `metadata = yaml.full_load(open(file))` ?

egarcialara, line 68: Why not:
```suggestion
    yaml_cont = yaml.full_load(open(file))
```
? It would require importing yaml.

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+        raise ValueError(f"There are either no or multiple genome builds present in your MAF file: \'{genome}\'")
+    try:
+        genome = int(genome[0])
+    except ValueError:
+        try:
+            genome = int(genome[0][-2:])
+        except ValueError:
+            genome = input("Can't decipher genome build. Please enter manually (37/38): ")
+            genome = int(genome)
+    genome = "GRCh" + str(genome)
```
egarcialara, line 123: Following the comment from Sander, could this be made with a mapping of accepted possibilities?
`{'19': 'GRCh37', 'hg19': 'GRCh37', ...}`
And finally
`if not ncbi_build.lower() in ['grch37', 'grch38']: ERROR`
The current code would accept e.g. GRCh25

egarcialara: I took the name `ncbi_build` from `validateData.py`, and maybe it's good to keep it consistent?

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    :param tmp_dir: str, path to temp folder
+    :return: path to pre-processed MAF, genome build
+    """
+
+    # Create temporary directory and copy file(s)
+    try:
+        os.mkdir(tmp_dir)
+    except FileExistsError:
+        answer = input(f"\nDirectory \'{tmp_dir}\' already exists. Remove directory (including contents) "
+                       f"and continue script? (y/n): ")
```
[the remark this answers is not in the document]
egarcialara: I agree with Sander. We seldom have scripts that ask for an input once it's running, and imo that's better not to forget running processes. If really necessary, it can be added as another flag.

core/src/main/scripts/mutationalSignatures.py
SRodenburg: some meta files turned out to contain extra colon chars in the description... 
Normally I would also opt for `yaml.full_load` but this is more robust

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    # TODO select which alternative allele?
+    alt_al = "Tumor_Seq_Allele2"
+    headers = ["Chromosome", "Start_Position", "End_Position", "Reference_Allele", alt_al, "Tumor_Sample_Barcode"]
+    maf = maf[headers]
+    # Remap maf file using important parts, as cbioportal MAF format does not conform to
+    #  GFC MAF file format (column "Consequence" is not placed on the correct location).
+    new_headers = ["NA" for i in range(16)]
+    for i, j in enumerate([4, 5, 6, 10, 12, 15]):
+        new_headers[j] = headers[i]
+    maf = maf.reindex(new_headers, axis=1)
```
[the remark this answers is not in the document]
MatthijsPon: The filling of columns is done because sigprofilermatrixgenerator takes columns by index, not by header. So the empty columns are needed. Alternative allele is something we discussed to determine automatically, because sometimes Tumor_Seq_Allele1 contains the variant. This still has to be discussed/implemented.

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    # Fill empty cells with NA string, otherwise they get parsed out by SigProfilerMatrixGenerator
+    maf = maf.fillna("NA")
```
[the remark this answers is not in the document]
MatthijsPon: Yes, but these are normally required columns in MAF files, so they cannot be empty. However, they are unnecessary for matrix generation. Thus they are without value. The way sigprofilermatrixgenerator parses MAF files, empty strings surrounded by tab characters are not seen as a column, thus a value needs to be present.

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: I'll rename it to cosmic_version, as not to confuse it with the TempoSig version.

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+    """Generate mutational signatures using tempoSig
+    
+    :param temposig_loc: str, path to tempoSig.R script
+    :param inf_matrix: str, path to input mutational matrix
+    :param outf: str, path to output file
+    :param cosmic: int, cosmic version to use (2, 3)
+    :param nperm: int, number of permutations for p-value estimation (default: 1000)
+    :return: None
+    """
+    command = f"{temposig_loc} --cosmic_v{cosmic} --pvalue --nperm {nperm} --cbio {inf_matrix} " \
```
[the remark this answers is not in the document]
MatthijsPon: What do you mean with making temposig external? Not performing the installation? Or also letting the user run temposig themselves, instead of from the script?

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: This does not work, since only the first value of the first row has to be removed, the rest of the first column has to be left in the file.  df = df.iloc[: , 1:] removes the whole first column. If you have other more efficient solutions, let me know :)

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+
+def check_install_temposig():
+    """Call R to check if tempoSig is installed and where, otherwise install using R
+    
+    :return: str, location of tempoSig.R script
+    """
+    temposig_location = robjects.r("system.file('exec', 'tempoSig.R', package = 'tempoSig')")[0]
+    if not temposig_location:
+        install_temposig()
+        temposig_location = robjects.r("system.file('exec', 'tempoSig.R', package = 'tempoSig')")[0]
```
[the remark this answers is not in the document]
MatthijsPon: If you suggest letting the user supply the installation directory, this whole function can be removed. Thoughts?

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: Nice, good catch! I will make a mapper! Do you also suggest removing the input option and adding a genome cml parameter? as to keep the program without user input?

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def generate_meta_file(study_id, meta_file, file_type, algorithm_string, out_dir):
+    """"""
+    profile_name = " ".join(meta_file.split("_"))
+    # Meta files follow structure of Gaofei Zhao mutational signatures data format example:
+    # https://github.com/cBioPortal/cbioportal-frontend/tree/master/end-to-end-test/local/studies/lgg_ucsf_2014_test_generic_assay
+    out_lines = f"cancer_study_identifier: {study_id}\n" \
+                f"genetic_alteration_type: GENERIC_ASSAY\n" \
+                f"generic_assay_type: MUTATIONAL_SIGNATURE\n" \
```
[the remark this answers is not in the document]
MatthijsPon: I also think this would be better, however, Gaofei Zhao of MSKCC put this value in his example files for mutational signatures

pvannierop:
[Mutational Signatures Backend](https://trello.com/c/XshVbHzO/12-mutational-signatures-backend)

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+            return "Tumor_Seq_Allele1"
+        elif row["Tumor_Seq_Allele2"] != row["Reference_Allele"] \
+            and regexp.fullmatch(row["Tumor_Seq_Allele2"]):
+            return "Tumor_Seq_Allele2"
+    raise ValueError("Alternative Allele in maf file could not be determined. "
+                     "Please use the --alt-allele parameter.")
+
+
+def preprocess_maf(mut_file: str, tmp_dir: str, args: argparse.Namespace) -> tuple:
+    """Preprocess the cBioPortal mutations data file for matrix generation
```
SRodenburg, line 245: Perhaps slightly more detail on what preprocessing involves, or why it's done

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    :param tmp_dir: path to temp folder
+    :param args: parsed argparse.ArgumentParser
+    :return: path to preprocessed MAF, NCBI build
+    """
+    maf = pd.read_csv(mut_file, sep="\t", comment="#", dtype=str)
+    # Get NCBI build from MAF file
+    if not args.ncbi_build:
+        ncbi_build = get_ncbi_build(maf)
+    else:
+        ncbi_build = args.ncbi_build
```
SRodenburg, line 257:
```suggestion
    ncbi_build = getattr(args, 'ncbi_build', get_ncbi_build(maf))
```

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    print(f"Using ncbi build: {ncbi_build}.")
+    # Determine alt allele
+    alt_al = args.alt_allele
+    if not alt_al:
+        alt_al = determine_alternative_allele(maf)
```
SRodenburg, line 262:
```suggestion
    alt_al = getattr(args, 'alt_allele', determine_alternative_allele(maf))
```

core/src/main/scripts/mutationalSignatures.py
SRodenburg: matter of preference, I like compact programming better. Feel free to maintain your own style

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    print(f"Using '{alt_al}' as alternative allele.")
+    # Remap maf file using important parts, as cbioportal MAF format does not conform to
+    #  GFC MAF file format (column "Consequence" is not placed on the correct location).
+    headers = ["Chromosome", "Start_Position", "End_Position", "Reference_Allele", alt_al, "Tumor_Sample_Barcode"]
+    maf = maf[headers]
+    new_headers = ["NA" for i in range(16)]
+    # SigProfilerMatrixGenerator selects alternative allele by location, so the header doesn't matter
+    for i, j in enumerate([4, 5, 6, 10, 12, 15]):  
+        new_headers[j] = headers[i]
+    maf = maf.reindex(new_headers, axis=1)
```
SRodenburg, line 272: I don't understand what you're trying to do here

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    # File needs the MAF file extension, otherwise it is not recognized by SigProfilerMatrixGenerator
+    outfile = osp.join(tmp_dir, "mutations.maf")
+    # Fill empty cells with NA string, otherwise they get parsed incorrectly by SigProfilerMatrixGenerator
+    maf.to_csv(outfile, sep="\t", index=False, na_rep="NA")
+    return outfile, ncbi_build
+
+
+def preprocess_vcf(args: argparse.Namespace, tmp_dir):
+    """Preprocess vcf files by copying them into the tmp dir
```
SRodenburg, line 281: if only copying, it's not really preprocessing :)

SRodenburg: would it be good to add a check that all filenames match a sample ID?

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+def prepare_input(args: argparse.Namespace) -> tuple:
+    """Wrapper function for preparing input files for matrix generation
+    
+    :param args: argparse parsed arguments 
+    :return: ncbi_build, study ID
+    """
+    # Init variables
+    maf_file, study_id, ncbi_build = [""] * 3
```
SRodenburg, line 298: Use `None` instead of empty string

```suggestion
    maf_file, study_id, ncbi_build = (None,) * 3
```

SRodenburg: also, `maf_file` is created in next block, so it's obsolete to init it here

SRodenburg: and `study_id` is not used

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    replace = {
+        ">": "-", ":": "_", "[": "_", "]": "_"
+    }
+    new = ["mutational_signatures_matrix_"]
+    for char in string:
+        if replace.get(char):
+            new.append(replace.get(char))
+        else:
+            new.append(char)
+    return "".join(new)
```
SRodenburg, line 358: is easier resolved using regex
```suggestion
import re

string = re.sub(r'>', '-', string)
string = re.sub(r'[\[\]:]', '_', string)
return string
```

SRodenburg: of course, put the `import re` to the opening of the script

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    :param matrix: path to input matrix
+    :param outfile: path to output file    
+    """
+    data = pd.read_csv(matrix, sep="\t", dtype=str)
+    data = data.rename({"MutationType": "NAME"}, axis=1)
+    data = data.set_index("NAME")
+    data = data.reset_index()
+    data["ENTITY_STABLE_ID"] = data["NAME"].apply(matrix_stable_id) 
+    column_order = ["ENTITY_STABLE_ID", "NAME"] + [x for x in data.columns if x not in ["ENTITY_STABLE_ID", "NAME"]]
+    data = data[column_order]
```
SRodenburg, line 373:
```suggestion
    id_cols = ['ENTITY_STABLE_ID', 'NAME'] 
    data = data[id_cols + data.columns.difference(id_cols).tolist()]
```

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    :param infile: path to input file
+    :param outfile: path to output file
+    :param annotation_file: path to annotation file
+    :param data_type: data type used to generate entity stable IDs
+    :return: path to output file
+    """
+    # read in TS output
+    data = pd.read_csv(infile, sep="\t", dtype=str, comment="#")
+    # Number of mutations column in unnecessary
+    data = data.drop("Number of Mutations", axis=1)
```
SRodenburg, line 420:
```suggestion
    data = data.drop("Number of Mutations", axis='columns')
```

SRodenburg (changes requested):
Nice work Matthijs,

Some code suggestions and a few spots where I think we can improve on clarity, structure or efficiency, as indicated.

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
MatthijsPon: I've tested it with python 3.10.9, the dockerized version of cbio uses 3.10.6

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+import glob
+import os
+import re
+import shutil
+import subprocess
+
+import os.path as osp
+import pandas as pd
+from SigProfilerMatrixGenerator import install as genInstall
+from SigProfilerMatrixGenerator.scripts import SigProfilerMatrixGeneratorFunc as matGen
```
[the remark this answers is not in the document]
MatthijsPon: The same way they install any other package, either via pip install or conda. Should this be added to the cbio image?

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+def determine_alternative_allele(maf: pd.DataFrame) -> str:
+    """Determine which alternative allele to use from MAF file
+    
+    :param maf: maf dataframe
+    :return: alternative allele to use 
+    """
+    regexp = re.compile(r'^[ATGC]*$')
+    # This is normally advised against due to speed, but we need only one good row.
```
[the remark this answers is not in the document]
MatthijsPon: Within cBio TSA1 is preferred over TSA2 when both are filled in and TSA1 differs from REF: https://docs.cbioportal.org/file-formats/#tumor-seq-allele-ambiguity
If a MAF is provided which sometimes puts data in TFA1 and other times in TFA2, its an input data issue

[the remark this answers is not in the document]
SRodenburg: regardless, that issue is fixed in the same way, right?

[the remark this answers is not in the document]
MatthijsPon: No, its not. If we take TSA2 as the one containing the mutations of interest: 
TSA1 may be filled with nucleotides, but be equal to the reference sequence
TSA2 may contain "-" upon deletions
Both of these cases will be incorrectly selected in your case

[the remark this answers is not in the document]
MatthijsPon: Also, the current implementation works, it's just not very pretty. But it does follow cBioPortal convention.
- It prefers TSA1 over TSA2 if both contain different values than reference
- Ignores columns equal to reference
- Skips over "-" values as these might be ambiguous

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+        "mm10": "GRCm38", "10": "GRCm38", "grcm38": "GRCm38"
+    }
+    ncbi_build = ncbi_build_options.get(genome)
+    if ncbi_build is None:
+        raise argparse.ArgumentTypeError(f"Genome: `{genome}` has not been implemented. "
+                                         f"The following genomes are implemented: {set(ncbi_build_options.values())}")
+    return ncbi_build
+        
+        
+def arg_conditionals(args: argparse.Namespace):
```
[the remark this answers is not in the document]
MatthijsPon: To my knowledge, there are no options in `ArgParse` for the conditionals in this function:
- 1 or more of these options is needed
- If option1 is chosen, option2 is required, otherwise not

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    # If using vcf-folder, genome is required
+    if args.vcf_folder and not args.ncbi_build:
+        raise argparse.ArgumentTypeError(f"If using the --vcf-folder argument, the --study-id and "
+                                         f"--genome arguments are required")
+    # Check if at least one of the signature files is given
+    if not (args.sbs_file or args.dbs_file or args.id_file):
+        raise argparse.ArgumentTypeError("At least one of the [-s,-d,-i] arguments is required.")
+
+
+def load_yaml(file: iter) -> dict:
```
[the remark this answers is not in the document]
MatthijsPon: Implemented.

core/src/main/scripts/mutationalSignatures.py
```diff
@@ -0,0 +1,562 @@
...
+    """Find and read mutation data meta file in cBioPortal study folder
+    
+    :param study_path: path to cBioPortal study folder
+    :return: contents of mutations meta file
+    """
+    meta_files = glob.glob(osp.join(study_path, "meta*"))
+    for file in meta_files:
+        with open(file) as f:
+            yaml_cont = load_yaml(f)
+        if yaml_cont.get("stable_id") == "mutations" and yaml_cont.get("datatype") == "MAF":
```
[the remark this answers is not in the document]
MatthijsPon: implemented

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: implemented

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: Oeh, did not know this existed. Love it. However, ncbi_build is in the namespace of args, so getattr() returns None unfortunately

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: I've rewritten the comments around the block. The matrix generator needs columns at specific locations in the MAF file, its not very flexible

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: Renamed to copy_vcf. On the filename part, I think that should be for the user. Otherwise we also have to load in a clinical sample file

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: Changed to [None]. For the variables, I can remove them, but this way the function wont crash if they don't get initialized

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: Nice clean solution, implemented

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: implemented

core/src/main/scripts/mutationalSignatures.py
MatthijsPon: implemented

core/src/main/scripts/mutationalSignatures.py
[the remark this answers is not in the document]
SRodenburg:
```python
try:
  import ...
except:
 install
```

MatthijsPon:
Implementation moved to datahub-study-curation-tools: [PR#48](https://github.com/cBioPortal/datahub-study-curation-tools/pull/48)

PR #9928, opened 2022-12-09
Fix entrezID not changed upon HUGO recognition in structural variants

MatthijsPon:
Fix: SQL fails when inserting structural variant entries where the EntrezID is unknown in the DB, but the HUGO symbol is known. Reason: the corresponding gene is found via HUGO, but the corresponding EntrezID is not updated when adding the variant to the structural variant table. This causes a foreign key error, since an unknown EntrezID is being added.

- Added updating the entrez ID in the structuralVariant object when the ID is different than the one found in the DB.
- Added test entry in (test/resources/) data_structural_variants.txt with unkown Entrez and known HUGO
- Updated test files to accommodate added test entry: TestImportProfileData.java, TestImportStructuralVariantData.java

# Checks
- [ ] Runs on heroku
- [x] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!


inodb (commented):
Thanks! This seems reasonable to me

Just wondering for validation of the data files. We also want to curate those properly for people looking to use the data files directly. I believe that for mutation data we have a flag in genome nexus to override the entrez id. Maybe we need something similar for the validator of SV data? The proposed solution is nice for someone that doesn't want to update their data files, but I think ideally one can opt-in to fix the data files

PR #9956, opened 2023-01-11
Fix genetic_alteration import for cna long format

BasLee:
# Breaking bugs
- CNA long format importer does not add a value to the comma separated list in `genetic_alteration.value` when the CNA event is not present in the data file for a specific gene*sample combination.
- The order of profile values (`genetic_alteration` table) does not match that of samples (`genetic_profile_samples` table). 

# Solution
- Add an empty value for missing observations (e.g. a second missing value would looks like: `2,,`).
- Correct the order of genetic alteration values. A guava HashBasedTable returns the sample columns in a different order when calling `geneSampleTable.row(entrezId).values()` or `geneSampleTable.eventsTable.columnKeySet()`. The latter call returns the samples in the correct order.

# Notes:
- This PR contains regression tests.


pvannierop (approved):
Tested and it fixes the bug!

PR #9961, opened 2023-01-16
Fix import of cna long entries when entrez ID is missing or wrong

BasLee:
# Bugs
1. CNA long format importer does not import an entry when only a hugo ID and no entrez ID is provided.
2. CNA long format importer does not import an entry when a correct hugo ID but a faulty entrez ID is provided.

# Solutions
1. Use the entrez ID that is linked to the provided hugo ID. This entrez ID lookup was already done, and it is now also used when creating a CnaEvent.
2. Try to find gene by entrez ID first, and if nothing found, try to find a gene by its hugo symbol.

# Notes
This PR contains regression tests.


pvannierop (approved):
Thanks for fixing!

dippindots (approved):

PR #9972, opened 2023-01-20
Empty namespace columns should result in NULL instead of {}

BasLee:
Fix the import of missing custom namespace columns for CNA and Struct Var.

At the moment the CNA and Struct Var importers create an empty json object (`{}`) for entries without custom namespace columns. However, entries without custom namespace columns should end up with a `NULL`, just like extended mutation data entries.

This PR also tests the importers with entries missing namespace columns.


dippindots (approved):
@BasLee Thanks for the quick fix, it looks good to me!

PR #10035, opened 2023-02-18
Do not call next study view filter when no samples left

pvannierop:
# Problem
Study View filter Appliers are called on empty lists. This results in errors in some of the Filters.

# Solution
Do not call next filter when not needed.


pvannierop:
[Terminate Study View filter applier when not needed](https://trello.com/c/eSvxT97M/15-terminate-study-view-filter-applier-when-not-needed)

dippindots (approved):
Thanks! LGTM!

PR #10037, opened 2023-02-19
Configure database connection via db.connection_string param

pvannierop:
This update is needed to allow connection parameters required to run MySQL8.

# Problem
In order to run MySQL 8 additional parameters must be provided to the database connection. At present similar parameters (e.g., _db.use_ssl_) are provided via dedicated properties in _portal.properties_ and, by extension, in java class _DatabaseProperties_. 

# Solution
In order to make it easier to provide custom database connection parameters, I propose in this PR to leverage the existing _db.connection_url_ parameter (that is unused as far as I can see) and make this the required way of configuring the connection. The _db.user_ , _db.password_ and _db.driver_ properties will still be used. The _db.host_, _db.database_, _db.port_ and _db.use_ssl_ properties become obsolete.

This PR will also update the python migration script to run migration using the _db.connection_string_ property. The python importer scripts all rely on confguration at the level of the Java code, and do not require update.

# Backward compatibility
This implementation is backwards compatible with configuring with the legacy _db.host_, _db.database_, _db.port_ and _db.use_ssl_ properties.

# Deprecation warning
A deprecation warning was added for the legacy style database connection params:
```
1:41:25.713 [RMI TCP Connection(2)-127.0.0.1] INFO  o.m.c.portal.util.GlobalProperties - Successfully read properties file

----------------------------------------------------------------------------------------------------------------
-- Deprecation warning:
-- You are connecting to the database using the deprecated 'db.host', 'db.portal_db_name' and 'db.use_ssl' properties.
-- Please use the 'db.connection_string' instead (see https://docs.cbioportal.org/deployment/customization/portal.properties-reference/).
----------------------------------------------------------------------------------------------------------------

11:41:25.718 [RMI TCP Connection(2)-127.0.0.1] INFO  o.s.c.s.PostProcessorRegistrationDelegate$BeanPostProcessorChecker - Bean 'businessDataSource' of type [org.mskcc.cbio.portal.dao.JdbcDataSource] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
...
11:41:29.012 [RMI TCP Connection(2)-127.0.0.1] INFO  o.s.web.context.ContextLoader - Root WebApplicationContext initialized in 6066 ms

----------------------------------------------------------------------------------------------------------------
-- You are connecting to the OncoKB public instance which does not include any therapeutic information.
-- Please consider obtaining a license to support future OncoKB development by following https://docs.cbioportal.org/2.4-integration-with-other-webservices/oncokb-data-access.
-- Thank you.
----------------------------------------------------------------------------------------------------------------

<6>Initializing Spring DispatcherServlet 'api'
11:41:29.079 [RMI TCP Connection(2)-127.0.0.1] INFO  o.s.web.servlet.DispatcherServlet - Initializing Servlet 'api'
```

# Verification
**MSK tasks:**
- [x] Java part 
  - [x] (backward compatible) importers
  - [x] (backward compatible) backend spring app database connection 
- [x] (backward compatible) local backend migration script
- [x] deployments ([beta portal](https://beta.cbioportal.org/))
- [ ] internal usage

**The Hyve tasks:**
- [ ] docker and docker-compose
  - [ ] backend database connection
  - [ ] migration script is still working

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
@@ -3,28 +3,44 @@ package org.mskcc.cbio.portal.dao;
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.mskcc.cbio.portal.util.DatabaseProperties;
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.util.Assert;
 
 /**
  * Data source that self-initializes based on cBioPortal configuration.
  */
 public class JdbcDataSource extends BasicDataSource {
+
     public JdbcDataSource () {
         DatabaseProperties dbProperties = DatabaseProperties.getInstance();
+
         String host = dbProperties.getDbHost();
         String userName = dbProperties.getDbUser();
         String password = dbProperties.getDbPassword();
         String mysqlDriverClassName = dbProperties.getDbDriverClassName();
         String database = dbProperties.getDbName();
-        String useSSL = (!StringUtils.isBlank(dbProperties.getDbUseSSL())) ? dbProperties.getDbUseSSL() : "false";
         String enablePooling = (!StringUtils.isBlank(dbProperties.getDbEnablePooling())) ? dbProperties.getDbEnablePooling(): "false";
-        String url ="jdbc:mysql://" + host + "/" + database +
-                        "?user=" + userName + "&password=" + password +
-                        "&zeroDateTimeBehavior=convertToNull&useSSL=" + useSSL;
+        String connectionURL = dbProperties.getConnectionURL();
+        
+        Assert.isTrue(
+            !defined(host) && !defined(database) && !defined(dbProperties.getDbUseSSL()),
+            "\n----------------------------------------------------------------------------------------------------------------" +
+                "-- Connection error:\n" +
+                "-- You try to connect to the database using the deprecated 'db.host', 'db.portal_db_name' and 'db.use_ssl' properties.\n" +
+                "-- Please remove these properties and use the 'db.connection_string' property instead. See https://docs.cbioportal.org/deployment/customization/portal.properties-reference/\n" +
+                "-- for assistance on building a valid connection string.\n" +
+                "----------------------------------------------------------------------------------------------------------------\n"
+        );
+        
+        Assert.hasText(userName, errorMessage("username", "db.user"));
+        Assert.hasText(password, errorMessage("password", "db.password"));
+        Assert.hasText(mysqlDriverClassName, errorMessage("driver class name", "db.driver"));
+
+        this.setUrl(connectionURL);
+
         //  Set up poolable data source
         this.setDriverClassName(mysqlDriverClassName);
         this.setUsername(userName);
         this.setPassword(password);
-        this.setUrl(url);
         // Disable this to avoid caching statements
         this.setPoolPreparedStatements(Boolean.valueOf(enablePooling));
         // these are the values cbioportal has been using in their production
@@ -36,5 +52,13 @@ public class JdbcDataSource extends BasicDataSource {
         this.setTestOnBorrow(true);
         this.setValidationQuery("SELECT 1");
         this.setJmxName("org.cbioportal:DataSource=" + database);
+    }
+    
+    private String errorMessage(String displayName, String propertyName) {
+        return String.format("No %s provided for database connection. Please set '%s' in portal.properties.", displayName, propertyName);
+    }
+    
+    private boolean defined(String property) {
+        return property != null && !property.isEmpty();
     }
 }
--- a/core/src/main/scripts/importer/cbioportalImporter.py
+++ b/core/src/main/scripts/importer/cbioportalImporter.py
@@ -419,8 +419,7 @@ def usage():
                            '--command [%s] --study_directory <path to directory> '
                            '--meta_filename <path to metafile>'
                            '--data_filename <path to datafile>'
-                           '--study_ids <cancer study ids for remove-study command, comma separated>'
-                           '--properties-filename <path to properties file> ' % (COMMANDS)), file=OUTPUT_FILE)
+                           '--study_ids <cancer study ids for remove-study command, comma separated>' % (COMMANDS)), file=OUTPUT_FILE)
 
 def check_args(command):
     if command not in COMMANDS:
--- a/core/src/main/scripts/importer/cbioportal_common.py
+++ b/core/src/main/scripts/importer/cbioportal_common.py
@@ -11,7 +11,9 @@ import csv
 import logging.handlers
 from collections import OrderedDict
 from subprocess import Popen, PIPE, STDOUT
-
+from typing import Dict, Optional
+import dsnparse
+import MySQLdb
 
 # ------------------------------------------------------------------------------
 # globals
@@ -36,6 +38,14 @@ IMPORT_CANCER_TYPE_CLASS = "org.mskcc.cbio.portal.scripts.ImportTypesOfCancers"
 IMPORT_CASE_LIST_CLASS = "org.mskcc.cbio.portal.scripts.ImportSampleList"
 ADD_CASE_LIST_CLASS = "org.mskcc.cbio.portal.scripts.AddCaseList"
 VERSION_UTIL_CLASS = "org.mskcc.cbio.portal.util.VersionUtil"
+
+PORTAL_PROPERTY_DATABASE_USER = 'db.user'
+PORTAL_PROPERTY_DATABASE_PW = 'db.password'
+PORTAL_PROPERTY_DATABASE_HOST = 'db.host'
+PORTAL_PROPERTY_DATABASE_NAME = 'db.portal_db_name'
+PORTAL_PROPERTY_DATABASE_URL = 'db.connection_string'
+PORTAL_PROPERTY_DATABASE_USESSL = 'db.use_ssl'
+REQUIRED_DATABASE_PROPERTIES = [PORTAL_PROPERTY_DATABASE_USER, PORTAL_PROPERTY_DATABASE_PW, PORTAL_PROPERTY_DATABASE_URL]
 
 # provides a key for data types to metafile specification dict.
 class MetaFileTypes(object):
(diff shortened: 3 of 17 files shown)


pvannierop:
[Allow cBioPortal to user MySQL](https://trello.com/c/S3N8YWll/14-allow-cbioportal-to-user-mysql)

core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
```diff
@@ -3,28 +3,45 @@
...
+        Assert.hasText(password, errorMessage("password", "db.password"));
+        
+        // For backward compatibility, build connection URL from individual properties.
+        if (connectionURL == null) {
+            Assert.hasText(host, errorMessage("host", "db.host") + 
+                " Or preferably, set the 'db.connection_string' and remove 'db.host', 'db.database' and 'db.use_ssl' (best practice).");
+            Assert.hasText(database, errorMessage("database", "db.database") + 
+                " Or preferably, set the 'db.connection_string' and remove 'db.host', 'db.database' and 'db.use_ssl (best practice).");
+            connectionURL = String.format(
+                "jdbc:mysql://%s/%s&zeroDateTimeBehavior=convertToNull&useSSL=%s",
```
n1zea144, line 34: @pvannierop This may be because I updated my local JDBC driver to 8.0.31, but I also had to add **allowLoadLocalInfile=true** to allow the bulk loader to work properly.

pvannierop: @n1zea144 The idea of this PR is that when you have the need for these params, you should provide the full connection string with the `d.connection_string` parameter. Remember that `allowLoadLocalInfile=true` is considered as a security risk. Activation as a default is not the way to go in my opinion.

pvannierop:
@inodb  The integration test fails because there is a missing Python dependency. This test will automatically become green when the new docker image is pushed to Dockerhub that has this dependency.

core/src/main/scripts/importer/cbioportalImporter.py
```diff
@@ -419,8 +419,7 @@ def usage():
                            '--command [%s] --study_directory <path to directory> '
                            '--meta_filename <path to metafile>'
                            '--data_filename <path to datafile>'
-                           '--study_ids <cancer study ids for remove-study command, comma separated>'
-                           '--properties-filename <path to properties file> ' % (COMMANDS)), file=OUTPUT_FILE)
+                           '--study_ids <cancer study ids for remove-study command, comma separated>' % (COMMANDS)), file=OUTPUT_FILE)
```
dippindots, line 422: I don't quite understand this. Are we stopped supporting loading the customized property file?
Why do we remove this `'--properties-filename <path to properties file> '`?

core/src/main/scripts/migrate_db.py
```diff
@@ -29,86 +29,63 @@
...
-        
+        url_elements = dsnparse.parse(portal_properties.database_url)
+        connection_kwargs = {
+            "host": url_elements.host,
+            "port": url_elements.port if url_elements.port is not None else 3306,
+            "db": url_elements.paths[0],
+            "user": portal_properties.database_user,
+            "passwd": portal_properties.database_pw,
+            "ssl": "useSSL" not in url_elements.query or url_elements.query["useSSL"] == "true"
+        }
```
dippindots, line 48: I see we parsed `database_url` here to get all DB properties. But it seems like the `database_url` is required and not backward compatible. In the description of this pr, I see `This implementation is backward compatible with configuring with the legacy db.host, db.database, db.port and db.use_ssl properties.`. So I was expecting to see it's fully compatible with current DB properties. Are there any reasons for making `database_url` a required property here?

dippindots (commented):
@pvannierop Thanks for making this pr. The JDBC part looks good to me, and I can see that's backward compatible, we can still use the old property file. But I don't know why don't we also make `migrate_db.py` backward compatible. Maybe I missed some important things, could you help me understand this?

core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
inodb: re `allowLoadLocalInfile=true` - I guess we would apply this just for the importer? Doesn't seem like much of a security risk in that setting

inodb:
I think this might need to be updated too: https://github.com/cBioPortal/cbioportal/blob/master/docker/web-and-data/docker-entrypoint.sh#L27-L33

docker/web-and-data/docker-entrypoint.sh
```diff
@@ -23,19 +23,53 @@ parse_db_params_from_config_and_command_line() {
...
 check_db_connection() {
-    eval "$(parse_db_params_from_config_and_command_line $@ | sed 's/^db\./db_/g')"
-    POTENTIAL_DB_PARAMS=$@
+    eval $(parse_db_params_from_config_and_command_line $@)
+    
+    if ([[ -n $db_host ]] || [[ -n $db_portal_db_name ]] || [[ -n $db_use_ssl ]]) && [[ -n $db_connection_string ]]
+    then
+        echo "The portal.properties file defines both db.connection_string and (one of) db.host, db.portal_db_name and "
+        echo "db.use_ssl. Please configure with either db.connection_string (preferred), or db.host, db.portal_db_name and db.use_ssl."
+        exit 1
```
oplantalech, line 66: We're promising backwards-compatibility but that's not true: if staying with properties as they are right now, the error message in line 64 will show up (since we have both obsolete and `db.connection_string` properties defined).

Removing the obsolete properties and leaving `db.connection_string` will not work, since the value of this property needs to be changed too (from `jdbc:mysql://localhost:3306/` to `jdbc:mysql://localhost:3306/cbioportal?useSSL=false`). 

If the user removes the `db.connection_string` and keeps the obsolete properties as they are in their setup, then the _promised_ backwards compatibility comes into play. But since we want to prioritise using the new setup (`db.connection_string`), and the user needs to modify `portal.properties` anyway, maybe it makes more sense to force to use `db.connection_string` with the correct format and clean up the usage of the obsolete properties?

@inodb What do you think?

oplantalech (changes requested):

core/src/main/scripts/importer/cbioportal_common.py
```diff
@@ -1000,3 +1010,110 @@ def run_java(*args):
...
+
+    try:
+        url_elements = dsnparse.parse(portal_properties.database_url)
+        connection_kwargs = {
+            "host": url_elements.host,
+            "port": url_elements.port if url_elements.port is not None else 3306,
+            "db": url_elements.paths[0],
+            "user": portal_properties.database_user,
+            "passwd": portal_properties.database_pw,
+            "ssl": "useSSL" not in url_elements.query or url_elements.query["useSSL"] == "true"
```
dippindots, line 1108: @pvannierop I tested the local migration script with the new URL parameter, but I got this error: 
```
  File "/Users/zhaog/Desktop/Github/cbioportal/core/src/main/scripts/importer/cbioportal_common.py", line 1108, in get_db_cursor
    "ssl": "useSSL" not in url_elements.query or url_elements.query["useSSL"] == "true"
```
Maybe it should be `url_elements.useSSL == "true"`?

dippindots: But I still get an error about this connection string:
`db.connection_string=jdbc:mysql://localhost:3306/cbioportal?zeroDateTimeBehavior=convertToNull&useSSL=false`
This is the result I get from the output (I hid the username and password)
```
{'host': None, 'port': 3306, 'db': 'mysql:', 'ssl': False}
```
I don't know how could this happen, could you take a look?

dippindots: And db has been parsed as `mysql` seems wrong. Not sure if we can find a better library to do the parsing task?

core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
pvannierop: >I guess we would apply this just for the importer
Correct

core/src/main/scripts/importer/cbioportalImporter.py
pvannierop: >It looks to be not used.
That was also my conclusion.

core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
```diff
@@ -3,28 +3,59 @@
...
+        
+        // For backward compatibility, build connection URL from individual properties.
+        if (connectionURL == null) {
+            Assert.hasText(host, errorMessage("host", "db.host") + 
+                " Or preferably, set the 'db.connection_string' and remove 'db.host', 'db.portal_db_name' and 'db.use_ssl' (best practice).");
+            Assert.hasText(database, errorMessage("database name", "db.portal_db_name") + 
+                " Or preferably, set the 'db.connection_string' and remove 'db.host', 'db.portal_db_name' and 'db.use_ssl (best practice).");
+            System.out.println("\n----------------------------------------------------------------------------------------------------------------");
+            System.out.println("-- Deprecation warning:");
+            System.out.println("-- You are connection to the database using the deprecated 'db.host', 'db.portal_db_name' and 'db.use_ssl' properties.");
```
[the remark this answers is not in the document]
pvannierop: Fixed!

docker/web-and-data/docker-entrypoint.sh
dippindots: How about we don't allow `db.connection_string` shows up with one of 'db.host', 'db.portal_db_name', and 'db.use_ssl' properties? If that's the case, we throw an error and point out that the users should use `db.connection_string` only instead.

dippindots: I know this is probably not 100% backward compatible, but I guess it's fine.

dippindots:
@oplantalech Thanks for help testing the docker part. I also have done my testing for the local build. And it looks good, this is currently pending @sheridancbio's testing on our internal importing code compatibility. I will work with @sheridancbio to finish the testing and then let you know.

@oplantalech Do you have any ideas about the failing tests? I think the `Labeler` test is fine. But We should figure out why `Core tests` and `Integration tests` are failing before we merge this.

dippindots (approved):
@pvannierop It looks all good to me, just the localdb tests are not working, but I guess that's because we haven't updated the [cbioportal-docker-compose](https://github.com/cBioPortal/cbioportal-docker-compose) repository yet. I created a pull request to make the related changes, could you help to take a look? https://github.com/cBioPortal/cbioportal-docker-compose/pull/22/files

I will include this update in the next release. @inodb @pvannierop Just wondering maybe we should update software's minor version to `5.4.0` because based on our [documentation](https://github.com/cBioPortal/cbioportal/blob/master/docs/development/Release-Procedure.md#cbioportal-software-version-numbers), this change might require additional effort for a deployer of cBioPortal

docs/Migration-Guide.md
```diff
@@ -1,22 +1,45 @@
 # Migration Guide
-This page describes various changes deployers will need to make as they deploy newer versions of the portal. 
+
+This page describes various changes deployers will need to make as they deploy newer versions of the portal. -
+
+## <v5.3.19 --> v5.3.20
```
dippindots, line 5:
```suggestion
## <v5.3.19 --> v5.4.0
```
We decided to change it to `v5.4.0`

dippindots (approved):

dippindots:
Merging this pull request. The localdb tests are failing because it's not using the latest URL format. Related pull request:
**frontend**: https://github.com/cBioPortal/cbioportal-frontend/pull/4710 
**cbioportal-docker-compose**: https://github.com/cBioPortal/cbioportal-docker-compose/pull/22

PR #10038, opened 2023-02-20
Update to Add Clinical Event Filtering to backend, and create new api…

### LINKED ISSUE #10000: New filtering chart for clinical event data (longitudinal data)
We want to add a new "chart" to the study view which will allow users to filter samples/patients based on longitudinal data in the clinical events table.  For example, we want to filter for patients who have at least one lab result, or treatment or imaging event.  We call this longitudinal data because it is dated and can be plotted on a timeline (in the patient page).  
The data in clinical events have a type property.   This is the category on which we wish to filter. 

Here is an example of a [patient timeline](https://www.cbioportal.org/patient/summary?caseId=P04&studyId=lgg_ucsf_2014) 

This is what the api response data looks like:
image

In the study view we will need to add a new chart type and new filter type.  We will need a new "counts" service that, like the other existing counts services, accepts a set of study view filters and then returns the counts of the clinical event categories for the samples which match those filters.   You should closely follow the pattern found in the existing count services.  E.g. clinical-data-counts.
https://www.cbioportal.org/api/clinical-data-counts/fetch

haynescd:
Fix #10000

- Update backend to add Clinical Event Type Filtering and add support clinical event type counts

# Checks
- [ ] Runs on heroku

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
+++ b/web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
@@ -36,6 +36,8 @@ public class StudyViewFilter implements Serializable {
     private List<List<String>> caseLists;
     private List<ClinicalDataFilter> customDataFilters;
     private AlterationFilter alterationFilter;
+
+    private List<DataFilter> clinicalEventFilters;
     
     @AssertTrue
     private boolean isEitherSampleIdentifiersOrStudyIdsPresent() {
@@ -202,5 +204,13 @@ public class StudyViewFilter implements Serializable {
 
     public void setPatientTreatmentTargetFilters(AndedPatientTreatmentFilters patientTreatmentTagetFilters) {
         this.patientTreatmentTargetFilters = patientTreatmentTagetFilters;
+    }
+
+    public List<DataFilter> getClinicalEventFilters() {
+        return clinicalEventFilters;
+    }
+
+    public void setClinicalEventFilters(List<DataFilter> clinicalEventFilters) {
+        this.clinicalEventFilters = clinicalEventFilters;
     }
 }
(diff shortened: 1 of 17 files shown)


web/src/main/java/org/cbioportal/web/util/appliers/StudyViewSubFilterApplier.java
```diff
@@ -5,8 +5,8 @@
 
 import java.util.List;
 
-public abstract class StudyViewSubFilterApplier {
-    public abstract List<SampleIdentifier> filter(List<SampleIdentifier> toFilter, StudyViewFilter filters);
+public interface StudyViewSubFilterApplier {
```
haynescd, line 8: Updated to interface, saw no need to keep this as an abstract class

web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
```diff
@@ -36,6 +36,8 @@ public class StudyViewFilter implements Serializable {
     private List<List<String>> caseLists;
     private List<ClinicalDataFilter> customDataFilters;
     private AlterationFilter alterationFilter;
+
+    private List<DataFilter> clinicalEventFilters;
```
haynescd, line 40: I reused DataFilter class to store my EventType filter, but it may be more concise to create a class to hold the filter

dippindots (approved):
Great, it looks good to me!

web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
dippindots: If this suits your case, I think DataFilter is good enough.

PR #10045, opened 2023-02-23
Fix validator bug with DFS_STATUS error log message

pvannierop:
# Problem
The python validator throws a python error when there is a data-dependent error in the `DFS_STATUS` column.

# Solution
This is caused by the passing of a f-string variable where there is no need for string replacement (no `%` present in message string). Removal of the second argument corrects this problem.

### DIFF
--- a/core/src/main/scripts/importer/validateData.py
+++ b/core/src/main/scripts/importer/validateData.py
@@ -2941,7 +2941,6 @@ class PatientClinicalValidator(ClinicalValidator):
                     self.logger.error(
                             'Value in DFS_STATUS column is not 0:DiseaseFree, '
                             '1:Recurred/Progressed, 1:Recurred, 1:Progressed',
-                            'DiseaseFree, Recurred/Progressed, Recurred or Progressed',
                             extra={'line_number': self.line_number,
                                    'column_number': col_index + 1,
                                    'cause': value})


SRodenburg (approved):
Looks good. Remarkable how this never came up before.

core/src/main/scripts/importer/validateData.py
```diff
@@ -2941,7 +2941,6 @@ def checkLine(self, data):
                     self.logger.error(
                             'Value in DFS_STATUS column is not 0:DiseaseFree, '
                             '1:Recurred/Progressed, 1:Recurred, 1:Progressed',
```
dippindots, line 2943: I think we should still keep these values recommendations, I was wondering if you delete the comma at the end of this line, will that fix the problem too?
```suggestion
                            '1:Recurred/Progressed, 1:Recurred, 1:Progressed'
```

core/src/main/scripts/importer/validateData.py
```diff
@@ -2941,7 +2941,6 @@ def checkLine(self, data):
...
-                            'DiseaseFree, Recurred/Progressed, Recurred or Progressed',
```
dippindots, line 2944: @pvannierop I see, I was wondering why we delete `DiseaseFree, Recurred/Progressed, Recurred or Progressed`. Now I get it. You are right. We don't support formats like these `DiseaseFree, Recurred/Progressed, Recurred or Progressed` anymore, we only support values with the prefix `0` or `1`. So we should delete this line.

dippindots (approved):
Approved, thanks for answering my question!

PR #10055, opened 2023-02-28
Add default parameters to DB schema migration script

oplantalech:
Both `--properties-file` and `--sql` properties can be inferred from the environment, so this PR proposes to use the default properties and sql files in the `migrate_db.py` script unless specified in the parameters.

### DIFF
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -7,6 +7,7 @@ import contextlib
 import argparse
 from collections import OrderedDict
 import MySQLdb
+from pathlib import Path
 
 # globals
 ERROR_FILE = sys.stderr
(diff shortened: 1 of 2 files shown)


core/src/main/scripts/migrate_db.py
```diff
@@ -421,18 +422,35 @@ def main():
     """ main function to run mysql migration """
     parser = argparse.ArgumentParser(description='cBioPortal DB migration script')
     parser.add_argument('-y', '--suppress_confirmation', default=False, action='store_true')
-    parser.add_argument('-p', '--properties-file', type=str, required=True,
-                        help='Path to portal.properties file')
-    parser.add_argument('-s', '--sql', type=str, required=True,
-                        help='Path to official migration.sql script.')
+    parser.add_argument('-p', '--properties-file', type=str, required=False,
+                        help='Path to portal.properties file (default: locate it '
+                             'relative to the import script)')
```
fedde-s, line 427: It isn't really an import script, but I don't know whether that will confuse the target audience of this usage message.

fedde-s:
I believe the docs for the script are in <https://github.com/cBioPortal/cbioportal/blob/v5.2.9/docs/Updating-your-cBioPortal-installation.md>, in the same repo. Should those tell people to make use of the simplified command?

fedde-s (approved):
I haven't tested the code either with or without defaults so I might have missed something, but if it works, it looks good to me. I did suggest two small improvments for the documentation that may be worth applying before the merge.

oplantalech:
@fedde-s Thanks! I also updated the documentation - I think it makes sense to keep it simple

oplantalech:
Close this PR because not all tests run - reopened in #10060

PR #10060, opened 2023-03-01
Add default parameters to DB schema migration script

oplantalech:
Reopen #10055 so all tests run

### DIFF
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -7,6 +7,7 @@ import contextlib
 import argparse
 from collections import OrderedDict
 import MySQLdb
+from pathlib import Path
 
 # globals
 ERROR_FILE = sys.stderr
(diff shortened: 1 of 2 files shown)


inodb (approved):
@oplantalech thanks for this suggestion! This overall seems relatively innocuous but it does feel a little odd to me that this script is aware of the location of these files. It seems like something that should be set at the docker image level and then it could e.g. use a docker specific wrapper script to call this with the appropriate paths

oplantalech:
@inodb Thanks for the review. The idea is follow the pattern of the rest of scripts, try to infer the parameters from `PORTAL_HOME` (or similar) so you don't need to explicitly define them. That is also less prone-errors because you will get the config of the cBioPortal you are calling the script from - and you still have the option to pass the parameters if you do really want to use specific files outside of your installation.

inodb:
Thanks @oplantalech ! Makes sense. I guess I mean that relying on file paths feels a little finicky even tho we do it in other places. I'm not sure what common patterns are for this kind of thing tho. It feels this issue stems from stuffing python scripts +web backend etc into one docker image

PR #10079, opened 2023-03-13
Update SQL to fix incompatibility with sqlmode OnlyFullGroupBy

### LINKED ISSUE #10080: BUG SQL Query incompatibility with sqlmode OnlyFullGroupBy Local e2e test failing
Local e2e test are failing due to utilizing a new version of mysql, which defaults ONLY_FULL_GROUP_BY mode to be enabled. 

After MySQL 5.7.5 sqlmode OnlyFullGroupBy is set by default. 

`Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'cbioportal.clinical_event.CLINICAL_EVENT_ID' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`

haynescd:
Fix #10080 

After MySQL 5.7.5 ONLY_FULL_GROUP_BY mode was added, which rejects queries which select list refer to non-aggregated columns that are neither named in the group by clause. 

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

Updated Query to utilize ANY_VALUE() 
https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_any-value

## MySQL vs H2 compatibility
Integration tests run agains a H2 in-memory database that does not support the `ANY_VALUE` statement. To solve this, knowledge of the database type is passed to MyBatis using the [MyBatis DatabaseIdProvider mechanism](https://mybatis.org/mybatis-3/apidocs/reference/org/apache/ibatis/mapping/DatabaseIdProvider.html) that will use `ANY_VALUE` statements only when MySQL database is used.

### DIFF
--- a/core/src/test/resources/applicationContext-dao.xml
+++ b/core/src/test/resources/applicationContext-dao.xml
@@ -67,7 +67,16 @@
 		<property name="maxTotal" value="100" />
 		<property name="poolPreparedStatements" value="true" />
 	</bean>
-	
+
+    <bean id="databaseIdProvider" class="org.apache.ibatis.mapping.VendorDatabaseIdProvider">
+        <property name="properties">
+            <props>
+                <prop key="MySQL">mysql</prop>
+                <prop key="H2">h2</prop>
+            </props>
+        </property>
+    </bean>
+    
     <!-- define the SqlSessionFactory -->
     <bean id="sqlSessionFactory" class="org.mybatis.spring.SqlSessionFactoryBean">
         <property name="dataSource" ref="businessDataSource" />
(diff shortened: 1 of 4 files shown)


dippindots (approved):
LGTM

pvannierop (approved):
Thanks for implementing this! I added the following comment on the PR description to document the change:

## MySQL vs H2 compatibility
Integration tests run agains a H2 in-memory database that does not support the `ANY_VALUE` statement. To solve this, knowledge of the database type is passed to MyBatis using the [MyBatis DatabaseIdProvider mechanism](https://mybatis.org/mybatis-3/apidocs/reference/org/apache/ibatis/mapping/DatabaseIdProvider.html) that will use `ANY_VALUE` statements only when MySQL database is used.

PR #10119, opened 2023-04-07
RFC68: Impl. Struct Var count endpoint for Study View

pvannierop:
# Description

Thsi PR will add the `/structuralvariant-counts/fetch` endpoint to the internal API. This endpoint handles counts for structural variant queries that relate to _gene1::gene2_ positional queries. This functionality is needed for the _Structural Variants_ panel on Study View.

# Note
To distinduish between gene-level and _gene1::gene2_ positional queries, meny menthod names have been renamed to include the `Gene` word as a reference to gene-level count methods.

# Tests
Unit tests for the count methods are included.


service/src/main/java/org/cbioportal/service/util/AlterationEnrichmentUtil.java
```diff
@@ -2,6 +2,8 @@
 
 import org.apache.commons.math3.stat.inference.ChiSquareTest;
 import org.apache.commons.math3.util.Pair;
+import org.cbioportal.model.AlterationCountBase;
+import org.cbioportal.model.AlterationCountBase;
```
dippindots, line 6: duplication, we can remove one

service/src/main/java/org/cbioportal/service/util/AlterationEnrichmentUtil.java
```diff
@@ -160,7 +163,7 @@ public List<AlterationEnrichment> createAlterationEnrichments(Map<String, Pair<L
     
     public long includeFrequencyForSamples(
             List<MolecularProfileCaseIdentifier> molecularProfileCaseIdentifiers,
-            List<T> alterationCountByGenes,
+            List<T> alterationCounts,
```
dippindots, line 166: Just wondering, could we change this `T` to `AlterationCountByGene` too?

service/src/main/java/org/cbioportal/service/util/AlterationEnrichmentUtil.java
```diff
@@ -189,7 +192,7 @@ public long includeFrequencyForSamples(
 
     public long includeFrequencyForPatients(
             List<MolecularProfileCaseIdentifier> molecularProfileCaseIdentifiers,
-            List<T> alterationCountByGenes,
+            List<T> AlterationCountBases,
```
dippindots, line 195: same here, if we can't change it, it's fine too

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
```diff
@@ -103,37 +106,37 @@ public void calculate(List<T> alterationCounts,
...
             }
-            alterationCountByGene.setMatchingGenePanelIds(allMatchingGenePanelIds);
+            alterationCount.setMatchingGenePanelIds(allMatchingGenePanelIds);
         }
 
         if (includeMissingAlterationsFromGenePanel) {
             Map<Integer, Boolean> genesWithAlteration = alterationCounts.stream()
-                    .collect(Collectors.toMap(AlterationCountByGene::getEntrezGeneId, x -> true));
+                    .flatMap(count -> Arrays.stream(count.getEntrezGeneIds()))
+                    .collect(Collectors.toMap(entrezGeneId -> entrezGeneId, x -> true));
```
dippindots, line 126: If we always set `true` as value, could we just use Set?

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -87,8 +85,9 @@ private void init() {
     @Autowired
     private ClinicalDataBinUtil clinicalDataBinUtil;
     @Autowired
+    private AlterationCountService alterationCountService;
+    @Autowired
```
dippindots, line 89: Is this service being used here? I don't see any reference, maybe we can delete it.

dippindots (approved):
@pvannierop Thanks for making this pr, LGTM. I just had some minor comments. They should be easy to address. Could you double-check the localdb tests and make sure there are no related failing tests? Thanks!

service/src/main/java/org/cbioportal/service/util/AlterationEnrichmentUtil.java
pvannierop: Removed!

service/src/main/java/org/cbioportal/service/util/AlterationEnrichmentUtil.java
pvannierop: See my previous comment. But I agree that this name at least should be `alterationCounts` as also used in -includeFrequencyForSamples_.

service/src/main/java/org/cbioportal/service/util/ProfiledCasesCounter.java
pvannierop: Well spotted. I have changed this into a Set.

alisman:
@pvannierop lets connect on this one

service/src/main/java/org/cbioportal/service/util/AlterationEnrichmentUtil.java
BasLee: not here I think: profiledCasesCounter.calculate also expects a T (` extends AlterationCountBase`)

PR #10124, opened 2023-04-12
Do not call next study view filter when no samples left

pvannierop:
This PR will fix MySQL query errors when the _manual submit_ mode of Study View is active and subsequent filters act on a  list of zero samples due to previous filters.


dippindots (approved):
@pvannierop Thanks for the quick fix!

PR #10141, opened 2023-04-26
the mutations validate issue #10134

### LINKED ISSUE #10134: Add validation rule for uncalled mutations
There should never be mutations that are in both a profile named `{x}` and `{x}_uncalled` where `{x}` can be any mutation profile name. The uncalled profile contains all mutations that weren't called. It would be nice to have a check in [the validator](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/scripts/importer/validateData.py) to confirms this. The logic should be something like:

- check if there exists a mutation profile `{x}_uncalled`. If so, there should never be duplicates when you combine the mutation files
- indicate which variants are duplicates in the validation report

Linfye:
Fix  #10134

add a function check_duplicate_mutations


core/src/main/scripts/importer/validateData.py
```diff
@@ -5545,6 +5545,37 @@ def validate_study(study_dir, portal_instance, logger, relaxed_mode, strict_maf_
...
+    def check_duplicate_mutations():
+        # check if data_mutations_uncalled.txt exists
+        if os.path.exists('data_mutations_uncalled.txt') is not None:
+            # read data_mutations.txt and data_mutations_uncalled.txt
+            with open('data_mutations.txt', 'r') as mutations_file, open('data_mutations_uncalled.txt', 'r') as uncalled_file:
+
+                next(mutations_file, uncalled_file)
+                mutations = set()
+                uncalled_mutations = set()
+                uncalled_mutation = set()
```
inodb, line 5558: @Linfye this variable does not seem to be used?
```suggestion
```

dippindots:
@inodb Do you know what's the status of this pr?

PR #10213, opened 2023-06-07
Support StudyViewMyBatisRepository integration tests

onursumer:
- Currently only mysql compatible sql syntax is supported
- Clickhouse specific features like `base64Encode` function breaks the testing environment, need to investigate further for clickhouse compatiblity


gblaih (approved):

inodb:
> Clickhouse specific features like base64Encode function breaks the testing environment, need to investigate further for clickhouse compatiblity

Just double checking - right now we are not using any of these clickhouse functions right?

PR #10214, opened 2023-06-08
RFC68: Impl. structural variant filtering in Study View filter

pvannierop:
# Description

Thsi PR will add filtering based on gene1::gene2 structural variants in the `/filtered-samples/fetch` endpoint to the internal API. This functionality is needed to define the cohort using the _Structural Variants_ panel on Study View.

# Tests
Unit tests for the filter methods are included.


model/src/main/java/org/cbioportal/model/StructVarFilterQuery.java
```diff
@@ -0,0 +1,48 @@
+package org.cbioportal.model;
+
+import org.cbioportal.model.util.Select;
+import org.springframework.lang.Nullable;
+
+import java.io.Serializable;
+
+public class StructVarFilterQuery extends BaseAlterationFilter implements Serializable {
```
onursumer, line 8: I think it is better to spell the name out as `StructuralVariantFilterQuery` for consistency. For example, we already have `StructuralVariantGeneSubQuery`.

web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
```diff
@@ -12,10 +13,11 @@
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import org.cbioportal.model.AlterationFilter;
 import org.cbioportal.model.GeneFilter;
+import org.cbioportal.model.StudyViewStructuralVariantFilter;
 import org.cbioportal.web.parameter.filter.*;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
-@JsonInclude(Include.NON_NULL)
+@JsonInclude(Include.ALWAYS)
```
onursumer, line 20: Why do we need to change this? To prevent null pointer exceptions?

web/src/main/java/org/cbioportal/web/util/appliers/StructuralVariantSubFilterApplier.java
```diff
@@ -0,0 +1,112 @@
...
+    private MolecularProfileService molecularProfileService;
+
+    @Autowired
+    private StructuralVariantService structuralVariantService;
+
+    @Autowired
+    private StudyViewFilterUtil studyViewFilterUtil;
+    
+    @Override
+    public List<SampleIdentifier> filter(List<SampleIdentifier> toFilter, StudyViewFilter filters) {
```
onursumer, line 33: This is probably part of a larger discussion, but we will eventually need to translate this filter to SQL if we want to migrate study view filter related logic into Clickhouse, right?

onursumer (commented):
Looks good in general. Thanks! Added a few minor comments/questions.

web/src/main/java/org/cbioportal/web/parameter/StudyViewFilter.java
BasLee: Reverted in [9665e46](https://github.com/cBioPortal/cbioportal/pull/10214/commits/9665e463a464f5506e36ab3f1d7130b377f8f059)

web/src/main/java/org/cbioportal/web/StudyViewController.java
```diff
@@ -466,7 +467,7 @@ public ResponseEntity<DensityPlotData> fetchClinicalDataDensityPlot(
         @ApiIgnore // prevent reference to this attribute in the swagger-ui interface. this attribute is needed for the @PreAuthorize tag above.
         @Valid @RequestAttribute(required = false, value = "interceptedStudyViewFilter") StudyViewFilter interceptedStudyViewFilter,
         @ApiParam(required = true, value = "Study view filter")
-        @Valid @RequestBody(required = false) StudyViewFilter studyViewFilter) {
```
JREastonMarks, line 469: Why no more @valid?

JREastonMarks:
@pvannierop Do we need to worry about the tests failing? Other than that it looks good to me.

PR #10241, opened 2023-06-27
Fix 10221 403 error; Unauthorized studies are displayed as authorized

### LINKED ISSUE #10221: 403 on tags/fetch when show_unauthorized_studies=true
v5.3.9 on Centos 7
When opening the home page after logging in, I get a 403 for

    POST https://server.url/api/studies/tags/fetch

this request includes a list of all study IDs on the server.
I have 

    authorization=true`
    skin.home_page.show_unauthorized_studies=true

and my account has access to some studies but not all.

 My global_message doesn't even contain any study tags but some studies do have tags set.

This doesn't happen if I have access to all studies or if I disable the show unauthorized. It did work in 3.7.22 FWIW.

Should perhaps ```getTagsForMultipleStudies``` be tagged with ```T(org.cbioportal.utils.security.AccessLevel).LIST``` instead of  ```T(org.cbioportal.utils.security.AccessLevel).READ``` ?

jagnathan:
Fix #10221  Unauthorized studies are displayed as authorized due to issue with using RequestParam


web/src/main/java/org/cbioportal/web/StudyController.java
```diff
@@ -98,9 +99,9 @@ public ResponseEntity<List<CancerStudy>> getAllStudies(
         @ApiParam("Name of the property that the result list is sorted by")
         @RequestParam(required = false) StudySortBy sortBy,
         @ApiParam("Direction of the sort")
-        @RequestParam(defaultValue = "ASC") Direction direction,
-        @ApiParam(hidden = true, required = false)
-        @RequestParam(required = false) Authentication authentication) {
+        @RequestParam(defaultValue = "ASC") Direction direction
+    ) {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
```
BasLee, line 104: Problem:
- The `authentication` param _without_ annotations generates faulty/messy Swagger documentation
- The `authentication` param _with_  annotations results in all studies being marked with  `readPermission=true`
Maybe better to move the authentication parameter and instead fetch it from the security context?
@dippindots @JREastonMarks

dippindots: This looks ok to me, @jagnathan is there any potential issue if we merge this pr as is?

jagnathan: @dippindots in the file org/mskcc/cbio/portal/util/internal/AccessControlImpl.java, there is some more complex logic for getting the authentication object. Do we need to follow that logic?

jagnathan:
opened a new PR https://github.com/cBioPortal/cbioportal/pull/10267 to replace this PR.

PR #10245, opened 2023-06-28
Extract column store endpoints into StudyViewControllerColumnStore

haynescd:


web/src/main/java/org/cbioportal/web/columnstore/StudyViewControllerColumnStore.java
```diff
@@ -0,0 +1,73 @@
...
+
+import javax.validation.Valid;
+import java.util.Collection;
+import java.util.List;
+
+@InternalApi
+@RestController
+@Validated
+@Api(tags = "Study View Column Store")
+public class StudyViewControllerColumnStore {
```
onursumer, line 32: maybe name it this way? not a big deal though, we will most likely end up renaming these to something different.

```suggestion
public class StudyViewColumnStoreController {
```

web/src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -127,7 +128,7 @@ public class InvolvedCancerStudyExtractorInterceptor extends HandlerInterceptorA
         if (!request.getMethod().equals("POST")) {
             return true; // no attribute extraction needed because all user supplied filter objects are in POST requests
         }
-        String requestPathInfo = request.getPathInfo();
+        String requestPathInfo = StringUtils.removeStart(request.getPathInfo(), "/column-store");
```
onursumer, line 131: or we can just do this if we want to avoid using a 3rd party utility class.

```suggestion
        String requestPathInfo = request.getPathInfo().replaceFirst("/column-store", "");
```

onursumer (approved):
Looks good!

PR #10276, opened 2023-07-14
Fix disabling SSL for MySQL connection

### LINKED ISSUE #10275: Migration script does not respect db.use_ssl property
Affected PR: https://github.com/cBioPortal/cbioportal/pull/10037

# Problem
When migrating the database the migration.py script tries to connect to the MySQL database over SSL, despite the _db.use_ssl_ property having value _false_.

# Analysis
The current and broken logic in migration.py to configure SSL is:

```
connection_kwargs = {here
if portal_properties.database_use_ssl == 'true':
    connection_kwargs['ssl'] = {"ssl_mode": True}
```

When I change this to the following, the _db.use_ssl_ property is respected:

```
connection_kwargs = {}
if portal_properties.database_use_ssl == 'true':
   connection_kwargs["ssl_mode"] = "REQUIRED"
else:
   connection_kwargs["ssl_mode"] = "DISABLED"
```

# _mysqlclient_ python library

Official documentation of _mysqlclient_ python library states (see [here](https://mysqlclient.readthedocs.io/user_guide.html#functions-and-attributes)):

>ssl_mode
>If present, specify the security settings for the connection to the server. For more information on ssl_mode, see the MySQL documentation. Only one of ‘DISABLED’, ‘PREFERRED’, ‘REQUIRED’, ‘VERIFY_CA’, ‘VERIFY_IDENTITY’ can be specified.
>If not present, the session ssl_mode will be unchanged, but in version 5.7 and later, the default is PREFERRED.
> This must be a keyword parameter.
>
>ssl
>This parameter takes a dictionary or mapping, where the keys are parameter names used by the [mysql_ssl_set](http://dev.mysql.com/doc/refman/en/mysql-ssl-set.html) MySQL C API call. If this is set, it initiates an SSL connection to the server; if there is no SSL support in the client, an exception is raised. This must be a keyword parameter.

The _ssl_ parameter can reference values mentioned [here](https://dev.mysql.com/doc/c-api/5.7/en/mysql-ssl-set.html).

# Discussion
- Should we accept the proposed fix to the problem?
- Does connection over SSL work at all with the current script? I have a hard time believing this because the _ssl_ parameter does not receive correct parameters.

pvannierop:
Fix #10275

# Problem
When migrating the database the migration.py script tries to connect to the MySQL database over SSL, despite the _db.use_ssl_ property having value _false_.

# Analysis
The current and broken logic in migration.py to configure SSL is:

```
connection_kwargs = {here
if portal_properties.database_use_ssl == 'true':
    connection_kwargs['ssl'] = {"ssl_mode": True}
```

When I change this to the following, the _db.use_ssl_ property is respected:

```
connection_kwargs = {}
if portal_properties.database_use_ssl == 'true':
    connection_kwargs["ssl_mode"] = "REQUIRED"
else:
   connection_kwargs["ssl_mode"] = "DISABLED"
```

# Changes in this PR
This PR will add the _ssl_mode_ to the kwargs of the migration script.

### DIFF
--- a/core/src/main/scripts/migrate_db.py
+++ b/core/src/main/scripts/migrate_db.py
@@ -61,6 +61,9 @@ def get_db_cursor(portal_properties):
         connection_kwargs['db'] = portal_properties.database_name
         if portal_properties.database_use_ssl == 'true':
             connection_kwargs['ssl'] = {"ssl_mode": True}
+            connection_kwargs['ssl_mode'] = 'REQUIRED'
+        else:
+            connection_kwargs['ssl_mode'] = 'DISABLED'          
         
         connection = MySQLdb.connect(**connection_kwargs)
     except MySQLdb.Error as exception:


dippindots:
Closing this and created https://github.com/cBioPortal/cbioportal/pull/10353

PR #10277, opened 2023-07-17
Filter out unauthorized studies when fetching study tags

BasLee:
This PR filters out unauthorized studies by adding the `@PreFilter` annotation to the `/studies/tags/fetch` endpoint.

Depends on fixing authorization in `/studies` (see https://github.com/cBioPortal/cbioportal/pull/10267).

At the moment a user with unauthorized studies gets an error screen on the index page. This is because the tags of all studies are fetched. However, you cannot fetch the study tags of unauthorized studies, resulting in a 403 and an error screen (see https://github.com/cBioPortal/cbioportal/issues/10221) .

By filtering the studies in the backend, the frontend does not need to have any knowledge of authorized studies (see also https://github.com/cBioPortal/cbioportal-frontend/pull/4653).

### DIFF
--- a/web/src/main/java/org/cbioportal/web/StudyController.java
+++ b/web/src/main/java/org/cbioportal/web/StudyController.java
@@ -22,6 +22,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreFilter;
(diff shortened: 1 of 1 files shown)


inodb:
Closing this as I believe we decided to not expose study tags for unauthorized studies due to risk of exposing sensitive data, but feel free to re-open if I'm misremembering the discussion. One alternative implementation would be to allow exposing study tags endpoint by allowing deployers to opt in to sharing the tags or indicating clearly in the docs on this and the portal configuration file with some warnings that it exposes the tags for a study as well

PR #10317, opened 2023-08-08
Add Karthik to Caris on About Page

inodb:
Karthik is back 🎊


dippindots:
@inodb Thanks for making this! I already done this: https://docs.cbioportal.org/about-us/#caris-life-sciences

PR #10336, opened 2023-08-22
Add json prop for custom patient view sample type colors

BasLee:
Extend the legend colors of sample types in the patient view by configuring custom sample types from a json file on the classpath. 
Applies the same mechanism that is used to configure default clinical tracks in the oncoprint.
See also the frontend PR: https://github.com/cBioPortal/cbioportal-frontend/pull/4704

### DIFF
--- a/docs/deployment/customization/portal.properties-Reference.md
+++ b/docs/deployment/customization/portal.properties-Reference.md
@@ -223,6 +223,13 @@ Define the colors of custom sample types in the patient view using a json object
 ```
 skin.patient_view.custom_sample_type_colors_json=classpath:/skin-patient-view-custom-sample-type-colors.json
 ```
+Example of json file contents:
+```json
+{
+    "Primary": "green",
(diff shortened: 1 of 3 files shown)


inodb (approved):
LGTM! Thanks!

Let's revisit when refactoring how this property passing should work

docs/deployment/customization/portal.properties-Reference.md
```diff
@@ -218,6 +218,19 @@ skin.patient_view.copy_number_table.columns.show_on_init=
 skin.patient_view.structural_variant_table.columns.show_on_init=
 ```
 
+### Define custom sample type colors
+Define the colors of custom sample types in the patient view using a json object with for each sample type a color:
+```
+skin.patient_view.custom_sample_type_colors_json=classpath:/skin-patient-view-custom-sample-type-colors.json
```
alisman, line 224: @inodb this documentation is wrong now, correct?  i mean, why is there a classpath there?

PR #10371, opened 2023-09-06
Fix Unit Tests in Web Module

haynescd:


web/src/main/java/org/cbioportal/web/ClinicalDataEnrichmentController.java
```diff
@@ -39,6 +39,7 @@
 
 @InternalApi
 @RestController
+@RequestMapping("/api")
```
inodb, line 42: if there's a way to configure this for e.g. all Controllers or e.g. `InternalApi` and `PublicApi` groups that might be better but no worries if not

inodb (approved):
Looks good to me! 

if there's a way to configure the `/api` prefix for all API Controllers or e.g. InternalApi and PublicApi groups that might be cleaner but no worries if not

PR #10387, opened 2023-09-15
Use RSA256 algorithm for SAML2 authN statement singing

pvannierop:
# Problem
Keycloak 22 does not allow login from cBioPortal because the authN request is signed with SHA1 algorithm. Keycloak dropped support for SHA1 because it is insecure.

# Solution
Configure cBioPortal to sign with RSA-256 algorithm.

This PR will reconfigure cBioPortal to use RSA-256 according to this [blogpost](https://myshittycode.com/2016/02/23/spring-security-saml-replacing-sha-1-with-sha-256-on-signature-and-digest-algorithms/).

### DIFF
--- a/security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/saml/SAMLBootstrapRSA256.java
+++ b/security/security-spring/src/main/java/org/cbioportal/security/spring/authentication/saml/SAMLBootstrapRSA256.java
@@ -1,0 +1,18 @@
+package org.cbioportal.security.spring.authentication.saml;
+
+import org.opensaml.Configuration;
+import org.opensaml.xml.security.BasicSecurityConfiguration;
+import org.opensaml.xml.signature.SignatureConstants;
+import org.springframework.beans.BeansException;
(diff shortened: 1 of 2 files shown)


dippindots:
Inviting @n1zea144 for review, @pvannierop with this change, will cbioportal be able to use keycloak v22?

pvannierop:
@dippindots Yes, this change is needed to operate with Keycloak v22. I tested backwards compatibility with Keycloak v16.

inodb:
I'll double check our keycloak version and then merge

inodb:
Looks like we're on v9 btw:

```
keycloak                                        default         1               2023-07-05 12:32:48.814831 -0700 -0700  deployed        keycloak-9.0.1          11.0.0
```
Gonna try with that version and retry

inodb:
It seems to work on our genie instance

@pvannierop it looks like there is some issue in the core tests tho: https://github.com/cBioPortal/cbioportal/actions/runs/6395996176/job/17360869425. Any clue?

inodb (changes requested):
Some issue in core tests but otherwise looks good

PR #10412, opened 2023-09-29
Add 'skin.left_logo' property

pvannierop:
This PR will implement a new property `skin.left_logo` that will be used by cbioportal-frontend to render a custom logo at the left side of the header.

### DIFF
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -37,6 +37,7 @@ skin.documentation.dat=Authenticating-Users-via-Tokens.md
 
 # setting controlling the logos
 skin.right_logo=
+skin.left_logo=
 skin.tag_line_image=tag_line.png
 
 # setting controlling which tabs to hide.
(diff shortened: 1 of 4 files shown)


src/main/resources/portal.properties.EXAMPLE
```diff
@@ -37,6 +37,7 @@ skin.documentation.dat=Authenticating-Users-via-Tokens.md
 
 # setting controlling the logos
 skin.right_logo=
+skin.left_logo=
 skin.tag_line_image=tag_line.png
```
inodb, line 41: I usually prefer commenting out properties as it makes it easier to control defaults in the future. Also for testing: we currently won't test what happens when the property is not set (as it uses EXAMPLE properties), which is going to be the default for anybody upgrading their instance

inodb (approved):
LGTM

src/main/resources/portal.properties.EXAMPLE
pvannierop: @inodb Ok, I matched the `skin.right_logo` entry. Are these different, or should I also comment out that one?

PR #10423, opened 2023-10-09
Fix context initialization when property 'dat.method' is missing

pvannierop:
# a Problem
When property _dat.method_ is not defined in portal.properties, context initialization complains about a missing TokenService bean. The cause of this error is that TokenService initialization does not include a default.

# Solution
This PR will set the 'none' value as the default when the property is undefined.


inodb (commented):
LGTM - thanks!

inodb (approved):
LGTM - Thanks!

PR #10424, opened 2023-10-09
Replace DB Prop Assertion with Warning 

averyniceday:
Original change introduced in PR #10037 - switching away from redundant properties (e.g. using db connection string instead of separate db host/name)

The switch to the connection string works for establishing a connection but the introduced assertion breaks the backend importer which depends on the cbio core module. Specifically, the importer uses the single properties to wire Spring beans (some for legacy features) and does not accept blank values. Conversely, if we do have the properties set, the importer fails at runtime due to the assertion failing. Given the scope of refactoring the backend pipeline codebase, the proposed solution is to switch the assertion to a warning and plan a separate effort for migrating the backend importer. 

- Assertion is non-compatible with backend pipeline importer which needs db props for bean initialization
- Replace assertion with warning

# Checks

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
+++ b/core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
@@ -4,11 +4,15 @@ import org.apache.commons.dbcp2.BasicDataSource;
 import org.mskcc.cbio.portal.util.DatabaseProperties;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.util.Assert;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Data source that self-initializes based on cBioPortal configuration.
  */
 public class JdbcDataSource extends BasicDataSource {
+    
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcDataSource.class);
 
     public JdbcDataSource () {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
```diff
@@ -20,17 +24,17 @@ public JdbcDataSource () {
...
-                "-- Connection error:\n" +
-                "-- You try to connect to the database using the deprecated 'db.host', 'db.portal_db_name' and 'db.use_ssl' properties.\n" +
-                "-- Please remove these properties and use the 'db.connection_string' property instead. See https://docs.cbioportal.org/deployment/customization/portal.properties-reference/\n" +
-                "-- for assistance on building a valid connection string.\n" +
-                "----------------------------------------------------------------------------------------------------------------\n"
-        );
-        
+
+        if (defined(host) || defined(database) || defined(dbProperties.getDbUseSSL())) {
+		LOG.info("\n----------------------------------------------------------------------------------------------------------------" +
```
dippindots, line 29: Should we use `Log.warning` as we discussed during our meeting?

dippindots (approved):

JREastonMarks (commented):
lgtm

PR #10439, opened 2023-10-19
Added setting for python-specific connection URL 

Floris-Hyve:
### Problem:

Standard database connection string does not work when using MySQL python connector for migration. For example, ssl connection is disabled for the standard connection URL, but ssl connection is required for python MySQL connection. 

### Fix: 

Added Python-specific connection URL, and implemented it in the get_database_properties function, so the correct URL is selected for the migration script.

Input URL is given in the portal.properties:

`db.connection_string_python=`

### DIFF
--- a/docs/deployment/customization/portal.properties-Reference.md
+++ b/docs/deployment/customization/portal.properties-Reference.md
@@ -31,6 +31,17 @@ the `db.connection_string` instead.
 db.tomcat_resource_name=jdbc/cbioportal
 ```
 
+### Python-specific connection URL (internal use only)
+
+``` 
+db.connection_string_python= 
+
+``` 
+
+Setting that allows a python-specific database connection URL which is different from the connection URL used by Java. This allows, for instance, to migrate the database with different connection properties in the docker entry point script. If provided, the python-specific connection URL takes precedence over the standard connection string. 
+
+The format of the `db.connection_string_python` is identical to `db.connection_string`
+
 ## cBioPortal Customization
 
 ### Hide tabs (pages)
--- a/src/main/resources/portal.properties.EXAMPLE
+++ b/src/main/resources/portal.properties.EXAMPLE
@@ -6,6 +6,8 @@ db.user=cbio_user
 db.password=somepassword
 db.driver=com.mysql.jdbc.Driver
 db.connection_string=jdbc:mysql://localhost:3306/cbioportal?useSSL=false
+# Connection string for when both python and java connections are used in the docker entrypoint. When a connection URL is given for python, it will be used to form a connection to the mySQL database instead of the standard connection string. This allows for different connection properties to be used (like ssl connection for example). 
+# db.connection_string_python= jdbc:mysql://cbioDB:3306/cbioportal
 # set tomcat_resource_name when using dbconnector=jndi instead of the default
 # dbconnector=dbcp. Note that dbconnector needs to be set in CATLINA_OPTS when
 # using Tomcat (CATALINA_OPTS="-Ddbconnector=jndi"). It does not get picked up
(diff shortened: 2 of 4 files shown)


docs/deployment/customization/portal.properties-Reference.md
```diff
@@ -31,6 +31,22 @@ the `db.connection_string` instead.
...
 ```
 
+### Python-specific connection URL
+
+``` 
+db.connection_string_python= 
+
+``` 
+
+Setting for python-specific connection URL in order to connect using both python and java at the entrypoint.
```
pvannierop, line 41:
```suggestion
Setting that allow a python-specific database connection URL that is different from the connection URL used by Java. This allows, for instance, to migrate the database with different connection properties in the docker entry point script.
```

docs/deployment/customization/portal.properties-Reference.md
```diff
@@ -31,6 +31,22 @@ the `db.connection_string` instead.
...
+
+The format of the `db.connection_string_python` is: 
+
+``` 
+jdbc:mysql://localhost:3306/cbiodb
+
+```
```
pvannierop, line 48:
```suggestion
The format of the `db.connection_string_python` is: 

``` 
jdbc:mysql://localhost:3306/cbiodb

``` 
```

The format of the `db.connection_string_python` is identical to `db.connection_string`.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -6,6 +6,8 @@ db.user=cbio_user
 db.password=somepassword
 db.driver=com.mysql.jdbc.Driver
 db.connection_string=jdbc:mysql://localhost:3306/cbioportal?useSSL=false
+# Added connection string for when both python and java connections are used in the entrypoint
```
pvannierop, line 9: Please better describe what this setting __does__, not only when it is used.

pvannierop (changes requested):
Some small changes.

src/main/resources/portal.properties.EXAMPLE
```diff
@@ -6,6 +6,8 @@ db.user=cbio_user
 db.password=somepassword
 db.driver=com.mysql.jdbc.Driver
 db.connection_string=jdbc:mysql://localhost:3306/cbioportal?useSSL=false
+# Connection string for when both python and java connections are used in the docker entrypoint. When a connection URL is given for python, it will be used to form a connection to the mySQL database instead of the standard connection string. This allows for different connection properties to be used (like ssl connection for example). 
+# db.connection_string_python= jdbc:mysql://cbioDB:3306/cbioportal
```
pvannierop, line 10: The comment on precedence should also  be made in the portal.properties-reference.md

pvannierop (changes requested):
Two proposed changes:
1. Clean up the commits (now there are 7!)
2. Harmonize information on precedence of the URL properties.

JREastonMarks:
From an RFC72 (6.0) perspective the core module has been moved out of the cbioportal code base and into its own code base here: https://github.com/cBioPortal/cbioportal-core

@pvannierop can you create another PR for this in that repo? That way it will work in both 5.X and 6.X versions of cBioPortal.

alisman:
@JREastonMarks can you provide direction on how re-target this PR in a 6.0 world.

JREastonMarks:
Hi @Floris-Hyve,
With the release of 6.0.0 (RFC72) we moved a lot of code around and updated some of the files that you reference in this PR. The core module has been moved to its own repository: https://github.com/cBioPortal/cbioportal-core. If you move this PR to the core package it should still be a quick change. Please let me know if you need any help.

JREastonMarks:
Closing but willing to reopen if needed.

PR #10452, opened 2023-10-31
Add documentation for quick select on homepage

sowmiyaa-kumar:
Add docs for quick select button on homepage:

image


inodb (approved):
missed this - thanks for sending! I rebased it

PR #10462, opened 2023-11-09
Label checker action reruns when needed

Luke-Sikina:
- The label checker GH action currently runs when a PR is posted, but not when labels are added
- Since most people add labels _after_ they post a PR, this makes the action mostly useless
- Fix: rerun action on add/remove label
- If this works, you could start making label checker a required check, which would make your release process a little cleaner, I think


inodb (approved):
Awesome! Thanks @Luke-Sikina !

PR #10487, opened 2023-12-11
Removing the filtering of amino acid changes with '+' as not relevant…

sowmiyaa-kumar:
… anymore

This PR addresses an issue identified in the current variant representation guidelines. According to Q1 under Q&A on [HGVS nomenclature rules](https://varnomen.hgvs.org/recommendations/DNA/variant/alleles/), the use of the '+' sign was previously a method to represent multiple changes together. This format is not recommended anymore, and it restricts the loading of 5' intronic splicing variants, such as c.357+1G>A.


rmadupuri (approved):

rmadupuri:
Looks good! Thanks for catching that!

PR #10505, opened 2023-12-22
move treatments endpoint to internal API

jagnathan:
Move Treatments Endpoint to Internal API - fix  cBioPortal/icebox#561 

Note:
- API endpoints stay at the same location, just in certain clients they might have to refer to the endpoint now via internal instead of public


jagnathan:
Move Treatments Endpoint to Internal API - fix #10345

alisman:
@jagnathan is this ready for review?  This will probalby require a frontend change so lets be careful about that.

jagnathan:
@alisman this is ready for review. as per my understanding, the frontend change is automatic. the API endpoint will move into internal section of the Swagger Doc. Are there other changes required on frontend?

inodb (approved):
LGTM ! Thank you!

PR #10513, opened 2024-01-02
Fix CacheMapUtil configurations

haynescd:


JREastonMarks (approved):

src/main/java/org/cbioportal/web/util/InvolvedCancerStudyExtractorInterceptor.java
```diff
@@ -76,17 +67,23 @@
...
+import java.util.stream.Collectors;
+
 public class InvolvedCancerStudyExtractorInterceptor implements HandlerInterceptor {
 
     @Autowired
     private UniqueKeyExtractor uniqueKeyExtractor;
 
     private ObjectMapper objectMapper = new ObjectMapper();
 
-    @Qualifier("staticRefCacheMapUtil")
```
inodb, line 89: Why is this one no longer necessary?

haynescd: That was there to force to a specific class when multiple IMPLs are available for a specific Interface. After adding the Conditionals for the CacheMapUtils there is now only one CacheMapUtils Implementation on the class path.

PR #10527, opened 2024-01-10
Add Installations Map Redirect

haynescd:


pom.xml
```diff
@@ -12,7 +12,7 @@
 
 	<groupId>org.cbioportal</groupId>
 	<artifactId>cbioportal</artifactId>
-	<version>6.0.1-SNAPSHOT</version>
+	<version>6.0.0</version>
```
alisman, line 15: is this intetnional?

alisman (approved):

pom.xml
haynescd: it should actually be 6.0.0-SNAPSHOT I will update now

PR #10532, opened 2024-01-10
Demo 6.0.0

JREastonMarks:
Fix issues with 6.0.0 RC deployment.
Includes proxy fixes
Fixes Swagger
Adds Sentry
Other smaller fixes

### DIFF
--- a/src/main/resources/application.properties.EXAMPLE
+++ b/src/main/resources/application.properties.EXAMPLE
@@ -408,4 +408,9 @@ server.max-http-request-header-size=16384
 #sentry.traces-sample-rate=1.0 Probably should change for prod
 #sentry.exception-resolver-order=-2147483647
 
+# Swagger Configuration
+springdoc.swagger-ui.disable-swagger-default-url=true
+springdoc.swagger-ui.path=/api/swagger-ui
+springdoc.api-docs.path=/api/v2/api-docs
+
 # EOL - Do not delete the following lines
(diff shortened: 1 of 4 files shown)


src/main/resources/application.properties.EXAMPLE
```diff
@@ -403,4 +403,14 @@ server.max-http-request-header-size=16384
...
 
+# Sentry Config
+#sentry.dsn=
+#sentry.traces-sample-rate=1.0 Probably should change for prod
+#sentry.exception-resolver-order=-2147483647
+
+# Swagger Configuration
+springdoc.swagger-ui.disable-swagger-default-url=true
+springdoc.swagger-ui.path=/api/swagger-ui
+springdoc.api-docs.path=/api/v2/api-docs
```
inodb, line 414: I feel slightly uncomfortable about making this the default in the EXAMPLE properties: springdoc.api-docs.path=/api/v2/api-docs. As I guess it is actually v3 of the api-docs. Might be cleaner if we can somehow serve both /api/v3/api-docs and /api/v2/api-docs . Wondering if in the future we can prevent this by just having e.g. a /api/api-docs. Anyway can be follow up ticket, filed on here: https://github.com/cBioPortal/cbioportal/issues/10534. Prolly ok for now

inodb (approved):
LGTM! Thank you! Added a follow up ticket re the change in API docs url

PR #10548, opened 2024-01-16
Fix SonarCube Coverage

JREastonMarks:


.github/workflows/sonarcloud.yml
```diff
@@ -36,4 +36,9 @@ jobs:
         env:
           SONAR_TOKEN:  de1b5cc660cd210dde840f492c371da6cc801763
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: mvn -DskipTests clean install verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Pcoverage
+        run: mvn clean install verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
+      - name: Code Coverage
+        env:
+          SONAR_TOKEN:  de1b5cc660cd210dde840f492c371da6cc801763
```
haynescd, line 42: do we want our token exposed?

haynescd: We could hide this with an env var

JREastonMarks: You are right but our token has been exposed for a while. I'll take a look at bringing it into a secure area.

PR #10549, opened 2024-01-17
Add updates to spring security config

haynescd:


src/main/java/org/cbioportal/security/config/Saml2SecurityConfig.java
```diff
@@ -69,7 +65,7 @@ public SecurityFilterChain samlFilterChain(HttpSecurity http) throws Exception {
             // described at https://docs.spring.io/spring-security/reference/6.1/servlet/saml2/logout.html
             // Logout Service POST Binding URL: http://localhost:8080/logout/saml2/slo
             .logout(logout -> logout
-                .logoutUrl(logoutUrl)
+                .logoutUrl("/logout")
```
JREastonMarks, line 68: Should be configurable and final as line 41 above

nr23730:
I tested this. Unfortunately the `claims` object in `OAuth2SecurityConfig.java:83` does not contain the `resource_access` key. However, I was able to verify that the raw jwt sent by Keycloak does indeed contain it. Does spring security process it in an earlier stage, so that it is not present here anymore?
Also, as a result the mappedAuthorities object will be returned empty. Shouldn't this result in an empty study list when the authorization is set to `oauth2` and not `optional_oauth2`?

haynescd:
@nr23730 Have you pulled down the latest code? 
We have two properties to set the roles path. `spring.security.oauth2.client.jwt-roles-path`  and `dat.oauth2.jwtRolesPath` which will account for the dynamic roles path.

For the last question it just depends on your configuration. But in general yes. 

depends on if they have this property set `always_show_study_group`

nr23730:
I tried again with e7930ee1f0fecd95c74517068c64d5099201b0b5 and it didn't work out of the box.
However, I was able to make it work it by putting the client roles in the userinfo similar to what they did with the realm roles here.
https://stackoverflow.com/questions/69331013/springboot-oauth2-with-keycloak-not-returning-mapped-roles-as-authorities
So it seems like it's possible to read the roles from the access_token. Probably it's worth either implementing this or at least to mention it in the docs.

According to the `always_show_study_group` Parameter - that did the trick! I migrated from 4.1.13 where this was disabled by default, so I expected it to work the same....

Thank you very for helping me resolving the issue!

PR #10567, opened 2024-01-25
Fix some issues and code cleanup

JREastonMarks:
Addresses some sonarcube issues that weren't picked up in the first merge.

### DIFF
--- a/src/main/java/org/cbioportal/security/config/OAuth2SecurityConfig.java
+++ b/src/main/java/org/cbioportal/security/config/OAuth2SecurityConfig.java
@@ -42,22 +42,19 @@ public class OAuth2SecurityConfig {
 
     @Bean
     public SecurityFilterChain filterChain(HttpSecurity http, ClientRegistrationRepository clientRegistrationRepository) throws Exception {
-        http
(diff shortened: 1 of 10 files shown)


haynescd (approved):
lgtm

src/main/java/org/cbioportal/security/config/OAuth2SecurityConfig.java
```diff
@@ -42,22 +42,19 @@ public class OAuth2SecurityConfig {
...
             )
             .oauth2Login(login ->
                 login
-                    // TODO: Add constants
                     .loginPage(LOGIN_URL)
                     .userInfoEndpoint(userInfo ->
                     userInfo.userAuthoritiesMapper(userAuthoritiesMapper())
                 )
-                    .failureUrl("/login?logout_failure")
+                    .failureUrl(LOGIN_URL + "/?logout_failure")
```
haynescd, line 57: I don't think there should be a forward slash?

/login/?logout_failure

JREastonMarks: good catch

PR #10568, opened 2024-01-25
Add dynamic host to legacy swagger api

haynescd:


inodb (approved):
Awesome - thanks for the quick fix!

src/main/resources/application.properties.EXAMPLE
```diff
@@ -409,9 +409,9 @@ server.max-http-request-header-size=16384
 #sentry.exception-resolver-order=-2147483647
 
 # Swagger Configuration
-springdoc.swagger-ui.disable-swagger-default-url=true
-springdoc.swagger-ui.path=/api/swagger-ui
-springdoc.api-docs.path=/api/v2/api-docs
+#springdoc.swagger-ui.disable-swagger-default-url=true
+#springdoc.swagger-ui.path=/api/swagger-ui
+#springdoc.api-docs.path=/api/v2/api-docs
```
inodb, line 414: wondering if we should make the EXAMPLE the same as public portal? I imagine that for most deployers it'll be useful if the v2 API keeps working for now

PR #10602, opened 2024-02-05
Update backend configuration

dippindots:
- msk slides viewer
- fix darwin config issue


src/main/java/org/cbioportal/web/util/TokenUtils.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+@Component
+public class TokenUtils {
+    @Autowired
+    private Environment env;
+    public String getMskWholeSlideViewerToken() {
```
haynescd, line 14: This should be a true util class.

So I would make this into a static function and pass the values in via function arg.

We should move away from the convention of having util classes as components

haynescd (changes requested):

src/main/java/org/cbioportal/web/IndexPageController.java
```diff
@@ -51,6 +55,8 @@ private Map<String, String> getFrontendProperties(HttpServletRequest request, Au
         properties.put("user_email_address", authentication != null ? authentication.getName(): "anonymousUser");
         // TODO: Support skin.user_display_name 
         properties.put("user_display_name", authentication != null ? authentication.getName(): "anonymousUser");
+        // Set MSK slide viewer token at runtime
+        properties.put("mskWholeSlideViewerToken", tokenUtils.getMskWholeSlideViewerToken());
```
haynescd, line 59: I would actually just create a helper function here (no need for TokenUtils)

src/main/java/org/cbioportal/web/IndexPageController.java
```diff
@@ -51,8 +56,28 @@ private Map<String, String> getFrontendProperties(HttpServletRequest request, Au
         properties.put("user_email_address", authentication != null ? authentication.getName(): "anonymousUser");
         // TODO: Support skin.user_display_name 
         properties.put("user_display_name", authentication != null ? authentication.getName(): "anonymousUser");
+        // Set MSK slide viewer token at runtime
+        properties.put("mskWholeSlideViewerToken", getMskWholeSlideViewerToken(env.getProperty("msk.whole.slide.viewer.secret.key"), authentication));
```
haynescd, line 60: I would suggest changing this to 
`
    @Value("${msk.whole.slide.viewer.secret.key:}")
    private String wholeSlideViewerKey;`

src/main/java/org/cbioportal/web/IndexPageController.java
```diff
@@ -35,6 +37,9 @@ public class IndexPageController {
     
     @Autowired
     private HttpRequestUtils requestUtils;
+
+    @Autowired
+    private Environment env;
```
haynescd, line 42: You then can remove this

haynescd (changes requested):

haynescd (approved):
lgtm

PR #10605, opened 2024-02-07
Add CheckDarwinAccess to proxy controller

haynescd:


src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -44,6 +48,19 @@ public class ProxyController {
...
+    @Value("${darwin.auth_url:}")
+    private String darwinAuthUrl;
+
+    @Value("${ddp.response_url:}")
+    private String ddpResponseUrl;
+
+    @Value("${cis.user:}")
+    private String cisUser;
+
+    @Value("${darwin.regex:Test}")
```
dippindots, line 61: Just curious, do we need a default for this value?

dippindots (approved):
Looks good to me

JREastonMarks (approved):

PR #10616, opened 2024-02-12
Update bravado documentation with persistent authorization

MatthijsPon:
This way, users do not have to pass the authorization headers every time they make a request, but only need to set it once


inodb (approved):
Great - thanks so much!

PR #10617, opened 2024-02-12
Update ER diagram to db schema 2.13.1

forus:
The layout of the ER diagram is also changed to highlight the records hierarchy:

- study
  - patient
    - sample
      - molecular data type

The respective tables are positioned on the main diagonal.


forus:
Tablese that were created/removed/updated: Screenshot 2024-02-12 at 13 35 19

JREastonMarks (approved):
lgtm
Did you use a script or some other process to create this?
It would be great to add this to our pipelines.

forus:
> lgtm Did you use a script or some other process to create this? It would be great to add this to our pipelines.

It would be great indeed. I thought about this.

I used MySQL Workbench GUI. It can synchronize the diagram with the DB schema (Database > Synchronize Model...). Unfortunately, I have not found how to do this from the command line. We also tend to group tables logically in colorful boxes and manually change the diagram's layout. The auto-layout is not so great.

PR #10620, opened 2024-02-12
Changes Immutable List to Mutable List

JREastonMarks:
#10618 was caused by issues with retainAll trying to mutate an immutable object.

This fix changes sampleIdentifiers to a mutable object and allows the retainAll function to work (i.e., remove which is part of retainAll only works on mutable lists).


inodb (approved):
Thanks!

Issue #10556, opened 2024-01-19
Customizing custom driver


kalletlak:
In continuation to https://github.com/cBioPortal/cbioportal-frontend/pull/4806

@jjgao and @inodb  suggested that we should customize the columns name.

- Adding fields to [application.properties](https://github.com/cBioPortal/cbioportal/blob/master/src/main/resources/application.properties.EXAMPLE)
  - skin.custom.driver.name=Custom Driver
  - skin.custom.driver.description=Custom driver annotation
  - skin.custom.driver.tiers.name=Custom Driver Tiers
  - skin.custom.driver.description=Custom driver tiers
  - above are default values. For Caris, let's use
    - skin.custom.driver.name=Caris Driver Annotation
    - skin.custom.driver.description=P/LP annotation by Caris Biomarker & Drugs Intelligence team
    - skin.custom.driver.tiers.name=Caris Driver Tiers
    - skin.custom.driver.description=Caris Driver Tiers
- use the name from application.properties for the column names with mouse-over message showing the description
- also change "Custom driver annotation" to use the name in menu below.
![image](https://github.com/cBioPortal/cbioportal-frontend/assets/840895/e1cc7931-1277-49f3-a993-f51f5719eaab)

PR #10630, opened 2024-02-14
support default_sort_column config for mutation, CNA, and SV tables

7xuanlu:
Fix https://github.com/cBioPortal/cbioportal/issues/10556

### Default sort columns on Mutation, Copy-Number and Structural Variant Tables

Define the column that are going to sort be default in the Mutation, Copy-Number and Structural Variant Tables on the Patient View and the Mutation Table in the Results View.
Column name should be exactly the same as shown in tables.
```
skin.results_view.tables.default_sort_column=
skin.patient_view.tables.default_sort_column=
```


JREastonMarks (approved):
lgtm

PR #10647, opened 2024-02-19
Add MSK-specific CIS function

inodb:
Add a way to hide user login name

### DIFF
--- a/src/main/resources/templates/index.html
+++ b/src/main/resources/templates/index.html
@@ -60,6 +60,18 @@
     </script>
 
     <script>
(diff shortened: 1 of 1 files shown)


inodb:
Hmm - actually it looks like this function exists already in the frontend: https://github.com/cBioPortal/cbioportal-frontend/blob/6cc6ec199285c2dfde5879791c99fc75c053f422/src/appBootstrapper.tsx#L143-L146

Not sure how the "NoReferenceError" shows up in CIS. Gonna close this for now

PR #10688, opened 2024-03-04
javax dependency

dippindots:
introduce javax dependency

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -316,6 +316,11 @@
 			<artifactId>h2</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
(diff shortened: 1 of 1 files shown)


JREastonMarks (commented):
I'm not 100% sure why this needed when we haven't had to bring it in yet.

dippindots:
@JREastonMarks This is related to the issue we see on the M series chip laptop when running v6 `cbioportal-docker-compose`
CC: @alisman

haynescd:
@dippindots How did you test this?

I propose just updating the deps in pom to be scoped to test 
org.testcontainers.mockserver
org.mock-server.mockserver-client-java

haynescd:
@alisman ok, awesome will do. I just don't think the mockserver should even be apart of the production build (which seems to be what pulls that need for the dep in)

This is all strange how its only on m1/m2 chips

PR #10700, opened 2024-03-12
Fix: groups and authorities filter issue

dippindots:
Filtering out empty groups and grantedAuthorities, it can cause issues if grantedAuthorities and groups both contain an empty string


src/main/java/org/cbioportal/security/CancerStudyPermissionEvaluator.java
```diff
@@ -284,6 +285,8 @@ private boolean hasAccessToCancerStudy(Authentication authentication, CancerStud
         // check if user is in study groups
         // performance now takes precedence over group accuracy (minimal risk to caching cancer study groups)
         Set<String> groups = new HashSet(Arrays.asList(cancerStudy.getGroups().split(";")));
```
haynescd, line 287: This is preference, but I would instead just filter out inline of creating the Set

`        Set<String> test = Arrays.stream(cancerStudy.getGroups().split(";"))
            .filter(String::isEmpty)
            .collect(Collectors.toSet());`
            
            
Same for the other one as well... Just incase someone adds something before this statement using groups and misses the removeIf statement

dippindots:
Thanks @kalletlak @haynescd, comments addressed

haynescd (approved):

Issue #10746, opened 2024-04-12
Enable cache for multi-study sessions


jjgao:
Currently cache is only available for single studies. It would be useful to enable it for multi-study sessions too.

PR #10764, opened 2024-04-22
Enable multi-study study-view cache

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/10746


dippindots (approved):
LGTM

alisman:
@kalletlak @jjgao @inodb i think we limited it to single study queries because we worried about permutations of studies overwhelming cache.  probably should have tried it first to see if there was a problem.  anyway, we can merge this but we should perhaps monitor the cache for the public portal to see if this changes things dramatically

alisman (approved):

PR #10777, opened 2024-05-02
Fix mutation-data-count API sync issue

7xuanlu:
Fix Swagger schema response


JREastonMarks (approved):
LGTM
May not be a bad idea to do a once over the API endpoints to make sure that this isn't the only one that needs correction.

inodb:
@7xuanlu Thanks for the fix! Could you do a pass on the other API endpoints as well as Jeremy proposed? Thanks so much!

7xuanlu:
> @7xuanlu Thanks for the fix! Could you do a pass on the other API endpoints as well as Jeremy proposed? Thanks so much!
Sure, I can hop on those. Let's just focus on endpoints whose return type is a list of an array?

PR #10778, opened 2024-05-03
Add missing POC clinical data binning function

onursumer:
Move over the missing functionality from #10265

### DIFF
--- a/src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
+++ b/src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
@@ -1,0 +1,133 @@
+package org.cbioportal.web.columnar;
+
+import org.cbioportal.model.*;
+import org.cbioportal.service.StudyViewColumnarService;
+import org.cbioportal.web.columnar.util.NewClinicalDataBinUtil;
+import org.cbioportal.web.parameter.*;
+import org.cbioportal.web.util.DataBinner;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component
+public class ClinicalDataBinner {
+    private final StudyViewColumnarService studyViewColumnarService;
+    private final DataBinner dataBinner;
+    
+    @Autowired
+    public ClinicalDataBinner(
+        StudyViewColumnarService studyViewColumnarService,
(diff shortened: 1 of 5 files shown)


src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
```diff
@@ -0,0 +1,121 @@
...
+        List<ClinicalDataBinFilter> attributes = dataBinCountFilter.getAttributes();
+        StudyViewFilter studyViewFilter = dataBinCountFilter.getStudyViewFilter();
+
+        if (shouldRemoveSelfFromFilter) {
+            studyViewFilter = NewClinicalDataBinUtil.removeSelfFromFilter(dataBinCountFilter);
+        }
+
+        List<String> attributeIds = attributes.stream().map(ClinicalDataBinFilter::getAttributeId).collect(Collectors.toList());
+
+        // a new StudyView filter to partially filter by study and sample ids only
```
alisman, line 43: @onursumer can you put some comments here explaining why we do this?  I know this is the same logic as legacy but it's not clear to me why we first do this partial filtering.

src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
```diff
@@ -0,0 +1,121 @@
...
+
+        // TODO make sure these keys don't need to be distinct
+        List<String> unfilteredUniqueSampleKeys = unfilteredSamples.stream().map(Sample::getUniqueSampleKey).collect(Collectors.toList());
+        List<String> filteredUniqueSampleKeys = filteredSamples.stream().map(Sample::getUniqueSampleKey).collect(Collectors.toList());
+        List<String> unfilteredUniquePatientKeys = unfilteredSamples.stream().map(Sample::getUniquePatientKey).collect(Collectors.toList());
+        List<String> filteredUniquePatientKeys = filteredSamples.stream().map(Sample::getUniquePatientKey).collect(Collectors.toList());
+
+        // TODO make sure we don't need a distinction between sample vs patient attribute ids here
+        //  ideally we shouldn't because we have patient clinical data separated from sample clinical data in clickhouse
+        List<ClinicalData> unfilteredClinicalDataForSamples = studyViewColumnarService.getSampleClinicalData(partialFilter, attributeIds);
```
alisman, line 60: also, we should comment here why we do this.  i think it's just because we want to show the filtered bins, right?  Just like the deselected pie charts?  lets use this opportunity to better document the business logic

Issue #10554, opened 2024-01-19
Internal API endpoints listed as default


jjgao:
Previously public endpoints were listed as default, now internal ones were. 

https://www.cbioportal.org/api/swagger-ui/index.html

PR #10800, opened 2024-05-17
Set default API springdoc properties to match those prior to v6

inodb:
Adds several defaults for springdoc to match configuration of the cBioPortal API prior to the v6 release:

- Fix https://github.com/cBioPortal/cbioportal/issues/10554. Default listing is alphabetical (so `internal` was shown first). This makes `public` always show up by default. Also remove old springfox properties file and replace it with a springdoc properties file to include defaults of the swagger UI page
- Also add [these parameters](https://github.com/knowledgesystems/knowledgesystems-k8s-deployment/blob/dd92c2cfe31b8e465eb845d2844a5f4ba6fa963b/public-eks/cbioportal-prod/cbioportal_spring_boot.yaml#L183-L186) that allow for backwards compatibility with v2 api-docs locations


haynescd (approved):
lgtm

PR #10823, opened 2024-06-06
Simplify clinical data binning related SQL

onursumer:
- Use clinical data counts derived from the new `clinical_data_derived` table instead of fetching all the actual clinical data
- Simplify NA count logic. Now NA data is actually in the database, so we just need the get that NA count.

### DIFF
--- a/src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
+++ b/src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
@@ -11,7 +11,6 @@ import org.springframework.stereotype.Component;
 
 import java.util.*;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 @Component
 public class ClinicalDataBinner {
@@ -27,6 +26,15 @@ public class ClinicalDataBinner {
         this.dataBinner = dataBinner;
     }
 
+    private List<ClinicalData> convertCountsToData(List<ClinicalDataCount> clinicalDataCounts)
+    {
+        return clinicalDataCounts
+            .stream()
+            .map(NewClinicalDataBinUtil::generateClinicalDataFromClinicalDataCount)
+            .flatMap(Collection::stream)
+            .toList();
+    }
+    
     @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
     public List<ClinicalDataBin> fetchClinicalDataBinCounts(
         DataBinMethod dataBinMethod,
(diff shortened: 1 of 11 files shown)


src/main/java/org/cbioportal/web/util/ClinicalDataBinUtil.java
```diff
@@ -310,17 +310,33 @@ public List<ClinicalDataBin> calculateStaticDataBins(
         List<String> filteredUniqueSampleKeys,
         List<String> filteredUniquePatientKeys
     ) {
-        return NewClinicalDataBinUtil.calculateStaticDataBins(
```
onursumer, line 313: This is to revert the legacy implementation for the legacy endpoint. We can't use `NewClinicalDataBinUtil.calculateDynamicDataBins` for legacy endpoint anymore since the underlying logic changed significantly.

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewClinicalDataBinMapper.xml
```diff
@@ -0,0 +1,86 @@
...
+    <select id="getSampleClinicalDataCountsForBinning">
+        SELECT
+            attribute_name as attributeId,
+            if(attribute_value='', 'NA', attribute_value) AS value,
+            count(value) as count
+        FROM clinical_data_derived
+        <include refid="sampleClinicalDataFromStudyViewFilter" />
+        AND type = 'sample'
+        GROUP BY attribute_name, value
+    </select>
```
onursumer, line 59: These queries are very similar to `getClinicalDataCounts` query. We may be able to directly use that one instead of adding new SQL.

https://github.com/cBioPortal/cbioportal/blob/b436b2b7b90e9fd72cc86d13d02abdffdf586008/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml#L66-L74

src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
```diff
@@ -26,6 +26,26 @@ public ClinicalDataBinner(
...
+                    d.setAttrId(c.getAttributeId());
+                    d.setAttrValue(c.getValue());
+                    data.add(d);
+                }
+                return data;
+            })
+            .flatMap(Collection::stream)
+            .toList();
+    }
+
```
haynescd, line 48: I think atleast for now the cleaner way would to have another function.

private List<ClinicalData> generateClinicalDataFromClinicalDataCounts(ClinicalDataCounts counts) { 
  nasty for loop
}

So then here you can just do this
`clinicalDataCounts
            .stream()
            .map(this::generateClinicalDataFromClinicalDataCounts)
            .flatMap(Collection::stream)
            .toList();`

src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
```diff
@@ -26,6 +26,26 @@ public ClinicalDataBinner(
         this.dataBinner = dataBinner;
     }
 
+    // TODO move this to a utility class?
+    public List<ClinicalData> convertCountsToData(List<ClinicalDataCount> clinicalDataCounts)
```
haynescd, line 30: This should be private

src/main/java/org/cbioportal/web/util/DataBinner.java
```diff
@@ -565,6 +647,16 @@ public Long countNAs(List<Binnable> clinicalData, ClinicalDataType clinicalDataT
         return count;
     }
 
+    /**
+     * Calculate number of clinical data marked actually as "NA", "NAN", or "N/A"
+     */
+    public Long countNAs(List<Binnable> clinicalData) {
```
haynescd, line 653: This is cast to Long and then later it grabs just the intValue(). Do you need the wrapper class?

        bin.setCount(countNAs(clinicalData).intValue());

src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
haynescd:
```suggestion
private List<ClinicalData> generateClinicalDataFromClinicalDataCounts(ClinicalDataCounts counts) {
nasty for loop
}

    clinicalDataCounts .stream() .map(this::generateClinicalDataFromClinicalDataCounts) .flatMap(Collection::stream) .toList();
```

src/main/java/org/cbioportal/web/util/DataBinner.java
onursumer: Probably we don't need a wrapper here. For now, just keeping it consistent with the legacy `countNAs` method.

https://github.com/cBioPortal/cbioportal/blob/de88ff79ad7bed99b706759e9ab0eff8e7d0fe1d/src/main/java/org/cbioportal/web/util/DataBinner.java#L621

We can revisit this once we completely get rid of the legacy functionality.

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -168,10 +171,13 @@
     </select>
     
 
-    <sql id="getCategoricalClinicalDataCountsQuerySample">
+    <sql id="getClinicalDataCountsQuerySample">
```
onursumer, line 174: renamed because this SQL returns both categorical and numeric

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -198,10 +204,13 @@
         value
     </sql>
 
-    <sql id="getCategoricalClinicalDataCountsQueryPatient">
+    <sql id="getClinicalDataCountsQueryPatient">
```
onursumer, line 207: renamed because this SQL returns both categorical and numeric

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -136,26 +136,31 @@
...
             <trim prefix="" prefixOverrides="AND">
                 <if test="dataFilterValue.value eq 'NA'">
-                    AND attribute_value = ''
+                    AND
+                    <include refid="isAttributeValueNA">
+                        <property name="attribute_value" value="attribute_value"/>
+                    </include>
+                </if>
+                <if test="dataFilterValue.start != null || dataFilterValue.end != null">
+                    AND match(attribute_value, '^[\d\.]+$')
```
onursumer, line 149: this ignores edge cases like `<18`, `>90`, etc. there is a separate ticket for this cBioPortal/rfc80-team#32

src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
```diff
@@ -43,53 +51,30 @@ public List<ClinicalDataBin> fetchClinicalDataBinCounts(
...
-        List<ClinicalData> filteredClinicalDataForPatients = studyViewColumnarService.getPatientClinicalData(studyViewFilter, attributeIds);
+        // we use the filtered data to calculate the counts for each bin, we do not regenerate bins for the filtered data
+        List<ClinicalDataCountItem> unfilteredClinicalDataCounts = studyViewColumnarService.getClinicalDataCounts(partialFilter, attributeIds);
+        List<ClinicalDataCountItem> filteredClinicalDataCounts = studyViewColumnarService.getClinicalDataCounts(studyViewFilter, attributeIds);
 
-        Map<String, ClinicalDataType> attributeDatatypeMap = NewClinicalDataBinUtil.toAttributeDatatypeMap(
-            unfilteredClinicalDataForSamples.stream().map(ClinicalData::getAttrId).collect(Collectors.toList()),
-            unfilteredClinicalDataForPatients.stream().map(ClinicalData::getAttrId).collect(Collectors.toList()),
-            Collections.emptyList() // TODO ignoring conflictingPatientAttributeIds for now
+        // TODO ignoring conflictingPatientAttributeIds for now
```
alisman, line 69: is there a ticket to address this?

src/main/java/org/cbioportal/web/util/DataBinner.java
```diff
@@ -45,10 +47,43 @@ public <T extends DataBinFilter> List<DataBin> calculateClinicalDataBins(T dataB
...
+            filteredClinicalData,
+            countNAs(filteredClinicalData)
+        );
+    }
+
+    public List<DataBin> recalcBinCount(
+        List<DataBin> dataBins,
+        ClinicalDataType clinicalDataType,
+        List<Binnable> clinicalData,
+        List<String> ids
```
alisman, line 73: can you improve the variable name "ids" to something more decsriptive?

src/main/java/org/cbioportal/web/columnar/util/NewClinicalDataBinUtil.java
```diff
@@ -109,33 +98,48 @@ public static List<ClinicalDataBin> calculateDynamicDataBins(
...
-                    .collect(toList());
+                    .toList();
                 clinicalDataBins.addAll(dataBins);
             }
         }
 
         return clinicalDataBins;
     }
+
+    /**
```
alisman, line 124: comment on why we are doing this kinda strange operation (for optimization and to leverage legacy bin counting logic)

src/main/java/org/cbioportal/web/columnar/ClinicalDataBinner.java
onursumer: Yes, there is. cBioPortal/rfc80-team#23. However, now we use the clinical data counts query as is. If we still need to handle the conflicting attributes, I think it needs to be done within the clinical data counts SQL.

src/main/java/org/cbioportal/web/util/DataBinner.java
onursumer: I think this one is now only used by the legacy endpoint. The new one doesn't need to deal with sample/patient ids anymore. We should probably add comments to the methods that are now only needed for the legacy endpoint so that we can remove them later once we completely migrate to the new implementation.

alisman (dismissed):

haynescd (approved):
👍

PR #10824, opened 2024-06-06
Feature/mutated genes total profiled counts and gene panels

haynescd:
Add Total Profiled counts  and matching Gene Panel ids for mutated genes


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -165,4 +163,68 @@
...
+            sample_unique_id IN ( <include refid="sampleUniqueIdsFromStudyViewFilter"/>
+            <if test="applyPatientIdFilters == true">
+                INTERSECT <include refid="getSampleIdsFromPatientIds"/>
+            </if>
+            )
+        </where>
+        GROUP BY gptg.gene;
+    </select>
+
+    <!-- Helper query to get counts for totalProfiledCount when no genePanel data is available-->
```
alisman, line 186: add in comment the detail that when no genePanel we consider it whole genome sequencing?

PR #10825, opened 2024-06-07
Add bean for CancerStudyPermissionEvaluator

Floris-Hyve:
Add a bean for `CancerStudyPermissionEvaluator` instead of instantiating directly within `createExpressionHandler`

### DIFF
--- a/src/main/java/org/cbioportal/security/CancerStudyPermissionEvaluator.java
+++ b/src/main/java/org/cbioportal/security/CancerStudyPermissionEvaluator.java
@@ -44,7 +44,6 @@ import org.cbioportal.persistence.cachemaputil.CacheMapUtil;
 import org.cbioportal.utils.security.AccessLevel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.*;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.authority.AuthorityUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/org/cbioportal/security/CancerStudyPermissionEvaluator.java
```diff
@@ -45,9 +45,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.*;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
```
haynescd, line 48: Can you remove unused imports

haynescd (changes requested):

src/main/java/org/cbioportal/security/CancerStudyPermissionEvaluator.java
dippindots: @Floris-Hyve I'm thinking the same thing, there are some unused imports, if we just need to change the way to create CancerStudyPermissionEvaluator, we don't need additional imports.

Floris-Hyve: Ah yes thanks for pointing out, removed them! This one seems unused also 
```import org.springframework.beans.factory.annotation.*;```

dippindots (approved):
LGTM

haynescd (approved):
LGTM

inodb (commented):
@Floris-Hyve Thanks for trying to solve this! It's unclear why changing how the bean is instantiated would solve an authorization issue. I don't see any issue with merging this but would be good to first update title+description to explain the reasoning

inodb (changes requested):
see comment re title/desciption

Floris-Hyve:
I think there are issues with spring recognizing the `CancerStudyPermissionEvaluator` when it is not instantiated as a bean, which causes the `expressionHandler` to be set incorrectly. I think its related to how dependency injection works in spring. See below screenshots:
- Without instantiating as bean, study is accessible to a user without permission (can explore it too)
image

- As bean, study not accessible
image

PR #10827, opened 2024-06-11
Implement custom-data-bin-counts endpoint

gblaih:


src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
```diff
@@ -419,5 +419,37 @@ public ResponseEntity<SampleTreatmentReport> fetchSampleTreatmentCounts(
         return new ResponseEntity<>(studyViewColumnarService.getSampleTreatmentReport(interceptedStudyViewFilter),
             HttpStatus.OK);
     }
-
+
```
alisman, line 422: @gblaih should this be commented out?

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -66,6 +66,45 @@
                     '${sampleIdentifier.studyId}_${sampleIdentifier.sampleId}'
                 </foreach>
             </if>
+            <if test="studyViewFilter.customDataFilters != null and !studyViewFilter.customDataFilters.isEmpty() and customDataSamples != null">
```
alisman, line 69: i'm not sure i understand how this customDataFilters collection (which seems to be a member of the studyViewFilter) maps back to the extracted customDataFilters which has been added as an extra parameter to the mapper methods.

alisman:
@haynescd @gblaih, as I discussed with Bryan, I would like to pursue a way to avoid passing customDataSamples through to all the methods.   This strikes me as something that should be dealt with using interception. 

I prospose the following:  we change the StudyViewFilter type to add a collection of custom data.   The intereceptor invokes ```extractCustomDataSamples``` and populates this collection with the result.   It will then automatically be available in the mappers without having to be passed through.  

We might want to consider extending or wrapping StudyViewFilter if we feel that this collection of customData is not native to a "filter object".

PR #10847, opened 2024-06-20
Do not use clinical data MVs

onursumer:
- [ ] use `clinical_data_derived` table when filtering both numerical and categorical clinical data
- [ ] use `clinical_attribute_meta` table to group clinical attribute names
- [ ] remove redundant clinical data count methods and SQL

### DIFF
--- a/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
+++ b/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
@@ -26,8 +26,6 @@ import java.util.stream.Collectors;
 @Service
 public class StudyViewColumnarServiceImpl implements StudyViewColumnarService {
 
-    private final Map<String, List<String>> clinicalAttributeNameMap = new HashMap<>();
-
 
     private final StudyViewRepository studyViewRepository;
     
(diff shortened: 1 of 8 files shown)


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -253,10 +213,17 @@
...
-        DISTINCT(attribute_name)
-        FROM ${tableName};
+            attr_id as attrId,
+            display_name as displayName,
+            description,
+            datatype as dataType,
+            patient_attribute as patientAttribute,
+            cancer_study_id as cancerStudyId,
+            cs.cancer_study_identifier cancerStudyIdentifier
+        FROM sling_db_2024_05_23_original.clinical_attribute_meta cam
```
haynescd, line 225: you have the specific db name here. This will fail when we switch the db out

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -108,12 +109,35 @@ private CategorizedClinicalDataCountFilter extractClinicalDataCountFilters(final
...
+                numericKey,
+                clinicalAttributes
+                    .stream()
+                    .filter(a -> 
+                        a.getDatatype().equals(ClinicalAttributeDataType.NUMERIC.getValue()) && 
+                            isPatientAttr == a.getPatientAttribute()
+                    )
+                    .map(ClinicalAttribute::getAttrId)
+                    .toList()
+            );
```
haynescd, line 140: I would say all this not needed anymore with your call to studyViewRepository.getClinicalAttributes() now.

We originally needed this because there were 4 separate tables for clinicalAttribute.

Now CategorizedClinicalDataFilter can be pushed down into the repository layer and have one call that grabs all clinicalAttribute_meta data. Then we can split patient and sample filtering

haynescd (dismissed):

PR #10849, opened 2024-06-20
Reimplement sample-lists-counts endpoint using Clickhouse

alisman:


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -121,6 +121,21 @@
         GROUP BY genetic_profile.stable_id, genetic_profile.name, sample_mv.cancer_study_identifier;
     </select>
     
+
+    <!-- for /sample-lists-counts/fetch (returns CaseListDataCount) -->
+    <select id="getCaseListDataCounts" resultType="org.cbioportal.model.CaseListDataCount">
```
haynescd, line 126: Have you profiled this query?

Does it make sense to use the sample_list_columnstore?

alisman: interesting.  i wonder who made sample_list_columnstore?    it's already subsecond as is but might as well use this.

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -32,6 +33,7 @@ public StudyViewColumnarServiceImpl(StudyViewRepository studyViewRepository, Alt
         this.alterationCountService = alterationCountService;
     }
 
+    @Cacheable(cacheResolver = "generalRepositoryCacheResolver", condition = "@cacheEnabledConfig.getEnabled()")
```
haynescd, line 36: Is this cacheable? I thought we weren't doing cacheable

haynescd (dismissed):

PR #10867, opened 2024-06-26
Consolidate SQL files used to create CH Tables

haynescd:


alisman (approved):

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -31,13 +14,40 @@ CREATE TABLE sample_to_gene_panel
 ) ENGINE = MergeTree()
 ORDER BY (gene_panel_id, alteration_type, sample_unique_id);
 
-CREATE TABLE gene_panel_to_gene
+INSERT INTO sample_to_gene_panel_derived
+SELECT
+    concat(cs.cancer_study_identifier, '_', sample.stable_id) AS sample_unique_id,
+    genetic_alteration_type AS alteration_type,
+    -- If a mutation is found in a gene that is not in a gene panel we assume Whole Exome Sequencing WES
+    ifnull(gene_panel.stable_id, 'WES') AS gene_panel_id,
```
sheridancbio, line 22: Recently I looked into this during the fixing of the 4 non-panel but profiled genie samples. According to the docs, the value 'NA' is specified to be used "When the sample is not profiled on a gene panel, or if the sample is not profiled at all"
So I'm not sure whether NA is supposed to be used for whole exome sequencing or not ... but it sounds like maybe. But then I guess the 'NA' value would pass through in this logic. I do wonder whether having a mixture of 'WES' and 'NA' would be confusing in the downstream logic. Perhaps our docs should be changed to reserve 'NA' only for cases where a sample was not profiled and to reserve 'WES' or something similar for whole exome sequencing.
https://docs.cbioportal.org/file-formats/#values

This comment applies to several tables below as well.

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -31,13 +14,40 @@ CREATE TABLE sample_to_gene_panel
...
+    g.hugo_gene_symbol AS gene
+FROM gene_panel gp
+         INNER JOIN gene_panel_list gpl ON gp.internal_id = gpl.internal_id
+         INNER JOIN gene g ON g.entrez_gene_id = gpl.gene_id
+UNION ALL
+SELECT
+    'WES' AS gene_panel_id,
+    gene.hugo_gene_symbol AS gene
+FROM gene
+WHERE gene.entrez_gene_id > 0;
```
sheridancbio, line 49: some of our databases use negative entrez_gene_id values (mainly for microRNAmature/precursor combinations related to expression profiling). I'm not sure if that matters here. Maybe a note should be added about the possible future need to support negative values? Or maybe those are intentionally being excluded here.

sheridancbio (approved):
Looks good, and just in time (I will need to create these views soon on the new version of the genie database, so I can run this).

I point out a couple of concerns, but don't think they need to hold up this work.

src/main/resources/db-scripts/clickhouse/clickhouse.sql
haynescd: Yea, this isn't finalized and would be interested jumping on a call and discussing this in the future. To figure out what the best approach is

PR #10884, opened 2024-07-08
Restrict proxy access to known hosts only

onursumer:
- Restrict proxy access to known hosts only
- Remove unused endpoints and methods
- Rename dev endpoint and add comments

Need to merge cBioPortal/cbioportal-frontend#4941 and cBioPortal/cbioportal-frontend#4942 before merging this one.

### DIFF
--- a/src/main/java/org/cbioportal/proxy/ProxyController.java
+++ b/src/main/java/org/cbioportal/proxy/ProxyController.java
@@ -13,7 +13,6 @@ import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.converter.StringHttpMessageConverter;
 import org.springframework.security.core.Authentication;
-import org.springframework.util.ObjectUtils;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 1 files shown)


src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -226,54 +213,30 @@ private <T> ResponseEntity<T> exchangeData(String body, URI uri, HttpMethod meth
         return restTemplate.exchange(uri, method, new HttpEntity<>(body, httpHeaders), responseType);
     }
 
+    private boolean isKnownHost(String pathInfo) {
+        String[] knownHosts = {
+            this.genomeNexusUrl,
+            this.genomeNexusGrch38Url,
+            "http://mutationaligner.org/api",
```
inodb, line 220: do we still use mutationaligner?

src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -226,54 +213,30 @@ private <T> ResponseEntity<T> exchangeData(String body, URI uri, HttpMethod meth
...
+            "http://docs.cbioportal.org/news"
```
inodb, line 221: should we fix cors on the docs site?

src/main/java/org/cbioportal/proxy/ProxyController.java
inodb: Looks like it, just to check for existence of this URL:

image

onursumer: We decided to always show the link. See cBioPortal/cbioportal-frontend#4942

src/main/java/org/cbioportal/proxy/ProxyController.java
onursumer: Looks like Github pages now allows CORS! See cBioPortal/cbioportal-frontend#4941

src/main/java/org/cbioportal/proxy/ProxyController.java
```diff
@@ -49,6 +42,12 @@ public class ProxyController {
     
     @Value("${oncokb.public_api.url:https://public.api.oncokb.org/api/v1}")
     private String oncokbApiUrl;
+
+    @Value("${genomenexus.url.grch38:}")
+    private String genomeNexusGrch38Url;
+
+    @Value("${genomenexus.url:}")
+    private String genomeNexusUrl;
```
haynescd, line 50: Where are these fields used?

haynescd (changes requested):

haynescd (approved):
👍

PR #10889, opened 2024-07-10
Fix sorting in clinicaldata tab of study view

alisman:
Sorting on clinical data tab of study view was not working when sorted by patient or sample ids


haynescd (approved):
LGTM

PR #10893, opened 2024-07-15
Switch off spring security on public VS endpoints

forus:
We are making this change to simplify publishing and unpublishing by using only our custom authorization token, without spring security tokens. The same way the `/api/cache` call does.
https://github.com/cBioPortal/cbioportal/blob/5c4aab003028c55f610cdf9c8a1863550b52203c/docs/Create-And-Publish-Virtual-Study.md?plain=1#L26-L34

cBioPortal does not have an admin role atm. Having the token makes you an admin.


pieterlukasse (approved):

pieterlukasse:
Looks good to me 👍 
It follows the security pattern of `/api/cache`, so I understand we're _not_ introducing a new pattern.

PR #10908, opened 2024-07-25
Fixed outdated keycloak docker image

### LINKED ISSUE #10896: Regarding keycloak deployment
https://docs.cbioportal.org/deployment/docker/using-keycloak/ 

Instructions here mention a no longer available version of keycloak's docker. (as far as i could find).

jboss/keycloak:4.8.3.Final

On quay.io, the oldest working version is:  
quay.io/keycloak/keycloak:9.0.3

Is there any updated version of this guide? I ran into some issues while configuring keycloak through the GUI (at the SAML part) and i wonder if the recommended versions and instructions changed for the newer versions of cbioportal.

ugurkoysuren:
Changed the Docker Image from jboss image to the actual keycloak image since jboss image is not maintained anymore

- Changed the Docker Image from jboss image to the actual keycloak image since jboss image is not maintained anymore

### DIFF
--- a/docs/deployment/docker/using-keycloak.md
+++ b/docs/deployment/docker/using-keycloak.md
@@ -1,5 +1,7 @@
 # Authenticating and Authorizing Users using Keycloak in Docker
 
+**⚠️ This documentation for keycloak might be outdated, see related [ticket](https://github.com/cBioPortal/cbioportal/issues/10360) ⚠️**
+
 This guide describes a way to Dockerise Keycloak along with cBioPortal, for authentication.
 
 First, create an isolated network in which the Keycloak and MySQL servers can talk to one another.
--- a/docs/deployment/docker/using-keycloak.md
+++ b/docs/deployment/docker/using-keycloak.md
@@ -22,7 +22,7 @@ docker run -d --restart=always \
     mysql:5.7
 ```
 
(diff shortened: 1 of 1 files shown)


inodb:
Thanks so much for this fix @ugurkoysuren !

One issue is that the docs will also need to be updated to be in sync: https://github.com/cBioPortal/cbioportal/blob/master/docs/deployment/authorization-and-authentication/Authenticating-and-Authorizing-Users-via-keycloak.md. Is this something you could help with?

gl00ten:
This additionally changes keycloak from version 4 to version 25. This is a huge increment. Did you confirm this much newer version of keycloak works? 
I wasn't able to make the cbioportal installation work with keycloak versions 8,9,10 and 25. 
Someone else confirm before merging this.

inodb:
Thanks @gl00ten ! @hweej tried v25 and ran into some issues that will require some code changes in the cBioPortal backend

ugurkoysuren:
Good Point, but the maybe we should stay in the same keycloak version and just change the provider first. Since jboss image cannot be pulled at all right now. @inodb

ugurkoysuren:
I tested it with the same version and changin the provider, it is currently working, needed to change the path as well.

Can someone review the PR and test it?

https://github.com/cBioPortal/cbioportal-docker-compose/pull/25

inodb (approved):
@ugurkoysuren Thanks so much for fixing this! I didn't get a chance to test it, but it anyway seems better than the current situation with a non-existent image

If you're able to provide another PR with updated keycloak documentation that would be super helpful!

PR #10911, opened 2024-07-26
Fix alteration count services

haynescd:
Fix Alteration Count Service by preventing a join on 'WES' gene panel. This would create a HUGE join with (TCGA) 625K samples * 20K Genes.

### DIFF
--- a/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
+++ b/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
@@ -232,7 +232,13 @@
         JOIN cancer_study cs on cs.cancer_study_id = cam.cancer_study_id
     </select>
 
-    <!-- Grab Total Profiled Counts -->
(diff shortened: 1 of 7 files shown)


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -233,22 +233,34 @@
...
-        SELECT
+    <select id="getTotalProfiledCounts" resultType="org.cbioportal.model.AlterationCountByGene">
+        SELECT 
         gene as hugoGeneSymbol,
         COUNT(*) as numberOfProfiledCases
         FROM sample_to_gene_panel_derived stgp
             INNER JOIN gene_panel_to_gene_derived gptg on stgp.gene_panel_id = gptg.gene_panel_id 
         <where>
-            stgp.alteration_type = '${alterationType}' 
+            stgp.alteration_type = '${alterationType}'
```
alisman, line 243: could use comment explaining why NOT WES

PR #10921, opened 2024-08-01
RFC87: add download_custom_buttons_json to application.properties schema and update documentation

pappde:
# Summary

This is the backend update for frontend PR https://github.com/cBioPortal/cbioportal-frontend/pull/4938.  This PR addresses the backend configuration requirement for [RFC87](https://docs.google.com/document/d/1aVs9uva7OEqoxOZLflsFbo3c4_uogKpDH8uLEVNmF9k/edit#heading=h.btohl122j9ni), which involves defining JSON to customize the frontend to display custom buttons for data tables.

This PR includes documentation for how to define your own custom button, and also includes the AVM configuration for RFC87. 

# Enabling Custom Buttons

By default, there are no custom buttons defined. The following can be added to an instance's application.properties to enable it:

```
download_custom_buttons_json=classpath:/custom-buttons/download_custom_button_avm.json
```

# Changes

## Code

- FrontendPropertiesServiceImpl.java - updated so that the property is treated as a reference to a JSON

## Documentation
- application.properties.EXAMPLE
- application.properties-Reference.md: added download_custom_buttons_json and link to new document
- download_custom_buttons-Reference.md: new document describing how to configure custom JSON

## Resources
- resources/custom_buttons/download_custom_button_avm.json: added, but unreferenced by default

# Notify reviewers
@alisman @jjgao


docs/deployment/customization/application.properties-Reference.md
```diff
@@ -798,3 +798,10 @@ is on by default but can be disabled using the following property.
 //boolean
 enable_study_tags=true|false
 ```
+
+# Add Custom Buttons to data tables
+Custom Buttons can be defined which will conditionally appear in all group comparison data tables (with CopyDownloadControls) to launch a custom URL. This can be used, for example, to launch a software application (that is installed on the user's system) with the data. This configuration can also customize new elements on the Visualize page. It points to a JSON file on the classpath. (See [custom_buttons reference](/deployment/customization/custom_buttons-Reference.md)).
```
pappde, line 803: I noticed an inconsistency. When you are viewing the documentation directly in github, these links (to /deployment) don't work, because the documents actually are under "/docs/deployment/....". Example, 
- https://github.com/cBioPortal/cbioportal/blob/master/docs/deployment/customization/application.properties-Reference.md is the correct URL
- https://github.com/cBioPortal/cbioportal/blob/master/deployment/customization/application.properties-Reference.md is where cross-links try to go (giving a 404 error)

When you view the documentation under "docs.cbioportal.org" the links work, because /docs/ is the root and does not appear in the URL path at all.  Example,
-  https://docs.cbioportal.org/deployment/customization/application.properties-reference/

This also means that if you change the link to "/docs/deployment/..." then it will work in the first scenario, but not in the live scenario.  

Not sure how to fix.  It makes it difficult to test the documentation cross-linking prior to it actually going live.

src/main/java/org/cbioportal/service/FrontendPropertiesServiceImpl.java
```diff
@@ -234,6 +234,7 @@ private String getPropertyValue(FrontendProperty property) {
             case "query_sets_of_genes":
             case "skin.patient_view.custom_sample_type_colors_json":
             case "oncoprint.clinical_tracks.config_json":
+            case "custom_buttons_json":
```
pappde, line 237: Since the property refers to a JSON file, we need to add it to the readFile() case.

src/main/resources/custom-buttons/custom_button_avm.json
```diff
@@ -0,0 +1,15 @@
...
+        "image_src": "https://aquminmedical.com/images/content/AquminLogoSimple.png",
+        "required_user_agent": "Win",
+        "required_installed_font_family": "AVMInstalled",
+        "url_format": "avm://?importclipboard&-AutoMode=true&-ProjectNameHint={studyName}&-ImportDataLength={dataLength}",
+        "visualize_title": "AVM for cBioPortal (Windows)",
+        "visualize_href": "https://bit.ly/avm-cbioportal",
+        "visualize_description": "Windows software that loads data into 3D Landscapes for interactive visualization and pathway analysis. Download table data directly from cBioPortal.",
+        "visualize_image_src": "https://github.com/user-attachments/assets/5c17f5ed-0357-4ffa-a6e1-5a9d435dd3c5"
+    }
+]
```
pappde, line 15: Included in the "resources" folder so it can be easily enabled with one line (per the EXAMPLE)

docs/deployment/customization/custom_buttons-Reference.md
```diff
@@ -0,0 +1,72 @@
+# Custom Buttons for Data Tables
```
alisman, line 1: i think we need to make this more specific.   custom_buttons_json doesn't imply group_comparison.   i guess ```group_comparison_custom_buttons```?

pappde: Ok, as to exactly what name: technically, it may appear anywhere that the CopyDownloadControls appear (with ShowDownload==true). I couldn't find an official name for the tables where these controls appear in this context, so I have been calling them "group comparison", but realize it could be a misnomer.  In the RFC, we also used the term "LazyMobX tables", but that doesn't seem right either.

@alisman preference?
A) download_custom_buttons_json
B) group_comparison_custom_buttons_json
C) data_custom_buttons_json

alisman:
If they can appear anywhere that the download controls component is used,
then download_custom_buttons_json

pappde:
Renamed property to download_custom_buttons_json

docs/deployment/customization/custom_buttons-Reference.md
pappde: using download_custom_buttons_json

src/main/resources/application.properties.EXAMPLE
```diff
@@ -426,5 +426,8 @@ spring.devtools.restart.enabled=false
 ## Study Tag functionality
 #enable_study_tags=true|false
 
+## Custom Buttons
+# download_custom_buttons_json=classpath:/custom_buttons/download_custom_button_avm.json
```
pappde, line 430: left resource path as "custom_buttons" in case other types are added later (less clutter with multiple folders)

alisman: @pappde the example should be commented out so that when people copy this example file, they don't inadvertantly ...

pappde: @alisman Yes, off by default, but easy to enable, and also on in the public instance.  For this reason, the **application.properties.EXAMPLE** has the the following line:

```
# download_custom_buttons_json=classpath:/custom_buttons/download_custom_button_avm.json
```

It is commented out here, but:

- Private Instance: can easily be enabled by copying and removing the #
- Public Instance: the line (uncommented) would be added to the application.properties

alisman: ah.  i misread this line, not seeing the hash above it (it's wrapping in review interface)

docs/deployment/customization/application.properties-Reference.md
inodb: Would always using relative URLs solve the issue (e.g. `./../deployment/`)?

pappde: Yes, I can confirm that works when viewing document in github.  I can't verify that it works under docs.cbioportal.org but it should. I'll submit a new PR that updates the links in this file to be relative

pappde: added PR #10938 to address this file and a few others.

PR #10935, opened 2024-08-12
Add tests for case list sample counts endpoint

alisman:


src/test/java/org/cbioportal/persistence/mybatisclickhouse/CaseListSamplesCountsTest.java
```diff
@@ -0,0 +1,186 @@
...
+public class CaseListSamplesCountsTest extends AbstractTestcontainers {
+    
+    private static final String STUDY_TCGA_PUB = "study_tcga_pub";
+    private static final String STUDY_ACC_TCGA = "acc_tcga";
+    
+    @Autowired
+    private StudyViewMapper studyViewMapper;
+    
+    @Autowired
+    private StudyViewMyBatisRepository studyViewMyBatisRepository;
```
alisman, line 39: @haynescd this is where it's throwing error.  doesn't want to autowire this

src/main/java/org/cbioportal/persistence/mybatisclickhouse/StudyViewMyBatisRepository.java
```diff
@@ -107,11 +109,35 @@ public Map<String, ClinicalDataType> getClinicalAttributeDatatypeMap() {
         
         return attributeDatatypeMap;
     }
+    
+    public static List<CaseListDataCount> mergeCaseListCounts(List<CaseListDataCount> counts) {
```
alisman, line 113: @haynescd i'm assuming that this should go in some util?  let me know the right place or if this is ok to just have it hanging out here.
FYI, i don't love that this merging is done at repository level because it suggest that the result of the underlying mapper is legitimate.  it's not.  the result must be merged. but i can't figure out a way to do that operation in SQL.  ideally, it would be done in the mapper, but the mapper methods have no java implementation so ... what to do?

haynescd: So, I would suggest doing this at the service layer.

Creating a class CaseListService? Or whatever where this logic lived.


And the Repository would return getCaseListDataCountsPerStudy()

That way the mapper is still correct and you can do the business logic of merging in the service layer

haynescd:
Small unused import but everything else looks good

haynescd (approved):
lgtm

Issue #10937, opened 2024-08-13
Documentation - absolute links do not work in Github repo browser [Development issue]


pappde:
## REPRO
1. go to a markdown document in the repo using the github website. E.g. https://github.com/cBioPortal/cbioportal/blob/master/docs/Architecture-Overview.md
2. note the URL path is {branchname}/docs/....
3. click on a markdown link to another document. E.g.
``` 
[OncoKB Data Access](/deployment/integration-with-other-webservices/OncoKB-Data-Access.md)
```
4. observe a 404 error as it is trying to open:

```
https://github.com/cBioPortal/cbioportal/blob/master/deployment/integration-with-other-webservices/OncoKB-Data-Access.md
```

## EXPECTED
4. it opens the correct URL with the /docs/: 
```
https://github.com/cBioPortal/cbioportal/blob/master/docs/deployment/integration-with-other-webservices/OncoKB-Data-Access.md
```

## NOTE
All of these links work fine in the documentation website (https://docs.cbioportal.org) since the domain name is the root.  For example, following the above path, you would end up at the correct page:
```
https://docs.cbioportal.org/deployment/integration-with-other-webservices/oncokb-data-access/
```

## PROPOSED SOLUTION

Changing all these links to relative should fix it so it works in both the github repo and the documentation website. E.g. since the Architecture Overview is in the root:
``` 
[OncoKB Data Access](integration-with-other-webservices/OncoKB-Data-Access.md)
```

pappde:
Added PR #10938 to address the files under /deployment/customization/

dippindots:
Issue is solved by connecting pr

PR #10938, opened 2024-08-13
Deployment/Customization Docs: changed links to relative so they work both in github website and docs website

pappde:
This is a partial fix to issue #10937, where cross-links using absolute paths give 404 errors in the github repo browser. It updates the cross-links in the /docs/deployment/customization/  documents.

# Checks
- Verified these updates in the github repo browser
- NOTE: cannot verify in documentation website, but it should work since I believe there are already a couple of cases where relative links are used.


inodb (approved):
Thanks @pappde !

PR #10943, opened 2024-08-15
Normalize clinical data

onursumer:
Fix cBioPortal/rfc80-team#40

Generalize the `normalizeAttributeValueNA` to include boolean values as well.


alisman (commented):
@onursumer do we know if this covers all the normalizations of the legacy codebase?

PR #10944, opened 2024-08-15
RFC87: updated application.properties example and documentation so works in default docker config

pappde:
Updated the application.properties example and download_custom_buttons_json documentation to a configuration that will work in the default docker config. Specifically, changed from using "classpath:" to 
```
download_custom_buttons_json=/cbioportal-webapp/custom_buttons/download_custom_button_avm.json
```

# Explanation
It seems that the "classpath:" syntax doesn't work with the readFile() method used by FrontendProperties. This commit changes the example to use the absolute path that matches how the default docker config is set (files installed to /cbioportal-webapp/).

# Further Improvements
Ideally, either a relative path or "classpath:" would work for the application.properties. Then all configurations would work with a copy/paste of the reference.

# Checks
Tested on local docker running 6.016.

# Notify reviewers
@alisman


pappde:
Putting this back in draft mode, since the instructions may not work as is for private instances. 

# ISSUE
Turns out that, at least for the public instance, the docker container uses the "/docker/web/Dockerfile" configuration which deploys everything as an app.jar. So readFile() will fail, since it looks for loose files on the filesystem.

# OPTIONS
Most likely, either

A) readFile() needs to be updated to be able to load using "classpath;" syntax, which appears to be the intent of the documentation/examples.   See getResourceStream() for an example of using the ClassLoader, which should work for the app.jar.

B) config files like this need to be deployed to be loose on the server (outside the app.jar). For example, modifying /docker/web/Dockerfile like this:
```
COPY --from=build ${DEPENDENCY}/BOOT-INF/classes/custom_buttons             $PORTAL_WEB_HOME/custom_buttons  
```

pappde:
closing - going to replace with new PR that fixes the issue

PR #10946, opened 2024-08-16
fix frontendproperties readFile() to work with classpath and update download_custom_buttons_json reference

pappde:
This fixes the readFile() method which is used by frontendproperties (application.properties) to read the contents of a file in, to work with the "classpath:" syntax.  This PR also updates the reference config for download_custom_buttons_json to the classpath syntax so it works.

Properties affected:
- frontendConfigOverride
- query_sets_of_genes
- skin.patient_view.custom_sample_type_colors_json
- oncoprint.clinical_tracks.config_json

# Explanation
The existing documentation for these properties mentioned that "classpath:" syntax could be used.  E.g. from the application.properties.EXAMPLE, I don't believe this would work:

```
DOES NOT WORK
oncoprint.clinical_tracks.config_json=classpath:/oncoprint-default-tracks.json
```

However, the readFile() method didn't support that syntax.  It would only work with the file system. Furthermore, production instances (at least the public cBioPortal) build the server to a single "app.jar" file (using the "web" docker config).  If the above file exists in the "web-and-data" loose configuration via the repo, but no longer exists as a loose file in the "web" configuration, then you have configuration that works in dev but fails in production.
```
WORKS IN DEV BUT NOT PRODUCTION
download_custom_buttons_json=/cbioportal-webapp/custom_buttons/download_custom_button_avm.json
```

That means that these properties only worked if they had an absolute path reference to a file, and the file is manually placed on the file system of the server (and not packed into the app.jar). E.g.
```
WORKS IF THE FILE IS MANUALLY PLACED ON THE SERVER:
download_custom_buttons_json=/somefolder/download_custom_button_avm.json
```

# Checks
[DONE] I have tested the docker "web-and-data" container with loose files. All 3 of these variants work successfully in that scenario:

```
WORKS NOW:
download_custom_buttons_json=/cbioportal-webapp/custom_buttons/download_custom_button_avm.json
download_custom_buttons_json=custom_buttons/download_custom_button_avm.json
download_custom_buttons_json=classpath:custom_buttons/download_custom_button_avm.json
```

[TODO] This needs to be tested in the "web" container with just the app.jar.  The "classpath" syntax should work, since this pattern is used in other places, but I have been unable to test this image to confirm. I get an error 
```
Error: Could not find or load main class org.cbioportal.PortalApplication
Caused by: java.lang.ClassNotFoundException: org.cbioportal.PortalApplication: 
```

[TODO] This needs to be regressed to make sure that existing instances that use these properties and are referencing loose files this way still work in the "web" container with just app.jar. For example, if there is an instance that uses something like this:

```
VERIFY STILL WORKS:
oncoprint.clinical_tracks.config_json=/somefolder/oncoprint-default-tracks.json
```

This should work based on the above tests with the "web-and-data" container, but to be safe, it should be tested with the "web" container.

# Reviewers
@alisman

### DIFF
--- a/src/main/java/org/cbioportal/service/FrontendPropertiesServiceImpl.java
+++ b/src/main/java/org/cbioportal/service/FrontendPropertiesServiceImpl.java
@@ -10,6 +10,7 @@ import org.springframework.stereotype.Service;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


pappde:
As noted above, need to figure out how to locally test the "app.jar" docker container to make sure that everything will work in the final deployment config, but running into issues testing that. Building it like this:

```
docker build -t cbioportal/cbioportal:my-dev-image -f docker/web/Dockerfile .
```

pappde:
confirmed that the new version of readFile works with app.jar resource and with the file system. Added logging that adds visibility to what files are loaded by this code, and from where (file or jar).  

app.jar resources (i.e. repo files) can be referenced both with relative paths, or with "classpath:{relativePath}". The "classpath:" prefix is not necessary.

src/main/java/org/cbioportal/service/FrontendPropertiesServiceImpl.java
```diff
@@ -291,18 +295,63 @@ private Map<String,String> cloneProperties() {
...
+                    log.info("Reading frontend config resource: {}", filePath);
+                } else {
+                    throw new Exception("File not found in system or classpath: " + filePath);
+                }
             }
+
+            // read the file and return the content as a single-line string
+            // PRE: inputStream != null
+            BufferedReader br = new BufferedReader(new InputStreamReader(inputStream));
+            return br.lines().map(String::trim).collect(Collectors.joining(""));
```
haynescd, line 348:
```suggestion
         Optional<File> file = attemptToFindFile(filePath);
         try (BufferedReader br = Files.newBufferedReader(Paths.get(fileName))) {
                return br.lines().map(String::trim).collect(Collectors.joining(""));
            } catch (Exception e) {
                log.error("Error reading frontend config file: {}", e.getMessage());
                return null;
            }
```

Good catch. Can we wrap the logic into a separate function that is responsible for checking if the file exists?

haynescd (changes requested):
Great Find/Fix! Just small changes requested

src/main/java/org/cbioportal/service/FrontendPropertiesServiceImpl.java
haynescd: Suggestion doesn't build, just an example

pappde: Yes, that's a good idea since the function is long. Will update.

pappde:
> Great Find/Fix! Just small changes requested

Added a commit that splits into an additional method "locateFile". It returns an InputStream, either from a "File" or a Resource inside the jar.

haynescd (approved):
lgtm

PR #10951, opened 2024-08-20
Improve numerical clinical data filter

onursumer:
Fix cBioPortal/rfc80-team#32
Fixes https://github.com/cBioPortal/rfc80-team/issues/44

Handle numerical clinical data values such as `<18`, `>90`, `<=3`, `>=9` when range filtering.

This PR also improves the numerical value matching. Previously only positive integers could be matched, with this PR negative values as well as decimal numbers are also matched.

### DIFF
--- a/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
+++ b/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
@@ -618,6 +618,29 @@
         </where>
     </sql>
     
+    <!--
+        Convert string values to a single numerical value for filtering purposes only.
+        Not designed as a general purpose decimal number parser.
+        
+        Examples conversions:
+        '6' to 6
+        '>=6' to 6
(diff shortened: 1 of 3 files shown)


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -620,6 +620,29 @@
...
+        Not designed as a general purpose decimal number parser.
+        
+        Examples conversions:
+        '6' to 6
+        '>=6' to 6
+        '<=6' to 6
+        '>6' to 6.00..1...
+        '<6' to 5.99..9...
+    -->
+    <sql id="castStringValueToFloat">
```
alisman, line 634: @onursumer @haynescd if it turns out this is a performance killer, we could precompute this right?

haynescd: Yea, you could do this when you create the table

PR #10953, opened 2024-08-21
Docker: improve instructions for testing the backend with docker

pappde:
This is a documentation-only update that improves the "Deployment with Docker" instructions for new developers. Additional key notes:
- documents how to deploy the "app.jar" version of the image (web/Dockerfile as opposed to web-and-data/Dockerfile).
- tells WSL users which docker software to install

# Notify reviewers
@alisman

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ cBioPortal consists of several components, please read the [Architecture docs](h
 
 ### Local Development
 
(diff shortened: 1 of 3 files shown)


inodb (approved):
Awesome! This is great! Thanks @pappde !

PR #10961, opened 2024-08-28
Demo rfc80 poc mutation&genomic data count unit tests

fuzhaoyuan:
Fix rfc80-team#38 and rfc80-team#43

- Use Charles's implementation for mutated count and profiled count

### DIFF
--- a/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
+++ b/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
@@ -133,15 +133,19 @@ public class StudyViewColumnarServiceImpl implements StudyViewColumnarService {
     @Override
     public List<GenomicDataCountItem> getMutationCountsByGeneSpecific(StudyViewFilter studyViewFilter, List<GenomicDataFilter> genomicDataFilters) {
         List<GenomicDataCountItem> genomicDataCountItemList = new ArrayList<>();
+        int totalCount = studyViewRepository.getFilteredSamplesCount(createContext(studyViewFilter));
+        List<AlterationCountByGene> alterationCountByGenes = alterationCountService.getMutatedGenes(createContext(studyViewFilter));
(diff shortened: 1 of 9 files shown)


src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -133,15 +133,19 @@ public List<GenomicDataCountItem> getCNACountsByGeneSpecific(StudyViewFilter stu
     @Override
     public List<GenomicDataCountItem> getMutationCountsByGeneSpecific(StudyViewFilter studyViewFilter, List<GenomicDataFilter> genomicDataFilters) {
         List<GenomicDataCountItem> genomicDataCountItemList = new ArrayList<>();
+        int totalCount = studyViewRepository.getFilteredSamplesCount(createContext(studyViewFilter));
+        List<AlterationCountByGene> alterationCountByGenes = alterationCountService.getMutatedGenes(createContext(studyViewFilter));
```
alisman, line 137: @fuzhaoyuan (@haynescd) if I understand correctly this PR switches from using it's own query (via the old getMututionCounts method of the repository) to just leveraging the getMutatedGenes method that was built for the mutated_genes endpoint.   This is slightly painful because it means we're calling this thing at least twice for every filtering operation in the study view.  I wonder if we could adapt getMutatedGenes so that it could take an optional gene parameter and then query only that, which would have massive savings.

alisman: Unfortunately the same could be said for getFilteredSamplesCount.  The study view is already calling that once.  @haynescd this is probably too ambitious for this phase, but the solution for this redundancy would be a global cache of promises where the keys are hashes of the studyview filters.  That way, we can fire off all these requests (endpoints) at the same time and whoever is first kicks off a given filtering operation and subsequent consumers just subscribe to that cached promise.

PR #10975, opened 2024-09-04
Demo rfc80 poc mutation data count specific gene parameter

fuzhaoyuan:
Fix cBioPortal/rfc80-team#38 and cBioPortal/rfc80-team#43

- Add basic unit tests
- Overload mutatedGenes and populateAlterationCounts related methods to provide a way to query limited list of genes, thus fasten the speed of endpoint 'mutation-data-count'

### DIFF
--- a/src/main/java/org/cbioportal/persistence/StudyViewRepository.java
+++ b/src/main/java/org/cbioportal/persistence/StudyViewRepository.java
@@ -27,7 +27,9 @@ public interface StudyViewRepository {
     
     List<ClinicalData> getPatientClinicalData(StudyViewFilterContext studyViewFilterContext, List<String> attributeIds);
     
-    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext, String specificHugoGeneSymbol);
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext);
+    
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext, List<String> hugoGeneSymbols);
     
     List<AlterationCountByGene> getStructuralVariantGenes(StudyViewFilterContext studyViewFilterContext);
     List<CopyNumberCountByGene> getCnaGenes(StudyViewFilterContext studyViewFilterContext);
@@ -42,11 +44,15 @@ public interface StudyViewRepository {
 
     List<CaseListDataCount> getCaseListDataCountsPerStudy(StudyViewFilterContext studyViewFilterContext);
 
-    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType, String specificHugoGeneSymbol);
+    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType);
+
+    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType, List<String> hugoGeneSymbols);
     
     int getFilteredSamplesCount(StudyViewFilterContext studyViewFilterContext);
     
-    Map<String, Set<String>> getMatchingGenePanelIds(StudyViewFilterContext studyViewFilterContext, String alterationType, String specificHugoGeneSymbol);
+    Map<String, Set<String>> getMatchingGenePanelIds(StudyViewFilterContext studyViewFilterContext, String alterationType);
+    
+    Map<String, Set<String>> getMatchingGenePanelIds(StudyViewFilterContext studyViewFilterContext, String alterationType, List<String> hugoGeneSymbols);
     
     int getTotalProfiledCountsByAlterationType(StudyViewFilterContext studyViewFilterContext, String alterationType);
     
@@ -65,6 +71,4 @@ public interface StudyViewRepository {
     List<GenomicDataCountItem> getCNACounts(StudyViewFilterContext studyViewFilterContext, List<GenomicDataFilter> genomicDataFilters);
     
     List<GenomicDataCountItem> getMutationCountsByType(StudyViewFilterContext studyViewFilterContext, List<GenomicDataFilter> genomicDataFilters);
-
-    AlterationCountByGene getMutatedGene(StudyViewFilterContext studyViewFilterContext, String specificHugoGeneSymbol);
 }
--- a/src/main/java/org/cbioportal/persistence/StudyViewRepository.java
+++ b/src/main/java/org/cbioportal/persistence/StudyViewRepository.java
@@ -27,7 +27,7 @@ public interface StudyViewRepository {
     
     List<ClinicalData> getPatientClinicalData(StudyViewFilterContext studyViewFilterContext, List<String> attributeIds);
     
-    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext);
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext, String specificHugoGeneSymbol);
     
     List<AlterationCountByGene> getStructuralVariantGenes(StudyViewFilterContext studyViewFilterContext);
     List<CopyNumberCountByGene> getCnaGenes(StudyViewFilterContext studyViewFilterContext);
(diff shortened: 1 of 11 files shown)


src/main/java/org/cbioportal/persistence/StudyViewRepository.java
```diff
@@ -27,7 +27,7 @@ public interface StudyViewRepository {
     
     List<ClinicalData> getPatientClinicalData(StudyViewFilterContext studyViewFilterContext, List<String> attributeIds);
     
-    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext);
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext, String specificHugoGeneSymbol);
```
alisman, line 30: @haynescd @fuzhaoyuan is there a way to indicate that this is optional rather than forcing people to pass null?  I think that's called overloading at method, right?

alisman: Also, I think maybe just "hugoGeneSymbol" is enough.  specificHugGeneSymbol seems redundant.   There should be a comment explaining that we pass in a genesymbol if we only want counts for that gene.   i also wonder whether we should just make it an array since it seems inevitable that we'll want to target multiple genes.

haynescd: I would honestly create a new method called getMutatedGene that would reuse the sql in the mapper file for getMutatedGenes...

Kinda doesn't make sense to have getMutatedGenes that returns one gene

src/main/java/org/cbioportal/persistence/StudyViewRepository.java
```diff
@@ -42,11 +42,11 @@ public interface StudyViewRepository {
 
     List<CaseListDataCount> getCaseListDataCountsPerStudy(StudyViewFilterContext studyViewFilterContext);
 
-    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType);
+    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType, String specificHugoGeneSymbol);
```
haynescd, line 45: Same as above. We are overloading something that returns multiples to return a single

src/main/java/org/cbioportal/persistence/StudyViewRepository.java
fuzhaoyuan: Yeah I'll look into it

haynescd: maybe two methods (they could be overloaded )
getMutatedGenes
getMutatedGenes(StudyViewFilter, String... geneToQuery)

src/main/java/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.java
```diff
@@ -19,15 +19,14 @@
...
-import java.util.Map;
 
 
 public interface StudyViewMapper {
     List<Sample> getFilteredSamples(@Param("studyViewFilterHelper") StudyViewFilterHelper studyViewFilterHelper);
 
     List<GenomicDataCount> getMolecularProfileSampleCounts(@Param("studyViewFilterHelper") StudyViewFilterHelper studyViewFilterHelper);
     
-    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterHelper studyViewFilterHelper, AlterationFilterHelper alterationFilterHelper);
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterHelper studyViewFilterHelper, AlterationFilterHelper alterationFilterHelper, String specificHugoGeneSymbol);
```
haynescd, line 29: If you just overload the Repository layer.. you could just keep this the same except pass a list or null then you could just reuse the sql as well

haynescd (changes requested):

src/main/java/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.java
fuzhaoyuan: Think I got it. Thanks!

src/main/java/org/cbioportal/service/impl/AlterationCountServiceImpl.java
```diff
@@ -254,27 +254,27 @@ public Pair<List<CopyNumberCountByGene>, Long> getPatientCnaGeneCounts(List<Mole
...
     }
 
     private < T extends AlterationCountByGene> List<T> populateAlterationCounts(@NonNull List<T> alterationCounts,
                                                                                 @NonNull StudyViewFilterContext studyViewFilterContext,
-                                                                                @NonNull AlterationType alterationType) {
+                                                                                @NonNull AlterationType alterationType, String specificHugoGeneSymbol) {
         final int profiledCountWithoutGenePanelData = studyViewRepository.getTotalProfiledCountsByAlterationType(studyViewFilterContext, alterationType.toString());
-        var profiledCountsMap = studyViewRepository.getTotalProfiledCounts(studyViewFilterContext, alterationType.toString());
-        final var matchingGenePanelIdsMap = studyViewRepository.getMatchingGenePanelIds(studyViewFilterContext, alterationType.toString());
+        var profiledCountsMap = studyViewRepository.getTotalProfiledCounts(studyViewFilterContext, alterationType.toString(), specificHugoGeneSymbol);
```
fuzhaoyuan, line 276: @haynescd Do you think I should change it to list or overload? Same applies to the one below.

src/main/java/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.java
```diff
@@ -62,7 +61,7 @@ public interface StudyViewMapper {
 
     List<GenomicDataCountItem> getCNACounts(StudyViewFilterHelper studyViewFilterHelper, List<GenomicDataFilter> genomicDataFilters);
 
-    Map<String, Integer> getMutationCounts(StudyViewFilterHelper studyViewFilterHelper, GenomicDataFilter genomicDataFilter);
-
     List<GenomicDataCountItem> getMutationCountsByType(StudyViewFilterHelper studyViewFilterHelper, List<GenomicDataFilter> genomicDataFilters);
+
+    AlterationCountByGene getMutatedGene(StudyViewFilterHelper studyViewFilterHelper, AlterationFilterHelper alterationFilterHelper, String specificHugoGeneSymbol);
```
haynescd, line 66: Am I missing this? I don't see it defined in the xml?

src/main/java/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.java
```diff
@@ -19,15 +19,14 @@
...
-import java.util.Map;
 
 
 public interface StudyViewMapper {
     List<Sample> getFilteredSamples(@Param("studyViewFilterHelper") StudyViewFilterHelper studyViewFilterHelper);
 
     List<GenomicDataCount> getMolecularProfileSampleCounts(@Param("studyViewFilterHelper") StudyViewFilterHelper studyViewFilterHelper);
     
-    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterHelper studyViewFilterHelper, AlterationFilterHelper alterationFilterHelper);
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterHelper studyViewFilterHelper, AlterationFilterHelper alterationFilterHelper, List<String> hugoGeneSymbols);
```
haynescd, line 29: maybe a better variable name is genesToQuery?

src/main/java/org/cbioportal/persistence/StudyViewRepository.java
```diff
@@ -29,6 +29,8 @@ public interface StudyViewRepository {
     
     List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext);
     
+    List<AlterationCountByGene> getMutatedGenes(StudyViewFilterContext studyViewFilterContext, List<String> hugoGeneSymbols);
```
haynescd, line 32:
```suggestion
    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType, String... genesToQuery);
```

This is what I want. So when calling the function you can do the following...

```java
        studyViewRepository.getTotalProfiledCounts(studyViewFilter, alterationType, "TP53", "GENEA");
        //or
studyViewRepository.getTotalProfiledCountsByAlterationType(studyViewFilterContext, alterationType.toString());
```

src/main/java/org/cbioportal/persistence/StudyViewRepository.java
```diff
@@ -43,11 +45,15 @@ public interface StudyViewRepository {
     List<CaseListDataCount> getCaseListDataCountsPerStudy(StudyViewFilterContext studyViewFilterContext);
 
     Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType);
+
+    Map<String, Integer> getTotalProfiledCounts(StudyViewFilterContext studyViewFilterContext, String alterationType, List<String> hugoGeneSymbols);
```
haynescd, line 49: Change the variable name to genesToQuery

haynescd (changes requested):
Changed to use vararg to make it easier

src/main/java/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.java
fuzhaoyuan: Oh sorry I must forgot to remove it

src/main/java/org/cbioportal/persistence/StudyViewRepository.java
fuzhaoyuan: Oh wow that's cool

PR #10972, opened 2024-09-03
Fix clinical data NA counts for multiple studies

onursumer:
Fix cBioPortal/rfc80-team#39

PR #10978, opened 2024-09-05
Fix clinical data NA counts for multiple studies

onursumer:
Fix cBioPortal/rfc80-team#39

This is an alternative solution to #10972

### DIFF
--- a/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
+++ b/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
@@ -107,7 +107,17 @@ public class StudyViewColumnarServiceImpl implements StudyViewColumnarService {
     
     @Override
     public List<ClinicalDataCountItem> getClinicalDataCounts(StudyViewFilter studyViewFilter, List<String> filteredAttributes) {
-        return generateDataCountItemsFromDataCounts(studyViewRepository.getClinicalDataCounts(createContext(studyViewFilter), filteredAttributes));
+        StudyViewFilterContext studyViewFilterContext = createContext(studyViewFilter);
+        List<ClinicalDataCount> dataCounts = studyViewRepository.getClinicalDataCounts(studyViewFilterContext, filteredAttributes);
+        List<ClinicalDataCountItem> clinicalDataCountItems = generateDataCountItemsFromDataCounts(dataCounts);
+        
+        return calculateMissingNaCountsForClinicalDataCountItems(
+            clinicalDataCountItems,
+            filteredAttributes,
+            this.getClinicalAttributeDatatypeMap(),
+            studyViewRepository.getFilteredSamplesCount(studyViewFilterContext),
+            studyViewRepository.getFilteredPatientCount(studyViewFilterContext)
+        );
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -97,14 +97,60 @@ public Map<String, ClinicalDataType> getClinicalAttributeDatatypeMap() {
     
     @Override
     public List<ClinicalDataCountItem> getClinicalDataCounts(StudyViewFilter studyViewFilter, List<String> filteredAttributes) {
-        return studyViewRepository.getClinicalDataCounts(createContext(studyViewFilter), filteredAttributes)
+        Map<String, ClinicalDataType> clinicalAttributeDatatypeMap = this.getClinicalAttributeDatatypeMap();
+        StudyViewFilterContext studyViewFilterContext = createContext(studyViewFilter);
+            
+        int filteredSamplesCount = studyViewRepository.getFilteredSamplesCount(studyViewFilterContext);
```
alisman, line 103: unless i'm mistaken, we are executing all the filtering logic three times in a row here with each of these calls to the repository layer?

alisman: @onursumer is this true?

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -97,14 +97,89 @@ public Map<String, ClinicalDataType> getClinicalAttributeDatatypeMap() {
...
             .entrySet().parallelStream().map(e -> {
                 ClinicalDataCountItem item = new ClinicalDataCountItem();
                 item.setAttributeId(e.getKey());
                 item.setCounts(e.getValue());
                 return item;
-            }).collect(Collectors.toList());
+            }).toList();
+
+        // This rest of the code is postprocessing to adjust NA counts for multiple studies
+        List<ClinicalDataCountItem> combinedClinicalDataCountItems = new ArrayList<>();
```
alisman, line 117: @onursumer lets put this in a nicely named helper function

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -97,14 +97,89 @@ public Map<String, ClinicalDataType> getClinicalAttributeDatatypeMap() {
...
             .stream().collect(Collectors.groupingBy(ClinicalDataCount::getAttributeId))
             .entrySet().parallelStream().map(e -> {
                 ClinicalDataCountItem item = new ClinicalDataCountItem();
                 item.setAttributeId(e.getKey());
                 item.setCounts(e.getValue());
                 return item;
-            }).collect(Collectors.toList());
+            }).toList();
+
+        // This rest of the code is postprocessing to adjust NA counts for multiple studies
```
haynescd, line 116: Can we break this up into pieces... this one function is just under 100 lines long

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
onursumer: I think it is true. Not sure if there is a way to avoid it.

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
```diff
@@ -157,7 +168,80 @@ private StudyViewFilterContext createContext(StudyViewFilter studyViewFilter) {
...
+                    combinedClinicalDataCountItems.add(clinicalDataCountItem);
+                }
+
+                // find "NA"
+                ClinicalDataCount naClinicalDataCount = clinicalDataCountItem
+                    .getCounts()
+                    .stream()
+                    .filter(c -> c.getValue().equals("NA"))
+                    .findFirst()
+                    .orElse(null);
```
haynescd, line 224: really could keep this as an optional and not return null.

then unpack the optional below.
opt.ifPresentOrElse

onursumer: good point!

actually found a better solution

```Java
.orElseGet(() -> {
    // this should only happen when there are multiple studies
    ClinicalDataCount count = new ClinicalDataCount();
    count.setAttributeId(attributeId);
    count.setValue("NA");
    count.setCount(0);
    return count;
});
```

src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java
alisman: lets keep keep it the way it is and revisit if performance needs dictate

PR #11007, opened 2024-09-23
Add missing study view filters for clinical data binning

onursumer:
This is to fix a legacy issue where certain study view filters could be ignored when applying filters for clinical data binning

### DIFF
--- a/src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
+++ b/src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
@@ -246,9 +246,16 @@ public class StudyViewFilterUtil {
         return (
             filter != null &&
                 filter.getClinicalDataFilters() == null &&
+                filter.getClinicalEventFilters() == null &&
+                filter.getMutationDataFilters() == null &&
+                filter.getStructuralVariantFilters() == null &&
                 filter.getGeneFilters() == null &&
                 filter.getSampleTreatmentFilters() == null &&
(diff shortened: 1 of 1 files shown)


src/main/java/org/cbioportal/web/util/StudyViewFilterUtil.java
```diff
@@ -246,9 +246,16 @@ public boolean shouldSkipFilterForClinicalDataBins(StudyViewFilter filter) {
         return (
             filter != null &&
                 filter.getClinicalDataFilters() == null &&
+                filter.getClinicalEventFilters() == null &&
```
junengineer, line 249: What do you think if we write this huge null checking like this?


```
return filter != null && 
       Stream.of(
           filter.getClinicalDataFilters(),
           filter.getClinicalEventFilters(),
           filter.getMutationDataFilters(),
           filter.getStructuralVariantFilters(),
           filter.getGeneFilters(),
           filter.getSampleTreatmentFilters(),
           filter.getPatientTreatmentFilters(),
           filter.getPatientTreatmentGroupFilters(),
           filter.getSampleTreatmentGroupFilters(),
           filter.getPatientTreatmentTargetFilters(),
           filter.getSampleTreatmentTargetFilters(),
           filter.getGenomicProfiles(),
           filter.getGenomicDataFilters(),
           filter.getGenericAssayDataFilters(),
           filter.getCaseLists(),
           filter.getCustomDataFilters()
       ).allMatch(Objects::isNull);
```

PR #11032, opened 2024-09-29
Fix intersection with parens

alisman:


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -139,16 +139,16 @@
...
-                    <choose>
-                        <!-- Currently we only handle CNA in categorical genomic data filters. Future categorical profile types will involve updates to this part-->
-                        <when test="genomicDataFilter.profileType == 'cna' or genomicDataFilter.profileType == 'gistic'">
-                            <include refid="categoricalGenomicDataFilterForCNA"/>
-                        </when>
-                        <otherwise>
-                            <!-- when you make selection in genomic binning chart, it will still send genomic data filter -->
-                            <include refid="numericalGenomicDataFilter"/>
-                        </otherwise>
+                    (
```
alisman, line 142: @fuzhaoyuan i think the problem was that we introduced a UNION ALL inside in order to handle the nulls and that needs to be scoped so as to avoid interfering with the INTERSECTION above it.

alisman: probably needs to be applied also to generic assay

Issue #11046, opened 2024-10-04
MolecularProfileSampleCount discrepancy on MSKCC portal


alisman:
http://localhost:8082/study/summary?id=lusc_cptac_2021%2Cluad_cptac_2020

produces slight discrepancy in counts for molecular profile sample count 

![image](https://github.com/user-attachments/assets/8049af3f-d76b-4541-94b5-2152ea6151d8)

curl 'http://localhost:8082/api/column-store/molecular-profile-sample-counts/fetch' \
  -H 'Accept: application/json' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Cookie: _ga=GA1.1.1887007066.1710956751; _ga_5260NDGD6Z=GS1.1.1721930557.11.1.1721930574.0.0.0' \
  -H 'Origin: http://localhost:8082' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://localhost:8082/study/summary?id=lusc_cptac_2021%2Cluad_cptac_2020' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: "Google Chrome";v="129", "Not=A?Brand";v="8", "Chromium";v="129"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  --data-raw '{"studyIds":["lusc_cptac_2021","luad_cptac_2020"],"alterationFilter":{"copyNumberAlterationEventTypes":{"AMP":true,"HOMDEL":true},"mutationEventTypes":{"any":true},"structuralVariants":null,"includeDriver":true,"includeVUS":true,"includeUnknownOncogenicity":true,"includeUnknownTier":true,"includeGermline":true,"includeSomatic":true,"includeUnknownStatus":true,"tiersBooleanMap":{}}}'

PR #11049, opened 2024-10-07
Fix molecular-profile-sample-counts

### LINKED ISSUE #11016: Clickhouse mutation profile has 5 fewer samples
https://rfc80-msk.cbioportal.aws.mskcc.org/study/summary?id=gbc_msk_chile_2021%2Cblca_mskcc_solit_2014%2Cblca_tcga

Mutations profile has 5 fewer samples than it used to.

fuzhaoyuan:
Fix #11016 and #11046

- Used another way to count profiled mutations inside `molecular-profile-sample-counts` endpoint SQL
- Changed LEFT JOINs to JOINs. Not sure if this will break anything.


fuzhaoyuan:
After further investigation it is confirmed that it's in the legacy code where it's using the total sample count for the mutation count, which it should be profiled count instead. Therefore no fix is needed from this PR. Follow up issue is at #11055

PR #11050, opened 2024-10-07
Rfc80/add na clinical data filtering

haynescd:


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -368,22 +408,76 @@
...
+            ON
+            <choose>
+                <when test="'${type}' == 'sample'">
+                    sd.sample_unique_id = categorical_clinical_data.sample_unique_id
+                </when>
+                <otherwise>
+                    sd.patient_unique_id = categorical_clinical_data.patient_unique_id
+                </otherwise>
+            </choose>
+            WHERE empty(attribute_value)
```
fuzhaoyuan, line 455: So for here you can bring in the WHERE conditions for non-NA samples like
```
WHERE
  for each dataFilterValue:
    if dataFilterValue is 'NA', empty(attribute_value)
    else, normalizedAttributeValue ILIKE datavalue.value
```
Then you can save a UNION

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -368,22 +408,76 @@
...
+                    AND (
                     <include refid="normalizeAttributeValue">
                         <property name="attribute_value" value="attribute_value"/>
                     </include>
-                ) ILIKE '${dataFilterValue.value}'
-            </trim>
-        </foreach>
+                    ) ILIKE '${dataFilterValue.value}'
+                </trim>
+            </foreach>
```
fuzhaoyuan, line 478: From above mentioned comment you can bring the 470-478 up into the first NA part

PR #11061, opened 2024-10-10
Pre-select samples to speed up JOINs for NA filtering

fuzhaoyuan:
- Pre-select samples to speed up LEFT JOINs used during NA filtering


fuzhaoyuan:
Unnecessary at this point. Might continue if run into bottleneck in the future.

PR #11091, opened 2024-10-18
Re-apply caching to all study view service methods

alisman:
We want to renable caching for the new clickhouse endpoints.  This will make for very zippy initial load.   Ultimately we want to only cache UNFILTERED requests so as not to put EVERYthing in cache.  We need to work on the condition to detect that.

### DIFF
--- a/src/main/java/org/cbioportal/service/impl/ClinicalDataDensityPlotServiceImpl.java
+++ b/src/main/java/org/cbioportal/service/impl/ClinicalDataDensityPlotServiceImpl.java
@@ -9,6 +9,7 @@ import org.cbioportal.model.DensityPlotData;
 import org.cbioportal.service.ClinicalDataDensityPlotService;
 import org.cbioportal.web.columnar.StudyViewColumnStoreController;
 import org.cbioportal.web.util.DensityPlotParameters;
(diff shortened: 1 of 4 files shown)


src/main/java/org/cbioportal/service/impl/ClinicalDataDensityPlotServiceImpl.java
```diff
@@ -18,8 +19,14 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
+
 @Service
 public class ClinicalDataDensityPlotServiceImpl implements ClinicalDataDensityPlotService {
+
+    @Cacheable(
```
alisman, line 26: @haynescd ultimately i want to make the condition "unfiltered" but if you look up the way singleStudyUnfiltered is implemented, it has to be calculated on every method call. Would like to see if there's a more general way.  But for now, we can just put everything in cache.  It's ok.  It won't overwhelm the cache on genie as there's not that much traffic.  Can explain more later.
```
    @Cacheable(
               condition = "@cacheEnabledConfig.getEnabled() && #singleStudyUnfiltered" 
    )
```

haynescd (approved):

PR #11122, opened 2024-10-24
Fix a filtering issue with the violin plot patient data conversion

onursumer:
Use samples filtered without numerical clinical data filter when converting patient clinical data to sample clinical data

### DIFF
--- a/src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
+++ b/src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
@@ -362,11 +362,19 @@ public class StudyViewColumnStoreController {
         List<ClinicalData> patientClinicalDataList = filterNonEmptyClinicalData(
             studyViewColumnarService.getPatientClinicalData(interceptedStudyViewFilter, attributeIds)
         );
-
-        List<ClinicalData> combinedClinicalDataList = Stream.concat(
-            sampleClinicalDataList.stream(),
-            convertPatientClinicalDataToSampleClinicalData(patientClinicalDataList, filteredSamples).stream()
-        ).toList();
+        
+        List<ClinicalData> combinedClinicalDataList;
+        if (patientClinicalDataList.isEmpty()) {
+            combinedClinicalDataList = sampleClinicalDataList;
+        } else {
+            // fetch samples again without the numerical clinical data filter
+            List<Sample> samplesWithoutNumericalFilter = studyViewColumnarService.getFilteredSamples(interceptedStudyViewFilter); 
+            
+            combinedClinicalDataList = Stream.concat(
+                sampleClinicalDataList.stream(),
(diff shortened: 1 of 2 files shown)


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -91,8 +91,8 @@
 
     <select id="getSampleClinicalDataFromStudyViewFilter" resultType="org.cbioportal.model.ClinicalData">
         SELECT
-            sample_unique_id as sampleId,
-            patient_unique_id as patientId,
+            replaceOne(sample_unique_id, concat(cancer_study_identifier, '_'), '') as sampleId,
+            replaceOne(patient_unique_id, concat(cancer_study_identifier, '_'), '') as patientId,
```
onursumer, line 95: This should probably be refactored at some point as stated here
https://github.com/cBioPortal/cbioportal/blob/9a5f9e3b7d905890fdd5796118ba6a1257eb4cd3/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml#L144-L147

src/main/java/org/cbioportal/service/impl/ClinicalDataDensityPlotServiceImpl.java
```diff
@@ -33,7 +33,7 @@ public DensityPlotData getDensityPlotData(List<ClinicalData> sampleClinicalData,
         result.setBins(new ArrayList<>());
 
         Map<String, List<ClinicalData>> clinicalDataGroupedBySampleId = sampleClinicalData.stream().
-            collect(Collectors.groupingBy(ClinicalData::getSampleId));
+            collect(Collectors.groupingBy(c -> c.getStudyId() + "_" + c.getSampleId()));
```
onursumer, line 36: this may not be necessary, but since we removed the study id from the sample id for clinical data this change will keep the ids same for density plot implementation.

alisman (approved):

src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
```diff
@@ -362,11 +362,19 @@ public ResponseEntity<ClinicalViolinPlotData> fetchClinicalDataViolinPlots(
...
-        List<ClinicalData> combinedClinicalDataList = Stream.concat(
-            sampleClinicalDataList.stream(),
-            convertPatientClinicalDataToSampleClinicalData(patientClinicalDataList, filteredSamples).stream()
-        ).toList();
+        
+        List<ClinicalData> combinedClinicalDataList;
+        if (patientClinicalDataList.isEmpty()) {
+            combinedClinicalDataList = sampleClinicalDataList;
+        } else {
+            // fetch samples again without the numerical clinical data filter
```
alisman, line 370: @onursumer can you add to comment WHY we fetch again without clinical data filter?

onursumer: done!

PR #11132, opened 2024-10-28
Fetch missing patient clinical data for the density plot

### LINKED ISSUE #11118: specific x vs y chart isn't loading in clickhouse
This loads in non-clickhouse genie:
image

But not in clickhouse:
image

I tried a different numeric-vs-numeric chart and it loads just fine in clickhouse, so not sure why this one is special

onursumer:
Fix #11118

Moved some of the shared logic into `ClinicalDataViolinPlotUtil` and renamed the class as `ClinicalDataXyPlotUtil`, because both violin and density plots are X vs Y plots.

### DIFF
--- a/src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
+++ b/src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
@@ -68,15 +68,12 @@ import org.springframework.web.bind.annotation.RestController;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
-import static org.cbioportal.web.columnar.util.ClinicalDataViolinPlotUtil.convertPatientClinicalDataToSampleClinicalData;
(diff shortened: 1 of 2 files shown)


src/main/java/org/cbioportal/web/columnar/StudyViewColumnStoreController.java
```diff
@@ -308,8 +304,18 @@ public ResponseEntity<DensityPlotData> fetchClinicalDataDensityPlot(
             .yAxisLogScale(yAxisLogScale)
             .build();
         
-        List<ClinicalData> sampleClinicalDataList = studyViewColumnarService.getSampleClinicalData(interceptedStudyViewFilter, xyAttributeId);
-        DensityPlotData result = clinicalDataDensityPlotService.getDensityPlotData(sampleClinicalDataList, densityPlotParameters, interceptedStudyViewFilter);
+        List<ClinicalData> combinedClinicalDataList = fetchClinicalDataForXyPlot(
+            studyViewColumnarService,
+            interceptedStudyViewFilter,
+            List.of(xAxisAttributeId, yAxisAttributeId),
+            false
```
onursumer, line 311: Not sure why we filter empty clinical data for violin plot but not for density plot. Maybe we should filter empty data here as well and get rid of this additional parameter.

alisman (approved):

PR #11153, opened 2024-11-01
:lock: Update Default Filter Chain Ordering for Spring Security

haynescd:
Fix https://github.com/cBioPortal/cbioportal/issues/11156. Accessing via API token wasn't working because of a race condition. This forces an order. See related issue: https://github.com/spring-projects/spring-security/issues/15220


dippindots (approved):

inodb (approved):
Awesome 🏆 !!

PR #11185, opened 2024-11-13
Add back NA counts when corresponding samples are filtered out of study view 

alisman:


haynescd:
@alisman can you address the sonar fixes

PR #11216, opened 2024-11-20
Check for non-numerical values in numerical data count filter

### LINKED ISSUE #11204: Combination of categorical filter and filering on "unknown" in numerical produces discrepancy
curl 'http://localhost:8082/api/column-store/filtered-samples/fetch' \
  -H 'Accept: application/json' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Cookie: _ga=GA1.1.1887007066.1710956751; _ga_5260NDGD6Z=GS1.1.1721930557.11.1.1721930574.0.0.0' \
  -H 'Origin: http://localhost:8082' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://localhost:8082/study/summary?id=genie_public' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: "Chromium";v="130", "Google Chrome";v="130", "Not?A_Brand";v="99"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  --data-raw '{"clinicalDataFilters":[{"attributeId":"CANCER_TYPE_DETAILED","values":[{"value":"BREAST INVASIVE DUCTAL CARCINOMA"}]},{"attributeId":"AGE_AT_SEQ_REPORT","values":[{"start":65,"end":70},{"start":70,"end":75},{"start":75,"end":80}]}],"studyIds":["genie_public"],"alterationFilter":{"copyNumberAlterationEventTypes":{"AMP":true,"HOMDEL":true},"mutationEventTypes":{"any":true},"structuralVariants":null,"includeDriver":true,"includeVUS":true,"includeUnknownOncogenicity":true,"includeUnknownTier":true,"includeGermline":true,"includeSomatic":true,"includeUnknownStatus":true,"tiersBooleanMap":{}}}'

onursumer:
Fix #11204

~~This only fixes numericalClinicalDataCountFilter.~~

Now applied the same fix to `numericalGenomicDataFilter` and `numericalGenericAssayDataFilter` as well by unifying the filter logic under one SQL.

TODO: add unit tests with variety of clinical, genomic, and generic assay data filter combinations

### DIFF
--- a/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
+++ b/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
@@ -288,11 +288,8 @@
                     such as "UNKNOWN", "NOT COLLECTED", "NOT RELEASED", etc.
                 -->
                 <if test="dataFilterValue.value != null">
-                    AND (
-                    <include refid="normalizeAttributeValue">
-                        <property name="attribute_value" value="${attribute_value}"/>
-                    </include>
-                    ) ILIKE #{dataFilterValue.value}
+                    AND 
+                    ${attribute_value} ILIKE #{dataFilterValue.value}
                 </if>
                 <!--
                     Special case: Numerical range values such as <18, >=90, etc.
@@ -481,11 +478,7 @@
                         </include> = 'NA'
                     </when>
                     <otherwise>
-                        (
-                        <include refid="normalizeAttributeValue">
-                            <property name="attribute_value" value="attribute_value"/>
-                        </include>
-                        ) ILIKE #{dataFilterValue.value}
+                        attribute_value ILIKE #{dataFilterValue.value}
                     </otherwise>
                 </choose> 
             </foreach> 
@@ -612,11 +605,7 @@
                         </include> = 'NA'
                     </when>
                     <otherwise>
-                        (
-                        <include refid="normalizeAttributeValue">
-                            <property name="attribute_value" value="value"/>
-                        </include>
-                        ) ILIKE #{dataFilterValue.value}
+                        value ILIKE #{dataFilterValue.value}
                     </otherwise>
                 </choose>
             </foreach>
--- a/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
+++ b/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
@@ -374,6 +374,13 @@
                         <property name="attribute_value" value="attribute_value"/>
                     </include>
                 </if>
+                <if test="dataFilterValue.value != null">
+                    AND (
+                    <include refid="normalizeAttributeValue">
+                        <property name="attribute_value" value="attribute_value"/>
+                    </include>
+                    ) ILIKE #{dataFilterValue.value}
+                </if>
                 <if test="dataFilterValue.start != null and dataFilterValue.end == null">
                     AND match(attribute_value, '^>?=?[-+]?[0-9]*[.,]?[0-9]+$')
                 </if>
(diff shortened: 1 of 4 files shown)


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -272,6 +272,77 @@
...
+    <sql id="applyNumericalDataFilter">
+        <foreach item="dataFilterValue" collection="${data_filter_collection}" open="((" separator=") OR (" close="))">
+            <trim prefix="" prefixOverrides="AND">
+                <!-- Special case: NA -->
+                <if test="dataFilterValue.value eq 'NA'">
+                    AND
+                    <include refid="isAttributeValueNA">
+                        <property name="attribute_value" value="${attribute_value}"/>
+                    </include>
+                </if>
```
fuzhaoyuan, line 284: Maybe we don't need to handle NA here if this SQL is used as the non-NA part? Just curious, it's also fine to have it here. And I think it's also possible to include the NA part as a whole SQL but this is great progress already. Appreciate it.

fuzhaoyuan (dismissed):
lgtm

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
onursumer: I think this NA part was implemented before we added the additional logic for NA.

You're right. We should probably move the NA filter logic into this same SQL as well. Might be better to deal with it in a separate PR tough.

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -272,6 +272,77 @@
...
+                <!--
+                    Special case: Non-numerical value within numerical filter collection.
+                    Even if the data type is numerical we can still have non-numerical values
+                    such as "UNKNOWN", "NOT COLLECTED", "NOT RELEASED", etc.
+                -->
+                <if test="dataFilterValue.value != null">
+                    AND (
+                    <include refid="normalizeAttributeValue">
```
haynescd, line 292: I wouldn't do the normalizeAttributeValue. You are just testing multiple values of NA?

haynescd (changes requested):
Looks good. Just had one question

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
onursumer: `normalizeAttributeValue` used to normalize values other than `NA` but we got rid of that part at some point by upper casing everything. So I guess normalizing here is redundant now.

We are still doing this in multiple places where we filter categorical data. For example, see https://github.com/cBioPortal/cbioportal/blob/d9459f39a92ab251178793349e1b9ba003148238/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml#L483-L489
and
https://github.com/cBioPortal/cbioportal/blob/d9459f39a92ab251178793349e1b9ba003148238/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml#L614-L620

These are also most likely redundant now, and I think we should address this duplication as well, but probably in a separate PR.

onursumer: Removed `normalizeAttributeValue` from all these 3 places. We may still want to address the code duplication for categorical filters at some point.

Issue #11195, opened 2024-11-15
Data sanitization for giant IN list when handling sampleIds


alisman:
In https://github.com/cBioPortal/cbioportal/pull/11191, we need to sanitize the values to prevent SQL injection.  We should also investigate whether we can "use temporary table (or a Table Valued Parameter or JSON if supported)."

PR #11241, opened 2024-11-29
Custom sample filtering performance and security improvement

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/11195

### DIFF
--- a/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
+++ b/src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
@@ -59,14 +59,14 @@
 
             </if>
 
-            <if test="studyViewFilterHelper.studyViewFilter.sampleIdentifiers != null and !studyViewFilterHelper.studyViewFilter.sampleIdentifiers.isEmpty()">
-                INTERSECT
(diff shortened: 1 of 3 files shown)


src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -96,7 +96,9 @@
                                     OR
                                     sample_unique_id NOT IN (
                                         <foreach item="sampleIdentifier" collection="studyViewFilterHelper.customDataSamples" separator=",">
-                                            '${sampleIdentifier.studyId}_${sampleIdentifier.sampleId}'
+                                            <if test="sampleIdentifier.getUniqueSampleId != null">
+                                                #{sampleIdentifier.getUniqueSampleId}
+                                            </if>
```
dippindots, line 101: I'm not sure if adding `test` at here will break anything. maybe we need to add another `'',` at here like we did in line 85. Maybe @gblaih can comment on this.

gblaih: there is a case if getUniqueSampleId is null for all customDataSamples, it will break since we cannot do NOT IN (). adding '' will prevent it from breaking, but it may cause unintended issues with the logic i think. is there a case where the custom samples have null studyIds and sampleIds?

dippindots:
closing this and opened a new pr

PR #11248, opened 2024-12-03
Optimize bins filtering by merging bins

fuzhaoyuan:
- Merge numerical bins to reduce the number of bins passed to the database filtering

### DIFF
--- a/src/main/java/org/cbioportal/service/util/StudyViewColumnarServiceUtil.java
+++ b/src/main/java/org/cbioportal/service/util/StudyViewColumnarServiceUtil.java
@@ -5,21 +5,14 @@ import org.cbioportal.model.ClinicalAttribute;
 import org.cbioportal.model.ClinicalDataCount;
 import org.cbioportal.model.ClinicalDataCountItem;
 import org.cbioportal.model.GenomicDataCount;
-import org.cbioportal.web.parameter.ClinicalDataFilter;
-import org.cbioportal.web.parameter.DataFilter;
-import org.cbioportal.web.parameter.DataFilterValue;
-import org.cbioportal.web.parameter.GenericAssayDataFilter;
-import org.cbioportal.web.parameter.GenomicDataFilter;
-import org.cbioportal.web.parameter.StudyViewFilter;
 
-import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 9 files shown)


haynescd:
@fuzhaoyuan Looks good. Just fix the one sonar issue and we are good

haynescd (approved):

src/main/java/org/cbioportal/service/util/StudyViewColumnarServiceUtil.java
```diff
@@ -155,5 +162,61 @@ public static List<ClinicalDataCount> normalizeDataCounts(List<ClinicalDataCount
...
+                // merge adjacent numerical bins
+                else {
+                    BigDecimal start = dataFilterValue.getStart();
+                    BigDecimal end = dataFilterValue.getEnd();
+
+                    if (mergedStart == null && mergedEnd == null) {
+                        mergedStart = start;
+                        mergedEnd = end;
+                    }
+                    else if (mergedEnd.equals(start)) mergedEnd = end;
```
onursumer, line 207:
```suggestion
                    else if (mergedEnd.equals(start)) {
                        mergedEnd = end;
                    }
```

src/main/java/org/cbioportal/service/util/StudyViewColumnarServiceUtil.java
```diff
@@ -155,5 +162,61 @@ public static List<ClinicalDataCount> normalizeDataCounts(List<ClinicalDataCount
...
+     * Merge the range of numerical bins in DataFilters to reduce the number of scans that runs on the database when filtering.
+     */
+    private static <T extends DataFilter> List<T> mergeDataFilters(List<T> filters) {
+        List<T> mergedDataFilters = new ArrayList<>();
+
+        for (T filter : filters) {
+            List<DataFilterValue> mergedValues = new ArrayList<>();
+
+            BigDecimal mergedStart = null, mergedEnd = null;
+            for (DataFilterValue dataFilterValue : filter.getValues()) {
```
onursumer, line 193: this logic only works if the bins are sorted properly, right? are we sure that bins are always sorted the way we want?

fuzhaoyuan: I had 2 assumptions: 1) they come in sorted and 2) they're continuous. I handled the discontinuous case by adjacent merging but if sorting is needed we'll need an extra part and way more combinations of possible cases. What would you suggest?

PR #11256, opened 2024-12-04
Update cBioPortal to dynamically load ch Components only when property clickhouse_enabled is set

haynescd:


src/main/resources/application.properties.EXAMPLE
```diff
@@ -4,15 +4,24 @@ app.name=cbioportal
...
 # database
-spring.datasource.mysql.url=jdbc:mysql://localhost:3306/cbioportal?useSSL=false
-spring.datasource.mysql.username=cbio
-spring.datasource.mysql.password=P@ssword1
-spring.datasource.mysql.driver-class-name=com.mysql.jdbc.Driver
-spring.datasource.clickhouse.url=jdbc:ch://localhost:8443/cbioportal
-spring.datasource.clickhouse.username=dummy
-spring.datasource.clickhouse.password=dummy
-spring.datasource.clickhouse.driver-class-name=com.clickhouse.jdbc.ClickHouseDriver
+
```
alisman, line 12: @haynescd @inodb maybe we should hide this configuration stuff for the time being?

PR #11275, opened 2024-12-12
Bin merge test & input validation

fuzhaoyuan:
- Add more tests and provide input validation for future referrence

### DIFF
--- a/src/main/java/org/cbioportal/persistence/helper/StudyViewFilterHelper.java
+++ b/src/main/java/org/cbioportal/persistence/helper/StudyViewFilterHelper.java
@@ -135,7 +135,12 @@ public final class StudyViewFilterHelper {
      * Merge the range of numerical bins in DataFilters to reduce the number of scans that runs on the database when filtering.
      */
     public static <T extends DataFilter> List<T> mergeDataFilters(List<T> filters) {
-        boolean isNonNumericalOnly = true;
+        // this should throw error or move to all binning endpoints in the future for input validation
+        if (!areValidFilters(filters)) {
+            return filters;
+        }
+        
+        boolean hasNumericalValue = false;
         List<T> mergedDataFilters = new ArrayList<>();
 
         for (T filter : filters) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/cbioportal/persistence/helper/StudyViewFilterHelper.java
```diff
@@ -138,29 +143,29 @@ public static <T extends DataFilter> List<T> mergeDataFilters(List<T> filters) {
                 // leave non-numerical values as they are
                 if (dataFilterValue.getValue() != null) {
                     nonNumericalValues.add(dataFilterValue);
+                    continue;
```
alisman, line 146: @fuzhaoyuan can we add a comment explaining why non-null means non numerical?  i know that is true i just think someone coming to this for the first time would be baffled by it.

alisman: also, a comment explaining why we're going to jump to next iteration (continue)

alisman: i actually think that making an else clause would be clearer/preferable to continue.

alisman: in general i think this logic here is very difficult to grok.   because the number of items (bins) is always going to be negligible, i would prioritize readability and adopt a functional approach where we just filter out the non numericals and numericals into a separate array, sort the numericals and then use reduce to iteratively merge them depending on adjacency.

fuzhaoyuan: I used continue not else because otherwise SonarQube would complain the cognitive complexity.
I have added comments throughout the process. Basically yes we filter out non-numericals and then for numericals we merge them depending on adjacency.
The sorting would be a hassle because of the structure of class DataFilterValue, the actual bin range is hidden deep inside, so I just prevented unsorted filters in the input validation.

haynescd (approved):

PR #11281, opened 2024-12-13
Fx casing

alisman:


src/main/java/org/cbioportal/service/util/StudyViewColumnarServiceUtil.java
```diff
@@ -27,11 +27,14 @@ public static List<ClinicalDataCountItem> mergeClinicalDataCounts(
...
-                .collect(Collectors.groupingBy(ClinicalDataCount::getValue));
+                .collect(Collectors.groupingBy(a -> a.getValue().toLowerCase()));
             List<ClinicalDataCount> res = countsPerType.entrySet().stream().map(entry -> {
                 ClinicalDataCount mergedCount = new ClinicalDataCount();
                 mergedCount.setAttributeId(attr.getAttributeId());
-                mergedCount.setValue(entry.getKey());
+                // we are just going to choose the value of the first item
+                // due to failure in data normalization in source files, we may find values
+                // have inconsistent casing.  we choose to merge and choose an arbitrary casing
+                mergedCount.setValue(entry.getValue().stream().findFirst().get().getValue());
```
onursumer, line 37: We can probably avoid modifying this method by reusing `normalizeDataCounts`. So, something like this after this line might be better.

https://github.com/cBioPortal/cbioportal/blob/540b57d37cb07ee30916efc46b4c18b355d08bcf/src/main/java/org/cbioportal/service/impl/StudyViewColumnarServiceImpl.java#L166

```Java
result.forEach(item -> item.setCounts(StudyViewColumnarServiceUtil.normalizeDataCounts(item.getCounts())));
```

PR #11290, opened 2024-12-16
 genetable version added to info api

rajeswari2904:
10299-enhancement

- Genetable version added to info api response and model class,maven properties file.
- Genetable version added to info table in cgds.sql and migration.sql scripts.
- Genetable version added to test cases and test properties file.
- Verified that the info api is returning genetable version in response.
- Ran all tests to assure nothing else was accidently broken in the java part by running:mvn integration-test.


pom.xml
```diff
@@ -31,7 +31,7 @@
 
         <!-- THIS SHOULD BE KEPT IN SYNC TO VERSION IN CGDS.SQL -->
 		<db.version>2.13.1</db.version>
-
+        <genetable.version>hgnc_v2023.10.1</genetable.version>
```
inodb, line 34: @rajeswari2904 should this be pulled from the SQL instead? Do we need to hardcode it in the pom?

inodb (commented):
Thanks so much @rajeswari2904 ! Made a minor comment

pom.xml
inodb: I think the hgnc version is a bit different from `db.version`

- the `db.version` indicates the version of the SQL schema. The backend is only compatible with a specific version of this schema
- the genetable version is fine if it differs with the backend code version. No need to hardcode the version in the pom.xml. Whoever loads the gene table data should be setting this version

src/main/resources/db-scripts/cgds.sql
```diff
@@ -755,5 +756,5 @@ CREATE TABLE `resource_study` (
 );
 
 -- THIS MUST BE KEPT IN SYNC WITH db.version PROPERTY IN pom.xml
-INSERT INTO info VALUES ('2.13.1', NULL);
+INSERT INTO info VALUES ('2.13.1','hgnc_v2023.10.1', NULL );
```
inodb, line 759: This version should maybe be `N/A` or empty string. Whoever loads the gene table data should set the version

src/main/resources/db-scripts/migration.sql
```diff
@@ -1027,4 +1027,6 @@ ALTER TABLE `clinical_event_data` MODIFY COLUMN `VALUE` varchar(3000) NOT NULL;
 CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
 CREATE INDEX idx_clinical_event_value ON clinical_event_data (`VALUE`);
 CREATE INDEX idx_sample_stable_id ON sample (`STABLE_ID`);
+ALTER TABLE `info` ADD COLUMN `GENE_TABLE_VERSION` varchar(24);
+UPDATE `info` SET `GENE_TABLE_VERSION`="hgnc_v2023.10.1";
```
inodb, line 1031: same here also good if the version is empty or `N/A`

rajeswari2904:
@inodb Thanks for reviewing my PR! I don’t have Slack at the moment. Would it be possible to 
connect via another platform like Google Hangout or Zoom? 
Also, could you share your official email ID so that I can coordinate accordingly?

Looking forward to your response. Thanks!

inodb:
@rajeswari2904 yes that works - can you send me an email (ino@ino.bio)? Thanks!

dippindots:
The ticket is closed now by merging https://github.com/cBioPortal/cbioportal/pull/11757, thanks for your work on this ticket!

PR #11310, opened 2024-12-31
Added comments classes

nikhilnambiarc:
Comments help in reducing code-smells and therefore detailed comments were added for better understanding of certain classes and its methods. The classes for which comments were added are: 

1. CopyNumberSegmentService
2. DiscreteCopyNumberService
3. GenericAssayService
4. GenesetHierarchyService
5. SampleListService

These classes can be found under: ~/cbioportal/src/main/java/org/cbioportal/service


haynescd:
Can you rebase and maybe update your tabs to spaces? it seems to be off when looking at the diff

PR #11336, opened 2025-01-15
Fix genie api page error

gblaih:


dippindots (approved):
I tested this and it's working for me locally. Thanks!

PR #11365, opened 2025-01-30
FIX-11364 Proxy Controller and Monkifier

JREastonMarks:
Fix # 11364
Updates the proxy controller to be able to handle multiple proxy accounts in a secure manner. Updates the Monkifier to be a static utility class instead of a component.


JREastonMarks:
These 2 security issues and 4 new vulnerabilities are from legacy code that was renamed in this project. I am not sure of how much they are used.

The third security issue is an introduced one but sonarcube does not like all HTTP requests to go into a single method. However, this is what is required of the proxy.

src/main/java/org/cbioportal/proxy/ProxyExchangeGateway.java
```diff
@@ -0,0 +1,197 @@
...
+			service = service.substring(0, endPosition);
+		}
+
+		if (!routes.containsKey(service)) {
+			throw new UnknownServiceException();
+		}
+
+		String updatedPath = proxy.path("/proxy/" + service).trim();
+
+		String proxyServerHost = routes.get(service).trim();
```
onursumer, line 68: Might be better to extract this logic to a separate method, something like `getValidatedService`

src/main/java/org/cbioportal/proxy/ProxyExchangeGateway.java
onursumer, line 1: Seems like the indentation in this file is using tabs instead of spaces. I suspect deleting `ProxyController` and adding a brand new file caused this issue. I think just renaming the file and modifying the contents would be better. That way we can still keep the change history.

onursumer (changes requested):
Looks good in general. Just requested a few minor changes.

src/main/java/org/cbioportal/proxy/ProxyExchangeGateway.java
haynescd: Yea, might as well

src/main/java/org/cbioportal/proxy/ProxyExchangeGateway.java
```diff
@@ -0,0 +1,197 @@
...
+				break;
+			case "POST":
+				response = proxy.uri(uriComponents.toUri()).body(body).post();
+				break;
+			case "PUT":
+				response = proxy.uri(uriComponents.toUri()).body(body).put();
+				break;
+			default:
+				throw new UnknownServiceException();
+
```
haynescd, line 92: I would you HttpMethod.GET and etc instead of raw string.

Also combine GET and PATCH

```java
static boolean itsTheWeekend(DayOfWeek dow) {
    return switch(dow) {
        case SATURDAY, SUNDAY -> true;
        default               -> false;
    };
}
```

PR #11368, opened 2025-02-03
♻️ Update cBioPortal Architecture to start transition to clickhouse only implementation

### LINKED ISSUE #11360: Update cBioPortal Backend to utilize Clean Archictecture Principles
https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html

This separation will allow us to easily switch out the db in the future to clickhouse, mysql etc....

```bash
src/main/java/org/cbioportal
├── application
│   ├── documentation
│   ├── proxy
│   │   └── util
│   ├── rest
│   │   ├── mapper
│   │   ├── parameter
│   │   ├── request
│   │   ├── response
│   │   └── vcolumnstore
│   ├── security
│   │   ├── basic
│   │   ├── config
│   │   ├── token
│   │   │   ├── config
│   │   │   ├── oauth2
│   │   │   └── uuid
│   │   └── util
│   └── studyview
├── cancerstudy
│   ├── repository
│   └── usecase
├── infrastructure
│   ├── config
│   └── repository
│       └── clickhouse
│           └── cancerstudy
```

Image

haynescd:
closes #11360 

- This PR intends to utilize Strangler Fig Pattern to refactor the current code base to utilize a clean architecture approach. 

This will allow cbioportal to methodically transition to `clickhouse only` implementation

- I have implemented the endpoint `/api/columnar-store/studies` 

With this PR we have added Spring Profiles, which allows the application to toggle clickhouse functionality by using the property `spring.profiles.active=clickhouse`

I moved most of the legacy code into a sub-package `legacy`


haynescd:
Appreciate the feedback @sheridancbio!

To respond to your first concern. I agree something does need to be in place that either automatically checks or formats the code for each PR. I am leaning towards the check... I don't want the PR manipulating code. I found [this](https://github.com/spotify/fmt-maven-plugin) maven plugin than can be used to check or format. I will need some help to establish these rules and collectively discourage people from ignoring them and just using their admin abilities to bypass those... This also applies to your comment about enabling sonar cloud for static analysis!! Probably need to take this convo somewhere where we won't forget about it... And start assigning some of these tasks to people 


So, about the other concerns dealing with changing anything that can break other projects that depends on cbioportal(cbio-core and others) I totally agree. I think we should be very careful changes these things especially org.cbioportal.model package ( which I think is the only thing users are depending on)

That is why the org.cbioportal.model will probably be the last thing to go. ( Future changes should be propagated to others)  

With this change I am not to worried since it is only a package change. ( changing folders) We should still create an issue that captures this.. So we are prepared when we do need to make a database change.

Def. appreciate the comments I need as much help as I can get to push this project into using better practices.

src/main/java/org/cbioportal/cancerstudy/CancerStudyMetadata.java
```diff
@@ -0,0 +1,24 @@
+package org.cbioportal.cancerstudy;
+
+import java.util.Date;
+
```
alisman, line 4: @haynescd hey, i came across this beauty.   this is to provide default null values for some of these props?

Issue #11373, opened 2025-02-05
Create Standalone Deployment Documentation


JREastonMarks:
**Problem**
Currently no documentation exists for standing up a completely isolated cBioPortal instance. This would include setting up NIGINX, cBioPortal, Session Service, OncoKB, Genome Nexus, and supporting services (e.g., DBs). This documentation should include changes to ensure that the environment is isolated from outside requests.

**Solution**
Create a documentation that explains how to setup an isolated version of cBioPortal.

PR #11378, opened 2025-02-06
Creates Standalone Configuration Information Section in Documentation

JREastonMarks:
#11373 Creates Standalone Configuration Information Section in Documentation


docs/deployment/standalone/README.md
```diff
@@ -0,0 +1,460 @@
...
+
+
+## Setup Instructions
+
+### Docker Configuration and Setup
+Our docker configuration is based off the default cBioPortal docker configuration however we have made some changes to support our requirements. We create two networks in this deployment. The first being a bridge network that allows the services to connect to the outside world. The second is an internal network that only allows for services to talk to other services in that network. Most of the services are configured to use the internal network, with the exception being the NGINX and cBioPortal services. The NGINX needs to be able to communicate with the outside world in order to make the cBioPortal instance available to the outside world. The cBioPortal instance needs to be able to communicate with the outside world in order to allow for authentication with an outside service. However, if the cBioPortal instance does not authentication then the service can be deployed only on the internal network.
```
fuzhaoyuan, line 10: Just a small typo
```suggestion
Our docker configuration is based off the default cBioPortal docker configuration however we have made some changes to support our requirements. We create two networks in this deployment. The first being a bridge network that allows the services to connect to the outside world. The second is an internal network that only allows for services to talk to other services in that network. Most of the services are configured to use the internal network, with the exception being the NGINX and cBioPortal services. The NGINX needs to be able to communicate with the outside world in order to make the cBioPortal instance available to the outside world. The cBioPortal instance needs to be able to communicate with the outside world in order to allow for authentication with an outside service. However, if the cBioPortal instance does not authenticate then the service can be deployed only on the internal network.
```

fuzhaoyuan (approved):
LGTM. Will it show up somewhere in the [Docs](https://docs.cbioportal.org/deployment/)?

inodb (approved):
This looks great to me! Thanks @JREastonMarks !

A few small things:

- To get the doc to show up in the TOC, you'll need to add it to `SUMMARY.md`
- We recently started adding an `addon/` folder to the cbioportal-docker-compose repo (see for clickhouse: https://github.com/cbioPortal/cbioportal-docker-compose#clickhouse-mode). Idea being that you could sorta layer in additional services etc by adding additional compose files. Maybe some of the compose code in the docs could go there. Not necessary to do now tho, but maybe a good follow up. I included @zainasir to review

PR #11393, opened 2025-02-12
Clickhousify samples endpoint

### LINKED ISSUE #11377: Clickhouseify the samples endpoint
Why does the uncached samples endpoint take 30 seconds to response?

onursumer:
Fix #11377


haynescd (changes requested):
Happy to discuss. I thought the new clickhouse work will utilize the clean arch approach. I can give you more details tomorrow

haynescd (changes requested):
Looks good (needs a couple of changes) still looking at the derived sample table

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -54,30 +54,82 @@ WHERE gene.entrez_gene_id > 0 AND gene.type = 'protein-coding';
...
+    type_of_cancer_id           String,
+    cancer_study_name           String,
+    cancer_study_description    String,
+    cancer_study_public         Int,
+    cancer_study_pmid           String,
+    cancer_study_citation       String,
+    cancer_study_groups         String,
+    cancer_study_status         Int,
+    cancer_study_import_date    DateTime,
+    reference_genome            String
```
haynescd, line 80: Also, would be curious how much speed we get with the wide column vs just doing a simple join or using a dictionary. Don't want to just randomly add more columns to these tables if they aren't needed

onursumer: Just tried moving everything, including the `sequenced` and `copy_number_segment_present` logic, to the SQL layer. It doesn't seem to degrade the performance significantly. The wide table is slightly faster but probably negligible.

onursumer: Reduced the number of columns by https://github.com/cBioPortal/cbioportal/pull/11393/commits/34f3b7748139f24cdd76aaff9dd731a40fac6553.

For now keeping `sequenced` and `copy_number_segment_present` as well as `sample_type` and `patient_internal_id`, but it is also possible to derive them in SQL if we don't want to add any new columns to `sample_derived` at all.

src/main/resources/mappers/clickhouse/sample/ClickhouseSampleMapper.xml
```diff
@@ -38,4 +98,199 @@
...
+    <sql id="filterByCaseAndStudy">
+        <where>
+            <if test="sampleIds == null and studyIds != null">
+                sample_derived.cancer_study_identifier IN
+                <if test="studyIds.isEmpty()">
+                    (NULL)
+                </if>
+                <if test="!studyIds.isEmpty()">
+                    <foreach item="item" collection="studyIds" open="(" separator="," close=")">
+                        #{item}
```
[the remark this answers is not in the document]
onursumer: @haynescd now using ArrayType Handler wherever possible (see https://github.com/cBioPortal/cbioportal/pull/11393/commits/db7cc547e87898318f38cf051eec04fb73e852e5)

There is still one query where we combine two separate arrays into tuples by using a `foreach` loop. We may want to refactor this one as well so that we can use ArrayType Handler. I think we need to somehow create an array of tuples here. https://github.com/cBioPortal/cbioportal/pull/11393/files#diff-02fbcfa1feba704441a563f1a28297988c5eecdc8d8895f11e3e58b7b0ed6a2aR243-R247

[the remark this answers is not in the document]
onursumer: now concatenating study id and sample id so that we can use ArrayType handler

see https://github.com/cBioPortal/cbioportal/pull/11393/files#diff-02fbcfa1feba704441a563f1a28297988c5eecdc8d8895f11e3e58b7b0ed6a2aR244-R249

src/main/java/org/cbioportal/domain/sample/Sample.java
```diff
@@ -1,18 +1,93 @@
...
+            null,
+            cancerStudyIdentifier,
+            null,
+            null,
+            uniqueSampleKey,
+            uniquePatientKey
+        );
+    }
+
+    public Sample(
```
alisman, line 63: We should have a comment here explaining the difference between these Sample overloads

dippindots (approved):

src/main/java/org/cbioportal/infrastructure/repository/clickhouse/sample/ClickhouseSampleRepository.java
```diff
@@ -27,4 +31,120 @@ public List<Sample> getFilteredSamples(StudyViewFilterContext studyViewFilterCon
...
         return mapper.getSampleCount(studyViewFilterContext);
     }
+
+    @Override
+    public List<Sample> fetchSamples(
+        List<String> studyIds,
+        List<String> sampleIds,
+        ProjectionType projection
+    ) {
+        return switch (projection) {
```
alisman, line 41: @onursumer where do pagination params go?

onursumer: @alisman This method doesn't support pagination. Legacy implementation supports pagination only for certain methods where we get all samples. I kept it the same way for the clickhouse implementation.

https://github.com/cBioPortal/cbioportal/blob/fa1847fbada528189e92c374a63f1b24c23bef50/src/main/java/org/cbioportal/infrastructure/repository/clickhouse/sample/ClickhouseSampleRepository.java#L79-L80

https://github.com/cBioPortal/cbioportal/blob/fa1847fbada528189e92c374a63f1b24c23bef50/src/main/java/org/cbioportal/infrastructure/repository/clickhouse/sample/ClickhouseSampleRepository.java#L110-L111

https://github.com/cBioPortal/cbioportal/blob/fa1847fbada528189e92c374a63f1b24c23bef50/src/main/java/org/cbioportal/infrastructure/repository/clickhouse/sample/ClickhouseSampleRepository.java#L133-L134

src/main/resources/mappers/clickhouse/sample/ClickhouseSampleMapper.xml
```diff
@@ -38,4 +98,198 @@
...
+        cs.name as "patient.cancerStudy.name",
+        cs.description as "patient.cancerStudy.description",
+        cs.public as "patient.cancerStudy.publicStudy",
+        cs.pmid as "patient.cancerStudy.pmid",
+        cs.citation as "patient.cancerStudy.citation",
+        cs.groups as "patient.cancerStudy.groups",
+        cs.status as "patient.cancerStudy.status",
+        cs.import_date as "patient.cancerStudy.importDate",
+        rg.name as "patient.cancerStudy.referenceGenome"
+    </sql>
```
haynescd, line 191: So, I would say this is a good abstraction but the extra abstraction makes it easier to get lost.

The balance is to not duplicate code, but also don't abstract so much

src/main/resources/mappers/clickhouse/sample/ClickhouseSampleMapper.xml
```diff
@@ -38,4 +98,198 @@
...
+        SELECT
+            <include refid="sampleColumnsDetailed" />
+        <include refid="fromJoinedTable" />
+    </sql>
+
+    <sql id="selectMetaSample">
+        SELECT
+            COUNT(*) as totalCount
+        FROM sample_derived
+    </sql>
```
haynescd, line 215: I would remove these, but keep the other abstractions

haynescd (changes requested):
Small change, but after that It is good

haynescd (approved):

PR #11399, opened 2025-02-18
Use BIGINT primary key for clinical_event table

jamesqo:
Fixes issue: knowledgesystems/pipelines-scrum#1434

We've recently run into problems where the 32-bit integer primary key, used for most of the tables in the cBioPortal DB, has overflown (eg for clinical events). We decided to address this issue by using a 64-bit integer as the primary key for the `clinical_event` table. This PR updates the cgds/migration scripts and other locations so we can support this.

# Checks

# Any screenshots or GIFs?

# Notify reviewers


pom.xml
```diff
@@ -31,7 +31,7 @@
 		<frontend.version>v6.0.25</frontend.version>
 
         <!-- THIS SHOULD BE KEPT IN SYNC TO VERSION IN CGDS.SQL -->
-		<db.version>2.13.1</db.version>
+		<db.version>2.13.556</db.version>
```
inodb, line 34: should this be `2.13.2`?

dippindots (commented):
@jamesqo It looks good, I guess you are going to change the db.version before merge it.
Another thing I'd like to say is that I'm working on syncing the `cgds.sql` and `migration.sql` files. These two files have slight differences. I will make a PR soon, and hopefully, we can deploy these changes together. Do you have a timeframe for your work (when do you want to merge and deploy)?

jamesqo:
This has been successfully tested (both via unit tests + importing records past the 32-bit limit) and should be ready for review / merge.

sheridancbio:
The description at the top of this PR currently claims that all integer primary index fields are to be updated, but this is no longer the case. The description should be edited.

src/main/resources/db-scripts/migration.sql
```diff
@@ -1024,3 +1024,44 @@ CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
...
+--     END
+
+DROP PROCEDURE IF EXISTS DROP_FOREIGN_KEY_IF_EXISTS;
+-- This all has to be on a single line for the migrate_db script to parse it correctly
+CREATE PROCEDURE DROP_FOREIGN_KEY_IF_EXISTS(IN tableName VARCHAR(64), IN constraintName VARCHAR(64)) BEGIN IF EXISTS (SELECT 1 FROM information_schema.table_constraints WHERE table_schema = DATABASE() AND table_name = tableName AND constraint_name = constraintName AND constraint_type = 'FOREIGN KEY') THEN SET @query = CONCAT('ALTER TABLE ', tableName, ' DROP FOREIGN KEY ', constraintName); PREPARE stmt FROM @query; EXECUTE stmt; DEALLOCATE PREPARE stmt; END IF; END;
+
+CALL DROP_FOREIGN_KEY_IF_EXISTS('clinical_event_data', 'clinical_event_data_ibfk_1');
+
+-- Step 2: Change dtype of primary autoincrement keys
+ALTER TABLE `clinical_event` MODIFY COLUMN `CLINICAL_EVENT_ID` BIGINT NOT NULL AUTO_INCREMENT;
```
sheridancbio, line 1057: Here we refer to the AUTO_INCREMENT attribute of the field. I believe this can accept a value to reset the next integer to be allocated on an auto increment field completion. When you do not provide a value, I suppose the next value to be allocated remains the same (as it was before changing the datatype). Is that known? Hopefully it does not reset the value to zero (which would cause data corruption). mysql is pretty mature and well developed so I suspect a shortcut approach of "just reset to zero when the datatype changes" would not be used.

src/main/resources/db-scripts/migration.sql
```diff
@@ -1024,3 +1024,44 @@ CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
...
+
+-- Step 3: Change dtype of foreign keys
+ALTER TABLE `clinical_event_data` MODIFY COLUMN `CLINICAL_EVENT_ID` BIGINT NOT NULL;
+
+-- Step 4: Re-add foreign key constraints
+ALTER TABLE `clinical_event_data` ADD CONSTRAINT FOREIGN KEY (`CLINICAL_EVENT_ID`) REFERENCES `clinical_event` (`CLINICAL_EVENT_ID`) ON DELETE CASCADE;
```
sheridancbio, line 1063: 👍

src/main/resources/db-scripts/migration.sql
```diff
@@ -1024,3 +1024,44 @@ CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
...
+--             DEALLOCATE PREPARE stmt; 
+--         END IF; 
+--     END
+
+DROP PROCEDURE IF EXISTS DROP_FOREIGN_KEY_IF_EXISTS;
+-- This all has to be on a single line for the migrate_db script to parse it correctly
+CREATE PROCEDURE DROP_FOREIGN_KEY_IF_EXISTS(IN tableName VARCHAR(64), IN constraintName VARCHAR(64)) BEGIN IF EXISTS (SELECT 1 FROM information_schema.table_constraints WHERE table_schema = DATABASE() AND table_name = tableName AND constraint_name = constraintName AND constraint_type = 'FOREIGN KEY') THEN SET @query = CONCAT('ALTER TABLE ', tableName, ' DROP FOREIGN KEY ', constraintName); PREPARE stmt FROM @query; EXECUTE stmt; DEALLOCATE PREPARE stmt; END IF; END;
+
+CALL DROP_FOREIGN_KEY_IF_EXISTS('clinical_event_data', 'clinical_event_data_ibfk_1');
+
```
sheridancbio, line 1055: Now that we have executed the stored procedure I think it would make sense to clean it up so that it is not left on the MYQSL system without a known future purpose. So I suggest adding another statement here for:
DROP PROCEDURE IF EXISTS DROP_FOREIGN_KEY_IF_EXISTS;

pom.xml
```diff
@@ -31,7 +31,7 @@
 		<frontend.version>v6.0.25</frontend.version>
 
         <!-- THIS SHOULD BE KEPT IN SYNC TO VERSION IN CGDS.SQL -->
-		<db.version>2.13.1</db.version>
+		<db.version>2.14.0</db.version>
```
sheridancbio, line 34: I had the impression that @dippindots believed that because we were doing something which was backwards compatible (only expanding the range of an integer field) that this change could be a simple point release increment rather than a minor number increment. But whatever makes sense ... I think this case is kind of "in between" the two examples given in the guidelines. We are not renaming a table or a field ... but we are altering a field (which would require an installer to update their deployed cBioPortal backend at the time of migration). Under that consideration maybe it is a minor number increment.

sheridancbio (approved):
I'm approving this PR, but please read my suggestions and concerns before merge.

src/main/resources/db-scripts/migration.sql
```diff
@@ -1024,3 +1024,44 @@ CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
 CREATE INDEX idx_clinical_event_value ON clinical_event_data (`VALUE`);
 CREATE INDEX idx_sample_stable_id ON sample (`STABLE_ID`);
 UPDATE `info` SET `DB_SCHEMA_VERSION`="2.13.1";
+
+##version: 2.14.0
+
+-- Step 1: Drop all foreign keys
+SET FOREIGN_KEY_CHECKS=0;
+
+-- CREATE PROCEDURE DROP_FOREIGN_KEY_IF_EXISTS(IN tableName VARCHAR(64), IN constraintName VARCHAR(64))
```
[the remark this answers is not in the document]
sheridancbio: I had included a longer discussion of why this stored procedure seems to be needed ... and expressed an inclination to find a way to not use stored procedures if it wasn't really necessary. Somehow my extra comments were truncated. In brief, they were:
- I guess we don't know for sure what name was auto-assigned to the foreign key constraint at the time of creation. Different versions of MYSQL may have different algorithms for creating constraint names, so referencing the constraint by name is not reliable (because it was not explicitly named in the first place).
- now we are dealing with only a single constraint between two tables, so I wondered if there was another way to simply query the system to discover the constraint name and then delete it through a name reference (rather than a stored procedure). I acknowledged that this might be more work than it is worth. Also we intend to move towards a system without a mysql database component. If we get there, this point becomes irrelevant.

src/main/resources/db-scripts/migration.sql
```diff
@@ -1024,3 +1024,44 @@ CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
 CREATE INDEX idx_clinical_event_value ON clinical_event_data (`VALUE`);
 CREATE INDEX idx_sample_stable_id ON sample (`STABLE_ID`);
 UPDATE `info` SET `DB_SCHEMA_VERSION`="2.13.1";
+
+##version: 2.14.0
+
+-- Step 1: Drop all foreign keys
+SET FOREIGN_KEY_CHECKS=0;
```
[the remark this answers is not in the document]
jamesqo: I think you're right, the runner script seems to create a new connection for every statement in migration.sql, so these lines don't actually have an effect because they are executed in separate transactions. I actually don't think they're needed in the first place because dropping / re-adding the constraints ignores this setting, so I'll remove them.

src/main/resources/db-scripts/migration.sql
```diff
@@ -1024,3 +1024,44 @@ CREATE INDEX idx_clinical_event_key ON clinical_event_data (`KEY`);
...
+--         END IF; 
+--     END
+
+DROP PROCEDURE IF EXISTS DROP_FOREIGN_KEY_IF_EXISTS;
+-- This all has to be on a single line for the migrate_db script to parse it correctly
+CREATE PROCEDURE DROP_FOREIGN_KEY_IF_EXISTS(IN tableName VARCHAR(64), IN constraintName VARCHAR(64)) BEGIN IF EXISTS (SELECT 1 FROM information_schema.table_constraints WHERE table_schema = DATABASE() AND table_name = tableName AND constraint_name = constraintName AND constraint_type = 'FOREIGN KEY') THEN SET @query = CONCAT('ALTER TABLE ', tableName, ' DROP FOREIGN KEY ', constraintName); PREPARE stmt FROM @query; EXECUTE stmt; DEALLOCATE PREPARE stmt; END IF; END;
+
+CALL DROP_FOREIGN_KEY_IF_EXISTS('clinical_event_data', 'clinical_event_data_ibfk_1');
+
+-- Step 2: Change dtype of primary autoincrement keys
```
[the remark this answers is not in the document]
jamesqo: Will do

src/main/resources/db-scripts/migration.sql
jamesqo: Yes, the auto_increment value will remain the same in MySQL when the datatype of the primary key is changed: https://stackoverflow.com/a/65998566

src/main/resources/db-scripts/migration.sql
jamesqo: Sure

src/main/resources/db-scripts/migration.sql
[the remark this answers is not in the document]
jamesqo: @sheridancbio 

- The naming convention `<table>_ibfk_<number>` is always used for foreign keys provided that InnoDB is being used as the storage engine. I assume that's always the case, or are we letting users change the storage engine? I don't think other engines have support for foreign key constraints, anyhow.
- You're right-- I have looked into the possibility of dynamically querying the constraint name rather than hard-coding it, but it seems to be more effort than it's worth.
- Provided that the constraint name is always standardized, the new constraint we add should have the same name too. But I agree that it's a bit shaky in any case and it wouldn't hurt to have explicit names in the future.

dippindots (approved):
@jamesqo I'm approving this PR as it looks good overall. The only comment I have is that @sheridancbio mentioned the "CREATE Routine" privileges that other instance admins might need to execute this routine.

We should document any necessary updates here: [Migration Guide](https://github.com/cBioPortal/cbioportal/blob/master/docs/Migration-Guide.md).

jamesqo:
OK sounds good! I will merge this PR then, and take care of the documentation updates separately.

dippindots:
This pr contains DB updates, our current integration tests can't support DB updates correctly, we created another pr to verify tests: https://github.com/cBioPortal/cbioportal/pull/11471
The tests look good. We will merge this PR as is. The tests will be broken for a while until the related `cbioportal-core` package is updated.
We need to do some follow-ups:
- [ ] Update `cbioportal-core` and make a new release
- [ ] Update `cbioportal-core` release in docker file: https://github.com/cBioPortal/cbioportal/blob/master/docker/web-and-data/Dockerfile#L47

PR #11414, opened 2025-02-24
Fix file paths in the docker documentation

rmadupuri:
- Made `study_es_0` accessible within the container. 
- Updated script paths so the example commands can be executed.

# Checks

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/docker/web-and-data/Dockerfile
+++ b/docker/web-and-data/Dockerfile
@@ -48,6 +48,7 @@ RUN wget https://github.com/cBioPortal/cbioportal-core/releases/download/v1.0.9/
 
 
 COPY --from=build /cbioportal/src/main/resources/db-scripts /cbioportal/db-scripts
+COPY --from=build /cbioportal/test/test_data/study_es_0 /cbioportal/test/study_es_0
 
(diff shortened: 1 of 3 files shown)


inodb (approved):
Thanks for the fixes @rmadupuri !

docker/web-and-data/Dockerfile
```diff
@@ -48,6 +48,7 @@ RUN wget https://github.com/cBioPortal/cbioportal-core/releases/download/v1.0.9/
 
 
 COPY --from=build /cbioportal/src/main/resources/db-scripts /cbioportal/db-scripts
+COPY --from=build /cbioportal/test/test_data/study_es_0 /cbioportal/test/study_es_0
```
inodb, line 51: I'm not sure if the test data should go into the main app image. Kinda feels like it would be better in e.g. a pre populated mysql image

Just curious: why did u want to include it here?

inodb: Ah nvm I see now. It was included originally before we moved core out. Ok that makes sense

Issue #11291, opened 2024-12-16
Structural Variant Genes showing TFG gene with freq 1100%


fuzhaoyuan:
Found by one of my co-worker Erica at DFCI.

Link: https://genie.cbioportal.org/study/summary?id=genie_public#filterJson={%22clinicalDataFilters%22:[{%22attributeId%22:%22CANCER_TYPE%22,%22values%22:[{%22value%22:%22NON-SMALL%20CELL%20LUNG%20CANCER%22},{%22value%22:%22MESOTHELIOMA%22},{%22value%22:%22THYMIC%20TUMOR%22},{%22value%22:%22LUNG%20CANCER,%20NOS%22},{%22value%22:%22SMALL%20CELL%20LUNG%20CANCER%22}]}],%22studyIds%22:[%22genie_public%22],%22alterationFilter%22:{%22copyNumberAlterationEventTypes%22:{%22AMP%22:true,%22HOMDEL%22:true},%22mutationEventTypes%22:{%22any%22:true},%22structuralVariants%22:null,%22includeDriver%22:true,%22includeVUS%22:true,%22includeUnknownOncogenicity%22:true,%22includeUnknownTier%22:true,%22includeGermline%22:true,%22includeSomatic%22:true,%22includeUnknownStatus%22:true,%22tiersBooleanMap%22:{}}}

<img width="402" alt="image" src="https://github.com/user-attachments/assets/3c0dfcfa-00bd-4d3e-9356-d621d96af5d6" />

adarsh7raj:
Is this is a bug?

PR #11447, opened 2025-03-12
Fix freqency error in ClickHouse

### LINKED ISSUE #11431: [BUG] In combined study scenario, profiled count is not properly reported
https://genie.cbioportal.org/study/summary?id=genie_public
mutate_genes endpoint shows 

https://genie.cbioportal.org/study/summary?id=genie_public%2Cerbb2_genie_public

You can see here that in combined study view, the profiled count for this gene is 2.  When looking only at genie study alone, the profiled count is ~200k.    Presumably because it's WES study.  
Image

We need to figure out why the profiledCount is wrong in combination scenario

Image

fuzhaoyuan:
Fix #11431 and #11291

## Fix: Corrects Inflated Alteration Frequencies in Study View

**Problem:**
Users were observing incorrect and often excessively high alteration frequencies (e.g., >1000%) in the Study View for mutations, CNAs, and structural variants (SVs). The formula for frequency is `alteration count / profiled count`. These inflated frequencies occurred when the `profiled count` (denominator) was inappropriately small for the given `alteration count` (numerator).

**Root Cause & Previous Logic:**
The core issue stemmed from how profiled counts were determined, especially when combining data across multiple studies or handling SVs:

1.  **Original Logic:**
    * If a gene had alterations and was covered by specific gene panels (e.g., targeted panels or WES), the `profiled count` was the sum of samples sequenced with those panels (`non-WES count + WES count`).
    * If a gene had alterations but was *not* explicitly covered by a panel for those altered samples (i.e., it was "Altered & Not Profiled" on a specific panel), the system would use the `Total (profiled) count` of all samples in that study as the denominator.

2.  **Multi-Study Breakdown:**
    * This logic failed in combined study views. For example, Gene MT-ND5 had 896 alterations in Study 1 (where it was not on a panel, so `Total count` of 197,980 was correctly used, giving ~0.5% frequency). Study 2 had 0 alterations but 2 samples where MT-ND5 *was* on a panel (WES).
    * When combined, the 896 alterations from Study 1 were incorrectly divided by the small panel count (2) from Study 2, because the aggregated data indicated *some* panel coverage for MT-ND5, leading to a frequency of 44,800%. The aggregation lost the critical context that Study 1's alterations should use Study 1's `Total count`.
image

3.  **Structural Variant (SV) Issue:**
    * A similar problem affected SVs even within single studies. An SV involves two genes (e.g., site1 and site2). If site1 was an SV partner (and thus had an "alteration") but was not itself on a targeted panel for many samples where the SV was found (perhaps only covered by WES in a few samples), its alteration count would be divided by this small WES panel count, again leading to highly inflated frequencies (e.g., TACC3 at 4044.4%).
image

**Initial Attempt & Performance Issues:**
A first attempt to fix the multi-study mutation bug involved fetching detailed, per-study counts to apply the correct denominator logic (panel vs. total count) before aggregation. While this corrected the MT-ND5 combined frequency, it was computationally expensive and resulted in very slow filter response times, and it didn't solve the underlying SV issue.

**Solution: The `off_panel` Filter**
The persistent SV issue led to the insight that alterations should only be counted if the gene was *intentionally targeted* by the sequencing panel used for that specific sample. Alterations found in genes not on the specific panel for a sample are termed "off-panel" alterations.

1.  **`off_panel` Flag:** An `off_panel` boolean flag was added to the `genomic_event` data. This flag is `True` if an alteration occurs in a sample where the gene was *not* on the panel used for that sample.
2.  **Filtering at Source:** Queries now include `WHERE off_panel = False`.

**Why This Works & Benefits:**
* **Correct Alignment:** By filtering out `off_panel` alterations, the `alteration count` (numerator) now only includes alterations from samples where the gene was *intentionally profiled* on the panel used.
* **Simplified Logic:** This eliminates the need for the complex conditional logic of choosing between `panel count` and `total count` as the denominator. The denominator is now consistently the count of samples profiled by panels that *include the gene*.
* **Comprehensive Fix:** This elegantly resolves both the multi-study mutation/CNA frequency bug and the single-study SV frequency bug. For MT-ND5, the 896 off-panel alterations in Study 1 are filtered out. For TACC3, its off-panel SV alterations are filtered.
* **Performance:** Adding a `WHERE` clause at the database query level is far more efficient than the previous complex data fetching and application-side logic. Pre-computing the `off_panel` flag during data loading means query-time impact is negligible.

This change ensures that alteration frequencies displayed in the Study View are accurate and reliably reflect the proportion of *intentionally profiled* samples that harbor a given alteration, regardless of the number of studies or alteration type.


src/main/java/org/cbioportal/domain/alteration/usecase/AbstractAlterationCountByGeneUseCase.java
```diff
@@ -74,10 +76,19 @@ private boolean hasGenePanelData(@NonNull Set<String> matchingGenePanelIds) {
...
     private int computeTotalProfiledCount(boolean hasGenePanelData, int alterationsProfiledCount,
-                                           int sampleProfileCountWithoutGenePanelData, int totalProfiledCount) {
-        int profiledCount = hasGenePanelData ? alterationsProfiledCount + sampleProfileCountWithoutGenePanelData
-                : sampleProfileCountWithoutGenePanelData;
-        return profiledCount == 0 ? totalProfiledCount : profiledCount;
+                                          Map<String, Integer> studyIdToWESProfiledCount, Map<String, Integer> studyIdToSampleProfiledCount) {
+        if (hasGenePanelData) {
+            return alterationsProfiledCount + studyIdToWESProfiledCount.values().stream().mapToInt(Integer::intValue).sum();
+        } else {
+            Map<String, Integer> updatedWESProfiledCount = new HashMap<>(studyIdToWESProfiledCount);
```
alisman, line 83: this logic here might be more clear if it used reduce to produce a sum rather than going through intermediary step of populating the hashmap.   
also, would it be possible to get rid of the confusing ```if hasGEnePanelData``` statement.  Isn't it always a sum of alterationProfiledCount + WEScount, just with possibility that alterationProfiledCount could be zero?

alisman: and lastly, we just need comment the CRAP out of all this code so that we don't lose the understanding next time

fuzhaoyuan: 1. This is the logic for
```
No gene panel data, then some studies could have WES count while others don't ->
First use sample count for all studies that have no WES count
Then add all WES count
```
Every study will have a sample count. If not using a HashMap we can't tell which sample count to fill in the missing WES count for which study.
2. Get rid of hasGenePanelData might be possible. But we're still pending the Alteration Count fix.
3. I'll add in comments, and tests

alisman: I see.  So we're falling back to studyIdToSampleProfiledCount when there is no WES count.  But this assumes that there is an alteration in the subject gene FOR THAT STUDY, right?   i mean, it's no just any study which is going to get this fallback, it's only when there is an alteration but no WES profile, we then make the assumption that it would have been possible for any sample i.e. we consider it profiled?

alisman (approved):

PR #11463, opened 2025-03-21
Update CancerStudy with imaging sample counts

gblaih:
Update CancerStudy type with field imagingSampleCounts and calculate these counts

### DIFF
--- a/src/main/java/org/cbioportal/legacy/model/CancerStudy.java
+++ b/src/main/java/org/cbioportal/legacy/model/CancerStudy.java
@@ -34,6 +34,7 @@ public class CancerStudy implements ReadPermission, Serializable {
     private Boolean readPermission = true;
     private Integer treatmentCount;
     private Integer structuralVariantCount;
+    private Integer imagingSampleCount;
     
     public Integer getCancerStudyId() {
         return cancerStudyId;
@@ -247,5 +248,13 @@ public class CancerStudy implements ReadPermission, Serializable {
 
     public void setStructuralVariantCount(Integer structuralVariantCount) {
         this.structuralVariantCount = structuralVariantCount;
+    }
+
+    public Integer getImagingSampleCount() {
+        return imagingSampleCount;
+    }
+
+    public void setImagingSampleCount(Integer imagingSampleCount) {
+        this.imagingSampleCount = imagingSampleCount;
     }
 }
--- a/src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
+++ b/src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
@@ -36,6 +36,7 @@
         COUNT(CASE WHEN sample_list.STABLE_ID = CONCAT(cancer_study.CANCER_STUDY_IDENTIFIER,'_3way_complete') THEN 1 ELSE NULL END) AS completeSampleCount,
         IFNULL(treatment.count, 0 ) as treatmentCount,
         COALESCE(structural_variant.count, 0) as structuralVariantCount,
+        IFNULL(imaging.count, 0 ) as imagingSampleCount,
         type_of_cancer.TYPE_OF_CANCER_ID AS "typeOfCancer.typeOfCancerId"
         <if test="projection == 'SUMMARY' || projection == 'DETAILED'">
             ,
@@ -73,6 +74,17 @@
                 INNER JOIN structural_variant on structural_variant.SAMPLE_ID = sample.INTERNAL_ID
                 GROUP BY patient.CANCER_STUDY_ID
             ) as structural_variant on structural_variant.CANCER_STUDY_ID = cancer_study.CANCER_STUDY_ID
+            LEFT JOIN
+            (
+                SELECT COUNT(DISTINCT rs.INTERNAL_ID) as count,
+                patient.CANCER_STUDY_ID as CANCER_STUDY_ID
+                FROM cancer_study
+                INNER JOIN patient on cancer_study.CANCER_STUDY_ID = patient.CANCER_STUDY_ID
+                LEFT JOIN sample on patient.INTERNAL_ID = sample.PATIENT_ID
+                INNER JOIN resource_sample rs on rs.INTERNAL_ID = sample.INTERNAL_ID
+                WHERE rs.RESOURCE_ID NOT IN ('CHROMOSCOPE', 'FIGURES')
+                GROUP BY patient.CANCER_STUDY_ID
+            ) as imaging on imaging.CANCER_STUDY_ID = cancer_study.CANCER_STUDY_ID
         </if>
     </sql>
 


dippindots (commented):
@gblaih Thanks for making efforts and making this aviablable! This is great!

Just one quick comment: I think we can go ahead and merge this as a solid initial approach. That said, we should also explore ways to automatically discover new resource types. For example, if we add a new imaging type like new_image_type, it would be ideal for it to appear on the homepage automatically. Even if this requires introducing a new endpoint or doing some additional refactoring, I believe it's worth the effort.

dippindots (approved):

src/main/java/org/cbioportal/legacy/service/impl/StudyServiceImpl.java
```diff
@@ -44,6 +46,24 @@ public List<CancerStudy> getAllStudies(
...
         studyRepository.getAllStudies(keyword, projection, pageSize, pageNumber, sortBy, direction);
+
+    List<ResourceCount> resources = studyRepository.getResourcesForAllStudies();
+    Map<String, CancerStudy> studyMap =
+        allStudies.stream()
+            .collect(Collectors.toMap(CancerStudy::getCancerStudyIdentifier, Function.identity()));
+
+    for (CancerStudy study : allStudies) {
+      study.setResources(new ArrayList<>());
+    }
```
dippindots: Do we need to initialize `resourceCounts` for all studies? I think we only need to initialize it for studies that actually have resource data.

src/main/java/org/cbioportal/legacy/service/impl/StudyServiceImpl.java
```diff
@@ -97,6 +117,9 @@ public CancerStudy getStudy(String studyId) throws StudyNotFoundException {
       throw new StudyNotFoundException(studyId);
     }
 
+    List<ResourceCount> resources = studyRepository.getResources(studyId);
+    cancerStudy.setResources(resources);
```
dippindots: we also need to add resourceCounts to method `fetchStudies`

src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
```diff
@@ -178,4 +178,86 @@
...
+            CANCER_STUDY_IDENTIFIER as cancerStudyIdentifier,
+            rd.CUSTOM_METADATA AS customMetaData,
+            COUNT(DISTINCT rs.INTERNAL_ID) AS sampleCount,
+            COUNT(DISTINCT p.INTERNAL_ID) AS patientCount
+        FROM resource_sample AS rs
+                 INNER JOIN resource_definition rd ON rd.RESOURCE_ID = rs.RESOURCE_ID
+                 INNER JOIN sample s ON s.INTERNAL_ID = rs.INTERNAL_ID
+                 INNER JOIN patient p ON p.INTERNAL_ID = s.PATIENT_ID
+                 INNER JOIN cancer_study cs ON cs.CANCER_STUDY_ID = p.CANCER_STUDY_ID
+        WHERE rs.RESOURCE_ID NOT IN ('CHROMOSCOPE')
```
dippindots: Let's remove the hardcoded values for `CHROMOSCOPE` and `FIGURES` from this PR. It would be better to handle this on the frontend—either by hardcoding them there or by using the new JSON field.

src/main/java/org/cbioportal/legacy/service/impl/StudyServiceImpl.java
```diff
@@ -44,6 +46,24 @@ public List<CancerStudy> getAllStudies(
...
+      study.setResources(new ArrayList<>());
+    }
+
+    for (ResourceCount rc : resources) {
+      String cancerStudyIdentifier = rc.getCancerStudyIdentifier();
+      CancerStudy study = studyMap.get(cancerStudyIdentifier);
+      if (study != null) {
+        study.getResources().add(rc);
+      }
+    }
```
dippindots: We should only add `resourceCounts` when projection is `SUMMARY` or `DETAILED`
See https://github.com/cBioPortal/cbioportal/blob/master/src/main/java/org/cbioportal/legacy/web/parameter/Projection.java

src/main/java/org/cbioportal/legacy/model/CancerStudy.java
```diff
@@ -33,6 +34,7 @@ public class CancerStudy implements ReadPermission, Serializable {
   private Boolean readPermission = true;
   private Integer treatmentCount;
   private Integer structuralVariantCount;
+  private List<ResourceCount> resources;
```
dippindots: Let's rename this and related stuff to `resourceCounts`

dippindots (commented):
@gblaih Overall, the PR looks great and works well for demo purposes. My comments are mainly focused on improving code clarity and aligning better with API design principles. Here's a breakdown:

- **Rename `resources` to `resourceCounts`**  
  To better reflect the data type (`List<ResourceCount>`), let's rename related variables and methods in the model, service, and repository layers. (Apologies for initially suggesting `resources` — your original `ResourceCount` naming is actually more descriptive.)

- **Conditionally include `resourceCounts` based on `projection`**  
  Since `projection` controls the response detail in our API, we should include `resourceCounts` only when `projection` is `SUMMARY` or `DETAILED`. We can handle this with a simple conditional in the service layer.

- **Add support in `fetchStudies`**  
  `fetchStudies` is the third method that needs to add `resourceCounts`, but it’s currently missing from the implementation.

- **Remove hard‑coded `CHROMOSCOPE` and `FIGURES`**  
  Let’s avoid hardcoding these in the backend. Instead, handle them in the frontend—either by hardcoding there or by leveraging the new JSON field.

src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
```diff
@@ -178,4 +178,98 @@
...
+            rd.CUSTOM_METADATA AS customMetaData,
+            COUNT(DISTINCT rs.INTERNAL_ID) AS sampleCount,
+            COUNT(DISTINCT p.INTERNAL_ID) AS patientCount
+        FROM resource_sample AS rs
+                 INNER JOIN resource_definition rd ON rd.RESOURCE_ID = rs.RESOURCE_ID
+                 INNER JOIN sample s ON s.INTERNAL_ID = rs.INTERNAL_ID
+                 INNER JOIN patient p ON p.INTERNAL_ID = s.PATIENT_ID
+                 INNER JOIN cancer_study cs ON cs.CANCER_STUDY_ID = p.CANCER_STUDY_ID
+        <where>
+            <if test="studyIds != null and studyIds.size > 0">
```
dippindots:
```suggestion
            <if test="studyIds != null and !studyIds.isEmpty()">
```
We can just use isEmpty() function to see if it has any elements here, this can also apply to line#264

src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
```diff
@@ -178,4 +178,98 @@
...
+            rd.OPEN_BY_DEFAULT AS openByDefault,
+            CANCER_STUDY_IDENTIFIER as cancerStudyIdentifier,
+            rd.CUSTOM_METADATA AS customMetaData,
+            COUNT(DISTINCT rs.INTERNAL_ID) AS sampleCount,
+            COUNT(DISTINCT p.INTERNAL_ID) AS patientCount
+        FROM resource_sample AS rs
+                 INNER JOIN resource_definition rd ON rd.RESOURCE_ID = rs.RESOURCE_ID
+                 INNER JOIN sample s ON s.INTERNAL_ID = rs.INTERNAL_ID
+                 INNER JOIN patient p ON p.INTERNAL_ID = s.PATIENT_ID
+                 INNER JOIN cancer_study cs ON cs.CANCER_STUDY_ID = p.CANCER_STUDY_ID
```
dippindots: There are some shared code blocks between these two methods, I think we can reuse some of the code by creating a shared sql to be referenced.
You can see reference implementation here.
https://github.com/cBioPortal/cbioportal/blob/master/src/main/resources/org/cbioportal/legacy/persistence/mybatis/GeneMapper.xml#L6-L20

dippindots (commented):
Your updates looks great! I just have two more comments related to the mapper file

src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
```diff
@@ -180,95 +220,43 @@
...
-                 INNER JOIN resource_definition rd ON rd.RESOURCE_ID = rs.RESOURCE_ID
-                 INNER JOIN sample s ON s.INTERNAL_ID = rs.INTERNAL_ID
-                 INNER JOIN patient p ON p.INTERNAL_ID = s.PATIENT_ID
-                 INNER JOIN cancer_study cs ON cs.CANCER_STUDY_ID = p.CANCER_STUDY_ID
+        <include refid="selectSampleResourceCounts"/>
+        <include refid="fromResourceSample"/>
         <where>
-            <if test="studyIds != null and studyIds.size > 0">
+            <if test="studyIds != null and !studyIds.isEmpty()">
                 cancer_study.CANCER_STUDY_IDENTIFIER IN
```
dippindots: I found the issue:
need to replace this with
```
cs.CANCER_STUDY_IDENTIFIER IN
```
The reason is when you give a table an alias (e.g. `cancer_study cs`), that alias `cs` replaces the original table name in the rest of the query. Using the full table name afterward will cause an error

src/main/resources/org/cbioportal/legacy/persistence/mybatis/StudyMapper.xml
```diff
@@ -180,95 +220,43 @@
...
-                 INNER JOIN patient p ON cs.CANCER_STUDY_ID = p.CANCER_STUDY_ID
-                 LEFT JOIN sample s ON p.INTERNAL_ID = s.PATIENT_ID
-                 INNER JOIN resource_patient rp ON rp.INTERNAL_ID = s.PATIENT_ID
-                 INNER JOIN resource_definition rd ON rd.RESOURCE_ID = rp.RESOURCE_ID
+        <include refid="selectPatientResourceCounts"/>
+        <include refid="fromResourcePatient"/>
         <where>
-            <if test="studyIds != null and studyIds.size > 0">
+            <if test="studyIds != null and !studyIds.isEmpty()">
                 cancer_study.CANCER_STUDY_IDENTIFIER IN
```
dippindots: same here

dippindots (commented):
@gblaih The tests are indicating an real issue, I have some explainations in comments

Issue #10806, opened 2024-05-22
Change COMMENTS column in SV table to longtext instead of varchar(255)


zhx828:
The comment in the structural_variant table is currently limited to 255 characters and often time we see the comment is longer. We'd like to propose to change the column type to longtext.

Example with a long comment: 
Patient view: https://cbioportal.mskcc.org/patient?sampleId=P-0013777-T01-IM5&studyId=mskimpact
Result view: https://cbioportal.mskcc.org/results/structuralVariants?tab_index=tab_visualize&Action=Submit&session_id=664e356f854f636a38612e11&plots_horz_selection=%7B%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D

sridhar-max24:
-- This code changes the data type of the 'comment' column to 'LONGTEXT', allowing for significantly longer comments.
      ALTER TABLE structural_variant MODIFY COLUMN comment LONGTEXT;

I'm new to open-source contributions and wanted to verify if this is the correct solution or if I misunderstood the issue. Please let me know if there are any other considerations or if a different approach would be more suitable."

dippindots:
We should find out what the longest comment in our database is and update the column type to be large enough to handle it. Please also see Rob’s comment about potential ClickHouse performance concerns: https://github.com/cBioPortal/cbioportal/pull/11465#issuecomment-3362705520

PR #11465, opened 2025-03-25
Change COMMENTS column in SV table to longtext instead of varchar(255)

rajeswari2904:
Resolves https://github.com/cBioPortal/cbioportal/issues/10806

- This PR implements the requested enhancements.
- 
The type of comments column in sv table has been changed to longtext from varchar(255).
- Db scripts were updated with relavant changes. 
- Ran all tests to ensure that nothing in the Java part was accidentally broken by running: mvn integration-test
- Please let me know if any modifications are needed. Thanks!

### DIFF
--- a/src/main/resources/db-scripts/cgds.sql
+++ b/src/main/resources/db-scripts/cgds.sql
@@ -399,7 +399,7 @@ CREATE TABLE `structural_variant` (
   `EVENT_INFO` varchar(255),
   `CLASS` varchar(25),
   `LENGTH` int(11),
-  `COMMENTS` varchar(255),
+  `COMMENTS` longtext,
   `SV_STATUS` varchar(25) NOT NULL DEFAULT 'SOMATIC' COMMENT 'GERMLINE or SOMATIC.',
   `ANNOTATION_JSON` JSON,
   PRIMARY KEY (`INTERNAL_ID`),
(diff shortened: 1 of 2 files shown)


sheridancbio:
@dippindots Now that I have looked carefully at this PR I realize that I misunderstood it before. This PR is expanding the capacity of the comments field rather than reducing it.

My comment here is that we currently have two fields of type LONGTEXT in our schema:
- table genetic_alteration has field VALUES
- table genetic_profile_samples has field ORDERED_SAMPLE_LIST

We are worried about fields of type LONGTEXT because during database transfer operations these fields seem to slow down transfer rates and consume more memory during transfer. In particular, we use sling to transfer data into the clickhouse OLAP columnar database and it does not seem to handle these very large fields well when transferring the genetic_alteration table.

I would urge that we consider a smaller maximum field size. According to this:
https://stackoverflow.com/questions/13932750/tinytext-text-mediumtext-and-longtext-maximum-storage-sizes

LONGTEXT fields can contain up to 4 gigabytes of data. I don't think we need to support gigabyte sized comments in our database. MEDIUMTEXT fields can hold up to 16 megabytes, which also seems pretty excessive. A regular TEXT field type can hold up to 64 kilobytes. That would allow for several pages of densely packed text ... something like 8 full screen pages worth of comments. I think that would be adequate. But if we do want to allow the potential of more comments to be included than that, I still think we can avoid LONGTEXT at least.

I think we also are considering reducing the maximum field sizes for the two fields mentioned above. I have looked for the longest values present in our databases for these fields. Out of millions of records in genetic_alteration, only 525 records currently have more than 300KB. None have 1MB. I think we can safely reduce the field size one step for these two tables.

dippindots:
@rajeswari2904 We’ve decided to defer this ticket for now since it may cause performance issues. I’ll close this PR in the meantime. Thanks for taking the time to work on it!

PR #11472, opened 2025-03-28
Enable HTTP streaming by eliminating unnecessary response copying.

forus:
To enable streaming for Study Data Export (RFC95). Unlike caching http requests `ContentCachingRequestWrapper` for enabling multiple request body reads, wrapping responses in `ContentCachingResponseWrapper` is not necessary. It is not used anywhere. The application caching behaviour (with `@Cachable` annotation) does not depend on any of these.


alisman:
@forus please open this again if this is still necessary.  i think this feature was already merged so hopefully this is defunct

PR #11482, opened 2025-04-04
RFC95: Generating Study data files

forus:
Generating the downloadable files on the fly. See more details here https://docs.google.com/document/d/1SQ033yQUc9VJquIuHon1WXnrnaIwORi114Q0daXLrl4/edit?tab=t.0

- enable downloading studies via /export/study/{study id}.zip
- only works when `dynamic_study_export_mode` flag is on.

FE: https://github.com/cBioPortal/cbioportal-frontend/pull/5185

# Checks

# Any screenshots or GIFs?

# Notify reviewers


src/main/java/org/cbioportal/application/file/model/MafRecord.java
pieterlukasse, line 1: This file reminds me a lot about the one here: https://github.com/cBioPortal/cbioportal-core/blob/main/src/main/java/org/mskcc/cbio/portal/model/ExtendedMutation.java

There is probably great overlap between https://github.com/cBioPortal/cbioportal-core/blob/main/src/main/java/org/mskcc/cbio/portal/model/ and this `application/file/model/` folder. It would be good to add some explanation/motivation on why we're not reusing the existing `org/mskcc/cbio/portal/model/` classes, so it won't confuse future devs. Related to this, I have a question on if/how we plan to deprecate the `org/mskcc/cbio/portal/model/` classes at some point.

forus: Thanks @pieterlukasse, that's a good point worth clarifying.

Our decision not to reuse the existing `org/mskcc/cbio/portal/model/` classes originates from a fundamental difference in their design and purpose. The existing classes are primarily structured to mirror our database tables, meaning they're optimized for direct interaction with and representation of our relational data.

In contrast, the exporter model is specifically designed to represent the structure of our export files. This means its primary function is to accurately map to the layout and content required for the exported data, which often differs significantly from how the data is stored internally in the database.

src/main/java/org/cbioportal/application/file/export/ExportConfig.java
haynescd, line 1: I would utilize Spring Profiles so that would stick with our convention we currently have for clickhouse @Profile("clickhouse")

src/main/java/org/cbioportal/application/file/utils/ZipOutputStreamWriterFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+public class ZipOutputStreamWriterFactory implements FileWriterFactory, Closeable {
```
haynescd: This doesn't really follow the convention of a factory class which is usually stateless 

I think a better name probably Manager or Service

src/main/java/org/cbioportal/application/file/export/ExportController.java
```diff
@@ -0,0 +1,47 @@
...
+      return ResponseEntity.notFound().build();
+    }
+
+    StreamingResponseBody stream =
+        outputStream -> {
+          try (BufferedOutputStream bos = new BufferedOutputStream(outputStream);
+              ZipOutputStreamWriterFactory zipFactory = new ZipOutputStreamWriterFactory(bos)) {
+            exportService.exportData(zipFactory, new ExportDetails(studyId));
+          }
+        };
```
haynescd: Technically you don't need the local variable and can just do this

    return ResponseEntity.ok()
        .contentType(new MediaType("application", "zip"))
        .header("Content-Disposition", "attachment; filename=\"" + studyId + ".zip\"")
        .body(out -> {
          try (BufferedOutputStream bos = new BufferedOutputStream(out);
              ZipOutputStreamWriterFactory zipFactory = new ZipOutputStreamWriterFactory(bos)) {
            exportService.exportData(zipFactory, new ExportDetails(studyId));
          }
        });

src/main/java/org/cbioportal/application/file/utils/ZipOutputStreamWriterFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private final OutputStream outputStream;
+  private final ZipOutputStream zipOutputStream;
```
haynescd: Curious if you tested out different Compression Levels?

src/main/java/org/cbioportal/application/file/export/services/VirtualStudyAwareExportService.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Set;
+import org.cbioportal.application.file.export.exporters.ExportDetails;
+import org.cbioportal.application.file.export.exporters.Exporter;
+import org.cbioportal.application.file.utils.FileWriterFactory;
+import org.cbioportal.legacy.service.VirtualStudyService;
+import org.cbioportal.legacy.web.parameter.VirtualStudy;
+import org.cbioportal.legacy.web.parameter.VirtualStudyData;
+import org.cbioportal.legacy.web.parameter.VirtualStudySamples;
+
+public class VirtualStudyAwareExportService implements Exporter {
```
haynescd: Better named a VirtualStudyExportDecorator

src/main/java/org/cbioportal/application/file/export/services/VirtualStudyAwareExportService.java
```diff
@@ -0,0 +1,90 @@
...
+          fileWriterFactory.setBasePath(exportAsStudyId);
+          ExportDetails newExportDetails =
+              new ExportDetails(
+                  virtualStudySample.getId(), exportAsStudyId, virtualStudySample.getSamples());
+          allStudiesExported &= exportService.exportData(fileWriterFactory, newExportDetails);
+        }
+        fileWriterFactory.setBasePath(null);
+        return allStudiesExported;
+      }
+    }
```
haynescd: Can we break this function into smaller pieces

src/main/java/org/cbioportal/application/file/export/services/ExportService.java
```diff
@@ -0,0 +1,75 @@
...
+    CancerStudyMetadata cancerStudyMetadata =
+        this.cancerStudyMetadataService.getCancerStudyMetadata(studyId);
+    return cancerStudyMetadata != null
+        && (authentication == null
+            || cancerStudyPermissionEvaluator == null
+            || cancerStudyPermissionEvaluator.hasPermission(
+                authentication, studyId, "CancerStudyId", AccessLevel.READ));
+  }
+
+  @Transactional
```
haynescd: Does anything write to the DB?

src/main/java/org/cbioportal/application/file/export/services/ExportService.java
```diff
@@ -0,0 +1,75 @@
...
+  @PreAuthorize(
+      "hasPermission(#exportDetails.studyId, 'CancerStudyId', T(org.cbioportal.legacy.utils.security.AccessLevel).READ)")
+  @Override
+  public boolean exportData(FileWriterFactory fileWriterFactory, ExportDetails exportDetails) {
+    boolean atLeastOneDataFileExportedSuccesfully = false;
+    for (Exporter exporter : exporters) {
```
haynescd, line 51: Nitpick but I would wrap this in a helper method calling runExporters

src/main/java/org/cbioportal/application/file/export/mappers/CancerStudyMetadataMapper.java
```diff
@@ -0,0 +1,11 @@
+package org.cbioportal.application.file.export.mappers;
+
+import java.util.List;
+import org.cbioportal.application.file.model.CancerStudyMetadata;
+import org.cbioportal.application.file.model.CancerType;
+
+public interface CancerStudyMetadataMapper {
```
haynescd: Can we add some javadocs to this. So that we can explicitly say why we have different mappers here.

haynescd: For all the mappers

src/main/java/org/cbioportal/application/file/export/services/CancerStudyMetadataService.java
```diff
@@ -0,0 +1,23 @@
+package org.cbioportal.application.file.export.services;
+
+import java.util.List;
+import org.cbioportal.application.file.export.mappers.CancerStudyMetadataMapper;
+import org.cbioportal.application.file.model.CancerStudyMetadata;
+import org.cbioportal.application.file.model.CancerType;
+
+public class CancerStudyMetadataService {
```
haynescd, line 8: This feature has created its own architecture... I would prefer it get pulled out sooner rather than later

We have skipped the Repository layer.

Service -> Mapper 

Our convention has been to wrap these in a Repository Layer

src/main/java/org/cbioportal/application/file/export/ExportConfig.java
forus: @haynescd 
Thanks for the suggestion. I understand that we already use `@Profile("clickhouse")`. But I think `@ConditionalOnProperty` is a better choice for feature flags.

In cBioPortal, we may have many feature flags active at the same time. If we use `@Profile`, we have to list all of them in `spring.profiles.active`. This can get messy and hard to manage.

`@ConditionalOnProperty` lets us turn each feature on or off with a simple setting in application properties. It also makes it easy to see all the features (docs) and their default values in one place.

AFAIK `@Profile` is good for setting up different environments, but not great for turning individual features on or off.

If there’s already a widely accepted plan for how we handle feature flags? I just think this way is easier and more flexible.

src/main/java/org/cbioportal/application/file/utils/ZipOutputStreamWriterFactory.java
forus: I agree. It's a good point.

src/main/java/org/cbioportal/application/file/export/services/VirtualStudyAwareExportService.java
forus: I'll name it `VirtualStudyExportDecoratorService` just to follow naming convention of classes in the services package. Let me know what you think.

src/main/java/org/cbioportal/application/file/utils/ZipOutputStreamWriterFactory.java
forus: no, I didn't. Do you think we should use non-default compression level instead?

src/main/java/org/cbioportal/application/file/export/services/CancerStudyMetadataService.java
forus: @haynescd  Yes, I realized that. Do you see any parts of the code in this feature that could fit well in the Repository layer? Or would it just end up calling the mappers? If that's the case, it doesn’t seem very practical at the moment. I'd prefer to postpone adding the Repository layer until there's a clearer need for it.

src/main/java/org/cbioportal/application/file/export/services/ExportService.java
forus: @haynescd No, nothing is written to the database.
I was getting `java.lang.IllegalStateException: A Cursor is already closed.` without it. Maybe due to the fact that no Spring-managed transaction was started.

Plus we get the default (`READ_COMMITTED`?) isolation level on read by using `@Transaction`.

But your question made me think why I didn't use `@Transactional(readOnly = true)`. I will update the code to use it.

src/main/java/org/cbioportal/application/file/export/services/ExportService.java
forus: it won't make the code any simpler though. The `exportData` method now does just that.

src/main/java/org/cbioportal/application/file/export/services/CancerStudyMetadataService.java
forus: We decided to introduce repositories layer for this feature. One useful thing they could do is to abstract mybatis specific classes (e.g. Cursor)

src/main/java/org/cbioportal/application/file/utils/ZipOutputStreamWriterFactory.java
forus: We might do it configurable in the future if needed.

src/main/java/org/cbioportal/application/file/export/ExportConfig.java
forus: We talked with @haynescd , and he will ask the community when we should use conditional properties and when to use profiles. Charles said we might want to use profiles because this feature is big, almost like its own app.

src/main/java/org/cbioportal/application/file/export/services/CancerStudyMetadataService.java
forus: @haynescd I've added repositories here https://github.com/cBioPortal/cbioportal/pull/11482/commits/b346ac5bb9d169fba3ec250da24d03b3e3ce2442

Please review if that what you meant.

forus:
@haynescd Please have a look. I've checked all solarqube reported issues and fixed ones I agree with.

haynescd (approved):

haynescd:
Def. make sure you squash this PR

PR #11490, opened 2025-04-09
Fix freqency error in Legacy

### LINKED ISSUE #10967: Mutation count in combined studies with genepanels seems to be incorrect
The combined studyview shows that the IPO13 gene has a 100% mutation frequency (in the mutations table). However, upon closer inspection, the mutations table does not take along profiled genes without mutations in it's calculation. According to the results view the actual mutation rate is 24% as there are multiple samples in the second study that are profiled but do not have mutations.

[Study view of combined study with 100% mutation counts](https://www.cbioportal.org/study/summary?id=brca_hta9_htan_2022%2Cnsclc_mskcc_2015). See bottom of image.

image

[Results view of IPO13 gene with 24% mutation rate](https://www.cbioportal.org/results?cancer_study_list=brca_hta9_htan_2022%2Cnsclc_mskcc_2015&tab_index=tab_visualize&profileFilter=mutations%2Ccna&case_set_id=all&Action=Submit&gene_list=IPO13&show_samples=true)

image

fuzhaoyuan:
Fix #10967 
### Fix freqency error in Legacy: Filter alteration counts by gene panel coverage per study

* For mutations and CNAs, if an altered gene is on a gene panel that doesn't come with its study, it's "off-panel" and hence filtered out.
* SVs are not affected because of their 2 gene-partners nature. The ClickHouse implementation has a workaround by breaking that partnership apart into individual genes and then they can be (and are) filered.

### DIFF
--- a/src/main/java/org/cbioportal/legacy/service/impl/AlterationCountServiceImpl.java
+++ b/src/main/java/org/cbioportal/legacy/service/impl/AlterationCountServiceImpl.java
@@ -6,15 +6,16 @@ import org.cbioportal.legacy.model.AlterationCountByGene;
 import org.cbioportal.legacy.model.AlterationCountByStructuralVariant;
 import org.cbioportal.legacy.model.AlterationFilter;
 import org.cbioportal.legacy.model.CopyNumberCountByGene;
+import org.cbioportal.legacy.model.GenePanel;
+import org.cbioportal.legacy.model.GenePanelData;
+import org.cbioportal.legacy.model.GenePanelToGene;
 import org.cbioportal.legacy.model.MolecularProfile;
 import org.cbioportal.legacy.model.MolecularProfileCaseIdentifier;
 import org.cbioportal.legacy.model.util.Select;
 import org.cbioportal.legacy.persistence.AlterationRepository;
 import org.cbioportal.legacy.persistence.MolecularProfileRepository;
-import org.cbioportal.legacy.persistence.StudyViewRepository;
 import org.cbioportal.legacy.service.AlterationCountService;
-import org.cbioportal.legacy.service.SignificantCopyNumberRegionService;
-import org.cbioportal.legacy.service.SignificantlyMutatedGeneService;
+import org.cbioportal.legacy.service.GenePanelService;
 import org.cbioportal.legacy.service.util.AlterationCountServiceUtil;
 import org.cbioportal.legacy.service.util.AlterationEnrichmentUtil;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -25,6 +26,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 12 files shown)


src/main/java/org/cbioportal/legacy/service/impl/AlterationCountServiceImpl.java
```diff
@@ -271,18 +268,26 @@ private <S extends AlterationCountBase> Pair<List<S>, Long> getAlterationGeneCou
...
                     List<S> studyAlterationCountByGenes = dataFetcher.apply(studyMolecularProfileCaseIdentifiers);
+
                     if (includeFrequency) {
+                        // This call will set matchingGenePanelIds for each alteration count
                         Long studyProfiledCasesCount = includeFrequencyFunction.apply(studyMolecularProfileCaseIdentifiers, studyAlterationCountByGenes);
                         profiledCasesCount.updateAndGet(v -> v + studyProfiledCasesCount);
                     }
-                    AlterationCountServiceUtil.setupAlterationGeneCountsMap(studyAlterationCountByGenes, totalResult);
+
+                    // Filter out genes not covered by gene panels in this study
```
alisman, line 281: lets try to make these comments more precise.   
```filter out alteration counts for "off-panel" genes -- genes which are not profiled by a gene panel```.

File: src/main/java/org/cbioportal/legacy/service/impl/AlterationCountServiceImpl.java [hunk 2]
```diff
@@ -271,18 +268,26 @@ private <S extends AlterationCountBase> Pair<List<S>, Long> getAlterationGeneCou
...
                 .collect(Collectors
                     .groupingBy(identifier -> molecularProfileIdStudyIdMap.get(identifier.getMolecularProfileId())))
-                .values()
-                .forEach(studyMolecularProfileCaseIdentifiers -> {
+                .forEach((studyId, studyMolecularProfileCaseIdentifiers) -> {
+                    // Get alteration data for this study
                     List<S> studyAlterationCountByGenes = dataFetcher.apply(studyMolecularProfileCaseIdentifiers);
+
                     if (includeFrequency) {
+                        // This call will set matchingGenePanelIds for each alteration count
```
alisman, line 276: lets improve this comment, it
's not clear to me

fuzhaoyuan: Addressed

src/main/resources/org/cbioportal/legacy/persistence/mybatis/GenePanelMapper.xml
```diff
@@ -186,4 +186,17 @@
             </choose>
         </where>
     </select>
+
+    <select id="findGeneIdsPresentInGenePanelList" resultType="java.lang.Integer">
```
alisman, line 190: a little worried about the performance of this but there may be no other way

src/main/resources/org/cbioportal/legacy/persistence/mybatis/GenePanelMapper.xml
```diff
@@ -186,4 +186,17 @@
             </choose>
         </where>
     </select>
+
```
alisman, line 189: @fuzhaoyuan  it's not ANY panel, it's the panels which are currently at play in the queried studies, right?   a gene might be profiled by some gene panel that is totally unassociated with the current query

dippindots:
not using this anymore

PR #11503, opened 2025-04-18
add property to set default view oncoprint

TJMKuijpers:
The `FrontendPropertiesServiceImpl.java` is missing the property to set the default view of oncoprint (sample vs patient). 
Added the following line:
`        oncoprint_defaultview("oncoprint.defaultview", null),
`

### DIFF
--- a/src/main/java/org/cbioportal/legacy/service/FrontendPropertiesServiceImpl.java
+++ b/src/main/java/org/cbioportal/legacy/service/FrontendPropertiesServiceImpl.java
@@ -41,6 +41,7 @@ public class FrontendPropertiesServiceImpl implements FrontendPropertiesService 
         civic_url("civic.url", null),
         oncoprint_custom_driver_annotation_binary_default(
             "oncoprint.custom_driver_annotation.binary.default", null),
+        oncoprint_defaultview("oncoprint.defaultview", null),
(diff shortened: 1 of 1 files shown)


alisman:
@TJMKuijpers there is a new sonar issue introduced.  can you fix it?

TJMKuijpers:
@alisman sure, I will have a look at the sonarcloud error

alisman:
@TJMKuijpers looks like there's a conflict now?

alisman:
@TJMKuijpers did you fix the solar issue?  it still seems to report 1 problem

TJMKuijpers:
@alisman 
I looked at the sonar issue but I'm not sure how/if the issue should be resolved:

Sonar does say that we should:
1. Rename oncoprint_defaultview to oncoprint_default_view
2. Capitalize oncoprint_default_view to ONCOPRINT_DEFAULT_VIEW

I think all the definitions in `FrontendPropertiesServiceImpl.java` are formatted like they are in `application.properties` (for example, we also don't separate patientview in `skin_patientview_show_mskcc_slide_viewer`). But I can change the name to `oncoprint_default_view` and create a frontend PR (since in the frontend we are looking for `oncoprint_defaultview`). I don't think that capitalizing the parameter is in line with the rest of the code.

PR #11536, opened 2025-05-12
Add CUSTOM_METADATA column into resource_definition table

### LINKED ISSUE #11468: Expand Resources Data Model
Current challenges:

- Can't have multiple viewers for one sample, e.g.:
    Image
    Note: schema seems to support it, but frontend or API maybe not?
- Some viewers would be nice to list on the homepage and study view (maybe we add some `allow_filtering` field for each resource)
   - Let's add it as a JSON column on the resource definition

dippindots:
Fix #11468 

- This database change will allow more customized metadata gets added for resources
- @sheridancbio made a fix to `label-check.yml`, this test was not able to recognize labels with more than one words


sheridancbio:
The Label Check test has failed. Perhaps we can add label 'includes db changes' ?

end to end tests failing with errors like:
[Chrome 87.0.4280.66 linux #0-30] 2) Custom Bins menu in study view chart header generates bins using min and bin size input fields
[Chrome 87.0.4280.66 linux #0-30] element ("[data-test=bin-size-input]") still not existing after 10000ms

I'm guessing the integration tests are being run in the testing PR (#11537). I'll look there now.

sheridancbio:
I see that in #11537 both the e2e tests and the integration tests are passing. So I guess the only fix needed here is the label. I will add the "includes db changes" label but feel free to change / add labels.

sheridancbio:
I added the "includes db changes" label but the Label Check is still failing. I've verified that the text in the label matches the valid label from labels.txt so I'm not sure why this is failing. Trying to debug.

sheridancbio (approved):
I've added a fix to label-check.yml

Now the 'includes db changes' label is recognized properly.

PR #11561, opened 2025-05-27
Remove Cosmic and MyCancerGenome

### LINKED ISSUE #11548: Remove COSMIC and mycancergenome
Let's remove both COSMIC and mycancergenome from:

- [x] website (hide it basically)
- [ ] SQL dump
- [ ] In backend remove related code

In the future we might bring back updated COSMIC with limited data, but we would prolly store them differently (maybe via Genome Nexus?)

onursumer:
Fix #11548


src/main/resources/db-scripts/migration.sql
```diff
@@ -65,7 +65,6 @@ ALTER TABLE `protein_array_target` MODIFY COLUMN `ENTREZ_GENE_ID` int(11) NOT NU
 ALTER TABLE `sanger_cancer_census` MODIFY COLUMN `ENTREZ_GENE_ID` int(11) NOT NULL;
 ALTER TABLE `gistic_to_gene` MODIFY COLUMN `ENTREZ_GENE_ID` int(11) NOT NULL;
 ALTER TABLE `cna_event` MODIFY COLUMN `ENTREZ_GENE_ID` int(11) NOT NULL;
-ALTER TABLE `cosmic_mutation` MODIFY COLUMN `ENTREZ_GENE_ID` int(11) NOT NULL;
```
dippindots, line 68: @onursumer This change in `migration.sql` looks strange to me, I think we should only adding new lines to the end of the file. e.g.
```sql
DROP TABLE IF EXISTS `cosmic_mutation`;
```

src/main/resources/db-scripts/cgds.sql
```diff
@@ -632,24 +631,6 @@ CREATE TABLE `copy_number_seg_file` (
...
-  `STRAND` varchar(2),
-  `CODON_CHANGE` varchar(255),
-  `ENTREZ_GENE_ID` int(11) NOT NULL,
-  `PROTEIN_CHANGE` varchar(255) NOT NULL,
-  `COUNT` int(11) NOT NULL,
-  `KEYWORD` varchar(50) DEFAULT NULL,
-  KEY (`KEYWORD`),
-  PRIMARY KEY (`COSMIC_MUTATION_ID`),
-  FOREIGN KEY (`ENTREZ_GENE_ID`) REFERENCES `gene` (`ENTREZ_GENE_ID`)
-);
```
dippindots, line 651: Also since we are dropping an existing table, we should also updating db scheme version

dippindots (commented):
@onursumer comments are mainly related to the database shceme update, let me know if you want to have a chat

Issue #11595, opened 2025-06-19
Clinical-data-counts discrepancy/bug


alisman:
https://public-data.cbioportal.org/study/summary?id=gbm_tcga_pan_can_atlas_2018%2Cccle_broad_2019

In subtype clinical data chart, we see that count for GBM_IDHwt is different from legacy.  114 (CH) vs 120 (Legacy).    This makes some sense because in this combo study, there are 7 more samples than patients.  Clickhouse is counting patients with this attribute wherease Legacy is counting samples.   The choice of one or the other seems arbitrary to me (when there are "conflicting" attributes between studies, one sample level the other patient).    We may want to figure out how legacy decides which one to use in this scenario and adopt it. But that's no so important.  The NA and frequency issue (see below). 

On top of this, the frequencies are wrong.  I suspect that this is because of the NA issue but not sure.  See frequency of the above GBM_IDHwt.  Legacy is clear correct.  i.e. 114/2300 is not 2.7%.  The denominator is clearly off.

PR #11600, opened 2025-06-24
Fix Clinical Data NA Count

### LINKED ISSUE #11594: [BUG] Subtype: NA samples are more than total samples
https://public-data.cbioportal.org/study/summary?id=msk_met_2021%2Ccoadread_tcga_pan_can_atlas_2018
Subtype: NA samples are more than total samples (26,504 > 26,369)
(this bug does not exist when turn on legacy=1)

Image

fuzhaoyuan:
Fix #11594 and #11595

# Fix Clinical Data NA Count Bug for Cross-Study Multi-Level Attributes

## Problem

When the same clinical attribute exists at different levels (sample vs patient) across studies, the NA count calculation was incorrect:

- **Study A**: `SUBTYPE` defined at sample level
- **Study B**: `SUBTYPE` defined at patient level

**Bug**: When counting sample-level NA values, Study B samples were incorrectly included as "NA" even though Study B doesn't define sample-level `SUBTYPE`. Similarly for patient-level counts.

**Impact**: 
- Inflated NA counts in clinical data charts
- Incorrect sample/patient filtering when users select "NA" values
- Misleading data interpretation

## Root Cause

The original logic calculated NA as `total_count - actual_data_count` using global totals, without considering which studies actually define the attribute at the queried level.

### For Clinical Data Counts (`getClinicalDataCountsQuery`)

Added study-level filtering to ensure NA calculation only includes studies that actually define the attribute at the current level:

```sql
WITH study_attribute_levels AS (
    SELECT DISTINCT cancer_study_identifier
    FROM clinical_data_derived 
    WHERE type='${type}' AND attribute_name IN (...)
),
corrected_total_counts AS (
    SELECT 
        sal.attribute_name AS attributeId,
        count(DISTINCT sd.sample_unique_id) AS total_count
    FROM study_attribute_levels sal
    JOIN sample_derived sd ON sal.cancer_study_identifier = sd.cancer_study_identifier
    GROUP BY sal.attribute_name
)
```

### For Clinical Data Filtering

Applied the same logic to `categoricalClinicalDataCountFilter` and `numericalClinicalDataCountFilter` to ensure consistent behavior between counting and filtering.

## Changes

- **SQL**: Modified `getClinicalDataCountsQuery` to use study-aware NA calculation
- **SQL**: Updated clinical data filters to use same study-level logic  
- **Tests**: Updated test expectations to reflect correct NA counts

## Test Changes

Updated test expectations where studies without the queried attribute level were previously (incorrectly) counted as NA:

```java
// Before: 4 acc_tcga + 7 study_genie_pub = 11 (incorrect)
// After: 7 study_genie_pub only = 7 (correct)
assertEquals(7, filteredSamples5.size()); 
```

### DIFF
--- a/src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
+++ b/src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
@@ -58,26 +58,62 @@
 
     <sql id="getClinicalDataCountsQuery">
         (
-        -- Step 1: Identify which studies actually have data at the current type level for each attribute
-        WITH study_attribute_levels AS (
+        -- Step 1: Find studies that have this attribute type, sample or patient
+        WITH studies_with_attribute_type AS (
             SELECT DISTINCT
-                cancer_study_identifier,
-                attribute_name
-            FROM clinical_data_derived
-            WHERE type='${type}'
-                AND attribute_name IN
+                cs.cancer_study_identifier,
+                cam.attr_id AS attribute_name
+            FROM clinical_attribute_meta cam
+            JOIN cancer_study cs ON cam.cancer_study_id = cs.cancer_study_id
+            WHERE
+                <choose>
+                    <when test="'${type}' == 'sample'">
+                        cam.patient_attribute = 0
+                    </when>
+                    <otherwise>
+                        cam.patient_attribute = 1
+                    </otherwise>
+                </choose>
+                AND cam.attr_id IN
                 <foreach item="attributeId" collection="attributeIds" open="(" separator="," close=")">
                     #{attributeId}
                 </foreach>
                 <if test="studyViewFilterContext.customDataFilterCancerStudies != null and !studyViewFilterContext.customDataFilterCancerStudies.isEmpty()">
-                    AND cancer_study_identifier IN
+                    AND cs.cancer_study_identifier IN
                     <foreach item="studyId" collection="studyViewFilterContext.customDataFilterCancerStudies" open="(" separator="," close=")">
                         #{studyId}
                     </foreach>
                 </if>
         ),
-        -- Step 2: Get actual data counts
-        clinical_data_query AS (
+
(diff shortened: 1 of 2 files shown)


src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
```diff
@@ -58,60 +58,97 @@
...
+                    </otherwise>
+                </choose>
+                AND attribute_name IN
+                <foreach item="attributeId" collection="attributeIds" open="(" separator="," close=")">
+                    #{attributeId}
+                </foreach>
+            GROUP BY attribute_name, attribute_value
+        ),
+        -- Step 3: Calculate sum of actual data for each attribute
+        clinical_data_sum AS (
```
alisman: clinical_data_attribute_value_count?

src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
```diff
@@ -58,60 +58,97 @@
...
+        -- Step 5: Return actual data counts
         SELECT * FROM clinical_data_query
         UNION ALL
-        SELECT attributeId,
-        'NA' AS value,
-        ((
-        <choose>
-            <when test="'${type}' == 'sample'">
-                <include
-                        refid="org.cbioportal.infrastructure.repository.clickhouse.sample.ClickhouseSampleMapper.getFilteredSampleCount"/>
```
alisman, line 99: this here is the problem the total from which subtraction should take place is not the total number of samples filtered, but the total number of samples with attributes of type sample.  Vice versa for patients.

fuzhaoyuan:
This solution is deemed over-complicated, slow and only solves the NA count issue, remaining the attribute conflicting issue entangled still. With a better understanding of the problem, a new solution is proposed at #11603 that addresses both NA count and the attribute conflicting issue.

PR #11610, opened 2025-07-03
Fix for issue #11609: group comparison should be case-insensitive

### LINKED ISSUE #11609: [BUG] Authorization is broken when cancer study groups are not in uppercase
**Describe the bug**
When authentication and authorization are turned on in application.properties
and when 'group' field in the 'cancer_study' table contains lowercase characters 
then authorization by groups does not work.

**To Reproduce**
1. set authentication=oauth2 and authorization=true in application.properties
2. add a record to 'cancer_study' table with 'groups' field set to 'Group_123' and 'public' field set to 0.
3. add a record to 'authorities' table with 'email' field set to user's email and 'authority' field set to 'Group_123'
4. Go to cbioportal, login and observe home page

**Expected behavior**
Cancer study, specified in the record, should be present at the home page

**Observed behavior**
Cancer study  is NOT present at the home page.

(it  also can not be accessed by any other means because the authority is not granted).

**Root cause**
In **CancerStudyPermisisonEvaluator.java** the below code checks if the user is granted authority by groups:
```
Set<String> grantedAuthorities = getGrantedAuthorities(authentication);
...
if (!Collections.disjoint(groups, grantedAuthorities)) {
      if (log.isDebugEnabled()) {
        log.debug("hasAccessToCancerStudy(), user has access by groups return true");
      }
      return true;
    }
}
```

**getGrantedAuthorities()** method converts all elements of 'grantedAuthorities'  to upper case even when they are not in upper case in the 'authorities' table:

```
private Set<String> getGrantedAuthorities(Authentication authentication) {
...
for (String au : allAuthorities) {
        if (au.toUpperCase().startsWith(appName + ":")) {
          grantedAuthorities.add(au.substring(appName.length() + 1).toUpperCase());
        }
 }
...
```
Elements of 'groups', however, come from 'cancer_study' table as they are. 

**Collections.disjoint()** uses equals and therefore it makes case-sensitive comparison. 
Therefore, 'Group_123" not being equal 'GROUP_123,' the access is not granted

The proposed fix is to implement utility method disjointIgnoreCase (can be private in that file) which should be a trivial task.

I can submit a pull request for this fix.

Best regards,
Alexander

akulyakhtin:
Closes #11609

This request fixes the following issue:
if we have `set authorities=true ` in app properties
and if we have a `group` value in `cancer_study `table `sample_group` 
and if we have authority 'sample_group'  (or 'cbioportal:sample_group', depending on the filtering option) in `authorities `table for a user.

Then the access to that cancer study is not granted to the use even though the authroity matches the cancer study group.

This happens because authorities are convereted to upper case while cancer group names are not.

To fix this the pull request changes `Collections.disjoint` (which is case-sensitive) to a case-insensitive utility method `caseInsensitiveDisjoint`.

### DIFF
--- a/src/main/java/org/cbioportal/application/security/CancerStudyPermissionEvaluator.java
+++ b/src/main/java/org/cbioportal/application/security/CancerStudyPermissionEvaluator.java
@@ -368,7 +368,7 @@ public class CancerStudyPermissionEvaluator implements PermissionEvaluator {
         Arrays.stream(cancerStudy.getGroups().split(";"))
             .filter(g -> !g.isEmpty())
             .collect(Collectors.toSet());
-    if (!Collections.disjoint(groups, grantedAuthorities)) {
+    if (!caseInsensitiveDisjoint(groups, grantedAuthorities)) {
       if (log.isDebugEnabled()) {
         log.debug("hasAccessToCancerStudy(), user has access by groups return true");
       }
@@ -391,6 +391,12 @@ public class CancerStudyPermissionEvaluator implements PermissionEvaluator {
       }
     }
     return toReturn;
+  }
+
+  private static boolean caseInsensitiveDisjoint(Collection<String> c1, Collection<String> c2) {
+    Set<String> upperC1 = c1.stream().map(String::toUpperCase).collect(Collectors.toSet());
+    Set<String> upperC2 = c2.stream().map(String::toUpperCase).collect(Collectors.toSet());
+    return Collections.disjoint(upperC1, upperC2);
   }
 
   private boolean hasAccessToCancerStudy(


adityayadav800 (approved):

KingAlex1985 (approved):

KingAlex1985:
Hi @akulyakhtin .
Take a look at Copilot's recommendations regarding null pointer checks in the pervious comment.
Perhaps you could include these changes as well.

haynescd (approved):
Yea, looks good. I would just address the suggested null checks.

Interesting how this is just now coming up... but the docs do say its not case sensitive.

KingAlex1985:
Hi @akulyakhtin 
Take please a look at Copilot's recommendations regarding null pointer checks in the pervious comment.
Can you include these changes as well.
I guess after these changes this PR will be accepted and merged into master branch.

KingAlex1985:
@haynescd 
Are you able to take over the Copilot's recommendations by yourself without waiting for @akulyakhtin.
Is there a function at github like a "one click take over Copilot's recommendations"?

alisman:
@haynescd is this pre or pos your refactor of that code?  do we need this pr?

PR #11611, opened 2025-07-08
Add posibility to create and drop published virtual studies with custom IDs

forus:
To support RFC96 Virtual Study Authorisation Model we need static human readable ids to assign permissions to them

- a
- b

# Checks

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/src/main/java/org/cbioportal/legacy/service/util/SessionServiceRequestHandler.java
+++ b/src/main/java/org/cbioportal/legacy/service/util/SessionServiceRequestHandler.java
@@ -1,6 +1,6 @@
 package org.cbioportal.legacy.service.util;
 
-import static org.cbioportal.legacy.utils.removeme.Session.*;
+import static org.cbioportal.legacy.utils.removeme.Session.SessionType;
 
 import com.mongodb.BasicDBObject;
 import java.io.Serializable;
@@ -180,6 +180,32 @@ public class SessionServiceRequestHandler {
   }
 
   /**
+   * Creates a virtual study with custom id
+   *
+   * @param virtualStudyData - definition of virtual study
+   * @return virtual study object with id and the virtualStudyData
+   */
+  public VirtualStudy createVirtualStudy(String id, VirtualStudyData virtualStudyData) {
+
+    String url =
+        UriComponentsBuilder.fromUriString(sessionServiceURL)
+            .pathSegment("virtual_study")
+            .pathSegment(id)
+            .build()
+            .toUriString();
+
+    ResponseEntity<VirtualStudy> responseEntity =
+        new RestTemplate()
+            .exchange(
+                url,
+                HttpMethod.POST,
+                new HttpEntity<>(virtualStudyData, getHttpHeaders()),
+                new ParameterizedTypeReference<>() {});
+
+    return responseEntity.getBody();
+  }
+
+  /**
    * Soft delete of the virtual study by de-associating all assigned users.
    *
    * @param id - id of virtual study to soft delete
@@ -206,6 +232,23 @@ public class SessionServiceRequestHandler {
             .toUriString();
 
     new RestTemplate().put(url, new HttpEntity<>(virtualStudy.getData(), getHttpHeaders()));
+  }
+
+  /**
+   * Drop virtual study
+   *
+   * @param virtualStudyId - id of virtual study to drop
+   */
+  public void dropVirtualStudy(String virtualStudyId) {
+
+    String url =
+        UriComponentsBuilder.fromUriString(sessionServiceURL)
+            .pathSegment("virtual_study")
+            .pathSegment(virtualStudyId)
+            .build()
+            .toUriString();
+
+    new RestTemplate().delete(url, new HttpEntity<>(getHttpHeaders()));
   }
 
   private List<PageSettings> getPageSettingsForUser(
--- a/src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
+++ b/src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
@@ -7,8 +7,10 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import org.cbioportal.legacy.service.CancerTypeService;
+import org.cbioportal.legacy.service.StudyService;
 import org.cbioportal.legacy.service.exception.AccessForbiddenException;
 import org.cbioportal.legacy.service.exception.CancerTypeNotFoundException;
+import org.cbioportal.legacy.service.exception.StudyNotFoundException;
 import org.cbioportal.legacy.service.util.SessionServiceRequestHandler;
 import org.cbioportal.legacy.web.parameter.VirtualStudy;
 import org.cbioportal.legacy.web.parameter.VirtualStudyData;
@@ -22,6 +24,7 @@ import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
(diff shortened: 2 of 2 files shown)


.github/workflows/integration-test.yml
```diff
@@ -47,7 +47,8 @@ jobs:
               sed 's|spring.datasource.username=.*|spring.datasource.username=cbio_user|' | \
               sed 's|spring.datasource.password=.*|spring.datasource.password=somepassword|' | \
               sed 's|session.service.url=.*|session.service.url=http://cbioportal-session:5001/api/sessions/my_portal/|' | \
-              sed 's|feature.study.export=.*|feature.study.export=true|' \
+              sed 's|feature.study.export=.*|feature.study.export=true|' | \
+              sed 's|#\?session.endpoint.publisher-api-key=.*|session.endpoint.publisher-api-key=TEST|' \
```
pieterlukasse, line 51: why does your replacement pattern need `|#\?` while others don't?

src/main/java/org/cbioportal/legacy/service/util/SessionServiceRequestHandler.java
```diff
@@ -183,6 +185,32 @@ public VirtualStudy createVirtualStudy(VirtualStudyData virtualStudyData) {
     return responseEntity.getBody();
   }
 
+  /**
+   * Creates a virtual study with custom id
+   *
+   * @param virtualStudyData - definition of virtual study throws DuplicateVirtualStudyException if
+   *     a virtual study with the same id or identical definition already exists
+   */
+  public void createVirtualStudy(String id, VirtualStudyData virtualStudyData) {
```
pieterlukasse:
```suggestion
  public void createVirtualStudy(String virtualStudyId, VirtualStudyData virtualStudyData) {
```

pieterlukasse: to better align with naming in `dropVirtualStudy`

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
```diff
@@ -113,13 +153,28 @@ private void unPublishVirtualStudy(String id) {
...
+
+  /**
+   * Drops public virtual study, removing it from the public list
+   *
+   * @param id - id of public virtual study to drop
+   */
+  private void dropPublicVirtualStudy(String id) {
+    VirtualStudy virtualStudyToUnPublish = sessionServiceRequestHandler.getVirtualStudyById(id);
+    checkIfVSWasPublished(id, virtualStudyToUnPublish.getData());
+    sessionServiceRequestHandler.dropVirtualStudy(virtualStudyToUnPublish.getId());
```
pieterlukasse: naming is a bit confusing here. So we have `getVirtualStudyById(id);` and `dropVirtualStudy(virtualStudyToUnPublish.getId())`. Why not `dropVirtualStudy(id)`?

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
```diff
@@ -113,13 +153,28 @@ private void unPublishVirtualStudy(String id) {
           "The virtual study with id=" + id + " has not been found in the public list.");
     }
     VirtualStudyData virtualStudyData = virtualStudyToUnPublish.getData();
+    checkIfVSWasPublished(id, virtualStudyData);
+    virtualStudyData.setUsers(Set.of(virtualStudyData.getOwner()));
+    sessionServiceRequestHandler.updateVirtualStudy(virtualStudyToUnPublish);
```
pieterlukasse, line 158: what happens when I try to publish a VS that has been unpublished? Will it fail because the ID is the same, will it just overwrite it? If the latter is the case, we might not really need `dropPublicVirtualStudy`.

test/integration/in_service_publish_vs.sh
pieterlukasse, line 1: are we missing a test for drop?

.github/workflows/integration-test.yml
forus: to make it more flexible. We replace the line be it commented or not.
In my previous implementation I chose to specify the feature flag like:
#session.endpoint.publisher-api-key=false

We could choose to change the flag is specified in the example file. But that would require a separate PR to the core project.

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
forus: @pieterlukasse This method publishes a virtual study by its ID. It does not create a new copy; instead, it updates the existing entity in place.
Because the ID remains the same, the operation won't fail. It simply updates the "published" flag without modifying any other fields.

If you need to completely redefine a public virtual study (e.g. change its content), use dropPublicVirtualStudy first.
As far as I know, there is currently no method to update the full payload of a virtual study.

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
forus: @pieterlukasse Did you mean to say "Why not dropVirtualStudyById(id)" ?

test/integration/in_service_publish_vs.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Publishing virtual study with id: $hash"
+publish_status=$(curl -X POST -o /dev/null -s -w "%{http_code}" "http://cbioportal-container:8080/api/public_virtual_studies/$hash" \
+    -H 'X-PUBLISHER-API-KEY: TEST')
+if [ "$publish_status" -eq 200 ]; then
+    echo "Virtual study with id $hash published successfully."
+else
+    echo "Failed with status $publish_status"
+fi
+
+echo "Unpublishing virtual study with id: $hash"
```
forus: @pieterlukasse Not sure, what happened with your comment to this file. We don't test for dropping VS indeed, but we should. I've added to my TODO list

forus: For tests to pass it makes sense to add integration tests for dropping VS after the session service PR is merged (https://github.com/cBioPortal/session-service/pull/58/files), docker released and used here (https://github.com/cBioPortal/cbioportal-docker-compose/blob/master/.env#L2).

So please start with reviewing the session service PR first.

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
pieterlukasse: @forus yes, that might help. But I think I was hinting at the fact that `id` is used in each part, and then in this line the id needs to be taken from `.getId()`. Not clear why the difference.

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
pieterlukasse: ok, got it.

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
forus: @pieterlukasse ah, I see. And yes, using `id` directly does sound like an improvement.

forus: @pieterlukasse Done https://github.com/cBioPortal/cbioportal/pull/11611/commits/0a430cf8151c861cb06add43b02b17d7c102a9e8

test/integration/in_service_publish_vs.sh
forus: Done https://github.com/cBioPortal/cbioportal/pull/11611/commits/b4905b21f85ae89546acf245ec61e0653e5657b7

test/integration/in_service_publish_vs.sh
forus: Done https://github.com/cBioPortal/cbioportal/pull/11611/commits/b4905b21f85ae89546acf245ec61e0653e5657b7

src/main/java/org/cbioportal/legacy/web/PublicVirtualStudiesController.java
```diff
@@ -92,13 +106,39 @@ public ResponseEntity<Void> unPublishVirtualStudy(
    * @param typeOfCancerId - if specified (not null) update type of cancer of published virtual
    *     study
    * @param pmid - if specified (not null) update PubMed ID of published virtual study
+   * @param virtualStudyData - if specified (not null) create new virtual study with this data,
```
pieterlukasse, line 109:
```suggestion
   * @param virtualStudyData - if specified (not null) create new virtual study with this metadata,
```
this "data" is really just metadata, right?

forus: Yes, but metadata is loaded term. I'd not use it. Also , the class name is `VirtualStudyData`. So let's stick with data.

Issue #11616, opened 2025-07-14
Get all samples of a patient in a study endpoint returns 500 error for invalid patient id


gblaih:
Get all samples of a patient in a study endpoint should return 404 error for invalid patient id

/api/column-store/studies/{studyId}/patients/{patientId}/samples

PR #11626, opened 2025-07-18
Fix get all samples of patient in a study endpoint returning 500 error for invalid patient id

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11616

### DIFF
--- a/src/main/java/org/cbioportal/legacy/web/error/GlobalExceptionHandler.java
+++ b/src/main/java/org/cbioportal/legacy/web/error/GlobalExceptionHandler.java
@@ -40,7 +40,7 @@ import org.springframework.web.bind.annotation.ExceptionHandler;
 // TODO
 // - consider extending extends ResponseEntityExceptionHandler
 // - check controllers for not catching exceptions themselves
-@ControllerAdvice("org.cbioportal.legacy.web")
+@ControllerAdvice({"org.cbioportal.legacy.web", "org.cbioportal.application.rest.vcolumnstore"})
 public class GlobalExceptionHandler {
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/cbioportal/legacy/web/error/GlobalExceptionHandler.java
```diff
@@ -40,7 +40,7 @@
 // TODO
 // - consider extending extends ResponseEntityExceptionHandler
 // - check controllers for not catching exceptions themselves
-@ControllerAdvice("org.cbioportal.legacy.web")
+@ControllerAdvice({"org.cbioportal.legacy.web", "org.cbioportal.application.rest.vcolumnstore"})
```
alisman, line 43: @haynescd seems like this just brings columnstore into line with legacy (re excewption handling).  seems reasonable?

dippindots: One concern is: this handler is currently located at `/legacy/web/error/GlobalExceptionHandler.java`. If we ever decide to clean up or remove legacy code, there's a risk we might accidentally delete it.
@alisman @haynescd — thoughts on whether we should move this somewhere safer?

alisman: I'll defer to @haynescd on this, but maybe duplicating is worse than leaving a dependency which should be easy to detect and fix if and when we finally get rid of legacy codebase?

dippindots (approved):
Thanks! Looks good to me!

PR #11628, opened 2025-07-21
Add samples to the treatments endpoint

### LINKED ISSUE #11618: New study view treatment services are missing sample collections
New treatment endpoints are missing these sample collections.  We removed them because we didn't realize they were actually being used in order to create comparison sessions.  We need to reintroduce those sample collections unf.

Image

onursumer:
Fix #11618

`/treatments/sample-counts/fetch` endpoint now returns list of samples when `DETAILED` projection is enabled.

`/treatments/patient-counts/fetch` endpoint is still missing samples, but the comparison feature seems working fine.

### DIFF
--- a/src/main/resources/mappers/clickhouse/treatment/ClickhouseTreatmentMapper.xml
+++ b/src/main/resources/mappers/clickhouse/treatment/ClickhouseTreatmentMapper.xml
@@ -66,49 +66,74 @@
             )
         </where>
     </select>
-
+    
     <resultMap id="SampleTreatmentCount" type="org.cbioportal.legacy.model.SampleTreatment">
         <constructor>
             <arg column="treatment" javaType="String"/>
             <arg column="preSampleCount" javaType="_int"/>
             <arg column="postSampleCount" javaType="_int"/>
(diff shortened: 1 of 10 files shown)


src/main/resources/mappers/clickhouse/treatment/ClickhouseTreatmentMapper.xml
```diff
@@ -66,49 +66,74 @@
...
         </where>
         GROUP BY patient_unique_id, ced.value, cancer_study_identifier
-        ),
-        <!-- Nested sub query to grab minimum treatment event. When there are multiple treatments that are of the same
-        type we only care about the first treatment.
-         -->
-        treatments AS (
+    </sql>
+    
+    <sql id="treatments">
```
alisman, line 94: @onursumer doesn't this just select all treatment rows?

onursumer: I think so. This was an existing sql, I just moved this (and also `sample_acquisition_events`) into a separate SQL to make `getSampleTreatmentCounts` easier to read.

alisman (approved):

Issue #11633, opened 2025-07-23
BadSQLGrammerException detected by DataDog


alisman:
[Datadog report](https://app.datadoghq.com/error-tracking?query=&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22f5343134-6799-11f0-a959-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1753204763640&to_ts=1753291163640&live=true)

```
Request processing failed: org.springframework.jdbc.BadSqlGrammarException: 
### Error querying database.  Cause: java.sql.BatchUpdateException: Code: 43. DB::Exception: Illegal type (Tuple()) of 3 argument of function and: In scope WITH study_attribute_levels AS (SELECT DISTINCT cancer_study_identifier FROM clinical_data_derived WHERE (attribute_name = 'MUTATION_COUNT') AND (type = 'sample') AND (cancer_study_identifier IN ('brca_mbcproject_2022'))) SELECT cdd.sample_unique_id FROM clinical_data_derived AS cdd INNER JOIN study_attribute_levels AS sal ON cdd.cancer_study_identifier = sal.cancer_study_identifier WHERE (cdd.attribute_name = 'MUTATION_COUNT') AND (cdd.type = 'sample') AND tuple(). (ILLEGAL_TYPE_OF_ARGUMENT) (version 25.4.1.37460 (official build))
```

There is a a SQL error here.  Looks like the AND tuple() which chatGPT interprets and AND (( )).  i..e looksl ike the condition that produces that AND clause is fault.

PR #11637, opened 2025-07-24
Add special case for null start end values for numerical study view filters

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11633

Add special case where if null start and end values are passed in a numerical filter on study view, don't apply a filter. Prevents sql error where empty tuple () occurs.

### DIFF
--- a/src/main/java/org/cbioportal/domain/studyview/StudyViewFilterContext.java
+++ b/src/main/java/org/cbioportal/domain/studyview/StudyViewFilterContext.java
@@ -55,8 +55,7 @@ public record StudyViewFilterContext(
     var firstValue = clinicalDataFilter.getValues().getFirst();
     var lastValue = clinicalDataFilter.getValues().getLast();
     // there is a case where a categorical filter will have a set of null filter values first
-    // check if the first set of filter values is null, and if so, use the last set of filter values
-    // instead
+    // check if the first set of filter values is null. if so, use the last set of filter values
     var filterValue =
         firstValue.getStart() == null
                 && firstValue.getEnd() == null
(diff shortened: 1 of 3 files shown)


haynescd:
I don't think I understand the issue? Can you add an integration test here that validates this edge case has been fixed

haynescd (changes requested):
Need tests

src/main/resources/mappers/clickhouse/studyview/ClickhouseStudyViewFilterMapper.xml
```diff
@@ -280,6 +280,12 @@
                     AND 
                     ${attribute_value} ILIKE #{dataFilterValue.value}
                 </if>
+                <!--
+                    Special case: In case null start, end, and value values go through, don't apply filter.
+                -->
+                <if test="dataFilterValue.start == null and dataFilterValue.end == null">
```
onursumer, line 286: The remaining `if test` conditions seem to cover this case so we shouldn't be applying any filter. I'm surprised we still somehow apply filters when everything is null.

src/main/java/org/cbioportal/domain/studyview/StudyViewFilterContext.java
```diff
@@ -52,7 +52,16 @@ public String[] filteredSampleIdentifiers() {
   }
 
   public boolean isCategoricalClinicalDataFilter(ClinicalDataFilter clinicalDataFilter) {
-    var filterValue = clinicalDataFilter.getValues().getFirst();
+    var firstValue = clinicalDataFilter.getValues().getFirst();
+    var lastValue = clinicalDataFilter.getValues().getLast();
+    // there is a case where a categorical filter will have a set of null filter values first
```
onursumer, line 57: Ideally we should filter out invalid filters before passing them to this function, probably in the controller or service layer, so that we don't have to add more logic here.

src/main/resources/mappers/clickhouse/studyview/ClickhouseStudyViewFilterMapper.xml
gblaih: In the case where everything is null, none of the if conditions will pass so no filter should be applied. However, this `applyNumericalDataFilter` sql block is constructed like `((dataFilterValue1), (dataFilterValue2), ...)` and if everything is null in the original code, we will get an empty tuple which causes the sql errors we were encountering. So I added this condition so that it would act as if no filter is passed while also preventing the empty tuple sql error.

PR #11650, opened 2025-08-01
Add mutation domain classes and use cases

i-am-leslie:
This draft pull request outlines the changes I’ve made so far and my current thought process. I’m aiming to structure the mutation/fetch endpoint functionality following Clean Architecture principles.

 Summary of Changes
 1) Mutation and Gene Records
Created Mutation as a record class to represent mutation data.
Gene is a separate record and is used as an attribute within Mutation, based on how the frontend receives mutation objects.

2) usecase Package
This folder contains classes with single responsibilities, each focusing on a specific use case related to mutations.
All use cases rely on a ValidateMolecularProfileExistsUseCase class to validate the molecularProfileId.
Example: FetchAllMutationsInProfileUseCase demonstrates the intended structure and logic flow for other use cases.

3) Validation Folder
Contains the validation logic.
Currently includes a single class to validate the molecularProfileId, which is injected into the use cases to decouple responsibilities.

4) repository Package (Interface Only)
Not yet implemented. But the plans for it are as follows 
Will define the contract for data access logic, acting as an abstraction layer for the actual repository implementation.

 Next Steps / Plan
Implement the actual repository layer and define relevant methods in the MutationRepository interface.
Complete all use cases and integrate them with the repository.
Ensure proper test coverage and validate expected data flow from the controller through to the use cases.

 Reviewer Notes
Please feel free to review the overall structure and approach. Feedback on the direction, naming, or anything architectural is welcome, especially in the use case logic and dependency setup


src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnMutationController.java
```diff
@@ -0,0 +1,165 @@
...
+        @Min(PagingConstants.MIN_PAGE_NUMBER)
+        @RequestParam(defaultValue = PagingConstants.DEFAULT_PAGE_NUMBER)
+        Integer pageNumber,
+        @Parameter(description = "Name of the property that the result list is sorted by")
+        @RequestParam(required = false)
+        MutationSortBy sortBy,
+        @Parameter(description = "Direction of the sort") @RequestParam(defaultValue = "ASC")
+        Direction direction) {
+
+        if (projection == Projection.META) {
```
fuzhaoyuan: See if you can encapsulate the whole META processing logic into the use case so here it'll just be `if (META) {return usecase.execute()}`.

src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnMutationController.java
```diff
@@ -0,0 +1,165 @@
...
+                        projection.name(),
+                        pageSize,
+                        pageNumber,
+                        sortBy == null ? null : sortBy.getOriginalValue(),
+                        direction.name());
+            }
+            return ResponseEntity.ok(mutations);
+        }
+    }
+    private void extractMolecularProfileAndSampleIds(
```
fuzhaoyuan: If you try to encapsulate usecases if might not be feasible to keep this private method here. Do we put it in a Util class? Or it's simple enough to just do it in each use case? It's up to you but we all want to take the Clean Arch transit as an opportunity to clean things (controller here) up so the logic stays where it should be.

src/main/java/org/cbioportal/domain/mutation/usecase/GetMutationDataUseCases.java
```diff
@@ -0,0 +1,19 @@
...
+@Service
+@Profile("clickhouse")
+/**
+ * A record representing a collection of use cases related to Mutation data operations. This record
+ * encapsulates instances of various use case classes, providing a centralized way to access and utilize 
+ * the use cases 
+ * @param  fetchAllMetaMutationsInProfileUseCase
+ * @param fetchAllMutationsInProfileUseCase
+ */
+public record GetMutationDataUseCases(
```
fuzhaoyuan: The naming should be just `Mutation` instead of `MutationData` since that's the object name. Also applies to other mapper namings.

src/main/java/org/cbioportal/domain/mutation/usecase/GetMutationDataUseCases.java
```diff
@@ -0,0 +1,19 @@
...
+    FetchAllMetaMutationsInProfileUseCase fetchAllMetaMutationsInProfileUseCase,
```
fuzhaoyuan: The use case names will need to be changed but we can wait until later

src/main/java/org/cbioportal/domain/mutation/Mutation.java
```diff
@@ -0,0 +1,36 @@
...
+    String variantType,
+    String refseqMrnaId,
+    Integer proteinPosStart,
+    Integer proteinPosEnd,
+    String keyword,
+    String chr,
+    String variantAllele, 
+    String uniqueSampleKey, 
+    String uniquePatientKey
+    ) implements Serializable {
```
fuzhaoyuan: I noticed we have a Clean Arch`CancerStudyMetadata` that transitioned from legacy `CancerStudy` that looks like this. Do you think you can reference on how that endpoint `GetCancerStudyMetadataUseCase` is implemented?

src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnMutationController.java
i-am-leslie: Ok, I did notice three parameters that are a little confusing in the controller 

1)  Collection<String> involvedCancerStudies - it doesn't have any usage, so I plan on removing it.
2) MutationMultipleStudyFilter interceptedMutationMultipleStudyFilter - This parameter gets intercepted when a request is sent, and I believe it gets changed based on a user's permission.
3)  MutationMultipleStudyFilter mutationMultipleStudyFilter - The same as the second variable(2) but not intercepted. Not sure what to do with the second and third variable for now, using the third one for testing

fuzhaoyuan: For parameters you can leave them as they are since it's part of the formatting for [the swagger API definitions](cbioportal.org/api)

This controller looks better now. The naming should be `ColumnStoreMutationController`

src/main/java/org/cbioportal/shared/MutationSearchCriteria.java
i-am-leslie, line 1: I am not 100% sure this is where I should keep this file, but I used this record to hold all search criteria for the endpoint I am currently transitioning to clean architecture. I would like some advice on this?

src/main/java/org/cbioportal/infrastructure/repository/clickhouse/mutation/ClickhouseMutationMapper.java
i-am-leslie, line 1: The snpOnly parameter is currently not used. In the legacy implementation, it was always hardcoded to false, and in the MyBatis SQL, it was only checked for true, meaning the condition was never satisfied. Should we clean it up or leave it?

src/main/resources/mappers/clickhouse/mutation/ClickhouseMutationDataMapper.xml
```diff
@@ -0,0 +1,197 @@
...
+                <foreach item="item" collection="entrezGeneIds" open="(" separator="," close=")">
+                    #{item}
+                </foreach>
+            </if>
+        </where>
+    </sql>
+
+<!--    For now only works for projection SUMMARY not DETAILED PROJECTION -->
+    <sql id="from">
+        FROM mutation
```
alisman: hi @i-am-leslie, these joins are indeed going to slow things down.  in clickhouse we are trying to use the *_derived tables as much as possible because these contain much of the data denormalized (pre JOINED).    I would look at genomic_event_derived and filter for mutation.  It already should have fields from mutation_event and gene.   You can also join on sample_derived, which will already have necessary patient data.  If you find that there are missing fields, we should add them to derived tables and i'm happy to help with that.

i-am-leslie: Thanks for your review on my PR, right now, it is difficult to support the SUMMARY and DETAILED projections for mutations. Using genomic_event_derived alone doesn’t give me all the fields required, so I end up falling back to heavy multi-joins (mutation, mutation_event, driver_annotation, etc.) which is what we are trying to avoid. 

### **Details i get from genomic_event_derieved needed for Detailed and Summary Projection** 
1) molecularProfileId
2) sampleId
3) patientId
4) entrezGeneId
5) studyId
       **From mutation**
6) mutationStatus
       **From mutation_event**
7) variantType
8) mutaionType
9)variant_type
       **From Gene** (Only for Detailed projection )
10) hugoGeneSymbol  
   **From alteration_driver_annotation**
11) driverFilter
12) driverTiersFilter

### **Details needed for data completeness of mutation that are not present in genomic_event_derived**

**Needed from mutation:**
*  center
* validation_status
* tumor_alt_count
* tumor_ref_count
* normal_alt_count
* normal_ref_count
* amino_acid_change
* annotation_json
**Needed from mutation_event:**
* chr
* start_position
* end_position
* reference_allele
* tumor_seq_allele
* protein_change
* ncbi_build
* refseq_mrna_id
* protein_pos_start
* protein_pos_end
* keyword
**Needed from alteration_driver_annotation:**
* driver_filter_annotation
* driver_tiers_filter_annotation
**Needed from gene: (Detailed projection)**
* gene.type 
**Needed from allele_specific_copy_number:(Detailed projection)**
* ascnIntegerCopyNumber
* ascnMethod
* ccfExpectedCopiesUpper
* ccfExpectedCopies
* clonal
* minorCopyNumber
* expectedAltCopies
* totalCopyNumber

Would it make sense to extend genomic_event_derived to include these fields (at least the ones needed for SUMMARY / DETAILED projections)?

Alternatively, should we consider a new mutation_derived table that flattens mutation + mutation_event + driver_annotation?

Happy to adjust my query approach, but right now it seems impossible to avoid multiple joins without these fields in the derived tables. Also, details about each projection can be found here https://docs.google.com/document/d/1DYG0xdx_GM8pJq-GgNJz2Q0SZNOi4M4VNqNCAn-5Qzg/edit?tab=t.0

inodb:
Closing this because we decided to go a different route - thank you! See: https://github.com/cBioPortal/cbioportal/pull/11750

PR #11659, opened 2025-08-15
Update StudyRepository cache to static cache and institute keyword condition

alisman:
We need to keep studies in the static cache, especially now that we have lowered ttl


src/main/java/org/cbioportal/legacy/persistence/StudyRepository.java
```diff
@@ -9,8 +9,8 @@
 public interface StudyRepository {
 
   @Cacheable(
-      cacheResolver = "generalRepositoryCacheResolver",
-      condition = "@cacheEnabledConfig.getEnabled()")
+      cacheResolver = "staticRepositoryCacheOneResolver",
+      condition = "@cacheEnabledConfig.getEnabled() && #keyword == null")
```
dippindots, line 13: @alisman This looks good, we should use static cache for all studies and all tags.

dippindots (approved):
LGTM

PR #11663, opened 2025-08-18
Adjust caching behavior (move some resources to from general to static and vice versa)

alisman:
Many endpoints are cached in the "general" cache repository, which is now cleared on an hourly basis.  For data resources which are static or for which permutations are bounded (e.g. gene panel ids) we should not be afraid to put the them in the static cache to improve performance

### DIFF
--- a/src/main/java/org/cbioportal/legacy/persistence/CancerTypeRepository.java
+++ b/src/main/java/org/cbioportal/legacy/persistence/CancerTypeRepository.java
@@ -8,18 +8,18 @@ import org.springframework.cache.annotation.Cacheable;
 public interface CancerTypeRepository {
 
   @Cacheable(
-      cacheResolver = "generalRepositoryCacheResolver",
+      cacheResolver = "staticRepositoryCacheOneResolver",
       condition = "@cacheEnabledConfig.getEnabled()")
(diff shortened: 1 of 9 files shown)


src/main/java/org/cbioportal/legacy/persistence/GeneRepository.java
```diff
@@ -59,22 +59,22 @@ List<Gene> getAllGenes(
...
-      cacheResolver = "staticRepositoryCacheOneResolver",
+      cacheResolver = "generalRepositoryCacheResolver",
       condition = "@cacheEnabledConfig.getEnabled()")
   List<String> getAliasesOfGeneByEntrezGeneId(Integer entrezGeneId);
 
   @Cacheable(
-      cacheResolver = "staticRepositoryCacheOneResolver",
+      cacheResolver = "generalRepositoryCacheResolver",
       condition = "@cacheEnabledConfig.getEnabled()")
   List<String> getAliasesOfGeneByHugoGeneSymbol(String hugoGeneSymbol);
```
dippindots: We don’t want to switch this to the `general` cache, right? I think we should keep gene-related data in Redis for as long as possible.

PR #11669, opened 2025-08-19
Improve molecular data query

onursumer:


src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -30,24 +30,50 @@
...
-            <include refid="org.cbioportal.legacy.persistence.mybatis.GeneMapper.select">
-                <property name="prefix" value="GENE."/>
-            </include>
-        </if>
-        FROM genetic_alteration
-        INNER JOIN genetic_profile ON genetic_alteration.genetic_profile_id = genetic_profile.genetic_profile_id
-        INNER JOIN gene ON genetic_alteration.genetic_entity_id = gene.genetic_entity_id
+        WITH dd AS (<include refid="geneticAlterationSubQueryWithProjection" />)
+        SELECT 
+            ga.values,
```
alisman: @onursumer maybe the fact that values is first (instead of entrezGeneId) affects ordering or response and could account for difference?

src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -3,15 +3,40 @@
...
+                    SELECT gp.genetic_profile_id
+                    FROM genetic_profile gp
+                    WHERE gp.stable_id IN 
+                        <foreach item="item" collection="molecularProfileIds" open="(" separator="," close=")">#{item}</foreach>
+                )   
             </if>
             <if test="entrezGeneIds != null and !entrezGeneIds.isEmpty()">
-                AND gene.entrez_gene_id IN
-                    <foreach item="item" collection="entrezGeneIds" open="(" separator="," close=")">#{item}</foreach>
+                AND genetic_alteration.genetic_entity_id IN (<include refid="selectMultipleEntrezGeneIds"/>)
```
alisman: It's confusing as hell that 'selectMultipleEntrezGeneIds' actually returns genetic_entity_ids.   What if we renamed this 'selectMultipleGeneticEntityIdsByEntrezGeneIds'.  I know that's a mouthful but otherwise, people will forever confused the entit ids with entrezids?

alisman: Or maybe we also include "AND genetic_alteration.genetic_entity_id =" inside of the <include>?  cuz now you can just called 'gene condition' or whatever and hide the fact that we're acually using entity ids?

onursumer: I think it is okay to include `genetic_alteration.genetic_entity_id =` as well and rename to something more general purpose

src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
onursumer: This was happening because we were incorrectly using genetic entity id instead of entrez gene id. This has been fixed by modifying the SQL.

src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
onursumer: Included these in the SQL and renamed to be less confusing

onursumer:
Currently core tests are failing, but they are also failing on the base branch which is probably better to address in a separate PR.

PR #11676, opened 2025-08-22
FIX:cacheENabledConfig restructured

Ajiboye-01:
FIX:cacheENabledConfig restructured

I changed
LOG.info("Cache is enabled: " + this.enabled); to LOG.info("Cache is enabled: {}", this.enabled) because of efficiency, the first one will execute whether there is something to log or not, but the second one won't log anything unless the log is necessary


alisman:
@Ajiboye-01 we actually do want to know if it's false.  also the comment isn't proper english.   and the change to immutable is fine but belongs in another PR.

PR #11685, opened 2025-08-28
Add Spring Boot Actuator monitoring

Floris-Hyve:
## Spring Boot Actuator 
PR introduces Actuator dependency which allows monitoring various operational information regarding the app.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,14 @@
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -939,4 +947,4 @@
     </profile>
   </profiles>
 
-</project>
+</project>
```
Ritabrata1080, line 950: Please add extra line at EOF to satisfy github

alisman:
@inodb do we want this?    if so, please reopen and merge

PR #11730, opened 2025-10-02
order genomic_event_derived table by study id first

### LINKED ISSUE #11718: Improve ordering of genomic_event_derived table
Genomic event derived table needs new ordering because we are frequently scannign the entire table.   It should not be ordered initially by variant type.  it hsould be ordered by study id

onursumer:
Fix #11718


alisman:
@onursumer i've actually come to the conclusion that it should be ordered by genomic_profile_id, which includes study id. This has a huge impact on existing study view queries.

PR #11738, opened 2025-10-08
Add statements to clickhouse.sql that fix primary keys of underlying MySQL tables

alisman:


src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -485,6 +492,49 @@ FROM
         JOIN cancer_study cs ON cs.cancer_study_id = subquery.cancer_study_id
         JOIN sample_derived sd ON sd.internal_id = subquery.sample_id;
 
+
+--Adds primary key to the sample_cna_event table for Clickhouse-only
```
alisman, line 609: @sheridancbio what do you think about fixing the primary keys of the "slung" tables as part of derivation.  we create new table with appropriate keys, copy data into it and then switch the table out.  
for the genetic_alteration table of public portal, this takes 5 minutes.   

the alternative is that, as you suggested, we put table definition in sling process, which i don't love.

alisman:
@sheridancbio agreed would better not do this.   for the public database the table is 65GB, this takes only five mins.

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -373,9 +375,10 @@ SELECT
     ifNull(ce.stop_date, 0) AS stop_date,
     ce.event_type AS event_type,
     cs.cancer_study_identifier
-FROM clinical_event ce
-         LEFT JOIN clinical_event_data ced ON ce.clinical_event_id = ced.clinical_event_id
-         INNER JOIN patient p ON ce.patient_id = p.internal_id
+
+FROM clinical_event_data ced
+    RIGHT JOIN clinical_event ce ON ced.clinical_event_id = ce.clinical_event_id
```
alisman, line 380: @sheridancbio @jamesqo  please review this change here

PR #11740, opened 2025-10-09
Optimize discrete cna query for new Clickhouse primary key

alisman:
We are adding a primary key (profile id) to the table and need to filter on it in order to limit row scan.  

This PR also swaps the order of JOIN for optimization reason.

### DIFF
--- a/src/main/resources/org/cbioportal/legacy/persistence/mybatis/DiscreteCopyNumberMapper.xml
+++ b/src/main/resources/org/cbioportal/legacy/persistence/mybatis/DiscreteCopyNumberMapper.xml
@@ -24,8 +24,8 @@
     </sql>
 
     <sql id="from">
-        FROM cna_event
-        INNER JOIN sample_cna_event ON cna_event.cna_event_id = sample_cna_event.cna_event_id
+        FROM sample_cna_event sample_cna_event
+        INNER JOIN cna_event ON cna_event.cna_event_id = sample_cna_event.cna_event_id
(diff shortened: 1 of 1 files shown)


src/main/resources/org/cbioportal/legacy/persistence/mybatis/DiscreteCopyNumberMapper.xml
```diff
@@ -24,8 +24,8 @@
     </sql>
 
     <sql id="from">
-        FROM cna_event
-        INNER JOIN sample_cna_event ON cna_event.cna_event_id = sample_cna_event.cna_event_id
+        FROM sample_cna_event_optimized sample_cna_event
```
onursumer: Include a CREATE statement for `sample_cna_event_optimized` table in `clickhouse.sql`?

src/main/resources/org/cbioportal/legacy/persistence/mybatis/DiscreteCopyNumberMapper.xml
```diff
@@ -229,14 +237,15 @@
     </select>
 
     <select id="getSampleCountByGeneAndAlterationAndSampleIds" resultType="org.cbioportal.legacy.model.CopyNumberCountByGene">
+        --ZIGGY
```
onursumer: we should probably remove this comment :)

src/main/resources/org/cbioportal/legacy/persistence/mybatis/DiscreteCopyNumberMapper.xml
alisman: good catch. fixing this

PR #11749, opened 2025-10-11
Add Spring AI for OQL Assistant

FlorisVleugels:
# Introduce Spring AI dependency (for OQL Assistant)

[Spring AI](https://docs.spring.io/spring-ai/reference/index.html) provides abstractions to integrate with various [models](https://docs.spring.io/spring-ai/reference/api/chatmodel.html)

## Key Changes
- Added Spring AI dependency
- Added a context resource so model can be given instructions
- New endpoint `/api/assistant` 

## Todo
Spring AI has a built in [evaluator](https://docs.spring.io/spring-ai/reference/api/testing.html) for determining whether the the AI model has produced a hallucinated response or not (using another ai call to validate the other). 
At the moment the context resource includes the instruction to add `OQL: TRUE` or `OQL:FALSE` to the generated text (see example below), which is used by the frontend to decide whether to render the message or an error. 

Could be better to let the evaluator decide and change response accordingly, or have some feedback loop for n iterations until valid oql is returned or max is reached?

## Example

```bash
curl -X POST \
  http://localhost:8080/api/assistant \
  -H "Content-Type: application/json" \
  -d '"Find all somatic driver mutations in TP53 and BRCA1, except missense mutations, and also get EGFR and MYC with mRNA expression greater than 2 SD. Include KRAS mutations only at positions 12 and 13, but exclude nonsense mutations."'
```
```json
{
  "aiResponse": "TP53: DRIVER_SOMATIC_MUT != MISSENSE BRCA1: DRIVER_SOMATIC_MUT != MISSENSE EGFR: EXP > 2 MYC: EXP > 2 KRAS: MUT = (12-13) != NONSENSE OQL: TRUE"
}
```

## Related PR
[frontend PR that adds the OQL Assistant UI](https://github.com/cBioPortal/cbioportal-frontend/pull/5281)

### DIFF
--- a/microservices/assistant/Dockerfile
+++ b/microservices/assistant/Dockerfile
@@ -1,4 +1,19 @@
-FROM openjdk:21-jdk-slim
-ARG JAR_FILE=target/gene-assistant-1.0.0-SNAPSHOT.jar
-COPY ${JAR_FILE} app.jar
-ENTRYPOINT ["java","-jar","/app.jar"]
+FROM maven:3-eclipse-temurin-17 as build
+
+WORKDIR /assistant
+
+COPY pom.xml .
+
+RUN mvn dependency:go-offline --fail-never
+
+COPY src ./src
+
+RUN mvn install package -DskipTests
+
+FROM eclipse-temurin:17-jre-jammy
+
+WORKDIR /assistant
+
+COPY --from=build /assistant/target/assistant.jar .
+
+ENTRYPOINT ["java","-jar","assistant.jar"]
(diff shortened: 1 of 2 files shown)


src/main/java/org/cbioportal/legacy/web/GeneAssistantController.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@Validated
+@RestController()
+@ConditionalOnProperty(name = "spring.ai.enabled", havingValue = "true")
```
inodb: wondering if we need to set CORS restrictions somewhere for these endpoints? So only users of the website can use m

alisman:
@inodb @zainasir this is defunct, yes?

PR #11751, opened 2025-10-13
In the molecular data truncation SQL, add case for zero values

alisman:
Add special case handling for zero values in the molecular data value
truncation logic to prevent them from being converted incorrectly.

### DIFF
--- a/src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
+++ b/src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
@@ -76,10 +76,12 @@
                     -- Leave non-numeric values as is, otherwise they will be converted to zero.
                     isNull(accurateCastOrNull(x, 'Float64')), 
                     x,
-                    -- Special Case 1b: Zero values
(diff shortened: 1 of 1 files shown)


src/main/resources/org/cbioportal/legacy/persistence/mybatis/MolecularDataMapper.xml
```diff
@@ -76,6 +76,9 @@
                     -- Leave non-numeric values as is, otherwise they will be converted to zero.
                     isNull(accurateCastOrNull(x, 'Float64')), 
                     x,
+                    -- Special Case 1b: Zero values
+                        toFloat64OrZero(x) = 0,
+                        '0',
                     -- Special Case 2: Very small numbers
```
onursumer: Let's fix the indentation, and also add more comments?

```suggestion
                    -- Special Case 2: Zero values
                    -- We need to handle zero values separately otherwise they will break the very small number logic
                    -- due to the fact that log10(0) is equal to -Infinity
                    toFloat64OrZero(x) = 0,
                    '0',
                    -- Special Case 3: Very small numbers
```

PR #11753, opened 2025-10-14
Fix sample list error (clickhouse-only)

dippindots:
@dippindots @haynescd if i understand correctly, CH doesn't support this this function and it actually clear what it was doing in the first place in the absnce of group by?  @haynescd this was causing it auth to fail.

### DIFF
--- a/src/main/resources/org/cbioportal/legacy/persistence/mybatis/SampleListMapper.xml
+++ b/src/main/resources/org/cbioportal/legacy/persistence/mybatis/SampleListMapper.xml
@@ -4,36 +4,34 @@
 <mapper namespace="org.cbioportal.legacy.persistence.mybatis.SampleListMapper">
 
     <sql id="select">
-        sample_list.list_id AS "${prefix}listId",
-        sample_list.stable_id AS "${prefix}stableId"
+        sample_list.list_id AS "listId",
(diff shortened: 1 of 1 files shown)


src/main/resources/org/cbioportal/legacy/persistence/mybatis/SampleListMapper.xml
```diff
@@ -4,22 +4,28 @@
...
+        sample_list.list_id AS "${prefix}listId",
+        sample_list.stable_id AS "${prefix}stableId"
         <if test="projection == 'SUMMARY' || projection == 'DETAILED'">
             ,
-            ANY_VALUE(sample_list.cancer_study_id) AS "${prefix}cancerStudyId",
-            ANY_VALUE(cancer_study.cancer_study_identifier) AS "${prefix}cancerStudyIdentifier",
-            ANY_VALUE(sample_list.category) AS "${prefix}category",
-            ANY_VALUE(sample_list.name) AS "${prefix}name",
-            ANY_VALUE(sample_list.description) AS "${prefix}description"
+            sample_list.cancer_study_id AS "${prefix}cancerStudyId",
```
haynescd: remove the prefix as well... it is confusing and is set to "" anyways

PR #11754, opened 2025-10-15
Fix failing core tests

### LINKED ISSUE #11772: [BUG] Legacy tests are failing on clickhouse only branch
Upgrade legacy tests to use clickhouse test containers and make sure all legacy tests are passing on Github actions

onursumer:
Fix #11772 

- Use clickhouse container for legacy mybatis tests
- Fix failing service tests by properly mocking replaced methods

### DIFF
--- a/src/main/resources/org/cbioportal/legacy/persistence/mybatis/ClinicalEventMapper.xml
+++ b/src/main/resources/org/cbioportal/legacy/persistence/mybatis/ClinicalEventMapper.xml
@@ -172,4 +172,110 @@
         GROUP BY clinical_event.event_type, patient.stable_id
     </select>
 
+    <select id="getTimelineEvents" resultType="org.cbioportal.legacy.model.ClinicalEvent">
+        SELECT
+            ANY_VALUE(patient.stable_id) AS patientId,
+            cancer_study.cancer_study_identifier AS studyId,
+            MIN(clinical_event.start_date) AS startDate,
+            GREATEST(MAX(clinical_event.start_date), COALESCE(MAX(clinical_event.stop_date),0)) AS stopDate
+        FROM clinical_event
+            INNER JOIN patient ON clinical_event.patient_id = patient.internal_id
+            INNER JOIN clinical_event_data ON clinical_event.clinical_event_id = clinical_event_data.clinical_event_id
+            INNER JOIN cancer_study ON patient.cancer_study_id = cancer_study.cancer_study_id
+        <where>
+            <if test="patientIds == null and studyIds != null">
+                cancer_study.cancer_study_identifier IN
+                <if test="studyIds.isEmpty()">(NULL)</if>
+                <if test="!studyIds.isEmpty()"><foreach item="item" collection="studyIds" open="(" separator="," close=")">#{item}</foreach></if>
+            </if>
+            <if test="patientIds != null">
+                <if test="@java.util.Arrays@stream(patientIds.toArray()).distinct().count() == 1">
+                    <bind name="patientArray" value="@org.cbioportal.legacy.persistence.mybatis.util.SqlUtils@listToArray(patientIds)" />
+                    cancer_study.cancer_study_identifier = #{studyIds[0]} AND
+                    patient.stable_id IN (
+                        #{patientArray, typeHandler=org.apache.ibatis.type.ArrayTypeHandler}
+                    )
+                </if>
+                <if test="@java.util.Arrays@stream(patientIds.toArray()).distinct().count() > 1">
+                    <bind name="patientUniqueKeys" value="@org.cbioportal.legacy.persistence.mybatis.util.SqlUtils@combineStudyAndPatientIds(studyIds, patientIds)" />
+                    CONCAT(cancer_study.cancer_study_identifier, ':', patient.stable_id) IN (
+                        #{patientUniqueKeys, typeHandler=org.apache.ibatis.type.ArrayTypeHandler}
+                    )
+                </if>
+            </if>
+            <if test="!clinicalEvents.isEmpty()">
+                AND
+                <foreach item="element" collection="clinicalEvents" open="(" separator="OR " close=")">
+                    <if test="element.attributes == null || element.attributes.isEmpty()">
+                        clinical_event.event_type = #{element.eventType}
+                    </if>
+                    <if test="element.attributes != null and !element.attributes.isEmpty()">
+                        (CONCAT(clinical_event.event_type, '_', clinical_event_data.key, '_', clinical_event_data.value)) IN
+                        <foreach item="attribute" collection="element.attributes" open="(" separator="," close=")">
+                            CONCAT(#{element.eventType}, '_', #{attribute.key}, '_', #{attribute.value})
+                        </foreach>
+                    </if>
+                </foreach>
+            </if>
+        </where>
+        GROUP BY
+        clinical_event.patient_id, cancer_study.cancer_study_identifier;
+    </select>
+
+    <resultMap id="clinicalEventsResultMap" type="org.cbioportal.legacy.model.ClinicalEvent">
+        <result property="eventType" column="eventType"/>
+        <collection property="attributes" ofType="org.cbioportal.legacy.model.ClinicalEventData">
+            <result property="key" column="key"/>
+            <result property="value" column="value"/>
+        </collection>
+    </resultMap>
+    
+    <select id="getClinicalEventsMeta" resultMap="clinicalEventsResultMap">
+        SELECT
+            LOWER(clinical_event.event_type) AS eventType,
+            LOWER(clinical_event_data.key) AS "key",
+            LOWER(clinical_event_data.value) AS value
+        FROM clinical_event
+            INNER JOIN patient ON clinical_event.patient_id = patient.internal_id
+            INNER JOIN clinical_event_data ON clinical_event.clinical_event_id = clinical_event_data.clinical_event_id
+            INNER JOIN cancer_study ON patient.cancer_study_id = cancer_study.cancer_study_id
+        <where>
+            <if test="patientIds == null and studyIds != null">
+                cancer_study.cancer_study_identifier IN
+                <if test="studyIds.isEmpty()">(NULL)</if>
+                <if test="!studyIds.isEmpty()"><foreach item="item" collection="studyIds" open="(" separator="," close=")">#{item}</foreach></if>
+            </if>
+            <if test="patientIds != null">
+                <if test="@java.util.Arrays@stream(patientIds.toArray()).distinct().count() == 1">
+                    <bind name="patientArray" value="@org.cbioportal.legacy.persistence.mybatis.util.SqlUtils@listToArray(patientIds)" />
+                    cancer_study.cancer_study_identifier = #{studyIds[0]} AND
+                    patient.stable_id IN (
+                    #{patientArray, typeHandler=org.apache.ibatis.type.ArrayTypeHandler}
+                    )
+                </if>
+                <if test="@java.util.Arrays@stream(patientIds.toArray()).distinct().count() > 1">
+                    <bind name="patientUniqueKeys" value="@org.cbioportal.legacy.persistence.mybatis.util.SqlUtils@combineStudyAndPatientIds(studyIds, patientIds)" />
+                    CONCAT(cancer_study.cancer_study_identifier, ':', patient.stable_id) IN (
+                    #{patientUniqueKeys, typeHandler=org.apache.ibatis.type.ArrayTypeHandler}
+                    )
+                </if>
+            </if>
+            <if test="!clinicalEvents.isEmpty()">
+                AND
+                <foreach item="element" collection="clinicalEvents" open="(" separator="OR " close=")">
+                    <if test="element.attributes == null || element.attributes.isEmpty()">
+                        clinical_event.event_type = #{element.eventType}
+                    </if>
+                    <if test="element.attributes != null and !element.attributes.isEmpty()">
+                        (CONCAT(clinical_event.event_type, '_', clinical_event_data.key)) IN
+                        <foreach item="attribute" collection="element.attributes" open="(" separator="," close=")">
+                            CONCAT(#{element.eventType}, '_', #{attribute.key})
+                        </foreach>
+                    </if>
+                </foreach>
+            </if>
+        </where>
+        GROUP BY clinical_event.event_type , clinical_event_data.key, clinical_event_data.value;
+    </select>
+
 </mapper>


onursumer:
Opened a temporary PR #11777 to trigger the github actions on the modified code

src/main/resources/org/cbioportal/legacy/persistence/mybatis/PatientMapper.xml
```diff
@@ -10,18 +10,18 @@
...
+            cancer_study.type_of_cancer_id       AS "${prefix}cancerStudy.typeOfCancerId",
+            cancer_study.name                    AS "${prefix}cancerStudy.name",
+            cancer_study.description             AS "${prefix}cancerStudy.description",
+            cancer_study.public                  AS "${prefix}cancerStudy.publicStudy",
+            cancer_study.pmid                    AS "${prefix}cancerStudy.pmid",
+            cancer_study.citation                AS "${prefix}cancerStudy.citation",
+            cancer_study.groups                  AS "${prefix}cancerStudy.groups",
+            cancer_study.status                  AS "${prefix}cancerStudy.status",
+            cancer_study.import_date             AS "${prefix}cancerStudy.importDate",
+            reference_genome.name                AS "${prefix}cancerStudy.referenceGenome"
```
dippindots, line 25: I noticed similar implementations in other mappers — we always use an empty string ("") for the prefix. It seems like adding the prefix property isn’t really necessary in this case.

onursumer: Some tests were failing because of this missing prefix. We should probably clean up these prefixes in a follow up PR.

src/main/java/org/cbioportal/legacy/persistence/mybatis/ClinicalEventMapper.java
```diff
@@ -34,7 +34,7 @@ List<ClinicalEvent> getSamplesOfPatientsPerEventType(
       List<String> studyIds, List<String> sampleIds);
 
   List<ClinicalEvent> getPatientsDistinctClinicalEventInStudies(
-      List<String> studyIds, List<String> patientIds);
+      List<String> studyIds, List<String> patientIds, List<ClinicalEvent> clinicalEvents);
```
alisman, line 37: @onursumer this seems strange.  was it broken all along or part of Clickhouse-only?

src/main/resources/org/cbioportal/legacy/persistence/mybatis/AlterationCountsMapper.xml
```diff
@@ -180,9 +180,9 @@
         SELECT
             gene1.entrez_gene_id AS gene1EntrezGeneId,
             <!-- TODO: check this-->
-            gene1.hugo_gene_symbol AS gene1HugoGeneSymbol,
+            ANY_VALUE(gene1.hugo_gene_symbol) AS gene1HugoGeneSymbol,
```
alisman, line 183: i could swear that these aren't valid in clickhouse, no?

alisman: i googled and found that ANY_VALUE is not supported in CH.  That makes me wonder how it passes here.  We are using CH now as our testing db or some fake db? @haynescd

src/main/java/org/cbioportal/legacy/persistence/mybatis/ClinicalEventMapper.java
onursumer: Probably a change related to clickhouse only branch, otherwise tests on the master branch would (hopefully) catch these

src/main/resources/org/cbioportal/legacy/persistence/mybatis/AlterationCountsMapper.xml
onursumer: We already use `ANY_VALUE` in the clickhouse only branch, so I just followed that example. Maybe they recently added support for this?

alisman: ok, confirmed that these do run on CH in existing queries.  weird that it sayd that function doesn't exist.

src/main/resources/org/cbioportal/legacy/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -348,7 +334,7 @@
             <when test="sortIsPatientAttr">
                 SELECT sample.internal_id sampleInternalId, sample.stable_id sampleId, patient.stable_id patientId, clinical_patient.attr_id sortAttrId,
                 <choose>
-                    <when test="sortAttrIsNumber">CONVERT(clinical_patient.attr_value, DECIMAL) sortAttrValue
+                    <when test="sortAttrIsNumber">toFloat64(clinical_patient.attr_value) sortAttrValue
```
alisman: @onursumer @haynescd lets check to make sure that this is legit coversion (decimal to float64)

alisman: we should probably change this to toDecimal

src/main/resources/org/cbioportal/legacy/persistence/mybatis/ClinicalDataMapper.xml
```diff
@@ -363,7 +349,7 @@
                 SELECT sample.internal_id sampleInternalId, sample.stable_id sampleId, patient.stable_id patientId, clinical_sample.attr_id sortAttrId,
                 <choose>
                     <when test="sortAttrIsNumber">
-                        CONVERT(clinical_sample.attr_value, DECIMAL) sortAttrValue
+                        toFloat64(clinical_sample.attr_value) sortAttrValue
```
alisman: toDecimal

src/main/resources/org/cbioportal/legacy/persistence/mybatis/CopyNumberSegmentMapper.xml
```diff
@@ -35,19 +35,16 @@
...
         <include refid="where"/>
-        AND EXISTS (
-            SELECT * from copy_number_seg  INNER JOIN sample ON sample.internal_id = copy_number_seg.sample_id
-            <if test="chromosome != null and !chromosome.isEmpty()">
-                <where>
-                    copy_number_seg.chr = #{chromosome}
-                </where>
-            </if>
-        )
+        <if test="chromosome != null and !chromosome.isEmpty()">
```
alisman, line 45: this needs to be reviewed

alisman: the join may have resulted in multiple rows (suggested by the DISTINCT) and so this needs review.

src/main/resources/org/cbioportal/legacy/persistence/mybatis/CopyNumberSegmentMapper.xml
```diff
@@ -84,16 +81,14 @@
         SELECT
         <include refid="select"/>
         <include refid="from"/>
-        WHERE cancer_study.cancer_study_identifier = #{studyId}
+        INNER JOIN sample_list_list ON copy_number_seg.sample_id = sample_list_list.sample_id
```
alisman, line 84: same here

src/main/resources/org/cbioportal/legacy/persistence/mybatis/PatientMapper.xml
```diff
@@ -10,18 +10,18 @@
...
-            cancer_study.name                   AS "cancerStudy.name",
-            cancer_study.description             AS "cancerStudy.description",
-            cancer_study.public                  AS "cancerStudy.publicStudy",
-            cancer_study.pmid                    AS "cancerStudy.pmid",
-            cancer_study.citation                AS "cancerStudy.citation",
-            cancer_study.groups                  AS "cancerStudy.groups",
-            cancer_study.status                  AS "cancerStudy.status",
-            cancer_study.import_date             AS "cancerStudy.importDate",
-            reference_genome.name                AS "cancerStudy.referenceGenome"
+            cancer_study.cancer_study_id         AS "${prefix}cancerStudy.cancerStudyId",
```
alisman, line 14: lets put a comment here to address this abstraction.  remove prefix and also quotes

src/main/resources/org/cbioportal/legacy/persistence/mybatis/SampleMapper.xml
```diff
@@ -107,16 +107,12 @@
 
     <select id="getMetaSamplesBySampleListIds" resultType="org.cbioportal.legacy.model.meta.BaseMeta">
         SELECT
-        COUNT(*) AS "totalCount"
+        COUNT(DISTINCT sample.internal_id) AS "totalCount"
```
alisman, line 110: lets review this because the DISTINCT suggests that the fundamental selection logic has changed.

src/test/java/org/cbioportal/legacy/persistence/mybatis/ClinicalEventMyBatisRepositoryTest.java
```diff
@@ -266,6 +275,7 @@ public void getPatientsDistinctClinicalEventInStudies() {
   }
 
   @Test
+  @Ignore("getTimelineEvents already removed from the legacy mapper")
```
alisman: @haynescd this seems probably like an oversight copying stuff.  the mappers don't exist.  lets confirm

src/test/java/org/cbioportal/legacy/persistence/mybatis/MolecularProfileMyBatisRepositoryTest.java
```diff
@@ -91,9 +101,9 @@ public void getAllMolecularProfilesDetailedProjection() throws Exception {
     Assert.assertEquals("brca", cancerStudy.getTypeOfCancerId());
     Assert.assertEquals("Breast Invasive Carcinoma (TCGA, Nature 2012)", cancerStudy.getName());
     Assert.assertEquals(
-        "<a href=\\\"http://cancergenome.nih.gov/\\\">The Cancer Genome Atlas (TCGA)</a> Breast"
-            + " Invasive Carcinoma project. 825 cases.<br><i>Nature 2012.</i> <a href=\\\"http://tcga-data.nci."
-            + "nih.gov/tcga/\\\">Raw data via the TCGA Data Portal</a>.",
+        "<a href=\"http://cancergenome.nih.gov/\">The Cancer Genome Atlas (TCGA)</a> Breast"
```
alisman, line 104: see if we can find where this actually gets rendered and test whether the escaping change is problematic

src/test/java/org/cbioportal/legacy/persistence/mybatis/MutationMyBatisRepositoryTest.java
```diff
@@ -278,7 +286,7 @@ public void getMutationsInMolecularProfileBySampleListIdDetailedProjection() thr
     Assert.assertEquals("genome.wustl.edu", mutation.getCenter());
     Assert.assertEquals((Long) 41244748L, mutation.getEndPosition());
     Assert.assertEquals("BRCA1 truncating", mutation.getKeyword());
-    Assert.assertEquals("Germline", mutation.getMutationStatus());
+    Assert.assertEquals("GERMLINE", mutation.getMutationStatus());
```
alisman: lets figure out why this changed.

src/test/java/org/cbioportal/legacy/persistence/mybatis/StudyMyBatisRepositoryTest.java
```diff
@@ -62,37 +73,40 @@ public void getAllStudiesSummaryProjection() throws Exception {
     Assert.assertEquals((Integer) 0, cancerStudy.getStatus());
     Assert.assertEquals(
         simpleDateFormat.parse("2011-12-18 13:17:17+00:00"), cancerStudy.getImportDate());
-    Assert.assertEquals((Integer) 14, cancerStudy.getAllSampleCount());
+    // TODO enable after fixing "1 AS allSampleCount"
```
alisman: remove TODO. this mapper is defunct.

src/test/java/org/cbioportal/legacy/persistence/mybatis/MutationMyBatisRepositoryTest.java
onursumer: actually the enum definition is only in one of the test db schema https://github.com/cBioPortal/cbioportal/blob/39caae5f3249815a43541d04a5ecba54d938dff4/src/test/resources/clickhouse_cgds.sql#L300

I think for now we should revert the test data change (which I introduced with this PR) and keep the test assertion unchanged

src/main/resources/org/cbioportal/legacy/persistence/mybatis/ClinicalDataMapper.xml
alisman: @onursumer i looked into this and i think we should change this to toDecimal32OrNull() or toDecimal64OrNull().  Not sure whether we need nullable.  i guess just depends if the column is nullable.

src/test/java/org/cbioportal/legacy/persistence/mybatis/ClinicalEventMyBatisRepositoryTest.java
onursumer: restored the accidentally deleted mapper SQL

src/main/resources/org/cbioportal/legacy/persistence/mybatis/ClinicalDataMapper.xml
onursumer: for some reason changing to `toDecimal32OrNull` causes tests to fail, but `toDecimal64OrNull` works fine. I'll go ahead with `toDecimal64OrNull`.

src/main/resources/org/cbioportal/legacy/persistence/mybatis/SampleMapper.xml
onursumer: Here is the chatGPT response. Note the "multiple list entries per sample" case at the very end.

image

src/main/resources/org/cbioportal/legacy/persistence/mybatis/CopyNumberSegmentMapper.xml
onursumer: this assertion fails if we remove DISTINCT

https://github.com/cBioPortal/cbioportal/blob/283a2d7f1fda209db5e4494d84a76fff66d098e3/src/test/java/org/cbioportal/legacy/persistence/mybatis/CopyNumberSegmentMyBatisRepositoryTest.java#L130

```
java.lang.AssertionError: 
Expected :2
Actual   :3
```

onursumer:
Superseded by #11784

PR #11798, opened 2025-11-05
Optimize getClinicalDataCounts query

### LINKED ISSUE #11794: Optimize clinical data filtering mapper
The current clinical data counts query duplicates the same sub query multiple times in a single SQL operation.  We want to execute it only once and reference it.  This could be done with a temp table.  We can't use a CTE because Clickhouse doesn't materialize them, it inlines them, which means they get executed redundantly just as we had left them in place.

Problem query:

[bad_clin.txt](https://github.com/user-attachments/files/23314370/bad_clin.txt)

Optimization using temp table:

[optimized_coadread_query.sql](https://github.com/user-attachments/files/23314794/optimized_coadread_query.sql)

onursumer:
Fix #11794

- Update Clickhouse Clinical Data Mapper to use temporary session scoped tables
- Move certain tests into the repository layer due to the introduction of new session scoped tables


alisman:
@onursumer this looks good.  i'm going to take the opportunity to create some new api tests around this endpoint before merging.

PR #11806, opened 2025-11-11
Add support for Meta Projection to return Study Counts 

haynescd:
Create new endpoint /studies/meta, also update /studies to handle projection Meta, update projection Summary to add allSampleCount, simulate Pagination at controller level

### DIFF
--- a/src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreStudyController.java
+++ b/src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreStudyController.java
@@ -2,13 +2,15 @@ package org.cbioportal.application.rest.vcolumnstore;
 
 import io.swagger.v3.oas.annotations.Hidden;
 import io.swagger.v3.oas.annotations.Parameter;
-import java.util.Collections;
+import jakarta.validation.constraints.Max;
+import jakarta.validation.constraints.Min;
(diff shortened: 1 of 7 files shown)


src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreStudyController.java
```diff
@@ -85,14 +92,88 @@ public ResponseEntity<List<CancerStudyMetadataDTO>> getAllStudies(
...
+
+    // Pagination should be handled at the DB layer, but currently our query is not
+    // setup to handle this with authorization
+    if (pageSize != null) {
+      studies = studies.stream().limit(pageSize).toList();
+    }
+    var headers = new HttpHeaders();
+    if (projection == ProjectionType.META) {
+      headers.add(HeaderKeyConstants.TOTAL_COUNT, String.valueOf(studies.size()));
+      headers.add(TOTAL_COUNT_HEADER, String.valueOf(studies.size()));
```
onursumer, line 125: why do we need two headers for the exact same count? backward compatibility?

haynescd: Yea, one is a standard and the other is legacy. I thought might as well switch to a standard format for displaying counts

PR #11817, opened 2025-11-18
Merge master to release branch

alisman:


sheridancbio:
I've looked over the changes here (bringing in the changes which are now in master into the rc-7.0-clickhouse-only branch)
This looks valid to me. schema numbers have been properly updated in clickhouse.sql and pom.xml, and needed changes are present. There may be some unneeded / non-standard whitespace, but I think this is usable and I will be using this changeset for the next msk portal import cycle.

PR #11822, opened 2025-11-21
Fix Clinical Attribute NA Filtering

### LINKED ISSUE #11808: NA filtering doesn't work in combined study setting
https://www.cbioportal.org/study/summary?id=msk_chord_2024%2Ccellline_nci60

In the combined study view above, find the ICD-O Histology Description chart.   There are 67 NA samples, all of which belong to the cellline_nci60 study (which has 67 samples).   Filter on NA

Expectation: we filter to 67 samples

Actual: we filter to zero samples

Probably has something to do with the fact that that one of the studies does not have the attribute at all.  So it seems that we don't have the ability to filter on that.  This might have been a known issue.

fuzhaoyuan:
Brought over to https://github.com/cBioPortal/cbioportal/pull/11901
-----
Fix #11808 

# Fix Clinical Attribute NA Filtering

## Summary
This PR fixes a issue in NA filtering for clinical attributes introduced by PR #11603 . The fix ensures studies without an attribute definition are properly treated as NA while preserving the attribute level conflict resolution.

## Problem
When PR #11603 added logic to handle attribute level conflicts (sample vs patient level), it inadvertently broke NA filtering. Samples from studies that completely lack an attribute definition were incorrectly excluded from NA filter results.

### Core NA Filtering Logic
This PR implements a comprehensive three-tier conditional logic for handling clinical attribute filtering:

#### **1. IF study has NO attribute definition:**
→ **All samples treated as NA** via `study_without_attribute` CTE and `UNION ALL`

```sql
study_without_attribute AS (
    SELECT DISTINCT cancer_study_identifier FROM sample_derived
    WHERE cancer_study_identifier NOT IN (SELECT cancer_study_identifier FROM study_with_attribute)
)
```

#### **2. ELSE IF study HAS the attribute at different level (sample vs patient):**
→ Use data from the corresponding level only  
→ Sample queries skip patient-level attributes, vice versa  
→ Prevents attribute level conflicts (preserves #11603 fix)

```sql
study_attribute_levels AS (
    SELECT DISTINCT cancer_study_identifier FROM clinical_data_derived
    WHERE attribute_name = ... AND type='${type}'  -- Filters by correct level
)
```

#### **3. ELSE IF study HAS the attribute at matching level:**
→ LEFT JOIN to retrieve data  
→ IF attribute_value IS NULL or empty → treated as NA  
→ ELSE filter by actual value

### Modified Files

**ClickhouseStudyViewFilterMapper.xml**
- Added `studyAttributeFilterCTEs` and `samplesFromStudiesWithoutAttributeAsNA` SQL fragments
- Refactored `numericalClinicalDataCountFilter` to use shared fragments and handle studies without attributes
- Refactored `categoricalClinicalDataCountFilter` with same improvements

**ClickhouseSampleMapperTest.java**
- Updated test expectations to reflect correct NA handling (restored pre-#11603 behavior)
- `acc_tcga` study samples (4 samples with no AGE/DEAD attributes) now correctly included in NA filters
- Test changes:
  - NA AGE filter: 7 → 11 samples (7 explicit NA + 4 from study without attribute)
  - NA AGE + UNKNOWN: 8 → 12 samples
  - NA DEAD filter: 17 → 21 samples (17 explicit NA + 4 from study without attribute)

### DIFF
--- a/src/test/java/org/cbioportal/infrastructure/repository/clickhouse/sample/ClickhouseSampleMapperTest.java
+++ b/src/test/java/org/cbioportal/infrastructure/repository/clickhouse/sample/ClickhouseSampleMapperTest.java
@@ -184,6 +184,21 @@ public class ClickhouseSampleMapperTest {
             StudyViewFilterFactory.make(
                 studyViewFilter, List.of(), studyViewFilter.getStudyIds(), null));
     assertEquals(17, filteredSamples9.size());
+
+    // NA subtype filter: attribute exists at sample-level in acc_tcga but patient-level in
+    // study_genie_pub (cross-study attribute level conflict scenario)
+    // acc_tcga (sample-level subtype): sample tcga-a1-a0sb-01 has subtype='' (empty) -> 1 NA
+    // study_genie_pub (patient-level subtype):
+    //   - patient 316 has subtype='NA', patient 318 has subtype='' (empty) -> 2 NA samples
+    //   - patients 301-312, 319-324 have no subtype data -> 21 NA samples
+    // Total: 1 + 2 + 21 = 24 samples
+    studyViewFilter.setClinicalDataFilters(
+        List.of(newClinicalDataFilter("subtype", List.of(newDataFilterValue(null, null, "NA")))));
+    var filteredSamples10 =
+        mapper.getFilteredSamples(
+            StudyViewFilterFactory.make(
+                studyViewFilter, List.of(), studyViewFilter.getStudyIds(), null));
+    assertEquals(24, filteredSamples10.size());
   }
 
   @Test


alisman (commented):
## Code Review

**Purpose:** Fix NA filtering for clinical attributes so that samples from studies that completely lack an attribute definition are correctly treated as NA. This was a regression from PR #11603.

The core approach is sound: introduce `study_with_attribute` and `study_without_attribute` CTEs to identify studies that lack an attribute entirely, and UNION ALL their samples into the NA results. The fix is applied to both `numericalClinicalDataCountFilter` and `categoricalClinicalDataCountFilter`.

---

### 1. Potential edge case: attribute exists at opposite level only (Medium)

The `study_with_attribute` CTE does **not** filter by level:
```sql
study_with_attribute AS (
    SELECT DISTINCT cs.cancer_study_identifier
    FROM clinical_attribute_meta cam
    INNER JOIN cancer_study cs ON cam.cancer_study_id = cs.cancer_study_id
    WHERE cam.attr_id = #{clinicalDataFilter.attributeId}
)
```

But `study_attribute_levels` **does** filter by level:
```sql
study_attribute_levels AS (
    ...
    WHERE cam.attr_id = #{clinicalDataFilter.attributeId}
        AND cam.patient_attribute = IF('${type}' = 'patient', 1, 0)
)
```

Consider this scenario:
- **Study X** defines attribute `SUBTYPE` at **patient** level only
- Query is for **sample** level `SUBTYPE` with NA filter

Result: Study X is in `study_with_attribute` (has the attribute at some level), so it is NOT in `study_without_attribute`. But it is also NOT in `study_attribute_levels` (wrong level). Study X samples fall into a gap: they are neither treated as NA nor included in non-NA results.

This may be the **intended** behavior from PR #11603 (to prevent level-conflict inflation), but it means "study has attribute at wrong level" is treated differently from "study has no attribute at all." If a study only defines `SUBTYPE` at patient level, its samples arguably should show as NA for sample-level `SUBTYPE` queries. Worth confirming this edge case is acceptable, or whether `study_with_attribute` should also filter by `patient_attribute` to match `study_attribute_levels`.

### 2. `numericalClinicalDataCountFilter` is missing the `studyAttributeFilterCTEs` include (High)

Looking at the PR diff for `numericalClinicalDataCountFilter`, it adds the `studyAttributeFilterCTEs` include to the `study_attribute_levels` WITH clause:

```xml
WITH <include refid="...studyAttributeFilterCTEs"/>,
study_attribute_levels AS (...)
```

And the NA branch gets a UNION ALL with `samplesFromStudiesWithoutAttributeAsNA`.

However, in the **current master** version of `numericalClinicalDataCountFilter` (which I can see at lines 419-491), the `studyAttributeFilterCTEs` fragment and the `samplesFromStudiesWithoutAttributeAsNA` fragment are **not present**. The PR diff shows these being added. This part looks correct in the diff.

**But**: the NA branch in `numericalClinicalDataCountFilter` wraps the UNION in parentheses:
```xml
<if test="userSelectsNA">
    (
    SELECT DISTINCT sd.${unique_id} ...
    UNION ALL
    <include refid="...samplesFromStudiesWithoutAttributeAsNA"/>
    )
</if>
```

When `userSelectsNA AND userSelectsNumericalValue` are both true, the generated SQL becomes:
```sql
( SELECT ... UNION ALL SELECT ... )
UNION ALL
SELECT ... -- numerical values
```

The outer parentheses on the first UNION ALL ensure correct precedence. This looks correct.

### 3. `categoricalClinicalDataCountFilter` wrapping may produce invalid SQL in edge cases (Medium)

In the categorical filter, the `samplesFromStudiesWithoutAttributeAsNA` UNION ALL is placed **after** the WHERE clause's foreach:

```xml
WHERE (( ... ) OR ( ... ))
<if test="userSelectsNA">
    UNION ALL
    <include refid="...samplesFromStudiesWithoutAttributeAsNA"/>
</if>
)
```

This means the UNION ALL is inside the outer parenthesis of the entire categorical filter subquery. The generated SQL would be:
```sql
(
    WITH study_with_attribute AS (...), study_without_attribute AS (...), study_attribute_levels AS (...)
    SELECT sd.sample_unique_id FROM sample_derived sd ...
    WHERE ((attribute_value IS NULL OR ... = 'NA') OR (attribute_value ILIKE 'SomeValue'))
    UNION ALL
    SELECT DISTINCT sd.sample_unique_id FROM sample_derived sd INNER JOIN study_without_attribute ...
)
```

This is syntactically correct. However, note that the UNION ALL happens **unconditionally for the study_without_attribute samples** regardless of which filter values are selected. If the filter has both NA and non-NA categorical values (e.g., `['NA', 'Luminal A']`), the study_without_attribute samples will be included in the results. This is correct since they are NA.

But if `userSelectsNA` is false and only non-NA values are selected, the UNION ALL is skipped, which is also correct. This looks fine.

### 4. `study_attribute_levels` query changed from `clinical_data_derived` to `clinical_attribute_meta` (Low, positive change)

The PR changes the `study_attribute_levels` CTE from querying `clinical_data_derived` (actual clinical data rows) to `clinical_attribute_meta` (attribute definitions). This is a more correct approach because:

- **Before**: A study would only appear in `study_attribute_levels` if it had **data rows** in `clinical_data_derived` for that attribute at that level. If the attribute was defined but no samples had data, the study would be missed.
- **After**: A study appears if the **attribute is defined** in `clinical_attribute_meta`, regardless of whether data rows exist.

This is a good improvement. However, note that `clinical_attribute_meta.patient_attribute` is `Nullable(Int32)` in the ClickHouse schema. If `patient_attribute` is NULL for any rows, the `IF('${type}' = 'patient', 1, 0)` comparison would not match NULLs. This is likely a non-issue in practice since `patient_attribute` should always be 0 or 1.

### 5. Legacy mapper not updated (Medium)

The legacy `StudyViewFilterMapper.xml` has its own copies of `numericalClinicalDataCountFilter` and `categoricalClinicalDataCountFilter` (lines 380-477). These are **not updated** in this PR and still have the pre-#11603 behavior. This means:
- The legacy controller path still has the original NA filtering bug
- The two code paths continue to diverge

If the legacy path is being phased out, this may be acceptable. But if both paths are active, users will get inconsistent results depending on which endpoint they hit.

### 6. No test for attribute-level conflict + NA filter combination (Low)

The tests cover the straightforward cases: study with no attribute at all. But there is no test for the scenario described in issue #1 (attribute exists at opposite level). Adding a test with an attribute that exists at patient-level in one study and sample-level in another, then filtering by NA, would validate the interaction between this fix and the #11603 fix.

### 7. `samplesFromStudiesWithoutAttributeAsNA` fragment does not apply study view filters (Medium)

The `samplesFromStudiesWithoutAttributeAsNA` fragment is:
```xml
SELECT DISTINCT sd.${unique_id}
FROM sample_derived sd
INNER JOIN study_without_attribute swa ON sd.cancer_study_identifier = swa.cancer_study_identifier
```

This joins against `study_without_attribute` (which is scoped to `customDataFilterCancerStudies` if present). However, it does **not** apply the main study view filter (`applyStudyViewFilterUsingSampleId`). If the study view has other active filters (e.g., genomic data filters, other clinical data filters), the samples from studies without the attribute will be included **unfiltered**.

This is likely acceptable because the clinical data filter results are INTERSECTed together in `applyClinicalDataCountFilter` (line 373), so other filters will narrow the results. But within this one filter's scope, it means ALL samples from studies without the attribute are returned as NA, not just samples that pass the study view filter. In most cases the INTERSECT handles this correctly, but it could cause performance issues if `study_without_attribute` contains many studies/samples.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | Medium | Attribute at opposite level falls into a gap (not NA, not queried) |
| 2 | - | Numerical filter UNION wrapping is correct (verified) |
| 3 | - | Categorical filter UNION placement is correct (verified) |
| 4 | Low | Positive: switching to `clinical_attribute_meta` is more correct |
| 5 | Medium | Legacy mapper not updated -- divergent code paths |
| 6 | Low | No test for attribute-level conflict + NA combination |
| 7 | Medium | `samplesFromStudiesWithoutAttributeAsNA` does not apply study view filters |

Overall this is a well-structured fix with clear intent and good test coverage for the primary use case. The main items to consider are the opposite-level edge case (#1) and whether the lack of study view filter application on the NA samples fragment (#7) could cause unexpected results.

fuzhaoyuan:
I like how we are using AI to work on PRs and doing code reviews at each other, just like Pokemon. I'll look into those

PR #11849, opened 2025-12-09
Fix core tests with latest docker 

alisman:


alisman:
@zainasir @onursumer after hours of debugging, this change seems to be what fixes the running of the core tests on my local machine.  my local machine btw was freshly installed a few days ago.  seems the latest Docker is not totally compatible with our testing infrastructure.  This change fixes the problem.  But it seems to break the tests in CI.   Not sure exactly what to do.  Would be nice to know if it was broken for others as well or maybe this is just some weird thing about my setup.

onursumer:
@alisman Did you try moving this file to your own home directory instead of keeping it in `src/test/resources`? I have the same file located at `~/.docker-java.properties` on my Ubuntu machine.

alisman:
@onursumer got it.  i will try that.  thanks

alisman:
Thanks @onursumer  that seems to work.  Do you have the same content as me as well?  I mean,  do you have it pinned to a version?

onursumer:
yes, it's pinned to `api.version=1.44` but it would probably work with a newer version as well

PR #11840, opened 2025-12-05
Fixes #11583 — Optimize molecular data multi-profile fetch for ClickHouse (reduce N+1 queries)

immortal71:
Fixes #11583

This PR addresses the performance bottleneck when using ClickHouse in multi-profile molecular data fetches. Instead of repeated per-gene queries (N+1), the ClickHouse repository now fetches per-sample rows from the `genetic_alteration_derived` table and aggregates them into the legacy `values` CSV format expected by the service layer. The service now requests all entrez gene IDs in a single call.

Key changes:
- New ClickHouse mapper & XML to query `genetic_alteration_derived`.
- ClickHouse repository that aggregates per-sample rows into `GeneMolecularAlteration`.
- Service change to call repository once with the full set of requested entrez IDs.
- Unit tests for clickhouse aggregation.

Notes & next steps: Add `entrez_gene_id` to the derived table to avoid a join to `gene` during the ClickHouse query for better perf.

PR #11856, opened 2025-12-11
Add molecular data counts endpoint for multiple profiles - Fixes #11761

### LINKED ISSUE #11761: molecular data meta endpoint needs to handle multiple molecular profiles
We need the ability to retrieve counts for multiple molecular profiles (data) with a single call/query.  It shoudl replace this foreach here:

https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/resultsView/ResultsViewPageStore.ts#L1821

Note that currently we use the Meta endpoint, which returns counts in the http header.   

The problem is that we have multiple hits to the database which forces Clickhouse to do the joins multiple times.

immortal71:
## Summary
Implements a new endpoint to retrieve molecular data counts for multiple molecular profiles in a single database call, eliminating the need for N HTTP requests when querying N profiles.

## Problem
The existing `/api/molecular-profiles/{molecularProfileId}/molecular-data/fetch?projection=META` endpoint returns counts in HTTP headers but requires one call per profile. This causes:
- N HTTP requests for N profiles
- N database queries with joins (slow on ClickHouse)
- Poor performance when dealing with multiple profiles

## Solution
Created new `/api/molecular-data/counts` POST endpoint that:
- Accepts multiple molecular profile IDs in a single request
- Returns JSON array with per-profile counts
- Uses single optimized database query leveraging PR #11840's optimization
- Returns structured response: `[{molecularProfileId: "...", count: 123}, ...]`

## Changes
- **New Model**: `MolecularDataCountItem` - represents count per molecular profile
- **Service Layer**: Added `fetchMolecularDataCountsInMultipleMolecularProfiles` method
- **Controller**: New `POST /api/molecular-data/counts` endpoint
- **Tests**: Unit tests for both service and controller layers
- **Documentation**: Implementation plan document included

## Testing
- Added unit tests for service layer (MolecularDataServiceImplTest)
- Added controller test (MolecularDataControllerTest)
- Tests verify correct grouping and counting per profile
- No compilation errors

## Related Issues
Fixes #11761

### DIFF
--- a/src/main/java/org/cbioportal/legacy/service/impl/MolecularDataServiceImpl.java
+++ b/src/main/java/org/cbioportal/legacy/service/impl/MolecularDataServiceImpl.java
@@ -181,14 +181,18 @@ public class MolecularDataServiceImpl implements MolecularDataService {
     List<Integer> allInternalSampleIds = new ArrayList<>();
 
     for (String molecularProfileId : distinctMolecularProfileIds) {
-      List<Integer> internalSampleIds =
-          Arrays.stream(
-                  commaSeparatedSampleIdsOfMolecularProfilesMap
-                      .get(molecularProfileId)
-                      .getSplitSampleIds())
-              .mapToInt(Integer::parseInt)
-              .boxed()
-              .collect(Collectors.toList());
+      List<Integer> internalSampleIds = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


immortal71:
@alisman @dippindots  hey I have make some changes according to copilot suggestion !! ,can you review this and let me know if there is any other thing i should do or if it good for merger !!

PR #11857, opened 2025-12-11
Add Mocha/TypeScript E2E test framework and replace old integration (api-tests) with it

alisman:
Replace backend api tests with a new javascript implementation which is intended to run against the public portal.  These tests are intended to protect against regression in backend pull requests.  They use the public data set in order to more easily reproduce edge cases which are costly to synthesize in test data (e.g combined studies)


.circleci/config.yml
```diff
@@ -539,12 +539,12 @@ workflows:
             requires:
               - build_frontend
               - build_web_and_data_image
-        - run_api_tests:
-            context:
-              - api-tests
-            requires:
-              - build_web_image
-              - pull_cbioportal_frontend_codebase
+#        - run_api_tests:
```
alisman, line 542: @zainasir i want to commandeer this test process and run the new mocha job against the server it stands up

PR #11861, opened 2025-12-12
Stable sort structural variant genes

### LINKED ISSUE #11859: Stabilize sort order for structuralvariant-genes service
The sorting of items in the structural variants chart of study view is non deterministic due to Clickhouse and cause e2e instability.  We need to put a default sorting so that it's consistent.

https://output.circle-artifacts.com/output/job/a5773943-5499-4f0c-b172-311494cc8a7a/artifacts/0/imageCompare.html

onursumer:
Fix #11859

### DIFF
--- a/src/main/resources/mappers/clickhouse/alteration/ClickhouseAlterationMapper.xml
+++ b/src/main/resources/mappers/clickhouse/alteration/ClickhouseAlterationMapper.xml
@@ -29,6 +29,7 @@
             </if>
         </where>
         GROUP BY entrez_gene_id, hugo_gene_symbol
+        ORDER BY totalCount DESC, hugo_gene_symbol ASC
     </select>
 
     <!-- /cna-genes/fetch (returns CopyNumberCountByGene) -->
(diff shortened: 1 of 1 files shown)


src/main/resources/mappers/clickhouse/alteration/ClickhouseAlterationMapper.xml
```diff
@@ -75,6 +75,7 @@
             />
         </where>
         GROUP BY entrez_gene_id, hugo_gene_symbol
+        ORDER BY totalCount DESC, hugo_gene_symbol ASC
```
onursumer, line 80: We should probably apply the same ORDER BY criteria to other queries in this mapper

src/main/java/org/cbioportal/domain/alteration/usecase/GetAlterationCountByGeneUseCase.java
```diff
@@ -1,6 +1,6 @@
 package org.cbioportal.domain.alteration.usecase;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
```
onursumer, line 3: `LinkedHashMap` makes sure to keep the original order of the input list as opposed to unstable order of `HashMap`

PR #11874, opened 2025-12-17
Disable integration tests and skip clickhouse integration tests

dippindots:
- Disable integration-test from github action workflow
- Skip clickhouse integration tests
- Skip e2e tests in sonarcloud


forus:
@dippindots Is there a ticket for enabling integration-test from github action workflow?

PR #11892, opened 2025-12-24
Use VirtualStudy service from public VS controller

forus:
### Summary

Submitting **dynamic virtual studies** without explicitly defined sample IDs previously resulted in a `NullPointerException`. This behavior was incorrect, as dynamic studies are designed to derive their sample IDs **dynamically from filters**, not require them upfront.

### Root Cause

The issue was traced to the **public Virtual Study controller**, which was invoking the session service directly. That service assumed precomputed sample IDs and therefore failed for dynamic studies.

### Fix

- Updated the public VS controller to use the ** VirtualStudy service**.
  - When handling public dynamic virtual studies, the controller now reuses the same computation from the virtualStudyService that is already implemented for regular virtual studies to **compute sample IDs dynamically** based on the provided filters.
- This aligns the implementation with the intended design and prevents null-related failures.
- Sample counts are now calculated dynamically as well, ensuring they stay in sync with the underlying filters.

### Result

- Public dynamic virtual studies can now be submitted safely without specifying sample IDs, as expected.
- Landing page sample counts are no longer stale and reflect the current filter-based computation.

*Please* merge the fix to the v6 version as well https://github.com/cBioPortal/cbioportal/pull/11891

### DIFF
--- a/src/main/java/org/cbioportal/legacy/service/VirtualStudyService.java
+++ b/src/main/java/org/cbioportal/legacy/service/VirtualStudyService.java
@@ -2,27 +2,42 @@ package org.cbioportal.legacy.service;
 
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.cbioportal.legacy.service.exception.CancerTypeNotFoundException;
+import org.cbioportal.legacy.service.exception.DuplicateVirtualStudyException;
+import org.cbioportal.legacy.service.exception.StudyNotFoundException;
 import org.cbioportal.legacy.service.util.SessionServiceRequestHandler;
 import org.cbioportal.legacy.web.parameter.SampleIdentifier;
 import org.cbioportal.legacy.web.parameter.VirtualStudy;
 import org.cbioportal.legacy.web.parameter.VirtualStudyData;
 import org.cbioportal.legacy.web.parameter.VirtualStudySamples;
 import org.cbioportal.legacy.web.util.StudyViewFilterApplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 @Service
 public class VirtualStudyService {
+  private static final Logger LOG = LoggerFactory.getLogger(VirtualStudyService.class);
+
+  public static final String ALL_USERS = "*";
   private final SessionServiceRequestHandler sessionServiceRequestHandler;
   private final StudyViewFilterApplier studyViewFilterApplier;
+  private final CancerTypeService cancerTypeService;
+  private final StudyService studyService;
 
   public VirtualStudyService(
       SessionServiceRequestHandler sessionServiceRequestHandler,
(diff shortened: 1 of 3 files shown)


pieterlukasse:
Thanks @forus . 
Can you please clarify the issue description? I'm not sure I understand this part:
> The controller now computes sample IDs dynamically based on the provided filters when handling public dynamic virtual studies.

Does this mean that computation of  "sample IDs dynamically based on the provided filters" has not been occurring yet? Or is this an issue specifically with the **public virtual studies** controller? If it is just this controller, why would it now need compute the sample IDs dynamically instead of just reusing the same computation already implemented for the regular virtual studies? Or is that what you meant?

src/main/java/org/cbioportal/legacy/service/VirtualStudyService.java
```diff
@@ -107,4 +126,91 @@ private Map<String, Set<String>> groupSampleIdsByStudyId(
...
+   * Publishes virtual study optionally updating metadata fields
+   *
+   * @param id - id of public virtual study to publish
+   * @param typeOfCancerId - if specified (not null) update type of cancer of published virtual
+   *     study
+   * @param pmid - if specified (not null) update PubMed ID of published virtual study
+   * @param virtualStudyData - if specified (not null) create new virtual study with this data,
+   *     otherwise updates virtual study with the given id
+   */
+  public void publishVirtualStudy(
```
pieterlukasse, line 140: should we move this method here if it is only used for public virtual studies? Why not leave it in the `PublicVirtualStudiesController`?

forus:
@pieterlukasse Thanks for the review!

Find my answers below.

> Does this mean that computation of "sample IDs dynamically based on the provided filters" has not been occurring yet? Or is this an issue specifically with the public virtual studies controller?

Indeed, the computation has not been running for the public virtual studies controller only, which lists public virtual studies on the landing page. However, the calculation has been occurring for the session service endpoints, which are used everywhere else for virtual studies.

>  If it is just this controller, why would it now need compute the sample IDs dynamically instead of just reusing the same computation already implemented for the regular virtual studies? Or is that what you meant?

It reuses the same computation in the `virtualStudyService` that is already implemented for regular virtual studies. That is precisely the point.

src/main/java/org/cbioportal/legacy/service/VirtualStudyService.java
forus: @pieterlukasse You’re right, it’s currently only used by the public virtual studies controller. I still believe it belongs here, as it contains substantial logic and low-level interactions with sessionServiceRequestHandler. That kind of code does not belong in a controller.

pieterlukasse (commented):
Thanks @forus , I see now. Basically the code is adjusted to correctly use [getUserVirtualStudies](https://github.com/cBioPortal/cbioportal/blob/8816456a8a6b3f9137b3cc7329820db9d8c2d5a4/src/main/java/org/cbioportal/legacy/service/VirtualStudyService.java#L65) instead of just `sessionServiceRequestHandler.getVirtualStudiesAccessibleToUser(ALL_USERS)`

PR #11893, opened 2025-12-24
Make study export clickhouse only

forus:
Drop MySQL support

### DIFF
--- a/src/main/resources/mappers/export/CnaSegmentMapper.xml
+++ b/src/main/resources/mappers/export/CnaSegmentMapper.xml
@@ -4,6 +4,7 @@
 <mapper namespace="org.cbioportal.application.file.export.repositories.mybatis.CnaSegmentMapper">
     <select id="getCnaSegments" resultType="org.cbioportal.application.file.model.CnaSegment"
             resultSetType="FORWARD_ONLY">
+        <bind name="sampleIdsArray" value="sampleIds != null ? sampleIds.toArray(new String[sampleIds.size()]) : null" />
         SELECT
             cns.seg_id,
             s.stable_id AS sampleId,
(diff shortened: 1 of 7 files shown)


dippindots (approved):
@forus Thanks for fixing this! It looks good to me. I’d like to loop in @onursumer for a final review before we merge.

dippindots:
@forus I'm not very familiar with this feature, how can I test it?

src/main/resources/mappers/export/CnaSegmentMapper.xml
```diff
@@ -5,45 +5,43 @@
...
-        JOIN sample s ON s.INTERNAL_ID = cns.SAMPLE_ID
+        JOIN cancer_study cs ON cs.cancer_study_id = cns.cancer_study_id
+        JOIN sample s ON s.internal_id = cns.sample_id
         <if test="sampleIds != null">
             JOIN (
                 SELECT *
                 FROM (VALUES
                 <foreach item="sid" collection="sampleIds" separator=",">
-                    ROW(#{sid})
+                    (#{sid})
```
alisman: @forus just a heads up that in Clickhouse scenario, using foreach on very large lists (e.g. sample ids) can be problematic.  Yes, foreach is executed in java prior to submission to clickhouse, however, it has to do with the way prepared statements work differently in CL than MySQL.   If you search the mappers for ```apache.ibatis.type.ArrayTypeHandler``` you'll see an alternative style of array handling which is much more performant.

forus:
> @forus I'm not very familiar with this feature, how can I test it?

@dippindots To enable this feature, set `feature.study.export=true` in the application properties file. Once enabled, you can download a study by accessing the `/export/study/{STUDY_STABLE_ID}.zip` endpoint.

forus:
@alisman Thanks, Aaron! Could you please take another look?

alisman:
@forus just noticed that there's a new sonar issue.  it's the TODO comment.  this is annoying and maybe we should loosen the Sonar rule @haynescd since the nature of the TODO is to remind for later?  What do you guys think?

forus:
> @forus just noticed that there's a new sonar issue. it's the TODO comment. this is annoying and maybe we should loosen the Sonar rule @haynescd since the nature of the TODO is to remind for later? What do you guys think?

@alisman I can understand that it’s annoying.

I don’t have a solution for the TODO at the moment, but I do know it should be done. So strictly speaking, it’s technical debt that we should track in one way or another.

Maybe we can just review Sonar new issues (including TODOs) and take them simply into account? They won’t appear as new issues on other PRs.

alisman:
@forus ok, we figured out how to "accept" the sonar issue so that it doesn't show up as a new issue.

PR #11902, opened 2026-01-06
Update derived table version check in sanity test

dippindots:
For the derived_table_version check, we now only verify that the first line of clickhouse.sql contains the same version as specified in pom.xml's <derived_table.version> property.

Previously, the script checked the derived table version across `cgds.sql` and `migration.sql`, but this was unnecessary for the current ClickHouse-specific derived table schema. 

The script now extracts the version from the comment on line 1 of clickhouse.sql using a regex pattern and compares it directly with pom.xml.


dippindots:
We decided to merge @sheridancbio’s PR (https://github.com/cBioPortal/cbioportal/pull/11903), since it also updates the API response to reflect the correct database version, which makes more sense overall.

PR #11904, opened 2026-01-07
Gene-specific charts include sample ids

### LINKED ISSUE #11771: gene specific charts do not have a 'compare groups' option
If I add a gene-specific mutation chart, it does not have should have a 'Compare Groups' option in the menu:
Image

But a gene-specific CNA chart does have 'Compare Groups':
Image

I'm not sure why mutation charts are different, but can we please add 'Compare Groups' for mutation gene-specific charts?

i-am-leslie:
Fix https://github.com/cBioPortal/cbioportal/issues/11771
-  This change adds optional support for including sample IDs in the mutation data counts response. When enabled, the backend populates sample IDs required by the comparison page on the frontend, while preserving existing behaviour when the flag is disabled. This extension is aimed at reducing the complexity required by the frontend to group and create a comparison session for gene-specific charts.

## Checks
- [x] Yes, the change has been tested to check the presence of a list of strings when **includeSampleId** is **true** and no list when **includeSampleId** is **false**. Additionally, the use case layer was tested to ensure that when **includeSampleId** is true, only a single gene should be present in the list.
- [x] Yes, the commit log is comprehensive.
- [x] No, it is not adding logic based on one or more clinical attributes

**_Note_**
**includeSampleId** is used exclusively for gene-specific comparison sessions. Although the summary and detailed projection accept multiple genes and render one chart per gene, the comparison feature operates at the individual chart level. Each chart represents a single gene, and comparison groups (Mutated vs Not Mutated) are defined relative to that gene. Therefore, I enforced exactly one gene when sample IDs are requested to maintain clear and consistent grouping behaviour.

### DIFF
--- a/src/main/resources/mappers/clickhouse/genomic_data/ClickhouseGenomicDataMapper.xml
+++ b/src/main/resources/mappers/clickhouse/genomic_data/ClickhouseGenomicDataMapper.xml
@@ -133,7 +133,8 @@
             replace(mutation_type, '_', ' ') as label,
             mutation_type as value,
             count(*) as count,
-            count(distinct(sample_unique_id)) as uniqueCount
+            count(distinct(sample_unique_id)) as uniqueCount,
+            arrayStringConcat(groupArray(DISTINCT sample_unique_id), ',') AS sampleIdsStr
             FROM genomic_event_derived
         <where>
             <include
@@ -156,6 +157,7 @@
             <result property="label" column="label"/>
             <result property="count" column="count"/>
             <result property="uniqueCount" column="uniqueCount"/>
+            <result property="sampleIds" column="sampleIdsStr"/>
         </collection>
     </resultMap>
 
--- a/src/main/resources/mappers/clickhouse/genomic_data/ClickhouseGenomicDataMapper.xml
+++ b/src/main/resources/mappers/clickhouse/genomic_data/ClickhouseGenomicDataMapper.xml
@@ -127,26 +127,97 @@
 
     <!-- for /mutation-data-counts/fetch - (returns GenomicDataCountItem objects) mutation type counts table part-->
     <select id="getMutationCountsByType" resultMap="GenomicDataCountItemResultMap">
-        SELECT
-            hugo_gene_symbol as hugoGeneSymbol,
-            'mutations' as profileType,
-            replace(mutation_type, '_', ' ') as label,
-            mutation_type as value,
-            count(*) as count,
-            count(distinct(sample_unique_id)) as uniqueCount,
-            arrayStringConcat(groupArray(DISTINCT sample_unique_id), ',') AS sampleIdsStr
-            FROM genomic_event_derived
-        <where>
+        WITH
+            <if test="includeSampleIds">
+                all_samples AS (
+                SELECT sample_unique_id
+                FROM sample_derived
+                WHERE cancer_study_identifier IN
+                <foreach collection="studyViewFilterContext.studyIds" item="studyId" open="(" separator="," close=")">
+                    #{studyId}
+                </foreach>
+                ),
+            </if>
+            mutated_samples AS (
+                SELECT
+                sample_unique_id,
+                hugo_gene_symbol,
+                mutation_type
+                FROM genomic_event_derived
+                <where>
+                    <include
+                        refid="org.cbioportal.infrastructure.repository.clickhouse.studyview.ClickhouseStudyViewFilterMapper.applyStudyViewFilterUsingSampleId"/>
+                    AND variant_type = 'mutation'
+                    <if test="includeSampleIds">
+                        AND hugo_gene_symbol = #{hugoGeneSymbol}
+                    </if>
+                    <if test="!includeSampleIds">
+                        <foreach item="genomicDataFilter" collection="genomicDataFilters" open=" AND (" separator=" OR " close=")">
+                            hugo_gene_symbol = #{genomicDataFilter.hugoGeneSymbol}
+                        </foreach>
+                    </if>
+                </where>
+            )
+        <if test="includeSampleIds">
+            ,
+            profiled_samples as (
+            SELECT DISTINCT sgp.sample_unique_id
+            FROM sample_to_gene_panel_derived sgp
+            JOIN gene_panel_to_gene_derived gpg ON sgp.gene_panel_id = gpg.gene_panel_id
+            WHERE
             <include
-                    refid="org.cbioportal.infrastructure.repository.clickhouse.studyview.ClickhouseStudyViewFilterMapper.applyStudyViewFilterUsingSampleId"/>
-            AND variant_type = 'mutation'
-            <foreach item="genomicDataFilter" collection="genomicDataFilters" open=" AND (" separator=" OR " close=")">
-                hugo_gene_symbol = #{genomicDataFilter.hugoGeneSymbol}
-            </foreach>
-        </where>
+                refid="org.cbioportal.infrastructure.repository.clickhouse.studyview.ClickhouseStudyViewFilterMapper.applyStudyViewFilterUsingSampleId"/>
+            AND gpg.gene = #{hugoGeneSymbol}
+            AND sgp.alteration_type = 'MUTATION_EXTENDED'
+            )
+        </if>
+               
+        SELECT
+            hugo_gene_symbol AS hugoGeneSymbol,
+            'mutations' AS profileType,
+            replace(mutation_type, '_', ' ') AS label,
+            mutation_type AS value,
+            COUNT(*) AS count,
+            COUNT(DISTINCT sample_unique_id) AS uniqueCount
+            <if test="includeSampleIds">
+                ,
+                arrayStringConcat(groupArray(DISTINCT sample_unique_id), ',') AS sampleIdsStr
+            </if>
+        FROM
+            mutated_samples
         GROUP BY
-        mutation_type,
-        hugo_gene_symbol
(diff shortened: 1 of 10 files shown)


src/main/resources/mappers/clickhouse/genomic_data/ClickhouseGenomicDataMapper.xml
```diff
@@ -133,7 +133,8 @@
             replace(mutation_type, '_', ' ') as label,
             mutation_type as value,
             count(*) as count,
-            count(distinct(sample_unique_id)) as uniqueCount
+            count(distinct(sample_unique_id)) as uniqueCount,
+            arrayStringConcat(groupArray(DISTINCT sample_unique_id), ',') AS sampleIdsStr
```
alisman: you're going to want to make this conditional using MyBatis.  I.e. we will pass in the "flag" all the way from the controller and only return the samples when we need to.

src/test/java/org/cbioportal/infrastructure/repository/clickhouse/genomic_data/ClickhouseGenomicDataMapperTest.java
```diff
@@ -119,7 +119,7 @@ public void getMutationCountsByType() {
     List<GenomicDataCountItem> actualMutationCountsByType =
         mapper.getMutationCountsByType(
             StudyViewFilterFactory.make(studyViewFilter, null, studyViewFilter.getStudyIds(), null),
-            List.of(genomicDataFilterMutation));
+            List.of(genomicDataFilterMutation), false);
```
alisman: can you add a test to this spec where it's true?

i-am-leslie: Yes, will do

alisman (commented):
## Code Review

**Purpose:** Add an `includeSampleIds` flag to the mutation-data-counts endpoint so the comparison page can retrieve which samples belong to each mutation type bucket.

---

### 1. Legacy `StudyViewMapper.xml` is not updated -- divergent code paths (Medium)

The legacy MyBatis mapper at `src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml` has its own copy of the `getMutationCountsByType` query. It was **not updated** with the new `includeSampleIds` parameter or the `sampleIdsStr` column. Similarly, the legacy `StudyViewMapper.java` interface still has the old 2-parameter signature.

The legacy `StudyViewController.java` calls `studyViewService.getMutationTypeCountsByGeneSpecific(studyIds, sampleIds, genomicDataFilters)` (the legacy `StudyViewService` interface), which goes through `StudyViewServiceImpl` and ultimately calls the legacy `StudyViewMyBatisRepository` -- a completely separate code path from the domain layer that was updated.

The legacy endpoint (`/api/mutation-data-counts/fetch` in `StudyViewController`) still works because it was never changed -- it just will not support the new `includeSampleIds` feature. However, two divergent code paths now exist. If the legacy controller is still active, clients may get different behavior depending on which endpoint they hit.

### 2. `GenomicDataCountItemResultMap` now has `sampleIds` for ALL queries that use it (Low)

The shared `GenomicDataCountItemResultMap` in `ClickhouseGenomicDataMapper.xml` now maps `sampleIdsStr` to `sampleIds` **unconditionally**. This same result map is used by `getCNACounts`. The CNA query does not produce a `sampleIdsStr` column, so MyBatis will map it as `null` -- which is harmless. However, **CNA count responses will now include a `sampleIds: null` field in the JSON output** that was not there before. If any frontend code does a strict schema check or if OpenAPI validation is enabled, this could cause issues.

The legacy `StudyViewMapper.xml` result map was not updated, so the legacy path will not have this field -- further divergence.

### 3. `setSampleIds(String)` setter has a type mismatch (Low)

In `GenomicDataCount.java`, the setter accepts a `String` (for MyBatis mapping from comma-delimited `sampleIdsStr`) but the getter returns `List<String>`. This is a JavaBean convention violation -- `setSampleIds(String)` and `getSampleIds()` returning `List<String>` do not match types. While Jackson serialization uses the getter (so JSON output will be `List<String>`), this could break:
- Any code calling `setSampleIds` with a `List<String>` at the Java level (there is no such overload)
- Standard JavaBean introspection tools
- The `equals()`/`hashCode()` methods **do not include `sampleIds`**, meaning two `GenomicDataCount` objects with different sample IDs would be considered equal

### 4. Hardcoded `genomicDataFilters[0]` -- wrong results for multi-gene queries (High)

In the new SQL in `ClickhouseGenomicDataMapper.xml`, the "Not Mutated" and "Not Profiled" UNION ALL blocks and the `profiled_samples` CTE use:
```sql
#{genomicDataFilters[0].hugoGeneSymbol}
```

This hardcodes the **first** gene filter only. If `genomicDataFilters` contains multiple genes, the "Not Mutated" and "Not Profiled" rows will only be computed for `genomicDataFilters[0]`, silently ignoring the rest. The `mutated_samples` CTE correctly uses a `<foreach>` loop over all filters, creating an inconsistency.

**Impact:** When `includeSampleIds=true` and multiple genes are passed, the "Not Mutated" and "Not Profiled" counts and sample IDs will be **wrong** -- they will only reflect the profiling status of the first gene.

### 5. Performance concern: unbounded `groupArray` (Medium)

When `includeSampleIds=true`, the query collects all distinct sample IDs into a comma-separated string via `arrayStringConcat(groupArray(DISTINCT sample_unique_id), ',')`. For large studies with thousands of samples, this can produce enormous strings that:
- Consume significant ClickHouse memory during aggregation
- Create very large JSON responses
- May hit ClickHouse max_query_size or max_memory_usage limits

There is no limit or pagination on the sample IDs list.

### 6. `includeSampleIds` silently ignored with SUMMARY projection (Low)

In `ColumnarStoreStudyViewController.java`:
```java
projection == Projection.SUMMARY
    ? studyViewService.getMutationCountsByGeneSpecific(studyViewFilter, genomicDataFilters)
    : studyViewService.getMutationTypeCountsByGeneSpecific(
        studyViewFilter, genomicDataFilters, includeSampleIds);
```

The `includeSampleIds` parameter only takes effect when `projection != SUMMARY`. If someone passes `?includeSampleIds=true&projection=SUMMARY`, the flag is silently ignored. This is not necessarily a bug, but it is undocumented behavior since the API accepts the parameter in both cases.

### 7. Test gap: no multi-gene test with `includeSampleIds=true` (Medium)

The new test `getMutationCountsByTypeAddSampleId` only tests with a single gene (`AKT1`). Given issue #4 above, a multi-gene test would expose the `genomicDataFilters[0]` bug.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | Medium | Legacy mapper/service not updated -- divergent code paths |
| 2 | Low | CNA results will now include `sampleIds: null` in JSON |
| 3 | Low | `setSampleIds(String)` type mismatch with getter; missing from equals/hashCode |
| 4 | **High** | Hardcoded `genomicDataFilters[0]` -- wrong results for multi-gene queries with `includeSampleIds=true` |
| 5 | Medium | Unbounded `groupArray` could cause performance issues on large studies |
| 6 | Low | `includeSampleIds` silently ignored with SUMMARY projection |
| 7 | Medium | No test coverage for multi-gene + `includeSampleIds=true` scenario |

alisman:
@i-am-leslie i asked Claude code to review the PR.  Maybe we can meet tomorrow  an go over it?  Let me know if you have time.   Some of the issues it raises are probably not a big deal.

i-am-leslie:
### Addressing Claude’s Review

1. **Legacy StudyViewMapper not updated (Medium)**

 The legacy StudyViewController / MyBatis path was intentionally not modified as part of the ClickHouse migration. 

2. **sampleIds added to shared result map (Low)**

This is backward compatible. Queries that don’t populate sampleIdsStr will map it as null, preserving previous behaviour. No functional regression.

3. **setSampleIds(String) type mismatch (Low)**

This was intentional for SQL → domain mapping (comma-delimited aggregation → parsed list). It’s not meant for general mutation via JavaBean usage.
sampleIds is excluded from equals/hashCode by design since it is informational, not part of identity semantics.

4. **Hardcoded genomicDataFilters[0] (High)**

The hardcoding has been removed from the MyBatis layer. A new argument (gene symbol) is passed from the use case when includeSampleIds = true.

Additionally, I now enforce:

If includeSampleIds = true, exactly one gene must be provided.

Otherwise, an error is thrown.

This is intentional and product-driven:

- The UI renders one chart per gene

- The comparison feature is scoped to a specific chart

- Each chart represents a single gene

- Mutation groups (Mutated / Not Mutated) are gene-relative

A sample’s mutation status only makes sense relative to one gene. This prevents ambiguous semantics and silent incorrect results.

5. **Performance concern: unbounded groupArray (Medium)**

includeSampleIds is only used when creating comparison sessions (explicit user action).
This is a new feature behaviour, not a regression. I can find out and document potential size implications and introduce safeguards later if needed.

6. **includeSampleIds ignored with SUMMARY projection (Low)**

Intentional.
includeSampleIds only applies when projection == DETAILED and includeSampleIds == true

7. **No multi-gene test with includeSampleIds=true (Medium)**

I have added a Multi-gene test; Multi-gene + includeSampleIds is invalid by design. We test that this scenario fails fast, which prevents the issue described in #4.

The implementation aligns with product behaviour and user expectations:

Comparison is chart-scoped

Charts are gene-scoped

Sample grouping is gene-relative

alisman (approved):

PR #11909, opened 2026-01-08
Fix 500 error when user has no data access token

### LINKED ISSUE #11905: [BUG] IndexOutOfBoundsException in UUID getDataAccessToken
**Describe the bug**
If user does not have a DAT and goes to Download Access Token, it raises a 500 internal server error.

**To Reproduce**
Setup a new system with oAuth2 login. 
Configure dat.method = Uuid. 
User logs in and select Data Access Token in dropdown. 
User clicks "Download token".
It opens a new page with error.

**Expected behavior**
If not DAT exists, it should create one and share it with user.

**Additional context**
https://github.com/cBioPortal/cbioportal/blob/master/src/main/java/org/cbioportal/legacy/service/impl/UuidDataAccessTokenServiceImpl.java#L102

```
public DataAccessToken getDataAccessToken(String username) {
    List<DataAccessToken> allDataAccessTokens =
        dataAccessTokenRepository.getAllDataAccessTokensForUsername(username);
    DataAccessToken newestDataAccessToken = allDataAccessTokens.get(allDataAccessTokens.size() - 1);
    return newestDataAccessToken;
  }
```

It does not check if allDataAccessTokens len is 0.

Gautam-aman:
### Problem
Calling getDataAccessToken for a user with no existing data access tokens
throws an IndexOutOfBoundsException, resulting in a 500 response.

### Solution
Guard against an empty token list and create a new data access token
when none exists.

### Impact
Prevents server error and aligns behavior with expected token lifecycle.

### Tests
- mvn test -Dtest=UuidDataAccessTokenServiceImplTest

Fixes #11905

### DIFF
--- a/src/main/java/org/cbioportal/legacy/service/impl/UuidDataAccessTokenServiceImpl.java
+++ b/src/main/java/org/cbioportal/legacy/service/impl/UuidDataAccessTokenServiceImpl.java
@@ -102,8 +102,12 @@ public class UuidDataAccessTokenServiceImpl implements DataAccessTokenService {
   public DataAccessToken getDataAccessToken(String username) {
     List<DataAccessToken> allDataAccessTokens =
         dataAccessTokenRepository.getAllDataAccessTokensForUsername(username);
-    DataAccessToken newestDataAccessToken = allDataAccessTokens.get(allDataAccessTokens.size() - 1);
-    return newestDataAccessToken;
+
+    if (allDataAccessTokens == null || allDataAccessTokens.isEmpty()) {
+      return createDataAccessToken(username);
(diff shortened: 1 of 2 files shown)


sachindigharra:
Issue: The user is still logged into the system but doesn’t have a token to access resources.
Solution: Since the user is already verified, we should generate and store a token instead of re-verifying. This ensures that future resource access is not interrupted.

Gautam-aman:
Yes, exactly this change ensures that already-authenticated users receive
a newly generated and persisted data access token when none exists, avoiding
re-verification and preventing 500 errors on resource access.

PR #11910, opened 2026-01-08
Return 404 instead of 500 for missing reference genome gene

Gautam-aman:
### Problem
Requesting a reference genome gene that does not exist results in a 500 server error.

### Solution
Add explicit null handling in ReferenceGenomeGeneController and return HTTP 404 when the gene is missing.

### DIFF
--- a/src/main/java/org/cbioportal/legacy/web/ReferenceGenomeGeneController.java
+++ b/src/main/java/org/cbioportal/legacy/web/ReferenceGenomeGeneController.java
@@ -69,7 +69,7 @@ public class ReferenceGenomeGeneController {
       responseCode = "200",
       description = "OK",
       content = @Content(schema = @Schema(implementation = ReferenceGenomeGene.class)))
-  @ApiResponse(responseCode = "404", description = "Gene not found")
+  @ApiResponse(responseCode = "404", description = "Reference genome or gene not found")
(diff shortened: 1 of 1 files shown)


Gautam-aman:
This PR applies a minimal defensive fix to return HTTP 404 instead of 500
when a reference genome gene is missing.

No business logic was changed and no new dependencies were introduced.
Please let me know if you would like me to add a controller test.

src/main/java/org/cbioportal/legacy/web/ReferenceGenomeGeneController.java
```diff
@@ -70,14 +69,21 @@ public ResponseEntity<List<ReferenceGenomeGene>> getAllReferenceGenomeGenes(
       responseCode = "200",
       description = "OK",
       content = @Content(schema = @Schema(implementation = ReferenceGenomeGene.class)))
+  @ApiResponse(responseCode = "404", description = "Gene not found")
```
onursumer: It might be better if the description says `Reference genome or gene not found` or something similar.

Gautam-aman: Updated the 404 description to “Reference genome or gene not found”. Thanks for the suggestion.

PR #11928, opened 2026-01-27
Fix allSampleCount in /column-store/studies endpoint

### LINKED ISSUE #11927: [BUG] Some studies have inflated allSampleCount from /studies endpoint
**Describe the bug**
allSampleCount for some studies returned by /api/column-store/studies are inflated

**To Reproduce**
1. `curl -X 'GET' \
  'https://www.cbioportal.org/api/column-store/studies?keyword=luad%20tcga&projection=SUMMARY&pageSize=10000000&pageNumber=0&direction=ASC' \
  -H 'accept: application/json'`
2. check study 'luad_tcga' it has 1166 allSampleCount

Image

**Expected behavior**
In https://www.cbioportal.org/study/summary?id=luad_tcga it only has 586 samples

Image

fuzhaoyuan:
Fix #11927

**Describe changes proposed in this pull request:**
- Use exact matching for sample_list instead of like matching

**Cause to the issue:**
For study 'luad_tcga', like matching matched both 'luad_tcga_all' (586 sample count) and 'luad_tcga_methylation_all' (580 sample count), then the allSampleCount is inflated to 586+580=1166

### DIFF
--- a/src/integration/java/org/cbioportal/infrastructure/repository/clickhouse/cancerstudy/ClickhouseCancerStudyRepositoryIntegrationTest.java
+++ b/src/integration/java/org/cbioportal/infrastructure/repository/clickhouse/cancerstudy/ClickhouseCancerStudyRepositoryIntegrationTest.java
@@ -3,6 +3,9 @@ package org.cbioportal.infrastructure.repository.clickhouse.cancerstudy;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import org.cbioportal.AbstractClickhouseIntegrationTest;
+import org.cbioportal.domain.sample.repository.SampleRepository;
+import org.cbioportal.infrastructure.repository.clickhouse.sample.ClickhouseSampleMapper;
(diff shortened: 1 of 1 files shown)


src/main/resources/mappers/clickhouse/cancerstudy/CancerStudyMapper.xml
```diff
@@ -8,20 +8,21 @@
             resultMap="CancerStudyMetadataResultMap">
         WITH sample_counts as (
             SELECT
-                sample_list.cancer_study_id,
-                countIf(stable_id LIKE '%_all') as allSampleCount,
```
alisman, line 12: @fuzhaoyuan @haynescd i don't actually understand how this worked at all in the past.  it seems like the counts would not be study specific at all but would just always count all the samples.   lets make sure we understand this.   @fuzhaoyuan i think we should put some tests for this.

fuzhaoyuan: Yeah it's because the `GROUP BY` below limited this to study specific. I'll add some tests

PR #11936, opened 2026-02-03
Add news item for VAF in Plots tab feature

inodb:
Adds a news item for the Variant Allele Frequency feature in the Plots tab.

Reference: https://github.com/cBioPortal/cbioportal-frontend/pull/5209


inodb:
Use this screenshot: 
image

PR #11941, opened 2026-02-06
Optimize clinical_event table (slung from MySQL) with primary key

alisman:
Fixes: https://github.com/cBioPortal/cbioportal/issues/11942


dippindots (approved):
LGTM

PR #11991, opened 2026-02-25
Add ClickHouse-backed co-expression endpoint

alisman:
## Summary

- Adds a new `/api/column-store/molecular-profiles/co-expressions/fetch` endpoint that computes Spearman rank correlations using ClickHouse's built-in `rankCorr()` aggregate function against the `genetic_alteration_derived` table
- Follows the existing DDD column-store architecture pattern (Controller → UseCase → Domain Repository → ClickHouse Infrastructure)
- The legacy co-expression endpoint remains completely untouched — both implementations coexist

## Motivation

The legacy co-expression implementation fetches all expression data (~20K genes × N samples) into Java heap memory and computes correlations one-by-one using Apache Commons Math3. This is memory-intensive (100-350MB heap per request) and slow (3-35s depending on study size).

The ClickHouse implementation pushes the computation into the database, returning only filtered results. Tested across 24 diverse studies:

- **2x faster** overall (112s vs 220s total across all tests)
- **61% less heap memory** (avg 71MB vs 182MB per request)
- **Zero committed memory growth** vs 5.7GB cumulative in legacy
- **97-100% correlation match** with default threshold (0.3)

## Implementation

| File | Description |
|------|-------------|
| `ColumnStoreCoExpressionController.java` | New REST controller at `/api/column-store/...` |
| `FetchCoExpressionsUseCase.java` | Domain use case — resolves profile metadata, calls repository, computes p-values |
| `CoExpressionRepository.java` | Domain repository interface |
| `ClickhouseCoExpressionRepository.java` | Infrastructure implementation delegating to MyBatis mapper |
| `ClickhouseCoExpressionMapper.java` | MyBatis mapper interface |
| `ClickhouseCoExpressionMapper.xml` | ClickHouse SQL using `rankCorr()` with direct WHERE parameters for sort key pruning |
| `CoExpressionResult.java` | DTO (entrezGeneId, spearmansCorrelation, numSamples) |

## Known Differences

- **Result count**: ClickHouse returns ~1500-2400 more genes per study because its gene table is larger than MySQL's. All legacy genes are present in ClickHouse results.
- **Spearman tie-handling**: ClickHouse `rankCorr()` and Apache Commons Math3 use different tie-breaking strategies. With the default threshold (0.3), this affects <0.3% of shared genes. Differences are concentrated in genes with many identical values (e.g., mostly zeros), where Spearman correlation is unreliable regardless.

## Test plan

- [x] Verify `/api/column-store/molecular-profiles/co-expressions/fetch` returns results for a sample study (e.g., ACC KRAS with `sampleListId=acc_tcga_all`)
- [x] Verify legacy endpoint `/api/molecular-profiles/co-expressions/fetch` still works unchanged
- [x] Test dual-profile case (different molecularProfileIdA and molecularProfileIdB)
- [x] Test with explicit sample IDs instead of sampleListId
- [x] Test with threshold=0 (returns all genes)
- [x] Verify p-values are non-zero for strong correlations

### DIFF
--- a/src/main/java/org/cbioportal/domain/coexpression/usecase/FetchCoExpressionsUseCase.java
+++ b/src/main/java/org/cbioportal/domain/coexpression/usecase/FetchCoExpressionsUseCase.java
@@ -106,11 +106,13 @@ public class FetchCoExpressionsUseCase {
               CoExpression coExpression = new CoExpression();
               coExpression.setGeneticEntityId(String.valueOf(result.getEntrezGeneId()));
               coExpression.setGeneticEntityType(EntityType.GENE);
-              coExpression.setSpearmansCorrelation(
-                  BigDecimal.valueOf(result.getSpearmansCorrelation()));
-              double pValue =
-                  computePValue(result.getSpearmansCorrelation(), result.getNumSamples());
-              coExpression.setpValue(BigDecimal.valueOf(pValue));
+              if (result.getSpearmansCorrelation() != null) {
+                coExpression.setSpearmansCorrelation(
+                    BigDecimal.valueOf(result.getSpearmansCorrelation()));
+                double pValue =
+                    computePValue(result.getSpearmansCorrelation(), result.getNumSamples());
+                coExpression.setpValue(BigDecimal.valueOf(pValue));
+              }
               return coExpression;
             })
         .toList();
--- a/src/main/resources/mappers/clickhouse/coexpression/ClickhouseCoExpressionMapper.xml
+++ b/src/main/resources/mappers/clickhouse/coexpression/ClickhouseCoExpressionMapper.xml
@@ -50,7 +50,25 @@
            AND uniq(toFloat64OrNull(d.alteration_value)) > count(*) / 2
            AND isFinite(spearmansCorrelation)
            AND abs(spearmansCorrelation) >= #{threshold}
-        ORDER BY abs(spearmansCorrelation) DESC
+
+        UNION ALL
+
+        SELECT
+            g.entrez_gene_id AS entrezGeneId,
+            NULL AS spearmansCorrelation,
+            count(*) AS numSamples
+        FROM genetic_alteration_derived d
+        INNER JOIN ref_values r ON d.sample_unique_id = r.sample_unique_id
+        INNER JOIN gene g ON d.hugo_gene_symbol = g.hugo_gene_symbol
+        WHERE d.cancer_study_identifier = #{cancerStudyIdentifierB}
+          AND d.profile_type = #{profileTypeB}
+          AND d.hugo_gene_symbol != #{hugoGeneSymbol}
+          AND d.alteration_value != '' AND d.alteration_value != 'NA'
+          AND toFloat64OrNull(d.alteration_value) IS NOT NULL
+        GROUP BY g.entrez_gene_id
+        HAVING count(*) &lt; 3
+            OR uniqExact(toFloat64OrNull(d.alteration_value)) = 1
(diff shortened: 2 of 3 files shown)


alisman:
@jjgao @inodb Claude's implementation of co-expression has the following difference, which results in discrepancies with legacy.  Can you read the below and judge whether this is a problem or now?

 Spearman Tie-Handling Difference: Summary

  Spearman rank correlation works by converting raw values to ranks, then computing correlation on the ranks. When multiple samples have the same value (a "tie"), they must be assigned ranks — and there
  are different valid strategies for doing this.

  Apache Commons Math3 (legacy) assigns tied values the average of their ranks (e.g., three samples tied for positions 5-7 all get rank 6.0), then computes Pearson correlation on the averaged ranks.

  ClickHouse rankCorr() uses a different internal algorithm. The exact tie-breaking strategy isn't documented, but produces slightly different correlations when ties are present.

  When does this matter?

  - For genes with mostly unique expression values, the two implementations produce virtually identical results (diff < 0.0001).
  - For genes with many identical values (e.g., 240 out of 244 samples with expression = 0), the rank assignments diverge, producing correlation differences > 0.1.
  - With the default threshold of 0.3, 97-100% of results match closely across all tested studies. The few large diffs appear on genes with heavy ties — which inherently have unreliable correlations
  anyway.
  - With threshold=0 (returning all genes), ~12% of results show large diffs, but these are all weak correlations near zero where tied values dominate.

  Bottom line: The differences are statistically insignificant and only affect genes whose expression data is too sparse/degenerate for Spearman correlation to be meaningful in the first place. No
  actionable co-expression relationship would be missed or misrepresented.

inodb:
@alisman nice explanation :) - the changes in handling ties seems fine to me

alisman:
This is the memory profile of the two endpoints.  First arrow shows legacy coexpression and second arrow shows new one. 
image

README.md
```diff
@@ -183,7 +183,7 @@ npm install
 CBIOPORTAL_URL=http://localhost:8082 npm test
 
 # Run a specific test suite
-npm test 'test/ColumnStoreStudyController/*.spec.ts'
+CBIOPORTAL_URL=http://localhost:8082 npm test 'test/ColumnStoreStudyController/*.spec.ts'
```
dippindots: @alisman let's do some cleanup before merge

.circleci/config.yml
```diff
@@ -539,8 +539,6 @@ workflows:
               - build_frontend
               - build_web_and_data_image
 #        - run_api_tests:
-#            context:
-#              - api-tests
```
dippindots, line 543: maybe remove this part for now?

alisman: i'm gonna follow up quickly with this one so lets just leave it.

src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreCoExpressionController.java
```diff
@@ -0,0 +1,76 @@
...
+      @Parameter(required = true, description = "Molecular Profile ID B") @RequestParam
+          String molecularProfileIdB,
+      @Parameter(required = true, description = "Co-Expression Filter") @Valid @RequestBody
+          CoExpressionFilter coExpressionFilter,
+      @Parameter(description = "Threshold") @RequestParam(defaultValue = "0.3") Double threshold)
+      throws MolecularProfileNotFoundException, GeneNotFoundException {
+
+    if (coExpressionFilter.getEntrezGeneId() == null) {
+      return new ResponseEntity<>(List.of(), HttpStatus.OK);
+    }
```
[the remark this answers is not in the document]
alisman: fixed

PR #11999, opened 2026-02-28
Update study specific cache invalidation endpoint URL to include '/api'

saxenanurag:
Study specific cache invalidation endpoint is missing `/api`. This fixes it.

There are no existing tests for this.


saxenanurag:
Hey Maintainers. I created this PR because I saw this bug in cBioPortal version 6 where you cannot clear the cache for a single study. I understand that this may not be patched in version 6. Hopefully, it will be fixed in version 7. Thank you.

Issue #5437, opened 2018-12-13
Study view: Number of samples incorrect in various pie charts


paulcwvandijk:
Steps to reproduce:
Go to url: http://www.cbioportal.org/beta/study?id=brca_tcga

Look at Pie Chart "Sample Type"

Look at sample and patient size at the header

Total of both (1,105 vs 1,108) don't add up to same amount.

![screenshot 2018-12-13 at 16 06 48](https://user-images.githubusercontent.com/44695112/49947393-70edb480-fef1-11e8-8fe9-69daeea0672c.png)
![screenshot 2018-12-13 at 16 09 25](https://user-images.githubusercontent.com/44695112/49947411-7c40e000-fef1-11e8-98c0-eaade59bbe33.png)

- Desired solution: Numbers should match

pieterlukasse:
notify: @jjgao @zhx828

ersinciftci:
The reason is, this query:
```
SELECT * FROM public_test.sample inner join patient on sample.PATIENT_ID = patient.INTERNAL_ID 
inner join cancer_study on cancer_study.CANCER_STUDY_ID = patient.CANCER_STUDY_ID
where cancer_study.CANCER_STUDY_IDENTIFIER = 'brca_tcga';
```
returns 1108 patients but this query:
```
SELECT * FROM public_test.sample_list_list inner join sample_list on sample_list_list.LIST_ID = sample_list.LIST_ID where sample_list.STABLE_ID = 'brca_tcga_all';
```
return 1105. So `brca_tcga_all` list doesn't have all the samples in `brca_tcga` for some reason. Seems like a data issue.

jjgao:
@ersinciftci @zhx828 instead of loading all samples, should we load a list, by default case_all? Frontend can pass this list to backend.

ersinciftci:
I think short term solution can be calling the `/filtered-samples/fetch` endpoint instead of `/samples/fetch` in the first page load (it's better to do that in terms of consistency anyways, we already call `/filtered-samples/fetch` for every subsequent filter). We can talk about getting rid of the case lists concept after the release.

pieterlukasse:
Getting rid of case lists could prove difficult,  since they are used to keep track of what has been sequenced for example. But perhaps this has been mostly made redundant by the gene panel implementation...not sure.

pieterlukasse:
[...]

PR #12010, opened 2026-03-05
Hackathon: Initial draft for resource data model

dippindots:

- a
- b

# Checks

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/plan.md
+++ b/plan.md
@@ -46,32 +46,57 @@ CREATE TABLE `resource_node` (
 | `NODE_TYPE=ITEM` | Leaf node; has URL; no children |
 | `PARENT_ID=NULL` | Root node directly under a `resource_definition` |
 | `TYPE` | Free-text domain label (e.g. `H_AND_E`, `IHC`, `CT`, `BAM`, `PDF`); no schema change needed for new domains |
+| `METADATA` | JSON blob for descriptive domain-specific data (e.g. stain info, acquisition params) |
+| `PRIORITY` | Integer for explicit display ordering; kept as a dedicated column (not in METADATA) so `ORDER BY PRIORITY` is efficient |
 
 ### Migration
 
-Existing `resource_sample`, `resource_patient`, `resource_study` rows each become a root-level `ITEM` node with `PARENT_ID=NULL`.
+Existing `resource_sample`, `resource_patient`, `resource_study` rows each become a root-level `ITEM` node with `PARENT_ID=NULL`. Migrated rows will have `TYPE=NULL` and `METADATA=NULL` since old tables have no equivalent columns — this is acceptable and expected.
+
+**Note on `DISPLAY_NAME` in migration:** old tables have no per-row display name. The resource ID is used as the node's `DISPLAY_NAME`, NOT `resource_definition.DISPLAY_NAME`, which is already used as the tree root label.
+
+**Note on `ENTITY_INTERNAL_ID` for `resource_study`:** study-scoped nodes have no patient/sample to reference. Use the study's own `CANCER_STUDY_ID` as `ENTITY_INTERNAL_ID`.
 
 ```sql
--- Example: resource_patient → resource_node
+-- resource_patient → resource_node
 INSERT INTO resource_node
   (RESOURCE_ID, CANCER_STUDY_ID, ENTITY_TYPE, ENTITY_INTERNAL_ID,
    PARENT_ID, NODE_TYPE, DISPLAY_NAME, URL)
 SELECT
-  rd.RESOURCE_ID, rd.CANCER_STUDY_ID, 'PATIENT', rp.INTERNAL_ID,
-  NULL, 'ITEM', rd.DISPLAY_NAME, rp.URL
+  rp.RESOURCE_ID, rd.CANCER_STUDY_ID, 'PATIENT', rp.INTERNAL_ID,
+  NULL, 'ITEM', rp.RESOURCE_ID, rp.URL
 FROM resource_patient rp
 JOIN resource_definition rd
   ON rp.RESOURCE_ID = rd.RESOURCE_ID;
 
--- Same pattern for resource_sample (ENTITY_TYPE='SAMPLE')
--- and resource_study (ENTITY_TYPE='STUDY')
+-- resource_sample → resource_node (ENTITY_TYPE='SAMPLE')
+INSERT INTO resource_node
+  (RESOURCE_ID, CANCER_STUDY_ID, ENTITY_TYPE, ENTITY_INTERNAL_ID,
+   PARENT_ID, NODE_TYPE, DISPLAY_NAME, URL)
+SELECT
+  rs.RESOURCE_ID, rd.CANCER_STUDY_ID, 'SAMPLE', rs.INTERNAL_ID,
+  NULL, 'ITEM', rs.RESOURCE_ID, rs.URL
+FROM resource_sample rs
+JOIN resource_definition rd
+  ON rs.RESOURCE_ID = rd.RESOURCE_ID;
+
+-- resource_study → resource_node (ENTITY_TYPE='STUDY', ENTITY_INTERNAL_ID=CANCER_STUDY_ID)
+INSERT INTO resource_node
+  (RESOURCE_ID, CANCER_STUDY_ID, ENTITY_TYPE, ENTITY_INTERNAL_ID,
+   PARENT_ID, NODE_TYPE, DISPLAY_NAME, URL)
+SELECT
+  rs.RESOURCE_ID, rd.CANCER_STUDY_ID, 'STUDY', rd.CANCER_STUDY_ID,
+  NULL, 'ITEM', rs.RESOURCE_ID, rs.URL
+FROM resource_study rs
+JOIN resource_definition rd
+  ON rs.RESOURCE_ID = rd.RESOURCE_ID;
 ```
 
 ---
 
 ## Import File Format
 
-Three optional columns added to existing data files
+Five optional columns added to existing data files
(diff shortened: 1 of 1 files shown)


dippindots:
Here are the key topics worth discussing before moving to implementation:

---

## 1. Controlled vocabulary for `TYPE`

- Should `TYPE` values (e.g., `H_AND_E`, `IHC`, `CT`) be a curated list enforced somewhere, or truly free-text with no validation?
- Who owns the list, and how do new types get added?
- Does `TYPE` drive any rendering behavior in the UI (e.g., iframe vs download link vs embedded viewer)?

## 2. Metadata on `GROUP` nodes

- Currently only `ITEM` nodes carry `TYPE` and `METADATA`. Should `GROUP` nodes also have metadata?
- Example: a tissue block `GROUP` might want to store anatomical site, collection date, or pathology notes alongside it.

## 3. Tree depth limits

- Is there a practical maximum depth the UI should support?
- Unlimited depth is possible in the schema but could create UX problems if curators go too deep.

## 4. Cross-entity queries

- Should users be able to query across patients — e.g., “show all `H_AND_E` slides in this cohort”?
- If yes, the current per-patient tree model works, but the API and UI need explicit design for it.

## 5. Incremental updates vs full reload

- When a study is re-imported, how are existing `resource_node` rows handled — full delete and reload, or upsert/diff?
- `GROUP_PATH` uniqueness: if two import files define the same path for the same patient, do they merge or conflict?

## 6. Backward compatibility timeline

- How long do the old `resource_sample`, `resource_patient`, and `resource_study` tables and their API endpoints stay alive for existing consumers?
- Is there a versioned migration path or a hard cutover?

## 7. API shape

- Should the API return the full tree in one response, or lazy-load children on expand?
- For cohort-level pages (not single patient), how should imaging resources be surfaced?

## 8. Who creates `GROUP` nodes — curators or the system?

- Current plan: importer auto-creates `GROUP`s from `GROUP_PATH`.
- Is there a use case where curators need to explicitly define and name a `GROUP` with its own metadata before adding items to it?

dippindots:
# PR Mismatch Findings

Sources reviewed:
- Plan: cbioportal#12010
- Backend schema+mapper: cbioportal#12018
- Importer: [cbioportal-core#119](https://github.com/cBioPortal/cbioportal-core/pull/119)
- Validator: [cbioportal-core#120](https://github.com/cBioPortal/cbioportal-core/pull/120)
- File format docs: cbioportal#12022
- Frontend: [cbioportal-frontend#5437](https://github.com/cBioPortal/cbioportal-frontend/pull/5437)

---

## 1. [Backend + Importer] Mapper queries columns that don't exist in the schema

**Files:** `ResourceDataMapper.xml` (PR #12018), `DaoResourceData.java` + `ImportResourceData.java` (PR core#119)

**Problem:** The MyBatis mapper's `selectNodeFields` SQL block queries:
```sql
resource_node.node_type   AS "nodeType"
resource_node.parent_id   AS "parentId"
```
But `cgds.sql` / `migration.sql` (same PR) defines `resource_node` with **`GROUP_PATH varchar(255)`** — no `NODE_TYPE` and no `PARENT_ID` columns. The importer also inserts `group_path` as a flat string rather than resolving it to parent rows.

The plan (PR #12010) designed the schema with `PARENT_ID` + `NODE_TYPE ENUM('GROUP','ITEM')`. The Java model (`NodeType.java`, `ResourceData.nodeType`, `ResourceData.parentId`) and mapper were written for that design, but the schema and importer drifted to a flat `GROUP_PATH` string instead.

**Will cause:** SQL runtime error the moment any resource API endpoint is called.

### Fix: Resolve `GROUP_PATH` to `PARENT_ID` at import time

The mapper and Java model already implement the `PARENT_ID` design correctly. The fix is confined to the schema and importer only.

**Schema changes** (`cgds.sql` + `migration.sql` in PR #12018):
- Replace `GROUP_PATH varchar(255)` with:
  - `NODE_TYPE ENUM('GROUP','ITEM') NOT NULL`
  - `PARENT_ID bigint DEFAULT NULL` with `FOREIGN KEY (PARENT_ID) REFERENCES resource_node(ID) ON DELETE CASCADE`
- Add a unique index on `(RESOURCE_ID, CANCER_STUDY_ID, ENTITY_TYPE, ENTITY_INTERNAL_ID, PARENT_ID, DISPLAY_NAME)` to allow efficient GROUP node upsert (two rows with the same parent + display name at the same scope are the same folder).

**Importer changes** (`DaoResourceData.java` + `ImportResourceData.java` in PR core#119):

1. Change the `INSERT` statement: replace `group_path` with `node_type` and `parent_id`.
2. Add a `findOrCreateGroupNode()` helper in `DaoResourceData` that does a SELECT-then-INSERT for each GROUP node:
```java
// SELECT ID ... WHERE node_type='GROUP' AND display_name=segment AND parent_id IS [NULL / = currentParentId]
// If not found: INSERT GROUP node, return generated ID
```
3. In `ImportResourceData`, before inserting each ITEM row, walk the `GROUP_PATH` segments to resolve the final `parentId`:
```
segments = GROUP_PATH.split("/")   // e.g. ["CT 2023-01-15", "Series 1: Axial T2"]
currentParentId = null

for each segment:
    currentParentId = DaoResourceData.findOrCreateGroupNode(
        resourceId, cancerStudyId, entityType, entityInternalId,
        segment, currentParentId)

INSERT ITEM node with parent_id = currentParentId, node_type = 'ITEM'
```
4. Keep an **in-memory cache** `Map<String, Long>` keyed on `"resourceId|entityId|parentId|segment"` within each import run. This avoids redundant DB lookups when many rows share the same folder (e.g. dozens of slides all under `Block A – Primary Tumor`).

5. **ClickHouse bulk load path**: GROUP nodes require a SELECT to get the generated ID before inserting children, so GROUP resolution must run outside the bulk loader. Only ITEM rows should go through `ClickHouseBulkLoader`; GROUP inserts always use the regular JDBC path.

**What does NOT need to change:**
- `ResourceDataMapper.xml` — already correct for `PARENT_ID` / `NODE_TYPE`
- `ResourceData.java`, `NodeType.java` — already correct
- File format docs (PR #12022) — `GROUP_PATH` stays as the user-facing column name
- Validator (PR core#120) — no changes needed
- Frontend — receives `parentId` in API responses, which it already handles in `ResourceNodeBase`

---

## 2. [Importer] `DISPLAY_NAME` is never read from the TSV — hardcoded to `resourceId`

**File:** `ImportResourceData.java` (PR core#119)
**Problem:** The importer calls:
```java
DaoResourceData.addPatientDatum(internalId, cancerStudyId, resourceId, resourceId, resourceURL, metadata, groupPath);
//                                                                        ↑ displayName arg is just resourceId again
```
There is no `DISPLAY_NAME_COLUMN_NAME` constant and no `findDisplayNameColumn()`. So whatever users put in the `DISPLAY_NAME` column of their data files is silently ignored, and `resourceId` is stored as the display name instead.

**Contradicts:** File format docs (PR #12022) and validator (PR core#120) both support `DISPLAY_NAME` as a meaningful optional column.

---

## 3. [Importer] `TYPE` column is not read from TSV and not stored in DB

**File:** `ImportResourceData.java` + `DaoResourceData.java` (PR core#119)
**Problem:** The DB `INSERT` statement only covers 8 columns:
```
resource_id, cancer_study_id, entity_type, entity_internal_id,
display_name, url, metadata, group_path
```
`TYPE` is completely absent — no column constant, no `findTypeColumn()`, not in the INSERT.

**Contradicts:** The DB schema has `TYPE varchar(64)`, file format docs document it, and the validator validates it. Data in `TYPE` column will be silently dropped on import.

---

## 4. [Importer] `PRIORITY` column is not read from TSV and not stored in DB

**File:** `ImportResourceData.java` + `DaoResourceData.java` (PR core#119)
**Problem:** Same as TYPE — `PRIORITY` has no column constant, no `findPriorityColumn()`, and is absent from the INSERT statement. The DB will always get the schema default of `0` regardless of what the data file says.

**Contradicts:** DB schema has `PRIORITY int(11) DEFAULT 0`, file format docs document it, and the validator validates it (checks it's an integer).

---

## 5. [Frontend] `ResourceNodeRow` is missing the `priority` field

**File:** `ResourceNodeTypes.ts` (PR #5437)
**Problem:** The TSV row interface:
```typescript
interface ResourceNodeRow {
  patientId: string; sampleId: string; resourceId: string;
  url: string; displayName: string; type?: string;
  groupPath?: string; metadata?: Record<string, any>;
  // no priority field
}
```
`priority` is defined on `ResourceNodeBase` (the tree node type) but missing from `ResourceNodeRow` (the parsed TSV row). So the TSV parser won't capture it and ordering will be undefined.

---

## 6. [Frontend] Data fetched from hardcoded `localhost:3000`, bypassing the backend API entirely

**Files:** Two separate components in PR #5437
**Problem:**
```typescript
const RESOURCE_BASE_URL = '//localhost:3000/';
const response = await fetch(`${RESOURCE_BASE_URL}${tsvFile}`);
```
This appears in two places. The frontend parses the raw TSV file client-side and never calls the REST API that the backend (PR #12018) built. This only works in local dev. In production, the `localhost:3000` URL will fail.

**Required work:** Wire the frontend to the actual `/api/...` resource endpoints instead of fetching TSVs directly.

---

## 7. [Frontend ↔ API] `metadata` type mismatch — object vs string

**Problem:** Frontend types `metadata` as `Record<string, any>` (already-parsed JS object). The backend `ResourceData.java` stores and returns it as `String`. When the frontend is eventually wired to the real API, calling `.stain` on a raw JSON string will silently fail or throw.

**Required work:** Frontend needs a `JSON.parse(metadata)` step when consuming the API response, or the API should serialize it as a JSON object.

PR #12039, opened 2026-03-18
Migrate legacy mappers to clinical_event_derived

alisman:
## Summary
- Migrates clinical_event_derived table to clinical_event_data_derived and updates all references accordingly.
- Migrates all legacy mybatis mapper files (`ClinicalEventMapper.xml`, `TreatmentMapper.xml`, `StudyMapper.xml`, `ClinicalAttributeDataMapper.xml`) from querying the base `clinical_event` table to using `clinical_event_derived`, which has a proper primary key and denormalized columns for better performance.

## Test plan
- [x] `ClinicalEventMyBatisRepositoryTest` — 17 tests passing
- [x] `TreatmentMyBatisRepositoryTest` — 7 tests passing
- [x] `StudyMyBatisRepositoryTest` — 12 tests passing
- [x] Export-related tests — 39 tests passing
- [x] E2E tests against running instance — 19 tests passing

### DIFF
--- a/src/main/resources/mappers/clickhouse/cancerstudy/CancerStudyMapper.xml
+++ b/src/main/resources/mappers/clickhouse/cancerstudy/CancerStudyMapper.xml
@@ -27,7 +27,7 @@
         ),
         treatment AS (
             SELECT
-                COUNT(DISTINCT patient_stable_id) as count,
+                COUNT(DISTINCT patient_unique_id) as count,
                 cancer_study_identifier
             FROM clinical_event_data_derived
             <where>


src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -372,7 +372,7 @@ FROM patient AS p
                          ON (p.internal_id = clinpat.internal_id) AND (clinpat.attr_id = cam.attr_id)
 WHERE cam.patient_attribute = 1;
 
-CREATE TABLE clinical_event_derived
+CREATE TABLE clinical_event_data_derived
```
alisman, line 404: this corrects the misnamed table

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -708,9 +708,9 @@ SELECT * FROM genetic_alteration;
 -- SWITCH THE TABLES
 EXCHANGE TABLES genetic_alteration_BACKUP AND genetic_alteration;
 
--- Adds primary key to the clinical_event table for Clickhouse-only
-DROP TABLE IF EXISTS clinical_event_BACKUP;
-CREATE TABLE clinical_event_BACKUP
+-- Creates and populates clinical_event_derived with a primary key for Clickhouse-only (original clinical_event table remains unchanged)
+DROP TABLE IF EXISTS clinical_event_derived;
```
dippindots: We should also move this `DROP` query to the beginning of the script.

alisman: i think this was done

sheridancbio: I agree ... this should be moved back up to become line 11. I'll see if I can make that change.
I also wonder why this is being defined down here and not adjacent to the clinical_event_data_derived table up higher in the file. I guess it does not have to be there, but I kind of like the division of the actual "derived" tables and the adjustments to the ORDER BY clauses of a few of the core tables. I think this now feels more at home up with the other table derivations.
This is especially true because this part is in a section which is commented with:
-- START: PRIMARY KEY ADDITIONS
--END: PRIMARY KEY ADDITIONS
But this is no longer a PRIMARY KEY ADDITION

I also would also suggest dropping some excessive whitespace.

src/main/resources/mappers/clickhouse/studyview/ClickhouseStudyViewFilterMapper.xml
```diff
@@ -188,7 +188,7 @@
                     ced.patient_unique_id AS patient_unique_id,
                     min(ced.start_date) AS time_taken,
                     ced.cancer_study_identifier AS cancer_study_identifier
-                FROM clinical_event_derived ced
+                FROM clinical_event_data_derived ced
```
sheridancbio: now that we have both a clinical_event_data_derived and also a clinical_event_derived, perhaps the alias assigned to this table should be "cedd" and not "ced"

src/main/resources/mappers/export/ClinicalAttributeDataMapper.xml
```diff
@@ -136,7 +136,7 @@
     <select id="hasClinicalTimelineData" resultType="java.lang.Boolean">
         <bind name="sampleIdsArray" value="sampleIds != null ? sampleIds.toArray(new String[sampleIds.size()]) : null" />
         SELECT count() > 0
-        FROM clinical_event ce
+        FROM clinical_event_derived ce
```
sheridancbio: alias could now be "ced", not "ce" (there are 4 more occurrences below in this file)

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewFilterMapper.xml
```diff
@@ -188,7 +188,7 @@
                     ced.patient_unique_id AS patient_unique_id,
                     min(ced.start_date) AS time_taken,
                     ced.cancer_study_identifier AS cancer_study_identifier
-                FROM clinical_event_derived ced
+                FROM clinical_event_data_derived ced
```
sheridancbio: table alias could be cedd (because we do have a table called clinical_event_data too and this might be confusing)

src/main/resources/org/cbioportal/persistence/mybatisclickhouse/StudyViewMapper.xml
```diff
@@ -629,7 +629,7 @@
     <select id="getTotalSampleTreatmentCounts" resultType="int">
         SELECT
             count(distinct ced.value) AS totalSamples
-        FROM clinical_event_derived ced
+        FROM clinical_event_data_derived ced
```
sheridancbio: table alias could be cedd (there is one more occurrence like this later in this file)

PR #12048, opened 2026-03-23
Fix/12012 discrete cna endpoint performance

### LINKED ISSUE #12012: [Performance] Discrete copy number endpoint
needs to look at discrete copy number endpoint
Image

VarshiniGunti:
Fix #12012

- Add pagination support (`pageSize`, `pageNumber`) for single-profile discrete CNA endpoints:
  - `GET /api/molecular-profiles/{molecularProfileId}/discrete-copy-number`
  - `POST /api/molecular-profiles/{molecularProfileId}/discrete-copy-number/fetch`
- Propagate pagination through controller → service → repository → MyBatis mapper for those endpoints.
- Add SQL `LIMIT/OFFSET` with deterministic ordering for `getDiscreteCopyNumbersBySampleListId` and `getDiscreteCopyNumbersBySampleIds`.
- Keep backward-compatible behavior by default (`pageSize=0` => no paging).
- Harden in-memory pagination in the non-HOMDEL/AMP code path (defensive offset handling and safe copy of paged sublists).
- Add focused tests for:
  - Controller pagination parameter forwarding
  - Service pagination behavior for non-HOMDEL/AMP path
  - MyBatis pagination behavior for single-profile fetch path

# Checks
- [x] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). We can fix this during merge by using a squash+merge if necessary
- [x] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!

# Any screenshots or GIFs?
Not applicable (backend/API change only; no UI change).

# Notify reviewers
Potential reviewers identified from file history (`git blame`/`git log`):
- `@ersinciftci `
- `@oplantalech `
- `@alisman `
- `@inodb `

### DIFF
--- a/src/test/java/org/cbioportal/legacy/service/impl/DiscreteCopyNumberServiceImplTest.java
+++ b/src/test/java/org/cbioportal/legacy/service/impl/DiscreteCopyNumberServiceImplTest.java
@@ -200,7 +200,8 @@ public class DiscreteCopyNumberServiceImplTest extends BaseServiceImplTest {
   }
 
   @Test
-  public void getDiscreteCopyNumbersInMolecularProfileBySampleListIdHomdelOrAmpNoPagingNormalizesPageNumber()
+  public void
+      getDiscreteCopyNumbersInMolecularProfileBySampleListIdHomdelOrAmpNoPagingNormalizesPageNumber()
       throws Exception {
 
     createMolecularProfile();
@@ -471,7 +472,8 @@ public class DiscreteCopyNumberServiceImplTest extends BaseServiceImplTest {
   }
 
   @Test
-  public void fetchDiscreteCopyNumbersInMolecularProfileNonDiploidPagedUsesSqlPath() throws Exception {
+  public void fetchDiscreteCopyNumbersInMolecularProfileNonDiploidPagedUsesSqlPath()
+      throws Exception {
 
     createMolecularProfile();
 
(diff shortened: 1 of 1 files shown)


onursumer:
@VarshiniGunti Thanks for working on this. 

I don't think the main performance issue is the lack of pagination. Merging this PR won't really fix the existing memory spikes we are experiencing. To properly fix this issue, we will most likely need to refactor the legacy code and possibly reimplement some of the existing logic in SQL.

VarshiniGunti:
@onursumer Thanks for the feedback. I updated this PR to target the memory-spike issue more directly, not just response paging.

What changed in the latest commit (`a8624f5c9e`):
1. For single-profile discrete CNA endpoints, I now route all **non-diploid** requests (`-2, -1, 1, 2`) through the SQL/repository path (with deterministic DB paging), instead of only HOMDEL/AMP.
2. I keep **diploid-including** requests on the molecular-data fallback path for correctness (since the `sample_cna_event/cna_event` path does not include diploid records).
3. I updated service tests to cover:
   - non-diploid -> SQL path
   - non-diploid paged -> SQL path
   - diploid -> molecular-data fallback

So this should reduce in-memory work for a much larger set of discrete CNA queries while preserving existing behavior for diploid cases.

PR #12093, opened 2026-04-08
add clickhouse schema.sql and seed db files

sheridancbio:
This is an initial step towards consolidating all required scripts and resources needed for database setup and maintenance into the cbioportal repository. The database schema and seed can be used both for creating portal deployments and also for Integration / CI tests.

- add clickhouse base schema definition file for cbioportal database (schema.sql)
- add seed database (clickhouse equivalent of data in current seedDB from datahub) in parquet and concatenated SQLInsert  format

# Checks

# Any screenshots or GIFs?

# Notify reviewers


inodb:
Thanks! Looks good! Is there a PR or ticket describing what will happen with the current files on datahub? Will the source of truth move to this repo?

sheridancbio:
Thanks @inodb  ... We have been focussed predominantly on the rollout of rfc100 to the msk environment. When we are ready to roll this out to the community we will need to have addressed this issue (about the relocation of the seed database into the cbioportal repository). I plan to reach out to the curation team and others involved in seed database maintenance next week. I want to put together some documentation about how we will maintain future seed releases. I've not fully thought it through yet. My first idea is that we can just keep one seed in the cbioportal repository at the HEAD of the master branch. When a new seed is released it must be compatible with the schema, code, and importer --- they must all be kept in sync. If someone wants to retrieve an old seed for old code they can checkout that timepoint/commit-hash in the cbioportal repository and access the seed/code/importer which was at the HEAD there. We can keep a log of seed releases there too with commit hashes and dates for each released seed to make it easy to rewind to the proper spot.

The older mysql based seed probably need to stick around for users who are running old mysql based versions of cbioportal. If we want to remove them from datahub, we could. Or we could just leave them there with documentation/README updates telling users that the newer versions of cbioportal which are for clickhouse-only operation are not compatible with the mysql seed databases and that they are remaining here for legacy users of old versions of cbioportal. That probably the easiest route.

sheridancbio:
Since this introduces new resources which are not referenced anywhere in the cbioportal codebase currently, it is possible that we could merge this PR now. However, I think we should think about it a little first. The seed database is in path:
src/main/resources/db-scripts/clickhouse/init

Will this be packed (by maven) into the running executable of the portal? We probably don't want that, because this resource is not actually needed by the running backend ... just the unit tests etc. It would add needless bloat. We should think about this and (maybe) relocate or update our maven pom.xml files to specify that these files should be excluded when building the backend image. There may be a bunch of other things we can exclude too --- we don't really need "cgds.sql" for the deployed backend either but maybe we are packaging that too.

PR #12097, opened 2026-04-09
Changes to backend + derived table script required for RFC100 rollout

jamesqo:
PR https://github.com/cBioPortal/cbioportal/pull/12093 adds a unified schema for the core tables to cBioPortal. We will be using this schema for the rollout of the direct-to-Clickhouse RFC100 work.

The current derived table script modifies a few of the core tables to use primary keys -- which is now redundant, because we now specify them in the base schema. For the sample_cna_event and mutation table, these are just harmless no-ops. The genetic_alteration table also does not change its primary key -- however, it is being changed to use `MergeTree` instead of `ReplacingMergeTree` as its engine when the derived table script runs. This will break the rfc100 importer, which expects the genetic_alteration table to be backed by a `ReplacingMergeTree`.

This PR removes the now-redundant primary key additions, so that we are not modifying the core tables and the genetic_alteration table is kept as a ReplacingMergeTree. This should not require any changes to the cBioPortal backend to work with the new schema changes.

---------

**edit Apr 16** In addition, we needed to make some changes to the backend / derived table code in order to account for columns that are no longer nullable in the unified ClickHouse schema. There are cases where we were checking the result of a JOIN for NULL to determine whether or not there was a matching record -- however, this was always returning false for non-nullable columns. These queries have been updated to check for default values instead (eg. '' for String), and any API response endpoints that return the result of a JOIN have also been updated


sheridancbio (approved):
Looks good. We can start using this in production by resetting the importer reference branches to this branch ... but we should not merge this until we are committed to the rollout (to the community) of the direct-to-clickhouse updates to cbioportal-core, which will include the new "unified" schema of #12093 which inherently includes these ORDER BY clauses.

src/main/resources/db-scripts/cgds.sql
```diff
@@ -744,4 +744,4 @@ CREATE TABLE `resource_study` (
 
 -- DB_SCHEMA_VERSION AND DERIVED_TABLE_SCHEMA_VERSION MUST BE KEPT IN SYNC WITH THE db.version AND derived_table.version PROPERTIES IN pom.xml
 INSERT INTO `info` (`DB_SCHEMA_VERSION`, `GENESET_VERSION`, `DERIVED_TABLE_SCHEMA_VERSION`, `GENE_TABLE_VERSION`)
-  VALUES ('2.14.5', NULL, '1.0.9', NULL);
+  VALUES ('2.14.5', NULL, '1.0.10', NULL);
```
dippindots, line 747: @jamesqo I remember @sheridancbio wants to renaming the database file name, maybe that's in the other PR?

src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
```diff
@@ -191,12 +191,12 @@
             cd.attribute_name as attrId,
             cd.attribute_value as attrValue,
             cd.attribute_name as attributeName,
-            cam.display_name as displayName,
-            cam.description as description,
-            cam.datatype as datatype,
-            cam.patient_attribute as patientAttribute,
-            cam.priority as priority,
-            cam.cancer_study_id as cancerStudyId,
+            if(cam.cancer_study_id = 0, NULL, cam.display_name) as displayName,
```
dippindots, line 194: is it possible that `cam.cancer_study_id = 0`

src/main/resources/org/cbioportal/legacy/persistence/mybatis/GenePanelMapper.xml
```diff
@@ -12,12 +12,12 @@
...
 
     <sql id="selectGenePanelData">
-        gene_panel.stable_id AS "genePanelId",
+        nullIf(gene_panel.stable_id, '') AS "genePanelId",
         sample.stable_id AS "sampleId",
         patient.stable_id AS "patientId",
         cancer_study.cancer_study_identifier AS "studyId",
         genetic_profile.stable_id AS "molecularProfileId",
-        sample_profile.genetic_profile_id IS NOT NULL AS "profiled"
+        sample_profile.genetic_profile_id != 0 AS "profiled"
```
dippindots, line 20: we are setting `0` when the sample is not profiled?

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -67,7 +77,7 @@ SELECT
     concat(cs.cancer_study_identifier, '_', sample.stable_id) AS sample_unique_id,
     genetic_alteration_type AS alteration_type,
     -- If a mutation is found in a gene that is not in a gene panel we assume Whole Exome Sequencing WES
-    ifnull(gene_panel.stable_id, 'WES') AS gene_panel_id,
+    if(gene_panel.stable_id = '', 'WES', gene_panel.stable_id) AS gene_panel_id,
```
sheridancbio, line 80: Was this just a bug?  Good fix.

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -348,9 +358,9 @@ SELECT sm.internal_id             AS internal_id,
 FROM sample_derived AS sm
          INNER JOIN cancer_study AS cs
                     ON sm.cancer_study_identifier = cs.cancer_study_identifier
-         FULL OUTER JOIN clinical_attribute_meta AS cam
+         INNER JOIN clinical_attribute_meta AS cam
```
sheridancbio, line 361: usually we don't specify "INNER" .. because JOIN defaults to INNER when not otherwise specified. If we consistently (in the code base ... examine the cbioportal persistence repository mappers) use plain "JOIN" rather than "INNER JOIN" then let's not introduce the INNER specifier here. We should just be consistent if possible.

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -348,9 +358,9 @@ SELECT sm.internal_id             AS internal_id,
...
                          ON cs.cancer_study_id = cam.cancer_study_id
-         FULL OUTER JOIN clinical_sample AS csamp
+         LEFT JOIN clinical_sample AS csamp
```
sheridancbio, line 363: Here we are dropping the "OUTER" because "LEFT" implies "OUTER". So, similar to the previous comment , if we are going to infer "OUTER" we can also infer "INNER"

src/main/resources/db-scripts/clickhouse/clickhouse.sql
```diff
@@ -764,7 +682,7 @@ SELECT
     ge.stable_id AS entity_stable_id,
     ge.entity_type AS entity_type,
     CAST(
-        (groupArray(gep.name), groupArray(gep.value))
+        (groupArrayIf(gep.name, gep.id != 0), groupArrayIf(gep.value, gep.id != 0))
```
sheridancbio: The only way id will be zero / missing is if some error occurred during import. But this is still a good check. 👍

src/main/resources/db-scripts/cgds.sql
sheridancbio: Yes @dippindots  ... what we actually will do when merging the rfc100 work to master (with the full -rc branch completed) will be to delete db-scripts/cgds.sql.  ... because there will no longer by a mysql database or any reason to create one. I plan to merge this PR into an -rc branch which includes other fixes (like updates to documentation and other tools) where we will complete that PR. This PR should **not** be directly merged to master. It will be redirected to a new -rc branch.

src/main/resources/db-scripts/migration.sql
```diff
@@ -1087,5 +1087,5 @@ UPDATE `info` SET `DB_SCHEMA_VERSION`="2.14.4";
 ##version: 2.14.5
 -- reminder : always set both the current DB_SCHEMA_VERSION and the current DERIVED_TABLE_SCHEMA_VERSION at each update
 ALTER TABLE `info` ADD COLUMN `GENE_TABLE_VERSION` varchar(24);
-UPDATE `info` SET `DERIVED_TABLE_SCHEMA_VERSION`="1.0.9";
+UPDATE `info` SET `DERIVED_TABLE_SCHEMA_VERSION`="1.0.10";
```
sheridancbio, line 1090: @dippindots similarly, we will not be migration mysql databases after the merge of this work to master .. so migration.sql will no longer be needed.

src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
sheridancbio: @dippindots I'm not clear on this. I guess that every clinical_attribute_metadata records is associated with a known cancer_study. The cancer study it is associated with will have a non-zero cancer_study_id which should be set in this field. Hypothetically an importer bug might cause the field to be uninitialized.
Then ... maybe the cancer study is deleted but something breaks during the deletion of the cancer study and the clinical_attribute_metadata records connected to the study are accidentally **not** deleted. But even then, we should still have a non-zero (although invalid) cancer study id in this field. Basically this is a safeguard to cover the case of something going wrong which leads to a "null - equivalent" value being set for the cancer_study_id.
The mysql database had this constraint:
FOREIGN KEY (`CANCER_STUDY_ID`) REFERENCES `cancer_study` (`CANCER_STUDY_ID`) ON DELETE CASCADE
And now without that constraint we are (I think) checking for accidents/bugs which leave this value unset

src/main/resources/org/cbioportal/legacy/persistence/mybatis/GenePanelMapper.xml
sheridancbio: @dippindots My comment here is similar to the comment above ... this should never happen under normal circumstances. Every record in sample_profile must have a valid, non-0 sample_id specified and a valid non-0 genetic_profile_id specified. Only the `panel_id` field can truly be null (for WGS/WXS) or contain a real panel id.

The test "!= 0" is equivalent to "IS NOT NULL" now that the field genetic_profile_id is no longer specified as being Nullable(). An "empty" value in the output of an OUTER join source will default to the type default value when there is no associated record to join to. So the "could not join this record" representation in the output of the OUTER join will be "0" for this integer field. It can no longer be NULL.
(maybe this is a more accurate answer to the question above actually)

sheridancbio (commented):
Looked over changes again, and looks good.

Note : this PR should not be merged directly to the master branch. This should be folded into a "-rc" branch for the upcoming rfc100 release to the community ... and only merged then in conjunction with other PRs in repos {cbioportal-core, cbioportal-docker-compose, ...)

sheridancbio:
Reminder : do not merge this PR into master. The plan is to redirect this into a compose -rc branch which will be merged later

PR #12155, opened 2026-05-19
Second batch of updates to RFC100 docs

jamesqo:
cc @averyniceday @sheridancbio @n1zea144 

This covers all topics that we discussed Tuesday morning. It's broken down by commit for easier diffing

### DIFF
--- a/docs/data-loading/README.md
+++ b/docs/data-loading/README.md
@@ -38,6 +38,8 @@ cp -r /path/to/your_study ./study/
 docker compose exec cbioportal metaImport.py -s /study/your_study -o
 ```
 
+> **Note:** `-o` overrides validation warnings and proceeds with the import. If you are confident your data will pass all validation checks without warnings, you can drop `-o`.
+
 The `-s` flag here refers to the path of the study _as seen inside the Docker container_.
 
 ## Study Structure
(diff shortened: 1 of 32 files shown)


docs/data-loading/README.md
```diff
@@ -38,6 +38,8 @@ cp -r /path/to/your_study ./study/
 docker compose exec cbioportal metaImport.py -s /study/your_study -o
 ```
 
+> **Note:** `-o` overrides validation warnings and proceeds with the import. If you are confident your data will pass all validation checks without warnings, you can drop `-o`.
```
sheridancbio, line 41: maybe add a note that "any error encountered will prevent imports from proceeding"

docs/deployment/deploy-without-docker/Build-from-Source.md
```diff
@@ -0,0 +1,13 @@
+> *Note:* This documentation was written for an earlier version of cBioPortal. The information presented here may not apply to cBioPortal v7 and onwards.
```
sheridancbio, line 1: I'd prefer a stronger emphasis than a simple bolded "Note:"  I think making this a Heading would be fine .. or at least I would add an emoji or icon like:

:warning:

The words also could be stronger, like "With the v7 release of cBioPortal, this page of documentation is out of date and must be revised before relying on it."

sheridancbio (approved):

PR #12178, opened 2026-05-26
Exclude empty clinical attribute values from legacy clinical-data reads

alisman:
## Summary

Fixes the group-comparison **Clinical** tab bar chart rendering an **unlabeled legend entry** where an **"NA"** category should appear (visible on staging/v7 but not production/v6).

### Root cause
The v7 ClickHouse import stores blank clinical cells as empty-string (`''`) rows in `clinical_patient`/`clinical_sample`, whereas v6 omitted them entirely. The legacy `POST /api/clinical-data/fetch` (`ClinicalDataController` → `ClinicalDataMyBatisRepository` → `ClinicalDataMapper.xml`, reading those tables directly) therefore returns a full cartesian — one row per patient/sample × attribute — with `''` for missing values.

The comparison frontend (`groupComparison/ClinicalData.tsx`) synthesizes the **"NA"** legend category only for samples that have **no** returned row, so the new empty-string values become a distinct category with a blank legend label.

Confirmed live for comparison session `5ce411c7e4b0ab4137874076` (acc_tcga): staging (v7.0.0) returned 77 patients × 7 attrs = 539 rows with 93 empty values; production (v6.4.4) returned only actual values (446 rows, 0 empty). Both run ClickHouse, so this is a v6→v7 import-behavior difference, not MySQL-vs-ClickHouse.

### Fix (mapper-level, option 2)
Filter `attr_value != ''` in the legacy clinical-data reads via a guarded `excludeEmptyValues` flag, enabled for:
- the row-returning reads `getSampleClinicalData` / `getPatientClinicalData` (backing `fetchClinicalData`, `fetchAllClinicalDataInStudy`, `getAllClinicalDataInStudy`, `getAllClinicalDataOf{Sample,Patient}InStudy`), and
- their count companions `getMetaSampleClinicalData` / `getMetaPatientClinicalData`, so the paginated endpoints' `totalCount` stays consistent with the rows actually returned.

The flag is required because the shared `whereSample`/`wherePatient` fragments are also reused by the value-distribution counts and clinical-table queries — and `getVisibleSampleInternalIdsForClinicalTable` has no `clinical_sample` in its FROM — so an unconditional filter would alter those (or be a SQL error). Study-view NA counting (columnar `clinical_data_derived` path) is untouched.

### Testing
- Added an empty-value fixture entity to `clickhouse_data_legacy.sql` — `TCGA-EMPTY-VAL-PT` / `TCGA-EMPTY-VAL-PT-01` in `acc_tcga`, each with one non-empty (`OTHER_PATIENT_ID` / `OTHER_SAMPLE_ID`) and one empty (`FORM_COMPLETION_DATE` / `DAYS_TO_COLLECTION`) attribute.
- Added regression tests in `ClinicalDataMyBatisRepositoryTest` covering patient fetch, sample fetch, `fetchClinicalData`, and meta-count, asserting the empty values are excluded and the non-empty ones returned.
- Updated `PatientMyBatisRepositoryTest` global counts (18 → 19) for the added fixture patient.
- ✅ Verified: all **398 `*MyBatisRepositoryTest`** tests pass against a ClickHouse Testcontainer (JDK 21 / Maven).

### Follow-ups (not in this PR)
- The authoritative fix is in the importer (not this repo): don't load blank clinical cells at all.
- `getPatientClinicalDataDetailedToSample` (clinical-data downloads) still returns blanks if full consistency is desired.

🤖 Generated with <a href="https://claude.com/claude-code">Claude Code</a>


alisman:
@copilot can you redo this PR with the fix against latest master

PR #12186, opened 2026-06-02
Embedding Schema

i-am-leslie:
Fix N/A

- Introduced initial ClickHouse schema for embedding support
- Added `Embeddings` table to store embedding metadata (name, description, type, reduction technique)
- Added `Embedding_Data` table to store embedding coordinates and sample-level mappings
-  Defined primary ordering strategy for efficient lookup by embedding, study, and sample identifiers
- This is an initial draft schema intended for review and feedback before implementation

# Checks
- [x] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). We can fix this during merge by using a squash+merge if necessary

# Any screenshots or GIFs?

# Notify reviewers

### DIFF
--- a/src/main/resources/db-scripts/clickhouse/init/schema.sql
+++ b/src/main/resources/db-scripts/clickhouse/init/schema.sql
@@ -555,7 +555,7 @@ CREATE TABLE 'embeddings'(
                              'DESCRIPTION' String,
                              'EMBEDDING_TYPE' String,
                              'REDUCTION_TECHNIQUE' String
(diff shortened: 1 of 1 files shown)


hweej:
Also, there seems to be single quotes being used here to define table columns rather than backticks. This will probably break the script.

i-am-leslie:
I need to change the type for study_id from string to integer because a reference to the internal study id, which is an integer, will be used, and also the embedding table and embedding table will need an internal_id for smoother joins .

PR #12199, opened 2026-06-08
Stream generic-assay-meta/fetch to bound web-server memory

alisman:
## Problem

The column-store `/api/generic-assay-meta/fetch` endpoint (and the two `GET` variants) materialized the **entire** `List<GenericAssayMeta>` in memory — and cached it — before serializing the response.

Measured against a production-data ClickHouse clone, a methylation (`hm450`) request resolves to ~396K–835K entities, each an object holding a `HashMap` of properties — **hundreds of MB of heap** to produce a response that gzips to a few MB. The `@Cacheable` on the use case pinned a **second** full copy on the heap, keyed by the exact profile/stable-id set, so for the large, varied requests that caused the problem its hit rate was ~0 while the memory cost was full.

## Change

Switch the meta fetch to **MyBatis `ResultHandler` streaming**:

- The mapper methods stream each mapped row to a `ResultHandler` instead of returning a `List`.
- The domain `GenericAssayRepository` exposes a `Consumer<GenericAssayMeta>`; the ClickHouse impl bridges `ResultHandler` → `Consumer`.
- `GetGenericAssayMetaUseCase.execute(...)` pushes each entity to the consumer as it is read.
- The controller writes a JSON array element-by-element via a `JsonGenerator` inside the `StreamingResponseBody`, so the result set is **never collected into a list**.

The SQL and result maps are unchanged. The per-request `@Cacheable` is dropped on this path — it is incompatible with streaming (nothing is returned to cache) and provided little benefit for the large requests that motivated this work.

## Validation

Benchmarked old (master) vs new, both booted against a production-data ClickHouse clone, same 396,065-entity `hm450` request, 6 sequential runs each:

| | response | latency (median of 6) | peak old-gen heap growth |
|---|---|---|---|
| **OLD** (materialize + serialize) | 55 MB / 396,065 rows | ~6.0s | **+255 MB** |
| **NEW** (streaming) | 55 MB / 396,065 rows | ~13.5s | **+3–5 MB** |

- **~15× lower peak heap**, constant regardless of result size.
- Output is **byte-identical** to the old impl (`cmp` confirms; both 55,332,682 bytes).
- All **1,329 core tests pass** (0 failures/errors), including a real-ClickHouse Testcontainers mapper test exercising the `ResultHandler` path.

### Latency tradeoff (honest)

Streaming was ~2.2× slower wall-clock **in this benchmark** — but that number is inflated by the test setup, and the tradeoff is favorable:

- The ClickHouse query itself is **~0.7–0.9s** server-side (verified via `system.query_log`: 671/817/893 ms, ~175 MiB peak server memory). The rest of the measured latency is **transferring 55 MB from ClickHouse Cloud to the test box over the public internet + serialization** — not query execution. In production the app is co-located with ClickHouse (LAN), so absolute latencies are far lower and the gap shrinks.
- The slowdown is intrinsic to incremental consumption (per-row read cadence throttles network throughput vs. a tight drain-into-a-List loop). It is **not** fixable by `fetchSize` or output buffering — both were tested and made no difference.
- The gap only affects pathological large pulls, which are exactly the OOM-risk requests. Typical few-hundred-entity requests are sub-second either way.

If large-pull latency ever proves to matter in production, it can be recovered with a bounded read-ahead buffer (producer drains the cursor fast into a small bounded queue; consumer serializes) — deferred as out of scope here.

## Follow-up (the bigger lever): #12201

Benchmarking surfaced the real root cause: this endpoint returns **~396K entities / 55 MB regardless of how many studies are selected** (hm450 profiles share one probe set), and the frontend reads only `NAME`/`DESCRIPTION`/`URL` — ~80% of the property payload is keys it never reads. Reducing the payload (trim `SUMMARY` / have the frontend request less) would cut memory **and** latency more than streaming does. This PR bounds the memory; #12201 tracks shrinking the payload.

### DIFF
--- a/src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreGenericAssayController.java
+++ b/src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreGenericAssayController.java
@@ -128,6 +128,12 @@ public class ColumnStoreGenericAssayController {
     StreamingResponseBody body =
         outputStream -> {
           try (JsonGenerator generator = objectMapper.getFactory().createGenerator(outputStream)) {
+            // Once the 200 + first bytes are flushed the status can no longer change, so a
+            // mid-stream failure cannot become a 500 (inherent to streaming). Disable
+            // AUTO_CLOSE_JSON_CONTENT so that on failure the array is left unclosed: the client
+            // receives detectably-invalid JSON (parse error) rather than a silently-truncated but
+            // well-formed array, and the propagated exception is still logged server-side.
+            generator.disable(JsonGenerator.Feature.AUTO_CLOSE_JSON_CONTENT);
             generator.writeStartArray();
             getGenericAssayMetaUseCase.execute(
                 stableIds,
--- a/src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreGenericAssayController.java
+++ b/src/main/java/org/cbioportal/application/rest/vcolumnstore/ColumnStoreGenericAssayController.java
@@ -1,5 +1,6 @@
 package org.cbioportal.application.rest.vcolumnstore;
 
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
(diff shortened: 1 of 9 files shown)


alisman:
Thanks @copilot — both addressed in 66e2c39:

1. **Null element → 500 (`sorted()` NPE):** real catch. It's pre-existing on `master` (the `molecularProfileIds.stream().distinct().sorted()` line is unchanged from before this PR), but since the method is being reworked I've hardened it: `molecularProfileIds` (and `stableIds`) are now `filter(Objects::nonNull)`-ed before `distinct()`/`sorted()`, so a null element is treated as absent and the empty-check short-circuits instead of throwing. Added a regression test (`execute_profilesWithNullElement_filtersNullsWithoutThrowing`).

2. **Misleading Javadoc:** correct. Reworded so the "never materialized" claim is scoped to the SUMMARY/DETAILED meta-fetch paths, and the doc now notes the lightweight `ID` projection still resolves matching stable IDs into an in-memory set (no per-entity properties, no meta query).

Issue #12201, opened 2026-06-08
generic-assay-meta/fetch returns full probe set + unused properties regardless of study selection


alisman:
## Summary

`/api/generic-assay-meta/fetch` returns the **full entity set regardless of how many studies are selected, and includes per-entity properties the frontend never reads**. For methylation (`hm450`) this is the dominant cost driver behind the web-server memory/latency problems that PR #12199 (response streaming) only partially addresses.

Streaming bounds the *server memory*, but the **payload itself is the root cause** and reducing it would cut memory, latency, CPU, and bandwidth all at once — more than streaming does.

## Evidence (measured against a production-data ClickHouse clone)

- The 30 `*_tcga_pan_can_atlas_2018_methylation_hm450` profiles from a real study-summary request resolve to **396,065 distinct entities** — because every hm450 study shares the same methylation probe set. **Selecting 1 study or 30 returns the same ~396K entities / ~55 MB JSON.** The payload does not shrink with fewer studies.
- The DB query is cheap (~0.7–0.9s server-side, ~175 MiB peak ClickHouse memory). The cost is entirely in **moving and materializing the 55 MB result in the web server**.
- Of the ~66 MiB of property text across all generic-assay entities, **~80% is keys the frontend never reads.** Frontend usage (verified in `cbioportal-frontend`):
  - Reads: `stableId`, and properties `NAME`, `DESCRIPTION`, `URL`, `MUTATION_TYPE`, `MUTATION_CLASS`.
  - Never reads: `entityType`, `TRANSCRIPT_ID` (7.9 MiB), `DETAILS` (2.2 MiB), `GENE_SYMBOL` (2.0 MiB), `PHOSPHOSITE(S)`, `ACETYLSITES`.
  - For hm450 specifically, each probe carries `NAME` + the **unused** `TRANSCRIPT_ID`.
- No call site sends a `projection`, so everything comes back as the default `SUMMARY`. Today `SUMMARY` and `DETAILED` are identical for memory purposes — only `ID` is lightweight.

## Proposed work (any subset is a win)

1. **Trim `SUMMARY`** to the consumed keys (`NAME`, `DESCRIPTION`, `URL`, `MUTATION_TYPE`, `MUTATION_CLASS`) + drop `entityType`; keep the full map under `DETAILED`. The frontend reads properties via `getGenericAssayMetaPropertyOrDefault()` with fallbacks, so missing keys degrade gracefully.
[...]

PR #12202, opened 2026-06-08
Stream copy-number-segments, mutations, and clinical-data fetch to bound web-server memory

alisman:
## Problem

Several column-store-backed `/fetch` endpoints built the entire result `List` in memory (and the repository `@Cacheable`d it) before serializing. Production `query_log` shows these are the largest materializers:

| endpoint | peak result | notes |
|---|---|---|
| `/api/copy-number-segments/fetch` | **~8.3M rows / 1.34 GiB** | largest single result |
| `/api/mutations/fetch` | ~800K rows / ~470 MiB | ~600 calls / 6 days |
| `/api/clinical-data/fetch` | ~1.3M rows / ~430 MiB | thousands of calls / 6 days |

## Change

Applies the `ResultHandler` streaming pattern validated in #12199 (generic-assay-meta) to all three. For each, a new **additive** streaming method is threaded controller → service → repository → mapper, and the controller's `/fetch` endpoint returns `StreamingResponseBody`, writing a JSON array element-by-element via a `JsonGenerator`. The result set is never collected into a list; peak heap becomes bounded regardless of result size.

Per-endpoint notes:
- **copy-number-segments**: distinct `streamCopyNumberSegments` mapper statement (unpaged, ordered by seg_id).
- **mutations** / **clinical-data**: the streaming mapper method is a `ResultHandler` **overload** of the existing `getMutationsInMultipleMolecularProfiles` / `getSampleClinicalData` / `getPatientClinicalData` statements — the **exact same SQL** (projection/paging/sort) is reused with zero duplication.

In all cases:
- Existing `List`-returning methods and their `@Cacheable` are left intact for other callers; the streaming paths are **not** cached.
- `META` projection branches (count headers) are unchanged. Only the multi-study `/fetch` POSTs are converted; per-profile/single-study and paginated GET variants are untouched. **Output and API contract are identical.**

## Testing

- Reworked the three fetch controller tests to the async `StreamingResponseBody` flow (assertions unchanged — same JSON).
- Added real-ClickHouse (Testcontainers) streaming repo tests for all three, which also confirm MyBatis accepts the `ResultHandler` overloads.
- All affected tests pass (102); `spotless:check` clean.

## Context

Follow-up to #12199, which established the pattern (~15× lower peak heap, byte-identical output; latency on huge single requests is dominated by transfer/serialization, not the sub-second DB query). `gene-panel-data` and `discrete-copy-number` were intentionally excluded — they have service-layer transforms that aren't clean passthroughs and need separate design. The payload-reduction follow-up is tracked in #12201.

### DIFF
--- a/src/main/java/org/cbioportal/legacy/persistence/ClinicalDataRepository.java
+++ b/src/main/java/org/cbioportal/legacy/persistence/ClinicalDataRepository.java
@@ -1,6 +1,7 @@
 package org.cbioportal.legacy.persistence;
 
 import java.util.List;
+import java.util.function.Consumer;
 import org.cbioportal.legacy.model.ClinicalData;
 import org.cbioportal.legacy.model.ClinicalDataCount;
 import org.cbioportal.legacy.model.meta.BaseMeta;
@@ -87,6 +88,19 @@ public interface ClinicalDataRepository {
       List<String> attributeIds,
       String clinicalDataType,
       String projection);
+
+  /**
+   * Streaming counterpart of {@link #fetchClinicalData}: each datum is passed to {@code consumer}
+   * as it is read so the full (potentially very large) multi-study result is never held in memory.
+   * Not cached.
+   */
+  void streamClinicalData(
+      List<String> studyIds,
+      List<String> ids,
+      List<String> attributeIds,
+      String clinicalDataType,
+      String projection,
+      Consumer<ClinicalData> consumer);
 
   @Cacheable(
       cacheResolver = "generalRepositoryCacheResolver",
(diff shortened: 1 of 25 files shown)


alisman:
Pushed b6f6a72:
- **@copilot** test-comment nit: reworded the `streamCopyNumberSegments` repo-test comment — it asserts the streamed rows match the expected segments (same inputs/assertions as the `fetchCopyNumberSegments` test); it does not compare the two methods' outputs directly.
- **Mid-stream failure hardening** (carried over from #12199): a `StreamingResponseBody` commits the 200 before the body is written, so a mid-stream failure can't become a 500 — and Jackson's `AUTO_CLOSE_JSON_CONTENT` (default on) would otherwise write the closing `]` on the exception path, yielding a *silently* truncated but well-formed array. All three fetch endpoints now `disable(AUTO_CLOSE_JSON_CONTENT)`, so a failure leaves the array unclosed → the client gets detectably-invalid JSON and the exception still propagates/logs. Added a representative mid-stream-failure controller test.

PR #12212, opened 2026-06-12
refactor: migrate ServerStatusController from field injection to cons…

Pradyuman-aviator:
N/A — This is a Proof-of-Concept PR to start a discussion. A follow-up issue can be created to track the full migration of all ~150 legacy classes if the team approves this approach.

## Describe changes proposed in this pull request:
- removes `@Autowired` field injection from `ServerStatusController`
- Replaces it with constructor injection using a `private final` field
- Removes the now-unnecessary `import org.springframework.beans.factory.annotation.Autowired`
- no behavioral changes — this is a pure refactor with zero impact on functionality

## Why this change?

While reviewing the legacy codebase, I noticed approximately 150 classes in the `org.cbioportal.legacy` package still use `@Autowired` field injection. Modern Spring Boot (which this project uses at v3.5.x with Java 21) strongly discourages field injection in favor of constructor injection for the following reasons:

1. **Immutability:** `private final` fields cannot be accidentally set to `null` at runtime
2. **Easier unit testing:** Dependencies can be passed directly via constructor without loading the full Spring context
3. **Fail fast on circular dependencies:** Spring detects circular deps at startup instead of causing random runtime crashes

This PR migrates `ServerStatusController` as a **proof-of-concept**. It is the smallest and simplest controller in the legacy package (only one dependency), making it ideal for team review.

If the team approves this pattern, I am happy to migrate the remaining ~150 legacy classes in small, focused batches to avoid large merge conflicts.

## Checks
- [x] The commit log is comprehensible and follows the 7 rules of great commit messages
- [x] No test changes needed — this is a pure refactor with no logic changes. Constructor injection actually makes future unit tests easier to write without needing the full Spring context.
- [x] This PR does not add any logic based on clinical attributes
- [x] Label: `refactoring`

## Screenshots
N/A — No visual changes. This is a backend code quality refactor only.

## Notify reviewers
Tagging @inodb, @dippindots , and @onursumer for review, as they have been involved in the ongoing legacy package cleanup and architecture modernization efforts.


dippindots:
@Pradyuman-aviator I think this proposal makes sense. One thing worth mentioning is that the legacy package contains a lot of classes that we eventually plan to delete, so part of this work may end up being removed later. Because of that, I’m not sure this is the best time to do a large refactor there.

That said, if updating all ~150 files is relatively straightforward and low-risk, then I think it’s reasonable to go for it.

Pradyuman-aviator:
@dippindots  Thanks for the context... Given the legacy package risk, I'll keep the changes surgical  just mechanical find-and-replace style updates with no logic changes. If classes get deleted later, the diff will be trivially revertable. Happy to proceed on that basis.

dippindots:
@Pradyuman-aviator sounds good, please feel free to proceed

dippindots (approved):
Thanks for opening this poc pr!

Pradyuman-aviator:
Hey @dippindots! Thanks again for approving this  I've now opened the batch 1 PR 
following the pattern we discussed but but..

I also have a separate performance PR open that I'd love your eyes on when you get 
a chance  it fixes a 5-year-old TODO in DataBinHelper.calcCounts(), reducing the 
complexity from O(n × m) to O((n+m) log(n+m)). I noticed a reviewer was assigned 
but it hasn't been reviewed yet  no rush at all.... just wanted to make sure it didn't 
slip through! 🙂

PR link: https://github.com/cBioPortal/cbioportal/pull/12213

alisman:
@Pradyuman-aviator @dippindots i wonder whether it makes more sense to leave this code untouched and instead focus on migrating endpoints to the new codebase, where best practices will be implemented.   i.e. given that we know we want to throw about this code and have it reimplemented by claude, why should be bother to improve it and take any risk.

Pradyuman-aviator:
@alisman  yeah this is a fair point  if this legacy code is already planned 
for replacement, optimizing it adds complexity and risk for no long-term 
gain. Happy to close this PR and shift focus to something more impactful 
on the new stack instead. Let me know what would be most useful to work on.

PR #12226, opened 2026-06-17
Add ClickHouse HTTP request logging for endpoint QC

alisman:
## What

Adds an optional HTTP request logger that captures every matching `/api/**` request (method, path, query, headers, body, response status) and writes it to ClickHouse for QC of endpoint implementations. Each distinct request is captured once and can be replayed (e.g. as curl) and searched by matched route.

The earlier iterations of this branch targeted MongoDB; this changes the sink to **ClickHouse**, reusing the application's primary datasource instead of standing up a separate database with its own connection and credentials.

## How it works

- **Append-only + eventual dedup.** Every observation is one row in a `ReplacingMergeTree` ordered by `id` (a SHA-256 of method + path + query + body). Duplicate observations of the same logical request collapse during background merges; read the deduplicated set with `FINAL` / `GROUP BY id`.
- **Low write impact.** Inserts use ClickHouse `async_insert` so the many small per-request writes are batched server-side. Writes run on a bounded background thread pool with drop-on-full, so request handling is never blocked.
- **No auto-DDL / retention.** The table is provisioned out-of-band (`src/main/resources/db-scripts/clickhouse/request_log.sql`) and targeted via `request-logging.table`. No TTL — rows persist until explicitly deleted.
- **QC traceability columns:**
  - `git_commit` — the full commit the backend was built from (suffixed `-dirty` when the build had uncommitted changes), read straight from the classpath `git.properties`.
  - `server_name` — the host the request was addressed to, for deployments serving the API on multiple domains.
- **Redaction.** Sensitive headers (`authorization`/`cookie`/`set-cookie`) are redacted by default; configurable query/body param redaction is available via `request-logging.redact-params`.

## Off by default

The whole unit is gated on `@ConditionalOnProperty(request-logging.enabled=true)` with no `matchIfMissing`. With no config nothing is wired — no filter, no service, no writes, zero overhead. MongoDB remains wired for its other (legacy) uses; only the request-logging unit is migrated.

## Provisioning

Run `db-scripts/clickhouse/request_log.sql` (admin) to create `cbioportal_qc.logged_requests`, then grant the runtime datasource user `INSERT` (plus `SELECT`/`OPTIMIZE`) on it, and set `request-logging.enabled=true` + `request-logging.table=...`.

## Testing

- Unit tests for the capture/redaction filter pass.
- Verified end-to-end against a live ClickHouse Cloud instance: API requests captured with correct endpoint patterns, bodies, headers (as JSON), `server_name`, and `git_commit`; non-`/api` requests excluded; dedup behavior confirmed.

### DIFF
--- a/src/main/java/org/cbioportal/infrastructure/requestlog/HttpHeader.java
+++ b/src/main/java/org/cbioportal/infrastructure/requestlog/HttpHeader.java
@@ -4,7 +4,7 @@ package org.cbioportal.infrastructure.requestlog;
  * A single captured request header. The full set of headers for a request is serialized to JSON as
  * an array of these {@code {name, value}} objects and stored in a single column.
  *
- * <p>Headers are kept as a list of name/value pairs rather than a map so that duplicate header names
- * are preserved and attacker-supplied header names never need to become structural keys.
+ * <p>Headers are kept as a list of name/value pairs rather than a map so that duplicate header
+ * names are preserved and attacker-supplied header names never need to become structural keys.
  */
 public record HttpHeader(String name, String value) {}
--- a/src/main/java/org/cbioportal/infrastructure/requestlog/HttpHeader.java
+++ b/src/main/java/org/cbioportal/infrastructure/requestlog/HttpHeader.java
@@ -1,10 +1,10 @@
 package org.cbioportal.infrastructure.requestlog;
 
 /**
- * A single captured request header, stored as a {@code {name, value}} sub-document.
+ * A single captured request header. The full set of headers for a request is serialized to JSON as
+ * an array of these {@code {name, value}} objects and stored in a single column.
  *
- * <p>Headers are kept as a list of these rather than a {@code Map} so that attacker-supplied header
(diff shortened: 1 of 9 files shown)


alisman:
Addressed the SonarCloud bug and the Copilot review in 03c0280:

- **Sonar `java:S2583`** (`getCharacterEncoding() != null` always true) — charset is now resolved via `Optional.ofNullable(...).map(Charset::forName).orElse(UTF_8)`, removing the condition the analyzer flagged while keeping the UTF-8 fallback.
- **Copilot — `bodyTruncated` under-reported on chunked transfer** — when the cache hits the cap, an unknown content length (`-1`) is now treated as possibly truncated instead of always `false`.
- **Copilot — default `table` vs. provisioning script mismatch** — `request-logging.table` now defaults to `cbioportal_qc.logged_requests`, matching `db-scripts/clickhouse/request_log.sql`, so enabling the feature without overriding the table no longer targets the wrong DB.
- **Copilot — EXAMPLE points to "project docs" for DDL** — updated to reference the actual `request_log.sql` script and the corrected default table.

PR #12230, opened 2026-06-18
perf(api): restore memory cut on gene-panel-data/fetch and generic_assay_meta/fetch

alisman:
## What

Restores the memory optimization from #12210 (commit `3b00a2688a`), which was reverted in #12225 (`30831bfb21`).

This is a clean cherry-pick of the original optimization onto current `master`. The only merge conflict — in `GenePanelServiceImpl.java` — came from a SonarCloud field-rename that the revert PR had bundled in; those fields are **deleted** by this optimization, so the rename was moot and resolved in favor of the optimized code.

## The optimization (unchanged from #12210)

- **`gene-panel-data/fetch` (sampleMolecularIdentifiers path):** use the SQL-filtered mapper (`fetchGenePanelDataInMultipleMolecularProfiles[ByPatientIds]`) instead of fetching gene panel data for every sample of each requested profile and filtering in Java. The sequenced-sample-list profiled annotation for MUTATION profiles is preserved via the new `annotateBySequencedSampleLists` helper.
- **`generic_assay_meta/fetch` (SUMMARY projection):** restrict the generic-entity properties query to the names the frontend consumes (NAME/DESCRIPTION/URL) via an optional `propertyNames` SQL filter, instead of loading every property for every entity. DETAILED still returns the full map.

Original benchmarks (ClickHouse Cloud clone, `-Xmx2g`, N=24):
- gene-panel (10 samples of a 54k-sample profile): heap alloc/req 118 → 44 MB (−63%), peak heap 570 → 285 MB (−50%), latency 0.49 → 0.34s; response byte-identical.

## Verification

- `mvn -o compile` succeeds against the current `master` base.
- Net diff vs `master`: +366 / −43 across 14 files — the exact inverse of the revert.

### DIFF
--- a/src/e2e/js/src/types.ts
+++ b/src/e2e/js/src/types.ts
@@ -287,6 +287,44 @@ export interface ClinicalDataCount {
 }
 
 /**
+ * GenePanelData - Gene panel profiling status for one (sample, molecular profile) pair.
+ * Source: https://www.cbioportal.org/api/v3/api-docs/public (GenePanelData schema)
+ * Required fields: molecularProfileId, sampleId, patientId, studyId, profiled
+ */
+export interface GenePanelData {
+  /** Molecular profile stable id (e.g. "acc_tcga_mutations") */
+  molecularProfileId: string;
+  /** Sample stable id */
+  sampleId: string;
+  /** Patient stable id */
+  patientId: string;
+  /** Study stable id */
+  studyId: string;
+  /** Whether the sample is profiled in this molecular profile */
+  profiled: boolean;
+  /** Gene panel stable id; absent/null when whole-genome profiled */
+  genePanelId?: string;
+  /** Base64(sampleId + studyId) */
+  uniqueSampleKey: string;
+  /** Base64(patientId + studyId) */
+  uniquePatientKey: string;
+}
+
+/**
+ * GenericAssayMeta - Metadata for a generic assay entity.
+ * Source: https://www.cbioportal.org/api/v3/api-docs/public (GenericAssayMeta schema)
+ * Required fields: stableId
+ */
+export interface GenericAssayMeta {
+  /** Generic assay entity stable id */
+  stableId: string;
+  /** Entity type (e.g. "GENERIC_ASSAY") */
+  entityType: string;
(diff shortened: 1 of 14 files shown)


alisman:
## QC results — replayed real public-portal traffic against this branch vs `master`

**Method:** Built both jars (OLD = `master`@47cdb38 reverted, NEW = this branch), ran them sequentially against the same remote ClickHouse clone (`cbioportal_public_clone_2026_05_27`) with `-Xmx2g -XX:+UseParallelGC` + GC logging. Requests sourced from real logged traffic (`cbioportal_qc.logged_requests`). Optimized classes confirmed byte-different between jars.

### ✅ Correctness — PASS
**780/780 requests byte-identical** (canonicalized JSON), zero status/response mismatches, zero errors — across 200 heavy real requests + 100 sequential + 480 concurrent pathological requests.

### ✅ Latency — no regression
- Broad real traffic (200 reqs): p95 +0.3%, p99 −2.6% (network-bound, ~flat).
- Pathological shape under load: **~2× faster** (see below).

### ✅ Memory — confirmed, on the shape this PR targets
The win is shape-dependent: it shows up when a request asks for **a few samples from a profile that has many samples** (the old code fetched *all* samples per profile and filtered in Java). On broad traffic spread across ~1,900 modest profiles, memory is ~flat (expected). On the target shape — 10 samples from the 54,331-sample `msk_impact_50k_2026_mutations` profile (the PR's own benchmark) — it's dramatic:

**Sequential (110 reqs):** alloc/req **264 → 60 MB (−77%)**; latency p50 478 → 359 ms.

**Concurrent N=24 (480 reqs):**

| Metric | OLD | NEW | Δ |
|---|---|---|---|
| Peak heap | 1844 MB | 809 MB | **−56%** |
| Peak RSS | 2423 MB | 1163 MB | −52% |
| Alloc/req | 133.9 MB | 53.1 MB | −60% |
| Latency p50 | 1993 ms | 955 ms | −52% |
| Latency p95 | 2740 ms | 1332 ms | −51% |

Under concurrency OLD pressed against the 2 GB heap cap (1844/2048 MB) and GC-thrashed, doubling latency; NEW stayed at 809 MB. This reproduces the PR's headline **−50% peak heap** (measured −56%) and **0.49→0.34s** latency claims.

**Verdict:** parity perfect, no latency regression, claimed memory reduction reproduced. 🟢

PR #12235, opened 2026-06-21
fix: sanitize owner and users fields for anonymous virtual study sess…

Pradyuman-aviator:
N/A — resolves a TODO written in the codebase asking for this exact sanitization.

## Describe changes proposed in this pull request:
- Adds input sanitization for anonymous (unauthenticated) users creating virtual 
  study or group sessions
- Anonymous users could previously inject arbitrary `owner` and `users` field values 
  into the session store by including them in the request body — these values were 
  never cleared
- The fix adds an `else` branch to the existing `isAuthorized()` check that 
  explicitly strips both fields: `setOwner(null)` and `setUsers(Collections.emptySet())`
- Authenticated users are completely unaffected — the existing logic that sets 
  owner/users from the Spring Security context is unchanged
- Removes the resolved `// TODO sanitize what's supplied` comment from line 109

# Checks
- [x] The commit log is comprehensible and follows the 7 rules of great commit messages
- [x] No test added — the fix is a single guard clause in an existing branch. 
  A dedicated integration test for anonymous session sanitization can be tracked 
  as a follow-up if needed.
- [x] This PR does not add logic based on clinical attributes
- [x] Label: `bug`

# Any screenshots or GIFs?
N/A — no visual changes. Backend security fix only.

# Notify reviewers
Used `git blame` on `SessionServiceController.java` to identify authors. 
Tagging @onursumer and @inodb for review as they have the most context on the 
session service layer.

### DIFF
--- a/src/main/java/org/cbioportal/legacy/web/SessionServiceController.java
+++ b/src/main/java/org/cbioportal/legacy/web/SessionServiceController.java
@@ -146,6 +146,10 @@ public class SessionServiceController {
         if (isAuthorized()) {
           customData.setOwner(userName());
           customData.setUsers(Collections.singleton(userName()));
+        } else {
+          // Sanitize: anonymous users must not inject arbitrary owner or users values.
+          customData.setOwner(null);
+          customData.setUsers(Collections.emptySet());
         }
 
         // use basic authentication for session service if set
--- a/src/main/java/org/cbioportal/legacy/web/SessionServiceController.java
+++ b/src/main/java/org/cbioportal/legacy/web/SessionServiceController.java
@@ -106,7 +106,6 @@ public class SessionServiceController {
         // JSON from file to Object
         VirtualStudyData virtualStudyData =
             sessionServiceObjectMapper.readValue(body.toString(), VirtualStudyData.class);
-        // TODO sanitize what's supplied. e.g. anonymous user should not specify the users field!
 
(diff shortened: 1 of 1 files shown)


Pradyuman-aviator:
@inodb  if you got time have look at these approach Thanks!!! eager to contribute more

Pradyuman-aviator:
Also extends the sanitization to the `custom_data` session type, which 
had the identical anonymous injection gap  anonymous users could POST 
to `/{type}` with `type=custom_data` and inject arbitrary owner/users 
fields into the session store.

`settings` and `custom_gene_list` are unaffected  both already return 
503 for unauthenticated requests at the endpoint level.

dippindots (commented):
Currently, both `VirtualStudyData` and `CustomAttributeWithData` treat "anonymous" as the default invariant. We can use "anonymous" instead of null, which still removes any attacker-controlled ownership information while preserving existing assumptions and behavior.

Pradyuman-aviator:
@dippindots  changing  virtualStudyData.setOwner(null); -----> virtualStudyData.setOwner(anonymous);  &  customData.setOwner(null);  ---> 
   customData.setOwner("anonymous");

Pradyuman-aviator:
Thanks for the Review made the changes you suggested

dippindots (approved):

PR #12248, opened 2026-06-25
Benchmark: no-explode import (direct exploded load vs ARRAY JOIN derive)

alisman:
## What

Adds a self-contained benchmark under `benchmarks/no-explode-import/` evaluating an alternative import strategy for gene×sample matrix profiles: have the importer write the **exploded** `(sample, gene, profile, value)` rows directly, instead of storing a packed `genetic_alteration.values` string and rebuilding `genetic_alteration_derived` with a server-side `ARRAY JOIN` derive.

The harness loads a real datahub study (`ccle_broad_2019` — 89M cells across a discrete CNA matrix and a continuous RPKM matrix) **two ways** into ClickHouse and compares correctness, latency, server memory, and storage.

## Findings

- **Correctness ✅** — the directly-loaded exploded table is **byte-identical** (matching `cityHash64` checksums) to the legacy `ARRAY JOIN` derive, for both discrete and continuous profiles.
- **Storage** — no-explode persists **~38% less** (198 MiB vs 319 MiB): it drops the redundant packed copy (123 MiB), which compresses poorly for continuous data.
- **Server memory** — the `ARRAY JOIN` derive peaked at **1.85 GiB (CNA) / 4.78 GiB (RPKM) for a single study**; the direct load has no such spike (insert-level). This is the multi-GiB OOM driver at full-corpus scale.
- **Latency** — direct is ~2.4× slower wall-clock **for a remote client**, because it ships the large exploded rows over the network while the derive ships tiny packed data and explodes server-side. Co-located with ClickHouse this gap collapses.
- **Caveat** — the legacy baseline here is a *scoped per-study* derive. Production `metaImport.py` fully rebuilds the **entire** derived corpus on every import, so against the real baseline the no-explode strategy wins decisively (seconds-per-study vs. a full ~11.8B-row rebuild).

## Contents

- `BENCHMARK.md` — full method, numbers, and interpretation
- `ddl.sql` — benchmark table definitions
- `run_benchmark.sh` — reproducible harness (downloads the study, runs both pipelines, prints metrics)

This is a benchmark/investigation only — no production code paths are changed.

### DIFF
--- a/benchmarks/no-explode-import/BENCHMARK.md
+++ b/benchmarks/no-explode-import/BENCHMARK.md
@@ -1,0 +1,78 @@
+# No-explode import benchmark: direct exploded load vs. packed + `ARRAY JOIN` derive
+
+## Question
+
+cBioPortal stores gene×sample matrix profiles (CNA, expression, …) in a **packed** base
+table `genetic_alteration` (one row per gene, a comma-joined `values` string), then builds
+the denormalized `genetic_alteration_derived` (one row per gene×sample) with a server-side
+`ARRAY JOIN` **derive** step that is *fully rebuilt from scratch on every import*.
+
+Can we instead have the importer write the **exploded** rows directly — no packed `values`,
+no `ARRAY JOIN`, no derive step — and is that strategy correct and faster/cheaper?
+
+## Method
+
+- **Source:** a real datahub study, **`ccle_broad_2019`**, two genuine matrix files:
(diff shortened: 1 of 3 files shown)


alisman:
Closing: this harness used clickhouse-local/native as a proxy, not the actual cBioPortal importer. Redoing against the real cbioportal-core importer with no-explode implemented as an optional mode.

PR #12251, opened 2026-06-26
Support dash-separated numeric ranges in ClinicalDataIntervalFilter

### LINKED ISSUE #12250: [BUG] ClinicalDataIntervalFilter does not support dash-separated numeric ranges (e.g. `20-30`)
While reviewing the interval filter implementation, I noticed that `ClinicalDataIntervalFilterApplier` contains the following TODO:

```java
// invalid range -- TODO: also support ranges like 20-30?
```

This suggests that dash-separated numeric ranges are currently treated as invalid rather than being parsed as an inclusive interval.

If this parser is used by the current clinical filtering workflow, users attempting to filter numeric clinical attributes using syntax such as `20-30` would not receive the expected results.

## Expected Behavior

A numeric interval entered as:

```
20-30
```

should be interpreted as:

```
value >= 20 && value <= 30
```

and return records whose values fall within the specified range.

## Actual Behavior

Dash-separated numeric ranges appear to be treated as invalid and are not parsed into an interval.

## Steps to Reproduce

1. Open a study containing numeric clinical attributes.
2. Navigate to a page that supports clinical data filtering.
3. Select a numeric clinical attribute (for example, Age or Overall Survival).
4. Enter a filter value using the format:

```
20-30
```

5. Apply the filter.

## Possible Cause

The current implementation contains the following TODO:

```java
// invalid range -- TODO: also support ranges like 20-30?
```

which indicates that this input format is not currently handled.

## Proposed Fix

Extend the interval parsing logic to recognize dash-separated numeric ranges (`min-max`) and convert them into an inclusive interval before applying the filter.

## Environment

* cBioPortal: latest `master`
* Browser: N/A (backend parsing logic)
* OS: Windows 11 (development environment)

Debajeet-1411:
fixes #12250

## Summary

This PR fixes the handling of clinical attribute values stored as dash-separated numeric ranges (e.g. `20-30` or `4.5 - 10.5`) in the backend interval parsing logic.

Previously, when a clinical attribute contained a range, the parser attempted to convert the entire string directly into a `BigDecimal`, causing a `NumberFormatException`. As a result, these values were treated as invalid and excluded from interval-based filtering.

This PR updates the parsing logic to correctly recognize and interpret dash-separated numeric ranges while preserving the existing behavior for single numeric values.

## Changes

* Updated `ClinicalDataIntervalFilterApplier` to detect and parse dash-separated numeric ranges.
* Updated `CustomDataFilterUtil` with the same parsing logic to keep interval handling consistent.
* Safely parses both bounds of a numeric range after trimming whitespace.
* Preserves support for single numeric values and negative numbers.
* Gracefully handles malformed ranges by falling back to the existing behavior.
* Removed the outdated TODO comments after implementing the missing functionality.

### Before

Clinical attribute value:

```text
20-30
```

Result:

* Parsing attempted to convert `"20-30"` directly into a `BigDecimal`.
* A `NumberFormatException` occurred.
* The value was treated as invalid and excluded from interval-based filtering.

### After

Clinical attribute value:

```text
20-30
```

Result:

* Parsed as the interval `[20, 30]`.
* Interval comparisons are performed correctly during filtering.

## Testing

Verified that:

* Single numeric values continue to behave as before.
* Dash-separated numeric ranges are parsed successfully.
* Negative numbers are not incorrectly interpreted as ranges.
* Invalid range values continue to be handled gracefully.
* Existing tests pass (`StudyViewFilterApplierTest`).

## Impact

This change ensures that patients with clinical attributes recorded as numeric ranges are no longer silently excluded from interval-based filtering. Researchers querying studies containing imprecise clinical measurements (for example, `20-30` or `4.5 - 10.5`) will now obtain results that correctly reflect the stored data.

### DIFF
--- a/src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
+++ b/src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
@@ -98,10 +98,21 @@ public class ClinicalDataIntervalFilterApplier extends ClinicalDataFilterApplier
         min = new BigDecimal(value.substring(gt.length()));
         startInclusive = false;
       } else {
-        min = max = new BigDecimal(attrValue);
+        int dashIndex = value.indexOf('-', 1); // skip the first character to allow negative numbers
+        if (dashIndex > 0) {
+          try {
+            min = new BigDecimal(value.substring(0, dashIndex).trim());
+            max = new BigDecimal(value.substring(dashIndex + 1).trim());
+          } catch (NumberFormatException nfe) {
+            min = max = new BigDecimal(value);
+          }
+        } else {
+          min = max = new BigDecimal(value);
+        }
       }
     } catch (Exception e) {
-      // invalid range -- TODO: also support ranges like 20-30?
+      // invalid range
+
       return null;
     }
 
--- a/src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
+++ b/src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
@@ -17,6 +17,11 @@ import org.springframework.stereotype.Component;
 @Component
 public class ClinicalDataIntervalFilterApplier extends ClinicalDataFilterApplier {
 
+  private static final String LTE = "<=";
+  private static final String LT = "<";
+  private static final String GTE = ">=";
+  private static final String GT = ">";
+
   @Autowired
   public ClinicalDataIntervalFilterApplier(
       PatientService patientService,
@@ -78,24 +83,19 @@ public class ClinicalDataIntervalFilterApplier extends ClinicalDataFilterApplier
 
     String value = attrValue.trim();
 
-    String lte = "<=";
-    String lt = "<";
-    String gte = ">=";
-    String gt = ">";
-
     boolean startInclusive = true;
     boolean endInclusive = true;
 
     try {
-      if (value.startsWith(lte)) {
-        max = new BigDecimal(value.substring(lte.length()));
-      } else if (value.startsWith(lt)) {
-        max = new BigDecimal(value.substring(lt.length()));
+      if (value.startsWith(LTE)) {
+        max = new BigDecimal(value.substring(LTE.length()));
+      } else if (value.startsWith(LT)) {
+        max = new BigDecimal(value.substring(LT.length()));
         endInclusive = false;
-      } else if (value.startsWith(gte)) {
-        min = new BigDecimal(value.substring(gte.length()));
-      } else if (value.startsWith(gt)) {
-        min = new BigDecimal(value.substring(gt.length()));
+      } else if (value.startsWith(GTE)) {
+        min = new BigDecimal(value.substring(GTE.length()));
+      } else if (value.startsWith(GT)) {
+        min = new BigDecimal(value.substring(GT.length()));
         startInclusive = false;
       } else {
         int dashIndex = value.indexOf('-', 1); // skip the first character to allow negative numbers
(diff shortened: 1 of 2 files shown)


src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
```diff
@@ -98,10 +98,21 @@ private Range<BigDecimal> calculateRangeValueForAttr(String attrValue) {
...
+            min = new BigDecimal(value.substring(0, dashIndex).trim());
+            max = new BigDecimal(value.substring(dashIndex + 1).trim());
+          } catch (NumberFormatException nfe) {
+            min = max = new BigDecimal(value);
+          }
+        } else {
+          min = max = new BigDecimal(value);
+        }
       }
     } catch (Exception e) {
```
harshkha3417: catch (NumberFormatException e) {
    return null;
} 
if later someone accidentally introduces NULLPOINTEREXCEPTION or ILLEGALARGUMENTEXCEPTION the current code hides this bug.Catching NUMBERFORMATEXCEPTION is more precise

src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
harshkha3417: private static final String lte = "<=";
private static final String lt = "<";
private static final String gte=">=";
private static final String gt=">";
as they are constants.

harshkha3417 (approved):
One suggestion: the interval parsing logic (Range<BigDecimal> calculateRangeValueForAttr(String attrValue))now exists in both CustomDataFilterUtil.java and ClinicalDataIntervalFilterApplier.java,would it make sense to extract the parsing into a shared utility (e.g. ClinicalRangeParser)? That would provide a single source of truth for interval parsing and make future maintenance and testing easier.

src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
Debajeet-1411: Done, extracted the operator strings into class-level constants. Thanks!

src/main/java/org/cbioportal/legacy/web/util/ClinicalDataIntervalFilterApplier.java
Debajeet-1411: Thanks! I've updated it to catch NumberFormatException instead of the broader Exception.

Debajeet-1411:
@harshkha3417 Thanks for the suggestion! I agree that having a shared parser would improve maintainability. Since this PR is focused on fixing the interval parsing bug, I'd prefer to keep its scope limited. If you think it would be better to include the refactoring in this PR, I'm happy to update it accordingly.

harshkha3417:
> @harshkha3417 Thanks for the suggestion! I agree that having a shared parser would improve maintainability. Since this PR is focused on fixing the interval parsing bug, I'd prefer to keep its scope limited. If you think it would be better to include the refactoring in this PR, I'm happy to update it accordingly.

@Debajeet-1411 That makes sense. Keeping the bug fix focused is reasonable. Extracting the shared parsing logic can be done in a follow-up refactoring PR to avoid expanding the scope of this change. Thanks for considering it!

Pradyuman-aviator (commented):
Hi @Debajeet-1411 — thanks for the contribution, the idea of supporting dash separated ranges is actually  useful improvement!

However, I noticed that CustomDataFilterUtil (line 27 in the diff) is annotated with @Deprecated(forRemoval = true), meaning this class is planned for removal as part of the ongoing migration to the new column store stack. Adding new logic here would be investing effort into code that will eventually be deleted.

It might be worth checking with the maintainers whether this feature should instead be implemented in the equivalent non deprecated path, or if there's a tracking issue for the new stack where this would fit better. Happy to help investigate if needed!

onursumer:
This is a deprecated legacy utility class which we should probably avoid modifying unless there is a need for a critical fix. I think the current functionality works fine for our use-cases.

PR #12272, opened 2026-07-13
Clickhouse migration mechanism

dippindots:
## Summary

Replaces the dead MySQL-era migration system (`migrate_db.py` / `migration.sql`), which has had
no working upgrade path since the ClickHouse cutover, with a ClickHouse-native equivalent. This PR
ships the mechanism only — no other schema content changes beyond the `3.0.0` cutover.

- New `db-scripts/clickhouse/migrate/migrate_schema.sql` (versioned, forward-only SQL sections)
  applied by a new `db-scripts/clickhouse/migrate/migrate_db.py` runner
- `db_schema_version` → `3.0.0`. There is a single version number covering both base and derived
  tables — no separate `derived_table_schema_version` track (see below)
- Renamed `clickhouse.sql` → `generate_derived_tables.sql` → `populate_derived_tables.sql`.
  Derived table *structure* now lives in `schema.sql` alongside every other table, versioned like
  everything else; this script is data population only (`TRUNCATE` + repopulate), has no version
  of its own, and is safe to run any time
- Removed legacy MySQL `cgds.sql` / `migration.sql`
- `migrate_db.py` advances `info.db_schema_version` itself after a section's SQL succeeds —
  section authors don't hand-write that line
- `migrate_db.py --populate-derived-tables` (off by default) repopulates derived tables whenever a
  run actually applies one or more pending migrations
- ClickHouse credentials are written to a temp `clickhouse client --config-file` (mode `0600`,
  deleted after the run) instead of passed as a subprocess argv, which would otherwise be visible
  to other users on the host via `ps aux`
- Rewrote `test/test_db_version.sh` to validate the ClickHouse schema files instead of the removed
  MySQL ones — this caught a real pre-existing drift on `master`: `schema.sql`'s
  `derived_table_schema_version` was `1.0.9` while `clickhouse.sql`'s header said `1.0.11`
- Added `SchemaVersionChecker`, a startup check that refuses to start the backend against a
  mismatched `db_schema_version` (or a missing one) unless
  `db.suppress_schema_version_mismatch_errors=true`
- Updated deployment docs (§7 derived tables, §11 Version Migration) describing when derived
  tables rebuild across Docker Compose vs. manual deployment, and the upgrade paths for each

Companion PRs (same branch name, `clickhouse-migration-mechanism`):
- cbioportal-docker-compose — wires `migrate_db.py` into `docker compose up`
- cbioportal-core — bumps `db.version`, fixes a hardcoded `clickhouse.sql`/`generate_derived_tables.sql`
  reference in `rebuild_derived_tables.py` that this rename would otherwise have broken

## Test plan

- [x] `mvn compile` / `mvn test-compile` pass
- [x] `test/test_db_version.sh` fails against current `master` (proves it catches the real
      `1.0.9`/`1.0.11` drift) and passes on this branch
- [x] `migrate_schema.sql` section parsing verified standalone against `migrate_db.py`
- [x] `migrate_db.py` exercised end-to-end against a stub `clickhouse client` (no ClickHouse
      available locally): base migration apply, idempotent no-op on re-run, and
      `--populate-derived-tables` triggering correctly
- [x] Every derived-table `CREATE TABLE`/`INSERT`/`OPTIMIZE` statement cross-diffed
      programmatically between the old single file and the new `schema.sql` /
      `populate_derived_tables.sql` split, to confirm byte-for-byte preservation across the move
- [ ] End-to-end `docker compose up` migration run (see companion `cbioportal-docker-compose` PR —
      not verified in this environment, needs a real Docker run before merge)
- [ ] Backend startup refusal on mismatched schema version, verified against a live ClickHouse
      instance


src/main/java/org/cbioportal/SchemaVersionChecker.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+@Component
+public class SchemaVersionChecker implements ApplicationRunner {
+
+  private static final Logger logger = LoggerFactory.getLogger(SchemaVersionChecker.class);
+
+  @Autowired private InfoService infoService;
+
+  @Value("${clickhouse_mode:false}")
+  private String clickhouseMode;
```
sheridancbio: Maybe outside the scope of this PR, but this setting is now utterly obsolete and should be removed from the codebase.

src/main/java/org/cbioportal/SchemaVersionChecker.java
```diff
@@ -0,0 +1,79 @@
...
+ * Refuses to let the application start against a ClickHouse database whose db_schema_version
+ * doesn't match the version this build expects (db.version, from pom.xml), since running with a
+ * mismatched schema silently produces wrong queries/results rather than a clear error.
+ *
+ * <p>Run the migration tool (db-scripts/clickhouse/migrate/migrate_db.py) to bring the database up
+ * to date, or set db.suppress_schema_version_mismatch_errors=true to downgrade this to a warning
+ * (e.g. when testing a feature unrelated to the schema change in the expected version).
+ */
+@Component
+public class SchemaVersionChecker implements ApplicationRunner {
```
sheridancbio, line 23: Spring boot can accommodate multiple classes implementing the ApplicationRunner interface. All such classes will have their run() method executed during application startup. If only one were allowed, I would have objected to making the one and only applicationrunner be called "SchemaVersionChecker". But since many are allowed, this should be no problem.

src/main/java/org/cbioportal/SchemaVersionChecker.java
```diff
@@ -0,0 +1,79 @@
...
+  @Value("${db.version}")
+  private String expectedDbSchemaVersion;
+
+  @Value("${db.suppress_schema_version_mismatch_errors:false}")
+  private boolean suppressSchemaVersionMismatchErrors;
+
+  @Override
+  public void run(ApplicationArguments args) {
+    if (!Boolean.parseBoolean(clickhouseMode)) {
+      return;
```
sheridancbio: I think that if the user has set clickhouse_mode to false we should simply cancel the startup of the application. (SpringApplication.exit())

But it would be better for us to simply remove this property/setting from the codebase.

src/main/java/org/cbioportal/SchemaVersionChecker.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    InfoDb infoDb = infoService.getInfoFromDb();
+    if (infoDb == null || infoDb.getDbSchemaVersion() == null) {
+      logger.warn(
+          "Could not read db_schema_version from the database; skipping schema version check.");
```
sheridancbio: Should we be stronger here and require that info.db_schema_version be set? Perhaps if db.suppress_schema_version_mismatch_errors is true we can skip the check, but if it is not set perhaps we should log the problem and then fail the check / exit the application.

src/main/java/org/cbioportal/SchemaVersionChecker.java
```diff
@@ -0,0 +1,79 @@
...
+      return;
+    }
+
+    String message =
+        String.format(
+            "Database schema version mismatch: the database is at db_schema_version '%s', but "
+                + "this cBioPortal build expects '%s'. Run "
+                + "db-scripts/clickhouse/migrate/migrate_db.py against your database to upgrade "
+                + "it, or downgrade the cBioPortal build to a version compatible with '%s'. To "
+                + "start anyway, set db.suppress_schema_version_mismatch_errors=true.",
```
sheridancbio: I'd add a caution note to the "stat anyway" part .. perhaps : "to allow application to start despite the database version mismatch (which can result in silent errors and data distortion)"

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -0,0 +1,237 @@
...
+RED = '\033[91m'
+GREEN = '\033[92m'
+END = '\033[0m'
+
+SECTION_HEADER_RE = re.compile(r'^##\s*db_schema_version:\s*(\S+)\s*$')
+DESCRIPTION_RE = re.compile(r'^##\s*description:\s*(.*)$')
+CUSTOM_RE = re.compile(r'^##\s*custom:\s*true\s*$', re.IGNORECASE)
+
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+DEFAULT_MIGRATE_SCHEMA_SQL = os.path.join(SCRIPT_DIR, 'migrate_schema.sql')
```
sheridancbio, line 42: Ok ... by default we expect migrate_schema.sql to be in the same directory as the python script .. but we can override with a command line argument. This seems ok.

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -0,0 +1,237 @@
...
+    return ch_props
+
+
+def _base_cmd(ch_props):
+    return [
+        'clickhouse', 'client',
+        '--host', ch_props['host'],
+        '--port', ch_props['port'],
+        '--user', ch_props['user'],
+        '--password', ch_props['password'],
```
sheridancbio: This is a security risk. It converts the credentials from environment variables (which have some risks but are better than command line arguments) into command line arguments which are passed to a subprocess (forked). Any user on the machine which is executing this command can use "ps aux" to see the credentials.

A better approach would be to write the needed credentials into a temporary file, formatted for use with `clickhouse client`. A file named clickhouse_client_migrate_db_temp_config.properties with this content:
```
user: value_from_ch_props['user']
password: value_from_ch_props['password']
host: value_from_ch_props['host']
port: value_from_ch_props['port']
database: value_from_ch_props['database']
```
could then be used like this:
`clickhouse client --config-file clickhouse_client_migrate_db_temp_config.properties`
as the base_cmd.

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -0,0 +1,237 @@
...
+            "clickhouse client not found. Install it with:\n"
+            "  curl https://clickhouse.com/install | sh"
+        )
+    if result.returncode != 0:
+        raise RuntimeError(f"clickhouse client failed (exit {result.returncode}):\n{result.stderr}")
+    return result.stdout
+
+
+def get_current_db_schema_version(ch_props):
+    version = run_query(ch_props, "SELECT db_schema_version FROM info LIMIT 1")
```
sheridancbio, line 189: interesting. What should we do if there are multiple records in the info table I wonder?  A question for another day I guess.

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -0,0 +1,237 @@
...
+
+    ALTER TABLE ... UPDATE/DELETE are async mutations in ClickHouse — a section isn't
+    actually done just because the client returned.
+    """
+    deadline = time.time() + timeout_secs
+    while time.time() < deadline:
+        pending = run_query(
+            ch_props,
+            "SELECT count() FROM system.mutations "
+            f"WHERE database = '{ch_props['database']}' AND is_done = 0",
```
sheridancbio, line 210: oh .. this is really helpful. I am working on a race condition in the importer where this might be the solution we need. 👍

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -0,0 +1,237 @@
...
+        time.sleep(poll_interval_secs)
+    raise TimeoutError(f"Mutations did not complete within {timeout_secs}s")
+
+
+def apply_section(ch_props, section):
+    print(f"Applying db_schema_version {section.version}: {section.description}")
+    if section.sql:
+        run_multiquery(ch_props, section.sql)
+        wait_for_mutations(ch_props)
+    if section.custom:
```
sheridancbio: This might be worth documenting ... that when a section has both a SQL block in migrate_schema.sql, and a custom code function in this script, we apply first the SQL block, *and then* the custom code is executed.

src/main/resources/db-scripts/clickhouse/migrate/migrate_schema.sql
```diff
@@ -0,0 +1,31 @@
...
+-- Format: one section per db_schema_version, applied strictly in ascending order by
+-- migrate_db.py (see migrate_db.py in this directory for the runner).
+--
+--   ## db_schema_version: <version>
+--   ## description: <one line>
+--   ## custom: true            (optional — see below)
+--   <SQL statements for this section>
+--
+-- Rules for writing a section:
+--   1. Sections must be idempotent (IF EXISTS / IF NOT EXISTS everywhere) — ClickHouse has no
```
sheridancbio: I'm not sure if this actually will always be possible ... but it is good to recommend it.

src/main/resources/db-scripts/clickhouse/migrate/migrate_schema.sql
```diff
@@ -0,0 +1,31 @@
...
+--   3. For ORDER BY / primary-key changes: create a new table, INSERT ... SELECT, RENAME —
+--      ClickHouse cannot ALTER these in place.
+--   4. Every section must end by advancing db_schema_version, e.g.:
+--        ALTER TABLE info UPDATE db_schema_version = '<version>' WHERE 1;
+--      migrate_db.py waits for this mutation (and any others the section triggered) to finish
+--      before treating the section as applied.
+--   5. Mark a section '## custom: true' when it needs logic beyond raw SQL (e.g. data reshaping
+--      that must be computed outside ClickHouse). migrate_db.py runs this section's SQL (if any)
+--      first, then calls the matching hardcoded Python function registered in
+--      migrate_db.py's CUSTOM_MIGRATIONS dict, keyed by this section's version.
```
sheridancbio: Ok .. this documents the point I mentioned above (SQL first, then custom function). 👍

src/main/resources/db-scripts/clickhouse/generate_derived_tables.sql
```diff
@@ -1,7 +1,8 @@
--- version 1.0.11 of derived table schema and data definition
+-- version 2.0.0 of derived table schema and data definition
```
sheridancbio: future work : we need a mechanism to keep the info table in sync with this version number (info.derived_table_schema_version). But maybe we will be engineering away from derived tables being a separate step soon and they will become part of the base table schema mechanism.

The easy way to accomplish this would be to have generate_derived_tables.sql set the version number in info.

It is also important that the construct_derived_tables.sql version the user is running is actually compatible with the base table schema (db_schema_version). We don't check for that, but if a user is using an "old" version of cbioportal, they should be using an "old" version of generate_derived_tables.sql.

test/test_db_version.sh
```diff
@@ -72,42 +72,48 @@ function find_delimited_substrings() {
...
+if [ "$pom_db_version" == "$schema_sql_db_version" ] &&
+        [ "$schema_sql_db_version" == "$migrate_schema_sql_db_version" ] ; then
     db_versions_all_match="yes"
 else
     db_versions_all_match="no"
 fi
-if [ "$pom_derived_table_version" == "$cgds_db_derived_table_version" ] &&
-        [ "$cgds_db_derived_table_version" == "$migration_derived_table_version" ] ; then
+if [ "$pom_derived_table_version" == "$schema_sql_derived_table_version" ] &&
+        [ "$schema_sql_derived_table_version" == "$generate_derived_tables_sql_version" ] ; then
```
sheridancbio: So we are checking for code matching in the source files (consistent use of db_schema_version, consistent use of derived_tables_schema_version), which is good. We will need to be careful though, because nothing here is forcing us to update the derived_table_schema_version when we update the base tables (db_schema_version). But ... if our update to the base tables affects the construction of the derived tables, then that will need to be revised too.

Maybe we should always (as a rule) update the generate_derived_table.sql when we update schema.sql. (bumping both version numbers). Even if there is no semantic change in generate_derived_table.sql. We can discuss ... and again, this might be irrelevant soon if we make the derived tables part of the base table schema.

sheridancbio (changes requested):
Good work 👍 My only real request for a change is related to the conversion of environment variables holding credentials (password) into command line arguments passed to clickhouse client. I suggest using a temporary config file instead. We can discuss.

src/main/java/org/cbioportal/SchemaVersionChecker.java
dippindots: That's true, I will remove this

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
dippindots: This is a good point, let me think about this

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
dippindots: Will that actually be the case? I thought our migration script always updates the same record. Do you mean other instance maintainers might choose to add new records manually?

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -175,6 +195,15 @@ def get_current_db_schema_version(ch_props):
     return version
 
 
+def get_current_derived_table_schema_version(ch_props):
```
sheridancbio: This is okay ... but it means we will need to be careful to keep the info table up to date as we swap in new derived table generation script versions into our importing process. The script itself does not update the info table, so we will need to do do manually. Perhaps we could add a line into the derived table generation script which *does* alter the info table value for DERIVED_TABLE_SCHEMA_VERSION ... but hopefully this is not needed soon (when we engineer away from a separate phase for constructing derived tables), so it does not need to be added right now.

src/main/resources/db-scripts/clickhouse/generate_derived_tables.sql
```diff
@@ -711,3 +711,8 @@ OPTIMIZE TABLE genetic_alteration_derived;
 OPTIMIZE TABLE generic_assay_data_derived;
 OPTIMIZE TABLE generic_assay_profile_entity_derived;
 OPTIMIZE TABLE generic_assay_meta_derived;
+
+-- Record the version of this script in info.derived_table_schema_version, so callers (e.g.
+-- migrate_db.py --regenerate-derived-tables) can detect whether derived tables are stale by
+-- comparing against this script's own header version, regardless of who/what last ran it.
+ALTER TABLE info UPDATE derived_table_schema_version = '2.0.0' WHERE 1;
```
sheridancbio: ok .. this answers the comment above. We did add an update to info every time we run the generate_derived_tables.sql script. 👍

sheridancbio (approved):
👍

docs/deployment/clickhouse/README.md
```diff
@@ -263,18 +271,43 @@ After importing studies and rebuilding derived tables, you can verify that your
...
+`db-scripts/clickhouse/migrate/migrate_schema.sql` (a forward-only, version-tagged set of SQL
+sections) applied by `db-scripts/clickhouse/migrate/migrate_db.py`. The runner reads the current
+`db_schema_version` from the `info` table, skips sections already applied, and applies the rest in
+order. Derived table schema updates (tracked by `DERIVED_TABLE_SCHEMA_VERSION`) version
+independently, and are applied by rebuilding derived tables with `generate_derived_tables.sql`.
+
+**Version-bump convention:** any migration that changes base tables must also bump
+`generate_derived_tables.sql`'s version (even with no semantic change to that script), since a
+base-table change may affect derived tables in ways that aren't obvious to every contributor.
+`derived_table_schema_version` can also bump on its own, with no corresponding base-table
```
forus: Why do we have separate versions for derived tables in the first place? This approach is error-prone.

src/main/java/org/cbioportal/SchemaVersionChecker.java
forus, line 23: Why not simplify this code so it performs the same function as `migrate_db.py`, and then remove `migrate_db.py` entirely? The SQL scripts could be bundled with the release JAR, which would eliminate the extra Docker Compose complexity required to invoke `migrate_db.py`.

src/main/resources/db-scripts/clickhouse/migrate/migrate_schema.sql
```diff
@@ -0,0 +1,36 @@
...
+--
+-- Format: one section per db_schema_version, applied strictly in ascending order by
+-- migrate_db.py (see migrate_db.py in this directory for the runner).
+--
+--   ## db_schema_version: <version>
+--   ## description: <one line>
+--   ## custom: true            (optional — see below)
+--   <SQL statements for this section>
+--
+-- Rules for writing a section:
```
forus, line 10: Could we automate some of these rules instead of relying entirely on developer discipline?

e.g. `db_schema_version` could be updated automatically after the section completes successfully.

src/main/resources/db-scripts/clickhouse/migrate/migrate_schema.sql
```diff
@@ -0,0 +1,36 @@
...
+--      has no transactions, so a crash mid-section ideally leaves it safe to re-run from the top.
+--      Not always achievable; use judgment, and note in the description when a section isn't.
+--   2. Never DROP a column in the same section that reads from it.
+--   3. For ORDER BY / primary-key changes: create a new table, INSERT ... SELECT, RENAME —
+--      ClickHouse cannot ALTER these in place.
+--   4. Every section must end by advancing db_schema_version, e.g.:
+--        ALTER TABLE info UPDATE db_schema_version = '<version>' WHERE 1;
+--      migrate_db.py waits for this mutation (and any others the section triggered) to finish
+--      before treating the section as applied.
+--   5. Mark a section '## custom: true' when it needs logic beyond raw SQL (e.g. data reshaping
```
forus: Do we really need this additional complexity? Do we have a concrete example of a computation that must run in Python and cannot be handled with pure SQL?

src/main/resources/db-scripts/clickhouse/migrate/migrate_schema.sql
```diff
@@ -0,0 +1,36 @@
+-- ClickHouse base-table migration file.
+--
+-- Format: one section per db_schema_version, applied strictly in ascending order by
+-- migrate_db.py (see migrate_db.py in this directory for the runner).
+--
+--   ## db_schema_version: <version>
```
forus, line 6: Another approach I’ve seen is to use a separate file for each change set, for example: `1_base_tables.sql`, `2_add_user_table.sql`, and `3_drop_unused_columns_from_att_table.sql`, ...

src/main/resources/db-scripts/clickhouse/init/schema.sql
```diff
@@ -40,15 +40,24 @@
 -- specifications for tables. These are needed for good performance during
 -- cBioPortal web service functionality and data import functionality.
 --
+-- This includes the derived tables (denormalized tables that exist purely for query
+-- performance, named with a `_derived` suffix) alongside the base tables, so their structure is
+-- versioned under the same single db_schema_version as everything else. Derived tables are
+-- created here empty; db-scripts/clickhouse/populate_derived_tables.sql fills them in and is
+-- safe to re-run any time (see docs/deployment/clickhouse/README.md §7).
```
sheridancbio: well ... sort of. If you re-run the script which re-populates the derived tables on a database which is currently connected to an in-production backend web service, the web service will be unstable (generate errors) when queries to the database fail for empty or partially filled derived tables.

It is safe in terms of database management, but not safe in terms of deployment management. But I think we instruct users to take their web service deployments offline any time they are making database updates. So in that case it would be safe.

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -227,53 +190,32 @@ def apply_section(ch_props, section):
...
-        wait_for_mutations(ch_props)
-    if section.custom:
-        custom_fn = CUSTOM_MIGRATIONS.get(section.version)
-        if not custom_fn:
-            raise RuntimeError(
-                f"Section {section.version} is marked '## custom: true' but no function is "
-                f"registered in CUSTOM_MIGRATIONS")
-        custom_fn(ch_props)
+    run_multiquery(ch_props, f"ALTER TABLE info UPDATE db_schema_version = '{section.version}' WHERE 1;")
+    wait_for_mutations(ch_props)
```
sheridancbio, line 223: Side comment : I've come to understand this a little better recently. In Clickhouse, the system.mutations table reflects tracked changes to the database which fit criteria. Speaking generally:
- INSERT operations are not tracked here
- DELETE and UPDATE operations are usually tracked.
- CREATE TABLE and other data definition operations are usually not tracked here
So I *think* this approach works, because any asynchronous alterations (DELETE/UPDATE) statements in the migration script statements will be waited for (to complete) with this call, and any data definition statements or INSERT statements should (normally) execute synchronously (i.e. will complete before returning from a statement execution).

However, our wait for mutations to complete is only occurring *in between* each schema version update.

I see two blind spots here.

1. we may try to write a schema version update step which contains multiple SQL statements and it may be the case that an early statement includes an UPDATE to data (which would be asynchronous). Then a late statement in that block might assume that the UPDATE has completed, and use the results of the update in that late stage step. (perhaps as a where condition filter). But we are only checking for mutation completion after the entire migration step has been done. So ... with this approach we need to be on guard and look for such cases. If we attempt to do anything of this nature, the migration needs to be broken up into at least 2 schema increments --- where the early stage UPDATE is part of a defined schema migration (to a new schema version number), and the late stage "WHERE updated_thing =" statement belongs to a subsequent schema increment. The alternative to that would be to either use statement settings such as "mutations_sync = 2" or "lightweight_deletes_sync = 2", or add some kind of SQL statement in between the two which would wait for the system.mutations table to show no in-progress mutations.
2.  we may forget about this problem and write a schema migration block which explicitly creates asynchronous INSERT statements (by including "SETTINGS async_insert = 1") or other such tricks to operate asynchronously. Those processes will not be tracked in system.mutations but will be tracked elsewhere (like system.asynchronous_inserts), and those would be overlooked by our wait_for_mutations() call.

But we can wait to see whether either of these problems come up. It would be more difficult to track asynchronous inserts to see if there were any running. There is no "is_done" flag in system.asynchronous_inserts. First, a number of buffers are created as the insert is absorbed. Then these buffers get flushed. According to this, you can track pending asynchronous insert flushes through system.metrics table :
https://clickhouse.com/blog/monitoring-asynchronous-data-inserts-in-clickhouse

perhaps we should just agree to avoid any asynchronous operations aside from DELETE and UPDATE.

src/main/resources/db-scripts/clickhouse/migrate/migrate_db.py
```diff
@@ -227,53 +190,32 @@ def apply_section(ch_props, section):
...
 
-def maybe_regenerate_derived_tables(ch_props, derived_table_sql_filepath=None):
-    """Regenerate derived tables if generate_derived_tables.sql's own version differs from the
-    database's current info.derived_table_schema_version. Runs unconditionally as a blanket
-    policy whenever that's the case — covers both a derived-table rebuild required by a
-    base-table migration and a derived-table-only version bump with no corresponding
-    migrate_schema.sql section."""
-    filepath = derived_table_sql_filepath or DEFAULT_GENERATE_DERIVED_TABLES_SQL
+def populate_derived_tables(ch_props, populate_derived_tables_sql_filepath=None):
+    """Repopulate derived tables by running populate_derived_tables.sql. Safe to call any time —
```
sheridancbio: again .. "sort of"

src/main/resources/db-scripts/clickhouse/migrate/migrate_schema.sql
```diff
@@ -15,22 +14,16 @@
...
---      before treating the section as applied.
---   5. Mark a section '## custom: true' when it needs logic beyond raw SQL (e.g. data reshaping
---      that must be computed outside ClickHouse). migrate_db.py runs this section's SQL (if any)
---      first, then calls the matching hardcoded Python function registered in
---      migrate_db.py's CUSTOM_MIGRATIONS dict, keyed by this section's version.
---   6. Any section that changes base tables must also bump ../generate_derived_tables.sql's
---      version (even with no semantic change to that file), since a base-table change may affect
---      derived tables in ways that aren't obvious to every author. derived_table_schema_version
---      may also bump on its own with no corresponding section here.
+--   4. Don't write db_schema_version updates yourself — migrate_db.py advances
```
sheridancbio, line 17: maybe we can add a rule here regarding asynchronous operations:

"Don't use unusual asynchronous operations (such as adding "SETTINGS async_insert = 1").
Beware of any within-version dependencies on the results of DELETE or UPDATE (which are
usually asynchronous). For valid outcomes, dependencies may need to run synchronously
using settings such as "mutations_sync = 2".

src/main/resources/db-scripts/clickhouse/populate_derived_tables.sql
```diff
@@ -1,22 +1,34 @@
...
-DROP TABLE IF EXISTS generic_assay_data_derived;
-DROP TABLE IF EXISTS mutation_derived;
+-- Populates cBioPortal's derived tables (denormalized tables that exist purely for query
+-- performance — see docs/deployment/clickhouse/README.md §7).
+--
+-- This script does NOT define table structure — derived table DDL lives in
+-- db-scripts/clickhouse/init/schema.sql (and, when it changes, in a
+-- db-scripts/clickhouse/migrate/migrate_schema.sql section), versioned together with every
+-- other table under the single db_schema_version. This script only clears and repopulates data,
+-- and is safe to call repeatedly, any time (e.g. after importing a study, or as a manual
```
sheridancbio: any time that there is no connected backend web service in production

src/main/resources/db-scripts/clickhouse/populate_derived_tables.sql
```diff
@@ -1,22 +1,34 @@
...
+--
+-- Callers: metaImport.py's rebuild_derived_tables.py after every import, migrate_db.py
+-- --populate-derived-tables after a migration, docker-compose's fresh-install init scripts, or
+-- run directly (e.g. `metaImport.py derive-tables`).
 
 SET function_sleep_max_microseconds_per_block = 3600000000;
 
+-- Clear derived tables before repopulating — this script doesn't recreate table structure, so
+-- data needs to be explicitly cleared rather than relying on a DROP+CREATE to do it implicitly.
+TRUNCATE TABLE IF EXISTS sample_to_gene_panel_derived;
```
sheridancbio: TRUNCATE is basically a big DELETE statement. I think it would be a good idea to add statements to make these operations synchronous.

Example:

`TRUNCATE TABLE IF EXISTS gene_panel_to_gene_derived SETTINGS alter_sync = 2;`

src/main/resources/mappers/clickhouse/clinical_data/ClickhouseClinicalDataMapper.xml
```diff
@@ -317,7 +317,7 @@
                     #{studyId}
                 </foreach>
             </if>
-            AND <!-- Table creation in generate_derived_tables.sql has ensured no NA values but extra caution is always appreciated -->
+            AND <!-- Table creation in schema.sql has ensured no NA values but extra caution is always appreciated -->
```
sheridancbio, line 320: I think it is the population script queries which ensure the avoidance of NA ... not the schema.sql table defintions. (same comment for 7 additional cases below)

src/test/resources/schema.sql
```diff
@@ -47,8 +47,11 @@ DROP TABLE IF EXISTS authorities;
 DROP TABLE IF EXISTS cancer_study;
 DROP TABLE IF EXISTS cancer_study_tags;
 DROP TABLE IF EXISTS clinical_attribute_meta;
+DROP TABLE IF EXISTS clinical_data_derived;
```
sheridancbio, line 50: I wonder whether we could simply use the production schema definition directly. Do we really need a "test" copy of the schema definition?

src/test/resources/schema.sql
```diff
@@ -531,19 +677,17 @@ CREATE TABLE structural_variant (
...
 ) ENGINE = MergeTree ORDER BY (type_of_cancer_id);
-
 CREATE TABLE users (
     `email` String,
     `name` String,
     `enabled` Int32
 ) ENGINE = MergeTree ORDER BY (email);
 
-INSERT INTO info (`db_schema_version`, `geneset_version`, `derived_table_schema_version`, `gene_table_version`) VALUES ('2.14.5', 'msigdb_v2025.1.Hs', '1.0.9', 'hgnc_v7_2025.10.7');
+INSERT INTO info (`db_schema_version`, `geneset_version`, `gene_table_version`) VALUES ('2.14.5', 'msigdb_v2025.1.Hs', 'hgnc_v7_2025.10.7');
```
sheridancbio, line 754: Maybe this is why we need a test copy? Because tests a written against an older version of the database schema?  But ... wouldn't it be more proper to update our tests to run correctly against the latest schema whenever we update the schema?

sheridancbio (approved):
This is great work 👍 

I've made a few suggestions, and I hope we can have a follow-up conversation about having the tests run under an out-of-date version of the database schema (which seems like a problem to me).

Happy to walk through these comments with you if you want.

forus (approved):
@dippindots Looks good! Thank you!

PR #12288, opened 2026-08-10
Use Schema decorator to fix frontend API client TS issues

### LINKED ISSUE #12244: Fix DTOs and type generation
1. We should be cofnirmting to Charles architecture: always send back DTOs, not server-side objects.  It seems we didn't objserve this new rule
2. Use the schema decorators to normalize the names of the DTOs to frontend expectations
3. Make sure the DTOs match the old objects (as defined in typscript)

onursumer:
Fix #12244

- Replace server side objects with DTOs for new endpoints
- Use Schema decorator and add more mappers to fix frontend API client TS issues
- Update core and api tests to be compatible with the new implementation
- Clean up defunct mixins


src/main/java/org/cbioportal/legacy/web/mixin/DiscreteCopyNumberDataMixin.java
alisman, line 1: maybe we can delete this Mixins? @onursumer

alisman:
@onursumer remid me how the DTO suffix gets removed?

src/main/java/org/cbioportal/legacy/model/DiscreteCopyNumberData.java
```diff
@@ -20,11 +30,28 @@ public void setAlteration(Integer alteration) {
...
   }
 
-  public Object getAnnotationJson() {
+  public String getAnnotationJson() {
     return annotationJson;
   }
 
   public void setAnnotationJson(String annotationJson) {
     this.annotationJson = annotationJson;
   }
```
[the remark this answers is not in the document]
onursumer: Not a good idea to do anything else than a simple `set` operation in a `setter`. If really needed, this should be handled somewhere else.

src/main/java/org/cbioportal/legacy/model/GeneFilterQuery.java
```diff
@@ -40,34 +38,6 @@ public GeneFilterQuery(
...
-        if (!token.isEmpty()) {
-          alterations.add(CNA.valueOf(token.toUpperCase()));
-        }
-      }
-    }
-    GeneFilterQuery query = new GeneFilterQuery();
-    query.setHugoGeneSymbol(hugoGeneSymbol);
-    query.setAlterations(alterations.isEmpty() ? null : alterations);
-    return query;
-  }
```
onursumer, line 69: Looks like this moved to a separate class called `GeneFilterDeserializer` which seems better than keeping logic in the model class

PR #12305, opened 2026-08-19
Refactor SQL queries in SecurityMapper.xml

### LINKED ISSUE #12304: [BUG] ClickHouse user authorization lookup fails when OAuth2 email claim contains uppercase characters
**Describe the bug**

In cBioPortal v7.0.6 using OAuth2/OIDC authentication and ClickHouse, an enabled user cannot log in when the identity provider emits an email address containing uppercase characters.

The user exists in both `users` and `authorities`, and an exact ClickHouse query finds the user. However, `SecurityMapper.getPortalUser` returns zero rows.

This appears to be caused by ClickHouse resolving the `email` alias from the `SELECT` list inside the `WHERE` clause.

The current mapper query is:

```sql
SELECT
    LOWER(email) AS email,
    name,
    enabled
FROM users
WHERE email = #{username}
```

For an OAuth2 email such as `USER@example.org`, ClickHouse appears to evaluate the predicate as:

```sql
WHERE LOWER(email) = 'USER@example.org'
```

This can never match because the left side is lowercase while the parameter remains uppercase.

The same aliasing pattern exists in `getPortalUserAuthorities`.

**Environment**

- cBioPortal image: `cbioportal/cbioportal:7.0.6`
- Database: ClickHouse 24.10
- Authentication: OAuth2/OIDC
- Identity provider: Microsoft Entra ID / Azure AD
- `authorization=true`
- `authenticate=oauth2`

**To Reproduce**

1. Configure OAuth2 authentication with database authorization.
2. Have the identity provider emit an email claim containing uppercase characters, for example:

   ```text
   USER@example.org
   ```

3. Add an enabled matching user:

   ```sql
   INSERT INTO users (email, name, enabled)
   VALUES ('USER@example.org', 'Test User', 1);
   ```

4. Add an authority:

   ```sql
   INSERT INTO authorities (email, authority)
   VALUES ('USER@example.org', 'ALL');
   ```

5. Attempt to log in.

6. MyBatis reports an exact uppercase parameter but zero results:

   ```text
   Preparing: SELECT LOWER(email) AS email, name, enabled
              FROM users WHERE email = ?

   Parameters: USER@example.org(String)
   Total: 0

   User USER@example.org is null
   ```

7. An exact query without the conflicting alias returns the user:

   ```sql
   SELECT email, name, enabled
   FROM users
   WHERE email = 'USER@example.org';
   ```

8. Running the mapper query directly reproduces the zero-row result:

   ```sql
   SELECT LOWER(email) AS email, name, enabled
   FROM users
   WHERE email = 'USER@example.org';
   ```

**Expected behavior**

Email lookup should be case-insensitive, or the authenticated username should be normalized consistently before querying. An enabled user should be authorized regardless of capitalization in the IdP's email claim.

Email addresses inserted through the mapper are already normalized using `LOWER(#{email})`, which suggests lowercase/case-insensitive matching is intended.

**Suggested fix**

Qualify the source columns to avoid ClickHouse alias substitution and normalize both sides of the comparison:

```xml
<select id="getPortalUser" resultType="org.cbioportal.legacy.model.User">
    SELECT
        LOWER(u.email) AS email,
        u.name,
        u.enabled
    FROM users AS u
    WHERE LOWER(u.email) = LOWER(#{username})
</select>

<select id="getPortalUserAuthorities" resultMap="userAuthoritiesResultMap">
    SELECT
        LOWER(a.email) AS email,
        a.authority
    FROM authorities AS a
    WHERE LOWER(a.email) = LOWER(#{username})
</select>
```

saxenanurag:
Fixes #12304

### DIFF
--- a/src/main/resources/org/cbioportal/legacy/persistence/mybatis/SecurityMapper.xml
+++ b/src/main/resources/org/cbioportal/legacy/persistence/mybatis/SecurityMapper.xml
@@ -14,19 +14,19 @@
 
     <select id="getPortalUser" resultType="org.cbioportal.legacy.model.User">
         SELECT
-            LOWER(email) AS email,
-            name,
-            enabled
-        FROM users
-        WHERE email = #{username}
+            LOWER(u.email) AS email,
+            u.name,
+            u.enabled
+        FROM users as u
+        WHERE LOWER(u.email) = LOWER(#{username})
     </select>
 
     <select id="getPortalUserAuthorities" resultMap="userAuthoritiesResultMap">
         SELECT
-            LOWER(email) AS email,
-            authority
-        FROM authorities
-        WHERE email = #{username}
+            LOWER(a.email) AS email,
+            a.authority
+        FROM authorities as a
+        WHERE LOWER(a.email) = LOWER(#{username})
     </select>
 
     <insert id="addPortalUser">


alisman:
@saxenanurag did you spin this up and test it?

saxenanurag:
@alisman I changed the code and created a custom docker image locally. This is what I am running right now on my dev server.

alisman (approved):

PR #12310, opened 2026-08-25
ci: make the security badge link to the Scout report it was computed from

alisman:
## What

The README's Security badge showed pass/fail with no way to see the scan behind it — it linked to a generic docs page. This PR makes the badge link to the exact Docker Scout report its state was computed from:

- **`.circleci/config.yml`** — `update_security_status_badge` now also commits `master_report.sbom` to the `cbioportal-test` repo as `security-report.json`, in the **same commit** as `security-status.json`. Badge color and report are updated atomically, so the link can never show a different scan than the one that set the color.
- **`README.md`** — the badge's link target changes from the docs page to [`cbioportal-test/security-report.json`](https://github.com/cBioPortal/cbioportal-test/blob/main/security-report.json).
- Incidental: `git diff` → `git diff --stat` in the push step, so the report's long advisory descriptions don't flood the job log.

## Notes

- The report JSON contains no timestamps, so `cbioportal-test` only gets a new commit when the CVE set actually changes — no per-build churn.
- The linked file won't exist until the first master run of `update_security_status_badge` after this merges; the badge itself is unaffected in the meantime.
- No new credentials needed — the job already pushes to `cbioportal-test` via its existing deploy key.

https://claude.ai/code/session_011AdoR5KtVUuJg54XXTaD3H

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -801,12 +801,16 @@ jobs:
               mv temp.json security-status.json
               echo "security-status.json has been updated:"
               cat security-status.json | jq
+              # Publish the scan the badge was computed from alongside the badge
+              # state, so the README badge can link to the exact report. Committed
+              # in the same commit as security-status.json to keep them atomic.
+              cp ../master_report.sbom security-report.json
         - run:
             name: Push new security status to cbioportal/cbioportal-test
             command: |
               cd cbioportal-test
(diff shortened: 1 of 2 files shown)


alisman:
@zainasir this PR would make the badge link through to the actual report.  let me know what you think.

Issue #12317, opened 2026-08-27
Improve the import flow for single-ClickHouse-instance deployments


fuzhaoyuan:
## Summary

This applies to deployments running a single ClickHouse instance — the default/documented community setup, not blue-green.

In this setup, importing a study through `metaImport.py` can leave the live site serving broken or incomplete data in two distinct ways:

### 1. Studies unrelated to the import can break while derived tables rebuild

`db-scripts/clickhouse/populate_derived_tables.sql` `TRUNCATE`s every derived table before repopulating them. Derived tables aren't scoped per study, so rebuilding them to pick up one study's new or changed data means every *other* study's Study View can hit empty or partially-rebuilt tables and error out or render wrong data for the whole rebuild. This isn't limited to the study currently being imported; a completely unrelated, previously-stable study can start failing the moment any import triggers a rebuild.

### 2. A study can become visible before its own derived data exists

Separately: `cbioportalImporter.py`'s `process_study_directory()` unconditionally marks the study `AVAILABLE` as its last step, before the caller (`metaImport.py`) goes on to run the derived-table rebuild afterward. A newly imported study can therefore appear in listings before any derived data for it has been built at all.

This is currently blocking study-loading work on DFCI single-instance deployment, and the same gap should affect any other institution running a single-instance setup that needs to stay reachable to users during imports.

## Proposed fixes

Two independent, complementary changes:

- **cbioportal** https://github.com/cBioPortal/cbioportal/pull/12318: rebuild `populate_derived_tables.sql` to populate each derived table under a `_temp` copy (structure cloned via `CREATE TABLE ... AS ...`, so it always matches the live schema) and swap all 12 into place via `EXCHANGE TABLES` in one batch at the end. `EXCHANGE` is a metadata-only, near-instant swap, so live queries see the complete old table set throughout the rebuild and the complete new set right after — no empty/partial window. Tradeoff: derived-table disk usage roughly doubles during rebuild on a single instance, ~1.5x on blue-green; rebuild time and CPU/memory are essentially unchanged since the same queries run either way.

[...]

PR #12318, opened 2026-08-27
Rebuild derived tables via temp-table swap instead of TRUNCATE

fuzhaoyuan:
Addresses #12317

## Summary

`populate_derived_tables.sql` currently `TRUNCATE`s every derived table before repopulating them, so a live instance querying the database mid-rebuild sees empty or partially-rebuilt tables and errors out. This affects every study's Study View, not just the study being imported.

## What changed

- Each derived table is now rebuilt under a `<name>_temp` copy instead of truncating the live one in place. The `_temp` copy's structure is cloned live via `CREATE TABLE ... AS ...`, so it automatically tracks future `schema.sql`/`migrate_schema.sql` changes with no manual upkeep.
- All 12 tables are cut over via `EXCHANGE TABLES` in one batch at the end of the script, back-to-back with nothing else interleaved. `EXCHANGE` is a metadata-only, near-instant swap, so live queries see the complete old table set throughout the rebuild and the complete new set right after — instead of an empty or half-populated one for the whole rebuild.

## Tradeoffs

Derived-table disk usage roughly doubles during the rebuild on a single-instance deployment (old and new copies coexist until the cutover), or rises to about 1.5x the steady-state total on a blue-green deployment (only the standby database being rebuilt doubles, while production is unaffected). Rebuild time and other resources like CPU and memory are essentially unchanged, since the same `SELECT`/`INSERT` work runs either way.

## Not in scope

This only addresses derived-table rebuilds affecting studies unrelated to a given import. A separate cbioportal-core change (in progress) addresses a newly-imported study becoming visible before its own derived data is ready — see #12317 and fix in https://github.com/cBioPortal/cbioportal-core/pull/178

### DIFF
--- a/src/main/resources/db-scripts/clickhouse/populate_derived_tables.sql
+++ b/src/main/resources/db-scripts/clickhouse/populate_derived_tables.sql
@@ -6,9 +6,9 @@
 -- db-scripts/clickhouse/migrate/migrate_schema.sql section), versioned together with every
 -- other table under the single db_schema_version. This script only clears and repopulates data
 -- (e.g. after importing a study, or as a manual rebuild step) — it does not need its own version
--- number. Only run it while no backend web service is connected to this database in production:
--- it TRUNCATEs derived tables before repopulating them, so a live instance querying the database
--- mid-run will see empty or partially-rebuilt derived tables and surface errors.
+-- number. Each table is rebuilt under a "<name>_temp" copy and swapped in via `EXCHANGE TABLES`,
+-- so it can in theory run against a live production instance -- running with the backend web
+-- service disconnected is still recommended.
 --
 -- Callers: metaImport.py's rebuild_derived_tables.py after every import, migrate_db.py
 -- --populate-derived-tables after a migration, docker-compose's fresh-install init scripts, or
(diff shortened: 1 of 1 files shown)


sheridancbio:
In a quick read through this PR, my first understanding is that the main purpose is to shorten the period of time where the derived table content is in an unusable / inconsistent state. The population process (which can take up to 2 hours with the larger portal database) currently begins with emptying all of the tables. If the tables are used in a live database during this time, errors will very likely be seen by users. With the approach in this PR, the original content of the derived tables remain visible to the persistence layer until we approach the end of the population process. Then the new state is exchanged (table by table) into visibility with EXCHANGE TABLE statements. That phase should take less than a minute or two. So the period of inconsistency is greatly reduced. The alterations in these tables will become visible not in an atomic operation ... so there is still a period of data inconsistency (with possible errors or distortions). But @fuzhaoyuan intends to mask these errors by leveraging the status / "available" property of the cancer_study table in order to hide the studies which have any data changes as they work their way through the repopulation process. I have not yet looked at those intended code changes (in cbioportal-core). I wonder whether it means that for the period of re-populating the derived tables those studies will not be visible to the users. Perhaps for studies which are being newly imported that would not be an issue .. but for studies which are being updated they might "disappear" for a couple of hours while they are under update.

Another issue : api clients still might have access to studies which are not "available". So scripted interactions with the cBioPortal API endpoints may still generate database (sql) exceptions when retrieving these studies which are not categorized as available. This should be tested. A quick look at src/main/java/org/cbioportal/application/security/CancerStudyPermissionEvaluator.java seems to indicate that users have access to studies based on the roles of their user authentication object after they log in to the portal, without regard to whether a study is marked as available or not. So it is likely that users will see not-available studies in the api/studies endpoint and could query for study data using this and other api endpoints.

alisman:
@sheridancbio @fuzhaoyuan i just want to make sure i understand this.  this PR will allow studies which have not been updated (and are thus compatible with the "old" derived tables) to continue to function while derivation is taking place.  studies which HAVE been updated cannot be used because they are no longer compatible with the "old" derived tables.  is there indeed a mechanism to hide studies which are getting updated?

alisman:
@fuzhaoyuan lets discuss this at standup

fuzhaoyuan:
> @sheridancbio @fuzhaoyuan i just want to make sure i understand this. this PR will allow studies which have not been updated (and are thus compatible with the "old" derived tables) to continue to function while derivation is taking place. studies which HAVE been updated cannot be used because they are no longer compatible with the "old" derived tables. is there indeed a mechanism to hide studies which are getting updated?

Yes and yes the mechanism to hide is proposed in the other PR https://github.com/cBioPortal/cbioportal-core/pull/178 but I only did the "delete and reimport" part, not "update" part.

alisman:
@fuzhaoyuan just asked Claude to assess whether "status" is actually respected in the portal codebase and it looks like it is ignored entirely.  so we would need to put that gate in and probably in a lot of different places.  is that in the plan?  
[Here is link to Claude analysis](https://claude.ai/share/36be121c-90cf-4528-86aa-4a68bb3a907f)

PR #12321, opened 2026-08-28
Fix study_es_0 to be a standardized format that can work with the current pipelines importer JAR

jamesqo:
works with public reference data


dippindots (approved):
LGTM, do you want to merge this?

PR #12341, opened 2026-09-16
Make clinical data table pagination deterministic

raylim:
## Summary

The clinical data table now needs to browse beyond the first 500 samples in bounded 500-row blocks. This change makes those blocks stable and avoids rebuilding the complete matching sample-ID list for every page request.

- Return a separate total count and only the requested ordered sample-ID block for paged requests.
- Add deterministic ordering, including an internal sample-ID tie-breaker for every sort and a default order when no sort is selected.
- Return ordered sample keys and include selected samples with no clinical values as empty rows.
- Preserve the existing unpaged path used by downloads.
- Include `orderedSampleKeys` in the REST response DTO so the frontend can use deterministic server pagination.

This is the backend companion to cBioPortal/cbioportal-frontend#5712.

## Validation

- Added repository count coverage.
- Added service coverage for out-of-range pages preserving the total count.
- Added controller coverage for the serialized `orderedSampleKeys` response field.
- Existing clinical-table repository and service tests cover search, sorting, paging, and empty input behavior.
- Local directory-backed validation passes: `mvn -Dmaven.resources.skip=true compile`, `mvn -Dtest=StudyViewControllerTest test`, `mvn -Dtest=ClinicalDataServiceImplTest test`, and `mvn spotless:check`.
- CI image builds and security tests pass. The current API job timed out waiting for the container health check before tests ran; CI browser failures are unrelated screenshot regressions in ASCN, CN-segments, and survival views.


alisman:
@raylim is this ready for review?  we want to put this on staging.   btw, what prompted this PR?

Issue #11980, opened 2026-02-21
Feature: Add API rate limiting to prevent abuse and protect server resources


purvanshjoshi:
**Is your feature request related to a problem? Please describe.**

cBioPortal's REST API currently has **no rate limiting or request throttling** mechanism. Any client — authenticated or anonymous — can send an unlimited number of requests at any rate, creating denial-of-service risk and database overload.

This is not a theoretical concern. PR #11946 was recently merged to make the `entrezGeneId` parameter required because:

> *"This fixes a major performance bottleneck when API users requested mutations while incorrectly specifying the gene param, resulting in ALL mutations being sent."*

Even with that parameter fix, nothing prevents a client from making the same expensive request thousands of times in rapid succession. Heavy endpoints like `POST /api/column-store/clinical-data-counts/fetch` and `POST /api/column-store/clinical-data-bin-counts/fetch` trigger expensive ClickHouse aggregation queries with no protection.

After reviewing the codebase, I confirmed:
- No rate limiting filter exists in any `SecurityFilterChain` ([NoSecurityConfig](cci:2://file:///d:/cbioportal/src/main/java/org/cbioportal/application/security/config/NoSecurityConfig.java:11:0-23:1), [OAuth2SecurityConfig](cci:2://file:///d:/cbioportal/src/main/java/org/cbioportal/application/security/config/OAuth2SecurityConfig.java:33:0-119:1), [ApiSecurityConfig](cci:2://file:///d:/cbioportal/original_ApiSecurityConfig.java:24:0-86:1))
- No rate limiting library is present (no Bucket4j, Resilience4j, Guava RateLimiter, or similar)
- No `429 Too Many Requests` response is defined in [GlobalExceptionHandler](cci:2://file:///d:/cbioportal/src/main/java/org/cbioportal/application/rest/error/GlobalExceptionHandler.java:42:0-255:1) or any controller
- The application uses 200+ `@Cacheable` methods that can be bypassed by varying request parameters, causing cache misses that directly hit the database

**Describe the solution you'd like**

A layered rate limiting approach integrated into Spring Boot's security filter chain:

1. **Global rate limiting filter** — A `RateLimitFilter extends OncePerRequestFilter` using token bucket algorithm (e.g., [Bucket4j](https://github.com/bucket4j/bucket4j)) for per-client rate limiting
2. **Client identification** — By IP address (unauthenticated) or API token (authenticated)
[...]

PR #12346, opened 2026-09-18
Add configurable API rate limiting

1sakshm:
## Summary

Adds an opt-in, in-memory token-bucket limiter for REST API and legacy webservice requests.

- limits each remote IP address using configurable sustained and burst limits
- returns `429 Too Many Requests` with `Retry-After`
- keeps bucket state in the domain layer and filter wiring in application security
- documents `rate-limit.*` configuration in `application.properties.EXAMPLE`

## Validation

- `mvn -Dtest=RateLimitServiceTest,RateLimitFilterTest test`
- `mvn spotless:check`
- `mvn integration-test` ran 1,418 tests; three existing `StudyControllerTest` import-date assertions failed because this environment formats their timestamp 5½ hours later.

Addresses #11980.


1sakshm:
@dippindots Could you please add the feature label to this draft PR, or point me to the right maintainer? The label-check workflow requires it.

Pradyuman-aviator (changes requested):
Nice work ,  the architecture is clean and the Clock injection 
for testability is a smart design choice.

A few things worth discussing:

1. **Memory leak in `buckets`**: The ConcurrentHashMap grows 
   unboundedly  every unique IP ever seen adds an entry that 
   is never evicted. Under real traffic this will grow without 
   bound. Consider a Caffeine cache with a TTL/max-size policy.

2. **Proxy-unaware IP detection**: `getRemoteAddr()` returns 
   the load balancer's IP behind a reverse proxy. Worth checking 
   `X-Forwarded-For` (with appropriate trust validation) first.

3. **Shared "unknown" bucket**: All requests with a blank 
   remote address share one rate-limit bucket, which could let 
   one client starve others unfairly.

1sakshm:
@Pradyuman-aviator Thank you for the thoughtful review and suggestions. I addressed the points in `0fe91c745`.

- Replaced the unbounded `ConcurrentHashMap` with a Caffeine cache capped at 10,000 buckets and expiring idle entries after one hour.
- Requests without a usable remote address now bypass rate limiting rather than sharing an `unknown` bucket.
- The limiter continues to use `getRemoteAddr()` only; it does not trust `X-Forwarded-For` without an explicit trusted-proxy design.

Pradyuman-aviator (approved):
LGTM 👍

alisman:
@1sakshm we've decided not to do rate limiting, at least not at the application level.  thank you for your time on this and we apologize

Issue #12332, opened 2026-09-04
[FEATURE REQUEST] :   Patient Genomic Similarity Search API endpoint


Pradyuman-aviator:
## Summary

Add a new API endpoint that finds the N most genomically similar patients 
to a reference patient within a study, ranked by Jaccard similarity over 
a user specified gene set.

## Motivation

Currently, researchers who want to find patients with similar mutation 
profiles must:
1. Export mutation data from cBioPortal
2. Build binary alteration matrices locally in R or Python
3. Compute pairwise similarity manually

This is a common workflow - oncologists frequently ask "have you seen a 
patient like mine before?" ,  but cBioPortal provides no API-level support 
for it. MatchMiner handles trial matching but not patient-to-patient 
genomic similarity.

## Proposed API

```
POST /api/studies/{studyId}/patient-genomic-similarity
```

**Request:**
```json
{
  "molecularProfileId": "brca_tcga_mutations",
  "referencePatientId": "TCGA-A2-A0T2",
  "hugoGeneSymbols": ["TP53", "PIK3CA", "BRCA1", "CDH1"],
  "topN": 10
}
```

**Response:**
```json
[
  {
    "patientId": "TCGA-B6-A0RS",
    "similarityScore": 0.75,
    "commonMutatedGenes": ["BRCA1", "CDH1", "TP53"]
  }
]
```

## Algorithm

Jaccard similarity over the queried gene set:

```
Jaccard(A, B) = |mutated_genes(A) ∩ mutated_genes(B)|
              / |mutated_genes(A) ∪ mutated_genes(B)|
```

Score 1.0 = identical profiles. Score 0.0 = nothing in common.

## Implementation note

@onursumer  , would love your thoughts on this before I submit a PR. 
I have a working implementation ready (Jaccard similarity over 
mutation profiles, 8 unit tests). Happy to open the PR immediately 
if this direction makes sense to you.

## Future extensions

- Include CNA and structural variant alterations (not just mutations)
- Cross-study search
- Cosine similarity as an alternative metric

Pradyuman-aviator:
## Future extensions

[...]

PR #12348, opened 2026-09-18
perf: add @Transactional(readOnly=true) to all domain read use cases

Pradyuman-aviator:
## Summary

All `Get*UseCase` classes in the `org.cbioportal.domain` layer perform 
read-only database operations but lacked `@Transactional(readOnly = true)`. 

Adding this annotation allows:
- **InnoDB** to skip gap-lock and shared-lock acquisition on reads
- **Spring** to set the Session flush mode to `NEVER`, skipping dirty-checking overhead
- **Connection pools** to route to a read replica if one is configured

## Scope

20 use cases across 10 domain packages:
`alteration`, `cancerstudy`, `clinical_attributes`, `clinical_data`, 
`clinical_event`, `generic_assay`, `genomic_data`, `patient`, `sample`, 
`treatment`

`GetPatientGenomicSimilarityUseCase` already carried this annotation 
(#12332). This PR applies the same pattern consistently across the rest 
of the domain layer.

## Changes

20 files  2 lines each (1 import + 1 annotation). No logic changes.

## Tests

`GetClinicalDataForXyPlotUseCaseTest` + `GetCancerStudyMetadataUseCaseTest`: 
**9/9 passing ✅**


dippindots (commented):
@Pradyuman-aviator could you elaborate on why we still need to care about InnoDB, given that we've fully migrated from MySQL to ClickHouse?

Pradyuman-aviator:
After looking more carefully at the repository layer 
the domain use cases delegte to MyBatis  & ClickHouse 
JDBC implementations, not Hibernate/JPA. That means the 
flush mode optimization from @Transactional(readOnly = true) 
doesn't actually apply here, and ClickHouse's read model 
makes the locking concern irrelevant too.

The annotation would only serve as documentation of intent, 
which isn't strong enough justification on its own. Closing 
this PR   thanks for the sharp question @dippindots

PR #12349, opened 2026-09-18
Block reads of unavailable studies with HTTP 423

fuzhaoyuan:
Related: #12317, cBioPortal/cbioportal-core#178, cBioPortal/cbioportal-frontend#5715

## Summary

`cancer_study.status` is set by the importer but never read by the portal. While a study is being (re)imported, its base and derived tables can be inconsistent, and the portal still serves it.

This PR makes the portal respect that flag. Reading a study whose status is not AVAILABLE returns **423 Locked**, so the frontend can tell "this study is being updated" apart from "you have no access".

## Behavior

- If the user has access and the study is unavailable, the response is 423 with `Study <id> is being updated. Please check back later.`
- If the user has no access, the response is 403, same as before. Availability is not revealed.
- If a request covers several studies, the access check runs on all of them first, so a study the user cannot access still produces a 403.
- The study list and `LIST` access are unchanged. Unavailable studies stay listed, with `readPermission=false`.
- Status is read with a lightweight query cached for 2s. It does not come from `CacheMapUtil`, which is only built at startup. If the query fails, the previous result is kept. Before the first successful query, reads are allowed.

## Flow

```
 API request
     │  @PreAuthorize hasPermission(id(s), type, READ)
     ▼
 CancerStudyPermissionEvaluator ── group check (unchanged) ── fails ────────────► 403
     │ passes
     ▼
 UnavailableCancerStudyIds.isUnavailable(id)
     │   refreshes every 2s via StudyMapper.getUnavailableStudyIds()  (StudyMapper.java / .xml)
     ├── available ─────────────────────────────────────────────────────────────► 200
     │ unavailable
     ▼
 sets request attribute UNAVAILABLE_STUDY_ATTRIBUTE, returns false
     │  AccessDeniedException
     ▼
 GlobalExceptionHandler ── attribute set? ── yes ───────────────────────────────► 423
                                          └─ no ────────────────────────────────► 403

 MethodSecurityConfig: passes UnavailableCancerStudyIds to the evaluator
```

## Testing

- Unit tests for the evaluator, the 2s cache, and the exception handler.
- Tested manually on a local ClickHouse + Keycloak setup with the frontend PR. Changing `status` in the database takes effect within about 2s.

## Known limitations / follow-ups

- The importer still has to mark a study UNAVAILABLE at the start of every import, and `metaImport.py` should only enable it after the derived tables are rebuilt. See cBioPortal/cbioportal-core#178.
- Results view shows the 423 message on its full-page error screen instead of a banner. Multi-study queries that include an unavailable study fail rather than leaving that study out.


src/main/java/org/cbioportal/application/security/CancerStudyPermissionEvaluator.java
```diff
@@ -185,6 +204,10 @@ public boolean hasPermission(
     if (log.isDebugEnabled()) {
       log.debug("hasPermission(), checking permissions on targetId");
     }
+    if (!userAuthorizationActive) {
```
alisman: @fuzhaoyuan will look into why we need to check whether auth is active given that availability check should be the same either way

PR #12351, opened 2026-09-19
Fix study export stable ID prefix removal

### LINKED ISSUE #12350: [Bug] Study export removes repeated study identifiers from profile and case-list IDs
## Summary

The study-export metadata helpers remove **all** occurrences of `<cancerStudyIdentifier>_` from a stable ID. Their documented contract is to remove the cancer-study identifier from the beginning of the ID.

For a study ID that is also present later in the profile or case-list identifier, the generated metadata stable ID and case-list filename are corrupted.

## Reproduction

Given `cancerStudyIdentifier = "demo"` and `stableId = "demo_demo_mutations"`:

```java
stableId.replace(cancerStudyIdentifier + "_", "")
```

returns `"mutations"`. The expected exported stable ID is `"demo_mutations"`: only the leading `"demo_"` belongs to the study prefix.

The same behavior occurs for case-list IDs, for example `demo_demo_sequenced` is exported as `sequenced` instead of `demo_sequenced`.

## Affected code

- `src/main/java/org/cbioportal/application/file/model/GeneticProfileDatatypeMetadata.java`
  - `getGeneticDatatypeStableId()` feeds exported `meta_*.txt` files and data filenames.
- `src/main/java/org/cbioportal/application/file/model/CaseListMetadata.java`
  - `getCaseListTypeStableId()` feeds `case_lists/cases_*.txt` filenames and metadata.

Both classes were introduced by RFC95 study-data-file export in #11482.

## Expected fix

Strip the prefix only when `stableId.startsWith(cancerStudyIdentifier + "_")`, using `substring(prefix.length())`; otherwise preserve the original ID. Add focused unit tests covering a repeated study identifier in genetic-profile and case-list IDs.

## Scope

This is a current `master` export-path bug. It is self-contained Java logic and can be verified with unit tests; no database migration, SQL change, or endpoint authorization change is involved.

1sakshm:
## Problem

RFC95 study-data-file export removes every occurrence of `<studyId>_` from genetic-profile and case-list stable IDs. For example, with study ID `study`, `study_study_mutations` became `mutations` rather than `study_mutations`.

## Fix

Remove only a leading study-ID prefix in both export metadata helpers, preserving any later occurrence. Add regression coverage for genetic-profile and case-list IDs.

Fixes #12350

## Validation

- `mvn -Dmaven.gitcommitid.skip=true -Dtest=ToMetadataKeyValuesTest test` — **8 tests passed**.
- Confirmed the two new cases fail against the prior implementation.
- `git diff --check` passed.

The local Maven command skips the repository Git-metadata plugin because it cannot resolve `HEAD` from this linked worktree; the compiled production and test sources ran normally.


alisman:
@pieterlukasse @forus  can you comment on this one?

